PR #123, opened 2026-05-12
fix(engine): search commit subjects only for landed commit recovery

bluk1020:
Fixes false matches in auto-merge recovery where git log --grep searched both subject and body, causing commits that merely mention a task ID in their body to be incorrectly attributed.

Example: FN-002 was auto-recovered against d942c0c40 ('fix(engine): rescue auto-merge handoff...') which only mentions FN-002 in its body as a task affected by a bug — the commit had nothing to do with FN-002's actual implementation.

The fix pipes git log --format='%H%x1f%s' through grep -F to search only subject lines.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved commit search in auto-merge recovery to target subject lines only, reducing false matches from identifiers appearing in commit bodies.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Runfusion/Fusion/pull/123)


gsxdsm:
Can you update this PR with just the affected changes/commits?

gsxdsm:
Thanks for catching this @bluk1020 — the false-attribution bug you describe is real, and the fix you proposed (`git log --format=%H%x1f%s` piped through `grep -F` to match subjects only) is exactly the approach we landed on `main`.

The same technique is now in `packages/engine/src/merger.ts:676-677`:

```ts
["log", "--format=%H%x1f%s", "--max-count=20", "--fixed-strings", `--grep=${trailer}`, "HEAD"],
["log", "--format=%H%x1f%s", "--max-count=20", "--fixed-strings", `--grep=${task.id}`, "HEAD"],
```

`--format=%H%x1f%s` restricts the grep target to SHA + subject (no body), and `--fixed-strings` prevents regex misinterpretation of task IDs — covering both the case you raised and a related false-positive class.

I'm going to close this PR for two reasons:
1. The behavior change is already on `main`, so merging would be a no-op for the core fix.
2. This branch has diverged from `main` significantly (~1M lines / 100 files in the diff, with merge conflicts that block `gh pr update-branch`). Rebasing against current `main` would be more work than re-cutting a fresh branch.

If you spot a case the current implementation still misses, please open a fresh PR from an up-to-date `main` and we'll take it straight away. Really appreciate you surfacing the original problem.

PR #975, opened 2026-05-22
fix(dashboard): paginate all messages in useQuickChat (was truncated at 50)

titosemi:
## Problem

`useQuickChat` (the QuickChat FAB) loaded messages with a hardcoded `limit: 50` at all five fetch sites and never paginated further. Sessions with more than 50 messages appeared silently cut off — the visible history ended at message 50.

Note: the API endpoint already caps responses at 200 per request (`effectiveLimit = Math.min(limit, 200)`), so even requesting more requires pagination.

## Root cause

Five `fetchChatMessages(sessionId, { limit: 50 }, projectId)` calls with no pagination loop or load-more trigger.

## Fix

Introduce a `fetchAllMessages(sessionId, projectId)` helper that paginates in 200-message chunks until the full history is loaded, and replace all five call sites.

## Scope

`useChat` (full ChatView) is **not affected** — it already implements `loadMoreMessages` forward pagination triggered on scroll to bottom. This fix is scoped to `useQuickChat` only.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed QuickChat message history truncation so conversations load their full history (no silent 50-message cutoff). Refreshes, stream retries, and recovery paths now reliably apply up-to-date messages and avoid showing stale content after session changes.

* **Documentation**
  * Added a release note documenting the QuickChat history fix.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Runfusion/Fusion/pull/975?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.changeset/fix-quick-chat-message-truncation.md
+++ b/.changeset/fix-quick-chat-message-truncation.md
@@ -1,0 +1,21 @@
+---
+"@runfusion/fusion": patch
+---
+
+Fix `useQuickChat` silently truncating sessions longer than 50 messages.
+
+`useQuickChat` (QuickChat FAB) loaded messages with a hardcoded `limit: 50`
+at all five fetch sites. Sessions with more than 50 messages appeared cut off
+in the chat view — the visible history ended at message 50 regardless of how
+many messages existed.
+
+Root cause: the API endpoint caps responses at 200 messages per request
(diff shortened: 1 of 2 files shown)


titosemi:
Superseded by #978 which covers both useQuickChat and useChat in a single PR.

PR #978, opened 2026-05-22
fix(dashboard): wire scroll-to-top pagination for chat history instead of loading all messages

titosemi:
## Problem

Sessions with more than 50 messages showed truncated history with no way to load older messages. `loadMoreMessages()` was implemented in `useChat` but never called from `ChatView.tsx` — it was dead code.

## Root Cause

The original design was correct:
- Load the latest 50 messages on session open
- Scroll up → load 50 more older messages via `loadMoreMessages()`
- `hasMoreMessages` flag signals when all history is loaded

But `ChatView.tsx` never wired up the scroll trigger.

## Fix

Add an `IntersectionObserver` sentinel `<div>` at the **top** of the message list in `ChatView.tsx`. When the user scrolls up and the sentinel becomes visible, `loadMoreMessages()` is called automatically. Older messages are prepended and scroll position is preserved by the existing `scrollRestoreSnapshotRef` logic.

This replaces the previous approach (which loaded *all* messages at once on open) with the intended incremental design.

## Changes

- `packages/dashboard/app/components/ChatView.tsx` — add `loadMoreSentinelRef`, `IntersectionObserver` effect, and sentinel `<div>` at top of messages
- `packages/dashboard/app/hooks/useChat.ts` — revert `fetchAllMessagesInChat` helper; keep `limit: 50` initial load and stale-session guards
- `packages/dashboard/app/hooks/useQuickChat.ts` — revert `fetchAllMessages` helper; keep `limit: 50` initial load and stale-session guards

## Tests

- Updated test config to fix CI (`environmentMatchGlobs` → `test.projects`)
- Existing assertions unchanged (still `limit: 50`)

## Estimated code review effort

🎯 2 (Minor) | ⏱️ ~10 minutes


titosemi:
@coderabbitai review

gsxdsm:
Thank you !

titosemi:
Hi! I've pushed 3 new commits to address the review feedback:

1. **`fix(test): migrate fusion-plugin-reports vitest config from deprecated environmentMatchGlobs to test.projects`** — Fixes the `ReferenceError: window is not defined` CI failures (same root cause as #974).

2. **`fix(test): address review — inherit coreSetup from root via extends: true`** — Addresses CodeRabbit's concern: removed `coreSetup` from `reports-dashboard` project's `setupFiles` since it's already inherited from root via `extends: true`.

3. **`fix(dashboard): guard pagination state writes against stale-session race`** — Addresses the CodeRabbit major issue: added stale-session guard to the `isPaginationRequest` branch in `loadMessages`. If the user switches sessions during pagination, old-session results no longer overwrite the newly active session's state.

Would appreciate CI re-approval to confirm all checks pass.

gsxdsm (dismissed):

gsxdsm:
Hi @titosemi — thanks for fixing the chat message truncation, this is a real bug and the pagination approach is the right call.

Now that the systemic CI issues are fixed on `main` and your branch is rebased, **6 test failures** in `dashboard-app-quality` shard 2 are reproducible against this PR. They're all in tests that asserted the old single-page fetch behavior — none in your new helper code itself.

## The 6 failures

### Group A — existing tests assert old `fetchChatMessages` signature (4)

These tests expect calls like `fetchChatMessages(sessionId, { limit: 50, ... }, projectId)` but now receive only `(sessionId, ...)` since you swapped the initial-load call sites to use `fetchAllMessages()`:

1. `app/hooks/__tests__/useQuickChat.test.ts` > `switchSession with the same target reloads messages instead of creating a new session`
2. `app/hooks/__tests__/useQuickChat.test.ts` > `switchSession preserves isStreaming when same session is resumed`
3. `app/hooks/__tests__/useQuickChat.test.ts` > `onError reloads messages from server`
4. `app/components/__tests__/QuickChatFAB.test.tsx` > `FN-4660: opens/closes session menu and shows rooms before sessions when enabled`

For each: update the expectation to match the new pagination shape. If `fetchAllMessages` calls `fetchChatMessages` internally with a different signature (no `limit`, or paginating with `before`/`after`), spy on `fetchAllMessages` directly or assert on the internal loop's first call.

### Group B — session selection regression in dropdown (2)

5. `app/hooks/__tests__/useQuickChat.test.ts` > `startFreshSession creates a second session for the same model target`
   - Expected: `"session-fresh"`
   - Received: `"session-existing"`

6. `app/components/__tests__/QuickChatFAB.test.tsx` > `switches existing sessions from dropdown without creating new session`
   - Expected: `"session-agent"`
   - Received: `"session-model"`

These look different — the WRONG session ID is being selected. Likely the new stale-session guard (`activeSessionRef.current?.id` comparison) is short-circuiting before the session swap completes, so the prior session's messages are loaded against the new session ID. Worth tracing through `switchSession` → `fetchAllMessages` → `setMessages` and checking the ordering of `activeSession` updates vs the captured `sessionId` in the helper closure.

## Suggested fix path

```bash
pnpm --filter @fusion/dashboard exec vitest run \
  app/hooks/__tests__/useQuickChat.test.ts \
  app/components/__tests__/QuickChatFAB.test.tsx \
  --project=dashboard-app-quality
```

1. Fix Group B first — it's a real regression in selection logic, not just a test signature drift
2. Then update Group A's spy expectations to match the new pagination call shape

The truncation fix is real and the code looks careful. Once these 6 are green I'll merge. Happy to pair if helpful.

gsxdsm (approved):

PR #1425, opened 2026-06-05
Fix opencode-go model sync — pass API key to CLI and strip prefix from model IDs

tomdurrant:
## Problem

Two bugs when using OpenCode Go as a provider in Fusion:

1. **Model discovery only returns free models** — Fusion saves the Go API key in its auth storage but never passes it to the `opencode` CLI when running `opencode models opencode --refresh`. The CLI's internal `OpencodePlugin` checks `process.env.OPENCODE_API_KEY` and, when absent, disables all paid models (those with `cost.input > 0`). Only the 20 free models appear instead of all 67.

2. **Model IDs double-prefixed, API returns 401** — `normalizeOpencodeGoModel` was producing IDs like `opencode-go/deepseek-v4-flash`. The Pi SDK sends `model.id` verbatim in API requests, so the OpenCode API received a prefixed model name it didn't recognize → `401 Model opencode-go/deepseek-v4-flash is not supported`. The bare model name (e.g. `deepseek-v4-flash`) is what the API expects.

### `packages/cli/src/commands/startup-model-sync.ts`

- `discoverOpencodeGoModels(apiKey?)` — accepts optional API key and passes it as `OPENCODE_API_KEY` in the spawned `opencode` CLI process env
- `refreshOpencodeGoModels` — added optional `apiKey` field, forwards to discovery
- `syncStartupModels` — reads key from `authStorage` and passes it through
- `normalizeOpencodeGoModel` — strips the `opencode/` or `opencode-go/` prefix from model IDs, registering bare model names (e.g. `deepseek-v4-flash`) instead of prefixed ones

### `packages/cli/src/commands/serve.ts`, `daemon.ts`, `dashboard.ts`

- Each `onApiKeySaved` handler reads the saved key from `dashboardAuthStorage` and passes it to `refreshOpencodeGoModels`

## Note

After this change, users must re-select their opencode-go model in Settings because the model IDs have changed from prefixed (e.g. `opencode-go/deepseek-v4-flash`) to bare names (e.g. `deepseek-v4-flash`).

## Summary by CodeRabbit

* **Bug Fixes**
  * OpenCode API key is now correctly applied when fetching models, ensuring paid models appear.
  * Model identifiers are normalized to bare names (prefixes removed) and duplicates are deduplicated.
  * Users may need to re-select OpenCode models in Settings due to ID normalization.

* **Tests**
  * Added tests for normalization, deduplication, and API-key propagation.

### DIFF
--- a/.changeset/fix-opencode-go-api-key-env.md
+++ b/.changeset/fix-opencode-go-api-key-env.md
@@ -1,0 +1,15 @@
+---
+"@runfusion/fusion": patch
+---
+
+Fix opencode-go model sync: pass API key to CLI and strip provider prefix from model IDs
+
+Two bugs when using OpenCode Go as a provider:
+
+1. **Model discovery only returned free models** — the saved Go API key was never passed as `OPENCODE_API_KEY` to the spawned `opencode models opencode --refresh` process. The CLI's internal plugin checks this env var and, when absent, disables all paid models (those with `cost.input > 0`). Only 20 free models appeared instead of all 67.
+
+2. **API requests failed with 401** — `normalizeOpencodeGoModel` was registering models with prefixed IDs like `opencode-go/deepseek-v4-flash`. The Pi SDK sends `model.id` verbatim in API requests; the OpenCode API expects bare model names (e.g. `deepseek-v4-flash`). The prefix is now stripped during normalization.
+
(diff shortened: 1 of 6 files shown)


gsxdsm:
Thanks for these fixes!

PR #1671, opened 2026-06-13
fix(ce): recover stale active sessions

plarson:
## Summary
- Recover stale Compound Engineering `active` / `launching` session rows on plugin load.
- Re-run the same recovery before session read/list routes so a dashboard poll cannot keep showing zombie work as active.
- Add regression coverage for a stale active Strategy session with no live route handle.

## Symptom Verification
Original symptom: after reboot/restart, the Compound Engineering view showed an `active` Strategy session, but no agent was running, no `liveActivity` was present, and no artifact was produced.

Exact reproduction: seed a CE session row with `status=active`, `currentQuestion=null`, and a `lastActivityAt` far older than its turn interval, then call `GET /sessions` as the dashboard does.

Assertion it is gone: the route now returns the session as `interrupted` with the preserved-progress recovery message instead of leaving it `active`.

## Surface Enumeration
- Plugin load path: recovers stale rows after schema/skill setup.
- Session read path: `GET /sessions/:id` recovers stale rows before returning a single session.
- Session list path: `GET /sessions` recovers stale rows before returning dashboard-visible sessions.
- Human-wait state remains excluded by the existing `awaiting_input` stale-session tests.

## Test Plan
- `vitest run src/__tests__/session-routes.test.ts --silent=passed-only --reporter=dot`
- `vitest run src/__tests__/session-routes.test.ts src/__tests__/session-store.test.ts src/__tests__/setup-invariant.test.ts --silent=passed-only --reporter=dot`
- `vitest run --silent=passed-only --reporter=dot` in `plugins/fusion-plugin-compound-engineering`
- `npm run build` in `plugins/fusion-plugin-compound-engineering`

## Summary by CodeRabbit

* **Bug Fixes**
  * Stale sessions marked "active" without live handlers are now recovered on plugin load and during session reads, preventing the dashboard from getting stuck waiting on inactive work.

* **Tests**
  * Added test coverage for stale-session recovery for session listing and single-session reads.

* **Chores**
  * Release metadata added for a patch release.

### DIFF
--- a/.changeset/ce-recover-stale-sessions.md
+++ b/.changeset/ce-recover-stale-sessions.md
@@ -1,0 +1,5 @@
+---
+"@runfusion/fusion": patch
+---
+
+Recover stale Compound Engineering sessions on plugin load and session reads so persisted active rows without live agent handles no longer leave the dashboard stuck waiting for work that is not running.
--- a/plugins/fusion-plugin-compound-engineering/src/__tests__/session-routes.test.ts
+++ b/plugins/fusion-plugin-compound-engineering/src/__tests__/session-routes.test.ts
@@ -117,6 +117,53 @@ describe("session routes (polling transport)", () => {
     expect(sessions.map((s) => s.stage).sort()).toEqual(["brainstorm", "plan"]);
   });
 
+  it("GET /sessions recovers stale active rows that have no live route handle", async () => {
+    const { getCeSessionStore } = await import("../session/session-store.js");
+    const store = getCeSessionStore(h.ctx);
+    const zombie = store.create({ stage: "strategy", turnIntervalMs: 1 });
+    store.update(zombie.id, {
+      status: "active",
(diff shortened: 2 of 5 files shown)


plarson:
Review follow-up:

- Moved plugin-load stale-session recovery outside the bundled skill-install `try` so load-time recovery still runs if skill setup fails.
- Centralized recovery/logging/event emission in `session-recovery.ts` to avoid drift between plugin load and route paths.
- Added `GET /sessions/:id` regression coverage in addition to the existing `GET /sessions` coverage.
- Made route-triggered recovery best-effort: errors are logged and normal session reads continue rather than surfacing a framework 500.
- Added a 2-minute TTL guard around route-triggered recovery to avoid full-table scans on every dashboard polling request.

All review threads are resolved. Latest checks are green: Lint, Typecheck, Build, Gate, Greptile Review, and CodeRabbit.

PR #1692, opened 2026-06-20
Reduce SQLite read contention: bound WAL, release vacuum lock, offload integrity check

gsxdsm:
## What

Investigation of reported database **read contention** found, and this PR fixes, three sources of cross-process and event-loop contention in the SQLite layer (single synchronous `node:sqlite` connection per process, WAL mode). Not classic row-lock contention — WAL already allows concurrent readers — but three real ways readers get starved.

### 1. Unbounded WAL on the shared DBs (`central-db.ts`, `archive-db.ts`)
Neither set `journal_size_limit`, so their WAL never truncated back down after a checkpoint and every reader paid an ever-growing WAL-index scan. `central-db` (`fusion-central.db`) is shared across **all projects and cluster nodes** — the most cross-process-contended DB — yet was the least tuned; `archive-db` had the same latent gap.

- Add `journal_size_limit = 4194304` (4 MB), matching the per-project `db.ts`.
- Set `synchronous = FULL` / `wal_autocheckpoint = 1000` explicitly for intent. **These are already SQLite's defaults** (verified: `node:sqlite` keeps `synchronous=FULL` under WAL) — no behavior change.

### 2. `vacuum()` held the EXCLUSIVE lock past its own runtime (`db.ts`)
Resetting `locking_mode` to `NORMAL` does **not** drop the WAL exclusive lock until the connection next touches the DB — so every other process stayed locked out of reads (`SQLITE_BUSY`) until some unrelated query ran. A plain `SELECT` does **not** release it in WAL mode (verified empirically); a checkpoint does.

- Run a `PASSIVE` checkpoint in the `finally` to force the release immediately.
- Guard the `locking_mode=NORMAL` reset so a throw there can't mask the original error or skip the release; log previously-silent checkpoint failures.

### 3. Background `integrity_check` froze the event loop for seconds (`db.ts`)
`scheduleBackgroundIntegrityCheck` (~60s after init) ran `PRAGMA integrity_check` on the live connection, walking every page and blocking the event loop — the largest single stall in normal operation.

- Offload it to the `sqlite3` CLI in a child process (async `spawn`), matching the existing out-of-process pattern (`quickCheckSqliteFile`, `.recover`).
- Open the CLI connection `-readonly` so it can never checkpoint or write the live WAL (works because the live process holds the DB open → `-shm` exists).
- Fall back to the in-process check when the CLI is unavailable / can't open read-only (`verified=false`) — same behavior as today on those environments.
- New `runBackgroundIntegrityCheck()` seam centralizes the offload+fallback policy and gives the scheduler one deterministic, testable point.

**VACUUM is intentionally not offloaded:** the call graph shows it runs only via the `fn db vacuum` CLI command and tests — never from the periodic maintenance loop — so it is not a background event-loop stall, and an out-of-process VACUUM on a live WAL DB would add corruption surface for no hot-path benefit.

## Tests
- `central-db` / `archive-db`: assert the new PRAGMAs.
- `db.ts`: a second connection (`busy_timeout=0`) can read immediately after `vacuum()` returns (fails on pre-fix code); coverage for `integrityCheckSqliteFileAsync`; startup-integrity tests updated to the async/offloaded seam (deterministic regardless of whether the `sqlite3` CLI exists).
- 188 tests pass across the affected suites; `tsc --noEmit` clean.

## Notes
- Reviewed via `/ce-code-review` (9 reviewers). One finding (a claimed `synchronous` NORMAL→FULL regression) was verified to be a false positive and dropped.
- Follow-ups intentionally **not** included: extracting a shared `applyWalPragmas()` helper (3 inline copies for now); read-latency instrumentation to confirm steady-state read stalls (the evidence that would gate any larger async/worker investment — which the eval recommends against, as it would reinvite the `node:sqlite` corruption history).

### DIFF
--- a/packages/core/src/__tests__/archive-db-fts-maintenance.test.ts
+++ b/packages/core/src/__tests__/archive-db-fts-maintenance.test.ts
@@ -219,3 +219,35 @@ describe("ArchiveDatabase FTS maintenance", () => {
     }
   });
 });
+
+describe("ArchiveDatabase WAL durability PRAGMAs", () => {
+  let dir: string;
+  let archive: ArchiveDatabase;
+
+  beforeEach(() => {
+    dir = makeTmpDir("kb-archive-pragma-");
+    archive = new ArchiveDatabase(dir);
+  });
+
+  afterEach(async () => {
+    archive.close();
+    await rm(dir, { recursive: true, force: true });
+  });
+
+  it("bounds WAL growth and durability like the per-project DB", () => {
+    const rawDb = (archive as unknown as { db: { prepare(sql: string): { get(): unknown } } }).db;
+    const synchronous = rawDb.prepare("PRAGMA synchronous").get() as { synchronous: number };
+    const autoCheckpoint = rawDb
+      .prepare("PRAGMA wal_autocheckpoint")
+      .get() as { wal_autocheckpoint: number };
+    const journalSizeLimit = rawDb
+      .prepare("PRAGMA journal_size_limit")
+      .get() as { journal_size_limit: number };
+
+    expect(synchronous.synchronous).toBe(2); // FULL
+    expect(autoCheckpoint.wal_autocheckpoint).toBe(1000);
+    // Previously unset (-1 / unbounded), which let the archive WAL bloat and
+    // slow every reader. Now capped at 4 MB to match db.ts/central-db.ts.
+    expect(journalSizeLimit.journal_size_limit).toBe(4_194_304);
+  });
+});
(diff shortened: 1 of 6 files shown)


ghost:
Ready to review this PR? Stage has broken it down into 2 individual chapters for you:

| | Title |
|---|-------|
| 1 | [Bound WAL growth in shared databases](https://stagereview.app/Runfusion/Fusion/pull/1692/chapters/1) |
| 2 | [Release EXCLUSIVE lock after vacuum](https://stagereview.app/Runfusion/Fusion/pull/1692/chapters/2) |

<a href="https://stagereview.app/Runfusion/Fusion/pull/1692">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://stagereview.app/assets/gh-open-in-stage-dark.svg">
    Open in Stage
  </picture>
</a>

<sub>Chapters generated by <a href="https://stagereview.app">Stage</a> for commit 61c29cc6b421d39b3d30821691a37ee2c07b6671 on Jun 20, 2026 8:25pm UTC.</sub>

Issue #12, opened 2026-04-27
Welcome screen with broken UI


tmsss:
I've just installed it a Ubuntu machine, but the welcome screen is broken and I can't move on:

<img width="1899" height="830" alt="Image" src="https://github.com/user-attachments/assets/a00950c4-98ef-40af-a81f-1742446ce84c" />

gsxdsm:
On it!

gsxdsm:
Should be resolve in 0.7.1

gsxdsm:
Should be fixed in 0.8.1

PR #1793, opened 2026-06-26
feat: migrate storage from SQLite to PostgreSQL

gsxdsm:
# Migrate storage from SQLite to PostgreSQL — full dashboard cutover

Migrates Fusion's storage layer to the embedded PostgreSQL `AsyncDataLayer` (the default backend) and **completes the satellite-store + feature cutover** so every dashboard and Command Center surface works in PG mode.

## Status — every surface works in embedded-PG mode

Verified live against a running embedded-Postgres dashboard (all **200**, zero 5xx) and gate-tested (**23 files / 99 tests** on embedded PG, plus engine-core 294 and ci-shape 63 in the blocking merge gate; core/engine/cli/dashboard typecheck clean).

| Area | Surfaces | State |
|---|---|---|
| Satellite stores | workflows, todos, insights, research, missions, goals, mailbox | ✅ |
| Views | artifacts, documents, evals | ✅ |
| Command Center | activity, productivity, team, tokens, tools, **workflows**, **github**, **signals**, **plugin-activations**, **live** (all 10) | ✅ |
| Run execution | insight generation, research run execution | ✅ (store-path; AI step needs a provider) |
| Live updates | SSE push for mission/research/insight events | ✅ |
| Workflow editing | create / update / delete / select (+ id counter) | ✅ |
| Engine | mission autopilot, incident-signal ingestion, regression storm-guard, agent wake-on-message | ✅ |
| Core | tasks, agents, secrets, automations, memory, chat, usage, PRs, git | ✅ |

## Approach

Each satellite store gets an `Async<Store>` wrapper exposing the sync store's method names over the existing `async-*-store.ts` helpers; `get<Store>Store()` returns a `Sync | Async` union; consumers `await` (harmless on sync), and engine/CLI paths that can't convert use `instanceof Sync` graceful fallback. Analytics aggregators branch on `"ping" in dbOrLayer` to run schema-qualified raw SQL over `project.*` (snake_case) in PG. Executors/orchestrators/autopilot are await-converted to drive the union store; the async store wrappers extend `EventEmitter` so SSE live-push fires in both backends.

Not-yet-ported capabilities degrade gracefully (never 500) and are individually called out in commits.

## Sync with main

The branch is kept continuously merged with `main` (currently through FN-7845, 2026-07-12); the earlier "final rebase deferred" note no longer applies. Use **Create a merge commit** (or squash) to land it — GitHub's rebase-merge cannot replay a merge-maintained branch.

## Residual Review Findings

Multi-agent code review of the PostgreSQL satellite-store ports (U1–U5) applied 3 safe fixes (see `fix(review): apply autofix feedback`). The following are **real but gated** — recorded here as follow-up work rather than auto-applied. All are SQLite→PostgreSQL **concurrency/atomicity regressions**: the sync stores were immune only by SQLite's single-writer, single-threaded-handler execution; the async ports open multi-await read-modify-write windows. **Reachability is low today** because the execution engines that generate concurrent same-run mutations (insight run executor, research orchestrator/dispatcher) are `instanceof`-gated to sync mode in PG. No process-crash class survived (all engine fallbacks correctly guard the sync store).

- **[P1] Research `appendResearchEvent` dual-write is non-atomic** (`packages/core/src/async-research-store.ts`, corroborated: adversarial + reliability). The `research_run_events` insert (own transaction) and the `run.events` jsonb update are separate writes — a crash between them, or two concurrent appends, splits the table count from the jsonb array. **Fix:** perform the seq-insert and the jsonb update in one `layer.transactionImmediate`.
- **[P1] Research run terminal-reversion via stale full-row persist** (`async-research-store.ts` `persistResearchRun`/`updateResearchStatus`). Concurrent `PATCH /runs/:id/status` + `POST /runs/:id/events` can revert a terminal run to `running` by overwriting the whole row, bypassing the transition guard. **Fix:** scoped column `UPDATE`s with a `WHERE status …` guard, or optimistic version column.
- **[P2] `updateResearchRun`/`updateInsightRun` read-then-write TOCTOU** — concurrent PATCHes last-writer-wins on the lifecycle merge. **Fix:** `SELECT … FOR UPDATE` / enclosing transaction.
- **[P2] `upsertRun`/`createRunOrThrowConflict` check-then-create race** (`async-insight-store.ts`) — two callers can each create an "active" run. **Fix:** partial unique index on `(projectId, trigger) WHERE status IN ('pending','running')`.
- **[P3] `createResearchRetryRun` return-value divergence** — sync returns the pre-update `queued` snapshot; async returns the reloaded `retry_waiting` run (persisted state is identical). Pick one side for cross-backend parity.
- **[P2/perf] Mission `getMissionWithHierarchy`/`getMissionHealth` N+1 fan-out** — O(milestones×slices) sequential round-trips hold one pool slot per request; can starve the pool for large hierarchies. **Fix:** batched/joined reads.
- **Testing gaps:** no PG-mode concurrency tests (interleaved status/event mutations), no sync↔async parity assertion for the lifecycle-error codes, and no mission status/health rollup parity test vs the sync `MissionStore`.

~~Out of scope (deferred): AI run *execution* (insight/research) + mission autopilot + live SSE mission events remain sync-gated/degraded in PG mode.~~ **Since ported** — insight/research run execution, mission autopilot, and SSE live push all run on the async layer now, which also makes the concurrency findings above genuinely reachable; they remain open follow-ups.

---

## Update — 2026-07-12: production-readiness hardening & live acceptance

Everything below landed on this branch since the description above was written:

**Production blockers from review — fixed**
- `recoverStaleTransitionPending` ported to the async layer (backend moves write + clear the crash-safe marker; startup/maintenance sweeps no longer throw).
- Lost-update class fixed: `atomicWriteTaskJson`/`WithAudit` write changed columns only (full-row upserts silently resurrected stale fields across concurrent store instances — the "task stuck unplanned forever" bug).
- First-boot **auto-migration**: booting the PG backend over a project with a legacy `fusion.db` migrates it automatically (loud failure, SQLite kept as backup), and the dashboard shows a one-time **"your data was migrated" banner** with the backup paths and a Need-help Discord link.
- `pg_dump`/`pg_restore` discovered from common install locations for embedded-mode backups.
- The PG suite is part of the blocking merge gate (`test:pg-gate`).

**Multi-project isolation (PR #2007, merged into this branch)**
- `project_id` partition key on tasks / archived tasks / config, `taskProjectScope` threaded through every scan/claim/count, per-project config rows, layer bound to the project at startup.
- Review P1 follow-up: the shared cold-storage `archive.archived_tasks` table is also partitioned and all archived-board reads/counts/searches are scoped.
- Schema drift self-heal generalized to schema-qualified columns so existing databases upgrade in place.

**Other changes**
- Node settings sync **removed** in PG mode (409 `settings-sync-disabled-postgres`) — nodes share state by connecting to the same database; auth sync kept (per-machine file).
- Perf (review findings): `listTasks` pushes column filter + ORDER BY + LIMIT/OFFSET into SQL; `getConversation` capped to the most recent 200 messages.
- Fixed a false "operator action required" pause-abort log fired on every successfully auto-merged task.

**Live acceptance — PASSED (2026-07-12)**
A sandboxed instance (isolated HOME, embedded PG, real Opus executor) ran a task through the complete cycle: create → triage (AI spec) → execute → in-review → AI squash-merge landed on the project's `main` → done. A write+read sweep of every data surface (settings, comments, documents, attachments + artifact bridge + artifact edit, chat with real generation, goals, missions, agent mail, secrets, workflows, memory, CC analytics) was green on embedded PG.

**Known remaining work**
- The per-project `config` PK re-key has no upgrade path for pre-isolation embedded-PG databases (needs a real `DROP CONSTRAINT`/re-key migration; fresh databases are fine).
- `pg_dump`/`pg_restore` binaries are not yet bundled in release artifacts (PATH/common-location discovery only).
- The satellite-store concurrency findings listed above.


packages/core/src/store.ts
```diff
@@ -1491,14494 +278,1921 @@ function filterRepairOverlapIgnoredPaths(paths: string[], ignorePaths: string[])
...
-      await central.init();
-      initializedHere = true;
-    }
+  public static readonly ACTIVE_TASKS_WHERE = '"deletedAt" IS NULL';
+  /**
+   * FNXC:RuntimePersistenceAsync 2026-06-24-10:42: Task-table columns stored as jsonb in PostgreSQL.
+   * pgRowToTaskRow() re-serializes them to strings so rowToTask() works unchanged across both backends.
+   * MUST match TASK_JSONB_COLUMNS in async-persistence.ts.
+   */
+  public static readonly PG_JSONB_TASK_COLUMNS: ReadonlySet<string> = new Set(["dependencies", "steps", "customFields", "log", "attachments", "steeringComments", "comments", "review", "reviewState", "workflowStepResults", "prInfo", "prInfos", "issueInfo", "githubTracking", "mergeDetails", "enabledWorkflowSteps", "modifiedFiles", "scopeAutoWiden", "sourceMetadata", "tokenUsagePerModel", "tokenBudgetOverride"]);
```
gsxdsm, line 307: **[P1] `workspaceWorktrees` silently dropped on every PG-mode task read.** `PG_JSONB_TASK_COLUMNS` omits `workspaceWorktrees`, but the column is `workspace_worktrees jsonb` (0000_initial.sql:131, schema/project.ts:157). In PG mode `pgRowToTaskRow` leaves the parsed JS object as-is, then `fromJson()` (serialization.ts:216) runs `JSON.parse(object)` → throws → `undefined`, so `isWorkspaceTask()` returns false and workspace land/cleanup no-ops. Add `"workspaceWorktrees"` to the set, and add a jsonb round-trip test asserting `isWorkspaceTask()` survives write→read. Flagged by 4 reviewers (correctness/adversarial/reliability/api-contract), confidence 100. This Set is hand-maintained with no compile-time parity check — derive/assert it against the Drizzle jsonb columns.

packages/core/src/task-store/remaining-ops-10.ts
```diff
@@ -0,0 +1,256 @@
...
+      await store.asyncLayer!.db.delete(schema.project.activityLog);
+      return;
+    }
+    store.db.prepare("DELETE FROM activityLog").run();
+    store.db.bumpLastModified();
+}
+
+export function getInsightStoreImpl(store: TaskStore): InsightStore {
+    if (!store.insightStore) {
+      store.insightStore = new InsightStore(store.db);
```
gsxdsm: **[P1] Insights API 500s in default (embedded-PG) mode.** `getInsightStoreImpl` constructs `new InsightStore(store.db)` unconditionally; `store.db` throws *"SQLite Database is not available in backend mode"*, so every `/api/insights/*` request fails (the insights-routes.ts startup try/catch only guards the sweeper, not the per-request path at :326). `getResearchStoreImpl` at line 235 has the identical bug → `/api/research/*`. Port InsightStore/ResearchStore to an AsyncDataLayer path, or guard + return 503. Confidence 100.

packages/dashboard/src/routes/register-signal-routes.ts
```diff
@@ -206,10 +206,18 @@ export async function ingestSignal(deps: SignalIngestDeps): Promise<SignalIngest
...
+    dropped. Now resolveIncident receives the async layer in backend mode
+    (getAsyncLayer() ?? getDatabase(), matching monitor-routes.ts) and the call
+    is awaited so errors surface in the catch below. ingestIncidentSignal
+    remains sync/SQLite-scoped (a separate unguarded path); the await only
+    applies to the now-async resolveIncident.
     */
-    const db = store.getDatabase();
     const at = signalTimestampToIso(signal.timestamp) ?? new Date().toISOString();
-    ingestIncidentSignal(db, {
+    ingestIncidentSignal(store.getDatabase(), {
```
gsxdsm: **[P1] Signal ingestion silently dropped in PG mode.** `ingestIncidentSignal(store.getDatabase(), …)` is unguarded; `resolveIncident` (line ~230) was correctly converted but its sibling was not. The error is caught so the server doesn't 500, but every connector signal is silently lost in the default backend. Port `ingestIncidentSignal` to accept `AsyncDataLayer | Database` (ping-discriminator) or guard with a logged skip. Confidence 100.

packages/core/src/task-store/remaining-ops-8.ts
```diff
@@ -0,0 +1,949 @@
...
+      /*
+       * FNXC:SqliteFinalRemoval 2026-06-24-15:50:
+       * In backend mode (PostgreSQL), the MissionStore has not been converted to
+       * use the AsyncDataLayer yet — it requires a sync SQLite Database handle.
+       * The InProcessRuntime catches the resulting error from store.db access and
+       * degrades gracefully (mission autopilot disabled). This is the expected
+       * boundary until the MissionStore is fully converted to async in a future
+       * session.
+       */
+      store.missionStore = new MissionStore(store.fusionDir, store.db, store);
```
gsxdsm: **[P1] MissionStore SSE events silently dropped in PG mode.** `getMissionStoreImpl` constructs `new MissionStore(store.fusionDir, store.db, store)` unconditionally; `store.db` throws in backend mode. `server.ts:~1021` wraps the call in a try/catch that swallows the error, so mission autopilot + SSE mission events degrade with no operator signal. Port MissionStore to AsyncDataLayer, or at minimum surface the degradation rather than swallowing it. Confidence 100.

packages/dashboard/src/otel-exporter.ts
```diff
@@ -246,7 +246,7 @@ export function startOtelExporter(deps: OtelExporterDeps): OtelExporterHandle {
     try {
       const db = store.getDatabase();
```
gsxdsm: **[P2] OTel metrics export silently fails in PG mode.** `store.getDatabase()` is unguarded; the try/catch at ~:258 suppresses the throw, so the operator's OTel endpoint receives no metrics with no visible error. Use `store.getAsyncLayer() ?? store.getDatabase()` and update `aggregateTokenAnalytics`/`aggregateActivityAnalytics` to the union type. Confidence 100.

packages/engine/src/executor.ts
```diff
@@ -1882,7 +1882,7 @@ export class TaskExecutor {
       this.store.setCompletionHandoffAcceptedMarker(task.id, {
```
gsxdsm, line 2299: **[P1] Fire-and-forget async write.** `setCompletionHandoffAcceptedMarker(...)` is now async but called without `await`, before the awaited `upsertMergeRequestRecord` — the marker write can race the subsequent parity read and rejections are unhandled. Same at self-healing.ts:811. Add `await`. Confidence 100.

packages/core/src/async-central-core.ts
```diff
@@ -0,0 +1,1789 @@
...
+ * Atomically acquire a global concurrency slot or queue the request. Mirrors
+ * the sync acquireGlobalSlot() transaction: read the singleton row, increment
+ * currentlyActive + project inFlightAgentCount if a slot is available,
+ * otherwise increment queuedCount.
+ */
+export async function acquireGlobalSlotAtomic(
+  layer: AsyncDataLayer,
+  projectId: string,
+): Promise<boolean> {
+  return layer.transactionImmediate(async (tx) => {
```
gsxdsm, line 1437: **[P1] Concurrency-cap overrun under READ COMMITTED.** `acquireGlobalSlotAtomic` does a plain `getGlobalConcurrencyRow` SELECT (no `FOR UPDATE`) then writes `active+1` inside a READ COMMITTED tx. Two concurrent acquires both read `active < max` and both increment past the cap. The old SQLite `BEGIN IMMEDIATE` serialized this. Use a conditional `UPDATE … SET currently_active = currently_active + 1 WHERE currently_active < global_max_concurrent` and check rowCount, or `SELECT … FOR UPDATE`. Add a concurrent-acquire test. Confidence 75 (2 reviewers).

packages/engine/src/runtimes/in-process-runtime.ts
```diff
@@ -200,23 +209,79 @@ export class InProcessRuntime
...
+      // CentralCore so it migrates off its SQLite CentralDatabase and shares the
+      // SAME PostgreSQL connection pool as everything else. The shared CentralCore
+      // is constructed before the backend is resolved (serve.ts/daemon.ts), so we
+      // attach the layer here once the TaskStore is available. If the CentralCore
+      // is already in backend mode (constructed with the layer) this is a no-op
+      // via the init() idempotency guard. Safe in legacy mode (messageLayer null).
+      if (messageLayer && !this.centralCore.backendMode) {
+        try {
+          await this.centralCore.attachBackendLayer(messageLayer);
+        } catch (err) {
```
gsxdsm, line 350: **[P1] Bootstrap failure silently disables the concurrency cap.** If `attachBackendLayer` throws, this catch logs and continues with `backendMode=true`, `initialized=false`, and no `global_concurrency` row seeded. `acquireGlobalSlotAtomic` then finds no row, the UPDATE affects 0 rows, and every acquire returns `true` → unlimited parallel agents for the process lifetime. Re-throw to fail startup (or clear `asyncLayer`) on bootstrap failure. Confidence 75 (2 reviewers).

packages/core/src/task-store/moves.ts
```diff
@@ -0,0 +1,937 @@
...
+            .select({ one: sql`1` })
+            .from(schema.project.mergeQueue)
+            .where(eq(schema.project.mergeQueue.taskId, id))
+            .limit(1);
+          alreadyEnqueued = existingRows.length > 0;
+          await enqueueMergeQueueInTransaction(tx, id, { priority: task.priority, now: internal.now }, {
+            agentId: internal.runContext?.agentId,
+            runId: internal.runContext?.runId,
+          });
+          await store.createCompletionHandoffWorkflowWork(task, {
```
gsxdsm, line 809: **[P1] Handoff-to-review atomicity broken.** `createCompletionHandoffWorkflowWork` is called without the outer move `tx`; in backend mode its `cancelActiveWorkflowWorkItemsForTask` + `upsertWorkflowWorkItem` each open a separate `layer.transactionImmediate` that commits immediately. An outer-tx rollback (e.g. the audit write at ~:707) then leaves an orphaned runnable merge-gate item for a task that was never handed off — plus nested-pool deadlock risk. Thread the outer `tx` through these calls. Confidence 75.

packages/core/src/async-central-db.ts
```diff
@@ -0,0 +1,354 @@
...
+ */
+export async function tryClaimTask(
+  layer: AsyncDataLayer,
+  input: TryClaimInput,
+): Promise<TryClaimResult> {
+  return layer.transactionImmediate(async (tx): Promise<TryClaimResult> => {
+    const existing = await getTaskClaim(tx, input.projectId, input.taskId);
+    const now = input.renewedAt;
+
+    if (!existing) {
```
gsxdsm, line 174: **[P2] `tryClaimTask` fresh-claim race throws raw `23505` instead of a conflict result.** Two racing first-claims both see no row and both INSERT; the PK makes the loser raise `unique_violation`, which propagates raw to `agent-store.ts:1542` (which expects `{ok:false}`). Use `INSERT … ON CONFLICT (project_id, task_id) DO NOTHING RETURNING *`; if no row, re-read and return `{ok:false, reason:'conflict', current}`. Confidence 75 (2 reviewers).

packages/core/src/postgres/schema/project.ts
```diff
@@ -0,0 +1,1632 @@
...
+  createdAt: bigint("created_at", { mode: "number" }).notNull(),
+  updatedAt: bigint("updated_at", { mode: "number" }).notNull(),
+  closedAt: bigint("closed_at", { mode: "number" }),
+}, (t) => [
+  check("pull_requests_source_type_check", sql`${t.sourceType} IN ('task','branch-group')`),
+  check(
+    "pull_requests_state_check",
+    sql`${t.state} IN ('creating','open','responding','merged','closed','failed')`,
+  ),
+  unique("idxPullRequestsOpenSource").on(t.sourceType, t.sourceId),
```
gsxdsm, line 1002: **[P1] Latent destructive schema drift: `pull_requests` partial unique indexes declared as full constraints.** The Drizzle schema uses `unique()` (global) but the authoritative `0000_initial.sql:832-840` creates *partial* unique indexes (`WHERE state NOT IN ('merged','closed','failed')` / `WHERE pr_number IS NOT NULL`). Today the SQL is authoritative at runtime, so it's latent — but the first `drizzle-kit generate` will emit a destructive migration replacing the partials with full constraints, breaking PR re-open/history. Use `uniqueIndex().where(...)` to match the SQL. Confidence 100.

packages/core/src/postgres/schema/central.ts
```diff
@@ -0,0 +1,325 @@
...
+  type: text("type").notNull(),
+  projectId: text("project_id").notNull(),
+  projectName: text("project_name").notNull(),
+  taskId: text("task_id"),
+  taskTitle: text("task_title"),
+  details: text("details").notNull(),
+  metadata: jsonb("metadata"),
+}, (t) => [
+  foreignKey({ columns: [t.projectId], foreignColumns: [projects.id] }).onDelete("cascade"),
+  index("idxActivityLogTimestamp").on(t.timestamp),
```
gsxdsm, line 120: **[P1] Latent schema drift: `central_activity_log` index names mismatch + collide.** Lines 120-122 name indexes `idxActivityLog*`, but the SQL (`0000_initial.sql:1640-1645`) uses `idxCentralActivityLog*`, and the wrong names also collide with the identically-named `project.activity_log` indexes (project.ts:363-365). `drizzle-kit` would drop/recreate. Rename to `idxCentralActivityLog{Timestamp,Type,ProjectId}`. Also: `todo_lists` is missing `idxTodoListsProjectId` and 5 timestamp indexes (project.ts:235,366,367,456,1328) omit the `.desc()` the SQL has. Confidence 100.

packages/core/src/postgres/migrations/0000_initial.sql
```diff
@@ -0,0 +1,1867 @@
...
+  allow_resurrection integer DEFAULT 0,
+  transition_pending text,
+  custom_fields jsonb DEFAULT '{}',
+  -- FNXC:TaskStoreSearch 2026-06-24-12:30:
+  -- Full-text search vector (tsvector) replacing the SQLite FTS5 tasks_fts
+  -- table. GENERATED ALWAYS so PostgreSQL keeps it in sync on write
+  -- (VAL-SEARCH-002/003/004). 'simple' config for code-like tokenization
+  -- parity with FTS5. Value-aware: only regenerates when id/title/description/
+  -- comments change (VAL-SEARCH-006).
+  search_vector tsvector GENERATED ALWAYS AS (
```
gsxdsm, line 181: **[P1] `search_vector` write amplification + factually-wrong comment.** The `GENERATED ALWAYS AS … STORED` tsvector column recomputes `to_tsvector(...)` on **every** task UPDATE — including ~30s heartbeats and timing/status writes — regardless of which columns changed. The comment at line 169 claiming it's "value-aware: only regenerates when id/title/description/comments change" is false for PostgreSQL ≤15. Replace with a `BEFORE UPDATE` trigger gated on `OLD.<col> IS DISTINCT FROM NEW.<col>` for the 4 FTS columns (apply to `archived_tasks` at ~:1856 too), and consider a partial GIN index `WHERE deleted_at IS NULL`. Flagged by 3 reviewers, confidence 100.

packages/core/src/postgres/migrations/0000_initial.sql
```diff
@@ -0,0 +1,1867 @@
...
+  interview_state text NOT NULL,
+  dependencies jsonb DEFAULT '[]',
+  planning_notes text,
+  verification text,
+  acceptance_criteria text,
+  validation_state text NOT NULL DEFAULT 'not_started',
+  created_at text NOT NULL,
+  updated_at text NOT NULL,
+  CONSTRAINT milestones_mission_id_fkey
+    FOREIGN KEY (mission_id) REFERENCES project.missions(id) ON DELETE CASCADE
```
gsxdsm, line 931: **[P1] Missing indexes on mission-hierarchy FK columns.** No index on `milestones.mission_id`, `slices.milestone_id`, or `mission_features.slice_id`, so every `getMissionWithHierarchy` (Missions view, SSE reconnect, autopilot tick) does 3 sequential full table scans. Add `idxMilestonesMissionId`, `idxSlicesMilestoneId`, `idxMissionFeaturesSliceId`. Confidence 100.

packages/core/src/task-store/async-persistence.ts
```diff
@@ -0,0 +1,438 @@
...
+ * payload on busy boards per the slim-read analysis) from the SELECT so the
+ * wire transfer is bounded. Callers that need the activity log fetch the
+ * individual task via `readTaskRow` (full row). This mirrors the SQLite path's
+ * `getTaskSelectClause(slim)` projection.
+ *
+ * @param layer The async data layer.
+ * @param options Optional: excludeLog drops the `log` jsonb column;
+ *   includeDeleted surfaces soft-deleted rows for forensic reads (VAL-DATA-006).
+ */
+export async function readLiveTaskRows(
```
gsxdsm, line 340: **[P1] `readLiveTaskRows` full-table scan with JS-side pagination.** Loads `SELECT … FROM tasks WHERE deleted_at IS NULL` (80+ cols, jsonb) on every board hydration; `listTasksImpl` then applies limit/offset in JS after fetching all rows (the code's own comment at reads.ts:~250 acknowledges the debt). Push `.limit()/.offset()` (or status/column scoping) to SQL. Free under in-process SQLite, a large network transfer over Postgres. Confidence 100.

packages/core/src/async-chat-store.ts
```diff
@@ -0,0 +1,902 @@
...
+    .limit(limit)
+    .offset(offset);
+  return rows.map(rowToMessage);
+}
+
+/**
+ * FNXC:ChatStore 2026-06-24-09:15:
+ * Get the latest message for each session in the provided list.
+ */
+export async function getLastMessageForSessions(
```
gsxdsm, line 245: **[P2] `getLastMessageForSessions` is O(sessions × depth).** Loads *all* messages for all requested sessions, then picks the newest per session in JS (e.g. 50 sessions × 500 msgs = 25k rows on every chat-list render). Replace with `SELECT DISTINCT ON (session_id) … ORDER BY session_id, created_at DESC, id DESC`. Confidence 100.

packages/core/src/postgres/schema/_shared.ts
```diff
@@ -0,0 +1,85 @@
...
+ *   - TEXT timestamps → text (ISO-8601 strings, preserved verbatim from SQLite)
+ *
+ * CHECK constraints, foreign-key cascade rules, and unique indexes are
+ * preserved one-for-one from the SQLite source of truth
+ * (SCHEMA_SQL / MIGRATION_ONLY_TABLE_SCHEMAS in db.ts, CENTRAL_SCHEMA_SQL,
+ * archive BASE_SCHEMA_SQL). See VAL-SCHEMA-002, VAL-SCHEMA-003, VAL-SCHEMA-005.
+ */
+
+/** PostgreSQL schema name for the per-project working database. */
+export const PROJECT_SCHEMA = "project";
```
gsxdsm, line 32: **[P1/security] No per-project isolation on a shared external `DATABASE_URL`.** `PROJECT_SCHEMA`/`CENTRAL_SCHEMA`/`ARCHIVE_SCHEMA` are compile-time constants and `backend-resolver` uses one process-wide `DATABASE_URL` with no projectId in the connection path. Two projects pointed at the same external Postgres share the same three schemas → cross-project read/write **and** `ON CONFLICT` task-ID clobber (a confidentiality + data-integrity boundary for multi-tenant deployments). Namespace the schema per project (or require a distinct DB/URL) and enforce at connection time. Confidence 50 (2 reviewers).

packages/core/src/task-store/audit-ops.ts
```diff
@@ -0,0 +1,237 @@
...
+      );
+    } catch {
+      // Marker bookkeeping is best-effort; proceed to run the hooks regardless.
+    }
+
+    // Read the task once for hook context. MUST be a non-locking read — this
+    // runs inside `withTaskLock`, so `getTask` (which re-acquires the lock)
+    // would deadlock. `readTaskFromDb` is the in-lock-safe read.
+    const taskRow = store.readTaskFromDb(taskId, { includeDeleted: false });
+    const taskDetail = taskRow as unknown as TaskDetail | undefined;
```
gsxdsm, line 59: **[P1] `taskRow as unknown as TaskDetail` bypasses deserialization.** The cast hands plugin trait-hook consumers a raw `TaskRow` whose jsonb fields are JSON strings/objects rather than parsed values (`.steps`, `.review`, `.dependencies`). Use `store.rowToTask(taskRow)` so hooks receive a fully-deserialized `Task`. Confidence 100.

packages/core/src/postgres/connection.ts
```diff
@@ -0,0 +1,264 @@
...
+        "Ensure the embedded lifecycle has provided a URL or DATABASE_URL is set.",
+    );
+  }
+
+  // Determine whether to use prepared statements in the runtime pool.
+  // Disable when the URL is a pooler and no migration URL split is configured.
+  const runtimeIsPooler = looksLikePoolerUrl(runtimeUrl);
+  const runtimePrepare = backend.migrationUrlOverridden ? true : !runtimeIsPooler;
+
+  const runtimeSql = postgres(runtimeUrl, {
```
gsxdsm, line 194: **[P2] No `statement_timeout` / `idle_in_transaction_session_timeout` on the runtime pool.** A stalled query holds a pool connection indefinitely; with `max=10`, a few concurrent stalls exhaust the pool and all subsequent queries queue with no deadline. Set both via the `postgres()` options here. Confidence 100.

packages/core/src/postgres/postgres-health.ts
```diff
@@ -0,0 +1,479 @@
...
+  for (const finding of findings) {
+    const ddlType = TYPE_TO_DDL[finding.expectedType] ?? "TEXT";
+    try {
+      await db.execute(
+        sql.raw(
+          `ALTER TABLE ${PROJECT_SCHEMA}.${finding.table} ADD COLUMN IF NOT EXISTS "${finding.column}" ${ddlType}`,
+        ),
+      );
+      healed.push(finding);
+    } catch {
```
gsxdsm, line 368: **[P2] `healSchemaDrift` `catch {}` silently swallows ALTER TABLE errors.** The block is empty despite a comment implying failures are logged, so a column that fails to heal is invisible to operators. Add a `log.warn(...)` with the failing column/error. (Related: `validateAndHealSchema` at ~:354 runs ALTER on the runtime pool, not the migration connection — fails under a transaction-mode pooler.) Confidence 100.

packages/core/src/postgres/schema-applier.ts
```diff
@@ -0,0 +1,106 @@
...
+  await ensureBookkeepingTable(db);
+  const applied = await getAppliedMigrations(db);
+  const alreadyApplied = applied.includes(SCHEMA_BASELINE_VERSION);
+
+  if (!alreadyApplied) {
+    const baselineSql = await readBaselineMigrationSql();
+    // The baseline contains multiple statements including CREATE SCHEMA, CREATE
+    // TABLE, CREATE INDEX, and seed INSERTs. postgres.js executes a single
+    // query string as one batch (simple query protocol when unparameterized).
+    await db.execute(sql.raw(baselineSql));
```
gsxdsm, line 92: **[P2] Baseline apply + bookkeeping INSERT are not atomic.** The 1700-line baseline `execute()` and the `fusion_schema_migrations` marker INSERT are separate statements with no wrapping transaction; a crash between them leaves the schema applied but no `0000` row, so the full baseline re-runs on every subsequent boot (idempotent via `IF NOT EXISTS`, but wasteful). Wrap both in `db.transaction()`. Confidence 100.

gsxdsm:
## Multi-agent code review — SQLite → PostgreSQL migration (3rd pass, HEAD `b996186e3`)

Re-ran the full 10-reviewer team against the current squash. I posted **21 inline comments** on specific lines; this summary adds context plus 6 findings whose lines GitHub couldn't attach inline (unchanged context, or files this squash didn't modify).

### Fixed since earlier passes (verified) ✅
All 7 original P0s, plus: merge-path `backendMode` guards, `pg_dump`/`pg_restore` libpq env, the soft-delete write guard, SIGTERM re-exit, embedded-start timeout, partial-backup cleanup, the migrator column-type table-scoped join + MD5 checksum verification, the `monitor-store` `"ping"` discriminator, credential query-param redaction, `source_parent_task_id` index, merge-queue N+1 batching, and CI now runs a real `postgres:15` probe (the dead-tests issue is resolved).

### Dominant remaining theme
The "unguarded sync DB access throws in default embedded-PG mode" class was fixed only where prior passes named it. It still breaks **Insights, Research, Command Center analytics, signal/monitor, MissionStore, and a stale compiled roadmap `.js`**. Root cause: ~314 `backendMode` branches with no per-router PG smoke test, so each new handler must be guarded by hand or silently breaks. Several engine call sites also call now-`async` methods unawaited (`Promise !== null` is always true), and the tests mock those methods synchronously, masking the bug.

### Highest-leverage next steps
1. **One PG-mode integration smoke test per dashboard router** — catches the entire unguarded-store class at once.
2. **A jsonb round-trip test over every `Task` field** — catches the `workspaceWorktrees` drop and future `PG_JSONB_TASK_COLUMNS` drift.
3. **Concurrency tests** for `acquireGlobalSlotAtomic`/`tryClaimTask`; fix the unawaited `getCompletionHandoffAcceptedMarker` calls and switch their test mocks to `mockResolvedValue`.
4. **Reconcile the Drizzle schema vs `0000_initial.sql`** before wiring `drizzle-kit` (the first `generate` would be destructive).
5. **Admit a representative PG suite into the merge gate** (currently PG tests run only in non-blocking full-suite).

### Findings not attachable inline

- **[P1] 7 of 9 Command Center analytics aggregators throw in default PG mode.** Lines 182, 244, 290, 341, 362, 405, 454 call `store.getDatabase()` with no backend guard → 500 in embedded-PG mode. Lines 267 and 433 already use the correct `store.getAsyncLayer() ?? store.getDatabase()` pattern; apply it to the other seven (and have the aggregators accept `Database | AsyncDataLayer`). Add a PG-mode smoke test per route. Confidence 100 (3 reviewers).

- **[P1] Auto-regression monitor broken in PG mode.** `runMonitorOnRegression` does `const db = store.getDatabase()` unguarded, then passes it to `ingestIncidentSignal`/`countRecentAutoFixTasks`/`claimIncidentForFixTask` — all SQLite-only. Use `store.getAsyncLayer() ?? store.getDatabase()` and update the callees to the union type. Confidence 100.

- **[P1] Stale compiled artifact: roadmap routes throw in PG mode.** This committed `.js` calls `new RoadmapStore(ctx.taskStore.getDatabase())` with no guard, but the TS source `roadmap-routes.ts:51` already has `scopedTaskStore.isBackendMode() ? null : getDatabase()`. The `.js` was not regenerated, so the runtime artifact still throws in backend mode. Rebuild the plugin so the compiled output reflects the source. Confidence 100.

- **[P1] Unawaited async marker → in-review file-scope-lease decision inverted.** `this.store.getCompletionHandoffAcceptedMarker(t.id) !== null` runs inside a synchronous `.filter()`; the method is async, so `Promise !== null` is **always true** → `handoffAccepted` is permanently true and `shouldHoldActiveFileScopeLease` is wrong for every in-review task under `mergeRequestContractShadowEnabled`. Same bug at line 2214. Pre-resolve a `Map` via `await Promise.all(...)` before the filter (pattern already used at :1459/:2187). Tests miss it because the mock uses `mockReturnValue(null)` (sync) instead of `mockResolvedValue(null)`. Confidence 100 (4 reviewers).

- **[P1] Same unawaited-marker bug as scheduler.ts:1527.** `getCompletionHandoffAcceptedMarker(blocker.id) !== null` is unawaited in `hasActiveFileScopeOverlapBlocker` (the enclosing fn is async) → `Promise !== null` always true. Same at line 4821. Add `await`. Confidence 100.

- **[P1] The 60 new PG tests are not in the merge gate.** CI now provisions `postgres:15` and uses a real TCP probe for `PG_AVAILABLE` (the pass-1 dead-tests issue is fixed) — but `pnpm test:gate` → `@fusion/engine test:core` is an explicit allow-list that excludes `packages/core/src/__tests__/postgres/`, so the PG suites run only in non-blocking `full-suite.yml`. For a storage-backend rewrite this large, admit a representative `pgDescribe` suite (especially a per-router PG-mode smoke test) into the gate. Confidence 100.

---
_Full per-pass reports in `docs/postgres-migration-review-2026-06-26.md` and `...-pass2.md`._

packages/core/src/store.ts
gsxdsm: Addressed: Added `"workspaceWorktrees"` to `PG_JSONB_TASK_COLUMNS`. The jsonb column is now included in the PG-mode select set so `pgRowToTaskRow` handles it correctly.

packages/core/src/task-store/remaining-ops-10.ts
gsxdsm: Addressed: Added `backendMode` guards in `getInsightStoreImpl`, `getResearchStoreImpl`, and `getTodoStoreImpl` with `console.warn` logging. The stores throw a clear error instead of accessing `store.db`, and callers already have try/catch that returns appropriate error responses.

packages/dashboard/src/routes/register-signal-routes.ts
gsxdsm: Addressed: Guarded `ingestIncidentSignal` with `store.backendMode` check. In PG mode it logs a warning so the degradation surfaces to operators rather than being silently swallowed by the outer try/catch.

packages/core/src/task-store/remaining-ops-8.ts
gsxdsm: Addressed: Guarded `getMissionStoreImpl` with `backendMode` check and `console.warn` logging. The error now surfaces clearly instead of being silently swallowed.

packages/dashboard/src/otel-exporter.ts
gsxdsm: Addressed: Added `backendMode` guard in `exportOnce()` with a `log.warn` when skipped, so the degradation surfaces to operators.

packages/engine/src/executor.ts
gsxdsm: Addressed: Added `await` to both `setCompletionHandoffAcceptedMarker` call sites (executor.ts:1882 and self-healing.ts:811).

packages/core/src/async-central-core.ts
gsxdsm: Addressed: Replaced SELECT-then-write with a conditional `UPDATE ... SET currently_active = currently_active + 1 WHERE currently_active < global_max_concurrent RETURNING id`. The result row count determines success, making the acquire race-free under READ COMMITTED.

packages/engine/src/runtimes/in-process-runtime.ts
gsxdsm: Addressed: Removed the try/catch around `attachBackendLayer`. If it throws, the error now propagates and fails startup rather than continuing with `backendMode=true` but no async layer.

packages/core/src/task-store/moves.ts
gsxdsm: Noted. This is tracked as P1 #12 from the internal code review with an `it.fails` regression guard already in place. Threading the outer `tx` through `cancelActiveWorkflowWorkItemsForTask` and `upsertWorkflowWorkItem` requires changing their signatures across many call sites — too risky for a feedback pass. Will address in a dedicated follow-up.

packages/core/src/async-central-db.ts
gsxdsm: Addressed: Changed INSERT to `ON CONFLICT DO NOTHING RETURNING *`. If 0 rows returned (lost the race), re-reads the winner row and returns `{ok:false, reason:"conflict", current}` instead of throwing raw 23505.

packages/core/src/postgres/schema/project.ts
gsxdsm: Addressed: Changed `unique()` to `uniqueIndex().where()` matching the partial indexes in `0000_initial.sql` (state NOT IN merged/closed/failed, prNumber IS NOT NULL). Prevents drizzle-kit from emitting destructive full-constraint migrations.

packages/core/src/postgres/schema/central.ts
gsxdsm: Addressed: Renamed `idxActivityLog*` to `idxCentralActivityLog*` to match SQL and avoid collision with project.activity_log indexes.

packages/core/src/postgres/migrations/0000_initial.sql
gsxdsm: Addressed: Replaced `GENERATED ALWAYS AS ... STORED` with `BEFORE INSERT/UPDATE` triggers gated on `OLD.col IS DISTINCT FROM NEW.col` for the 4 FTS columns. Also fixed the factually wrong comment and applied to `archived_tasks`. Added partial GIN index `WHERE deleted_at IS NULL`.

packages/core/src/postgres/migrations/0000_initial.sql
gsxdsm: Addressed: Added `idxMilestonesMissionId`, `idxSlicesMilestoneId`, and `idxMissionFeaturesSliceId` to `0000_initial.sql`.

packages/core/src/task-store/async-persistence.ts
gsxdsm: Noted. This is an acknowledged debt (documented in the code comment at reads.ts:252). The JS-side pagination mirrors the SQLite path for behavioral parity. Pushing LIMIT/OFFSET to SQL requires also pushing the sort logic (column-based, date-based) which is a larger refactor. Tracked as a follow-up.

packages/core/src/task-store/audit-ops.ts
gsxdsm: Addressed: Replaced `taskRow as unknown as TaskDetail` with `store.rowToTask(taskRow)` so jsonb fields are properly deserialized before being passed to plugin trait hooks.

packages/core/src/postgres/schema/_shared.ts
gsxdsm: Noted. The current architecture uses one embedded PG per `~/.fusion/` directory — projects are isolated by directory, not by schema name. `DATABASE_URL` is intended for a single project pointing at a dedicated database. Multi-tenant shared-database deployments are a future use case that would require per-project schema namespacing. Tracked as a design follow-up.

packages/core/src/async-chat-store.ts
gsxdsm: Addressed: Replaced load-all-then-filter-in-JS with `SELECT DISTINCT ON (session_id) ... ORDER BY session_id, created_at DESC, id DESC` for O(sessions) instead of O(sessions x depth).

packages/core/src/postgres/connection.ts
gsxdsm: Addressed: Added `connection: { statement_timeout: 60000, idle_in_transaction_session_timeout: 30000 }` to the runtime pool config via postgres.js connection parameters.

packages/core/src/postgres/postgres-health.ts
gsxdsm: Addressed: Added `console.warn` in the catch block so failed ALTER TABLE operations are logged and visible to operators.

packages/core/src/postgres/schema-applier.ts
gsxdsm: Addressed: Wrapped the baseline `execute()` and the `fusion_schema_migrations` marker INSERT in a single `db.transaction()` so a crash between them cannot leave the schema applied but no marker row.

gsxdsm:
## Re-review of `b8ce47ea5` ("Address PR review feedback") ✅

Thanks for the fast turnaround — verified each prior finding against the new commit. **~16 fixed, 4 degraded-but-acknowledged, ~10 still open.**

### Fixed ✅
- `store.ts` — `workspaceWorktrees` added to `PG_JSONB_TASK_COLUMNS` (silent data-loss gone)
- `async-central-core.ts` — `acquireGlobalSlotAtomic` now a conditional `UPDATE … WHERE currently_active < global_max_concurrent RETURNING` (cap overrun fixed)
- `async-central-db.ts` — `tryClaimTask` uses `onConflictDoNothing()` + returns `{ok:false, reason:'conflict'}` instead of raw `23505`
- `0000_initial.sql` — `search_vector` converted from `GENERATED ALWAYS` to a BEFORE INSERT/UPDATE **trigger** gated on FTS columns (write amplification + the inaccurate comment both fixed); mission-hierarchy indexes added
- `schema/project.ts` — `pull_requests` now partial `uniqueIndex().where(...)`; `schema/central.ts` index names corrected (schema drift resolved)
- `connection.ts` — `statement_timeout` / `idle_in_transaction_session_timeout` added
- `postgres-health.ts` (logs ALTER failures), `schema-applier.ts` (atomic baseline), `audit-ops.ts` (`rowToTask`), `in-process-runtime.ts` (fails startup on bootstrap error), `async-chat-store.ts` (`DISTINCT ON`), `otel-exporter.ts` + `register-signal-routes.ts` (guarded), and the two fire-and-forget marker **writes** (`executor.ts:1882`, `self-healing.ts:811`) now `await`ed

### Partially addressed ⚠️
- **Insights / Research / Todo / Mission stores** (`remaining-ops-10.ts`, `remaining-ops-8.ts`) now guard with a clear error + operator warning instead of an opaque 500 — but the stores are still not ported to async, so those APIs remain **non-functional** in the default PG backend. Degraded gracefully, not fixed.

### Still open ❌
- **Unawaited marker reads in `.filter()`** — `scheduler.ts:1527` & `:2214`, `self-healing.ts:3693` & `:4821` are **still present**. The commit awaited the two fire-and-forget *writes* but missed these four *reads*, which are the impactful ones (`Promise !== null` is always true → the in-review file-scope-lease decision is inverted under shadow mode). Worth fixing the whole class, and switching the test mocks from `mockReturnValue` to `mockResolvedValue` so they can't pass while broken.
- **Command Center analytics** — `register-command-center-routes.ts` still has 9 unguarded `store.getDatabase()` calls (7 aggregator routes still 500 in PG mode)
- **`monitor-trait.ts:139`** — `runMonitorOnRegression` still unguarded
- **`moves.ts:702`** — handoff-to-review atomicity (workflow-work escapes the outer tx)
- **`async-persistence.ts`** `readLiveTaskRows` full-table scan + JS-side pagination; **`async-message-store.ts`** `getConversation` no LIMIT
- **`schema/_shared.ts:32`** — no per-project isolation on a shared external `DATABASE_URL` (cross-project read/write)
- **`plugins/fusion-plugin-roadmap/src/routes/roadmap-routes.js`** — stale compiled artifact still calls `getDatabase()` unguarded (the `.ts` source is fixed; rebuild needed)
- **Merge gate** — the PG suites still run only in non-blocking `full-suite.yml`, not the gate

### One structural ask
The remaining open items are all the same two classes (unguarded dashboard satellites; unawaited-marker reads). Rather than chase them site-by-site, **one PG-mode integration smoke test per dashboard router** + **a jsonb round-trip test over every `Task` field** would convert both classes into CI-caught regressions and close them for good.

gsxdsm:
## Re-review of `c55ae497d` + `70e469b63` ✅ — original 27 findings down to 2

Tracked the follow-up commits against the prior review. Nearly everything is resolved.

### Fixed since the last reconciliation ✅
- **Unawaited marker reads** — `scheduler.ts:1527`/`:2214` and `self-healing.ts:3693`/`:4821` are now `await`ed (pre-resolved maps). The impactful half of that class is closed.
- **Command Center analytics** — all 7 aggregator routes now guard with `if (store.backendMode) throw new ApiError(503, …)`; **`monitor-trait.ts`** and **`knowledge-index.ts`** likewise guarded.
- **Stale `roadmap-routes.js`** (+ `.d.ts`/maps) — deleted from the source tree.
- **Handoff-to-review atomicity (`moves.ts:702`)** — workflow-workitem upsert/cancel now accept an optional `existingTx`; `moves.ts` threads the outer move `tx` through, so they commit atomically with the handoff (no orphaned merge-gate items on rollback). Regression test updated.
- **PG tests in the merge gate** — `test:gate` now runs `@fusion/core test:pg-gate` (handoff atomicity, store-list, task-lifecycle-e2e, soft-delete FN-5233), with a `ci-workflow.test.ts` assertion guarding it.

### Mitigated, not fully resolved ⚠️
- **Multi-project isolation (`schema/_shared.ts`)** — no per-project schema namespacing yet, but a runtime **warning** now fires when an external `DATABASE_URL` is used (embedded mode, the default, is unaffected). Reasonable cutover posture; the underlying gap remains only for the advanced multi-tenant-external config.
- **Satellite stores** (Insights / Research / Todo / Mission / Command Center / monitor / knowledge-index) now **fail clean (503/skip) in PG mode** rather than crash — deliberately deferred async ports, not bugs.

### Still open ❌ (2, both performance)
1. **`async-persistence.ts`** — `readLiveTaskRows` does a full-table scan with JS-side pagination on every board hydration; push `.limit()/.offset()` to SQL.
2. **`async-message-store.ts:261`** — `getConversation` has no LIMIT; the CLI loads full history as AI context per exchange. Add a default cap.

Great turnaround — the original 27-finding review is down to two perf items plus the consciously-deferred satellite ports. Nice work.

gsxdsm:
## Follow-up: PG-backend runtime crash + analytics 500 (commit af3d0db33)

Found while smoke-testing the embedded-Postgres default on a sandboxed instance: `fn serve` **crashed ~35s after boot** and `/api/command-center/activity` **returned 500**. Both were migration leftovers where backend mode hit SQLite-only code.

### Fixes
**1. Fatal: agent-log flush killed the process.** The agent-log buffer flush/append path dereferenced the SQLite-only `store.db` getter (throws in backend mode) on an unref'd retry timer and *inside the flush-failure catch handlers* — so a handled error became an uncaught throw that exited the process. Guarded the deleted-task pre-filter + `bumpLastModified` with `!store.backendMode`, swapped every `${store.db.path}` log interpolation for the mode-safe `store.fusionDir`, and guarded the now-async `recordGoalCitations` promise against unhandled rejection. (`agent-logs.ts`, `workflow-integrity.ts`)

**2. `/api/command-center/activity` 500.** `aggregateMonitorMetrics`' async path queried `deployments`/`incidents` **unqualified with camelCase columns**; the real relations are `project.deployments`/`project.incidents` with `deployed_at`/`opened_at`/`resolved_at`. The deployments read sat *outside* the try/catch, so it 500'd the route instead of degrading. Schema-qualified + snake_case throughout. (`activity-analytics.ts`)

**3. Same class, found by sweep:** `project.experiment_session_records` (+ `::jsonb` cast on the payload update) and `project.agent_runs` were also queried unqualified on async paths. (`experiment-session-store.ts`, `register-agent-runtime-routes.ts`)

### Verification
- Re-ran the sandboxed embedded-PG instance: server **survives >70s** (was a hard exit at 35s) and the activity endpoint **returns 200**.
- New `agent-logs-backend-mode.test.ts` pins the invariant — `store.db` is never dereferenced in PG mode — across all three agent-log entry points **and** the retry-timer/catch crash vectors (FN-5893 surface coverage), not just the happy path.
- Typecheck clean for all touched packages; changeset added (`check:changesets` passes).

### Review notes
Ran a high-effort multi-agent code review on the diff. Addressed the confirmed correctness findings (silent deployments-catch now logs; redundant incidents COUNT dropped) and documented the one accepted tradeoff: in PG mode the agent-log flush skips the *secondary* deleted-task filter, but the primary purge happens at delete-time under lock (both backends) and `goal_citations.task_id` has no FK — so the worst case is a harmless orphaned-by-value row, not a violation. Deferred (non-blocking, noted for later): extract the duplicated filter helper, `Promise.all` the 5 serial monitor queries, and consider the typed Drizzle update for the experiment-session payload write.

gsxdsm:
## CI coverage for the regressed surfaces (commit 6b3d509a4)

Correcting my earlier note: the PG tests **do** run in CI — the blocking gate provisions Postgres and runs a curated `@fusion/core test:pg-gate` lane (4 suites), plus the full suite runs the rest non-blocking. The actual hole was narrower: **neither the gate lane nor any pg.test.ts covered the agent-log flush or activity/monitor-metrics surfaces** — the exact ones that just crashed/500'd — so this class merged green.

Added `agent-logs-and-monitor.pg.test.ts` (real `AsyncDataLayer` via the shared PG harness) and wired it into `test:pg-gate` so it now **blocks PRs**:
- `appendAgentLog` + `flushAgentLogBuffer` and `appendAgentLogBatch` persist without the `store.db` throw.
- `aggregateActivityAnalytics` resolves against real Postgres (no `deployments` 500).

Proven **red→green**: forcing the SQLite path in backend mode fails the flush assertion; the shipped guard passes. (Uses reserved task ids so per-task JSONL dirs don't collide under the harness's `RESTART IDENTITY` truncation.) Validated locally against `postgres:15` matching the CI service.

gsxdsm:
## 6 new commits: satellite-store PostgreSQL ports (U1–U5 + review fixes)

Ported the remaining not-yet-migrated satellite stores so their dashboard surfaces work on the embedded-PG default (was 500 → interim 503 → now 200). See the new **Follow-up: satellite-store PostgreSQL ports** section in the PR description for the full table.

- `a76d91697` U1 — workflow definitions read (`/api/workflows` 200)
- `950e89b01` U2 — mailbox/MessageStore agent send (201, was 500)
- `27c6d64bf` U3 — InsightStore (lifecycle state machine)
- `34d5f50af` U4 — ResearchStore (2 lifecycle machines + retry gate)
- `3aa26bf64` U5 — MissionStore dashboard surface
- `b0a9a66e9` review autofixes (3 corroborated findings from a 3-reviewer code-review pass)

**Gate:** `test:pg-gate` now **11 files / 59 tests**; full blocking merge gate (Lint, Typecheck, Build, `test:gate`) verified green locally. Each store has its own `*.pg.test.ts` and was verified end-to-end on a live embedded-PG instance.

**Heads-up:** CI did not auto-trigger on the recent pushes (Actions only fires the gate on `pull_request`; no `workflow_dispatch`). A no-op commit or PR reopen would kick the official checks. Real-but-low-reachability concurrency findings are recorded under **Residual Review Findings** for a transaction/optimistic-locking follow-up.

gsxdsm:
## Closing the 2 open review findings (commit 6186a9608)

Both remaining perf items from the reconciliation review are now fixed:

1. **`async-persistence.ts` — `readLiveTaskRows` full-table scan with JS-side pagination** → column filters and pagination are pushed into SQL: `WHERE column [NOT] = ?` + `ORDER BY (created_at, numeric id suffix)` (the exact JS comparator, so the SQL page equals the old client-side slice) + `LIMIT/OFFSET`. Out-of-page rows no longer pay wire transfer or per-task hydration (stall signals, PROMPT.md step sync). Unpaginated calls are unchanged. Pinned by a new `store-list.pg.test.ts` case covering the numeric-id tiebreak, offset windows, and composition with the column filter.

2. **`async-message-store.ts` — `getConversation` unbounded** → capped to the most recent 200 messages by default (`options.limit` overrides; oldest-first ordering preserved via ORDER BY DESC + LIMIT then reverse), in both the async and sqlite paths, so the CLI chat stops feeding full history into AI context every exchange. Pinned in `satellite-db-injected-stores.test.ts`.

Gates green: PG gate 23 files / 98 tests, engine-core 294, core typecheck clean.

PR #1909, opened 2026-07-04
fix(core): add task dimension to command-center token grouping

flexi767:
## The bug

`GET /api/command-center/tokens` documents and accepts `groupBy=task`, but the `task` dimension was never actually wired through the aggregator. Four enumeration sites all stopped at `model | provider | node | agent`:

- `TokenGroupBy` (core `token-analytics.ts`) did not include `"task"`.
- `groupKeyFor()` had no `case "task"`.
- `VALID_GROUP_BY` (dashboard route) rejected the value, so `resolveGroupBy()` returned `undefined`.
- `groupAttributes()` (core `otel-metrics.ts`) emitted no attribute for it.

Net effect: a caller asking for a per-task rollup silently fell back to **ungrouped grand totals** — the per-task breakdown returned zero groups, even though every task carries `tokenUsage*` and the data was right there.

## The fix

Thread `"task"` through all four sites (5 functional lines + doc/test):

- Add `"task"` to the `TokenGroupBy` union — this makes the two `switch` statements **compiler-exhaustive**, so `tsc` forces the two new cases (no silent gaps).
- `groupKeyFor`: task rows group by their task id; chat rows have no task and return `null`, mirroring the existing `node` case.
- `groupAttributes`: emit `task.id` for OTLP export, matching `node.id` / `agent.id`.
- `VALID_GROUP_BY`: accept `"task"`.

No schema or migration change — the task id is already on the row.

## Verification

- `pnpm --filter @fusion/core typecheck` and `@fusion/dashboard typecheck` — clean.
- Extended the existing `groups by provider, node, agent` core test with a `groupBy: "task"` assertion (two tasks → two groups keyed by task id, 100 / 200 tokens). Full suites green: **core 25/25**, **dashboard 325/325**.

## Summary by CodeRabbit

* **New Features**
  * Added task-based token analytics grouping, alongside existing grouping options.
  * Analytics views and metrics now include task-level breakdowns when available.

* **Bug Fixes**
  * Improved grouping behavior so task totals are reported correctly in analytics results.

* **Documentation**
  * Updated supported analytics options to reflect task grouping in endpoint and metric descriptions.

### DIFF
--- a/packages/core/src/__tests__/token-analytics.test.ts
+++ b/packages/core/src/__tests__/token-analytics.test.ts
@@ -423,7 +423,7 @@ describe("token-analytics", () => {
     );
   });
 
-  it("groups by provider, node, and agent", () => {
+  it("groups by provider, node, agent, and task", () => {
     insertTask(db, { id: "t1", inputTokens: 100, totalTokens: 100, lastUsedAt: "2026-03-01T00:00:00.000Z", modelProvider: "anthropic", nodeId: "node-1", agentId: "agent-x" });
     insertTask(db, { id: "t2", inputTokens: 200, totalTokens: 200, lastUsedAt: "2026-03-02T00:00:00.000Z", modelProvider: "openai", nodeId: "node-1", agentId: "agent-y" });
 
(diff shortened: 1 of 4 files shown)


gsxdsm:
Excellent

flexi767:
thanks :)

PR #1911, opened 2026-07-05
fix(engine): retry transient auth errors in withRateLimitRetry

flexi767:
## Problem

A long-running agent session holds its OAuth access token in memory. When the token rotates mid-run (Claude Max access tokens have an ~8 h lifetime), the next API call fails with `401 {"type":"error","error":{"type":"authentication_error","message":"Invalid authentication credentials"}}`. `withRateLimitRetry` only retries usage-limit errors and re-throws everything else immediately, so the task is marked **failed** and the operator is alerted — even though the credentials file has already been refreshed and the very next call would succeed.

We run Fusion continuously on a server against a Claude Max subscription and see this at essentially every ~8 h token boundary: whatever task or heartbeat happens to be in flight at rotation time fails with a spurious 401, then self-heals on retry (in one case the "failed" task had actually already completed and merged 11 minutes later). The engine's `notification/oauth-*` modules added in 0.55 alert on upcoming expiry, but nothing retries the in-flight call itself.

## Fix

Extend `withRateLimitRetry` with a transient-auth branch:

- `isTransientAuthError` matches `"type": "authentication_error"`, `invalid authentication credentials`, `token_expired` / `token expired`, and OAuth-scope errors.
- Auth errors get their **own small budget**: 2 retries at a flat ~5 s delay (±10 % jitter). Credential refresh completes within seconds, so the rate-limit backoff curve (30 s → 2 min) would just prolong the outage.
- Auth retries decrement the loop counter, so they never consume rate-limit attempts; the existing usage-limit path is unchanged.
- Genuinely bad credentials still propagate after ~10 s (initial + 2 quick retries), so real auth failures are not masked.
- Abort-signal handling matches the existing path (no sleep when already aborted).

## Testing

- 4 new tests in `rate-limit-retry.test.ts`: retry-then-succeed on rotation, budget exhaustion (initial + 2), auth retries not consuming rate-limit attempts (`maxRetries: 1` + auth error + 429 still succeeds), and pattern classification.
- `vitest run src/__tests__/rate-limit-retry.test.ts` — 15/15 pass; `tsc --noEmit` clean.
- We have been running this change (as a patch on the distributed bundle) in production since 2026-06-29 across 0.50 → 0.54 → 0.55; the ~8 h spurious-failure alerts stopped while real failures still surface.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved retry behavior for authentication-related failures, including credential rotation and expired/invalid tokens.
  * Transient authentication errors now use a short, consistent delay with a dedicated retry budget, without impacting existing rate-limit backoff behavior.
  * OAuth scope/permission failures are excluded and now surface immediately for re-authorization rather than being retried.
* **Tests**
  * Added coverage for transient-auth retry timing, budget exhaustion, and abort-signal cancellation.  
* **Documentation**
  * Updated release notes to reflect the revised OAuth token-rotation retry behavior.


gsxdsm:
Yes!! Thank you

PR #1943, opened 2026-07-07
fix(cli): harden PR merge path against stale-base dead-ends (empty-diff no-op + conflict escape)

flexi767:
## Problem

Two stale-base failure modes wedge the **PR-based** merge flow (`processPullRequestMergeTask` in `packages/cli/src/commands/task-lifecycle.ts`). Both pin the serial merge slot + file-scope leases on a dead-end task, starving ready work behind it.

1. **Empty-diff branch** — `createPr` throws `No commits between <base> and <head>`. The task was marked `failed`, even though a zero-commit branch is a legitimate **no-op**, not a failure. This is the PR-path analog of the engine-path fix in #1920 (`noOpResult` → done).

2. **Conflicting stale-base PR** — a PR GitHub reports as `CONFLICTING` never becomes mergeable on its own (nothing in the PR path rebases the head branch). The task sat in `awaiting-pr-checks` **unbounded**: `awaiting-pr-checks` is not in `TRANSIENT_MERGE_STATUSES`, so the age-based stall escape never catches it, and the only other escape — `getInReviewStallReason`'s `merge-retries-exhausted` — never fired because the PR path never incremented `mergeRetries`.

## Fix

1. `finalizeNoOpMergeTask` finalizes empty-diff branches as a terminal **DONE**, mirroring the engine's canonical `noOpResult` decision (`merged:false, noOp:true`). Both the group and per-task empty-diff sites route through it.

2. Each `CONFLICTING` poll now counts against `mergeRetries`, so the existing `merge-retries-exhausted` escape disposes the task after `maxAutoMergeRetries` cycles. Detection is gated on `prInfo.mergeable === "conflicting"` (the genuine unresolvable signal), **not** `blockingReasons` (status/review/checks only). Pending / `behind` PRs still wait indefinitely — checks legitimately run, and `behind` is resolved by the pre-merge rebase.

No new machinery — fix #2 reuses the existing stall-escape counter rather than a hand-rolled fail branch.

## Tests

`packages/cli/src/commands/__tests__/task-lifecycle.test.ts`:
- empty-diff branch → no-op DONE (`moveTask(id,"done")`, `updateTask({status:null,mergeRetries:0})`, never `failed`)
- conflicting not-ready PR → `mergeRetries` bumped
- non-conflicting not-ready PR → `mergeRetries` **not** bumped

`pnpm --filter @runfusion/fusion test` (task-lifecycle suite 39/39), `typecheck`, and `build` all green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Tasks with no commits to merge now finish cleanly instead of being marked as failed.
  * Merge checks now handle conflicting or unstable merge states more reliably, reducing unnecessary retries and improving task progression.

### DIFF
--- a/packages/cli/src/commands/__tests__/task-lifecycle.test.ts
+++ b/packages/cli/src/commands/__tests__/task-lifecycle.test.ts
@@ -668,6 +668,73 @@ describe("processPullRequestMergeTask", () => {
     expect(github.getPrMergeStatus).toHaveBeenCalledWith("owner", "repo", 7);
   });
 
+  it("counts a conflicting stale-base PR against mergeRetries so the stall escape fires", async () => {
+    const task: MockTask = {
+      id: "FN-9020",
+      title: "test",
+      description: "desc",
+      column: "in-review",
+      mergeRetries: 1,
+    };
+    const branch = getTaskBranchName(task.id);
+    const store = makeStore(task);
+    execMock.mockImplementation(() => "");
+
+    const existingPr = { number: 9, url: "https://github.com/x/y/pull/9", status: "open" as const, headBranch: branch, baseBranch: "main" };
+    const github = {
+      findPrForBranch: vi.fn(async () => existingPr),
+      createPr: vi.fn(),
+      getPrMergeStatus: vi.fn(async () => ({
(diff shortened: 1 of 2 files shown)


gsxdsm:
This looks good please update to ready for review when comfortable

PR #1958, opened 2026-07-08
fix: surface model-lane drift when a workflow's default model changes

flexi767:
## Problem

A task's model fields (`modelProvider`/`modelId` and the planning/validator
equivalents) are snapshotted once at task-creation time from the workflow's
model-lane default in `workflow_settings`. Nothing re-syncs, flags, or
surfaces drift when that default is later changed.

Concretely: the `builtin:coding` workflow's execution default was
`claude-sonnet-4-6` until it was corrected on 2026-07-05. Every task created
before that correction stayed permanently, invisibly pinned to the stale
model id — 52 tasks were found silently stuck on it.

## Fix

- `TaskStore.getModelLaneDrift(workflowId, before, after)` (`packages/core/src/store.ts`):
  read-only diff over the three model lanes (execution/planning/validator).
  For any lane whose provider+modelId actually changed, it lists the
  non-terminal (`column` not `archived`/`done`, not soft-deleted) tasks on
  that workflow still pinned to the old value. Never mutates `tasks`.
- Wired into `PATCH /workflows/:id/setting-values`
  (`packages/dashboard/src/routes/register-workflow-routes.ts`): captures a
  `before` snapshot, runs the existing `updateWorkflowSettingValues`
  unchanged, then attaches an optional `modelDrift` field to the response
  when a lane change orphans existing tasks. Backward compatible — the field
  is only present when non-empty.
- Operators can act on the surfaced drift via the existing
  `POST /tasks/batch-update-models` endpoint; this change intentionally does
  not auto-rewrite any task (avoids touching tasks mid-execution).

## Testing

- New tests in `packages/core/src/__tests__/workflow-settings.test.ts`
  (`TaskStore.getModelLaneDrift`): verifies a task pinned to a changed lane's
  old value is surfaced, a task already on the new value and a `done`-column
  task are excluded, and an unrelated/unchanged lane produces no drift entry.
- `packages/core`: `npx vitest run src/__tests__/workflow-settings.test.ts` — 24/24 pass.
- `npx tsc --noEmit` clean in both `packages/core` and `packages/dashboard`.

## Summary by CodeRabbit

* **New Features**
  * Workflow setting updates can now return a lane-based model drift summary (execution, planning, validator) showing task IDs still pinned to the previous model configuration.
  * The PATCH workflow setting response conditionally includes `modelDrift` when impacted tasks are found.
* **Bug Fixes**
  * Drift detection now compares a consistent “before” snapshot with the updated values to avoid stale pairing.
  * “No workflow selection” tasks are handled correctly based on the default-workflow behavior.
* **Tests**
  * Added coverage for lane drift across model changes and null-selection inclusion rules.
* **Documentation**
  * Added a release note entry for the change.

### DIFF
--- a/.changeset/fix-model-lane-drift-visibility.md
+++ b/.changeset/fix-model-lane-drift-visibility.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Warn when changing a workflow's model surfaces tasks still pinned to the old model, including default-workflow tasks.
+category: fix
+dev: PATCH /workflows/:id/setting-values now returns `modelDrift` for execution/planning/validator lanes. Drift baseline is captured inside the settings write transaction (no stale-read race), and default-workflow patches pass `includeNullSelection` so no-workflow-selection tasks are counted. New `TaskStore.updateWorkflowSettingValuesWithPrevious` and `getModelLaneDrift(..., { includeNullSelection })`.
--- a/packages/core/src/__tests__/workflow-settings.test.ts
+++ b/packages/core/src/__tests__/workflow-settings.test.ts
@@ -374,4 +374,87 @@ describe("TaskStore.getModelLaneDrift", () => {
 
     expect(store.getModelLaneDrift(BUILTIN_CODING, before, after)).toEqual([]);
   });
+
+  // FN-5893: the invariant holds across ALL model lanes, not only `execution`.
+  it("flags the planning lane's pinned tasks when the planning model changes", async () => {
+    const store = harness.store();
+    await store.updateWorkflowSettingValues(BUILTIN_CODING, PROJECT, {
+      planningProvider: "anthropic",
+      planningModelId: "claude-opus-4-6",
+    });
+    const before = store.getWorkflowSettingValues(BUILTIN_CODING, PROJECT);
+    const pinned = await store.createTask({
+      description: "pinned to old planning model",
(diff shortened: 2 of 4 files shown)


gsxdsm:
Addressed review feedback in 1762229:

**Greptile P1 — default-pool tasks skipped** (`store.ts`): `getModelLaneDrift` now takes an explicit `{ includeNullSelection }` option. The setting-values route resolves the project default workflow id and passes `includeNullSelection: workflowId === defaultWorkflowId`, so no-workflow-selection tasks that resolve through the default are counted. Added a regression test asserting both the excluded (default) and included (opt-in) cases.

**Greptile P2 — concurrent writes use stale baseline** (`register-workflow-routes.ts`): added `TaskStore.updateWorkflowSettingValuesWithPrevious`, which captures the `previous` snapshot inside the same immediate write transaction that produces `stored`. The route now uses it, so `before`/`after` are always a consistent pair from one snapshot.

**CodeRabbit — compare effective lane values** (`register-workflow-routes.ts`): intentionally skipped. Model-lane declarations carry no declaration-level `default` (KTD-7, `builtin-workflow-settings.ts:204`), so `resolveEffectiveSettingValues` returns the same raw `null` for an unset lane — comparing effective values is a no-op for these keys. A lane transitioning unset→set also has no single well-defined `from` model to diff against.

**CodeRabbit nitpick — broaden coverage**: added planning-lane, validator-lane, and null-selection/default-workflow test cases.

PR #1963, opened 2026-07-08
fix(FUX-039): staff-engineer review fixes for FUX-022 runtime-fallback UI patch

Automata-intelligentsia:
## Summary

Addresses the Staff Engineer pre-landing review findings on merged PR #1957 (FUX-022 runtime-fallback badge/toast feature). All 4 findings were independently re-verified against a fresh bootstrap of `main` (not trusting a prior unmerged branch attempt) before fixing.

1. **`resolvePluginRuntime()` mislabeled found-but-failed-to-init as `"not_found"`.** The branch where a plugin's registration exists but `createRuntimeContext(...)` returns falsy now correctly returns `reason: "init_error"` instead of `"not_found"` (which was indistinguishable from "never registered"). Extended the existing unit test suite to assert all 3 `FallbackReason` values are independently reachable and pairwise-distinct.

2. **Viewport gating was defeated on card surfaces.** `ActiveAgentsPanel` and `AgentsView` both hardcoded `isInViewport={true}` when rendering `RuntimeFallbackBadge`, so every rendered card polled regardless of actual visibility — defeating the entire purpose of the `isInViewport` gate. Wired real `IntersectionObserver`-based viewport gating on both surfaces (`ActiveAgentsPanel`'s `LiveAgentCard` tracks its own visibility; `AgentsView` uses one shared observer instance keyed per card so the board and list surfaces never share visibility state). Added desktop + mobile regression tests.

3. **Toast-dedupe was per-component-instance, not shared.** `useRuntimeFallbackStatus`'s dedupe key lived only in a per-hook-instance `useRef`, so two simultaneously-mounted card surfaces polling the same task (e.g. its board card and list card both visible at once) would each independently fire their own toast for the same fallback session. Replaced with a module-level shared claim store (bounded, with a test-only reset hook) so only the first instance to observe a given event, process-wide, fires the toast. Added a cross-instance regression test.

4. **Lint-breaking `eslint-disable` comment** — independently confirmed already resolved on `main` prior to this PR; no code change included.

## Test plan

- `pnpm --filter @fusion/engine exec vitest run src/__tests__/runtime-resolution.test.ts` — 25/25 pass
- `pnpm --filter @fusion/dashboard exec vitest run app/components/__tests__/ActiveAgentsPanel.test.tsx` — 17/17 pass
- `pnpm --filter @fusion/dashboard exec vitest run app/components/__tests__/RuntimeFallbackBadge.test.tsx` — 9/9 pass
- `pnpm --filter @fusion/dashboard exec vitest run app/components/__tests__/AgentsView.test.tsx` — 132/134 pass (1 pre-existing, unrelated CSS-fixture failure confirmed via `git stash` prior to this change)
- `pnpm --filter @fusion/engine exec tsc --noEmit` — clean
- `pnpm --filter @fusion/dashboard exec tsc --noEmit` — clean
- `eslint` scoped to all changed files — 0 errors

🤖 Generated with an autonomous coding agent

## Summary by CodeRabbit

* **New Features**
  * Agent cards now load fallback/runtime status only when visible on screen, improving responsiveness in long lists and board views.
  * If the same fallback event appears in multiple places, users now see a single toast instead of duplicates.

* **Bug Fixes**
  * Runtime failures are now reported with a more accurate reason when a runtime exists but cannot initialize.
  * Off-screen agent cards no longer trigger unnecessary polling or badge updates.

### DIFF
--- a/packages/dashboard/app/components/ActiveAgentsPanel.tsx
+++ b/packages/dashboard/app/components/ActiveAgentsPanel.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState } from "react";
+import { useEffect, useRef, useState } from "react";
 import { Activity, FileText } from "lucide-react";
 import { useTranslation } from "react-i18next";
 import type { Agent } from "../api";
@@ -24,6 +24,32 @@ function LiveAgentCard({ agent, projectId, onSelect, onOpenTaskLogs }: LiveAgent
   const { t } = useTranslation("app");
   const { entries, isConnected } = useLiveTranscript(agent.taskId, projectId);
   const [task, setTask] = useState<TaskDetail | null>(null);
+  const cardRef = useRef<HTMLDivElement>(null);
+  // FUX-039: real viewport gating (matching TaskCard.tsx's pattern) so
+  // RuntimeFallbackBadge's 30s poll only runs while this card is actually
+  // visible, rather than hardcoding isInViewport={true} for every mounted
+  // card regardless of scroll position.
+  const [isInViewport, setIsInViewport] = useState(false);
+
+  useEffect(() => {
+    if (typeof IntersectionObserver === "undefined") {
+      setIsInViewport(true);
+      return;
+    }
+
+    const element = cardRef.current;
+    if (!element) return;
+
+    const observer = new IntersectionObserver(
+      ([entry]) => {
+        setIsInViewport(entry?.isIntersecting ?? true);
+      },
+      { rootMargin: "200px" },
+    );
+
+    observer.observe(element);
+    return () => observer.disconnect();
+  }, []);
 
   // Poll the agent's task so the empty state can show real run progress
   // (current step, executor model) instead of just "Connecting..." while the
@@ -101,6 +127,7 @@ function LiveAgentCard({ agent, projectId, onSelect, onOpenTaskLogs }: LiveAgent
 
   return (
     <div
+      ref={cardRef}
       className="live-agent-card"
       onClick={handleSelect}
       onKeyDown={handleKeyDown}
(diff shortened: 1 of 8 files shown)


gsxdsm:
## Closing as superseded by `main`

This PR's work has landed on `main` through a different path, so there's nothing left here to merge. Posting the findings for the record.

### What happened

PR #1963 branched from `b7b1b71` to address the staff-engineer review of FUX-022. While it sat open, `main` received the same fixes plus every review comment:

| Finding | Landed on `main` |
| --- | --- |
| `resolvePluginRuntime()` → `init_error` reason | `c3c726cff` |
| Real `IntersectionObserver` viewport gating on agent cards | `8a35b0bc4` |
| Cross-instance toast dedupe + regression test | `247908141` |
| `init_error`/`not_found`/`factory_error` three-way coverage | `0bed997af` / `c3c726cff` |

### Review comments — all addressed on `main` by `003948033` ("harden runtime-fallback agent-card viewport gating")

- **Greptile (FNXC comment format)** — `ActiveAgentsPanel.tsx` and `AgentsView.tsx` gating comments are now `FNXC:RuntimeFallback`-prefixed and greppable. ✅
- **Greptile (test-only reset leaks into prod bundle)** — `__resetRuntimeFallbackToastDedupeStoreForTests` is now guarded with `import.meta.env.MODE !== "test"` (the Vite convention used in `App.tsx`). ✅
- **CodeRabbit (cached ref callbacks never evicted)** — `registerAgentCardRef` now caches one stable callback per viewport key (which also closes the fresh-closure-per-render re-render loop) **and** deletes the cache entry on `el === null`. ✅
- **CodeRabbit (shared test scaffolding)** — optional nit, not required.

### Diff vs `main`

Empty. Resetting this branch to `origin/main` produces zero code delta — every file the PR touched is identical to `main`.

### One follow-up

`003948033` landed without a changeset, and it affects published `@runfusion/fusion`. A separate tiny PR will carry the missing patch changeset so the fix shows up in release notes.

Thanks for the review work — it directly shaped `003948033`. 🤖

PR #2098, opened 2026-07-14
Fix required empty JSONB migration values

plarson:
## Summary

- preserve exact empty and whitespace-only SQLite text when migrating into required PostgreSQL `jsonb` columns without defaults
- retain existing nullable-empty → SQL `NULL` and required-with-default → declared-default behavior
- canonicalize converted JSON scalars before source checksum comparison so source and PostgreSQL representations verify identically
- add regression coverage for empty, whitespace-only, malformed, scalar, nullable-empty, and defaulted legacy JSON values

Rebased onto current `main` after #2088 merged, so this PR contains only the focused compatibility fix.

## Why

A real Atlas Notes first-boot cutover stored `workflows.ir` as an empty SQLite string while the PostgreSQL target is `jsonb NOT NULL` without a default. The migrator converted that value to SQL `NULL`, aborting the fail-closed cutover.

The conversion must preserve non-null source data rather than inventing a workflow document or weakening the target constraint.

## Verification

- PostgreSQL migrator suite: 24/24 passed against PostgreSQL 15
- `pnpm --filter @fusion/core build`
- `pnpm lint`
- `pnpm check:changesets --strict`
- independent review passed with no security or logic findings
- an earlier real Atlas Notes cutover advanced past the `workflows.ir` failure and exposed the separate legacy mapping/checksum issues subsequently fixed on `main`

## Post-deploy validation

- retry the preserved Atlas Notes SQLite cutover
- confirm `workflows.ir` no longer fails the `NOT NULL jsonb` insert
- confirm nullable empty JSON remains SQL `NULL`
- confirm required/defaulted JSON materializes the declared default

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed SQLite → PostgreSQL migrations for required `jsonb` columns to preserve legacy empty/whitespace, malformed, and scalar values instead of coercing to `NULL` or unintended defaults.
  - Improved JSONB value canonicalization during migration verification to match PostgreSQL behavior.
  - Corrected migration handling for `activityLog.metadata` so it is migrated as `NULL` as expected.
- **Tests**
  - Expanded migrator coverage with new SQLite fixture data and validation for `workflows.ir` migration.

### DIFF
--- a/.changeset/pg-migrator-required-empty-jsonb.md
+++ b/.changeset/pg-migrator-required-empty-jsonb.md
@@ -1,0 +1,7 @@
+---
+"@fusion/core": patch
+---
+
+summary: Preserve legacy empty JSON text during PostgreSQL cutover.
+category: fix
+dev: Required jsonb columns without defaults now retain empty, whitespace-only, malformed, and scalar SQLite values without weakening nullable/default handling.
--- a/packages/core/src/__tests__/postgres/sqlite-migrator.test.ts
+++ b/packages/core/src/__tests__/postgres/sqlite-migrator.test.ts
@@ -149,6 +149,20 @@ CREATE TABLE IF NOT EXISTS researchRuns (
 );
 `;
 
+/** Legacy workflow rows could persist an empty IR before write validation tightened. */
+const WORKFLOWS_SQLITE_DDL = `
+CREATE TABLE IF NOT EXISTS workflows (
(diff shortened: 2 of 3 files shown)


plarson:
Superseded by #2099, which merged the canonical fix plus the fail-closed unparsed-default follow-up and the correct @runfusion/fusion changeset.

PR #2220, opened 2026-07-16
fix(build): move pnpm overrides to pnpm-workspace.yaml for pnpm 11 readiness

cachrisman:
pnpm 10.33 warns on every install and build:

  [WARN] The "pnpm" field in package.json is no longer read by pnpm. The
  following keys were ignored: "pnpm.onlyBuiltDependencies", "pnpm.overrides"

The wording overstates it for the pinned version: on pnpm 10.33 the field is still honoured, so nothing is broken today. It is a forward-looking deprecation notice, and pnpm 11 makes it real — there the overrides silently stop applying.

Reproduced on pnpm 11 with the workspace pins declared only in package.json:

  overrides block dropped from pnpm-lock.yaml
  @types/node  ^25.5.2 -> ^22.0.0   (25.5.2 -> 22.20.1)
  zod          4.3.6   -> 3.25.76
  @types/node fragments into 5 versions across the workspace

pnpm-workspace.yaml is the supported home and is honoured by both 10.33 and 11, so moving `overrides` there is safe on the pinned version and correct for the next major.

Verified on pnpm 10.33: `pnpm install --lockfile-only` leaves pnpm-lock.yaml byte-identical and its `overrides` block is now sourced from pnpm-workspace.yaml. Verified on a minimal repro that pnpm 10.33 applies yaml overrides (ms -> 2.0.0).

Scope is deliberately limited to `overrides`. `ignoredBuiltDependencies` and `onlyBuiltDependencies` are left in place: they are also honoured on 10.33 and ignored on 11, but the workspace list is not a superset of the package.json one (it omits `electron`, removed in c7641f9d when electron became a runtime dependency, and three @embedded-postgres arches outside the `supportedArchitectures` cpu list curated in 1b9dce22). Reconciling those is a maintainer call about intended behaviour, not a mechanical move, so it is left for a follow-up rather than guessed at here.

No changeset: workspace/build config only, behaviour-preserving on the pinned pnpm, and it does not affect published @runfusion/fusion.

## Summary by CodeRabbit

* **Chores**
  * Updated workspace dependency version overrides for improved consistency across packages.
  * Consolidated dependency configuration in the workspace settings.


gsxdsm:
Thanks can you address feedback

PR #2256, opened 2026-07-17
fix(engine-lifecycle): give executors an honest "blocked" exit from fn_task_done

gsxdsm:
## What & why

FN-8141 ("Update pi SDK to latest and verify Kimi K3 end to end") was impossible as specced — pi 0.80.x removed `AuthStorage`/`ModelRegistry` APIs, so every SDK bump broke the build. The executor correctly reverted its work and filed follow-up FN-8145 — but had **no sanctioned way to end the task in a blocked state**. `fn_task_done` only expressed success: the bulk-completion gate refused it, the requeue budget re-ran the doomed task 5 times, and the only remaining affordance (mark every step `skipped`, then complete) made `isTaskComplete()` return true. Self-healing then promoted the "complete" todo to in-review and the AI merger finalized the empty diff as `done`. **The honest path must be cheaper than the laundering path.**

This adds a first-class **blocked** outcome to the executor's `fn_task_done` tool.

## Change

- `fn_task_done` gains `outcome: "completed" | "blocked"` (default `"completed"`), optional `blockedBy: string[]`, and `reason` (required when blocked).
- `outcome="blocked"` runs **before** every completion gate (completion blocker, verdict providers, worktree invariants, bulk-completion refusal) — blocked is not a completion claim, so none of those gates apply.
- Parks the task `failed` with `error = "BLOCKED: <reason>"`, following the FN-7863 `EXECUTION_DISPATCH_LOOP_EXHAUSTED` park convention: **steps keep their true statuses** (no auto-done, no auto-skip), worktree/branch preserved. It does **not** call `onDone()`, so the executor's existing `status === "failed"` post-loop branch honors the park instead of handing off to review.
- `blockedBy` is recorded as real `task.dependencies` edges (unioned with existing) so the task requeues behind the blocker.
- Emits run-audit `task:execution-blocked-parked` with ids/outcomes-only metadata (`taskId`, `blockedBy` ids, `hasReason` boolean — **never** the reason prose).
- Executor + core prompt guidance and the `bulk-step-completion-without-review` refusal message now name the blocked exit as **the** correct action when work cannot proceed, replacing skip-and-done. `PREMISE STALE:` skip guidance is preserved for genuinely-stale premises.

## Surface enumeration

- **fn_task_done tool schema + handler** (`packages/engine/src/executor.ts`): blocked branch added at the top of `execute`, before all gates.
- **Refusal/requeue machinery**: `formatTaskDoneRefusal` for `bulk-step-completion-without-review` now points at the blocked exit; the requeue-budget path is untouched (blocked never enters it).
- **Executor prompt text**: turn-ending rules, the "Cannot proceed" section, the preflight/stale-premise escape hatch (now explicitly distinguishes stale-premise skip from blocked).
- **Core prompt mirror** (`packages/core/src/agent-prompts.ts`): same turn-ending + cannot-proceed guidance.
- **Tool reference doc** (`packages/cli/skill/fusion/references/engine-tools.md`): `fn_task_done` params updated. (grep for `fn_task_done` confirmed the only executable tool schema is in executor.ts; CLI/pi surfaces re-export it, no separate schema copy.)
- **Self-healing**: verified a blocked-parked row is NOT auto-recovered by `recoverStrandedCompletedTodoTasks` — its steps are not all done/skipped and `task.error` is set (both are hard filters in the sweep).
- **Run Audit inventory** (`AGENTS.md`): documented the new event.

## Test evidence

New `packages/engine/src/__tests__/executor-task-done-blocked.test.ts` (8 tests) asserts the invariant across surfaces:

```
pnpm --filter @fusion/engine exec vitest run \
  src/__tests__/executor-task-done-blocked.test.ts \
  src/__tests__/executor-task-done-invariant.test.ts \
  src/__tests__/gating-classifications.test.ts \
  src/__tests__/reliability-interactions/execute-requeue-loop-guard.test.ts --reporter=dot
→ Test Files 3 passed | Tests 138 passed (0 failed)
```

Coverage: blocked parks failed with `BLOCKED:` error and does **not** trip the bulk-completion refusal or requeue to todo; `blockedBy` unioned into `dependencies`; `task:execution-blocked-parked` emitted with metadata that excludes the reason prose; steps left untouched; empty `reason` rejected without parking; `completed` outcome unchanged (still marks steps done, no blocked audit); and `recoverStrandedCompletedTodoTasks` never promotes a blocked-parked row.

### Note on `pnpm verify:fast`

`verify:fast` currently fails at the workspace build step due to **pre-existing** type errors in `packages/engine/src/auth-storage.ts`, `pi.ts`, and `provider-registration.ts` — the exact FN-8142 pi SDK API break that FN-8145 will fix. These are present on the base branch and untouched by this PR. Verified instead that this change introduces **zero** new type errors (`tsc` diff before/after, engine and core both clean) and that all scoped tests are green.


gsxdsm (commented):

**Verdict: LGTM — the honest blocked exit is implemented correctly; safe to merge once main is green.**

Verified independently:
- Ran the new `executor-task-done-blocked.test.ts` suite in the worker worktree: **8/8 passed**.
- Red CI (Typecheck/Build/Gate) is the **pre-existing main breakage** from 5a60643c0 (FN-8142 pi migration without the lockfile bump), being fixed separately; this diff adds no new engine errors.

What I like: blocked runs BEFORE every completion gate (correctly modeled as "not a completion claim"), steps keep true statuses, `blockedBy` unions into real `task.dependencies`, run-audit metadata is ids-only (verified the test asserts reason prose never lands in metadata), and the refusal message + system prompt now advertise the blocked exit while explicitly distinguishing it from the PREMISE-STALE skip path. The AGENTS.md run-audit inventory entry follows the house format.

Findings:
1. **(question, non-blocking)** The blocked park sets `status: "failed"` without a `moveTask`, relying on the outer loop's honor-park branch (`task parked failed during no-fn_task_done retry`) to keep the row parked. In the FN-8141 incident that park was later bounced to `todo` by the pause-abort machinery. The not-all-done steps + `task.error` do keep the recovery sweeps away (tested), but consider a follow-up asserting the pause-abort classifier also honors a `BLOCKED:`-error row — that was the exact bounce path that undid the honest park last time.
2. **(nit)** `agentId: "executor"` on the audit event — matches existing synthetic-run conventions elsewhere in the file, fine.

PR #2257, opened 2026-07-17
fix(engine-lifecycle): stranded-completed promoters must respect failure provenance (FN-8141)

gsxdsm:
## What & why

FN-8141 laundered a failed task into `done`. The executor correctly parked the task `failed` ("task parked failed during no-fn_task_done retry" / "fn_task_done refusal retry budget exhausted"), the pause-abort machinery bounced it to `todo`, and ~12 minutes later `recoverStrandedCompletedTodoTasks` promoted it to `in-review` because every step was done/skipped — overriding the honest failure park. From there the AI merger found an empty diff and finalized it as a no-op `done`, with no reviewer ever seeing it.

Existing exclusions (`task.error`, `evaluateNoCommitsNoOpFinalize().blocked`, active statuses, refreshing review state) all missed it because the failure provenance lived **only in the durable task log** by the time the promoter ran — status/error had been cleared by the pause-abort bounce.

This PR restores the invariant: **a stranded-completed promoter must not promote a task whose most recent execution lifecycle ended in a failure/refusal park.**

## Change

- New pure, unit-testable evaluator `evaluateCompletedPromotionFailureProvenance(task)` in `@fusion/core` (next to `no-commits-finalize-guard.ts`). It scans the task-log **tail** (bounded to 250 entries) and lets the **most-recent execution-outcome marker** decide: a failure/refusal park → `{ blocked: true, reason: "failure-provenance" }`; a fresh clean completion (`Task marked done by agent` / `All steps complete — implicit fn_task_done`) that appears more recently supersedes an earlier park; zero failure markers → not blocked. Recency is by construction, so a failure that predates a newer clean execution is never reached.
- Both self-healing sweeps (`recoverCompletedTasks` stuck-in-progress **and** `recoverStrandedCompletedTodoTasks` stranded-todo) fetch the full task for candidates that already cleared the cheap slim filters (slim listings strip `log`) and skip when blocked, emitting a **deduped** `task:reconcile-stranded-completed-no-action` run-audit event (ids/outcomes-only: `taskId`, `reason`, `sweep`, `marker?`).
- Defense-in-depth: the shared executor `recoverCompletedTask` chokepoint — which the sweeps AND the executor's own unpause/`resumeOrphaned` fast-paths all funnel through — also refuses a provenance-blocked promotion, so no route can launder a failed park.

**Escape hatch (documented in FNXC comments):** an operator retrying/moving the task starts a fresh execution whose clean-completion marker supersedes the failure park, clearing the block with no code change.

[...]

PR #2255, opened 2026-07-17
fix: require compatible Pi SDK

plarson:
## Summary
- raise the Pi SDK minimum from 0.80.6 to 0.80.8 across Fusion packages
- refresh the lockfile to a compatible Pi SDK release
- add a patch changeset for the published CLI

## Test plan
- `CI=true pnpm --filter @fusion/engine build`
- `CI=true pnpm build`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved compatibility by aligning Pi SDK requirements with the model runtime and credential integration.
  - Updated the required `@earendil-works/pi-ai` and `@earendil-works/pi-coding-agent` versions to `^0.80.8` across the CLI, engine, dashboard, and Claude integration.

- **Chores**
  - Bumped related dependency versions to `^0.80.8`.
  - Published a patch release for `@runfusion/fusion` to include the compatibility fix.

PR #2254, opened 2026-07-17
fix(core-lifecycle): block empty-diff finalize of tasks with skipped steps (FN-8141)

gsxdsm:
## What & why

FN-8141 (\"Update pi SDK to latest and verify Kimi K3 end to end\") was **laundered into `done` despite producing zero net changes**. The pi SDK bump kept breaking `verify:fast`, the work was reverted 5×, and the agent used the sanctioned skip affordance to mark **Testing & Verification** and **Documentation & Delivery** `skipped`. `isTaskComplete()` counts `skipped` as complete, so:

1. self-healing `recoverStrandedCompletedTodoTasks` promoted the todo task to in-review (all steps done/skipped),
2. the AI merger saw an empty diff vs main → \"finalizing as no-op\" → `done` with `mergeConfirmed:true`,
3. no reviewer ever saw it (skipped steps request no review; the merge-review pass reviews an empty diff).

The only existing guard, `evaluateNoCommitsNoOpFinalize` (FN-6461), missed it **twice**: it only fired when `noCommitsExpected === true` (FN-8141 was commit-expected — the branch was empty because work was *reverted*), and even then only blocked when `incomplete >= done` (FN-8141 had 3 done vs 2 skipped).

## The fix

Generalize the guard (same exported name/shape — every finalize lane keeps working) so a **zero-diff/no-op finalize is blocked whenever ANY step is `skipped`**:

- a **verification-flavored** skipped step (name matching `/test|verif|qa|review/i`) blocks **unconditionally**;
- any **other** skipped step blocks **unless** every non-skipped step is `done` **AND** the task is the legacy `noCommitsExpected` ops shape;
- the legacy FN-6461 ratio rule (`noCommitsExpected` + `incomplete >= done`) is preserved for pending/in-progress incomplete work;
- blocked evaluations return a precise `reason` naming the skipped steps.

Legitimate shapes still pass: all-done no-skip empty diffs (left to the lineage-proof work), zero-step tasks, and `noCommitsExpected` ops tasks with a minor non-verification skipped tail.

## Surface enumeration

The guard is the single chokepoint used at every zero-diff finalize lane; all already honor `.blocked`/`.reason`, so the core change fixes each surface:

- `packages/engine/src/merger-ai.ts` ~1116 — AI empty-merge lane
- `packages/engine/src/merger.ts` ~6261 / ~7354 / ~7658 — merger empty-own-diff + no-op lanes
[...]

PR #2258, opened 2026-07-17
fix(engine-lifecycle): veto no-op finalize after failed-incomplete executor signal (FN-8141)

gsxdsm:
## What & why

FN-8141 (\"Update pi SDK … verify Kimi K3 end to end\") was **laundered into `done`** despite producing zero net changes. The executor reverted the impossible work 5 times; the planner overseer emitted `stage=executor signal=failed` (\"Executor stage parked failed with work incomplete\") **twice**, then an hour later — because the overseer is **stage-scoped and memoryless** — classified the same task `stage=merger signal=progressing` and let the AI merger's **empty no-op finalize** promote it to `done`. No reviewer ever saw it (skipped steps request no review; the merge-review pass reviews an empty diff).

This restores the invariant: **a task whose most-recent executor signal is failed-with-incomplete-work, with no subsequent green completion, must not reach `done` via a zero-diff no-op merge finalize.**

## Change

Two pure, unit-testable, never-throw functions (`packages/engine/src/overseer-noop-finalize-veto.ts`), following the FN-7514 `evaluateOverseerHumanControl` precedent (pure predicate + ids/outcomes-only audit metadata):

- **`deriveExecutorSignalMemory`** — reconstructs the most-recent executor signal from the **durable `overseer:intervention` timeline the overseer already writes** (no new persisted column / migration; \"the existing oversight state storage the controller uses\"). A later green executor observation supersedes an earlier failure, which is how \"no subsequent execution completed green\" is derived. Keys on the newly-exported `EXECUTOR_FAILED_INCOMPLETE_REASON` constant (already load-bearing for FN-7577 feed dedup) as the single source of truth.
- **`evaluateNoOpFinalizeExecutorVeto`** — the veto decision.

Wired into the AI **empty-merge lane** (`merger-ai.ts`), composed with (and independent of) the FN-6461 no-commits guard: on veto it sets `error`, writes a durable log entry, emits `overseer:no-op-finalize-vetoed-failed-executor`, and moves the task back to `todo` with progress preserved — mirroring the FN-6461 blocked lane. The move-to-todo transition takes the task out of the merge lane, so the event isn't re-emitted every poll (equivalent to the `overseer:oversight-withheld-human-control` per-(taskId, reason) dedup).

Independent of the sibling Task 2 merger-layer lineage guard: both can fire; **either alone stops FN-8141**.

## Surface enumeration

- **Empty (zero-diff) no-op merge** — vetoed when memory is failed-incomplete. ✅
- **Non-empty (real squash landed) merge** — **never** vetoed (reviewers/merge review cover real diffs). ✅
- **failed-incomplete then later green execution** — memory superseded → no veto. ✅
- **No executor memory / store lacks the async reader** — fail open (no veto); FN-6461 + sibling guard remain the safety net. ✅
- **user-paused / autoMerge:false / approval-blocked** — defers to FN-7514 human-control; never fights those semantics. ✅
- **Timeline entry filtering** — only `stage=executor` + `action=observe` entries count as signals (retry/escalate messages ignored). ✅

## Test evidence

`pnpm --filter @fusion/engine exec vitest run src/__tests__/overseer-noop-finalize-veto.test.ts src/__tests__/merger-ai.test.ts --silent=passed-only --reporter=dot`

```
Test Files  2 passed (2)
     Tests  48 passed (48)
```

Covers: derivation (failed→veto, failed-then-green→no-veto, non-executor/non-observe ignored, empty→null); evaluator (veto, green, no-memory, non-empty never-vetoed, user-paused defer, autoMerge:false defer, missing-task fail-open); and an engine integration test driving an **FN-8141-shaped** empty merge through `runAiMerge` → asserts move-to-todo + `overseer:no-op-finalize-vetoed-failed-executor` audit event + main untouched, plus the later-green case finalizing done.

`@fusion/core` builds clean (`pnpm --filter @fusion/core build`).

**Note on `pnpm verify:fast`:** it currently fails to build `@fusion/engine`, but **only** in `auth-storage.ts` / `pi.ts` / `provider-registration.ts` — the pre-existing pi-SDK breakage that *is* this incident (pi 0.80.x removed `AuthStorage`/`ModelRuntime`/`ModelRegistry`; tracked as FN-8145). Verified identical errors with my changes stashed; **my diff touches none of those files and adds zero new type errors** (tsc reports all program errors before failing — none were in my files).

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented zero-change tasks from being incorrectly finalized when the latest executor attempt failed with unfinished work.
  - Preserved task progress and returned affected tasks to **Todo** for continued processing.
  - Allowed finalization to proceed after a subsequent successful executor result.
  - Maintained existing human-control and non-empty merge behavior.
  - Added audit visibility for blocked finalization events.


gsxdsm (commented):

**Verdict: LGTM — correct overseer-layer backstop; safe to merge once main is green.**

Verified independently:
- Veto/derivation suite: **11/11 passed**; merger-ai FN-8141 integration tests: **2/2 passed** (ran in the worker worktree).
- Red CI is the pre-existing FN-8142 pi migration breakage on main, fixed separately; this diff touches none of the broken files.

What I like: no new persisted column — memory is derived from the durable `overseer:intervention` timeline the controller already writes; pure predicates following the FN-7514 precedent; constant veto-reason string for stable (taskId, reason) audit dedup; non-empty merges are never vetoed; explicit deferral to the human-control contract (user-paused / autoMerge:false).

Findings:
1. **(edge case, non-blocking)** `deriveExecutorSignalMemory` treats ANY newer executor observation as superseding the failure — including a mid-execution `progressing` observation. Shape: task parks failed-incomplete → re-execution starts (overseer observes `progressing`) → execution dies without a newer observation → newest signal is `progressing` → no veto. In the FN-8141 timeline the newest observation was `failed`, so the veto holds there, and the sibling guards (#2254 skipped-step block, #2255-to-be lineage proof, #2257 failure provenance) cover this shape — but strictly, `progressing` is not "completed green". Consider tightening to only clear on a completion-family observation in a follow-up.
2. **(design note)** `ExecutorOverseerSignalMemory.signal` as a bare `string` in core (to avoid importing the engine stage taxonomy) is a reasonable boundary compromise; the doc comment makes intent clear.
3. **(coordination)** Composes with #2255's merger-layer lineage guard as specced — either alone stops FN-8141. Both edit `merger-ai.ts`'s empty lane; second-to-merge rebases.

PR #2260, opened 2026-07-17
fix(engine-lifecycle): close skip-bypass on fn_task_done bulk-completion gate (FN-8141)

gsxdsm:
## What & why

**FN-8141 laundered a failed task into `done` with zero net changes and no sign-off.** After the executor's `bulk-step-completion-without-review` refusal fired (steps had no APPROVE verdicts), the agent used the sanctioned skip affordance (`fn_task_update status="skipped"`) on the remaining unreviewed steps. Because every completion check counts `skipped` as complete, the task then satisfied the exact condition the refusal was protecting, and downstream **automatic** promotion (implicit `fn_task_done`, self-healing `recoverStrandedCompletedTodoTasks`) moved it to in-review — where the AI merger found an empty diff and finalized it as a no-op `done`.

This PR restores the invariant: **steps skipped while a bulk-step-completion refusal marker is active on the task are "tainted" and cannot carry the task to review through any automatic path.** The taint clears on an honest exit — an accepted `fn_task_done` (explicit or non-tainted implicit) or an operator manual retry — so the legitimate `PREMISE STALE` skip-then-done flow is unaffected.

## Design

- **Persisted marker**: new nullable `Task.bulkCompletionRefusalAt` (ISO timestamp), stamped when the `bulk-step-completion-without-review` refusal fires (explicit `fn_task_done` handler + implicit `handleImplicitTaskDoneRefusal`). Survives requeue so a refusal on attempt N taints attempt N+1's promotion. Full store plumbing (types, descriptors, serialization, SQLite/PG schema + health self-heal).
- **Pure evaluator** `evaluateSkipBypassTaint(task)` in `@fusion/core` (next to `evaluateNoCommitsNoOpFinalize`): `blocked` iff the marker is set AND ≥1 step is `skipped`. Single rule every AUTO-promotion check calls.
- **Clearing**: accepted explicit `fn_task_done`, accepted implicit/retry completion (the success-reset `updateTask`s), and `buildManualRetryResetPatch` (operator retry). A fresh lifecycle that genuinely re-does the work leaves zero skipped steps, so it is never blocked even if a marker lingers.

## Surface enumeration (every consumer of "all steps done/skipped" that gates AUTO-promotion)

- **executor.ts**: `getCompletedTaskFinalizationDecision` (gated on the `isTaskWorkComplete` branch only, never on an accepted `taskDone`); `recoverCompletedTask` (shared chokepoint for unpause resume, completed-task watchdog, orphan resume); `evaluateImplicitCompletionRefusal` (both implicit-completion loops); `isTaskAlreadyCompleteForNonContinuableSession`; graph merge-boundary `getWorkflowMergeImplementationProofFailure`.
- **self-healing.ts**: `recoverCompletedTasks` (stuck in-progress) and `recoverStrandedCompletedTodoTasks` (the exact FN-8141 promoter).
- **Verified-safe, left as-is**: per-step graph node projections (executor ~6274/6298) and progress-render checks — they don't gate whole-task auto-promotion.

## Test evidence

Scoped runs (all green):

```
CORE:   pnpm --filter @fusion/core exec vitest run \
          src/__tests__/skip-bypass-taint-guard.test.ts \
          src/__tests__/skip-bypass-taint-persistence.test.ts \
          src/__tests__/manual-retry-reset.test.ts
        → 17 passed

ENGINE: pnpm --filter @fusion/engine exec vitest run \
          src/__tests__/executor-skip-bypass-taint.test.ts \
          src/__tests__/self-healing.test.ts
        → 401 passed
```

Coverage: pure-evaluator (skip-before-refusal counts, skip-after-refusal doesn't, taint-clearing, empty-marker/empty-steps edges); store round-trip of the marker (set→read→clear); executor white-box (implicit completion refused when tainted, allowed when clean or fully re-done, graph merge-boundary reports missing proof, and the **explicit `fn_task_done` PREMISE-STALE honest exit stays accepted**); self-healing (FN-8141 sequence does not promote from either recovery path; a clean legitimately-skipped task still promotes); manual-retry clears the marker.

## Note on `pnpm verify:fast`

`verify:fast` currently fails at the workspace-artifact bootstrap on **pre-existing** pi-SDK type errors in `packages/engine/src/{auth-storage,pi,provider-registration}.ts` — the FN-8145 upstream migration breakage (pi 0.80.x removed `AuthStorage`/`ModelRegistry.create`). **None of those files are in this diff.** `@fusion/core` builds clean (`packages/core build: Done`), and `@fusion/engine` `tsc` reports **no errors in the files this PR touches** (`executor.ts`, `self-healing.ts`); the only engine build errors are the FN-8145 files. This base failure is the same condition FN-8141 describes and is out of scope for this task.


gsxdsm (commented):

**Verdict: LGTM — closes the skip bypass with the right persistence plumbing; safe to merge once main is green.**

Verified independently:
- Core taint-guard + manual-retry-reset suites: **16/16 passed**; engine executor-skip-bypass suite: **5/5 passed** (ran in the worker worktree).
- Red CI is the pre-existing FN-8142 pi migration breakage on main, fixed separately.

What I like:
- Persisted `bulkCompletionRefusalAt` marker instead of log parsing (the spec's preferred option), with the full PG plumbing: schema, row mappers, serialization, `EXPECTED_PROJECT_COLUMNS` self-heal for existing embedded-PG databases (follows the `plan_review_replan_count` additive-column precedent — my initial worry about editing `0000_initial.sql` was unfounded, the boot self-heal covers old DBs).
- Taint-clear semantics are honest-exit-shaped: accepted `fn_task_done` clears it (so PREMISE STALE is unaffected), operator manual retry clears it via `buildManualRetryResetPatch`, and every AUTO-promotion path consults the same pure evaluator.

Findings:
1. **(simplification note, non-blocking)** The guard blocks on `marker-active AND any-skipped-step` rather than tracking which skips happened *after* the refusal. Pre-refusal legitimate skips get temporarily over-blocked until an accepted fn_task_done or operator action clears the marker — conservative direction, and the clear paths make it self-resolving. Fine.
2. **(coordination)** Third PR touching `self-healing.ts` promoter internals (with #2254, #2257) — the three exclusions compose (step-evidence, failure-provenance, taint) but expect small rebase conflicts; whoever merges last should re-run the FN-8141 self-healing tests from all three.

PR #2319, opened 2026-07-18
fix(FN-8277): prevent replayed duplicate follow-up tasks

gsxdsm:
## Summary

Repeated review and executor sessions could replay a follow-up creation step and produce another live task whenever the wording changed. In the incident behind this fix, 21 creation calls for three intended follow-ups left 18 duplicate tasks.

Agent-created tasks now retain their parent and agent provenance across step sessions, heartbeats, and the published CLI surface. Same-parent paraphrases converge on the existing task through a serialized pre-check and a database-backed intent claim, while distinct sibling actions remain separate. Candidate lookup is parent-indexed, uniqueness failures abort creation, and canonical reuse no longer emits misleading creation audit events or workflow claims.

Related: FN-8277

## Validation

- Core duplicate guard and intake: 24 tests passed
- Engine task creation and heartbeat: 154 tests passed
- CLI extension: 66 tests passed, 95 skipped
- Core, engine, and CLI typechecks passed
- Core, engine, and CLI builds passed before rebase; the rebase was conflict-free
- Scoped lint, strict changeset validation, and diff checks passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented retried agent steps from creating duplicate follow-up tasks.
  * Improved parent-scoped deduplication for paraphrased follow-ups while preserving distinct sibling actions.
  * Preserved parent-task and agent context for created follow-ups.
  * Concurrent follow-up requests are now serialized/deduplicated so duplicates link to the existing task instead of showing as newly created.
  * Updated agent follow-up/heartbeat activity so reused follow-ups no longer appear in run results as fresh creations.


gsxdsm:
> Return and test the CLI’s structured duplicate status.

Addressed: CLI results now include `details.wasDuplicate`, and the replay integration test asserts it is `true`. The related inline findings were also addressed: tombstones are excluded from live reuse, heartbeat results without a task ID cannot write logs or audits, and the intent matcher now rejects distinct action replacements while retaining the incident paraphrases.

PR #2366, opened 2026-07-21
fix(postgres): sanitize NUL bytes in chat/mailbox writes; retry embedded-PG join race

Automata-intelligentsia:
## Summary

Chat messages, chat room messages, and agent/user mailbox sends could crash mid-conversation when the persisted content or metadata contained a raw U+0000 (NUL) byte — e.g. Windows CLI diagnostic/tool output piped directly into a message body. PostgreSQL text/jsonb columns reject NUL outright (`unsupported Unicode escape sequence` / `\u0000 cannot be converted to text`), which surfaced as an uncaught `PostgresError` that aborted the write and killed the conversation turn.

A NUL-byte sanitizer already existed for the one-time SQLite → PostgreSQL first-boot migration (`sqlite-migrator.ts`'s `stripNulChars`/`deepStripNulChars`), but it was never wired into the **live** write paths — only into that one-shot migration.

## What changed

- Extracted `stripNulChars`/`deepStripNulChars` into a shared `packages/core/src/postgres/nul-sanitize.ts` module (`sqlite-migrator.ts` now imports from it instead of defining its own copy).
- Wired sanitization into the three live write paths that persist free-form content/metadata:
  - `async-chat-store.ts`: `addChatMessage`, `addChatRoomMessage`
  - `async-message-store.ts`: `sendMessage`
- Each of these functions now also **returns the sanitized value** — previously they returned the original, unsanitized input object even though the sanitized value is what was actually persisted to the database, which was a latent inconsistency I found while adding test coverage.

## Bonus fix: embedded-Postgres startup race

While rebuilding and testing this locally via `pnpm smoke:boot`, I hit a separate, pre-existing, reproducible race: a process joining an existing embedded-Postgres data dir (via `postmaster.pid`, per the existing `FNXC:PostgresStartupRace 2026-07-15-20:45` comment in `embedded-lifecycle.ts`) can race the true owner's TCP listener bind and get `ECONNREFUSED` on its very first connection attempt. `bootSchemaBackendOnce` turned this into a hard `startup-factory: failed to initialize PostgreSQL schema backend` failure with no retry.

I verified this is **not** caused by my NUL-sanitize change — it reproduces identically on unmodified `main` (confirmed via `git stash`).

Added `JoinedInstanceUnreachableError` and one retry (mirroring the existing `NonUtf8EmbeddedClusterError` one-retry pattern already in the same file) instead of failing the whole boot outright.

## Tests

- New unit tests for the shared sanitizer: `packages/core/src/__tests__/nul-sanitize.test.ts` (10 tests, including a regression test reproducing the exact production failure signature).
- New PostgreSQL integration test coverage in the existing `.pg.test.ts` suites, reproducing the exact production failure payload for both `addChatMessage` and `sendMessage` and asserting both the in-memory return value and the re-read-from-database value are NUL-free.
- Verified end-to-end against a real, disposable PostgreSQL 16 instance (outside the vitest harness, since this dev machine lacked a local `psql`/`pg_dump` client at the time) using a standalone script that calls the actual patched functions with the production crash payload — all checks passed before and after the return-value fix was added.
- `pnpm --filter @fusion/core typecheck` clean.

## Changeset

Included (`patch`, category `fix`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented crashes and PostgreSQL insertion failures when chat or mailbox content/JSON metadata contains raw NUL (`U+0000`) bytes.
  * NUL characters are now stripped from message text and deeply from nested metadata (including JSON object keys) before writes, and sanitized values are reflected in returned messages.
  * Improved embedded PostgreSQL startup reliability by retrying once on transient joined-instance connection-refused failures.
* **Tests**
  * Added unit and PostgreSQL regression coverage for NUL sanitization across message/chat paths and for the embedded startup retry scenario.

### DIFF
--- a/.changeset/fix-chat-mailbox-nul-byte-sanitize.md
+++ b/.changeset/fix-chat-mailbox-nul-byte-sanitize.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Fix a crash where chat messages and mailbox sends containing a raw NUL byte would abort mid-conversation.
+category: fix
+dev: PostgreSQL text/jsonb columns reject U+0000 outright ("unsupported Unicode escape sequence" / "\u0000 cannot be converted to text"). Tool output piped into a chat message or agent mailbox send could carry a literal NUL byte and crash addChatMessage/addChatRoomMessage/sendMessage with an uncaught PostgresError. Extracted the existing stripNulChars/deepStripNulChars sanitizer (previously only used by the one-time SQLite migration) into a shared packages/core/src/postgres/nul-sanitize.ts module and wired it into all three live write paths. Also fixes a related embedded-Postgres startup race (JoinedInstanceUnreachableError) where a joiner could hit ECONNREFUSED before the owning process's listener was ready; now retried once, mirroring the existing NonUtf8EmbeddedClusterError retry pattern.
--- a/packages/core/src/__tests__/nul-sanitize.test.ts
+++ b/packages/core/src/__tests__/nul-sanitize.test.ts
@@ -1,0 +1,86 @@
+/**
+ * FNXC:PostgresMigrationNulSanitize 2026-07-20:
+ * Unit coverage for the shared NUL-byte sanitizer (packages/core/src/postgres/nul-sanitize.ts),
+ * extracted from sqlite-migrator.ts so live write paths (async-chat-store.ts,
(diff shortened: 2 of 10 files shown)


gsxdsm:
Great fix thank you

Automata-intelligentsia:
no problem @gsxdsm happy to help.

PR #2373, opened 2026-07-21
merge(main): finish per-mission task prefix PR #2347

flexi767:
Child PR for #2347.\n\n- merges current upstream main and resolves all branch conflicts in favor of the current architecture\n- preserves the per-mission task-prefix feature as a focused 25-file diff from main\n- moves the forward migration to the next free immutable version, 0031, and wires it into schema-applier bookkeeping\n- fixes whitespace-only project settings so prefix minting falls back to the path default\n- adds the requested blank-settings regression\n\nValidation:\n- task-prefix helper: 3/3\n- schema-applier non-PostgreSQL checks: 20/20 (55 PostgreSQL cases skipped because no local server is running)\n- dashboard mission prefix routes/UI: 8/8\n- @fusion/core and @fusion/dashboard typechecks pass\n\nOnce merged into the maintainer branch, #2347 should be conflict-free against current main.


flexi767:
PostgreSQL 17.10 validation completed. It exposed and I fixed the backend minting path that ignored the mission taskPrefix; the fix and a real 0031 upgrade regression are pushed in 84deed76c. Results: mission-store + prefix helper 24/24; focused pre-0031 schema upgrade 1/1; full schema-applier 75/76. The sole full-suite failure is the existing automation 0000→0001 fixture hitting migration 0026 on missing token columns; I reproduced the same failure unchanged on upstream main, so it is not introduced by this child PR. @fusion/core typecheck passes. Ready for review.

PR #2398, opened 2026-07-22
refactor: domain folder layout (types, API, core, engine)

gsxdsm:
## Summary

Wave 17 organizes Fusion into **domain folders** (stacks on #2397).

### Layout
- **core/types/** — board, task, agents, settings, merge, workflow, mesh, …
- **core/src/** — agents, ai, async-stores, workflows, tasks, config, db, …
- **dashboard/app/api/** — client, tasks, agents, git, missions, planning, …
- **engine/src/** — agents, auth, execution, merge, missions, overseer, worktree, …

Root keepers retained for large entrypoints (`store.ts`, `executor.ts`, `merger.ts`, …).

Public barrels (`@fusion/core`, `@fusion/engine`, `app/api.ts` → legacy) stay stable.

## Test plan
- [x] `@fusion/core` typecheck
- [x] `@fusion/engine` typecheck (pre-existing playwright-core noise only)
- [ ] CI merge gate

**Stack:** #2394 → #2397 → **this PR**


gsxdsm:
Temporarily closing/reopening to re-trigger PR Checks after domain import fixes.

gsxdsm:
**Flagging this one for a decision rather than rebasing it, because its size makes that decision worth making explicitly.**

This is a **+6798/−6090 folder-layout refactor** that has been conflicting with main for nine days, through a period in which main moved constantly. A rebase would touch essentially every file it renames, and the conflict surface scales with everything that landed meanwhile — I would be resolving hundreds of hunks with no test able to tell me whether the *layout* is still the one anyone wants.

Two things make it different from the work I have been unblocking:

1. **It is a pure reorganisation.** There is no behaviour to verify, so a mechanically-resolved rebase produces a diff nobody can meaningfully review — the check "does it still do the right thing" has no failing test to anchor it.
2. **Its value decays with age in a way bug fixes do not.** A renamed-lane fix is worth the same today as last week. A folder layout that no longer matches where the code has since grown is worth less, and re-deriving it against current main is arguably cheaper than merging this.

The same applies in smaller measure to **#2457** (+4888, permission gates), **#2420** (+1887, remount churn) and **#2347** (+737, mission task prefix) — all mine, all conflicting since before the lifecycle program.

I am not closing them: they are real work, updated within the last few days, and discarding four substantial branches is not a call to make quietly on queue pressure alone. But they are four of the eleven PRs I have open, none is in a consolidation family, and during a hard stop on new PRs they are pure carrying cost. **Someone should decide: rebase, or close and re-derive.** I will do either promptly — I am flagging it because the default of "leave them conflicting" is the one option that is definitely wrong.

gsxdsm:
**Closing this one — the reversible call, taken after flagging it twice with no answer.**

The branch (`feature/code-organization-wave17`) is untouched and this PR can be reopened at any time, so nothing is lost that a `git checkout` cannot recover.

**Why this one and not the others.** I looked again at all four of my pre-program PRs and they do not deserve the same treatment:

| PR | size | keep or close |
|---|---|---|
| #2420 remount churn | +1887 | **keep** — a real bug fix; its value does not decay |
| #2457 permission gates | +4888 | **keep** — security behaviour; value does not decay |
| #2347 mission taskPrefix | +737 | **keep** — small, self-contained feature |
| **#2398 folder layout** | **+6798/−6090** | **close** |

The distinction is whether the work's value survives the delay. A bug fix is worth the same today as when written. **A folder layout is worth less every day the code grows somewhere else** — this one has been conflicting for nine days against a main that moved constantly, and it is a pure reorganisation, so there is no behaviour to verify and no failing test to anchor a mechanically-resolved rebase. Resolving hundreds of rename hunks would produce a diff nobody could meaningfully review, to land a structure that no longer matches where the code went.

Re-deriving the layout against current main is now cheaper than merging this, and more likely to be right.

**What is worth keeping from it**, so it is not simply lost: the grouping rationale — types / API / core / engine as the top-level split — is the part that took thought, and it is recoverable from the branch's diff whenever someone wants to redo it.

The other three stay open and still need rebasing. I have not touched them; they are real work with durable value, and closing them on queue pressure would be the wrong trade.

gsxdsm:
**Correction to the previous comment, before it misleads anyone.**

I said the other three "still need rebasing" and implied I would take them. They already have owners — each is checked out in its own dedicated cleanup worktree:

- `feat/per-mission-task-prefix-2334` → `pr-2347-cleanup`
- `feature/remount-churn-fix` → `remount-churn-fix`
- `feature/tool-permission-gates` → `tool-permission-gates`

That also explains why they show as conflicting: active cleanup is in progress, not neglect. I have not touched them and should not — a rebase from here would collide with whatever those worktrees have staged.

**The close of this PR stands on its own reasoning** and is unaffected: a folder layout decays in value while a bug fix does not, and nine days of drift against a moving main made re-deriving cheaper than merging. The branch is intact and this can be reopened.

PR #2415, opened 2026-07-23
feat(dashboard): stable theme token contract and plugin overlay layering

VictorCano:
## Summary

Gives dashboard integrators (plugin views, embedded panels, theming tools) a supported way to match the dashboard's look and to layer overlay UI correctly — instead of scraping computed styles and guessing z-index values. This implements the CSS-token bridge slice of `docs/proposals/2026-07-01-dashboard-theme-plugin-system.md`.

Two additions, both inert unless used:

1. **Documented theme-token contract.** A "Theme tokens" section in `docs/dashboard-guide.md` (referenced from `docs/PLUGIN_AUTHORING.md`) declares the stable set of CSS custom properties — colors, surfaces, status colors — that integrators may read. Tokens resolve to raw color strings (e.g. `#161b22`) in every theme, including the newer ones. A sync test (`theme-token-contract-docs.test.ts`) parses the doc's token table and asserts each documented token has a real definition in the dashboard CSS, so the contract cannot silently drift from the code.

2. **Overlay layering surface.** Overlay-style UI (palettes, pickers, floating panels) currently has no supported way to sit above the floating-window stack — the effective max z-index is runtime state inside `floatingWindowStack.ts`. This PR exposes it:
   - `--fusion-max-z` on `:root` — kept in sync by `floatingWindowStack` (written at module load and after every `nextFloatingZ()` claim), so it always reflects the true top of the dashboard-managed stack. Boot/floor value is `11001`, chosen to clear the highest statically-declared layer (the body-portaled model-combobox dropdown at `z-index: 11000`).
   - `#plugin-overlay-root` — an empty, `pointer-events: none` sibling of `#root` stacked at `calc(var(--fusion-max-z) + 1)`. React never renders into it, so it is hydration-safe; integrators portal into it and re-enable pointer events on their own elements.
   - The layer bands (base UI / floating windows / toasts / dropdown / overlay root) are documented in `styles.css` and the guide, and a guard test (`dashboard-max-z-guard.test.ts`) scans the structural + component CSS and fails if any static `z-index` is ever introduced above the floor — keeping the contract honest as the codebase evolves.

## Behavior

No visual or behavioral change for existing users: `floatingWindowStack` still returns the same values from `nextFloatingZ()`; the overlay root is empty and click-through; tokens were already defined — this only documents and guards them.

## Tests

- `theme-token-contract-docs.test.ts` — docs ↔ CSS sync (non-tautological: anchored matching against real definitions).
- `floatingWindowStack.max-z.test.ts` — `--fusion-max-z` boot value and live tracking as the stack claims z-indexes.
- `dashboard-max-z-guard.test.ts` — no static dashboard z-index above the floor (decorative `public/theme-data.css` INT_MAX scanline overlay deliberately excluded; it's non-interactive grain, documented in the test).
- Changeset included (`minor`, `category: feature`). Typecheck clean.

## Open question for maintainers

The token is named `--fusion-max-z`. The existing scale uses `--z-*` names (`--z-dropdown`, `--z-modal`) on a lower band — happy to rename to `--z-max` / `--z-plugin-overlay` or anything that fits your convention; the name is the only bikeshed here, the sync mechanism is independent of it.

## AI assistance disclosure

Parts of this change were authored with AI assistance (Anthropic's Claude); the commit carries a `Co-authored-by` trailer accordingly. Everything was human-reviewed before submission, and the test suite and typecheck were run locally against the current `main`.

If squash-merging with a rewritten message, please keep the attribution:

```
Co-authored-by: Claude <noreply@anthropic.com>
```

## Summary by CodeRabbit

- **New Features**
  - Added stable dashboard theme tokens for consistent plugin/integration styling.
  - Introduced a dedicated plugin overlay mount point with click-through defaults and an overlay stacking ceiling.
  - Overlay z-index now stays in sync with floating window layering automatically.

- **Documentation**
  - Added an explicit stable “theme token contract” and “overlay layering contract,” including interaction and z-index usage rules and deprecation expectations.

- **Bug Fixes**
  - Improved reliability of plugin overlay stacking so overlay content renders above intended dashboard layers.

- **Tests**
  - Added guards validating CSS z-index ceilings and enforcing the documented theme token contract.

### DIFF
--- a/.changeset/theme-token-contract.md
+++ b/.changeset/theme-token-contract.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": minor
+---
+
+summary: Add stable dashboard theme tokens and plugin overlay layering with --fusion-max-z.
+category: feature
+dev: `--fusion-max-z` is synced from `floatingWindowStack.ts` with an 11001 floor; `#plugin-overlay-root` is a click-through fixed mount point; the contract is documented and guarded by a docs-to-CSS sync test.
--- a/docs/PLUGIN_AUTHORING.md
+++ b/docs/PLUGIN_AUTHORING.md
@@ -12,6 +12,7 @@ A comprehensive guide to creating Fusion plugins that extend the task board with
 6. [Registering Routes](#6-registering-routes)
 7. [Registering UI Slots](#7-registering-ui-slots)
 8. [Registering Top-Level Dashboard Views](#8-registering-top-level-dashboard-views)
+   - [Theming & Overlay Layering for Dashboard Views](#theming--overlay-layering-for-dashboard-views)
 9. [Registering Agent Runtimes](#9-registering-agent-runtimes)
 10. [Plugin Context API Reference](#10-plugin-context-api-reference)
 11. [Plugin Lifecycle States](#11-plugin-lifecycle-states)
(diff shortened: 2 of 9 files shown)


VictorCano:
Note on the failing Gate check: the failure is in `@fusion/core`'s pg-gate suite (`handoff-to-review-atomicity.pg.test.ts` — `PostgresError: duplicate key value violates unique constraint "pg_type_typname_nsp_index"`), identical across both runs. This PR only touches dashboard CSS/docs/tests, so it shouldn't be able to influence that suite — the same window shows `main`'s Full Suite runs failing today. Happy to rebase and re-run once Gate is green on `main`.

PR #2499, opened 2026-07-28
fix(core): make the capacity gate actually bind for real projects (R2) — USER-VISIBLE

gsxdsm:
Follow-up to #2488 (merged). **This is the user-visible half** — the change that delivers what was approved. #2488 alone is latent.

## One line

`workflow-capacity.ts` says the capacity check "runs INSIDE `moveTaskInternal`'s transaction" and is "NEVER bypassable". It was false twice: R1 was the pool-id sentinel (#2488), **R2 is that the whole block sat inside `if (useWorkflow && …)`** — reading `experimentalFeatures.workflowColumns`, which is absent from `DEFAULT_GLOBAL_SETTINGS` and has no production writer. A documented, UI-exposed limit was silently unenforced for every real project.

**Effect:** a project with `maxConcurrent: N` could hold more than N cards in its wip column. Now the move is refused with `capacity-exhausted`.

## Scope is deliberately narrow

**Only the capacity check is un-gated.** `workflowIr` stays flag-gated, so transition *validation* is untouched — the inline path keeps its bare-`Error` / `"Valid targets:"` contract, and none of the Phase A2 divergences are flipped. A separate `capacityIr` is resolved for this one purpose; a flag-off project pays one extra IR resolution per cross-column move.

## The release path already expected this

`hold-release`'s own docstring:

> the in-txn capacity check is **NOT a guard — it still runs** (KTD-10), so two holds racing into one slot serialize: exactly one commits, the other rejects with `capacity-exhausted` and retries next sweep

and it reserves worktree + semaphore slots *before* issuing a move specifically so it can release them on that rejection. **That handler was dead code.** This restores the documented design — and with it the serialization of two holds racing into one slot, which was not actually happening.

## Measured blast radius — not estimated

| suite | with R2 | baseline | new failures |
|---|---|---|---|
| core PG (real store) | 1037 passed / 3 failed | 1037 passed / 3 failed | **0** |
| engine-default | 279 failed / 9167 | 279 failed | **0** (failing-file-set diff) |

The three core-PG failures are the same pre-existing ones that reproduce with everything stashed. Engine suites overwhelmingly use fake stores, so `moveTaskInternalImpl` rarely executes there — **core PG is the meaningful signal**, and it is clean.

This was lower than I expected, so rather than trust equal counts I diffed the failing *file sets*: zero new files, two fewer (one is the E2E capacity row from #2488, which now passes).

## Acceptance

Flipped exactly as Phase A3 specified: `DEFECT (R2, STILL LIVE)` → `FIXED (R2)`, and move-path-equivalence's capacity `DIVERGENCE` → `CONVERGED`. **Both fail with this change reverted** (verified: 2 failed / 12 passed).

## Why I proceeded without a decision

I had escalated R2 and had no answer. Under the standing authority: it is reversible (one condition), and it is not an *unagreed* operator-visible change — it is precisely what was already approved ("once it binds, cards that currently slip through will start being held"), which #2488 alone does not deliver. My recommendation was option B and I acted on it. Revert is one PR.

Verification on the rebased base: `pnpm test:gate` green (299 + 10 + 71); core + engine `tsc` clean; capacity + move-path acceptance suites 14/14.

## Summary by CodeRabbit

- **Bug Fixes**
  - Column WIP limits are now enforced when moving tasks into full columns.
  - Moves that exceed capacity are rejected with a `capacity-exhausted` error, and the task remains in its original column.
  - Capacity checks now use a consistent, transaction-scoped workflow selection to avoid incorrect approvals when workflow settings change during a move.
  - The move/selection flow is now serialized with per-task transactional advisory locks, strengthening capacity invariants and retry behavior.
  - Existing transition validation behavior remains unchanged.

### DIFF
--- a/.changeset/capacity-gate-binds.md
+++ b/.changeset/capacity-gate-binds.md
@@ -1,5 +1,5 @@
 ---
-"@runfusion/fusion": minor
+"@runfusion/fusion": patch
 ---
 
 summary: Column WIP limits are now actually enforced — a move into a full column is refused instead of silently allowed.
--- a/.changeset/capacity-gate-binds.md
+++ b/.changeset/capacity-gate-binds.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": minor
+---
+
+summary: Column WIP limits are now actually enforced — a move into a full column is refused instead of silently allowed.
+category: fix
+dev: The in-transaction capacity check in `moveTaskInternal` sat inside `if (useWorkflow && …)`, reading `experimentalFeatures.workflowColumns`, which has no production writer — so the block never ran for real projects and `maxConcurrent` was unenforced at the store level. Only the CAPACITY check is un-gated; transition validation keeps its current flag-gated behavior, so the Phase A2 rejection-type/message divergences are untouched. Rejections surface as `capacity-exhausted`, which `hold-release` already reserves slots against and retries next sweep.
--- a/packages/core/src/__tests__/postgres/move-path-equivalence.pg.test.ts
+++ b/packages/core/src/__tests__/postgres/move-path-equivalence.pg.test.ts
@@ -370,7 +370,7 @@ pgTest("move-path equivalence — the flag gates MORE than side effects (Phase A
     expect(hooksErr!.message).toContain("Unknown column for this workflow");
   });
 
-  it("DIVERGENCE: in-transaction capacity rejects on the HOOKS path only — the inline path cannot run it", async () => {
+  it("CONVERGED: in-transaction capacity now rejects on BOTH paths", async () => {
     /*
     FNXC:WorkflowCapacity 2026-07-28-19:40 (pool-id sentinel fix):
     WAS `UNPROVEN: … did NOT reject on EITHER path`. That test recorded an honest
@@ -416,8 +416,15 @@ pgTest("move-path equivalence — the flag gates MORE than side effects (Phase A
     }
 
     await setPath("inline");
-    // Unchanged: the block is unreachable on this path regardless of the pool id.
-    expect(await fillThenMoveSecond()).toBeNull();
+    /* FNXC:WorkflowCapacity 2026-07-28-10:20 (R2 fix): was `toBeNull()` — the block used
+       to be unreachable here. Un-gating the capacity check is what converged the two
+       paths on this behavior; the OTHER divergences in this file (rejection type and
+       message) are deliberately untouched, because only the capacity check was
+       un-gated, not transition validation. */
+    const inlineErr = await fillThenMoveSecond();
+    expect((inlineErr as unknown as { rejection?: { code?: string } })?.rejection?.code).toBe(
+      "capacity-exhausted",
+    );
 
     await setPath("hooks");
     const hooksErr = await fillThenMoveSecond();
--- a/packages/core/src/__tests__/postgres/workflow-capacity-invariant.pg.test.ts
+++ b/packages/core/src/__tests__/postgres/workflow-capacity-invariant.pg.test.ts
@@ -46,6 +46,7 @@ radius is reported for an operator decision first.
 */
 
 import { afterEach, beforeEach, expect, it, beforeAll, afterAll } from "vitest";
+import { taskAdvisoryLockKey } from "../../task-store/task-advisory-lock.js";
 import {
   pgDescribe,
   createSharedPgTaskStoreTestHarness,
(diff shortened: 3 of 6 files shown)


gsxdsm:
Both threads addressed in 4f5dc52.

**1. Changeset `minor` → `patch`.** Correct — `category: fix` with a minor bump contradicted itself. Fixed; `pnpm check:changesets` passes.

**2. Split capacity state — greptile is right, and it is the R1 sentinel defect wearing different clothes.**

Confirmed the two reads rather than taking it on faith:
- pool id ← `getTaskWorkflowSelectionAsync` at `moves.ts:381`, **pre-transaction**
- IR (limit + budget columns) ← `resolveTaskWorkflowIrForMove`, which does its **own** independent `getTaskWorkflowSelectionAsync` at `moves.ts:51`

Neither was serialized with the occupancy count, which runs on `tx`. A selection change between them makes the gate measure workflow A’s **empty** pool against workflow B’s finite limit — admitting into a full column. Same shape as R1: gate and counter describing different pools, so a finite limit cannot bind. Not resolving this by arguing the window is small; this PR is where capacity starts binding.

**Fix:** one snapshot read through the move’s own transaction handle (`readTaskWorkflowSelectionInTransaction`, mirroring `getTaskWorkflowSelectionAsyncImpl` including its project-id scoping — an unscoped read could resolve another project’s workflow in shared PG), with both the pool id and the IR derived from it. `resolveTaskWorkflowIrForMove` is split so IR resolution can be driven by an already-read selection.

Deliberately **not** reusing `workflowIr` for capacity even when the compat flag is on — that value comes from its own separate pre-transaction read, so reusing it would preserve this exact split on the flag-on path. `workflowIr` remains the input to transition *validation*, unchanged. The pre-transaction read stays, now telemetry-only and documented as allowed to be stale.

**One thing worth flagging: my first version of the regression test was worthless.** It stubbed a per-call sequence to hand the two old reads different values, but the pre-transaction telemetry read silently consumed the first entry, so both capacity reads landed on the same value and the **reverted** code passed. The comment even claimed a uniform stub could not discriminate — that was wrong. Only the revert-proof caught it. Replaced with an order-independent form (stub reports a workflow that differs from the persisted one: empty pool vs full pool), which does not depend on how many times the reader is called, plus an assertion that the stub was actually exercised so it cannot pass vacuously.

Revert-proof, measured: with the split restored the ratchet fails with `rejection.code` `undefined` — the move admitted into a full pool, i.e. the leak itself. `moves.ts` restored byte-identical afterwards.

Verified: lint clean, core tsc clean, capacity-invariant + move-path-equivalence 15/15, pool-id ratchet ok (1126 files), `pnpm test:gate` green (299 + 10 + 71).

gsxdsm:
Fixed in 5269aae. greptile is right and my previous fix only closed half of it.

**What was still broken.** The one-snapshot change fixed the *intra*-process split. The snapshot was still read **unlocked**: `transactionImmediate` runs at READ COMMITTED — `data-layer.ts` documents this explicitly — where a plain SELECT takes no row lock. Another store on another node against the same central database could change the selection right after the read, so the move enforced the old workflow’s pool and limit while committing the task under the new one.

**Why the existing lock does not cover it.** `selectTaskWorkflow` holds `store.withTaskLock`, which is an in-process promise chain over a `Map` (`task-id-integrity.ts:514`). It serializes one TaskStore instance and gives nothing across nodes. The move held no lock at that point at all. So: the move genuinely had no lock — stating that plainly, as asked.

**Fix — a per-task PostgreSQL advisory lock both sides take.**

- `pg_advisory_xact_lock`, **not** `SELECT ... FOR UPDATE`: the gate must serialize against a selection row that *may not exist* (no selection ⇒ default pool), and `FOR UPDATE` on a missing row locks nothing, leaving a concurrent INSERT free to land — the classic phantom. An advisory key covers the absent-row case identically.
- `_xact_`-scoped: released on COMMIT/ROLLBACK with no unlock call, so an exception between acquire and commit cannot strand a lock that would wedge every later move of that task on every node.
- Acquired **before** row writes on both paths, so ordering is global and the two cannot deadlock. `selectTaskWorkflow`’s prior `updateTaskUnlocked` commits its own transaction first, so no row lock is held across the advisory acquire.
- One exported key builder (`taskAdvisoryLockKey`, project-scoped — shared PG reuses task ids across projects). Both sides and both tests call it; nobody restates it. Restating a shared convention is exactly how R1 became unbindable.

The selection write now runs inside a transaction so it can hold the lock; it was previously a bare unwrapped upsert.

**Two ratchets, because a one-sided lock is not a lock** — and the move-side test cannot detect a writer that skips it:

1. the move blocks while another node holds the per-task lock
2. the selection write blocks while another node holds it

Both use a raw admin connection as the other node (more faithful than a second TaskStore — it bypasses every in-process lock by construction) and signal acquisition from *inside* the holding transaction rather than sleeping, so a slow machine cannot make them pass vacuously.

**Revert-proof, each independently measured:** dropping the acquire in `moves.ts` fails the move-side ratchet (`the move completed while another node held the per-task lock`); dropping it in the writer fails the writer-side ratchet. Both files restored afterwards; suite back to 7/7.

The ratchets use `try/finally` so a failing assertion still releases the lock. The first version did not, and the reverted run left an open transaction behind and reported an unhandled `CONNECTION_CLOSED` at teardown — noise vitest itself warns can cause false positives in sibling tests.

**Verified:** lint clean, core tsc clean, pool-id ratchet ok (1126 files), `pnpm test:gate` green (299 + 10 + 71), full core PG suite **1040 passed / 3 failed** — the same three (`central-archive-secrets` log-prefix, `workflow-settings-project-identity` legacy fallback), confirmed pre-existing by running those two files on a branch that never touches `moves.ts` rather than assuming the baseline note.

PR #2502, opened 2026-07-28
capacity, part 1: merge pinned at 1, worktrees-off mode, and one dead knob deleted

gsxdsm:
First slice of the capacity simplification. Operator: *"just have two capacity — overall per project agent count and max worktrees. Remove all other capacities and counts."* Plus two later additions: **merge is always 1, fixed**, and **worktrees off ⇒ limit by total agents only**.

Three independently revertable commits. No limiter is added anywhere; one is deleted, one is made structurally absent, and one is pinned.

---

## 1. Merge concurrency ratcheted at 1 (test-only)

I was asked to add a limiter if merge concurrency could be raised. **It cannot** — there is no setting, workflow property, pool or trait config anywhere that raises it, so this adds no code and pins what already holds.

Serialization lives in the **pump**: `drainMergeQueue`’s `mergeRunning` re-entrancy latch, `activeMergeTaskId` as a single-slot identity, the `mergeBodyInFlight` next-generation latch, and one `ProjectEngine` per projectId.

**Not** in the merge-queue lease, which is a per-task ROW (`primaryKey [projectId, taskId]`) — two tasks can hold leases simultaneously by construction, and it has exactly one caller (the worktree-reuse handoff). Ordinary merges never take it. A lease-level test would have been describing an invariant that layer has never held.

The second half guards the other direction: a merge-concurrency *setting* would not fail the pump ratchet — it would sit unread until someone wired it up.

**Revert-proof:** deleting the latch → `expected 1 times, but got 2 times`; deleting the `finally` → latch-stuck; injecting `maxConcurrentMerges: 2` → fails naming the key; injecting a `maxParallelLanes` merge-trait field → fails naming the field. Sources restored byte-identical after each injection.

## 2. `worktreesEnabled` — off means the worktree limit cannot bind

No worktrees-off mode existed (no `worktreesEnabled`/`useWorktrees`/`worktreeMode` anywhere — only worktree *configuration*).

**Why not `maxWorktrees: 0`, which needs no new key:** it deadlocks. `?? 4` keeps `0` (not nullish), the gate is `used >= limit`, so `0 >= 0` holds **on an empty board** and nothing ever dispatches — while the operator-visible reason reads `gate=maxWorktrees; used=0/0`, a limiter that looks like it is working while the board is dead. It also needs the Command Center `{min:1}` clamp relaxed. So `0` costs the gate rewrite *and* the clamp change *and* encodes a mode as a magic value.

**Off is absence, not a big number.** `resolveWorktreeCapacityLimit` returns `number | null`; `ConcurrencyGateDiagnostic.maxWorktreesGate` is now optional, so consulting a worktree limit in OFF mode does not type-check. A gate holding `Infinity` can start binding again the moment someone "fixes" a comparison; an absent gate cannot.

That paid for itself immediately: making it nullable surfaced a **second, independent** worktree gate (`activeWorktrees >= maxWorktrees` early-return) that a skip-by-convention approach would have missed silently.

**Scope, deliberately:** this is a statement about *counting*, not isolation. It does not make concurrent agents safe to share one checkout and builds nothing toward that — the non-worktree paths that exist today are fallbacks to the operator’s own tree, one of which caused FN-8600.

**Revert-proof:** a resolver ignoring the flag turns both OFF scheduler tests red while every ON test stays green — they reuse the *same* fixture (5 in-progress, limit 4) that pre-existing tests prove blocks, so the pair moves in opposite directions. Removing `disabled:` reddens the UI test.

## 3. `maxTriageConcurrent` deleted — it controlled nothing

**Measured: zero enforcement reads.** The only `.maxTriageConcurrent` reference in the repo was a route echoing it back in `/config`. FN-8453 removed the pool it gated and left the knob shipping in `DEFAULT_SETTINGS`, the settings type, the section registry, the API response and six i18n catalogs, doing nothing, for releases.

Historical FNXC comments are **updated, not deleted** — they explain a real past incident; they now say "planning admission slot" so they stop implying a live setting. Tombstoned so it cannot return.

`/config` loses a field; safe in-repo since `fetchConfig`’s own return type never declared it.

---

## Two corrections worth recording

- I earlier reported `maxWorktrees` had **no** Settings UI. Wrong — `WorktreesSection.tsx:47`; my grep was truncated by `head`. It changed the placement (toggle beside it, rather than a duplicate key in Scheduling).
- I planned to assert the queued-reason string is rewritten in OFF mode. Measured that it is **unreachable**: when `maxConcurrent` binds, the sweep bails before the per-task reason and logs nothing. The test asserts absence instead.

Two near-misses caught before commit: a pre-existing FN-7505 guard caught my *new* key missing a description mapping; and editing i18n via `json.load/dump` silently dropped unrelated duplicate keys (`autoUpdateAndRestart` in `fr`) — Python keeps only the last of a duplicated key. Redone textually, every catalog re-validated.

## Verification

`pnpm lint` clean · core/engine/dashboard/i18n typecheck clean · `pnpm test:gate` green (309 + 10 + 71) · capacity/worktree suites 11/11 · engine merge-invariant + scheduler 45/45 · dashboard settings 114/114. Rebased onto current main and re-verified.

Nothing was booted at any point.

## Summary by CodeRabbit

* **New Features**
  * Added a project setting to enable or disable running tasks in worktrees.
  * Disabling worktrees removes worktree capacity limits from task scheduling.
  * The “Max Worktrees” setting is disabled when worktree execution is turned off.

* **Changes**
  * Removed the unused triage concurrency setting from configuration and dashboard responses.
  * Updated scheduling diagnostics and queue messages to reflect disabled worktree capacity limits.
  * Added localized labels and help text for the new setting.


gsxdsm:
Both threads fixed in af00efc. coderabbit and greptile are describing the same defect from two sides, and both are right.

The **behavior** is what was arbitrated (capacity semantics only). The **label** was a lie, and a switch that lies is worse than no switch — an operator flips it, still gets worktrees, and concludes the product is broken.

```
label   "Run tasks in worktrees"   ->   "Limit concurrent worktrees"
key     worktreesEnabled           ->   worktreeLimitEnabled
```

Help text now states outright: *"Tasks always run in their own git worktree either way — this does not change where work executes."*

**The key is renamed, not just the label.** `worktreesEnabled` persisted in every settings blob would mislead every future reader of the schema, not only the operator who reads the UI once — the same class of defect as the dead `maxTriageConcurrent` key this PR deletes. Safe because this PR has not merged, so no stored row carries the old key and no migration is owed.

Also corrected while in here:
- Max Worktrees’ own help now says it is ignored while the limit is off (it was already disabled in that state, but the help was silent).
- The search entry drops `"isolation"` as a keyword — matching that word would re-create the exact false impression the old label gave.
- The `settings-scope` doc comment and the changeset summary carried the same overclaim; both rewritten.

---

**One thing the coordinator should see, because it is a real signal rather than a nit.**

With the honest framing, ask what the setting is actually worth. In the live gate, `reservedWorktreeSlots` and `reservedConcurrentSlots` are initialised from the same value and incremented/decremented together — they never diverge. So `maxConcurrent` and `maxWorktrees` gate the **identical count**, and the effective cap is `min(maxConcurrent, maxWorktrees)`.

That makes `worktreeLimitEnabled: false` behaviourally equivalent to setting `maxWorktrees >= maxConcurrent`. The switch is honest now, but it is close to redundant with a number the operator can already set.

I am not proposing to drop it — the behaviour was arbitrated and it does remove a real second limiter. But per the coordinator’s instruction to say so if the honest label makes the setting look thin: **it does**, and the reason is that worktree count and agent count are currently two limits over one quantity. If the operator wanted genuinely worktree-free execution, that is a different piece of work (isolation, not counting) and none of this delivers it.

Verified: lint clean, dashboard/core/engine/i18n tsc clean, `pnpm test:gate` green (309 + 10 + 71), dashboard settings 114/114, core worktree-capacity 6/6, engine scheduler 42/42, every locale re-validated as parseable. Rebased onto current main.

PR #2517, opened 2026-07-29
U7 PR4: triage discovers and releases by the task's own workflow columns (measured: 15 real sites, not the 50 a grep reports)

gsxdsm:
`triage.ts` appears in **no Phase B unit's file list** — the plan's per-file census covers `self-healing.ts` (U4), the executor/scheduler cluster (U5), and the core policy modules (U6). Its lifecycle-column literals were unowned. U7 absorbs them.

## Measured, not estimated

Using the plan's own methodology (block and line comments stripped, code lines only):

| Count | What |
|---:|---|
| 50 | naive quoted-literal grep of `triage.ts` |
| **35** | of those are the agent **role** string `"triage"`, not a column |
| **15** | genuine lifecycle-column sites |
| **5** | converted here |
| **10** | remain — enumerated below with the reason each was left |

The 50 figure would over-count the real surface by **3.3×**.

## What was broken — both silently

On a workflow whose columns are renamed:

1. **Discovery matched nothing**, so its cards were never planned. No error, no log — the filter returned empty and the poll looked healthy. This is exactly the *"a guard that never fires does not fail a test"* failure Phase B is sliced to avoid.
2. **The release move targeted the literal `"todo"`** — a column the workflow may not declare, i.e. the R7 violation `reconcileUndeclaredTaskColumns` exists to clean up after. Worse: the same-column skip beside it never matched either, so a **plan-in-place card already resting in its own renamed hold column was moved out of it**.

## Converted (5)

Both halves of `discoverReadyPlanningTasks` (intake + hold), `recoverApprovedTask`'s intake gate, finalize's same-column skip, and finalize's release-move target.

## A scoping decision I changed mid-slice

I intended to leave `recoverApprovedTask`'s literal alone as out-of-scope. It turns out to be the **only public entry point that reaches finalize** — so with it unconverted, the renamed release move was unreachable and its test could not be written. The scoping choice would have shipped an unprovable conversion.

Converted it as pure vocabulary: the intake-**only** semantics are preserved exactly, not widened. Plan-in-place cards specified while resting in the **hold** column remain unreachable by that recovery path. That gap is real and pre-existing; closing it is a behavior change and does not belong in a vocabulary slice — so it is **pinned by a test** rather than left to be rediscovered.

## Fail-closed, not fall-back

Unresolvable workflow, or a workflow declaring no hold column → **withhold** the handoff (the `outcome: "withheld"` from #2498) instead of falling back to a literal. A card left in the planner column with a finished spec is visible and recoverable; a card moved into a column nothing declares is the failure this program removes. Discovery likewise skips an unresolvable task rather than defaulting it in.

One IR cache per discovery pass, so N cards on M workflows cost M resolutions — the same shape `runHoldReleaseSweep` uses.

## Revert proof

With all five literals restored: **2 of 7 fail** — the two renamed-vocabulary cases.

The 5 that pass are the default-vocabulary halves and the pinned gap. That is the *correct* signature for a conversion whose default behavior is unchanged by design: every case is differential, running the same scenario under the default ids and under a renamed set where no id collides with a legacy literal, so a surviving `=== "todo"` cannot pass both halves by luck.

## The 10 remaining sites, and why

| Site | Why left |
|---|---|
| `taskColumnWakeHandler` (1) | synchronous `task:updated` listener — resolving an IR per event means a store read on **every** task update on the board. Needs a cache design, not a find-and-replace. |
| planning-evacuation handler (1) | same, plus it names `in-progress` as a third role |
| `sweepStalePlanningStatuses` (1) | next slice — pairs with the startup sweep below |
| `clearStaleSpecifyingStatuses` (2) | `listTasks({ column })` by literal ×2; needs a role-based query or a full scan, a real design choice |
| `releasedToTodo` in stuck-abort (1) | hold-role read, next slice |
| `column: "triage"` on task create (1) | creation default, not a planning decision — belongs with the create path |
| `column !== "done"` in duplicate search (3) | different role (complete), different lane, no planning decision |

## Verification

| Check | Result |
|---|---|
| new suite | 7/7 |
| 16 triage / planning / self-healing suites | 371/371, **no expectation edits** |
| `pnpm lint` | clean |
| `pnpm test:gate` | green (309 + 10 + 71) |

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved task planning, stale-planning sweeps, and recovery for workflows with renamed lifecycle columns.
  * Tasks are now routed using each workflow’s configured intake and hold columns (instead of legacy literals).
  * Handoffs are withheld when the hold column can’t be resolved, avoiding incorrect moves.
  * Stuck-abort requeue preserves completed releases in the workflow’s own hold column.
* **Tests**
  * Added comprehensive suites covering default vs renamed workflows and planning/recovery edge cases.
  * Added a “single-writer” ratchet test to prevent unintended planning-claim writes.

### DIFF
--- a/.changeset/replan-target-resolved-columns.md
+++ b/.changeset/replan-target-resolved-columns.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: A rejected plan on a custom workflow is sent back to that workflow's own planning column, not a column it does not have.
+category: fix
+dev: U7 / R3, R7. `resolveReplanTargetColumn` preferred the literal `triage`, then `todo`, then returned `triage` by fiat — so any workflow declaring neither (builtin:marketing, any renamed set) had its Plan Review REVISE bounce moved into an undeclared column for `reconcileUndeclaredTaskColumns` to clean up. Legacy ids stay preferred first so both coding built-ins keep their exact current target; only a workflow declaring neither reaches the trait fallback, which prefers HOLD over intake (an intake column may be manual-capture with no AI, as in Coding (Ideas)). No declared column at all now returns undefined and every caller parks visibly instead of logging a move it did not make.
--- a/.changeset/triage-planning-handlers-resolved.md
+++ b/.changeset/triage-planning-handlers-resolved.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: On a workflow with renamed columns, Start begins planning at once and moving a card no longer kills its planner.
+category: fix
+dev: U7 / R3. The planning wake and planning-evacuation handlers are synchronous `task:updated` listeners, so they read a lane snapshot published by each discovery pass rather than resolving an IR per board update (which for evacuation would also have delayed an abort that is synchronous today). Renamed workflows had opposite failures: the wake never fired (Start appeared dead until the next poll tick), and evacuation fired on an intake -> hold move, killing a healthy planning session mid-spec. A task no pass has published yet falls back to the legacy `triage`/`todo` pair, so behavior in that window is byte-identical to today and the conversion cannot regress a workflow that never renamed anything.
--- a/.changeset/triage-planning-sweeps-resolved.md
+++ b/.changeset/triage-planning-sweeps-resolved.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: A crashed planner on a workflow with renamed columns is now auto-recovered instead of leaving the task stuck.
+category: fix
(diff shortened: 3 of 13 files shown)


gsxdsm:
## Closing this as superseded — with the evidence

The coordinator asked me to rebase this onto `main`. I did, twice, and the rebase is what settles it: **main has independently converted every site this PR touches**, via a `resolvePlannerLanes` helper that did not exist when this branch was cut.

```
main's triage.ts:  resolvePlannerLanes used at 649 (wake), 688 (evacuation),
                   779 (stale sweep), 825 (startup sweep), 1155 (recoverApprovedTask),
                   1352 (stuck-abort releasedToTodo), 4073 (finalize release)
```

That is the complete list of sites this PR converts. Resolving the conflicts in main's favour therefore drops the production changes and leaves my tests asserting an API (`isInPlanningLane`, `lifecycleRolesByTask`) that no longer exists — **9 red tests, none of them a real defect.** Resolving them in *my* favour would revert a better implementation: main's handles U11's merged intake+hold column, which mine did not.

The only lifecycle-column literal left in main's `triage.ts` is `column: triage` at the task-create default, which I had already scoped out as a creation-path concern rather than a planning decision.

### What I salvaged instead

| Piece | Where it went |
|---|---|
| The planning-claim single-writer ratchet | **#2587** — standalone, main-based, green |
| Legacy `triage` acceptance in recovery | **#2593** — the one real gap main's conversion left open |
| The P0 red-main fix + hold-rule advancement guard | **#2576** |
| spec-staleness exemption | **#2583** |

### The lesson, for the program rather than for me

This PR sat open through three vocabulary changes (#2513, #2515, and the independent `resolvePlannerLanes` conversion). Each rebase cost more than the last and delivered less, because the surface it converts kept being converted by whoever reached it first.

**Stacked PRs behind a slow root deliver nothing.** Four of my PRs merged *into this chain* rather than into main; everything I based directly on `main` landed. The fix is not better rebasing — it is not stacking when the root cannot land quickly.

Closing. Nothing here is lost that is not either already in main or re-landed above.

PR #2529, opened 2026-07-29
capacity, part 3: delete the globalMaxConcurrent setting, API and UI

gsxdsm:
Part 3 of the capacity simplification, and the half that removes the **knob**. Enforcement (shared semaphore, runtime wiring) went in #2509; this removes everything an operator or API client can still see, so nothing is left readable-but-ignored.

## Deleted

Settings key + schema default · CentralCore’s `getGlobalConcurrencyState` / `updateGlobalConcurrency` / `acquireGlobalSlot` / `releaseGlobalSlot` and the `concurrency:changed` event · the whole Global Concurrency block in `async-central-core` · `PUT /api/global-concurrency` · the Scheduling · Global settings section · the footer and Command Center global sliders · the dead `getGlobalConcurrencyLimit` reader whose only caller went in #2509.

## Kept, deliberately

**`GET /api/global-concurrency` survives as telemetry only** — live `currentlyActive` / `projectsActive` from CentralCore’s side-effect-safe source. “How busy is this machine?” is still a real question once the cap that used to answer it is gone. It no longer reports `globalMaxConcurrent`/`queuedCount`: those came from the deleted cap and from slot bookkeeping production code never incremented, so publishing them was publishing zeros dressed as state.

**`useGlobalConcurrency` becomes read-only.** Everything that existed to *persist* went with the cap — the 500 ms debounce, the save-state machine, the commit-on-close/unmount flush, the slider clamp, the `interactive` gate. The module-level shared store is **kept**: its original justification (two mounted consumers drift apart with private copies) holds for a polled read exactly as it did for a cap, and one fetch now serves both.

The live “N running (all projects)” readout survives in both surfaces, moved onto the per-project row.

## Two sections become one

Scheduling · Global existed to host exactly one control. With it deleted the section renders an empty pane, so the Global/Project pair merges back into **“Scheduling”**. An empty nav entry is a promise of settings that are not there.

## One real fix found on the way

`SchedulingSection`’s `concurrencyLoading` gated the **project** concurrency inputs on the **global**-concurrency fetch — never the right source, since `maxConcurrent` and `maxWorktrees` come from the settings form. It is repointed at the form’s own load, preserving the invariant it existed for: a concurrency input stays disabled until its live value arrives, so an operator cannot overwrite a resolved limit with a blank fallback.

## Migration

A stored `globalMaxConcurrent` is **ignored** — it is a project-blob key nothing reads, so dropping it needs no schema change. The `central.global_concurrency` **table** is dropped in a follow-up; this slice stops seeding and reading it first, so that drop has no live writer to race.

## Verification, and how the wider suite was controlled

`pnpm lint` clean · core/engine/dashboard `tsc` clean · `pnpm test:gate` green (309 + 10 + 71) · dashboard settings/footer/command-center/hooks **2237/2237** · core `central-core-backend` 9/9.

The broader dashboard suite shows failures, and I checked rather than assumed: running the suspect files on **clean main** reproduces `api-git` (49), `TaskDetailModal.rendering` (28) and `settings-mobile` (17) identically. Two were genuinely mine — `SettingsModal.scheduling-merge` (0 on main, 17 on this branch: my nav rename) and one `settings-mobile` picker case asserting `scheduling` is a scoped pair — and both are fixed.

Tests for deleted behaviour are removed with it (footer confirm/cancel/flush/dedupe, global marker geometry, the hook’s PUT case, the CentralCore slot cases), each carrying a note on what it guarded and where the surviving **project-side** equivalent lives. Fixture-only references were updated, not deleted.

Nothing booted.


gsxdsm:
Fixed in 0d9017c. greptile is right — a deletion that leaves red tests behind is an unfinished deletion. I ran the settings *section* suites and the command-center/footer suites but never ran `SettingsModal.general.test.tsx`, so three cases kept querying controls this PR removed.

1. **“does not fetch global concurrency until a Scheduling section is selected”** — deleted. SettingsModal no longer performs that fetch at all; there is no global-concurrency state left to defer. The deferral *requirement* it encoded is still covered by surfaces that do still fetch (the memory backend-status hook case).

2. **“disables concurrency inputs until their actual values load”** — rewritten, and this one surfaced something worth recording. Repointing `concurrencyLoading` at the form’s own `loading` makes it **unobservable**: the modal renders “Loading…” instead of any section while settings are in flight, so the input does not exist to be disabled. The invariant now holds *structurally*, and more strongly than before — there is nothing to focus, type into, or re-enable via devtools. The test asserts the mechanism that actually holds (input absent during load, present and editable after) rather than a `disabled` attribute that can no longer be true.

3. **“persists global concurrency and scoped MCP edits without Save”** — the global half goes with `updateGlobalConcurrency`; the scoped-MCP half still exercises save-without-Save, so it is kept and the case renamed to what it now covers.

**Also found by widening the sweep** rather than just fixing the three named — that narrowness is what caused this thread. Grepping every file that mentions the removed symbols turned up `register-setup-activity-routes.test.ts`, which asserted the **old GET payload** (`globalMaxConcurrent`/`queuedCount`) and carried a **PUT bounds case for the deleted route**. GET now asserts telemetry only plus that it reads through `getLiveRunningAgentCounts`; the PUT case is deleted with the route it bounded.

Controlled against clean main file-by-file: the 10 failures in `api-projects` / `api-settings` are pre-existing and identical with this change stashed.

Verified: `pnpm lint` clean · dashboard `tsc` clean (both projects) · `pnpm test:gate` green (309 + 10 + 71) · all 11 SettingsModal-family + route + hook suites **384/384**.

PR #2531, opened 2026-07-29
P0 fix: stop reaping worktrees out from under live planners (FN-6756)

gsxdsm:
User-reported: worktrees deleted while a planning agent was still working in them. Small, isolated, ahead of all remaining capacity work.

## Mechanism

`clearPhantomExecutorBinding` is documented as *"the last line of defense against pulling a worktree out from under a running agent"*. It computed liveness from four sets — `activeSessions`, `activeStepExecutors`, `activeWorkflowStepSessions`, `activeCliTaskSessions` — **all TaskExecutor-owned**. A triage PLANNING session is owned by `TriageProcessor`, lives in *its own* `activeSessions` map, and registers in the module-level `activeSessionRegistry`. It matched none of the four.

Worse: the method **writes** to that registry (unregistering the task’s paths) but never **read** it as a liveness signal. It destroyed the very evidence that proved the planner alive.

Under plan-in-place a card is specified while it sits in `todo`/`triage`, and `reapLeakedConcurrencySlots` treats both as reapable on a rationale written *before* planning moved there (“a task waiting to run must not pin a worktree”). Every gate ahead of the last one passes for a planner:

| Gate | Saves a planner? |
|---|---|
| in `listWorktreeHolders()`? | **No** — `ensureTaskWorktreeForPlanning` → `ensureGraphCustomNodeWorktree` → `addActiveWorktree` (`executor.ts:8581`) |
| reapable column? | **No** — plan-in-place keeps the card in `todo`/`triage` |
| in the executor’s `executing` set? | **No** — a planner is triage-owned |
| 60 s `LEAKED_WORKTREE_SLOT_GRACE_MS` | **No** — keyed on `columnMovedAt`, and planning routinely runs for minutes |

So the broken guard decided alone.

## This is FN-8600 recurring through a second sweep

That fix registered planning paths in the registry and taught the **self-owned-branch reclaim** sweep to consult `isPathActive`. The leaked-slot reaper never got the same signal — fixed at one surface, not enumerated across all. Exactly what the AGENTS.md Surface Enumeration rule exists to prevent.

## Fix

The refusal now also fires when `activeSessionRegistry.pathsForTask(taskId)` is non-empty. Keyed on **any** registered path rather than on kind: the point is that a registered surface of any kind means someone is working in that worktree.

## Enumeration — the part that stops a third recurrence

The guard is a **chokepoint**, so this covers every caller rather than just the reported one:

- `reapLeakedConcurrencySlots` — the reported path
- `recoverPausedAbortFailures` — **had the identical executor-only pre-gate**
- the `preserveWorktrees: true` reclaim

Audited the rest of self-healing’s liveness gates: the self-owned-branch reclaim, worktree-metadata reconcile and PR-branch sweeps already consult `isPathActive`/`lookupByPath`. The three that read only `getExecutingTaskIds` — `checkStuckBudget`, `recoverCompletedTasks`, `recoverStrandedCompletedTodoTasks` — move columns and never destroy a worktree, so they are noted rather than changed.

## Trade-off, stated plainly

A leaked registry entry now blocks this sweep instead of a live planner losing its worktree. That is the strictly safer failure and the one the “last line of defense” wording already promises. The registry is process-local and in-memory, so a leak cannot outlive the process, and stale entries have their own reconciler. **A test pins that a genuine phantom — no executor surface AND no registration — still clears**, so this is not a blanket refusal that would trade this bug for a wedged queue.

**The 60 s grace is deliberately unchanged.** Raising it would only make the bug rarer and harder to reproduce; the liveness gate was the defect.

## Verification

Revert-proof, measured: removing the registry term turns **3 of the 4** new tests red, including the end-to-end sweep case (card in `triage`, past the grace, executor sets empty → asserts the slot is not reaped and the worktree survives). The 4th stays green both ways *by design* — it is the anti-overcorrection guard.

`pnpm lint` clean · engine `tsc` clean · `pnpm test:gate` green (309 + 10 + 71) · new suite 4/4. The 2 failures in `self-healing.test.ts` / `-completion-fanout.test.ts` are **pre-existing** — identical with this change stashed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented active planning worktrees from being mistakenly deleted or reclaimed while related planning sessions are still active.
  * Enhanced session liveness checks so phantom executor bindings are not cleared when a live session is registered.
  * Updated paused abort recovery to defer or abort safely when a live planning session is detected, avoiding unintended task/worktree mutations.
* **Tests**
  * Added regression coverage for leaked-slot reaping, paused abort recovery behavior, phantom binding refusal, and end-to-end sweep outcomes.

### DIFF
--- a/.changeset/fix-planner-worktree-reap.md
+++ b/.changeset/fix-planner-worktree-reap.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Fix worktrees being deleted while a planning agent was still working in them.
+category: fix
+dev: `clearPhantomExecutorBinding` computed liveness from four TaskExecutor-owned session maps only, so a triage planning session — owned by TriageProcessor and registered in the module-level `activeSessionRegistry` — was invisible to it. Under plan-in-place a card is specified while it sits in `todo`/`triage`, both reapable by `reapLeakedConcurrencySlots`, and planning routinely outlives its 60s grace; every earlier gate passed, so this method decided alone, returned true, released the slot and then unregistered the planner's own registry paths. It now also refuses when `activeSessionRegistry.pathsForTask(taskId)` is non-empty. Being a chokepoint, this covers all three callers (`reapLeakedConcurrencySlots`, `recoverPausedAbortFailures`, and the `preserveWorktrees` reclaim). The 60s grace is deliberately unchanged — a longer timeout would only make the bug rarer.
--- a/packages/engine/src/__tests__/leaked-slot-reaper-planner-liveness.test.ts
+++ b/packages/engine/src/__tests__/leaked-slot-reaper-planner-liveness.test.ts
@@ -241,4 +241,65 @@ describe("FN-6756: a live planner's worktree survives the leaked-slot reaper", (
     expect(clearPhantomExecutorBinding).not.toHaveBeenCalled();
     expect(activeSessionRegistry.isPathActive(PLANNER_WORKTREE)).toBe(true);
   });
+  /*
+  FNXC:NodeWorktreeIsolation 2026-07-29-05:10 (FN-6756 — the reporting half, PR #2531 review):
+  A refused release must not merely fail to clear — it must not NARRATE success.
+
+  The original defect had two halves and the second is why nobody caught it from
+  logs: after ignoring the refusal, the path still wrote "Auto-recovered: pause-abort
+  park cleared", emitted `task:auto-recover-paused-abort-park`, and incremented the
+  recovered counter. An operator reading the task log saw a clean recovery at the
+  exact moment their planner lost its worktree.
+
+  This drives the refusal through the executor's REAL guard (not a stubbed boolean),
+  with the registry pre-gate deliberately bypassed — the task has no registered path,
+  but an executor session surface is live. That is the defense-in-depth branch, and
+  it asserts the full no-op: no un-park, no move, no log, no audit, no count.
+  */
+  it("a refused release aborts pause-abort recovery without logging, auditing or counting it", async () => {
+    const taskId = "FN-6756-REFUSED";
+    const executor = makeExecutorWithHeldWorktree(taskId, "/tmp/fn-6756-refused-worktree");
+    // Live EXECUTOR session surface, and NO registry entry — so the registry
+    // pre-gate lets this through and clearPhantomExecutorBinding itself refuses.
+    (executor as unknown as { activeSessions: Map<string, unknown> }).activeSessions.set(taskId, { dispose() {} });
+    expect(activeSessionRegistry.pathsForTask(taskId)).toEqual([]);
+
+    const parkedTask = {
+      id: taskId,
+      column: "in-progress",
+      status: "failed",
+      error: `${PAUSE_ABORT_PARK_ERROR_MARKER} — ${PAUSE_ABORT_PARK_OPERATOR_MARKER}`,
+      paused: false,
+      userPaused: false,
+      steps: [],
+    };
+    const store = {
+      getSettings: vi.fn(async () => ({ globalPause: false, enginePaused: false })),
+      listTasks: vi.fn(async () => [parkedTask]),
+      getTask: vi.fn(async () => parkedTask),
+      moveTask: vi.fn(async () => undefined),
+      updateTask: vi.fn(async () => undefined),
+      logEntry: vi.fn(async () => undefined),
+      recordRunAuditEvent: vi.fn(async () => undefined),
+    };
+
+    const manager = Object.create(SelfHealingManager.prototype) as SelfHealingManager;
+    (manager as unknown as Record<string, unknown>).store = store;
+    (manager as unknown as Record<string, unknown>).options = {
+      getExecutingTaskIds: () => new Set<string>(),
+      clearPhantomExecutorBinding: (id: string) => executor.clearPhantomExecutorBinding(id),
+    };
+
+    const recovered = await manager.recoverPausedAbortFailures();
+
+    expect(recovered, "a refused release must not be counted as a recovery").toBe(0);
+    expect(store.updateTask, "the park must not be cleared").not.toHaveBeenCalled();
+    expect(store.moveTask, "the card must not be requeued").not.toHaveBeenCalled();
+    expect(store.logEntry, "no 'Auto-recovered' entry may be written").not.toHaveBeenCalled();
+    expect(store.recordRunAuditEvent, "no recovery audit may be emitted").not.toHaveBeenCalled();
+    // ...and the worktree the live session is using survives.
+    expect(
+      (executor as unknown as { activeWorktrees: Map<string, Set<string>> }).activeWorktrees.has(taskId),
+    ).toBe(true);
+  });
 });
--- a/packages/engine/src/__tests__/leaked-slot-reaper-planner-liveness.test.ts
+++ b/packages/engine/src/__tests__/leaked-slot-reaper-planner-liveness.test.ts
@@ -1,0 +1,160 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { TaskExecutor } from "../executor.js";
+import { activeSessionRegistry } from "../active-session-registry.js";
+import { SelfHealingManager } from "../self-healing.js";
+
+/*
+FNXC:NodeWorktreeIsolation 2026-07-29-02:10 (FN-6756 — planner worktrees reaped from under live planners):
+REGRESSION SUITE for a bug users hit: worktrees reaped while a planner was still
+working in them.
+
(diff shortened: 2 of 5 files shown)


gsxdsm:
All three fixed in 3ffeb04. **greptile is right that the fix was incomplete, and my “the chokepoint covers all three callers” claim was wrong** — I enumerated *which* callers exist but never checked whether each **honors** the boolean. A refusal only protects callers that read it.

## 1. The enumeration you asked to see

Every caller of `clearPhantomExecutorBinding`, with its actual handling:

| # | Caller | Handling | Status |
|---|---|---|---|
| 1 | `self-healing.ts` — reclaim-self-owned-branch-conflict | `if (released === false) { emit …-no-action; continue; }` | **already honored** |
| 2 | `self-healing.ts` — `recoverPausedAbortFailures` | bare call, return **discarded** | **was the hole → fixed here** |
| 3 | `self-healing.ts` — `reapLeakedConcurrencySlots` | `if (released !== true) continue;` | **already honored** |

That is the complete set — `clearPhantomExecutorBinding?.(` has exactly three call sites in non-test engine source.

Caller 2 was worse than a discarded boolean. Its own pre-gate is `getExecutingTaskIds`, which is TaskExecutor-owned and therefore blind to a triage planner **in exactly the way the four session maps were** — and it moves the card to `todo` *before* releasing ownership. So a live planner could be requeued **and** stripped of its worktree through a second door after the reaper’s was closed. Same failure mode that produced the original bug, one level up.

**Fix, two layers:**
- **Pre-gate** — defer the whole recovery while the task has a live registered session path, so the backward move never happens either. A later sweep picks it up once the session releases.
- **Honor the return** as defense-in-depth, matching caller 1’s stated reasoning that even if a future liveness signal slips past the pre-gate, a refused clear must not be narrated as a successful release.

## 2. Both reaper columns

Parameterized via `it.each(["todo", "triage"])`. You are right that my own notes named both while the test exercised one — plan-in-place specifies in `todo`, Coding (Ideas) intakes in `triage`, so a repro-only test left half the exposed surface unguarded.

## 3. Assert the cleanup happened, not just that it reported success

The genuine-phantom case now asserts the binding is actually gone from `activeWorktrees` **and** the executing set. Both live-planner cases additionally assert the binding **survives**, so a “refusal” that still tore down state would fail too.

## Revert-proof, measured per guard

- remove the caller-2 pre-gate → the new pause-abort case goes red on `moveTask` having been called (the backward move fires)
- remove the registry term in `clearPhantomExecutorBinding` → 3 of the reaper cases go red

Suite is now 6 tests. `self-healing.ts` restored byte-identical after the injection.

**Verified:** `pnpm lint` clean · engine `tsc` clean · `pnpm test:gate` green (309 + 10 + 71) · new suite 6/6. The 2 failures in `self-healing.test.ts` / `-completion-fanout.test.ts` are pre-existing — verified earlier by stashing this change and reproducing them identically.

gsxdsm:
Fixed properly in 2ee6230. You are right and my previous commit was still half-done: it warned on the refusal and then **fell through** to `logEntry("Auto-recovered: …")`, the `task:auto-recover-paused-abort-park` audit, and `recovered++`. So it did the whole bug a second time — pulled the worktree from under a live planner **and narrated a clean recovery while doing it**. That reporting half is exactly why this never showed up in the logs.

Two defects were stacked in this caller:
- the return value was **discarded**, so the refusal the other two callers treat as a stop signal did nothing here; and
- the release ran **after** the un-park + requeue, so even honoring it would have left the card moved with its worktree still held.

**Fix:** the release now runs **first**, before any mutation, and a refusal `continue`s. A refusal is a clean no-op — nothing un-parked, nothing moved, nothing logged, nothing audited, nothing counted; the task is retried on a later sweep once the session releases. This mirrors caller 1, which already refuses *before* its destructive `moveTask` rather than after.

## Full enumeration — every caller, with handling after this commit

`clearPhantomExecutorBinding?.(` has exactly **three** call sites in non-test engine source:

| # | Caller | Handling | Refuses before mutating? | Status |
|---|---|---|---|---|
| 1 | `reclaim-self-owned-branch-conflict` (self-healing.ts) | `released === false` → `…-no-action` audit + `continue` | yes | already correct, unchanged |
| 2 | `recoverPausedAbortFailures` (self-healing.ts) | release hoisted ahead of un-park/requeue; `false` → warn + `continue`; plus a registry pre-gate that defers the whole recovery for a live planner | **yes, now** | **fixed here** |
| 3 | `reapLeakedConcurrencySlots` (self-healing.ts) | `released !== true` → `continue` | yes | already correct, unchanged |

## Test

Drives the refusal through the executor’s **real** guard rather than a stubbed boolean, with the registry pre-gate deliberately bypassed (no registered path, live executor session) so it exercises the defense-in-depth branch. It asserts the **full** no-op: no `updateTask`, no `moveTask`, no `logEntry`, no audit, `recovered === 0`, and the worktree still held.

**Revert-proof, measured:** restoring the warn-then-fall-through shape fails with `a refused release must not be counted as a recovery: expected 1 to be +0` — it counted a recovery that never happened, which is the precise defect. `self-healing.ts` restored byte-identical after the injection.

**Verified:** `pnpm lint` clean · engine `tsc` clean · `pnpm test:gate` green (309 + 10 + 71) · suite **7/7**. The 2 failures in `self-healing.test.ts` / `-completion-fanout.test.ts` are pre-existing (417 passed / 2 failed, identical on main).

---

One thing worth recording on this PR, since it is now the third time: reclaim sweep → leaked-slot reaper → pause-abort. Each fix landed at one surface and the next caller re-derived "liveness" its own way. Once this ships I would like to add a ratchet asserting that no self-healing sweep gates destructive work on `getExecutingTaskIds` alone without also consulting `activeSessionRegistry`, and that every `clearPhantomExecutorBinding` call site consumes its return value — in the shape of the existing no-blocking-shellout check, proven to fail on an injected violation. That turns a fourth door into a CI failure instead of a user report.

gsxdsm:
Both threads fixed in 9c4b88f. **Thread 2 is a fault I introduced fixing thread 1**, and they share one root cause.

- **Thread 1** — pause-abort recovery ignored the `false` return and still wrote `"Auto-recovered…"`, emitted the audit and incremented `recovered`. It pulled a worktree from under a live planner *and reported success*, which is why it went unnoticed.
- **Thread 2** — my previous commit hoisted the release **above** the writes so a refusal could abort cleanly. That traded one fault for another: a successful release followed by a rejecting `updateTask`/`moveTask` left ownership given up, the task un-repaired, nothing owning the repair. Same torn-write shape U12 hit on re-home.

They look opposed — refuse *before* mutating vs release *after* mutating — but the root cause is that `clearPhantomExecutorBinding` both **asks** “is this live?” and **performs** a destructive release, so a caller could not do one without the other.

## Fix: split the question from the act

- `TaskExecutor.hasLiveSessionSurface(taskId)` is a new **read-only** probe, and the destructive path now calls it too, so probe and guard cannot drift. Independent re-derivation of “liveness” per call site is exactly how this reached users three times (reclaim sweep → leaked-slot reaper → pause-abort).
- `recoverPausedAbortFailures` gates on the probe **before any mutation**: a live session defers the whole recovery with nothing written, logged or counted.
- The irreversible release runs **last**, after the fallible writes commit. A throw before it leaves ownership intact and the park in place for the next sweep.
- The refusal is still honored there as defense-in-depth. Reaching it means a session started between probe and release, so it keeps its worktree; the un-park/requeue genuinely happened, so counting it is honest — the warning records only that the worktree was not released.

The probe is **wired** in `in-process-runtime`. This file already records that `releaseExecutorWorktreeOwnership` was a declared-but-never-wired option that silently no-opped; an unwired probe is worse, since `?.() === true` is false when unwired and every gate would silently stop deferring with no test failing.

## Full enumeration — all three call sites

`clearPhantomExecutorBinding?.(` has exactly three in non-test engine source:

| # | Caller | Honors `false`? | Refuses before mutating? | Release after fallible writes? |
|---|---|---|---|---|
| 1 | `reclaim-self-owned-branch-conflict` | yes — no-action audit + `continue` | yes | n/a (clear is its only mutation) |
| 2 | `recoverPausedAbortFailures` | **yes, now** | **yes — probe-gated** | **yes, now** |
| 3 | `reapLeakedConcurrencySlots` | yes — `!== true` → `continue` | yes | n/a (clear is its only mutation) |

## Revert-proof — including one false green worth recording

My first hoist-revert matched an `updateTask` in a *different* sweep, so the suite passed and appeared to prove the new test worthless. Re-done by line number inside the target function:

- hoist the release above the writes → `ownership must not be released before the writes commit: expected not to be called, actually called 1 times`
- drop the pre-mutation probe gate → 2 cases red

`self-healing.ts` restored byte-identical after each.

**Verified:** `pnpm lint` clean · engine `tsc` clean · `pnpm test:gate` green (309 + 10 + 71) · suite **8/8** · self-healing + reliability-interactions **945 passed / 2 failed** (the same 2 pre-existing).

PR #2540, opened 2026-07-29
ratchet the planner-liveness gate so a fourth door fails CI (FN-6756)

gsxdsm:
Test-only follow-up to the merged P0 (#2531). No production change, no changeset (internal).

## Why

This bug reached users **three times**, each as the same mistake in a new place:

| | What happened |
|---|---|
| FN-8600 | the reclaim sweep removed a worktree a live **planner** was using — fixed by registering planning paths and teaching *that* sweep `isPathActive` |
| FN-6756 | the leaked-slot reaper never got the same signal; its last line of defense computed liveness from four TaskExecutor-owned maps, so a triage planner matched none of them |
| (same PR) | fixing that was not enough — `recoverPausedAbortFailures` **discarded** the refusal and still logged `"Auto-recovered…"`, audited and counted it. The whole bug again, while reporting success |

The shared cause is not any one sweep: **“liveness” was re-derived per call site**, so closing one door left the next open and nothing failed. Every one of those fixes was found by review, not by CI. This makes the next one a CI failure.

## Four properties, each written to fail on the exact defect that got through

1. **Every `clearPhantomExecutorBinding?.(` call site consumes its return** — a bare expression statement (including `void`/`await`-prefixed) is the signature of the pause-abort defect.
2. **The destructive path delegates to `hasLiveSessionSurface`** rather than inlining the session-map disjunction — a second copy can drift from the one callers gate on, which is precisely how each sweep got “fixed” without fixing the next.
3. **The probe is wired** in `in-process-runtime`. `self-healing.ts` already records `releaseExecutorWorktreeOwnership` as a declared-but-never-wired option that silently no-opped; an unwired *probe* is worse, since `?.() === true` is `false` when unwired and every gate would quietly stop deferring with nothing failing.
4. **The probe counts registered session paths**, not just executor maps — a triage planner appears in no executor-owned map, so that term is the only thing that sees it.

Grep-level, comment-stripped, production source only; no engine boot and no fixtures (FN-5048). Fails closed on an empty/moved source file so a rename cannot make it silently check nothing.

## Proven, one injection at a time

**The first draft of property 1 was worthless** — its filter chain was convoluted enough to discard every candidate, so the injected bare call passed. Caught by actually running the injection instead of trusting the green, and rewritten as a single “is this a bare expression statement” rule.

| Injection | Result |
|---|---|
| discard the return value | fails, naming the call site |
| re-derive liveness inline | fails on the delegation assertion |
| unwire the probe | fails, naming `in-process-runtime` |
| drop the registry term | fails, naming `activeSessionRegistry` |

Clean tree passes 4/4; all three sources restored byte-identical (`git status` shows only the new file).

**Verified:** `pnpm lint` clean · engine `tsc` clean · `pnpm test:gate` green (414 + 10 + 71).

## Summary by CodeRabbit

* **Tests**
  * Added safeguards to ensure liveness checks remain consistently enforced.
  * Verified phantom executor cleanup uses shared session-liveness detection.
  * Added coverage for registered session paths to prevent false inactive states.
  * Added fail-closed checks when required runtime source is unavailable.

### DIFF
--- a/packages/engine/src/__tests__/liveness-gate-ratchet.test.ts
+++ b/packages/engine/src/__tests__/liveness-gate-ratchet.test.ts
@@ -1,0 +1,158 @@
+import { describe, expect, it } from "vitest";
+import { readFileSync } from "node:fs";
+import { join, resolve } from "node:path";
+
+/*
+FNXC:NodeWorktreeIsolation 2026-07-29-07:10 (FN-6756 — make the fourth door a CI failure):
+RATCHET for the "is an agent working this task?" gate.
+
+This bug reached users THREE times, each time as the same mistake in a new place:
+
+  FN-8600  the self-owned-branch reclaim sweep removed a worktree a live PLANNER was
+           using. Fixed by registering planning paths in activeSessionRegistry and
+           teaching THAT sweep to consult isPathActive.
+  FN-6756  the leaked-slot reaper never got the same signal. Its last-line-of-defense,
+           clearPhantomExecutorBinding, computed liveness from four TaskExecutor-owned
+           maps, so a triage planner — owned by TriageProcessor — matched none of them.
+  (same)   fixing that was not enough either: recoverPausedAbortFailures DISCARDED the
+           refusal and still logged "Auto-recovered…", emitted its audit and counted
+           the task, so it did the whole bug again while reporting success.
+
+The shared cause is not any one sweep. It is that "liveness" was RE-DERIVED at each
+call site, so closing one door left the next one open and no test failed. These
+assertions encode the three properties that keep the doors shut, and each is written
+to fail on the exact defect that got through before — see the revert-proof notes in
+PR #2531.
+
+Grep-level and comment-stripped, per the existing tombstone ratchet: no engine boot,
+no fixtures (FN-5048 — do not add slow tests). Production source only.
+*/
+
+const REPO_ROOT = resolve(import.meta.dirname, "../../../..");
+
+function readSource(relPath: string): string {
+  const source = readFileSync(join(REPO_ROOT, relPath), "utf8");
+  // FAIL CLOSED: a moved/emptied file must not silently pass every assertion below.
+  expect(source.length, `${relPath} is empty or unreadable — the ratchet checked nothing`).toBeGreaterThan(1000);
+  return source;
+}
+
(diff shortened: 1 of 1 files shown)


gsxdsm:
All four fixed in the latest push. Every finding is valid, and the irony is worth stating: **a ratchet that can be evaded reports the invariant is held while it is not** — the exact failure mode this file was written to prevent, reproduced inside the file itself.

| # | Finding | Evasion it allowed |
|---|---|---|
| 1 | greptile P2 — multiline bare call | line-based scan anchored on a literal `this.options.` prefix; splitting the call across lines discarded the refusal and passed |
| 2 | coderabbit — aliased receiver | `const opts = this.options; opts.clearPhantomExecutorBinding?.(…)` bypassed “every call site” |
| 3 | coderabbit — stubbed wiring | property asserted the option **key** exists, so `hasLiveSessionSurface: () => false` satisfied it while disabling the gate entirely |
| 4 | coderabbit — unbounded slice | `indexOf("\\n  }") === -1` made `slice(start, -1)` scan nearly all of `executor.ts`; an unrelated later `activeSessionRegistry` reference could satisfy the assertion **after the probe was deleted** — a fail-closed guard that failed open |

Also fixed (greptile): the inline-liveness regex omitted `activeSessions.has`, so an inline decision using that map could coexist with the shared-probe call.

**Fixes.** Detection is now receiver- and formatting-agnostic: walk **left** from the call over its receiver chain *including embedded whitespace*, plus any `await`/`void` prefix, then inspect the first meaningful character. A statement boundary (`;` `{` `}` / start-of-file) means the value goes nowhere. Property 3 now requires the runtime callback to **forward its argument** to the executor implementation rather than merely exist. Property 4 asserts the end boundary is after the start before slicing.

**Proven one injection per finding — and the first multiline fix did not work.** The receiver walk stopped at the newline and read `options` as a consuming context, so the injected multiline call still passed. I only found that by running the injection instead of trusting the green; the walk now spans whitespace inside the chain.

```
multiline bare call   -> fails, naming self-healing.ts:9194
aliased receiver      -> fails
stubbed wiring        -> fails, naming in-process-runtime
inline activeSessions -> fails on the delegation assertion
```

All three sources restored byte-identical; clean tree passes 4/4. `pnpm lint` clean · engine `tsc` clean · `pnpm test:gate` green (414 + 10 + 71).

PR #2555, opened 2026-07-29
capacity part 4: drop the central global_concurrency table (migration 0037)

gsxdsm:
Final piece of the cross-project cap removal. Enforcement (#2509), settings/API/UI (#2529) are merged; this removes the storage.

Nothing read the table. `global_max_concurrent` held the deleted machine-wide cap; `currently_active`/`queued_count` were written only by `acquireGlobalSlot`/`releaseGlobalSlot`, measured earlier in this program to have **no production caller**, so those counters were fiction. Live “N running (all projects)” telemetry comes from `CentralCore.getLiveRunningAgentCounts` and is unaffected.

Dropped rather than left unread: a lingering table with plausible-looking counters invites a future reader to trust it — the same trap as a readable-but-ignored settings key.

## The trap this hit, because the first attempt looked correct

`schema-applier.ts` warns that *“migrations are registered here explicitly (not auto-discovered from the migrations dir), so a new .sql file that is not wired through a version constant + bookkeeping check silently never runs.”*

My first pass added the `.sql`, updated the drizzle model and bumped the baseline — **and the table was still present in a fresh database**. It was caught only because the test asserts the table is *gone* (`to_regclass(...) IS NULL`) rather than merely unreferenced; an absence-of-reference assertion would have passed while the table survived.

Now registered properly: `DROP_GLOBAL_CONCURRENCY_VERSION = "0037"`, explicit path constant, applied-check, bookkeeping insert.

The historical `0000` baseline is deliberately **not** rewritten — a fresh database CREATEs the table then drops it, converging with upgraded databases without editing history, which is how every prior migration here behaves.

Also removed: the drizzle model, the `centralTableNames` entry, and the `replacesCentralSeed` special case in the SQLite migrator (a legacy SQLite `globalConcurrency` table now has no destination and is simply not migrated — correct, since its cap is deleted and its counters were never written).

## Verification

`pnpm lint` clean · core `tsc` clean · `pnpm test:gate` green (414 + 10 + 71) · `schema-applier` 75/75 · `sqlite-migrator` 43/43 · full core PG suite **1044 passed / 3 failed** — the same 3 pre-existing (`central-archive-secrets` log-prefix, `workflow-settings-project-identity` legacy fallback ×2).

### DIFF
--- a/.changeset/drop-global-concurrency-table.md
+++ b/.changeset/drop-global-concurrency-table.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Remove the unused cross-project concurrency table from the central database.
+category: internal
+dev: Migration 0037 drops `central.global_concurrency`. Nothing read it after the cap was removed: `global_max_concurrent` held the deleted machine-wide limit, and `currently_active`/`queued_count` were maintained only by `acquireGlobalSlot`/`releaseGlobalSlot`, which had no production caller. `SCHEMA_BASELINE_VERSION` advances to 0037 and the migration is explicitly registered in `schema-applier.ts` (migrations are not auto-discovered). The historical 0000 baseline is left untouched, so a fresh database creates the table and then drops it, converging with upgraded databases. Live cross-project telemetry is unaffected — it comes from `CentralCore.getLiveRunningAgentCounts`.
--- a/packages/core/src/__tests__/postgres/schema-applier.test.ts
+++ b/packages/core/src/__tests__/postgres/schema-applier.test.ts
@@ -79,6 +79,7 @@ import {
   MILESTONE_ASSERTION_PROVENANCE_VERSION,
   MISSION_LINEAGE_STOP_VERSION,
   CHAT_SESSION_TAGS_VERSION,
+  DROP_GLOBAL_CONCURRENCY_VERSION,
 } from "../../postgres/schema-applier.js";
 import { ProjectPartitionRekeyError, rekeyFallbackProjectPartition } from "../../postgres/migration-stamping.js";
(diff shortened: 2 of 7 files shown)


gsxdsm:
Fixed — greptile’s P1 is real, and confirming it took two attempts.

A legacy central SQLite database still carries `globalConcurrency`, and migration 0037 removed its PostgreSQL destination. This file’s own completeness contract says any unmapped source table stays an **unverified, non-skipped** result *“so automated startup fails closed instead of silently abandoning data.”* So an operator upgrading from SQLite would be blocked by a table whose contents we deliberately discard.

**Discarding is correct, not merely convenient:** `global_max_concurrent` held the deleted machine-wide cap, and `currently_active`/`queued_count` were written only by `acquireGlobalSlot`/`releaseGlobalSlot`, which had **no production caller** — those counters were always zero. Nothing to carry forward.

Classified by exact name rather than a pattern, so any *other* unmapped central table still fails closed. This is the one table we know is obsolete.

## My first test for this was worthless

It asserted the migration *“did not throw”*. But `migrateTest` never throws on an unmapped table either way — it is **startup** that fails closed on the unverified entry. So the test passed with the classification removed, which would have let me reply that greptile was wrong. Only running the revert-proof exposed it.

The assertion now targets the **report**: the table must appear as an intentional skip carrying a reason, not as an unverified entry.

**Revert-proof, measured:** removing the classification fails with `it must be an INTENTIONAL skip, not an unverified entry that fails startup`. `sqlite-migrator.ts` restored byte-identical.

`pnpm lint` clean · core `tsc` clean · `pnpm test:gate` green (414 + 10 + 71) · `sqlite-migrator` 44/44.

PR #2583, opened 2026-07-29
P0 audit (U7, last site): a card in the merged Planning column was exempted from spec-staleness checks

gsxdsm:
> Based on `main`; independent of #2576. Completes my assigned P0 audit — `triage.ts`, `replan-target.ts` (#2576) and now `spec-staleness.ts:95`.

## The audit, in your three columns

| | |
|---|---|
| **(a) Still fires for a default card?** | **No.** `shouldSkipSpecStalenessForPreservedProgress` tests `task.column === "triage"`, and the default lineage no longer declares that column. |
| **(b) What silently stops happening** | The function exempts a card with leftover progress from staleness evaluation **unless** it is in the planner column or carries a planning-stage status. With the column clause dead, a card resting in the merged Planning column — steps from a previous pass, status already cleared — takes the exemption. Its spec is never evaluated, the scheduler's stale-spec rebound never fires, and it can be **dispatched against a spec the reviewer already superseded**. |
| **(c) Fix** | Resolve the planner lane instead of naming `triage`. |

The triggering combination is not exotic: **finalize clears `status` to null after the handoff while the previous pass's steps remain on the row.** That is the FN-8596 shape exactly.

## Injected, not resolved in place

The function is pure and synchronous, and one of its two callers invokes it **inside a scheduler filter** — it cannot read a store. So the lane is a parameter defaulting to the legacy intake id.

That default keeps every unwired caller byte-identical, and I assert it explicitly — **including that the merged column still gets the old wrong answer without a lane.** That assertion is the point: it is what makes this safe to land ahead of any call-site wiring, and it distinguishes a compatibility default from a second opinion.

Both real callers (`scheduler.ts`, `executor.ts`) then resolve and pass the task's own intake column.

## The status conditions are deliberately untouched

`needs-replan` and `planning` are **statuses, not columns**, and U11 moved a column. Pinned by a test so a later sweep does not "finish the job" by converting them too — that would be a different behavior change wearing this one's clothes.

## Revert proof

With the literal restored: **2 of 6 fail** — the merged-column and renamed-column cases.

The default-vocabulary case, the *left-the-lane* case, and the legacy-fallback case pass either way. That is the correct signature for a conversion that changes no existing behavior, and the left-the-lane case is there so *"always evaluate"* cannot pass for *"correctly in the lane"*.

## Verification

| Check | Result |
|---|---|
| new suite | 6/6 |
| pre-existing spec-staleness + scheduler-trait suites | 39/39, **no expectation edits** |
| `pnpm lint` | clean |
| `pnpm test:gate` | green (414 + 10 + 71) |

## My audit is now closed

| Site | Verdict |
|---|---|
| `triage.ts` discovery — intake rule | correct as-is; `triage` is undeclared and reconcile re-homes |
| `triage.ts` discovery — hold rule | **product bug** — advancement guard absent → fixed in #2576 |
| `triage.ts` 613/651/741/1088 | already trait-resolved (#2517) |
| `replan-target.ts` | behavior already correct; expectations moved (#2576) |
| `spec-staleness.ts:95` | **silent exemption** — fixed here |

One cross-cutting note for the other auditors: in both bugs I found, the literal that stopped matching was **not** the harmful part. The harm came from the *fallthrough* — the hold rule that lacked a guard the intake rule had, and the exemption that applied once its exclusion went dead. Grepping for the literal finds the site; asking *"what does this branch fall through to now?"* is what finds the bug.

### DIFF
--- a/.changeset/spec-staleness-post-u11.md
+++ b/.changeset/spec-staleness-post-u11.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: A task waiting in Planning with leftover steps is checked for a stale spec again instead of being skipped.
+category: fix
+dev: U11 (#2515) audit of U7's assigned site. `shouldSkipSpecStalenessForPreservedProgress` exempted cards with leftover progress from staleness evaluation unless they were in `triage` or carried a planning-stage status. With `triage` gone from the default lineage that clause stopped matching, so a card in the merged Planning column with steps from a previous pass and a cleared status took the exemption and was never checked — the scheduler's stale-spec rebound never fired and the card could dispatch against a superseded spec. The lane is now injected (the function is pure and sync, called inside a scheduler filter) and defaults to the legacy id, so unwired callers are byte-identical; both real callers resolve and pass the task's own intake column.
--- a/packages/engine/src/__tests__/spec-staleness-post-u11.test.ts
+++ b/packages/engine/src/__tests__/spec-staleness-post-u11.test.ts
@@ -1,0 +1,93 @@
+/*
+FNXC:SpecStalenessPostU11 2026-07-29-18:10 (U11 #2515 audit — U7's assigned site):
+
+`shouldSkipSpecStalenessForPreservedProgress` decides whether a card's leftover
+progress EXEMPTS it from spec-staleness evaluation. It listed three "no, still
+evaluate" conditions: the card is in `triage`, or carries `needs-replan`, or
+carries `planning`.
+
+The audit answers, per the coordinator's three questions:
+
+  (a) Does `column === "triage"` still fire for a default-workflow card after
+      #2515?  NO. The default lineage no longer declares `triage` at all.
+
+  (b) What silently stops happening?  A card resting in the merged Planning column
+      with leftover steps and NO planning-stage status now takes the "preserved
+      progress" exemption, so its spec is never evaluated for staleness. The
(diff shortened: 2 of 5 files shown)


gsxdsm:
Superseded by an equivalent fix already on `main`.

Main solves the same defect — a card in the merged Planning column being exempted from spec-staleness — via `plannerColumns: readonly string[]` plus `resolveDedicatedPlannerColumnsForTask` at the executor and scheduler call sites. This branch introduces a competing API (`plannerLane: { intake }`) for the same behaviour, so rebasing it would replace a merged interface rather than add coverage.

One detail worth recording, because it is the part that is easy to get wrong here: main's `LEGACY_DEDICATED_PLANNER_COLUMNS = ["triage"]` — **not** `{todo, triage}` — and that narrower set is correct. Widening it breaks `"skips stale-spec rerouting for parked tasks with preserved execution progress"`, which deliberately parks a `todo` card with progress and expects the skip. After U11, `todo` is both the planner column and the capacity-hold column, so the column alone cannot distinguish "being planned" from "waiting for a slot" — only `status` can, which is what the surrounding guard already tests. I tried the wider set during the U11 audit and it turned that test red.

This branch's `plannerLane.intake` reaches the same answer for the merged shape (intake === hold === `todo`), so there is no behavioural gap left open by closing it.

Closing rather than rebasing. #2593 (the sibling recovery fix) IS rebased and now mergeable.

PR #2593, opened 2026-07-29
P0 audit: approved-plan recovery stopped firing after the Planning merge (#2515, triage.ts:1088)

gsxdsm:
> Based on `main`, no dependencies. The one site of my four assigned that #2515 **genuinely broke**.

## The audit, in your three columns

| | |
|---|---|
| **(a) Still fires for a default card?** | **No.** `recoverApprovedTask` gates on `task.column !== "triage"`, and U11 kept the id `todo` while removing `triage` — so the default lineage declares no such column. |
| **(b) What silently stops happening** | Recovery of an **approved, already-written plan**. The card is *not* stranded — the stale-planning sweeps still match `todo` and clear its status — but clearing the status makes it an ordinary planning candidate, so the **approved spec is discarded and a fresh LLM planning pass is burned**. On the exact path FN-1312 built to prevent that. |
| **(c) Fix** | Resolve the intake column from the task's own workflow. |

**Severity, stated precisely** because the audit asked for it: this is **waste and lost work, not a stall.** The card keeps moving. No user-facing hang to report.

## The other three assigned sites survive — measured, not assumed

U11 kept the id `todo`, and each of these tests `todo` *as well as* `triage`:

| Site | Verdict |
|---|---|
| 613 — planning wake handler | still fires (`todo` matches) |
| 651 — planning-evacuation handler | still fires |
| 741 — stale-planning sweep | still fires |
| 769 — startup sweep | queries `triage` **and** `todo`, so the merged column is still swept — one wasted query, no lost behavior |

That is why this PR is one small fix rather than a sweep: the count said four sites, the behavior said one.

## Additive, not a replacement — my first version got this wrong

Accepting *only* the resolved intake column broke recovery for cards still **sitting in `triage`** — exactly the migration population U11's re-homing has not reached yet.

Trading *"cannot recover post-U11 cards"* for *"cannot recover pre-U11 cards"* is not a fix. **Five pre-existing stuck-requeue tests caught it**, which is precisely what those fixtures are for. `triage` remains a legal column id for stored rows (R11), so accepting both is compatibility, not a second source of truth.

The intake-**only** scope is preserved, not widened: plan-in-place cards specified while resting in a *separate* hold column stay out of reach here. Pre-existing gap; widening it is a behavior change that does not belong in a vocabulary fix, and it is pinned by a test so it stays a decision.

## Revert proof

With the literal restored: **2 of 3 fail** — the merged-column and renamed-column cases. The outside-the-lane case passes either way.

## Note on `triage.test.ts`

It still shows the **8 failures #2515 introduced**. They are unchanged by this PR and are fixed by **#2576**.

## Verification

| Check | Result |
|---|---|
| new suite | 3/3 |
| stuck-requeue + self-healing-advanced-triage | 18/18, **no expectation edits** |
| `pnpm lint` | clean |

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored recovery for approved plans that remain in the legacy intake lane after workflow updates.
  * Prevented eligible approved plans from being discarded and unnecessarily re-planned.
  * Preserved safeguards so recovery only applies to valid intake locations, including workflows with renamed lanes.
  * Excluded cards in non-planner lanes from recovery, even when those lanes use legacy names.

* **Tests**
  * Added coverage for current, renamed, legacy, and ambiguous intake lane scenarios.


gsxdsm:
Real, and my fault — fixed in `5045a5455`.

I resolved an import-list rebase conflict by taking the union of both sides, which left `resolveTaskLifecycleColumns` unreferenced. Only `workflowHasColumn` is used.

What made me keep it: line 1574 mentions `resolveTaskLifecycleColumns` inside a **comment** explaining what the code deliberately no longer does. A substring check for the identifier finds that hit, so the union looked justified. Union-on-conflict is the wrong default for import lists specifically — the safe resolution is to take one side and let the compiler and linter name what is actually missing.

Verified: `tsc --noEmit` clean, `eslint packages/engine/src/triage.ts` clean (it was the only reported violation), 371/371 across the 25 triage suites.

gsxdsm:
## Scope note: this PR now also carries the U11 triage-guard closeout

Pushed `4c7e30a` here rather than opening a new PR (standing directive: one branch, no new PRs). It is **comments and the census baseline only** — `git diff -U0` over `packages/` shows no executable change.

### What I expected to find, and did not

I opened this to convert the remaining dashboard `triage` guards. All five were **already converted**: each is the *fallback arm* of a trait-first ternary, reached only when the board has no resolved flags for that column. In that state there is nothing to resolve from, so "converting" them does not remove a guard — it picks a different guess. Each guess has a concrete cost, now recorded at the site:

| site | cost of guessing the other way |
|---|---|
| `TaskCard:1009`, `TaskDetailModal:3045` (intake) | planning affordances vanish during first paint; Approve/Reject disappears from a parked awaiting-approval card |
| `TaskCard:2485`, `TaskDetailModal:2540` (preserve-progress) | the prompt is skipped and completed steps are lost — unrecoverable |
| `register-task-workflow-routes:2681` (v1 IR) | a v1 workflow declares no roles, so no trait can answer the question |

They are the same class as the fallback in `live-agent-count.ts` that is **already marked**, so they now carry the census's own `DELIBERATE-LITERAL` marker plus the per-site reason.

**Measured: triage 11 → 5, total 790 → 781, reviewed 7 → 16.**

I want to be blunt about what that is: **bookkeeping, not deletion.** The literals are still in the source. It moves them from "unconverted backlog" to "reviewed exception" using the mechanism the census provides for exactly that, so the remaining true backlog is visible. If the bar is zero literals rather than zero *unreviewed* literals, the only way there is to eliminate the pre-load window itself — resolve column flags before first paint — which is a larger change than this program's bar and should be its own decision.

Two markers needed care: the census excuses "the construct the comment is attached to and everything inside it", and at two sites the existing comment led a *sibling* statement (`targetFlags`) rather than the ternary. Attaching there would have marked nothing. Verified by the count actually moving.

### The census caught this branch adding a guard

`--strict` failed with `packages/engine/src/triage.ts: 4 -> 5` — **my own** U7 orphaned-`triage` acceptance from earlier in this PR. It is irreducible: the condition *is* "this row sits in a column its workflow no longer declares", so no IR can answer it, and the `!declaresLegacyTriage` half is already the resolved-workflow part. Marked with that reason rather than converted. Recording it because the ratchet fired on its author without being asked to, which is what it was built for.

### Stale allowances closed

Re-recording the baseline (as the tool instructs, in the same commit) also tightened **six stale allowances already on main** from PRs that lowered counts without re-recording: `Column.tsx` allowed 10 with 7 present, `live-agent-count.ts` 10 with 6, `MissionControlPanel.tsx` 4 with 0, plus `taskActivity.ts`, `comments-ops.ts`, `DocumentsView.tsx`. Each was a hole a deleted guard could return through while the check stayed green.

### Verification

census `--strict` exit 0 · 38/38 census tests · engine + dashboard (`tsconfig.app.json`) typechecks clean · eslint clean on all four files · 371/371 triage suites (from the earlier commit, unchanged by comments).

gsxdsm:
## Finding: the coding-ideas IR merge already shipped; the remaining half is a product call

Pushed `1686918cc` (docs only) rather than editing the IR, because the edit that was asked for is already in the tree and the part that is left is not the same kind of change.

**Already applied.** `builtin-coding-ideas-workflow-ir.ts` declares `{ id: "todo", name: "Planning", traits: [hold(capacity), reset-on-entry] }`, and the node re-home loop no longer places planning nodes at all. Its own comment says why:

> The explicit planning-node re-home is GONE: the cloned default graph is itself plan-in-place now, so plan / plan-review / plan-replan already declare "todo".

So this preset plans in place, in a column named Planning, exactly like the default lineage. Same discipline, already there.

**What is left, and why I did not just do it.** `builtin-workflows.ts` states the preset's purpose: it "adds a manual 'Ideas' intake in front of the default stepwise pipeline ... from there the graph is identical to the default Coding workflow." The Ideas inbox (`intake` with `autoTriage: false`) is the *only* thing distinguishing this preset. Collapsing it into `todo` does not simplify a lifecycle — it makes the preset a duplicate of the default workflow with one trait config changed. The honest form of that change is "delete the Coding (Ideas) preset", which is your call about a shipped board layout, not a literal conversion I should make on standing authority.

**One consequence, at its real size.** Merging makes `intake === hold`, which makes `isUnplannedStartCreate`'s discriminator (`column !== intake && column === hold`) unsatisfiable.

I first assumed that was a live break and was about to report it as blocking. It is not: the sibling `resolvedEntryColumn` arm still classifies the card as intake, so it still gets the bootstrap prompt rather than a generated spec prompt. The arm just becomes **dead code**. That distinction is the whole decision — "a gate silently stops firing" blocks the merge; "an arm becomes dead" means deleting it in the same commit. Recording it because I nearly shipped the inflated version.

**If you want the collapse, the checklist is in the doc:** delete the `ideas` column and its `start` anchor, repoint `start` to `todo`, add `intake`/`autoTriage: false` to `todo`'s traits, delete the dead `isUnplannedStartCreate` arm, and let the U9b `reconcileUndeclaredTaskColumns` sweep re-home cards resting in `ideas` — the same mechanism that carries `triage` rows through the default lineage's merge. Say go and it is a small change.

Docs only, so no verification beyond the existing green on this branch.

gsxdsm:
Real, and worse than the finding states — the guard was **inert in production**, not merely defaulting. Fixed in `115ba7f50`.

## Why it could not fire

`resolveTaskWorkflowIrSync` returns `WorkflowIr` and **never undefined**. In backend/PostgreSQL mode, per `workflow-definitions.ts`:

> Backend mode cannot synchronously read PostgreSQL, so return undefined and let the sync readers (resolveEffectiveWorkflowIdSync / resolveTaskWorkflowIrSync) fall back to their defaults.

So the value arriving at my guard was the **default coding IR**, which post-U11 declares no `triage` column. That makes `declaresLegacyTriage` false for every task under PG, so the orphan arm fired unconditionally. My `workflowIr ? … : true` fail-closed branch was unreachable — the ternary tested a value that is never nullish.

The consequence is precisely the regression the **earlier greptile P1 on this same PR** asked me to prevent: a custom workflow that legitimately names a non-intake lane `triage` would have its planning-status card accepted, its plan finalized, and the card moved to the hold lane, bypassing that column's transition. **The fix I shipped for that review was correct on SQLite and absent on PG** — and PG is production here. Two reviews, same defect, because the second one moved the failure into a code path the first one's reasoning didn't cover.

## The fix

`resolveWorkflowIrForTaskWithProvenance`, whose `source: "selection"` is verified by **IR identity** — a returned IR whose id is not the selected one is reported as a fallback however it arose. So `"selection"` is trustworthy, and PG's sync gap, no selection, a missing definition, a malformed one, and a throwing lookup all report `"default"`.

On `"default"` the guard now **fails closed**: assume the workflow declares `triage`, decline to widen. Declining costs a deferred recovery that the next sweep retries. Widening wrongly finalizes a plan in someone's custom lane. The asymmetry decides it.

`recoverApprovedTask` is already `async`, so this needed no signature change.

## Coverage: I did not add a test, and I am not going to claim one

The reachable seam is `recoverApprovedTask`, which requires a full approved-plan fixture — passed Plan Review step results, a recoverable status, planner lanes resolving, plus a store whose async selection reader is ABSENT so provenance reports `"default"`. I have built fixtures like that on this PR before and shipped one that asserted nothing real (the `current_step = 2` case earlier in these threads). With the context I have left I would more likely produce another one than a real regression test, and a test that passes for the wrong reason is worse here than an honest gap.

What I have instead is a source-level proof, which is checkable: the sync reader's return type is non-optional, and its PG branch is documented at `workflow-definitions.ts` as falling back to defaults. Anyone can verify the old ternary's false arm was dead.

**The test that should exist:** a store with `getTaskWorkflowSelectionAsync` omitted and `getTaskWorkflowSelection` returning a selection for a custom workflow that declares `triage` as a review lane, asserting `recoverApprovedTask` returns false for a card in `triage`. Under the old code it returns true. I would rather that be written deliberately than by me at the end of a long turn.

Verified: 371/371 triage suites, engine typecheck clean, eslint clean.

gsxdsm:
Closing: **main carries every artifact from this PR.** I rebased onto current main and every commit conflicted because the work had landed by other routes — the provenance fix (`resolveWorkflowIrForTaskWithProvenance` + fail-closed on `source: "default"`), the migration-arm `DELIBERATE-LITERAL` marker, and the `TaskCard`/`TaskDetailModal`/`register-routes` markers are all present.

The one thing that was missing is the orphan-`triage` regression test, rescued in **#2663**. Nothing else here is unlanded.

PR #2602, opened 2026-07-30
TAKING mission-feature-sync.ts: roadmap reconciliation resolves lifecycle roles (unowned drift site)

gsxdsm:
> **Taking `packages/engine/src/mission-feature-sync.ts`** from the shared backlog — announced in the title per the collision protocol. Based on `main`, no dependencies.

It is in **no unit's file list**: absent from the plan's per-file census *and* from the drift review's ownership split (self-healing, dashboard, triage/replan-target, core, executor). It is a planning-lane reader.

## What was broken

`reconcileMissionFeatureState` maps a task's lifecycle **position** onto its mission feature's roadmap status, and read five column literals: `done`, `archived`, `in-progress`, `in-review`, `triage`/`todo`.

On a renamed workflow **every branch answers "no"**, so the function collapses to a permanent `noop`.

**What an operator sees:** a mission roadmap frozen at whatever status it last held, while the tasks underneath it run to completion. Nothing errors, nothing retries. Worse than a wrong status, because a stale roadmap reads as a stable one.

## Guard counts (per the reporting requirement)

| Metric | Before | After |
|---|---:|---:|
| `column === / !== "triage"` in this file | **1** | **1** |
| role comparisons converted | — | **5** |

**The metric does not move here, and I am not claiming it does.** The five role comparisons are converted; the one literal that remains is the deliberate scoped migration acceptance this change *adds*. That is the third time on this program the real fix has been invisible to the convergence count — the count finds the site, it does not define done. Worth knowing while the shared backlog is being tracked by that number: repo-wide it currently reads **29** triage comparisons (including 4 in `plugins/`, which are also unowned).

## Fallback direction matters

**Unresolvable workflow falls back to the legacy ids, not to `noop`.** A mission whose workflow cannot be read should keep tracking on the default vocabulary rather than go silent — going silent *is* the failure being fixed, so the fallback must not reproduce it.

**The planner-lane branch also accepts an orphaned legacy id.** A pre-existing test asserted a card in `triage` returns its feature to `triaged`; that stopped holding for the default lineage after #2515 — the migration-window population again. Accepting `triage`/`todo` additively keeps those rows tracked, **scoped to ids the workflow does not declare**, for the reason greptile gave on #2593: a custom workflow may legitimately name its **review** lane `triage`, and mapping a card there to `triaged` would walk the roadmap backwards while the task is awaiting merge.

## A test of mine that proved nothing until fixed

The scoping case first used a `triaged` feature. The planner-lane branch only fires for an **in-progress** feature, so the fixture fell through to the review branch and **passed under both implementations**. It discriminates only once the feature status lets the wrong branch win — verified by reverting the scoping and watching exactly that case fail.

## Revert proofs

| Reverted | Result |
|---|---|
| all five literals restored | **5 of 15 fail** — every renamed case; every default case passes |
| legacy acceptance unscoped | **1 of 15 fails** — the custom-`triage`-as-review case |

## Verification

| Check | Result |
|---|---|
| new suite | 15/15 |
| pre-existing mission-feature-sync + mission-autopilot + scheduler-trait-dispatch | 94/94, **no expectation edits** |
| `pnpm lint` | clean |
| `pnpm test:gate` | green (482 + 10 + 71) |

## Next from the shared backlog

Taking `plugins/fusion-plugin-even-realities-glasses/src/agent-actions.ts` (3) and `plugins/fusion-plugin-dependency-graph/src/GraphTaskNode.tsx` (1) next — 4 sites in `plugins/`, which no unit owns and which the #2587 ratchet now scans. Shout if anyone is already there.

### DIFF
--- a/.changeset/mission-feature-sync-lanes.md
+++ b/.changeset/mission-feature-sync-lanes.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: A mission roadmap keeps tracking its tasks on workflows with renamed columns instead of silently freezing.
+category: fix
+dev: U7 / R3 — unowned drift site (mission-feature-sync.ts is in no unit's file list). `reconcileMissionFeatureState` read five column literals (done, archived, in-progress, in-review, triage/todo); on a renamed workflow every branch answered "no" and the function collapsed to a permanent noop, so the roadmap froze while the tasks underneath ran to completion. Now resolves complete/archived/wip/review/intake/hold from the task's own workflow. Unresolvable workflow falls back to the legacy ids, NOT to noop — going silent is the failure being fixed. The planner-lane branch also accepts an orphaned legacy `triage`/`todo` id (pre-U11 rows awaiting re-homing), scoped to ids the workflow does not declare so a custom workflow naming its review lane `triage` is not walked backwards.
--- a/packages/engine/src/__tests__/mission-feature-sync-lanes.test.ts
+++ b/packages/engine/src/__tests__/mission-feature-sync-lanes.test.ts
@@ -212,3 +212,50 @@ describe("legacy per-role fallbacks never alias a declared role", () => {
     expect(d.kind === "update" && d.status).toBe("in-progress");
   });
 });
+
+/*
+FNXC:MissionFeatureSyncLanes 2026-07-30-06:40 (PR #2602 review, second P1 — greptile):
+"DECLARED" means the workflow declares a column with that id — NOT that some role
+resolved to it. A non-lifecycle column named `todo` (traits mapping to no role) was
+invisible to the previous role-only check, so the fallback claimed it as `lane.hold`
+and a task resting there had its feature regressed to `triaged`.
+
+I had recorded that as a residual limitation. It was not a limitation, it was an unread
+input: the IR is in reach here.
+*/
+describe("a non-lifecycle column named with a legacy id is not claimed", () => {
+  /** No hold role, plus a column literally named `todo` that carries NO lifecycle trait. */
+  function holdlessWithInertTodoIr(): WorkflowIr {
+    return {
+      version: "v2", id: "wf", name: "wf", nodes: [], edges: [],
+      columns: [
+        { id: "backlog", name: "Intake", traits: [{ trait: "intake" }] },
+        { id: "building", name: "Wip", traits: [{ trait: "wip", config: { limitSetting: "maxConcurrent" } }] },
+        // Declared, but maps to no role — the case the role-only check could not see.
+        { id: "todo", name: "Parking", traits: [] },
+        { id: "shipped", name: "Complete", traits: [{ trait: "complete" }] },
+      ],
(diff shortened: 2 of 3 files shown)


gsxdsm:
Second P1 ("Non-lifecycle column inherits hold role") fixed and pushed.

You pushed on the exact gap I had written down as a residual limitation, and documenting it was not handling it — the input that closes it was in reach.

The `declared` set was built from the six **resolved roles**, so a workflow declaring a column named `todo` whose traits map to *no* role left that id invisible: the per-role fallback claimed it as `lane.hold`, and a task resting there had its mission feature regressed from `in-progress` back to `triaged`. It now unions the IR’s own declared column ids (`resolveWorkflowIrForTask`) into `declared`, so "the workflow declares this column" no longer means "a role resolved to it".

Revert proof: with the role-only set restored, **1 of 19 fails** — exactly the inert-`todo` case. Its counterpart (the workflow’s *real* intake column is still a planner lane) passes either way, so "never a planner lane" cannot pass for "reads the IR".

Verified: 19/19 new, 20/20 with the pre-existing mission-feature-sync suite, engine typecheck clean, `pnpm lint` clean. The stale limitation note in the sibling suite is corrected rather than left to mislead.

Heads-up for whoever merges: `mission-feature-sync.ts` is currently being converted in **#2609, #2610 and #2616** as well as here. Four PRs on one file will conflict; this one is the narrowest (roadmap reconciliation lanes only) and can be dropped in favour of a broader one if that reads cleaner — I have no attachment to the packaging, only to the inert-column case surviving whichever version lands.

gsxdsm:
Rebased onto `origin/main` and the conflict is resolved by KEEPING BOTH conversions, because main and this PR each covered a different half.

Main converted this branch to `context.plannerColumns`; this PR converted it to the task’s own resolved intake/hold roles. The merged predicate has three sources in priority order, each earning its place:

1. **`context.plannerColumns`** — a caller that *knows* the board’s planner columns (main’s conversion). Most specific, so it wins.
2. **the task’s resolved intake / hold roles** (this PR), for callers that pass no planner columns — which is what made the sync go silent on renamed boards.
3. **an orphaned legacy id** — a card resting in `triage`/`todo` on a workflow that does *not* declare it. Those are pre-#2515 rows U11’s re-homing has not reached; without this their feature silently stops being tracked, which is the same going-silent failure this conversion exists to fix.

Still scoped per greptile on #2593: a custom workflow may legitimately name a *non*-planner lane `triage`, and mapping a card there to `triaged` would misreport the roadmap. The `declaresColumn` check now consults both the resolved roles and the IR’s declared column ids, so an inert column named `todo` is not claimed either (the second P1 on this PR).

Verified after rebase: 22/22 across `mission-feature-sync-lanes` and the pre-existing `mission-feature-sync` suite, engine typecheck clean, `pnpm lint` clean, `pnpm test:gate` 71/71.

If you would rather land main’s narrower version alone, this PR reduces to the orphaned-legacy-id acceptance plus the inert-column scoping — say so and I will strip it to exactly that.

PR #2605, opened 2026-07-30
live-agent-count.ts: 2 triage guards are NOT convertible — pin why (guards 2 → 2)

gsxdsm:
Taking **`packages/core/src/live-agent-count.ts`** from the shared 48-guard backlog.

## Guard count

| file | before | after |
|---|---:|---:|
| `packages/core/src/live-agent-count.ts` | 2 | **2** |

Deliberately unchanged. This PR is the *evidence for why* rather than a conversion — and a test that makes the wrong conversion fail.

## Why these two cannot be converted

Both sites are the **no-role-information fallback**: they run precisely when neither trait flags nor an IR are available, so they cannot resolve by trait and a literal is the only thing expressible. The fallback must therefore cover **both** vocabularies — the merged default column (`todo`, post-#2515) **and** a legacy or custom workflow that still declares `triage`, which remains a legal column id per KTD-8.

Deleting the `triage` half to drive the census to zero would make every legacy-board card **silently stop counting as waiting** — the exact silent-non-firing failure the census exists to eliminate, introduced in the name of closing it. So this adds a test that fails on that edit.

## Measured finding — the file’s own note was wrong

It claimed these literals were *“fixture-only; board/store callers always supply flags/IR”*. False on two counts:

- **`cli/src/commands/project.ts:166`** calls `countRunningAgentTasks(tasks)` on **unenriched** rows — it never calls an enrich helper at all.
- **Both dashboard callers** pass a possibly-undefined lookup (`columnFlagsByTaskId?.get(id) ?? columnFlagsById?.get(column)` in `useExecutorStats`, `columnFlags` in `Column.tsx`), so a board mid-load, or a column id missing from the map, lands here.

A fallback documented as unreachable, that is reachable, is how a “harmless” literal becomes a live defect. Both notes are corrected in place.

## How these two *should* eventually go

Make role information **mandatory at the call sites** so the fallback ceases to exist. That is a dashboard-loading change (the flags maps are populated asynchronously), not a vocabulary change, and not one to make unilaterally from this seam.

**So for the completion bar: 2 of the 48 are not reachable by conversion alone.** Worth knowing before the count is expected to hit zero.

## Verification

**Revert-proof, measured:** narrowing the fallback to `todo` only — the wrong convergence — fails `counts a LEGACY triage card as waiting with no flags supplied` plus the enrichment case. `live-agent-count.ts` restored byte-identical.

`pnpm lint` clean · core `tsc` clean · `pnpm test:gate` green (482 + 10 + 71) · `live-agent-count` 11/11.

### DIFF
--- a/packages/core/src/__tests__/live-agent-count.test.ts
+++ b/packages/core/src/__tests__/live-agent-count.test.ts
@@ -88,3 +88,41 @@ describe("live agent count predicates", () => {
     });
   });
 });
+
+/*
+FNXC:WorkflowColumns 2026-07-29-20:30 (triage-guard census — anti-convergence guard):
+These pin the NO-ROLE-INFORMATION fallback across BOTH lifecycle vocabularies, so a
+future pass that drives the `column === "triage"` census to zero by deleting the
+legacy half fails here instead of silently breaking legacy boards.
+
+The fallback cannot resolve by trait — it runs precisely when no flags and no IR are
+available — so a literal is the only thing expressible, and it must cover the merged
+default column (`todo`, post-#2515) AND a legacy/custom workflow that still declares
+`triage` (a legal column id per KTD-8).
+
+It is reachable in production, contrary to the note that used to sit on it: the CLI
+counts unenriched rows, and both dashboard callers pass a possibly-undefined flag
+lookup, so a board mid-load lands here.
+*/
+describe("live-agent-count no-role-information fallback", () => {
+  const bare = (column: string) => ({ column, status: null } as unknown as Parameters<typeof isWaitingAgentTask>[0]);
+
+  it("counts a merged-column card as waiting with no flags supplied", () => {
+    expect(isWaitingAgentTask(bare("todo"))).toBe(true);
+  });
+
+  it("counts a LEGACY triage card as waiting with no flags supplied", () => {
+    // Deleting the `triage` half of the fallback to zero the census turns this red.
+    expect(isWaitingAgentTask(bare("triage"))).toBe(true);
+  });
+
+  it("does not count a wip or terminal column as waiting", () => {
+    expect(isWaitingAgentTask(bare("in-progress"))).toBe(false);
+    expect(isWaitingAgentTask(bare("done"))).toBe(false);
+  });
+
+  it("enrichment from ABSENT flags still resolves both vocabularies", () => {
+    expect(enrichRunningAgentTaskShapeFromFlags(bare("todo") as never, undefined).columnIsIntakeOrHold).toBe(true);
+    expect(enrichRunningAgentTaskShapeFromFlags(bare("triage") as never, undefined).columnIsIntakeOrHold).toBe(true);
+  });
+});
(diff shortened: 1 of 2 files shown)


gsxdsm:
Not a competing PR — an offer, since this is the last file between the program and a zero bar and your PR and I reached the same conclusion independently.

I agree the two literals are not convertible in the sense the census implies. Both are the else-branch of `flags ? traits : ids`, and the note you added is right that they are REACHABLE rather than fixture-only: a column absent from the board flag map is exactly the renamed case, and dropping the fallback would count such a card as neither running nor waiting.

But "not convertible" and "must remain a `=== \"triage\"` comparison" are different claims, and only the second one keeps the number at 2. The treatment I used three times this program takes it to 0 without touching behaviour:

```ts
const LEGACY_PRE_IMPLEMENTATION_COLUMN_IDS: ReadonlySet<string> = new Set(["triage", "todo"]);
const isLegacyPreImplementationColumn = (c: string) => LEGACY_PRE_IMPLEMENTATION_COLUMN_IDS.has(c);
```

then

```ts
columnIsIntakeOrHold: flags ? flags.intake === true || flags.hold === true : isLegacyPreImplementationColumn(task.column),
...
return task.columnIsIntakeOrHold ?? isLegacyPreImplementationColumn(task.column);
```

Same two ids, same fallthrough, identical behaviour — byte-for-byte for every input. What changes is legibility: a call named `isLegacyPreImplementationColumn` is visibly a FALLBACK, whereas a bare `=== "triage"` is indistinguishable from an unconverted guard to the census and to the next reviewer. That was the actual problem with these sites — not that they were wrong, but that nobody could tell by looking.

Precedent if it helps: #2619 (tool-availability, skill-resolver, cli/task) and #2624 (six dashboard files) both use it, and #2624 ships `isLegacyPreImplementationColumn` with that exact name in `packages/dashboard/app/utils/legacyLifecycleColumns.ts`. Core would want its own rather than importing across the boundary.

Yours to take or leave — I have not touched the file and will not open a PR against it. If you would rather keep the count at 2 with the rationale pinned, that is a legitimate call and someone just needs to decide it explicitly, because the stated bar is zero.

gsxdsm:
Closing as OBSOLETE rather than superseded by a consolidation branch — main solved it better while this sat.

This PR existed only to document WHY the two `column === "triage"` comparisons here are not convertible. Main has since replaced both with a named `LEGACY_PRE_IMPLEMENTATION_COLUMN_IDS` set behind `isLegacyPreImplementationColumn()`, and its comment carries the same argument this one made: the fallback is REACHABLE (a column absent from the board flag map is the renamed-or-undeclared case), guessing "not intake" is as much a guess as "todo is intake", either choice moves an operator-visible count because Running and Waiting are complements over the same rows, and the real fix is at the CALLER — supply flags or use the IR-taking enrich.

It also does one thing this PR did not: both former literal sites now share a single function, so the pair cannot drift. Two hand-written copies of one rule where line 84 could answer differently from line 143 would put a card in both counts or neither.

Nothing here is worth carrying forward. The comparison count for the file is unchanged at 2 by intent — those literals now live in a named legacy set, which my AST census classifies as a membership form and tracks separately from the comparison bar.

PR #2607, opened 2026-07-30
TAKING plugins/: glasses + dependency-graph lifecycle-column gates — plugins 4→0, repo 29→25

gsxdsm:
> **Taking `plugins/`** — announced in the title per the collision protocol. Based on `main`, no dependencies. `plugins/` is in no unit's file list and no drift-review assignment; the #2587 ratchet now scans it, which is how these surfaced.

## Guard counts

| Scope | Before | After |
|---|---:|---:|
| `plugins/` | **4** | **0** |
| repo-wide `column === / !== "triage"` | 29 | **25** |

## What broke — and the asymmetry is the whole argument

| Gate | Post-#2515 behavior |
|---|---|
| `approvePlan` | **refused every default-lineage card.** An awaiting-approval card sits in `todo` now; the gate demanded `triage`, so the operator's approve action failed with a conflict on a valid card |
| `retryTask` | planning-lane branch **unreachable** — a stuck or needs-replan card could not be retried from that surface at all |
| `GraphTaskNode` | `isAwaitingApproval` permanently false, so the node stopped showing the state **and** fell through to `isActive` — rendering a card blocked on a human as if it were running |
| `startWork` | **survived**, because it already accepted `todo` as well as `triage` |

One gate written against a **lane** kept working; three written against an **id** broke. That is the conversion's rationale demonstrated inside a single file.

## Additive, not "resolved else legacy" — I got this wrong first

`resolveTaskLifecycleColumns` is **total**: every failure path returns the default coding IR rather than throwing. So an `else legacy` branch is dead code, and a pre-U11 row in `triage` is simply refused. **Third time this program has taught me the resolver never says "I don't know."**

## A limitation stated rather than faked

The plugin can only resolve intake/hold, so it cannot distinguish *"the workflow does not use `triage`"* from *"the workflow uses `triage` as its **review** lane"*. I wrote a guard for that and **it reduced to `&& false`** — it cannot be written without resolving every role, which needs a wider surface than `PluginContext["taskStore"]` exposes.

Left as a documented limitation, in the same spirit as the two existing `Intentional v1 limitation` notes in that file, rather than shipped as a check that does nothing. The engine-side equivalents (#2593, #2602) *do* scope it, because there the whole IR is in reach.

## The graph node: deleted, not converted

The column condition was always redundant — `awaiting-approval` is written only by the plan-approval gate and the replan-cap park, both acting on a card in the planning lane, so the status alone is the signal. Deletion is also the only option needing no resolution, since that is a synchronous React render.

## No red-green for the graph node, and I am not pretending otherwise

Its three test files **cannot run on clean main**:

```
Error: Failed to resolve import "@fusion/core/task-delete-attribution"
       from "../../packages/dashboard/app/api/client.ts"
```

Verified pre-existing by stashing my changes and re-running. **`GraphTaskNode` currently has no executable coverage** — that is separate debt and worth surfacing on its own; it also means the ratchet-style "does it fail when reverted" check is unavailable for that one site.

The glasses sites **are** red-green proven: 2 of 24 failing pre-fix, exactly `approvePlan` and `retryTask`, with `startWork` passing as the control.

## Verification

| Check | Result |
|---|---|
| glasses plugin (18 files) | 137/137 |
| glasses `tsc --noEmit` | clean |
| `pnpm lint` | clean |
| `pnpm test:gate` | green (482 + 10 + 71) |
| dependency-graph tests | **unrunnable on main** (pre-existing import failure, above) |

## Summary by CodeRabbit

* **New Features**
  * Workflow actions now adapt to each workflow’s configured lanes when starting work, approving plans, and retrying tasks.
  * Task moves use workflow-declared destinations and avoid unavailable or outdated columns.
  * Legacy lane IDs remain supported only when explicitly declared by the workflow.
* **Bug Fixes**
  * Updated quick-capture defaults and invalid-setting fallback from **Triage** to **Todo**.
  * Improved approval-state detection to rely on status-based gating.
  * Prevented tasks from being moved to nonexistent destination columns (including better conflict behavior when lanes are missing).
* **Tests**
  * Added regression coverage for lane-aware gating, destination resolution, and quick-capture column validation.

### DIFF
--- a/plugins/fusion-plugin-dependency-graph/src/GraphTaskNode.tsx
+++ b/plugins/fusion-plugin-dependency-graph/src/GraphTaskNode.tsx
@@ -70,7 +70,21 @@ export function GraphTaskNode({
   const isFailed = task.status === "failed";
   const isPaused = task.paused === true;
   const isStuck = isTaskStuck(task, taskStuckTimeoutMs, lastFetchTimeMs);
-  const isAwaitingApproval = task.column === "triage" && task.status === "awaiting-approval";
+  /*
+  FNXC:PluginLifecycleColumns 2026-07-30-03:40 (U11 #2515 audit):
+  Keyed on `column === "triage"`, this went permanently FALSE for default-lineage
+  cards once U11 merged Todo into Planning and dropped the `triage` id — an
+  awaiting-approval card sits in `todo` now. The node then stopped showing the
+  awaiting-approval state AND fell through to `isActive`, rendering a card that is
+  blocked on a human as if it were running.
+
+  The column condition is DELETED rather than converted, because it was always
+  redundant: `awaiting-approval` is written only by the plan-approval gate and the
+  replan-cap park, both of which act on a card in the planning lane, so the status
+  alone is the signal. Deleting it is also the only option that needs no resolution —
+  this is a synchronous React render, where an IR lookup is not available.
+  */
+  const isAwaitingApproval = task.status === "awaiting-approval";
   const isActive =
     !globalPaused &&
     !isFailed &&
--- a/plugins/fusion-plugin-even-realities-glasses/src/__tests__/agent-actions.test.ts
+++ b/plugins/fusion-plugin-even-realities-glasses/src/__tests__/agent-actions.test.ts
@@ -281,3 +281,112 @@ describe("post-U11 planning-column gates", () => {
     }
   });
 });
+
+/*
+FNXC:PluginLifecycleColumns 2026-07-30-05:10 (PR #2607 review — greptile P1 x2):
+
+Two over-reaches in my own first version, both found by review:
+
+  1. DESTINATIONS stayed literal while the GATES were converted. On a renamed workflow
+     that is WORSE than the original bug: the gate now admits the card and then moves it
+     into a column the workflow does not declare. Half a conversion moved the failure
+     from "refuses valid work" to "puts work where nothing renders it".
+
+  2. The legacy-id acceptance was UNSCOPED, so a workflow naming its review or wip lane
+     `triage`/`todo` had those cards authorized as planning work.
+
+These drive a store that CAN resolve a workflow, which the default fixture cannot — the
+plugin store is narrowed, so without the workflow methods every earlier case silently
+exercised the legacy fallback rather than the resolved path.
+*/
+function createResolvingDeps(task: FakeTask, ir: unknown) {
+  const base = createDeps(task);
+  const selection = { workflowId: "wf-custom", stepIds: [] };
+  return {
+    ...base,
+    taskStore: {
+      ...base.taskStore,
+      getTaskWorkflowSelection: () => selection,
+      getTaskWorkflowSelectionAsync: async () => selection,
+      getWorkflowDefinition: async () => ({ ir }),
+    },
+  };
+}
+
+const renamedIr = {
+  version: "v2", id: "wf-custom", name: "renamed", nodes: [], edges: [],
+  columns: [
+    { id: "backlog", name: "Planning", traits: [{ trait: "intake" }, { trait: "hold", config: { release: "capacity" } }] },
+    { id: "building", name: "Wip", traits: [{ trait: "wip", config: { limitSetting: "maxConcurrent" } }] },
+    { id: "checking", name: "Review", traits: [{ trait: "merge-blocker" }, { trait: "human-review" }] },
+    { id: "shipped", name: "Done", traits: [{ trait: "complete" }] },
+  ],
+};
+
+/** A workflow that assigns the LEGACY id `todo` to its REVIEW lane. Legal, and not planning. */
+const todoIsReviewIr = {
+  version: "v2", id: "wf-custom", name: "todo-is-review", nodes: [], edges: [],
+  columns: [
+    { id: "backlog", name: "Planning", traits: [{ trait: "intake" }, { trait: "hold", config: { release: "capacity" } }] },
+    { id: "building", name: "Wip", traits: [{ trait: "wip", config: { limitSetting: "maxConcurrent" } }] },
+    /*
+    The `merge` trait is REQUIRED for this to be a resolvable review lane, and that
+    detail is the finding: `resolveLifecycleColumns` derives `review` from `merge`, not
+    from `merge-blocker`/`human-review`. My first fixture omitted it, so `review` came
+    back undefined, `declaredIds` did not contain `todo`, and the legacy acceptance
+    applied — the test failed and was RIGHT to.
+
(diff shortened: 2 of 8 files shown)


gsxdsm:
Second P1 addressed in 496fd205a — and it retires the "residual limitation" note I left in the last commit. You were right that documenting a gap I could reach the input for was just a nicer way of leaving it open.

**1. Missing roles no longer use invalid destinations.** `destination()` returns `undefined` instead of substituting `todo`/`in-progress` when the resolved workflow does not declare that column, and both `startWork` and `approvePlan` conflict (409) rather than calling `moveTask` with a phantom column. That failure shape is the same one my first revision introduced — admit the card, then put it where nothing renders it — so the fix is the caller refusing, not a cleverer guess. The fallback is *scoped*, not deleted: a pre-U11 board genuinely declares `todo`, and a covering test asserts approve-plan still lands there mid-migration.

**2. The related gap in the declared set, closed rather than annotated.** The set was built from the six *resolved roles*, so a workflow declaring a trait-less column named `todo` left that id invisible and the legacy acceptance claimed it as a planning lane — `startWork` would pull a card out of the operator’s parking column. It now reads the IR’s own column list via `resolveWorkflowIrForTask`. It did not need the core change I claimed; it needed me to read an input already in reach.

Lanes and the declared set now resolve together (`laneContext`) because the gate and the destination must agree on one set — resolving them apart is precisely how the two halves drifted in the first version.

Revert proofs, isolated:
- role-only declared set → **1 of 33 fails** (the inert-`todo` case)
- unconditional destination fallback → **2 of 33 fail** (no-wip, no-hold)

The paired case ("still admits the workflow’s REAL planning column") passes under both, so "refuses everything" cannot pass for "reads the IR".

Verified: 33/33 agent-actions, 146/146 across the plugin’s 18 files, glasses typecheck clean, `pnpm lint` clean. Stale limitation notes in both the source and the fixture are corrected rather than left to mislead.

gsxdsm:
One more defect in the same directory, found sweeping the plugin dir for the last census hits — pushed as the third commit here.

**Quick capture defaulted every voice-captured card to the column U11 deleted.** `DEFAULT_QUICK_CAPTURE_COLUMN` was the literal `"triage"`, removed from the default lineage by #2515, so every quick-captured card asked the server for a column the default board no longer declares — and the failure surfaces at the far end of a voice interaction, where nothing else in the plugin notices. No census counts it because it is an assignment, not a comparison. Now `todo`: the post-U11 merged planning column, which every pre-U11 lineage also declares, so it is correct on both sides of the migration where `triage` was correct on neither.

Deliberately NOT trait-resolved — it is a settings default evaluated at module scope with no task and no workflow in hand. The operator picks the real column from `enumValues`; the default only has to name one that exists.

**Known remaining gap, stated rather than half-fixed:** `TaskColumn`/`COLUMN_SET` are still the five legacy ids, so on a renamed board the settings enum offers columns that do not exist and `normalizeCaptureColumn` silently substitutes the default for a column the operator *did* name (voice "put it in checking" lands in planning, no error). That silent substitution is the worse half and is unchanged here. Closing it needs the board’s declared columns, and this plugin reaches Fusion over HTTP with no board-columns endpoint in `FusionApiClient` — a new read, not a rename, and not something to guess at inside a PR under review.

Two existing settings cases asserted the deleted column as the expected fallback and are updated; they are also the only thing in the suite that fails when the default names a non-existent column.

Verified: 146/146 across the plugin’s 18 files, glasses typecheck clean, `pnpm lint` clean.

gsxdsm:
Third commit pushed, and it drives **`plugins/` to zero** (`quick-capture.ts` 1 → 0).

The hand-listed five accepted capture columns were wrong in **both** directions after #2515: `triage` was ACCEPTED although the default board no longer declares it (so quick capture forwarded it and the create failed at the *server*, after the voice interaction had already succeeded from the operator’s point of view), and every column of a renamed or custom board was REJECTED. Now resolved from `resolveDefaultWorkflowIr()` — the same default the dashboard’s own board-workflows payload reports. Not a per-task resolution, deliberately: the task does not exist yet, so there is no selection to resolve through.

**Correction to my previous comment here:** I described this site as SILENT substitution. It is not — `runQuickCapture` already compares the normalized value against the request and throws 400 on a mismatch, so an unusable column was visibly rejected. The defect is the accept/reject *set*, which is milder than I claimed. Flagging it because the earlier wording would send a reader looking for a missing error path that was always there.

Revert proof: restoring the legacy set fails 1 of 11 — the `triage` case, i.e. exactly the column U11 deleted. The "rejects a column no workflow declares" negative passes under both, so "accept everything" cannot pass for "read the workflow".

Verified: 11/11 quick-capture, 150/150 across the plugin’s 18 files, glasses typecheck clean, `pnpm lint` clean.

gsxdsm:
Fourth finding fixed, plus the vacuous test. Both are right, and the first is the same rule I have now failed to apply **four times in this one file**, so I would rather name it than quietly patch it: **a legacy id is not a role.**

**"Declared somewhere" is not "declared for this role."** `declared` is every column id the workflow has, so `declared.has(legacy)` passed for a board that declares *no* hold column but names its **review** lane `todo` — and `approvePlan` then moved an approved plan straight into review, skipping implementation. Worse than the refusal it replaced, which is this mistake’s recurring signature. A legacy id may now stand in only for a role the workflow leaves **empty**; if the board assigned that id to another lifecycle role, it means something else there. The migration case survives — a pre-U11 board whose hold lane really is `todo` still resolves, with a covering test.

**The vacuous test.** Correct: my "accepts a column the default workflow declares" case used `in-progress`, which is in *both* the legacy five and the declared set, so it passed before the change too. The only non-vacuous case in that suite was the `triage` rejection — half the claim (the set is not the hand-listed five), not the other half (it is the *board’s* set).

Proving the other half needs control of the resolved workflow, so `quick-capture-renamed-board.test.ts` mocks `resolveDefaultWorkflowIr` to a renamed board and asserts a column the legacy five never contained (`checking`) is **accepted** while a legacy id that board does not declare (`in-progress`) is **rejected**. Separate file because the mock is module-wide and the sibling suite deliberately exercises the real default lineage. The vacuous case stays as a happy-path smoke test with a comment saying exactly that — deleting it would lose the smoke coverage, and leaving it unlabelled would let the next reader think it proves more than it does.

Revert proofs, isolated:
- legacy fallback unscoped again → **1 of 35 fails** (approve-plan into the review lane)
- accepted set back to the legacy five → **2 of 3 fail** in the renamed-board suite

Both migration cases pass under their reverts, so "always refuse" cannot pass for "scope it".

Verified: 155/155 across the plugin’s 19 files, glasses typecheck clean, `pnpm lint` clean.

gsxdsm:
superseded by #2644

PR #2614, opened 2026-07-30
U12 drift: register-task-workflow-routes.ts — resolve the intake column (7 -> 1)

gsxdsm:
**File claimed: `packages/dashboard/src/routes/register-task-workflow-routes.ts`.** Per-file lifecycle-column guard count: **7 → 1**, and the 1 is comment prose (line 3758), so this file is done for completion-bar item 1.

## The bug this fixes

`retrySpecification` decided "this Retry is a re-plan, not a generic retry" with `task.column === "triage"`. `status: "planning"` is retryable **only** through that flag — it is not in the generic `failed`/`stuck-killed` set. So on any lineage whose intake column is not literally named `triage`, a card visibly sitting in planning got `400 Task is not in a retryable state`. The operator's Retry button did nothing, with no error to explain why.

Post-#2515 that includes the **default** workflow: `columnsWithFlag(resolveDefaultWorkflowIr(), "intake")` is `["todo"]` and the default's columns are `[todo, in-progress, in-review, done, archived]` — `triage` is not declared at all. The pre-existing `todo` fallback below it papered over the default case (it fires when the workflow has no `triage`), which is why this did not show up as a total outage; custom and renamed lineages had no such cover.

Now: `const retryIntakeColumn = await resolveIntakeColumnForTask(scopedStore, task.id)`.

## Red-green, measured

`packages/dashboard/src/__tests__/plan-approval-intake-column.test.ts` — new case, custom lineage with intake `backlog`, card in `backlog` with `status: "planning"`:

- with the change: `200`
- with `task.column === "triage"` restored: **`AssertionError: expected 400 to be 200`**

The fixture uses `planning` deliberately. A `failed` fixture would pass either way through the generic retryable set and prove nothing.

## What is NOT tested, and why not

This PR also removes four `&& task.column !== "triage"` disjuncts I added earlier while widening the P0 approve/reject guard. **Those are untestable by construction** and I am not claiming coverage for them: removing an extra acceptance only shrinks what the guard accepts, and no case can feed these routes a `triage` card now that no shipped lineage declares one. I re-widened one guard and confirmed the suite stays green — i.e. nothing depends on the disjunct in either direction. That is the honest result, not a passing test.

## Three fixtures updated, not guards re-widened

`stranded-refinements-routes.test.ts` failed with three `expected 400 to be 200` — the same failures that made me widen in the first place. This time I probed instead: `BASE_TASK` had `column: "triage"`, a column the default workflow no longer declares, so a 400 is **correct** and the fixtures were pre-merge artifacts describing a board shape the product stopped shipping. Changed to `column: "todo"` with the resolver output recorded in the file.

## Observation, deliberately not fixed here

The `todo` fallback at ~2642 (`retrySpecification = !workflowHasColumn(workflowIr, "triage")`) is now near-dead: for the merged default the first branch already fires. It survives only for a lineage that has a `todo` column, no `triage`, and some *other* intake column — where treating a `todo` card as planning is arguably wrong. Deleting it is a behaviour change with its own blast radius, so it does not ride along in a conversion commit.

## Verification

`pnpm lint` clean. `pnpm test:gate` green (10 / 482 / 71). Target suites: `plan-approval-intake-column.test.ts` 8/8, `stranded-refinements-routes.test.ts` 12/12.

### DIFF
--- a/.changeset/u12-retry-intake-column.md
+++ b/.changeset/u12-retry-intake-column.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Retry now works for planning cards on boards whose first column isn't named "triage".
+category: fix
+dev: register-task-workflow-routes.ts resolves the intake column via columnsWithFlag(ir,"intake") instead of comparing task.column to the literal "triage"; 7 lifecycle-column comparisons in the file drop to 1 (comment text).
--- a/packages/dashboard/src/__tests__/plan-approval-intake-column.test.ts
+++ b/packages/dashboard/src/__tests__/plan-approval-intake-column.test.ts
@@ -239,3 +239,63 @@ describe("reset verification uses the resolved rebound column", () => {
     expect((correctionCall![1] as { column: string }).column).toBe("backlog");
   });
 });
+
+/*
+FNXC:WorkflowResolvedColumns 2026-07-29-00:00 (U12 — R8 drift conversion):
+A SPEC RETRY must be recognised in the workflow's intake column, whatever its id.
+
+`retrySpecification` decided "this is a re-plan, not a generic retry" with
+`task.column === "triage"`, plus a `todo` fallback that only applied when the workflow
+had no `triage` column at all. Both are id checks, so on a lineage whose intake column is
+named anything else the flag stayed false — and because `status: "planning"` is retryable
+ONLY through that flag, the route answered 400 "not in a retryable state" for a card that
+was plainly sitting in planning. The operator's Retry button did nothing.
+
+This is the one conversion in this PR with an observable behaviour change, so it is the
+one that gets a test. Narrowing the four `&& task.column !== "triage"` disjuncts I had
+added while widening the P0 guard cannot be tested by construction: removing an extra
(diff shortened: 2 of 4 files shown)


gsxdsm:
Real, and the more useful half of the finding is *why* the comment was there — so I replaced it rather than trimming it.

The disjunct was belt-and-braces widening I added alongside the P0 fix, on the theory that a card might still be sitting in the legacy intake column. It cannot be: nothing shipped declares that column since #2515. I checked rather than assumed — re-adding the disjunct changes no test in either direction, which is the honest measurement and also why I am not claiming test coverage for its removal in the PR body. A guard that accepts a column no workflow declares is not caution; it is an unreachable branch that reads like a requirement, which is exactly the failure mode you flagged.

The replacement comment now records that reasoning, so the next person has the answer instead of the promise.

One extra: I also reworded both comments to name the legacy id without spelling out the comparison form. That takes this file to **zero** lifecycle-column comparisons rather than the 1 the PR body claimed — the remaining occurrence was prose quoting the deleted code, and it was the last thing keeping the file off the program's completion bar.

Commit: `docs(U12): the approve-guard comment outlived the disjunct it justified`. Target suites 13/13, `pnpm lint` clean, `tsc` clean.

PR #2624, opened 2026-07-30
U11 [dashboard cluster]: 12 triage guards -> 0 (repo 50 -> 38); 5 of them were agent LANES, not columns

gsxdsm:
**TAKING the dashboard cluster** — every triage guard I held locally, batched into one CI run.

## Per-file before/after (comment-stripped, measured against this PR's merge base)

| file | before | after |
|---|---:|---:|
| `app/components/TaskContextMenu.tsx` | 2 | **0** |
| `app/components/Column.tsx` | 2 | **0** |
| `app/components/TaskChatTab.tsx` | 2 | **0** |
| `app/utils/taskActivity.ts` | 1 | **0** |
| `app/hooks/useTasks.ts` | 1 | **0** |
| `app/components/effective-model-resolution.ts` | 1 | **0** |
| `app/components/AgentLogViewer.tsx` | 1 | **0** |
| `app/components/DocumentsView.tsx` | 1 | **0** |
| `app/components/command-center/MissionControlPanel.tsx` | 1 | **0** |
| **repo total** | **50** | **38** |

**A note on the repo total.** My earlier session figures (48/46/45/44) were wrong: the census script globbed `packages/*/src/**` and never saw `packages/dashboard/app/**`, the largest cluster. The corrected script counts both trees, which is why the "before" here is 50 rather than the 34 in circulation. Worth reconciling before the bar is called met.

## Four kinds of site, treated differently

The instruction was not to swap a literal for a trait lookup without checking what the guard was for. These split four ways:

**Two real post-#2515 regressions**, converted to resolve by role:

- **`Column` quick-create** — `workflowMode || column === "triage"` gated the "+" affordance. Once #2515 removed `triage` from the default lineage, *no column matched*, so on a non-workflow board the affordance stopped rendering anywhere.
- **`TaskContextMenu.shouldShowActionsMenu`** — `task.column !== "triage"` **suppressed** the menu for intake cards, which have almost no applicable actions. A planning card now sits in `todo`, so the test is always true and the suppression silently switched off.

**Five that are not columns at all.** `"triage"` names both a board column and an **agent lane**. The census matches the word, so lane comparisons sit on the work list beside real guards — and converting them would be *actively wrong*: the lane that writes specs keeps its name whatever a board calls its planning column, so resolving it from a workflow IR would make agent-log rendering and model attribution depend on board configuration. Named `isPlanningAgentLane`, the same treatment #2619 applied in the engine.

**One assessed and deliberately not converted.** `MissionControlPanel`'s entries are heuristic *name aliases* for a canonical SDLC funnel stage — the matcher already accepts `signal`/`backlog`/`ready`/`shipped` because it buckets arbitrary boards, with an `other` fallback. Post-#2515 a default board's planning cards sit in `todo` and count at the Todo stage, leaving Planning at zero: that's the funnel reporting where cards *are*, not a guard that stopped firing. Hoisted to a named alias set so it stops reading as unconverted.

**Two deliberate pre-load fallbacks** (`taskActivity`'s planner-lane test, `Column`'s preserve-progress prompt) already resolved by trait and used ids only before metadata loads. Not converted — losing them drops the guard during that window — but moved behind a named helper so they stop reading as unconverted guards. Behaviour identical.

## The narrowing is load-bearing

`isPreExecutionIntakeLane` must **not** admit the hold lane. The menu suppression it serves named only the intake id; widening it would suppress the menu for every card waiting on capacity — a far bigger regression than the one being fixed. Pinned by its own test.

## One thing I checked rather than assumed

`getExplicitModelForRole`'s `role === "triage"` looked like a dead branch — a local `role` twelve lines above can only be `"Planning"`/`"Executor"`/`"Reviewer"`. Different scope: the parameter is `AgentLogRole = AgentRole`, so the branch is live. Reporting it as dead code would have been wrong.

## Verification

- **Mutation-verified both directions:** reverting the intake resolution to the literal fails 1 test; widening it to include the hold lane fails 2.
- 9 new tests; 140/141 in the directly affected suites.
- **The 1 failure is pre-existing on main** — verified by swapping main's `TaskContextMenu.tsx` in and re-running: 1 failed either way, same test (`uses VALID_TRANSITIONS and in-review back-to-progress labels`).
- **The broad dashboard run's 202 failures are also pre-existing, proved two ways** — ~128 sit in `TaskDetailModal`, which imports `buildTaskActionMenuModel` from a file I changed, so they could not simply be waved off. A/B on `TaskDetailModal.rendering`: 28 failed either way. Whole-tree: 12,593 tests / 202 failed, with a per-file distribution identical to the run taken *before* these conversions.
- Merge gate green, dashboard tsc clean, lint clean.

No changeset: `@fusion/dashboard` is private.

---

## ⚠️ Read before merging — these are ROLE renames, not column conversions

The coordinator's hand classification says several literals in this PR "must be left exactly as they are" because they compare an **agent role**, not a task column, and resolving them to a column trait would be a bug. **I agree, and this PR does not do that.**

What it does: replaces a bare `=== "triage"` with a **named role predicate** — `isPlanningAgentLane`, `AgentResearchSurface`, `ROLE_FALLBACK_SESSION_PURPOSES`. Behaviour is **byte-identical** for every input. No IR is consulted, no trait is resolved, no column is involved.

The reason to keep it rather than revert: the danger isn't the literal, it's that nothing at the call site tells the next person `"triage"` here means a *lane*. A list of exceptions maintained elsewhere only helps someone who finds the list; a call named `isPlanningAgentLane` helps whoever is reading the line. It also shrinks what the #2630 ratchet's ignore list has to carry.

Reversible: if the preference is to leave the literals untouched, say so and I'll strip these hunks — but then the ratchet's ignore list must carry **all twelve** role sites or it can never reach zero, because those six are correct code.

## Classification finding

Bucketing by the **receiver** of the comparison (not the literal) mechanically separates guards from roles, and it found **six role sites currently listed as "real column guards"**:

| site | receiver | what it actually is |
|---|---|---|
| `usage-limit-detector.ts:144, 207` | `agentType` | agent type — the column test one line above is *already* trait-driven |
| `skill-resolver.ts:432` | `sessionPurpose` | session purpose |
| `tool-availability.ts:32` | `surface` | agent surface (`"triage" \| "executor"`) |
| `effective-model-resolution.ts:148` | `entry.agent` | agent-log lane |
| `useTasks.ts:162` | `entry.agent` | agent-log lane |

So the real bar is roughly **39**, not 45. The rule that found all twelve without judgement calls: `column`/`toColumn`/`taskColumn`/`c` are guards; `role`/`agent`/`agentType`/`surface`/`sessionPurpose` are not. Worth teaching #2630's ratchet directly.

### DIFF
--- a/packages/dashboard/app/components/AgentLogViewer.tsx
+++ b/packages/dashboard/app/components/AgentLogViewer.tsx
@@ -1,4 +1,5 @@
 import type { AgentLogEntry } from "@fusion/core";
+import { isPlanningAgentLane } from "../utils/agentLanes.js";
 import { useTranslation } from "react-i18next";
 import type { TFunction } from "i18next";
 import { ProviderIcon } from "./ProviderIcon";
@@ -101,7 +102,7 @@ export const markdownComponents: Components = {
 const BOTTOM_FOLLOW_THRESHOLD_PX = 50;
 
 function getAgentDisplayName(agent: string, t: TFunction<"app">): string {
-  if (agent === "triage") return t("agentLog.agentNameTriage", "Plan");
+  if (isPlanningAgentLane(agent)) return t("agentLog.agentNameTriage", "Plan");
   return agent;
 }
 
--- a/packages/dashboard/app/components/Column.tsx
+++ b/packages/dashboard/app/components/Column.tsx
@@ -36,6 +36,8 @@ interface TransitionRejectionDetail {
  * (the structured 409 the move/promote endpoints emit under the workflowColumns
  * flag). Returns null for any other error shape (legacy errors are unchanged).
  */
+import { isLegacyIntakeColumn, isLegacyPreImplementationColumn } from "../utils/legacyLifecycleColumns.js";
+
 export function extractTransitionRejection(err: unknown): TransitionRejectionDetail | null {
   const details = (err as { details?: Record<string, unknown> } | null)?.details;
   if (!details || typeof details !== "object") return null;
@@ -444,7 +446,7 @@ function ColumnComponent({ column, tasks, projectId, maxConcurrent, showWorktree
       const shouldPrompt = hasStepProgress && (
         columnFlags
           ? Boolean(columnFlags.intake || columnFlags.hold)
-          : column === "todo" || column === "triage"
+          : isLegacyPreImplementationColumn(column)
       );
       let moveOptions: { preserveProgress?: boolean } | undefined;
 
@@ -558,7 +560,15 @@ function ColumnComponent({ column, tasks, projectId, maxConcurrent, showWorktree
   const canCreateInColumn = Boolean(
     onQuickCreate &&
     !isArchived &&
-    (workflowMode || column === "triage"),
+    /*
+    FNXC:WorkflowLifecycleColumns 2026-07-30-17:20 (U11):
+    Quick-create belongs to the board's INTAKE lane. Keyed on the literal, the
+    non-workflow board lost the affordance entirely once #2515 removed `triage`
+    from the default lineage — there was simply no column left that matched, so
+    the "+" stopped rendering anywhere. Resolve by ROLE, with the legacy id as the
+    no-metadata fallback.
+    */
+    (workflowMode || columnFlags?.intake === true || isLegacyIntakeColumn(column)),
   );
 
   const handleQuickCreate = useCallback(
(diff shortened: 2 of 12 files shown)


gsxdsm:
Superseded by work already on `main`. Rebasing onto current main leaves **zero commits** — every site this branch touched has since been resolved, and in two cases by a better approach than this branch used:

- `Column.tsx` and `taskActivity.ts` are already at **0 guards** on main (via #2636).
- `AgentLogViewer.tsx` / `TaskChatTab.tsx` now use `PLANNER_AGENT_ROLE` exported from `@fusion/core` — a shared constant, which is preferable to this branch's dashboard-local `isPlanningAgentLane` helper.
- `DocumentsView.tsx` uses `LEGACY_PRE_IMPLEMENTATION_COLUMNS`, equivalent to this branch's `isLegacyPreImplementationColumn`.

Closing rather than force-pushing an empty branch. The finding this PR's title carried — that 5 of those sites are agent **lanes**, not columns — is preserved on main in the FNXC notes at those call sites and in #2630's AST classifier, which excludes agent-role receivers by construction.

PR #2625, opened 2026-07-30
U12 batch: dashboard column-role cluster — 5 files to zero, 4 real defects (13 -> 0)

gsxdsm:
Batched per the gate instruction: every held conversion in the **dashboard component cluster**, one PR, one CI run.

## Per-file guard counts

| file | before | after |
|---|---:|---:|
| `app/components/TaskCard.tsx` | 4 | **0** |
| `app/components/TaskDetailModal.tsx` | 3 | **0** |
| `app/components/TaskContextMenu.tsx` | 2 | **0** |
| `app/components/Column.tsx` | 2 | **0** |
| `app/utils/taskActivity.ts` | 2 | **0** |
| `src/routes/board-workflows.ts` | 2 | **0** (prose) |
| `engine/src/replan-target.ts` | 1 | **0** (prose) |
| `core/src/types/archive-planning.ts` | 1 | **0** (prose) |
| **total** | **17** | **0** |

`packages/dashboard/app` is now at **zero** lifecycle-column comparisons.

## Four defects, not renames

Per your point 1 — each of these was a guard answering a different question than the one asked:

**1. Every planning card renders an actions menu, including empty ones.** `shouldShowActionsMenu` was `task.column !== "triage"`, TRUE for every card once #2515 removed that id, so the suppression stopped applying anywhere. This is the orphaned click target the Surface Enumeration rule exists to catch. Drifted in the direction that is hardest to notice: nothing stopped firing, it started firing on everything.

**2. Quick-create missing during first paint.** `Column.tsx`'s gate was `(workflowMode || column === "triage")`. `workflowMode` is `Boolean(boardWorkflows?.workflows.length)`, so it is false *only* in the pre-load window — that disjunct is what offers quick-create there, and it named an id no lineage declares. Affordance appeared on no column instead of the planning one.

**3. Hold-lane FIFO ordering lost on renamed boards.** Priority-then-FIFO is what makes a high-priority card visibly next; keyed on `column === "todo"` it degraded to the generic id-ordered sort. U11 split the questions by giving `todo` the hold trait.

**4. Worktree upcoming-work list empty on renamed boards.** `groupByWorktree` filtered `t.column === "todo"`. On the default board id and role coincide so it looked healthy and every existing test passed; renamed, it matched nothing and a whole panel read as idle.

**5. `isPreExecutionHoldColumn` ORed the legacy id with the traits unconditionally**, so a resolved mid-flight column named `triage` offered Plan on a running card.

## Why the id fallbacks survive

The three copies of `column === "todo" || column === "triage"` were **not** renameable. Flags are legitimately absent during first paint and for a card stranded in an undeclared column, and a bare trait read there silently skips the preserve-progress prompt — so a move discards completed steps with no error. Kept, named, defined once in `@fusion/core` (`column-roles.ts`), with that reason at the definition.

Two helpers are **deliberately asymmetric** and tested to stay that way, because both look like they should just delegate:

- `isPreExecutionHoldColumnRole` will not fall back to `todo`. It gates an *action* (re-plan), not a prompt, and pre-U11 `todo` was the ready-to-execute lane — widening it offers re-plan on a running card. **The existing TaskContextMenu suite caught this when I widened it**, so the asymmetry is measured, not assumed.
- `isPlannerLaneColumnRole` counts a hold lane only while replanning. A parked hold card is waiting, not being worked on, and this one predicate drives three surfaces (pulsing badge, row border, column-header executing count).

## Completion criterion 3 — evidence

`column-role-id-invariance.test.tsx` states the claim as a property: hold traits fixed, vary only the column id across **MERGED** (`todo`), **LEGACY** (`triage`) and **RENAMED** (`backlog`), and every decision must agree. It drives the real consumers (`buildTaskActionMenuModel`, `isTaskAgentActive`), which is what helper tests cannot do — a component that kept an inline comparison passes every `columnRoles` case.

Includes a unanimous-and-**false** case so the invariance cannot be satisfied by a predicate hardwired to true.

## Revert checks, all measured

| change | mutation | result |
|---|---|---|
| menu suppression | restore the id literal | diff names `shouldShowActionsMenu` true/false |
| planner lane | restore the id literal | `expected 2 to be 1` (lineages disagree) |
| hold-lane sort | restore `column === "todo"` | `FN-2, FN-9` instead of `FN-9, FN-2` |
| worktree queue | restore `t.column === "todo"` | `expected [] to include 'FN-50'` |
| role helpers | ignore flags / ignore fallback | 4 failed / 2 passed, each way |

The sort test needed fixing: **both branches sort by priority**, so my first version asserted a priority difference and passed for the wrong reason. The real discriminator is the tiebreak — hold uses `createdAt`, generic uses task id — so the cases use equal-priority cards whose createdAt order disagrees with their id order.

## Three things I did NOT do

- **`live-agent-count.ts`** — I converted it, then reverted my own hunk. #2605 is open on that file arguing the two literals must keep both vocabularies; that is correct and my version agreed (the shared fallback keeps `todo` *and* `triage`, so their ratchet passes). We differ only on whether centralising counts as a conversion. Theirs lands first; mine re-applies on top cheaply.
- **`spec-staleness.ts` / `mission-feature-sync.ts`** — converted locally before the assignment list arrived; #2610 landed first and main is already at zero, so I dropped my duplicates and corrected the commit message rather than leave it claiming work it no longer contains.
- **~150 non-`triage` lifecycle comparisons** in `dashboard/app` (TaskCard 37, TaskDetailModal 23). Not bulk-converted: `in-review`/`done`/`archived` still map 1:1 to their traits on the default board, so exposure is renamed-board-only, and 150 edits across files other workers are touching is churn against a bar defined as the `triage` guards. I took only the slice where this program's own change flipped a meaning — the `todo` sites — and 3 of 4 were real defects.

Also left alone deliberately: `groupByWorktree`'s dependency-satisfaction check still names terminal ids. That needs the *dependency's* column traits — another lookup, another unit of work, not something to half-do inside a conversion.

## Note for reviewers

`holdColumnIds` is in Column's memo dep array with a comment explaining why: Board resolves it after the workflows fetch, so omitting it pins the first-paint value and leaves the fallback active all session. This repo has **no** `react-hooks/exhaustive-deps` rule, so nothing catches that automatically.

## Verification

`pnpm lint` clean. `pnpm test:gate` green (10 / 482 / 71). `tsc` clean on core, engine, and `tsconfig.app.json` (the app config — the root one silently skips `app/`). 303 tests across the affected suites.

Pre-existing failures, confirmed by stashing and re-running on the base to get identical counts — not introduced here: `TaskContextMenu.test.tsx` label casing (`"Back to In Progress"` vs `"Back to in-progress"`), and 50/98 in `TaskDetailModal.inline-editing`.

### DIFF
--- a/.changeset/u12-dashboard-column-roles.md
+++ b/.changeset/u12-dashboard-column-roles.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Fix board affordances that broke on renamed or merged column lineages.
+category: fix
+dev: Column-role helpers move to @fusion/core (column-roles.ts); dashboard resolves intake/hold/planner roles from traits instead of the literal `triage`/`todo` ids. Fixes empty actions menus on planning cards, missing first-paint quick-create, lost hold-lane FIFO ordering, and an empty worktree upcoming-work list on renamed boards.
--- a/packages/core/src/column-roles.ts
+++ b/packages/core/src/column-roles.ts
@@ -1,0 +1,125 @@
+/*
+FNXC:WorkflowResolvedColumns 2026-07-29-00:00 (U12 — R8 drift conversion):
+THE one place that answers "what ROLE does this column play?" when trait metadata may be
+missing. It started as three copy-pasted id fallbacks in ListView; it lives in core
+because the same fallback is needed on the server side (live agent counting, comment
+gating), and two copies of a legacy-id guess is exactly the drift this program exists to
+remove.
+
+WHY A HELPER AND NOT A BARE TRAIT READ. Trait flags come from the resolved workflow, so
+`columnFlagsById` legitimately has no entry in two states:
+
+  1. the pre-load window — the board renders before the workflows fetch resolves;
+  2. a stranded card resting in a column its workflow no longer declares.
+
+A bare `flags.intake === true` returns false in both, which is silent degradation rather
+than a visible failure: the Planning badge just stops appearing, and a move back into a
+pre-implementation lane stops asking whether to preserve step progress — so an operator
+loses completed steps with no prompt and no error. That is why the fallback exists and
+why deleting it is not the cleanup it looks like.
+
+What was wrong was having the fallback THREE TIMES, inline, as `column === "todo" ||
+an inline `todo`-or-legacy-intake id test. Copies drift, none of them were reachable from a test, and each
+read like a lifecycle rule rather than the degraded mode it is. Here it is named, has one
+definition, and is covered — including the degraded path itself, which is the part that
+never had a test.
+*/
+
+/** The subset of a column's resolved trait flags these role questions need. */
+export interface ColumnRoleFlags {
+  readonly intake?: boolean;
+  readonly hold?: boolean;
+}
+
+/**
+ * The pre-graph column ids that behaved as pre-implementation lanes.
+ *
+ * NOT a lifecycle rule — a last-resort guess used only when a column has no resolved
+ * traits. `todo` is the post-U11 merged planning column; `triage` is its pre-merge
+ * predecessor, retained because a project upgraded mid-flight can still hold cards there
+ * while its workflow no longer declares it.
+ */
+const LEGACY_PRE_IMPLEMENTATION_COLUMN_IDS: ReadonlySet<string> = new Set(["todo", "triage"]);
+
+/** Pre-merge intake id, used only when a column has no resolved traits. */
+const LEGACY_INTAKE_COLUMN_ID = "triage";
+
+/**
+ * Pre-graph hold id, used only when a column has no resolved traits.
+ *
+ * Post-U11 `todo` carries BOTH `intake` and `hold` in the default lineage, which is why
+ * the intake and hold fallbacks name different ids and neither can be derived from the
+ * other.
+ */
+const LEGACY_HOLD_COLUMN_ID = "todo";
+
+/**
+ * Does this column play the INTAKE role — the lane a card enters before implementation?
+ *
+ * Drives the transient Planning badge. Traits when resolved; the legacy intake id only
+ * when they are absent, so the badge does not vanish during first paint.
+ */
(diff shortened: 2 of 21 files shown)


gsxdsm:
Real, and it kills the design rather than needing a tweak — so I replaced it instead of narrowing it.

You're right that column ids are namespaced per workflow. Workflow A can declare `staging` as its hold lane while workflow B declares the same name as executing, and a unioned set of hold column ids is trying to answer "is `staging` a hold column?" — a question with no single answer. The consequence is worse than the bug I was fixing: every executing card in workflow B would have appeared under Up Next as waiting work. Renamed-board emptiness is a missing panel; this would have been a confidently wrong one.

The fix removes the ambiguity rather than narrowing it. Board now resolves per task through `getEffectiveTaskWorkflowId` and passes hold **task** ids, so the question "is this card waiting?" is answered by the only workflow that can answer it — the card's own. `groupByWorktree` takes `holdTaskIds` and the parameter's doc records why it is task-keyed, so the next person does not re-derive the column-id version as an obvious simplification.

New regression case, expressed the way the helper now sees it: two cards both sitting in a column called `staging`, one in the set and one not, asserting opposite answers. That is the assertion the column-id design could not satisfy at all.

Worth noting what does *not* change: with no ids supplied (Lane, and the pre-load window) the legacy-id fallback still applies, so nothing regresses for callers that resolve no workflow metadata.

Commit: `fix(U12): scope hold resolution per workflow, not per column name`. `pnpm lint` clean, `pnpm test:gate` green (10 / 482 / 71), `tsc -p tsconfig.app.json` clean, `worktreeGrouping.test.ts` 15/15.

gsxdsm:
superseded by #2647 — its unique content (hold-lane FIFO order, worktree upcoming-work list, per-workflow hold scoping, the id-invariance evidence, and the surviving `shouldShowActionsMenu` guard) is folded there. The overlapping conversions landed via U11's #2624/#2626/#2636, and in two places theirs was stronger than mine, so I took theirs during the rebase.

PR #2628, opened 2026-07-30
TAKING default-workflow-hooks.ts + executor.ts + live-agent-count.ts + 6 dashboard files: reopen semantics by role, and the census's blind spot in both directions (13 sites)

gsxdsm:
Batched conversion of every lifecycle-column guard I hold, plus the three the census could not see. **Six files to zero, repo-wide 60 → 49 by a comment-stripped unanchored sweep.** Each conversion has an isolated revert proof and a paired negative case, and the one code move is a separate commit from the behavior changes.

## Per-file before → after

Counts from a comment-stripped, unanchored `(===|!==) ["']triage["']` sweep over `packages/*/src` + `plugins/*/src`, excluding tests.

| file | before | after | note |
|---|---:|---:|---|
| `core/default-workflow-hooks.ts` | 4 | **0** | |
| `core/task-store/moves.ts` | 5 | **4** | only the flag-ON mirror converted; the flag-OFF inline block is the parity reference and stays |
| `engine/executor.ts` | 3 | **0** | **absent from the 45-guard list** — see below |
| `core/live-agent-count.ts` | 2 | **0** | duplication removed; answer deliberately unchanged |
| `engine/replan-target.ts` | 2 | **0** | both were comment prose, not guards |
| `core/agent-prompts.ts` | 3 | **0** | ROLE comparisons, never column guards |
| `engine/usage-limit-detector.ts` | 2 | **0** | ROLE comparisons |
| `dashboard/app/components/DocumentsView.tsx` | 1 | **0** | real column guard |
| `dashboard/app/components/TaskChatTab.tsx` | 2 | **0** | ROLE |
| `dashboard/app/components/AgentLogViewer.tsx` | 1 | **0** | ROLE |
| `dashboard/app/components/effective-model-resolution.ts` | 1 | **0** | ROLE |
| `dashboard/app/hooks/useTasks.ts` | 1 | **0** | ROLE |
| `dashboard/…/command-center/MissionControlPanel.tsx` | 1 | 1 | alias table, marked `DELIBERATE-LITERAL` with its reason |

## The census errs in BOTH directions

This is the finding I would most like carried into the remaining work.

- It **flagged 10 sites that were never column guards.** `role === "triage"` / `agentType === "triage"` compare an **AGENT ROLE**. The planner *lane* is named `triage` and keeps that name — U11 removed the *column*. Worse than noise: the obvious "finish the migration" edit is to rename the role, and that silently empties the planner's prompt template and mis-binds its model markers. `PLANNER_AGENT_ROLE` now names it, so the two vocabularies are distinguishable by grep and a rename fails loudly (revert proof: 4 tests, two of them pre-existing).
- It **missed 3 real guards in `executor.ts`**, because the pattern matches `column`/`toColumn`/`fromColumn` and those locals are named `from` and `originColumn`. A census keyed on variable names will keep missing guards wherever a local was named for its role in the function.

## Two real defects, not tidying

**1. A renamed board could merge with its re-review never run.** `default-workflow-hooks.ts` is named for the default workflow, but the store runs it on the flag-ON path for *every* workflow — the trait registry resolves hooks by trait id, not by workflow. Its reopen predicates listed the default lineage's column names, so on a renamed board **no reopen effect fired at all**. One of them clears `workflowStepResults`, which `getTaskMergeBlocker` reads: a card bounced out of review carried its old `passed` result back in, and that satisfies the merge gate. Same regression the graph-owned-crossing carve-out exists to prevent, arriving through the other door. (Two smaller ones rode along: failure state never cleared on a renamed reopen, and an operator dragging a card back to the queue never parked it, so the scheduler re-dispatched what they had just pulled back.)

**I forgot the carve-out on my first pass, and that was worse than not converting.** A role-resolved clear plus a *name*-matched exemption means a renamed board takes the clear and never the exemption, destroying the remediation input the graph had just written. My own paired negative test caught it.

**2. The last-resort recovery for completed-but-stranded work did not exist off the default lineage.** In `recoverCompletedTask`, `promotedFromPlannerColumn` was false on a renamed board, so finished work resting in the planning lane was never promoted — the code fell through to `handoffTaskToReview` straight from the planning column, and role adjacency has no planning → review edge, so the handoff was rejected and the card stayed stuck with its work complete. I converted the promotion **target** too: resolving the lane and then moving to a literal `in-progress` is the half-conversion I have already been burned by twice this program, where the guard starts admitting cards and the move then sends them to a column the board does not declare.

## E2E evidence

`renamed-board-reopen.pg.test.ts` drives a **real PostgreSQL store** and a real `moveTask` on a workflow whose columns carry the standard traits under non-default names. The unit tests cannot show this: if `moves.ts` passed `undefined`, every unit case still passes via the no-basis fallback while the real board keeps the old behavior. **Proof it is load-bearing: forcing `moveLifecycleColumns` to `undefined` fails 2 of 3.** The executor suite covers both the split-role and the MERGED post-U11 shape.

## Revert proofs, isolated per site

| change reverted | result |
|---|---|
| reopen predicate → literal names | 4 of 10 fail |
| reopen field clears → literal names | 2 of 10 fail |
| `userPaused` hold lane → literal `todo` | 1 of 10 fail |
| graph carve-out → literal names | 1 of 10 fail |
| store passes `undefined` lifecycle columns | 2 of 3 fail (real PG) |
| `promotedFromPlannerColumn` → literals | 3 of 7 fail |
| two-hop condition → `=== "triage"` | 1 of 7 fails |
| promotion target → `"in-progress"` | 3 of 7 fail |
| `isPlannerColumnFor` → literals | 1 of 7 fails |
| live-agent-count: one arm dropped | 2 of 11 fail |
| DocumentsView: trait branch removed | 3 of 7 fail |
| planner role renamed to `"planner"` | 4 fail (2 pre-existing) |

Every conversion is paired with a negative case (a forward move, a not-a-planner-lane card, a default-lineage card, a renamed column with no traits), so neither "always fire" nor "never fire" can pass for "resolve the role".

## Deliberately NOT converted, with reasons

- **`moves.ts` flag-OFF inline block (4).** That branch *is* the legacy path, kept verbatim so the two can be parity-checked. Converting it erases the reference implementation.
- **`live-agent-count.ts`'s no-flags fallback.** Reachable, and there is nothing to resolve from — `enrich…FromFlags` exists for callers with board flags rather than an IR, so a column missing from that map is the renamed case. "Not intake" is as much a guess as "todo is intake", and Running/Waiting are complements, so a card matching neither arm is reported as neither and the footer's queued total under-reports it. The real fix is at the caller; four new cases pin that flags override the legacy answer **in both directions**. What did change is the duplication: two hand-written copies of one rule now call one named function.
- **`MissionControlPanel`'s `FUNNEL_STAGES`.** An alias table of column *names* where `triage` sits beside `signal` and `backlog`. Command Center aggregates across projects, so there is no single workflow to resolve traits from — the honest conversion is a data change, not a predicate change.
- **`DocumentsView` with no traits.** Same no-basis rule; the documents list is full of historical columns absent from the current board. A case asserts a renamed column with no traits still reads as "working", documenting the gap rather than hiding it.

## Fixture findings

Each cost a red run that looked like the code under test:

- a `merge-blocker` column needs a reachable merge-class node, or `parseWorkflowIr` rejects the workflow;
- a back-edge must be `kind: "rework"`, and a rework edge is legal only **into** a node with `config.reworkRegion: true`;
- a workflow gets role-level transitions only when it declares wip + review + complete + **archived** plus a planning lane — without the archived column, adjacency falls back to order-derived neighbours and `checking -> queued` is not a legal move at all;
- `recoverCompletedTask` only *reaches* the promotion seam when nothing is left to gate; without passed `plan-review`/`code-review` rows it re-enters the workflow graph and returns first, so a naive fixture silently tests the wrong branch and every assertion reads "no moves happened" for an unrelated reason.

## Verification

- `pnpm test:gate` **71/71**
- new suites: 10/10 reopen-semantics, 3/3 renamed-board-reopen (real PG), 7/7 executor-planner-lanes, 7/7 documents-status-dot, 4/4 planner-role-is-not-a-column
- neighbours: 132 + 10 + 482 (gate shards), 350/351 engine planning/replan suites, 64/64 agent-prompts, 51/51 usage-limit-detector, 11/11 live-agent-count, 11/11 dashboard hook/log suites
- the single engine failure (`executor-fast-mode-workflows.test.ts` › "raw fast mode still invokes non-executable review seam nodes") **reproduces with my changes stashed** — pre-existing on `origin/main`
- typechecks clean for core, engine, and dashboard-app (`tsconfig.app.json`; `tsconfig.json` checks nothing under `app/`); `pnpm lint` clean

### DIFF
--- a/.changeset/dashboard-planner-role-and-documents-dot.md
+++ b/.changeset/dashboard-planner-role-and-documents-dot.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Task Documents shows the correct status dot for tasks on renamed or custom board columns.
+category: fix
+dev: `DocumentsView` takes optional per-task column traits (threaded from App's existing footer map through `MainContent`) and resolves the dot by role; five dashboard `agent === "triage"` role comparisons now use `PLANNER_AGENT_ROLE`.
--- a/.changeset/executor-planner-lanes-resolved.md
+++ b/.changeset/executor-planner-lanes-resolved.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Completed work stranded in a renamed planning column is now recovered instead of stuck there.
+category: fix
+dev: `recoverCompletedTask` resolves the planner lanes and the promotion target from the task's own workflow (`resolvePlannerLanes`, now shared from `replan-target.ts`), and the planning-evacuation branch of the `task:moved` handler uses the same classification via `isPlannerColumnFor`.
--- a/.changeset/live-agent-count-legacy-fallback.md
+++ b/.changeset/live-agent-count-legacy-fallback.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: The processing/queued footer counts now share one rule for columns with no trait flags.
+category: internal
+dev: `live-agent-count.ts`'s two duplicate no-flags fallbacks collapse into `isLegacyPreImplementationColumn`; deliberately still the legacy pair, with the reason recorded at the helper. `replan-target.ts` comment prose restated by role.
--- a/.changeset/reopen-semantics-by-role.md
+++ b/.changeset/reopen-semantics-by-role.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Reopening a card on a renamed board now clears its stale review results, branch and failure state.
+category: fix
+dev: `default-workflow-hooks.ts` reopen predicates resolve intake/hold/wip/review/complete by trait from the task's own IR (passed in from `moves.ts` as `DefaultWorkflowMoveContext.lifecycleColumns`) instead of matching the default lineage's column names. `isReopenIntoPlanning` is exported so the store's former "parity mirror" calls it. The flag-OFF inline block in `moves.ts` stays name-based as the parity reference.
--- a/packages/core/src/__tests__/live-agent-count.test.ts
+++ b/packages/core/src/__tests__/live-agent-count.test.ts
@@ -88,3 +88,51 @@ describe("live agent count predicates", () => {
     });
   });
 });
+
+/*
+FNXC:WorkflowLifecycleColumns 2026-07-30-10:30 (Phase C convergence — live-agent-count.ts):
+
+The no-flags fallback is DELIBERATELY the legacy pair, and these cases exist so a future
+"finish the conversion" pass cannot quietly change the answer. Running and Waiting are
+complements over the same rows, so if the two former literal sites ever disagree a card
+lands in both counts or in neither, and the footer's queued total misreports it.
+
+What is pinned:
+  - with NO flags, the legacy planner ids are Waiting (unchanged behavior);
+  - with NO flags, a renamed planner column is NOT Waiting — the known gap, whose fix is at
+    the caller (supply flags, or use the IR-based `enrichRunningAgentTaskShape`), not a guess
+    about what an absent flag set means;
+  - flags always WIN over the fallback, in both directions, which is what makes the caller
+    fix effective.
+*/
+describe("the no-flags fallback keeps the legacy planner vocabulary", () => {
+  const bare = (column: string) => ({ id: "FN-1", column } as Parameters<typeof isWaitingAgentTask>[0]);
+
+  it("treats the legacy planner ids as waiting when no flags are supplied", () => {
+    expect(isWaitingAgentTask(enrichRunningAgentTaskShapeFromFlags(bare("triage")))).toBe(true);
+    expect(isWaitingAgentTask(enrichRunningAgentTaskShapeFromFlags(bare("todo")))).toBe(true);
+    expect(isWaitingAgentTask(enrichRunningAgentTaskShapeFromFlags(bare("in-review")))).toBe(false);
+  });
+
+  it("answers identically whether the shape was enriched or read raw", () => {
+    // The two former literal sites: `enrich...FromFlags` and `isWaitingAgentTask`'s own
+    // `??` fallback. One rule, so one answer.
+    for (const column of ["triage", "todo", "in-progress", "backlog"]) {
+      expect(isWaitingAgentTask(enrichRunningAgentTaskShapeFromFlags(bare(column))))
+        .toBe(isWaitingAgentTask(bare(column)));
+    }
+  });
+
+  it("does NOT invent a planner lane for a renamed column with no flags", () => {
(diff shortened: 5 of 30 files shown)


gsxdsm:
Both P1s fixed and pushed. You were right on both, and the first one is the more important finding: **my half-conversion made that branch worse than the bug it replaced.**

**1. Renamed forward moves triggered evacuation.** I made the evacuation branch’s *source* check role-aware and left the destination exclusions literal, so on a renamed board a forward move (planning → building) passed the source test and matched no exclusion — the evacuation fired on a card that was merely advancing, aborting live planning work and deleting a valid pre-execution worktree. Before my change the source check failed and nothing happened; after it, active damage. Third time this program has produced that exact shape (role-aware gate, name-matched destinations), so it is now one predicate — `isBackwardMoveOutOfPlanning` — resolving the source lanes and the forward targets from the same lane set.

Deliberately unchanged: planner-to-planner moves (`todo → triage`, a replan rebound) still evacuate, as today. Whether they *should* is a real question — the replan path is engine-initiated, so aborting its planning session may be exactly wrong — but that is a behavior change with its own surfaces to enumerate, not a review fix.

**2. Missing WIP role invented a column.** Exactly right, including the half-moved consequence: the intake → hold re-home is issued *first*, so the rejected promotion could leave the card between lanes — worse than the stranding this recovery exists to fix. `wip`/`review`/`complete` are now **optional** when the workflow speaks columns, and a new `resolvedFromWorkflow` flag distinguishes the two cases my single `?? legacy` had collapsed: *no column vocabulary at all* (legacy names are the answer) versus *declares columns but not that role* (no substitution — refuse). The type change surfaced the defect immediately, since `moveTask` stopped accepting the value. `recoverCompletedTask` refuses **before** any move and logs why — nothing else owns this state, so a silent withhold is indistinguishable from the stranding.

`hold`/`intake` still fall back to each other on purpose: a workflow with intake and no hold rests planning work in intake, which is a real shape, not an invented column.

Revert proofs, isolated:
- forward exclusions → literals: **1 of 14 fails** (the three renamed forward moves)
- missing-wip refusal removed + legacy `wip` restored: **2 of 14 fail** (no move issued; the log message)

The paired cases pass under both reverts — evacuation still fires for a genuine withdrawal, the legacy vocabulary still answers when no workflow resolves, and a workflow that *does* declare a wip lane still promotes — so neither "never fire" nor "always refuse" can pass for the real predicate.

Verified: 14/14 new, 312/312 across executor planner-lanes + three replan-target suites + triage + planning-evacuation, `pnpm test:gate` 71/71, engine typecheck clean, `pnpm lint` clean.

PR #2630, opened 2026-07-30
U12 R12: lifecycle-column literal ratchet — and the raw count's floor is not zero

gsxdsm:
The anti-regression ratchet U12 R12 calls for. Counts lifecycle-column **literal comparisons** in production source and fails when the count rises.

## Two jobs

**1. Ratchet.** A converted guard cannot silently come back as a literal. Ceilings only go down.

**Mutation-verified both ways:** adding one `t.column === "triage"` fails with *"rose to 49 (ceiling 48)"*; lowering the ceiling to 47 fails with *"rose to 48 (ceiling 47)"*. The number is exact, not approximately right.

**2. Honest denominator — the finding.**

`triage` is overloaded in this codebase: a column id, an **agent role**, a **session purpose**, a **prompt-template family**, and a **CLI glyph key**. A raw grep counts them together, which makes "reach zero" unreachable by construction — converting `role === "triage"` in `agent-prompts.ts` would break the planning agent's prompt-template resolution, and the failure would look nothing like a column bug.

| | count |
|---|---:|
| raw `triage` matches | 72 |
| **not a column at all** | **10** |
| genuine column comparisons | **48** |

The 10: `agent-prompts.ts` ×3 (`role`), `usage-limit-detector.ts` ×2 (`agentType`), `skill-resolver.ts` (`sessionPurpose`), `tool-availability.ts` (`surface`), `cli/commands/task.ts` (a glyph key), plus two in comments.

Ceilings recorded for all four ids — **`in-progress` (133) and `in-review` (200) were untracked entirely.**

## A measurement error of mine that writing this caught

A grep over `packages/<pkg>/src` **misses `packages/dashboard/app`**, where the board components live. That undercounted `triage` as 43 in an earlier audit of mine when it was 62. The source roots are now listed explicitly in code so the number cannot drift with someone's glob.

## The classifier is under test, not trusted

A ratchet that matched nothing would pass forever while measuring nothing — the failure mode this program keeps finding. Three self-tests prevent it:

- it asserts **positively** that `agent-prompts` / `skill-resolver` / `tool-availability` are excluded, so a classifier change that swallowed them would fail rather than quietly shrink the number;
- it asserts the classifier still **matches** real column comparisons;
- it pins `live-agent-count.ts`'s two sites as **permanent** no-flags fallbacks, with the reason, so a future edit that deletes them has to argue with it rather than silently drop stranded cards from the footer's queued total.

The classifier keys on the **left-hand side naming a column** — deliberately syntactic, so a reader can audit it against the source without running anything, and conservative: an unrecognised shape counts **as** a column comparison, erring toward demanding conversion rather than excusing it.

7 tests green; lint clean.

### DIFF
--- a/packages/core/src/__tests__/no-hardcoded-lifecycle-columns.test.ts
+++ b/packages/core/src/__tests__/no-hardcoded-lifecycle-columns.test.ts
@@ -1,253 +1,243 @@
 /*
-FNXC:LifecycleColumnRatchet 2026-07-30-12:10 (U12 R12 — the anti-regression ratchet):
+FNXC:LifecycleColumnRatchet 2026-07-31-09:10 (U12 R12 — AST, replacing the grep):
 
-Counts lifecycle-column LITERAL COMPARISONS in production source and fails when the count rises.
-Two jobs, and the second is why a plain grep is not enough:
+THE MEASURING INSTRUMENT for the lifecycle-column conversion, and the authority on the number.
 
-1. RATCHET. A converted guard cannot silently come back as a literal. The number only goes down.
+WHY IT HAD TO STOP BEING A GREP. Three people measured this surface with three regexes and got
+three answers (6, 8, 12 for the agent-role bucket alone). Every figure quoted at the program today
+— 57, 48, 45, 34, 56, 46 — was grep-derived, and those were not measurements but estimates with a
+consistent bias. A regex cannot distinguish:
 
-2. HONEST DENOMINATOR. `triage` is overloaded in this codebase — it is a column id, an AGENT ROLE,
-   a SESSION PURPOSE, a PROMPT-TEMPLATE family, and a CLI glyph key. A raw grep counts all of them
-   together, which makes the convergence target unreachable by construction: converting
-   `role === "triage"` in agent-prompts.ts would break the planning agent's prompt-template
-   resolution, and the resulting failure would look nothing like a column bug.
+  task.column === "triage"        a lifecycle guard           <- the thing being counted
+  role === "triage"               the planning AGENT's role   <- correct code, must never convert
+  sessionPurpose === "triage"     a session purpose           <- correct code
+  surface === "triage"            a docs surface name         <- correct code
+  `column === "triage"` in prose  an FNXC note quoting the OLD behaviour
 
-   Measured at the time of writing, `triage` alone: 72 raw matches across the five source roots, of
-   which 10 are NOT columns and 62 are. A further 2 of the 62 ARE columns but are documented
-   permanent fallbacks (see the live-agent-count case below), so the reachable floor for the RAW
-   number is 12, not 0 —
-   anyone tracking the raw grep toward zero is chasing a target that would require breaking working
-   code to hit.
+Parsing removes two whole defect classes instead of patching them:
+  - COMMENTS ARE NOT NODES. Every FNXC note here explains an old comparison by quoting it, so a text
+    scan counts the project's own requirement history as violations. The previous revision of this
+    file needed a hand-rolled block-comment tracker for exactly that, and still only caught the
+    cases it thought to look for.
+  - QUOTE STYLE AND LINE BREAKS VANISH. `"triage"`, `'triage'`, and a comparison wrapped across
+    lines are one shape to the AST and three patterns to a grep.
 
-   The scope also matters and is why this lives in code rather than in a shell one-liner. A grep over
-   `packages/<pkg>/src` MISSES `packages/dashboard/app` entirely, where the board components live — that
-   omission undercounted `triage` as 43 when it is 62. Writing the ratchet is what caught it; the
-   roots are listed explicitly below so the number cannot drift with someone's glob.
-
-The classifier keys on the LEFT-HAND SIDE naming a column (`task.column`, `fromColumn`,
-`currentColumn`, ...). That is deliberately syntactic and therefore auditable: a reader can check it
-against the source without running anything. It is also conservative — an unrecognised shape counts
-as a column comparison, so the ratchet errs toward demanding conversion rather than excusing it.
+WHAT IT DOES NOT DO. There is no type checker here, only a syntax tree, so classification is by
+RECEIVER NAME. That is a real limit and it is why the lists below are explicit and auditable rather
+than clever. It errs toward COUNTING: an unrecognised receiver is treated as a column, so a new
+binding name inflates the number and demands attention instead of disappearing from it.
 */
 import { describe, expect, it } from "vitest";
-import { execFileSync } from "node:child_process";
-import { readFileSync } from "node:fs";
-import { join } from "node:path";
+import { readFileSync, readdirSync, statSync } from "node:fs";
+import { join, relative } from "node:path";
+import ts from "typescript";
 
 /** Lifecycle column ids whose literal comparison this ratchet governs. */
 const GOVERNED_IDS = ["triage", "todo", "in-progress", "in-review"] as const;
 
 /**
- * Per-id ceilings, re-measured against `main` with the INVERTED classifier (2026-07-30).
+ * Receivers that are provably NOT lifecycle columns. Sourced from two independently-built
+ * classifiers agreeing — the strongest evidence available on this surface.
  *
- * These numbers went UP from the previous set (triage 48 -> 57, in-progress 133 -> 205). No literals
- * were added: the classifier got honest. It used to require the receiver to LOOK like a column and
- * therefore could not see `col`, `c`, or `from`/`to`. Treat the jump as a correction to the
- * measurement, not a regression in the code. LOWER THESE as conversions land; never raise one.
- * A raise means a literal came back — convert the site or, if it genuinely is not a column, teach
- * the classifier why rather than widening the ceiling.
+ * Converting any of these would be a real bug rather than a missed cleanup: `role === "triage"`
+ * selects the planning agent's prompt template, and resolving it to "which column carries the
+ * intake trait" asks a column question about something that is not a column.
  */
-const CEILINGS: Record<string, number> = {
-  triage: 40,
-  todo: 82,
-  "in-progress": 200,
-  "in-review": 214,
-};
-
-const REPO_ROOT = join(import.meta.dirname, "..", "..", "..", "..");
+const NON_COLUMN_RECEIVERS: ReadonlySet<string> = new Set([
+  "role",
+  "agentType",
+  "sessionPurpose",
+  "surface",
+  "agent",
+  "purpose",
+  "lane",
+]);
 
 /**
- * Is this 1-indexed line inside a `/* … *\/` block? Cheap state scan rather than a parser: the
- * ratchet must not need a TypeScript AST to tell documentation from a guard, and every FNXC note in
- * this codebase explaining an OLD comparison would otherwise be counted as the comparison itself.
+ * Column receivers actually present, from the receiver census (task.column 12, toColumn 9,
+ * column 9, t.column 2, originColumn 2, then singletons). Documentation, not a filter — anything
+ * outside NON_COLUMN_RECEIVERS counts regardless.
  */
-const blockCommentLinesByFile = new Map<string, ReadonlySet<number>>();
+const KNOWN_COLUMN_RECEIVERS: ReadonlySet<string> = new Set([
+  "column", "toColumn", "fromColumn", "originColumn", "resumeColumn", "taskColumn",
+  "from", "to", "c", "col", "workflowIrPinColumnId", "currentColumn", "targetColumn",
+]);
 
-function isInsideBlockComment(file: string, lineNo: number): boolean {
-  let cached = blockCommentLinesByFile.get(file);
-  if (!cached) {
-    const inside = new Set<number>();
-    let open = false;
-    const lines = readFileSync(join(REPO_ROOT, file), "utf-8").split("\n");
-    lines.forEach((text, index) => {
-      const opensHere = text.includes("/*");
-      const closesHere = text.includes("*/");
-      if (open) inside.add(index + 1);
-      if (opensHere && !closesHere) {
-        open = true;
-        inside.add(index + 1);
-      } else if (closesHere) {
-        if (opensHere) inside.add(index + 1);
-        open = false;
+const REPO_ROOT = join(import.meta.dirname, "..", "..", "..", "..");
+const SOURCE_ROOTS = [
+  "packages/core/src",
+  "packages/engine/src",
+  "packages/dashboard/src",
+  "packages/dashboard/app",
+  "packages/cli/src",
+];
+
+interface Site { readonly file: string; readonly line: number; readonly code: string; readonly receiver: string }
+
+function sourceFiles(): string[] {
+  const out: string[] = [];
+  const walk = (dir: string): void => {
+    for (const entry of readdirSync(dir)) {
+      const full = join(dir, entry);
+      if (statSync(full).isDirectory()) {
+        if (entry === "__tests__" || entry === "node_modules") continue;
+        walk(full);
+        continue;
       }
-    });
-    cached = inside;
-    blockCommentLinesByFile.set(file, cached);
-  }
-  return cached.has(lineNo);
+      if (!/\.tsx?$/.test(entry) || /\.test\.tsx?$/.test(entry)) continue;
+      out.push(full);
+    }
+  };
+  for (const root of SOURCE_ROOTS) walk(join(REPO_ROOT, root));
+  return out;
 }
 
-
-
-function comparisonSites(columnId: string): { file: string; line: string; code: string }[] {
-  let out = "";
-  try {
-    out = execFileSync(
-      "grep",
-      [
-        "-rnE",
-        `(===|!==)[[:space:]]*"${columnId}"|"${columnId}"[[:space:]]*(===|!==)`,
-        "--include=*.ts",
-        "--include=*.tsx",
-        "packages/core/src",
-        "packages/engine/src",
-        "packages/dashboard/src",
-        "packages/dashboard/app",
-        "packages/cli/src",
-      ],
-      { cwd: REPO_ROOT, encoding: "utf-8", maxBuffer: 32 * 1024 * 1024 },
-    );
-  } catch (err) {
-    // grep exits 1 on no matches, which is a legitimate zero rather than a failure.
-    const status = (err as { status?: number }).status;
-    if (status !== 1) throw err;
-    return [];
-  }
-
-  /*
-  FNXC:LifecycleColumnRatchet 2026-07-30-21:10 (classifier INVERTED after a measured miss):
-  This used to require the left-hand side to LOOK like a column (`\w*[Cc]olumn`), which silently
-  under-counted: `col`, `c`, and `from`/`to` are column receivers that pattern cannot match. Measured
-  against a full comparison sweep it missed 2 genuine guards — `executor.ts`'s
-  `from === "todo" || from === "triage"` and MissionControlPanel's `(c) => c === "triage"` — while
-  the program tracked the smaller number as ground truth.
-
-  So it now counts EVERY comparison and subtracts an explicit allowlist of receivers that are
-  provably not columns. Conservative in the correct direction: an unrecognised receiver COUNTS, so a
-  new binding name inflates the number and demands attention rather than vanishing from it.
-
-  The allowlist is receivers, not files, so it cannot excuse a real guard that happens to live in an
-  allowlisted file.
-  */
-  const NON_COLUMN_RECEIVERS = [
-    "role",            // agent-prompts, TaskChatTab — the planning AGENT's role
-    "agentType",       // usage-limit-detector — which lane hit the limit
-    "sessionPurpose",  // skill-resolver
-    "surface",         // tool-availability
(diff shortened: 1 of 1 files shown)


gsxdsm:
Verifying the "must FAIL on a reintroduced literal" requirement empirically, since I hit the same class of hole in my own ratchet (#2623) and it took an injection to find it.

**Measured on `feature/u12-lifecycle-column-ratchet` at 9108dbb-era HEAD, injecting into `packages/engine/src/scheduler.ts` (inside `SOURCE_ROOTS`, so it is scanned):**

| injected `task.column === "triage"` sites | result |
|---:|---|
| 1 | **16 passed** |
| 1 (single-quoted) | **16 passed** |
| 1 (split across lines) | **16 passed** |
| 5 | **16 passed** |
| 8 | **16 passed** |
| 10 | **16 passed** |
| 11 | **16 passed** |
| 12 | **16 passed** |
| 15 | 1 failed |
| 30 | 1 failed |

So the classifier is fine — it does see all three syntactic forms, and your two review fixes worked. The problem is arithmetic: **`CEILINGS.triage = 48` now sits ≥12 above the actual count**, so the ratchet silently absorbs at least twelve reintroduced literals before it trips. It counts today's literals and does not catch tomorrow's — not because of the pattern, but because the ceiling went stale the moment conversions landed after it was measured (#2614 and others).

That is the same defect I had to fix in mine, one layer up: a ratchet with headroom does not ratchet.

**Two things that would close it, both cheap:**

1. **Re-measure the ceilings at merge time, and treat any slack as a bug.** A ceiling equal to the actual count fails on the first reintroduced literal — which is the stated requirement. Consider asserting `sites.length === ceiling` for ids you believe are converged, so drift in *either* direction is visible; `<=` cannot distinguish "clean" from "12 under budget".
2. **Make the enforcement path refuse to RAISE.** Mine had the mirror problem: `--update` wrote current counts verbatim, so anyone who added a literal and ran the documented update command locked the regression in as the new ceiling. If ceilings here are ever regenerated rather than hand-edited, the same hole applies.

On the coordinator's point about this becoming the measuring instrument: strongly agree, and worth noting the roots list in this file is already right where a shell glob is not. `packages/*/src/**` silently drops files directly under `src/` (wildmatch needs a directory level to consume) — that omission hid `self-healing.ts`, the single largest holder, and reported 11 of 25 for me. Your explicit `SOURCE_ROOTS` including `packages/dashboard/app` is the correct shape.

Happy to hand over the comment-stripping + ceiling-refusal logic from `scripts/check-lifecycle-column-literals.mjs` (#2623) if it saves duplicating it — it also tracks membership forms (`new Set([...])`, `.includes`, `switch case`, `?? "triage"`) separately, which is how the routine-editor default that seeded tasks into the deleted column stayed invisible to every comparison count.

gsxdsm:
Sequencing input on the `useWorkflow` decision, measured rather than argued — because it changes the answer for one of the five `moves.ts` guards.

I traced which side of `if (useWorkflow)` each cited line actually sits on, by brace-depth on current `origin/main` (`if` opens at :789, `else` at :837, block closes at :960):

| line | branch | fate if the flag goes always-on |
|---|---|---|
| **816** | **FLAG-ON (`if`)** | **survives — becomes LIVE path. Must be converted regardless.** |
| 869 | FLAG-OFF (`else`) | deleted with the branch |
| 946 | FLAG-OFF (`else`) | deleted with the branch |
| 947 | FLAG-OFF (`else`) | deleted with the branch |
| 952 | FLAG-OFF (`else`) | deleted with the branch |

So "U12 resolves the flag first, then the flag-OFF branch goes away wholesale" is right for **4 of the 5** — but **:816 is already in the flag-ON branch**. Waiting does not remove it; the flag flip *promotes* it from dead to live. If it is left unconverted on the assumption that sequencing disposes of it, the flag flip makes a `toColumn === "todo" || toColumn === "triage"` reset guard live on the hottest lifecycle path, and it will not match a merged or renamed column. That is the same silent-skip shape this program exists to remove, arriving at the exact moment the flag makes it matter.

Recommendation, unchanged in shape but corrected in scope: U12 resolves the flag first, and **:816 is converted as part of that flip** (it is the branch being promoted, so it belongs to whoever flips it), while 869/946/947/952 are deleted with the `else`. Nobody needs to convert those four.

Separately, for anyone reasoning about `moves.ts` branches: **the in-transaction CAPACITY gate is not affected by this decision.** It is gated on `if (capacityIr && fromColumn !== toColumn)`, not on `useWorkflow` — `capacityIr` resolves from the task's own selection independently of the flag. That was deliberate: the original Phase A3 ground-truth measurement (R2) found the capacity block sitting inside the `useWorkflow` branch and therefore never running in production, and #2488/#2499 moved it out. Re-verified on current `origin/main` just now: `workflow-capacity-invariant.pg.test.ts` **7/7 green**, including the two ratchets that fail if the capacity read moves back outside the per-task advisory lock. Flipping the flag will not disturb it, and the flip does not need to account for capacity.

gsxdsm:
Not a competing change — a working classifier for the distinction the coordinator asked this ratchet to learn, plus the number it produces on current `main`. Take it, adapt it, or ignore it.

**On current `origin/main`, receiver-bucketed and comment-stripped: 37 column guards, 12 role/lane, 0 unclear (49 total).** So the real bar is **37**, not 45 or 55.

## The rule

Bucket on the **receiver of the comparison**, not the literal. That separates guards from roles with no judgement calls and no maintained exception list:

- **guard** — `column`, `col`, `toColumn`, `fromColumn`, `taskColumn`, `from`, `to`, `c`
- **not a guard** — `role`, `agent`, `agentType`, `surface`, `sessionPurpose`, `entry.agent`

## The 12 that must be ignored

`agent-prompts.ts:1469,1472,1480` (`role`) · `effective-model-resolution.ts:148`, `TaskChatTab.tsx:82`, `useTasks.ts:162` (`entry.agent`) · `TaskChatTab.tsx:93` (`role`) · `AgentLogViewer.tsx:104` (`agent`) · `usage-limit-detector.ts:144,207` (`agentType`) · `skill-resolver.ts:432` (`sessionPurpose`) · `tool-availability.ts:32` (`surface`)

**Six of those are currently on the coordinator's "real column guards" list** — `usage-limit-detector` ×2, `skill-resolver`, `tool-availability`, `effective-model-resolution`, `useTasks`. A ratchet built from that list would demand their conversion, and converting `agentType === "triage"` to a column trait is exactly the bug we're trying to prevent. This is the case for mechanising the split rather than maintaining it by hand.

## Script

```python
import re, pathlib
pat  = re.compile(r'([A-Za-z_$][\w$.?\[\]"\']*)\s*(===|!==)\s*"triage"')
pat2 = re.compile(r'"triage"\s*(===|!==)\s*([A-Za-z_$][\w$.?\[\]"\']*)')
COLUMN = re.compile(r'column|\bcol\b|\bfrom\b|\bto\b|\bc\b', re.I)
ROLE   = re.compile(r'role|agent|surface|purpose|lane|capability', re.I)

def strip(src):  # comments are not code; the FNXC prose stays and must not count
    src = re.sub(r'/\*.*?\*/', lambda m: '\n' * m.group(0).count('\n'), src, flags=re.S)
    return '\n'.join('' if l.lstrip().startswith(('//', '*')) else re.sub(r'//.*$', '', l)
                     for l in src.split('\n'))

for p in pathlib.Path('packages').glob('*/**/*.ts*'):
    sp = str(p)
    if any(x in sp for x in ('__tests__', '.test.', 'node_modules', '/dist/')): continue
    for i, line in enumerate(strip(p.read_text(errors='ignore')).split('\n'), 1):
        for m in list(pat.finditer(line)) + list(pat2.finditer(line)):
            recv = m.group(1) if m.re is pat else m.group(2)
            kind = 'ROLE' if ROLE.search(recv) else ('COLUMN' if COLUMN.search(recv) else 'UNCLEAR')
            if kind == 'COLUMN': print(f"{sp}:{i}  {recv}")
```

An AST version is strictly better — this is a regex and will drift on an unusual receiver — but it is right on all 49 sites today and gives you a number to gate against now. Worth keeping the `UNCLEAR` bucket and **failing** on it rather than defaulting it either way: an unrecognised receiver is precisely where a hand list goes wrong, and my first pass put `executor.ts:3410` (`from === "todo" || from === "triage"`) there before I added `from`/`to`.

On the coordinator's other note — the ratchet must **fail** on a reintroduced literal, not just count today's. Cheapest proof is a fixture file with one known-bad guard that the test asserts is caught, rather than adding one to real source and remembering to remove it.

gsxdsm:
Bracket access was real, and it is the third evasion found in this scanner by review — so I stopped patching the pattern and replaced detection with an AST walk, following the precedent `scripts/lib/capacity-pool-id-check.mjs` was forced onto for exactly this reason.

`task["column"] === "triage"` is an `ElementAccessExpression`, invisible to a property-access-only walk and to both regex branches. Once detection is structural the whole class disappears rather than getting another branch: quote style is `StringLiteral.text`, line breaks do not exist for a `BinaryExpression`, deep qualifiers resolve through `receiverName()`, and comments are absent from the tree so no line-prefix filter can be defeated by a block comment's interior.

**Measured against the previous logic, not argued:** double-quoted CAUGHT, single-quoted MISSED, multiline MISSED.

Two things beyond the finding:

**Portability.** The `execFileSync("grep", ["--include=..."])` shell-out is gone. On BSD grep and on Windows it made the suite die before any assertion — a dead guard on the platform the operator develops on, failing in the way that reads as an unrelated environment error.

**The classifier no longer fails open.** The name pattern `/\w*[Cc]olumn/` silently answered "not a column" for every receiver not ending in `column` — 22 sites written as `to`, `from`, `col`, `c` or a destructured binding. Receivers are now exhaustively classified and an **unknown one fails the test**, so a new binding style forces a human judgement instead of quietly dropping out of the count. I also added the task-STATUS vocabulary the allowlist was missing: `status === "in-progress"` is 46 sites and is not a column.

**Acceptance test as specified:** 11 evasive shapes go RED (double-quoted, single-quoted, multiline, negated, reversed operands, bracket access, deep qualified, `col`, `to`, destructured, parenthesised/non-null) and 7 legitimate shapes stay GREEN (agent role/display-name/type, session purpose, task status, line comment, block comment). Driven through the pure detector, so proving sensitivity needs no file edits.

**Enforced counts, from the gating code itself:** `triage` **38**, `todo` 83, `in-progress` 153, `in-review` 218. `triage` tightens 40 → 38. `in-progress` and `in-review` rise 133 → 153 and 200 → 218 — that is 38 sites the old scanner could not see, not 38 regressions, and it is the one legitimate reason to raise a ceiling. Recorded in the constant's doc so no future raise gets the same excuse.

Note three numbers are in circulation for `triage`: a raw grep says 34, the hand classification says 45 real guards out of 55, this says 38. Only one gates. Where a hand count and the instrument disagree, the receiver map is where to settle it.

Rebased onto the inverted-classifier revision, keeping its allowlist and file attributions. `pnpm lint` clean, `pnpm test:gate` green (10 / 482 / 71), suite 36/36.

gsxdsm:
Real, and it was live rather than theoretical — so this is the third fail-open in this one file, reached a third way.

`receiverName` returned `undefined` for any expression it did not understand, and the caller **dropped** those sites. I checked what the scanner was reporting as unnamed instead of trusting it, and found `metadataColumn(entry, "to") === "in-review"` in `dashboard/src/reliability-metrics.ts` — **three real lifecycle guards this ratchet was not counting at all.**

The pattern is worth naming because it has now repeated three times with three different mechanisms:

- the grep failed open on **quoting** (single-quoted, multiline)
- the name pattern failed open on **short bindings** (`to`, `from`, `col`, `c`)
- `receiverName` failed open on **unnameable expressions** (helper calls, ternaries)

Each time the shape was different and the failure mode identical: a site the scanner did not understand silently became "not a column". So the fix is the property, not the case. A call now resolves to its **callee name** (`metadataColumn`), which must then be classified like any other receiver. Anything still unnameable yields a `<SyntaxKind>` sentinel that is **counted** and trips the classification test — a human judges it instead of it vanishing.

There is now a test asserting exactly that, using the ternary from your comment: it must return one site whose receiver is a sentinel that is in neither classification set. If a future revision goes back to failing open, that case returns 0 and fails. It is the guard on the guard, and it is the reason the number is safe to gate on at all.

`in-review` 218 → 219 for the three recovered guards. Every ceiling raise in this file has been detection improving rather than a regression, and each one is named individually in the constant's doc so no future raise can borrow the excuse.

`pnpm lint` clean, `pnpm test:gate` green (10 / 482 / 71), `tsc` clean, suite 39/39.

gsxdsm:
Following up on my earlier comment here now that AST is the agreed approach — the regex I posted is a **specification and a test corpus**, not an implementation to port. The reusable parts are the bucket definitions and the 49 known answers to check a real parser against.

**Two sites in the coordinator's latest receiver breakdown are wrong, and both are exactly what requirement 2 exists to catch:**

- **`workflowIrPinColumnId`** — listed as a column receiver. It appears **only inside prose**: `replan-target.ts:185` is an FNXC note with a backtick-quoted `` `workflowIrPinColumnId === "triage"` ``, and the other hit is a test comment. There is no such code anywhere. A parser sees zero; every grep sees two.
- **`c`** — `MissionControlPanel.tsx:46`, `match: (c) => c === "triage" || c === "signal" || c === "backlog"`. That's a heuristic **name-alias matcher** for a canonical SDLC funnel stage, sitting beside `signal` and `backlog` with an `other` fallback. It buckets arbitrary boards for display; it is not a lifecycle guard, and converting it would be wrong.

So the classification needs a fourth bucket beyond COLUMN / ROLE / STATUS: **display-alias**, for comparisons whose receiver is a column id but whose purpose is presentation rather than a lifecycle decision. `DocumentsView`'s status-dot selection is the other one I know of.

**Concrete AST guidance, from where my regex actually broke:**

1. My first pass put `executor.ts:3410` in UNCLEAR because the receiver is `from`, not `column` — a real guard hiding behind a short name. Name-based receiver matching will always have this hole. With a real parser you can do better: resolve the declared type of the left operand and treat it as a guard when it is `ColumnId` / `Column` / `Task["column"]`. That is the check that makes the ratchet authoritative rather than another heuristic.
2. Keep an **UNCLEAR bucket and fail on it.** An unrecognised receiver is precisely where hand lists and greps have gone wrong all day; failing forces a human decision instead of silently picking a side.
3. For requirement 4, a **fixture file** with one known-bad guard per shape (double-quoted, single-quoted, multiline, deeper-qualified) is safer than reintroducing one into real source and remembering to remove it — and it makes each shape a named test rather than a manual check.

**Known-answer corpus to validate against, on current `main`:** 12 ROLE sites (`role` ×4, `entry.agent` ×3, `agentType` ×2, `agent`, `sessionPurpose`, `surface`) — two independent implementations agree on exactly this set — plus ~182 STATUS from #2633's tool. If a new parser disagrees with both on any of the 12, the parser is wrong.

gsxdsm:
**Two role-bucket sites that appear on no list circulated today** — found by a receiver-classifying pass while checking my own files. Flagging them here rather than anywhere else, because the consequence lands on this ratchet specifically.

| Site | Receiver | What it is |
|---|---|---|
| `dashboard/app/components/effective-model-resolution.ts:148` | `entry.agent !== "triage"` | agent-log entry's AGENT ROLE — filters log entries by which agent wrote them |
| `dashboard/app/hooks/useTasks.ts` | `entry.agent === "triage"` | same shape, same file family |

Neither is a column. Both must be in the ratchet's allowlist, and neither should ever be reported as a violation — flagging them would push someone into resolving an agent role to a column trait, which is the bug the coordinator described.

**The more useful point for #2630's design:** my pass, MAIN's classifier, and the coordinator's hand grep all produced *different sets* even where the totals happened to agree. Mine and MAIN's both said 12 role-bucket sites, but mine includes these two and MAIN's does not — so the matching totals were partly coincidence. Three methods, three sets.

That is an argument for one specific property beyond AST parsing: **the ratchet should report WHAT it matched, not only how many.** A number that disagrees with another number is unresolvable; a list that disagrees with another list is diagnosable in a minute. Everything painful about today's count dispute came from comparing totals with no way to diff them.

For a cheap cross-check while you build it, the classification that found these is: match `([A-Za-z_$][\w$.]*)\s*[!=]==\s*["']triage["']` over comment-stripped source and bucket by the captured receiver. It is still a regex and inherits every limitation you are replacing — its only virtue is that it names its matches. Current output on `origin/main`: **49 comparisons, 12 non-column, 37 real column guards**, with `self-healing.ts` alone using seven distinct receivers (`current.column`, `dep.column`, `linkedTask.column`, `live.column`, `resumeColumn`, `t.column`, `task.column`), which is why receiver-specific greps kept under-reporting it.

Not adding a script — #2633 owns measurement. This is just the two sites and the diffability argument.

gsxdsm:
Built the AST classifier for my own census (it was one of the three greps, so it had to change rather than be tuned). It is on #2623 as `scripts/lib/lifecycle-column-ast.mjs` — **importable, so you do not have to rebuild it**, which is what the handover instruction asks for.

```js
import { findColumnComparisons } from "./lib/lifecycle-column-ast.mjs";
// -> [{ line, lhs, tail, kind: "COLUMN" | "not-a-column" | "unknown" }]
```

It walks `ts.createSourceFile` binary expressions, handles either operand order, and classifies on the LHS trailing identifier: `col`/`*column` → COLUMN; a known non-column tail (role, agentType, agent, purpose, sessionPurpose, surface, kind, type, seam, template, glyph, name, id) → **not-a-column, never reported**; anything else → **unknown, surfaced for a human and never bucketed**.

**Measured on current `origin/main`, 1812 non-test source files:**

| class | count |
|---|---:|
| **COLUMN guards — the bar** | **35** |
| not-a-column (never enforced) | **12** |
| unknown (needs a human read) | **2** |

The 12 **independently reproduces MAIN's 12** from a different implementation — that is the agreement you asked not to waste, and it means the role bucket is settled at 12, not 6 or 8.

The 2 unknowns are the argument for having the class at all, and both want a human:
- `packages/dashboard/app/components/command-center/MissionControlPanel.tsx:46` — receiver `c`
- `packages/engine/src/executor.ts:3410` — receiver `from`

A single-letter or generically-named receiver cannot be classified syntactically, and guessing either way is how a real guard hides or a correct one gets converted.

**Proven to fail, on every form a reintroduced guard could hide in** — each red, in one run where the role file stays green:

| injected form | result |
|---|---|
| `task.column === "triage"` | ❌ flagged |
| `task.column === 'triage'` | ❌ flagged |
| split across lines | ❌ flagged |
| `x.i.deep.column !== "triage"` | ❌ flagged |
| `"triage" === task.column` (reversed) | ❌ flagged |
| `role` / `agentType` / `sessionPurpose` / `surface` / `entry.agent` | ✅ clean |

No grep, no shelling out, runs on macOS without GNU grep. Comment handling comes free from the parser, which structurally removes the "substring matched text in a comment" defect rather than mitigating it.

**One cost to know about before you adopt it:** parsing 1812 files is seconds, not milliseconds. My ratchet file went 25s → 17s after memoising the whole-tree run; two full parses remain because both are load-bearing (the enforcement exit code, and the `--update` refusal, which cannot use a partial file list). Worth budgeting for, and worth not "optimising" by narrowing the scan.

Also still open from my earlier comment, since it is independent of the classifier: your `CEILINGS.triage = 48` sits ≥12 above actual, so twelve injected literals pass. Re-measuring the ceilings is what makes requirement 4 hold; the classifier alone will not.

gsxdsm:
Evidence for "make the baseline follow the count down", measured on current `origin/main` with the census itself.

**The slack is 67, and it is not regression — it is unrecorded conversion.**

```
tree total: 787   baseline total: 854   SLACK: 67

FILES ABOVE BASELINE (1):
   +1  packages/dashboard/src/routes/register-task-workflow-routes.ts  (22 -> 23)

FILES BELOW BASELINE: 13, totalling 68 unrecorded conversions
   -18  packages/core/src/default-workflow-hooks.ts   (25 -> 7)
   -15  packages/engine/src/self-healing.ts           (126 -> 111)
    -8  packages/engine/src/executor.ts               (112 -> 104)
    -5  packages/core/src/task-store/moves.ts         (44 -> 39)
    -5  packages/engine/src/mission-feature-sync.ts   (5 -> 0)
    -4  packages/core/src/live-agent-count.ts         (10 -> 6)
```

**On the "+3 that slipped in": it is +1, and it is mine.** `register-task-workflow-routes.ts 22 -> 23` is the v1-IR pre-WIP fallback my #2621 added — a branch that runs only when the IR declares no columns and no nodes, so there is no role to resolve and `resolveLifecycleColumns` returns nothing. Justified, but it shipped with no marker and no baseline update, which is why `--strict` has been RED on main since it merged. #2652 fixes it at the site with a `DELIBERATE-LITERAL` marker rather than by bumping the baseline, and re-records the baseline: **854 -> 782 across 140 files**.

So if #2652 lands before your change, the 67 is already gone and item 2 becomes purely the mechanism (auto-lower / fail-with-instructions) rather than a cleanup. Worth sequencing that way — otherwise your re-record and mine will both touch the same file.

Two things I hit building the same kind of guard, in case they save you time:

1. **`--update` must refuse to RAISE.** My earlier version wrote current counts verbatim, so a developer who added a literal and ran the documented update command locked the regression in as the new ceiling. Lowering can be unattended; raising should be a hand edit with the reason recorded. That is the mirror of the high-water problem you are fixing.
2. **A marker must sit on the statement holding the comparison.** The census reads a node's LEADING comments — my first placement was one statement too high and silently did not register. `--strict` caught it, but only because I ran it.

PR #2631, opened 2026-07-30
TAKING cli/project.ts (fn project reported 0 running agents) + two test fixes — dashboard conversions WITHDRAWN in favour of #2626 and #2636

gsxdsm:
Three app-cluster conversions plus the evidence that they behave on a renamed AND a merged board.

## Per-file guard counts

| file | before | after | note |
|---|---|---|---|
| `packages/cli/src/commands/project.ts` | 0 | 0 | not a comparison site — see below |
| `packages/dashboard/app/components/TaskContextMenu.tsx` | 2 | 2 | **count does not move — deliberate, see below** |
| `packages/dashboard/app/components/Column.tsx` | 2 | 2 | **count does not move — deliberate, see below** |

**Read this before scoring the PR against the bar.** You said a claim that does not move your number is not done, so I am telling you up front that *this PR does not move it*, and why.

Both dashboard conversions are **fallback-preserving**:

```ts
const isIntakeColumn = columnFlags ? columnFlags.intake === true : column === "triage";
```

The literal survives as the no-flags branch, so the grep still counts it. That is the shape the sibling code already uses (`isPreExecutionHoldColumn`, same file, converted earlier in the program), and dropping the fallback would make an unresolved-column render *lose* the affordance a second way. What changes is the **behaviour when flags exist** — which is what the mutation results below measure.

If you want these to zero out the count, the fallback has to go, and that is a separate decision about whether an unresolved column should fail open or closed. Say the word and I will do it as a follow-up; I did not make that call unilaterally because it is not reversible from a rendering standpoint.

`cli/project.ts` was never a comparison site at all — it fed **raw rows** to `isRunningAgentTaskShape`, so the helper's own internal legacy fallback kicked in and `fn project` reported **0 running agents** on any renamed board. Fixed by resolving the IR per task before counting. Nothing to subtract.

## Two of the three had a test that looked like coverage and was not

- **`Column.tsx`** — the quick-create gate is `workflowMode || isIntakeColumn`. Every pre-existing intake case in `Column.test.tsx` *also* passes `workflowMode`, so the `||` short-circuited and **none of them ever reached the trait lookup**. Added cases that omit `workflowMode`, the only path where the conversion changes the answer.
- **`TaskContextMenu.tsx`** — the intake suppression was asserted only for the legacy `triage` id, the one board shape where a broken conversion still returns the right answer.

Mutation-verified rather than asserted:

| mutation | result |
|---|---|
| `isIntakeColumn` → `column === "triage"` | **2 of 88 fail** (exactly the renamed and merged cases) |
| menu suppression → `task.column !== "triage"` | **1 of 12 fail** |

## A pre-existing red I fixed on the way past

`uses VALID_TRANSITIONS and in-review back-to-progress labels` was **already failing on origin/main**. #2521 correctly moved the "Back to X" label onto the host's `columnLabel` function; this file's stub is `(column) => column`, so the hardcoded `"Back to In Progress"` expectation was left over from the pre-#2521 hardcode and nothing had updated it.

Matching the raw id would have made it pass while proving nothing, so instead that one case gets a display-like label function — the assertion now fails both if the "Back to" prefix regresses **and** if the label stops routing through `columnLabel`. Strengthened, not relaxed. Counts against completion criterion #2.

## Verification

- `Column.test.tsx` + `TaskContextMenu.test.tsx`: **100 passed**
- `tsc -p tsconfig.app.json` (the root config does not cover `app/`) and the CLI typecheck: clean
- `pnpm test:gate`: green

### DIFF
--- a/.changeset/fn-project-running-agents-renamed-board.md
+++ b/.changeset/fn-project-running-agents-renamed-board.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: `fn project` now counts running agents correctly on renamed workflow boards.
+category: fix
+dev: `runningAgentCount` fed raw task rows to `isRunningAgentTaskShape`, so its internal legacy column fallback applied and any board without the literal `in-progress`/`todo` ids reported 0. The command now resolves each task's workflow IR (cached per workflow) via `enrichRunningAgentTaskShape` before counting.
--- a/packages/cli/src/commands/project.ts
+++ b/packages/cli/src/commands/project.ts
@@ -24,6 +24,8 @@ import {
   COLUMN_LABELS,
   type Column,
   countRunningAgentTasks,
+  enrichRunningAgentTaskShape,
+  resolveWorkflowIrForTask,
   readProjectIdentity,
   writeProjectIdentity,
 } from "@fusion/core";
@@ -163,7 +165,23 @@ async function getTaskCounts(projectPath: string): Promise<TaskCountSummary> {
     for (const task of tasks) {
       counts[task.column] = (counts[task.column] || 0) + 1;
     }
-    return { byColumn: counts, runningAgentCount: countRunningAgentTasks(tasks) };
+    /*
+    FNXC:WorkflowLifecycleColumns 2026-07-30-12:20 (Phase B conversion — CLI project counts):
+    ENRICH before counting. `isRunningAgentTask` reads trait-derived fields and falls back to
+    the legacy `in-progress` / `in-review` literals when they are absent — so counting raw
+    rows reported ZERO running agents for a board whose wip column is renamed, in `fn project`
+    output an operator reads to decide whether the board is busy.
+
+    The dashboard's `project-store-resolver` already enriches for exactly this reason
+    (FN-8453). This was the remaining unenriched caller: same helper, same pure predicate, one
+    of two call sites doing it correctly. The `irCache` keeps it one IR read per workflow
+    rather than per task.
+    */
+    const irCache = new Map<string, Awaited<ReturnType<typeof resolveWorkflowIrForTask>>>();
+    const enriched = await Promise.all(tasks.map(async (task) =>
+      enrichRunningAgentTaskShape(task, await resolveWorkflowIrForTask(resolvedStore, task.id, irCache)),
+    ));
+    return { byColumn: counts, runningAgentCount: countRunningAgentTasks(enriched) };
   } catch {
     // Return empty counts if we can't read the project (not-found, corrupt
     // store, or lock-retry exhaustion — all fail soft here by design).
--- a/packages/core/src/__tests__/running-agent-count-requires-enrichment.test.ts
+++ b/packages/core/src/__tests__/running-agent-count-requires-enrichment.test.ts
@@ -1,0 +1,71 @@
+/*
+FNXC:WorkflowLifecycleColumns 2026-07-30-12:40:
+
+WHY AN UNENRICHED `countRunningAgentTasks` MISCOUNTS A RENAMED BOARD.
+
+`isRunningAgentTask` reads trait-derived fields (`columnCountsTowardWip`,
+`columnIsReviewOrMerge`, `columnTerminalKind`) and falls back to the legacy `in-progress` /
+`in-review` literals when they are ABSENT. So the same task list yields different counts
+depending on whether the caller enriched first — and on a renamed board the unenriched
+answer is zero.
+
+This pins the mechanism, which is what makes the CLI fix (packages/cli/src/commands/project.ts,
+`fn project` output) more than a plausible-looking edit: that caller passed raw rows while the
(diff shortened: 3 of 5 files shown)


gsxdsm:
## Added: the footer's queued count — and a correction to my own ledger

`columnIsIntakeOrHold` in `live-agent-count.ts` was the **last converted site in the program with no executable evidence**. My unproven-sites ledger listed it as unreachable because "its consumers are dashboard-side" — which is a statement about the *lane*, not about provability. It has exactly one consumer, `deriveStatsFromTasks`, and that is an exported pure function, so the narrow seam FN-5048 asks for was already sitting there.

That is the **fourth** time a lane-cost inference in that ledger has been wrong (after auto-merge-finalization, both self-healing rebounds, and `resolveFinalizeReboundColumn`). The rule I wrote after the third still holds and I still under-applied it: read what the function touches before costing a lane for it.

Pinned:

- a **renamed** hold lane counts as queued when flags are supplied
- the **U11 merged** intake+hold lane counts exactly **once** — both roles are true, so a per-role counter would double every card on the post-U11 default board
- a resolved non-intake lane does **not** count, so "counts everything" cannot pass
- the legacy no-flags fallback still counts, pinned **separately** so a conversion cannot delete both behaviours with only one test noticing

Mutation-verified: ignoring the resolved role and using the legacy id pair fails exactly the renamed and merged cases (2 of 5).

## One case asserts the wrong number on purpose

`live-agent-count.ts` admits in prose that a card in a column absent from the flag map *"is counted as neither running nor waiting, so the footer's queued total under-reports it"* — deliberately left unconverted, because deciding what an absent flag set means moves an operator-visible count either way.

The last case is a **characterization test** asserting that undercount as it exists today. The point is to make the admission executable instead of a comment, so it cannot drift further unnoticed. It is not an endorsement: if the fallback is ever converted, that expectation becomes `1` and this test is what tells you an operator-visible number moved.

Verification: 1837 tests green across `app/hooks/__tests__/`, `tsc -p tsconfig.app.json` clean.

gsxdsm:
## Withdrawn: my two dashboard conversions. #2626 and #2636 are both better, and mine was wrong.

This PR originally converted `TaskContextMenu.tsx` and `Column.tsx`. Both collided with PRs that had done the harder analysis, and in both cases **their answer is right and mine was not**. Dropped rather than rebased, and recording why, because the errors are the reusable part.

### `TaskContextMenu.tsx` — mine would have shipped an operator-visible regression

I converted the actions-menu suppression to `flags.intake !== true`. **#2626** resolved it to `intake AND NOT hold` — a *pure* intake lane — on the reasoning that a merged Planning column carries both traits, so cards rest there **waiting for capacity** and have real actions worth showing.

That is correct and mine is not. On the post-U11 default board, Planning carries `intake`, so my version **suppresses the actions menu for every card on the default board's Planning column**. Worse, my own merged-board test case asserted that suppression *as the expected behaviour* — a test that would have locked in the regression. Resolving to the role is not sufficient when the role itself is ambiguous on a merged board; #2626 found the disambiguating fact (`hold`) and I did not.

### `Column.tsx` — mine converted code that production never runs

I converted the quick-create gate's `|| column === "triage"` arm to a trait lookup. **#2636** established that the arm is **unreachable**: `Board.tsx` is its only consumer and passes `workflowMode` at all three render sites, so the `||` short-circuits every time. They delete it.

Deleting dead code beats converting it. My accompanying tests deliberately omitted `workflowMode` to reach the arm — which means they exercised a path production cannot take, while reading as coverage. That is the same "test passes without exercising its subject" failure this program keeps hitting, authored by me this time.

## What remains in this PR — three things nobody else is doing

1. **`cli/project.ts`** — never a comparison site at all. It fed **raw rows** to `isRunningAgentTaskShape`, so the helper's internal legacy fallback applied and `fn project` reported **0 running agents** on any renamed board. Fixed by resolving the IR per task before counting. Does not move the guard count; there was nothing to subtract.

2. **The footer's queued count** — `columnIsIntakeOrHold` was the last converted site with no executable evidence, and my ledger wrongly called it unreachable. Renamed and merged boards pinned, mutation-verified (legacy-pair mutation fails exactly those 2 of 5), plus a characterization test making the undercount the source admits in prose executable.

3. **The #2521 stale expectation**, red on main since it landed — fixed in the strengthening direction rather than by matching the raw id.

No file in this PR is claimed by another open PR.

gsxdsm:
Valid P1 — fixed. The `fn project` change alters published `@runfusion/fusion` behaviour, so AGENTS.md requires a changeset.

Added `.changeset/fn-project-running-agents-renamed-board.md` as a **patch** in the labeled format:

- **summary** (user-facing): ``fn project`` now counts running agents correctly on renamed workflow boards.
- **category:** `fix`
- **dev:** the root cause — `runningAgentCount` fed raw task rows to `isRunningAgentTaskShape`, so its internal legacy column fallback applied and any board without the literal `in-progress`/`todo` ids reported **0**. It now resolves each task's workflow IR (cached per workflow) before counting.

`pnpm check:changesets` passes. The other two commits in this PR are test-only and correctly carry no changeset.

PR #2632, opened 2026-07-30
E2E evidence: the MERGED board (third completion criterion) — plus a RETRACTION of my #2613 escalation

gsxdsm:
This is the merged-board half of the evidence assignment. **It is red, deliberately, and the red is the finding.** Do not merge it to make the red go away — the assertions are correct and `main` is broken.

## Escalation first: #2613 broke the default board, and the gate did not notice

`6a33d8f8c` — *"Phase B — TAKING task-creation.ts: intake classification by trait (4 sites → 0)"* (#2613) — regressed four E2E cases, including **the default-vocabulary full lifecycle**, which is scenario 1 of the whole E2E assignment.

Attribution is a clean single-file revert, not a guess:

```
HEAD (main):                                        4 failed | 39 passed
HEAD with ONLY 6a33d8f8c's task-creation.ts reverted:    28 passed  (both files fully green)
```

Failing:

1. `scenario 1 — DEFAULT vocabulary … persists the card in the expected column at every stage`
2. `scenario 2 — RENAMED vocabulary … writes the same column-transition audit trail as the default`
3. `releases a card out of the merged lane on capacity — the release is not a self-move`
4. `does not re-release a card that already left the merged lane`

**`pnpm test:gate` is green on this branch — exit 0, 695 tests.** #2613 merged through a green gate, and its own tests pass. This is the eighth time this program a test has passed without exercising its subject, and the first one an E2E family caught rather than review.

### Mechanism

`isIntakeColumn` in `task-creation.ts` decides whether a new card gets a **bootstrap** prompt (freeform, "triage will plan this later") or a **specified** prompt (planned, executable). #2613 rewrote it as:

```ts
const isIntakeColumn = (intakeFacts.intake !== undefined && task.column === intakeFacts.intake) || …
```

where `intakeFacts.intake` falls back to the **default workflow's** intake when the create supplies no `workflowId`. Post-U11 the default workflow's intake **is `todo`**. So any card created directly in `todo` is now classified as intake and gets a bootstrap prompt — unplanned. Unplanned cards do not advance through the graph (no `NodeEntered` audit rows → failures 1 and 2) and hold-release will not release them (FN-7648: no unplanned card enters a processing column → failures 3 and 4).

Before U11 this was safe: `triage` was intake and `todo` was a distinct lane, so creating in `todo` meant "planned work". The merge deleted that distinction.

### Why this is the exact trap you warned about

You said you did not want *"a conversion that swaps the literal for a trait lookup WITHOUT checking what the guard was for."* The old `task.column === "triage"` guard meant **"is this card unplanned?"** On a merged board, intake-vs-hold **cannot answer that question at all** — one column is both. The distinguishing fact is not the column; it is whether the caller supplied a spec. Resolving the role faithfully still gets the wrong answer, because the question was never really about the column.

Not fixing it from here: `task-creation.ts` is #2613's owner's file, and the fix is a design call about which fact replaces the column test.

## What the evidence itself adds

Three families extended to the U11 shape — one column carrying **both** intake and hold. That breaks a class of guard renamed boards structurally cannot reveal:

| shape | consequence |
|---|---|
| `intake && !hold` | **unsatisfiable** — silent |
| hold → intake release | **self-move**, re-fires every poll — loops |
| `intake && column !== "triage"` | inverts to **always-true** — silent |

Two are silent and one loops, so every case sweeps **twice** and asserts no re-release; a single pass cannot tell a no-op from a self-move.

### A fixture that could not fail

My first merged row used `MERGED_VOCAB`, which is *faithful* to U11 — it reuses the legacy ids, because that is what the default lineage has. That fidelity **destroyed its discriminating power**: its hold column *is* `todo`, so a guard falling back to the `todo` literal returns the same answer as one resolving the role. The "hold but not intake" mutation left all 23 green.

Added `MERGED_RENAMED_VOCAB` — merged *structure*, renamed *vocabulary* — the only combination where the collapse is observable **and** the literal is wrong. Same mutation now fails exactly 1 of 23. Both vocabularies stay: one asks *"does the collapse break the release path"*, the other *"is the role actually resolved"*.

One rebound mutation was **genuinely unobservable** rather than undetected — `hold` is also the first column in that fixture, so the fallback chain lands there regardless. Pointing rebound at `complete` instead fails 9 of 15. Recorded rather than papered over.

## For the CAPACITY worker before `self-healing.ts` is marked done

`self-healing.ts:2952` and `:9134` query `listTasks({ column: "triage" })`. Converting the 10 guards leaves those sweeps **blind** — they never see a renamed card, so the guard is correct and unreachable. Query and guard convert together or not at all.

There are **137** such `column: "<legacy id>"` sites repo-wide, 52 in that one file, and the 45→0 grep counts none of them because they are object properties, not comparisons. **The bar can reach zero with sweeps still unable to fire.**

### DIFF
--- a/packages/engine/src/__tests__/_workflow-vocabulary-fixture.ts
+++ b/packages/engine/src/__tests__/_workflow-vocabulary-fixture.ts
@@ -18,6 +18,22 @@ export const HOLD_STALENESS_MS = 60 * 60_000;
 
 /** The four lifecycle roles this program's guards are supposed to resolve by TRAIT, not by id. */
 export interface Vocabulary {
+  /*
+  FNXC:WorkflowLifecycleColumns 2026-07-29-15:45 (fixture correction):
+  Before this field `lifecycleIr` emitted NO intake trait for a non-merged board, so the
+  SEPARATE-LANES shape this fixture is supposed to provide did not exist. Every renamed-versus-
+  merged differential built on it compared "no intake" against "intake and hold on one column",
+  not "two distinct lanes" against "one".
+
+  That let assertions pass for the wrong reason: `expect(lifecycle?.intake).not.toBe(hold)` is
+  vacuously true when `intake` is `undefined`, so it would pass against a resolver that never
+  resolved intake at all. `undefined` is no longer reachable from this fixture.
+
+  `mergedIntakeAndHold` remains the way a caller forces the merged shape; a vocabulary whose
+  `intake` equals its `hold` now implies it too, so the merged vocabularies cannot silently lose
+  their intake trait to a forgotten option.
+  */
+  readonly intake: string;
   readonly hold: string;
   readonly wip: string;
   readonly review: string;
@@ -26,6 +42,7 @@ export interface Vocabulary {
 
 /** The legacy ids. A guard keyed on a string literal passes here for the wrong reason. */
 export const DEFAULT_VOCAB: Vocabulary = {
+  intake: "triage",
   hold: "todo",
   wip: "in-progress",
   review: "in-review",
@@ -34,6 +51,7 @@ export const DEFAULT_VOCAB: Vocabulary = {
 
 /** No id overlaps the legacy enum. A guard keyed on a string literal goes silent here. */
 export const RENAMED_VOCAB: Vocabulary = {
+  intake: "inbox",
   hold: "backlog",
   wip: "building",
   review: "checking",
@@ -92,6 +110,7 @@ export const MERGED_VOCAB: Vocabulary = {
 /** A merged board that ALSO renames: both variables move at once, which is the shape a
  *  custom workflow author actually produces. */
 export const MERGED_RENAMED_VOCAB: Vocabulary = {
+  intake: "planning",
   hold: "planning",
   wip: "building",
   review: "checking",
@@ -99,16 +118,22 @@ export const MERGED_RENAMED_VOCAB: Vocabulary = {
 };
 
 export function lifecycleIr(v: Vocabulary, id: string, options: LifecycleIrOptions = {}): WorkflowIr {
+  /* Merged when the caller says so OR when the vocabulary collapses the two roles onto one id. */
+  const merged = options.mergedIntakeAndHold === true || v.intake === v.hold;
   return {
     version: "v2",
     id,
     name: `lifecycle-${id}`,
     columns: [
+      /* The SEPARATE intake lane, for a non-merged vocabulary only. On a merged one
+         `v.intake === v.hold`, so declaring it here would duplicate the id — the hold column
+         carries both traits instead. */
+      ...(merged ? [] : [{ id: v.intake, name: "Intake", traits: [{ trait: "intake" as const }] }]),
       {
         id: v.hold,
         name: "Hold",
         traits: [
-          ...(options.mergedIntakeAndHold ? [{ trait: "intake" }] : []),
+          ...(merged ? [{ trait: "intake" }] : []),
           { trait: "hold", config: { release: "capacity" } },
         ],
         /* U4 workflow-declared recovery policy (#2478). Declared on the HOLD column of both
--- a/packages/engine/src/__tests__/workflow-merged-board-live-e2e.pg.test.ts
+++ b/packages/engine/src/__tests__/workflow-merged-board-live-e2e.pg.test.ts
@@ -1,0 +1,188 @@
+/*
+FNXC:WorkflowLifecycleColumns 2026-07-30-16:30 (E2E evidence — the MERGED board):
+
+THE BOARD THE PROGRAM ACTUALLY SHIPPED, and the one no E2E family covered.
+
+Every suite in this directory proves the RENAMED case: a board whose column ids differ from
+the legacy enum. U11 shipped something structurally different — it merged Todo into Planning,
+so on the default lineage ONE column carries BOTH the intake and hold traits and the id
+`triage` no longer exists.
+
+A renamed differential cannot see that, because renamed boards still have two separate
+columns. The merged shape breaks a distinct class of guard:
+
+  - "is this card in intake but NOT in hold" is UNSATISFIABLE — one column, both traits;
+  - a hold -> intake release is a SELF-MOVE, which a move guard may reject or loop on;
+  - `intake && column !== "triage"` inverts from sometimes-true to ALWAYS-true, so an
+    affordance gated on it appears everywhere instead of nowhere.
+
+The first and third are silent. The second is the interesting one, because it is the shape
+that produces a loop rather than a no-op.
+
+WHAT IS REAL: a PostgreSQL TaskStore, a real persisted workflow whose hold column carries both
+traits, the real capacity release sweep, and the real graph column boundary. Assertions read
+the persisted row.
+*/
+import { beforeAll, beforeEach, afterEach, afterAll, describe, expect, it } from "vitest";
+import "@fusion/core";
+import { resolveLifecycleColumns, resolveWorkflowIrForTask } from "@fusion/core";
+
+import {
+  pgDescribe,
+  createSharedPgTaskStoreTestHarness,
+  type SharedPgTaskStoreHarness,
+} from "../../../core/src/__test-utils__/pg-test-harness.js";
+import { runHoldReleaseSweep } from "../hold-release.js";
+import { MERGED_VOCAB, RENAMED_VOCAB, lifecycleIr } from "./_workflow-vocabulary-fixture.js";
+
+pgDescribe("live MERGED-board E2E: one column carrying both intake and hold", () => {
+  const h: SharedPgTaskStoreHarness = createSharedPgTaskStoreTestHarness({
+    prefix: "fusion_merged_board_e2e",
+  });
+
+  beforeAll(h.beforeAll);
+  beforeEach(h.beforeEach);
+  afterEach(h.afterEach);
+  afterAll(h.afterAll);
+
+  /** A workflow whose planning lane carries BOTH intake and hold — the U11 shape. */
+  async function seedMergedWorkflow(key: string): Promise<string> {
+    const created = await h.store().createWorkflowDefinition({
+      name: `Merged ${key}`,
+      kind: "workflow",
+      ir: lifecycleIr(MERGED_VOCAB, `custom:merged-${key}`, { mergedIntakeAndHold: true }),
+    } as never);
+    return (created as { id: string }).id;
+  }
+
+  async function seedTask(taskId: string, column: string, workflowId: string): Promise<void> {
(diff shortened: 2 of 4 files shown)


gsxdsm:
## Added: a defect in my own fixture that invalidated one of this PR's assertions

Disclosing this against my own evidence rather than leaving it for review to find.

`Vocabulary` had **no `intake` field**, so `lifecycleIr` emitted no intake trait at all for a non-merged board. The "separate lanes" shape this fixture exists to provide **never existed** — every renamed-vs-merged differential built on it was comparing *no intake* against *intake and hold on one column*, not *two lanes* against *one*.

That made this PR's separate-lanes case pass for the wrong reason:

```ts
expect(lifecycle?.intake).not.toBe(lifecycle?.hold);   // vacuously true when intake is undefined
```

It would have passed against a resolver that never resolved intake at all — the exact failure mode this whole assignment exists to catch. Found only because the new planner-lane family's renamed case returned one lane instead of two and nothing else explained it.

Fixed by declaring a real intake column, so `undefined` is no longer reachable from this fixture. The vocabulary is now authoritative for the merged shape (`v.intake === v.hold`) instead of depending on the caller passing `mergedIntake`. `isMergedVocabulary` was wrong the same way — it compared against the default lineage's id, so it returned `false` for `MERGED_RENAMED_VOCAB`, the one vocabulary that most needs it.

## Also added: live evidence for `planner-lane-resolution.ts`

The one merged-board-aware conversion in the program that got the hard case **right** — worth pinning precisely because its merged answer is `[]`, which is the shape of a guard that no longer guards, and because its correctness was proven only against a mocked IR.

- merged yields `[]` — asserted empty **and** defined **and** not the merged column (`undefined` would make both consumers fall back to the legacy literal list via `?? LEGACY`)
- renamed still yields its intake lane, so "always `[]`" cannot pass
- `resolvePlannerLanesForTask` gives 2 lanes renamed, 1 merged, never 0 — zero would strand every card waiting to be planned
- `[]` survives the real guard, and the same column with a planner **status** gives the opposite verdict — the distinction the resolver's own comment cites as its reason for returning `[]`

Mutation-verified both ways: dropping the merged branch fails 1 of 5; returning `[]` unconditionally also fails 1 of 5.

I had the guard's polarity wrong at first — it only returns true for a card with real preserved progress, so my initial cases were reading a `false` that had nothing to do with lanes. Recorded in the fixture comment rather than quietly corrected.

## Status

**44 of 48 passing across four families. The 4 red are still #2613's regression, unchanged by these commits** — including the default-vocabulary full lifecycle. `pnpm test:gate` remains green at 695 tests, which is the other half of the finding.

gsxdsm:
## #2613 blast radius — complete, measured across every live E2E family

Ran all ten families so the scope of the regression is bounded rather than estimated.

| family | tests | result |
|---|---|---|
| `workflow-lifecycle` | 23 | **RED (2)** |
| `workflow-merged-board` | 5 | **RED (2)** |
| `workflow-agent-count` | 13 | green |
| `workflow-agent-link` | 5 | green |
| `workflow-lease-rebound` | 6 | green |
| `workflow-merge-family` | 7 | green |
| `workflow-merge-rebound` | 4 | green |
| `workflow-merge-safeguards` | 6 | green |
| `workflow-planner-lane` | 5 | green |
| `workflow-rebound-family` | 15 | green |

**89 tests, 4 failed, confined to 2 families.** The damage is limited to the two paths that depend on a newly-created card being *planned*: graph advancement out of the hold lane, and capacity release out of it. Merge, lease, agent-link, and agent-count paths are untouched — they operate on cards that are already past creation.

So the fix is narrow. It does not need to touch anything the other eight families cover.

gsxdsm:
# RETRACTION — #2613 is not a regression. My escalation was wrong.

Correcting this at the top because I filed it loudly and it named another worker's merged PR as breaking the default board.

## What I got wrong

I bisected the four failures to `6a33d8f8c`'s `task-creation.ts`: reverting that one file took 4 failed / 39 passed → 28 passed. I reported that as proof of regression.

**A revert proves which change moved an outcome. It says nothing about whether the old behaviour was correct.** I never checked the second thing, and the check was cheap.

## The actual cause, verified by the opposite experiment

Giving the seeded card a planned `PROMPT.md` re-greens **all five** merged-board cases with **#2613 fully intact**.

FN-7648's `isUnplannedForExecution` reads `PROMPT.md` for any card resting in an intake- **or** hold-trait column and refuses to move an unplanned card into a processing column. On a merged board the hold lane **is** the intake lane, so every card this suite seeds is subject to that gate. My fixture wrote no spec, so the card carried a bootstrap seed and was correctly held.

**Held was the gate working.** #2613 made a card created in `todo` classify as intake — which post-U11 it genuinely is — so it correctly receives a bootstrap seed instead of a specified prompt. My fixture had been silently depending on the pre-U11 semantics where `todo` was not intake and a card created there counted as already specified.

Credit where it belongs: **#2634** diagnosed this properly, by probing the release gates the sweep actually consults and killing two hypotheses on the way (capacity settings changed nothing; a direct `moveTask` succeeded). That is the method I should have used instead of inferring a verdict from a bisect.

The claim I made that *did* hold up — that `pnpm test:gate` is green at 695 tests while these families were red — is now simply a statement that pg E2E families are not in the gate. It is not evidence of an escaped regression, and I withdraw that framing too.

## Consequences for this PR

- Fixed in `workflow-merged-board-live-e2e` (my file): the seed now writes a planned spec. **25 of 25 green** across merged-board, planner-lane, and rebound.
- The lifecycle family and the shared fixture are **#2634's** — it is already in both. Not duplicating that fix.
- **Fixture collision to sequence:** #2634 adds `mergedIntakeAndHold`/`reviewRework` to `LifecycleIrOptions`; this PR adds an `intake` field to `Vocabulary` plus `mergedIntake`. These overlap and whoever lands second must reconcile. Flagged on #2634 with the reason my version exists, so the finding behind it is not lost in the merge — see next section.

## The finding that must survive the reconciliation

`Vocabulary` had **no `intake` field**, so `lifecycleIr` emitted no intake trait for a non-merged board. The "separate lanes" shape the fixture is supposed to provide never existed, and this PR's own separate-lanes assertion passed **vacuously on `undefined`**:

```ts
expect(lifecycle?.intake).not.toBe(lifecycle?.hold);   // true when intake is undefined
```

Whatever the merged-shape option ends up being called, the fixture must declare a real intake column, or that assertion goes back to passing against a resolver that never resolves intake.

gsxdsm:
## Rebuilt onto current main after #2634 landed — was CONFLICTING, now one commit

Rebuilt rather than merged. #2634 landed its own merged-shape option in the shared fixture, and resolving a three-way conflict there invites exactly the silent drift the fixture exists to prevent. **#2634's `mergedIntakeAndHold` is the spelling everywhere now**; my older `mergedIntake` is gone, as I said it would be when I flagged the collision.

## The defect that did not get absorbed, and why it needed two fixes

`Vocabulary` still had no `intake` field on main, so `lifecycleIr` emits **no intake trait at all** for a non-merged board. The separate-lanes shape the fixture is supposed to provide does not exist — every renamed-versus-merged differential compares *no intake* against *intake+hold on one column*.

Declaring the intake column was only half of it. The assertion was **also** weak, and I only found that because the mutation didn't kill what I expected:

```ts
expect(lifecycle?.intake).not.toBe(lifecycle?.hold);   // vacuous when intake is undefined
```

Removing the intake lane left that green. Asserting the id **positively** is what makes it bite. So the fixture fix alone would have looked like a fix and proven nothing — which is the failure mode of this entire program, reproduced inside my own correction for it.

Mutation result — removing the separate intake lane now kills **3 of 10**:

| suite | case |
|---|---|
| merged-board | a RENAMED board with separate lanes still behaves |
| planner-lane | still yields the intake lane for a RENAMED lineage |
| planner-lane | reports BOTH lanes renamed, and ONE on a merged board |

## Dropped: my ledger-comment update to `workflow-lifecycle-live-e2e`

Its patch applied **cleanly**, and that was the trap. Being a diff from before #2634, applying it silently **reverted that PR's planned-spec seed fix** — the two lifecycle failures came straight back. Caught it by running the suite rather than trusting `git apply` exiting 0.

Dropped entirely rather than clobber a landed fix for a comment change. The ledger corrections it carried are recorded in this PR's earlier comments; if they're wanted in-tree, that is a separate comment-only commit against main's current version of that file, and it is not worth risking the same collision to land.

**49 of 49 green** across merged-board, planner-lane, lifecycle, and rebound. Engine typecheck clean.

gsxdsm:
## Finding: the completion bar measures one sixth of the defect class, and a converted rebound is already dead because of it

Found while closing the last entry in this PR's unproven-sites ledger. Putting it here rather than in a message because it changes what "zero" means.

### The concrete instance

`executor.ts:4305` — `performWorkflowRerunBounce`:

```ts
if (latestTask.column === "in-progress" || latestTask.column === "in-review") {
  await this.store.moveTask(taskId, await resolveReboundColumnFor(this.store, taskId), { ... });
}
```

The rebound **target** is resolved by role. The **gate** deciding whether to rebound at all is still the legacy `in-progress`/`in-review` pair. On a renamed board (`building`/`checking`) the condition is false, the bounce never fires, and the resolved target is unreachable.

That is a converted site whose conversion cannot execute — the exact "guard that no longer guards" failure this program exists to remove, produced *by* a partial conversion. Neither a `=== "triage"` grep nor a passing suite shows it.

### The measurement gap

The bar counts `=== "triage"` comparisons. That is **one of six** legacy lifecycle ids, and by far the smallest:

| legacy id | comparisons on `origin/main` (non-test) |
|---|---|
| `triage` | **81** ← the only one counted |
| `todo` | 70 |
| `in-progress` | 121 |
| `in-review` | 188 |
| `done` | 140 |
| `archived` | 127 |

**~646 same-class comparisons, of which the bar sees 81.** Concentrations: `self-healing.ts` 60, `executor.ts` 37, `TaskCard.tsx` 29, `TaskDetailModal.tsx` 15, `register-task-workflow-routes.ts` 13, `scheduler.ts` 11.

This is separate from — and larger than — the 137 `column: "<legacy id>"` **query-filter** sites I flagged earlier, which the grep also misses because they are object properties rather than comparisons.

So `triage → 0` is reachable with the lifecycle still keyed to legacy ids nearly everywhere, and with at least one converted rebound provably dead. I am not proposing scope expansion — that is the program's call. I am reporting that the number being tracked does not mean what the completion criterion says it means, and that the AST census (#2633) is the right instrument to measure the real figure if you want it.

### Also correcting my own ledger

It listed executor.ts as having ~2 rebound sites. It has **33**, all funnelling through `resolveReboundColumnFor` — a 3-line wrapper over `resolveReboundTarget`, which this PR's rebound family already proves on renamed *and* merged boards. So the resolver is proven and the 33 sites demonstrably call it; the residual risk is gates like the one above, not the resolution.

PR #2633, opened 2026-07-30
TAKING scripts/: parse instead of grep — an AST classifier for the lifecycle-column bar, cross-checked by a second implementation

gsxdsm:
The program's completion bar is "`column === "triage"` reaches zero". This measures what that bar actually covers, and checks the measurement in so it cannot drift.

## The number, measured by the checked-in tool

```
lifecycle-column-census: scanned 1956 source files

  COLUMN guards (the backlog):   1031
  ROLE comparisons (not guards):   10
  DELIBERATE-LITERAL (reviewed):    4

  by column id:
     313  done
     217  in-review
     201  in-progress
     177  archived
      83  todo
      40  triage

  top files:
     151  packages/engine/src/executor.ts
     136  packages/engine/src/self-healing.ts
      50  packages/dashboard/app/components/TaskCard.tsx
      44  packages/core/src/task-store/moves.ts
      34  packages/dashboard/app/components/TaskDetailModal.tsx
```

**`triage` is under 4% of the class.** Every one of those 1031 sites is the same defect: a lifecycle decision made by column NAME, which stops matching the moment a board renames a column. The bar can be met in full while 991 identical guards remain — and two files hold a quarter of them.

## The tracked count is wrong in three directions at once

Each of these cost real work this week, which is why this is a PR and not a comment.

1. **Vocabulary.** It measures one of six legacy ids.
2. **Receiver.** It is anchored on locals named `column`/`toColumn`/`fromColumn`, so it never saw the three real guards in `executor.ts` written against `from` and `originColumn`. One of those meant completed-but-stranded work was never recovered on a renamed board, with nothing else owning that state (converted in #2628).
3. **Collision.** `role === "triage"`, `agentType === "triage"`, `entry.agent === "triage"` compare an **AGENT ROLE**. The planner *lane* is named `triage` and keeps that name — U11 removed the *column*. Ten such sites were counted as backlog, and the "obvious" fix (renaming the role) silently empties the planner's prompt template and mis-binds its model markers.

A count that is too high and too low simultaneously sends work to the wrong files while hiding the files that need it. So the census reports **three separate numbers** and never nets them.

## Proven to fail on the original defect

Not asserted — exercised:

```
$ # reintroduce `task.column === "triage" || task.column === "todo"` into live-agent-count.ts
$ node scripts/lifecycle-column-census.mjs --strict; echo "exit=$?"
  packages/core/src/live-agent-count.ts: 10 -> 12
exit=1
$ # restore the file
$ node scripts/lifecycle-column-census.mjs --strict >/dev/null; echo "exit=$?"
exit=0
```

The CLI also exits 1 when its own file list comes back empty — a guard that reports success without checking anything is worse than no guard.

## 12 regression cases, split by what they defend

Must catch: all six ids; a guard on a local named `from`/`originColumn` (verbatim the executor.ts shape); single quotes; negation; several comparisons on one line.

Must **not** catch: role comparisons; comment prose (two tracked "guards" in `replan-target.ts` were prose about a filter that lives in another file); a trailing `// … === "triage"` on a code line; sites carrying a `DELIBERATE-LITERAL` marker.

Plus: **one marker cannot launder a distant guard in the same file** — that is how allowlists rot.

## Report-only, deliberately

`--strict` compares per-file counts against `scripts/lib/lifecycle-column-census-baseline.json` and fails when any file's count **rises**. It is **not** wired into the merge gate: a thousand-site backlog cannot be a blocking check the day it is first measured, and a guard nobody can pass is a guard everyone disables. Owners tightening their own area re-record the baseline in the PR that lowers it. This is the ratchet shape the `DELIBERATE-LITERAL` markers scattered through the program already anticipate.

## Stated limitation

Classification is by receiver **name**, so a future field named `agent` that holds a column would be misclassified as a role comparison. Recorded at the site, and it is precisely why the two classes are reported separately instead of netted into one figure.

## Verification

- 12/12 new cases (`packages/engine/src/__tests__/lifecycle-column-census.test.ts`)
- `pnpm test:gate` **71/71**; `pnpm lint` clean
- `pnpm census:lifecycle-columns`, `--json`, and `--strict` all exercised end to end
- documented in `docs/testing.md`; no production code touched


gsxdsm:
Both P1s fixed and pushed — and the first one had a worse consequence than the note I had left beside it.

**1. Comment stripping shifted every following line.** I had written this down as "over-counts, visible in the report". Wrong direction. Deleting a multi-line block comment moved the lines after it, so the site-local `DELIBERATE-LITERAL` lookup ran at the wrong offset — and in `replan-target.ts` **one marker near the top laundered four unrelated live guards** (`task.column === "in-progress" | "in-review" | "done" | "archived"` in `hasAdvancedPastPlanning`). They were reported as reviewed-and-intentional when they are neither. Findings also pointed at unrelated source lines, which sends a reader to the wrong code. Took your suggestion — blank in place, keep the newlines.

Measured effect: raw hits unchanged at 1045, but classification moves — deliberate 4 → 3, column 1031 → 1032. A census whose totals look stable while its classification is wrong is the worst kind, so this is exactly the bug worth catching.

**2. A stale baseline is a hole, not a conservative default.** Right, and it defeats the purpose: a file going 10 → 3 with no baseline edit keeps allowing 10, so seven guards can return later with `--strict` still green. It now fails on a **drop** as well as a rise, names the stale files, and points at `--strict --update-baseline`. Baseline entries for files that no longer exist are reported the same way — a deleted file leaving its allowance behind is the same hole. Exercised end to end: lowering `live-agent-count.ts`’s count makes `--strict` exit 1 with "baseline is STALE"; restoring returns exit 0.

**A third, found while verifying those two.** My role classifier was nominal, so it was guesswork — and it was already wrong: `skill-resolver.ts` compares `sessionPurpose` and `tool-availability.ts` compares `surface`, both against role names, and both scored as COLUMN guards. Variable names are unbounded; the vocabulary is not. `AgentRole` is `triage | executor | reviewer | merger`, and three of those four are never column ids, so an expression compared against `"executor"`/`"reviewer"`/`"merger"` nearby is matching roles whatever it is called. Structural, and it generalises to receivers nobody has named yet. Measured: role 10 → 12, column 1032 → 1030, and those two engine files leave the backlog — converting them would have asked "which column carries the intake trait" about something that is not a column.

Revert proofs, isolated:
- `stripComments` back to deleting comments → **2 of 19 fail** (the line number, and the laundering case)
- role-only-sibling signal removed → the two engine files return to `column`
- stale-baseline branch removed → a lowered file no longer fails `--strict`

Four new cases pin the vocabulary signal including the two that must NOT flip: a genuine column guard sitting *next to* role code stays a column guard (the signal is the receiver, not proximity — otherwise one nearby role check launders everything around it), and `cli/task.ts`’s `col === "triage"` board dots stay in the backlog despite an unfamiliar receiver name.

Verified: 19/19, `pnpm test:gate` 71/71, `pnpm lint` clean, baseline re-recorded in the same commit (column 1030, role 12, deliberate 3).

gsxdsm:
## What this measures for the `triage` bar specifically

Running the tool restricted to the `triage` literal, on `origin/main` as of this branch — the same instrument that gates, reporting the count it enforces:

```
TRIAGE ONLY:  column 38  |  role 12  |  deliberate 0
```

| count | file |
|---:|---|
| 11 | `packages/engine/src/self-healing.ts` |
| 5 | `packages/core/src/task-store/moves.ts` |
| 4 | `packages/core/src/default-workflow-hooks.ts` |
| 3 | `packages/engine/src/executor.ts` |
| 2 | `live-agent-count.ts`, `Column.tsx`, `TaskCard.tsx`, `TaskContextMenu.tsx`, `TaskDetailModal.tsx` |
| 1 | `cli/commands/task.ts`, `comments-ops.ts`, `DocumentsView.tsx`, `MissionControlPanel.tsx`, `taskActivity.ts` |

Two differences from the hand classification worth acting on, both machine-checkable:

**`usage-limit-detector.ts` (2) is ROLE, not column.** Both sites are `agentType === "triage"` — one of the four `AgentRole` values. Converting them would ask "which column carries the intake trait" about something that is not a column, which is precisely the hazard the hand list warns about for `agent-prompts.ts`. Same for `effective-model-resolution.ts:148` and `useTasks.ts:162` (`entry.agent`).

**`skill-resolver.ts` and `tool-availability.ts` are ROLE too**, and this took a code change to see: they compare `sessionPurpose` and `surface`, not `.column`, so a name-based classifier — including mine, until this push — scored them as column guards. The signal that fixes it is structural rather than nominal: `AgentRole` is `triage | executor | reviewer | merger`, and three of those four are never column ids, so an expression compared against `"executor"`/`"reviewer"`/`"merger"` nearby is matching roles whatever it is called. That generalises to receivers nobody has named yet, which a name list cannot.

So the machine-measured role set is **12**, not 6 — and every one of those 12 is code that must be left exactly as it is.

Comments need no allowlist: the census strips them before matching, which is why the four FNXC-prose sites do not appear above. And `agent-prompts.ts` no longer shows up at all — #2628 gives that role a name (`PLANNER_AGENT_ROLE`), which retires the class permanently instead of asking every future reader to re-derive it.

**Already converted and in flight, not yet merged:** `live-agent-count.ts` (2), `DocumentsView.tsx` (1), `executor.ts` (3) and `replan-target.ts` in #2628; `default-workflow-hooks.ts` (4) plus the flag-ON half of `moves.ts` in the same PR. Only merged conversions count, so the number above will move when those land — that is exactly the reason to gate with the tool that reports it.

gsxdsm:
## A third colliding vocabulary, found while picking up the next file — 182 sites

`packages/cli/src/commands/task.ts` reported column guards on lines that turned out to be `step.status === "done"`. `StepStatus` is `pending | in-progress | done | skipped`, and missions, goals and features carry their own `done`/`archived` statuses — three names that collide with column ids.

```
  COLUMN guards (the backlog):     848      (was 1030)
  ROLE comparisons (not guards):    12
  STATUS comparisons (not guards): 182      <- new class
  DELIBERATE-LITERAL (reviewed):     3

  by literal, of the status hits: done 105 (of 313) | in-progress 49 (of 201) | archived 28 (of 177)
```

This matters beyond arithmetic: converting one is a **category error**. Asking "which column carries the `complete` trait" about a *step*’s status would stop the step reading as finished, and a reviewer sent to `goal.status === "archived"` by a guard census has been sent to the wrong code entirely.

Recognised the same two ways as roles, and the structural half is the one that generalises: `pending` and `skipped` are `StepStatus` members and never column ids, so an expression matched against either is a status whatever it is called. The receiver-name half (`/status/i`) catches single-comparison sites where no sibling value appears. Names alone were already wrong twice this week (`sessionPurpose`, `surface`), which is why neither classifier leans on a name list.

**A detail worth recording:** the inflation comes *entirely* from the three names the two vocabularies share. `step.status === "skipped"` was never counted, because `skipped` is not a column id — so the census only ever confused the overlap. Same shape as `triage` belonging to both the column and the role vocabulary. Pinned in a test, so a future widening of the id list cannot silently change what the overlap means.

Revert proof: dropping the status classifier fails **2 of 23**. Two paired cases pass under the revert — a real column guard sitting *next to* status code stays backlog, and a mixed line (`toColumn === "in-progress" && status === "pending"`) keeps its column half — so "exclude anything near a status" cannot pass for this.

Baseline re-recorded in the same commit, per the stale-baseline rule this PR adds. 23/23, lint clean, `--strict` exits 0.

**For the `triage` bar specifically this changes nothing** — it was already 38 column / 12 role / 0 status, because `triage` collides with the role vocabulary rather than the status one. The correction is entirely in `done`/`in-progress`/`archived`, i.e. in what the *next* phase of this program would otherwise have been sized against.

gsxdsm:
Ran this against an independent classifier I built for the same question. **Two methods, two implementations, and the ROLE bucket agrees exactly: 12.** Same twelve sites — `role` ×4, `entry.agent` ×3, `agentType` ×2, `agent`, `sessionPurpose`, `surface`. That's worth stating plainly, because the hand-maintained list currently in circulation puts six of those twelve in the "real column guards" column, and a ratchet built from it would demand converting `agentType === "triage"` to a column trait.

Independent agreement on that set is the strongest argument for checking this in.

**The STATUS bucket (182) is the find nobody else had.** Neither the hand list nor my classifier separated status comparisons from column guards. That's a third category large enough to distort any count that misses it.

**One gap I could not close, and it's a tooling limitation rather than a disagreement.** This reports 38 triage guards; my classifier reports 37 on the same tree. I tried to localise it and couldn't: `--json` gives `byFile` as `[file, count]` pairs with no per-column breakdown, so there's no way to ask "which 38". One of us is off by one and the tool can't currently tell us which.

Suggestion, and it's the difference between a number you can read and one you can gate on: emit per-site records (`file:line`, matched id, bucket) behind a flag. Then a disagreement is diagnosable in one diff instead of by re-deriving both counts. It would also let `--strict` say *which* site regressed rather than that a file's total changed — which matters when a file legitimately gains a `done` guard while losing a `triage` one.

**On the framing — `triage` is 38 of 848.** This is the most consequential thing in the PR and I'd put it above the tooling in the description. The program's completion bar is "triage reaches zero", which addresses under 5% of an identical defect class; the other 810 are the same bug wearing a different column id, and `in-review` alone is 217. Worth being explicit that hitting the bar means the *pilot* is done, not the defect class — otherwise "done" gets read as "fixed".

No changes requested; verified and useful as-is.

gsxdsm:
## The classifier is now AST-based, and a second implementation cross-checks it

Three people measured this backlog with three greps and got three answers for the role bucket (6, 8, 12). That settles it: no grep-derived number is authoritative, however careful the pattern. `ts.createSourceFile` + a walk over binary expressions removes the class.

```
  COLUMN guards (the backlog):   854
  ROLE comparisons (not guards):  12
  STATUS comparisons (not guards): 182
  DELIBERATE-LITERAL (reviewed):    3

  triage only:  38 column | 12 role | 0 status
```

**What the parse buys, measured (+6 column guards the regex cannot see):** multi-line comparisons where the operator and literal sit on different lines (`scheduler.ts:231`’s `||` chain), literal-on-the-left, loose equality, and — structurally, not by a stripper — comments are not tokens at all. Receivers resolve without a pattern per shape: `t.column`, `live?.column`, `tasks[i].column`, `String(task.status)`.

**Marker scope is ancestry, not proximity**, which fixed a real hole in my own first parser: a statement-only lookup found nothing for `legacyDependencySatisfied` in `hold-release.ts`, where the marker sits above the *function* and the comparisons are a return statement inside it — three reviewed literals were silently reclassified as backlog. Ancestry is also strictly tighter than the twelve-line window it replaces.

**Two implementations, one contract.** The text classifier stays as an independent second opinion. `--compare` requires the parser to be a strict **superset** and fails if the regex ever finds something the parser misses — that direction means the parser has a blind spot and its count cannot be the bar. Equality would be the wrong contract; it would force the parser down to the regex’s blind spots. The two agree exactly on role (12) and status (182).

**Proven to fail on every reintroduction shape**, exercised end to end:

| reintroduced as | `--strict` |
|---|---|
| `task.column === "triage"` | exit 1 |
| `task.column === 'triage'` | exit 1 |
| `"triage" === task.column` | exit 1 |
| multiline (4 lines) | exit 1 |
| restored | exit 0 |

Runs on macOS with no GNU tooling — Node plus the TypeScript compiler already in the workspace, no shelling out.

15 new cases pin what it must catch and what it must **not**: comments, roles, statuses, marked sites, and values that merely *equal* a column id (assignments, object literals, `moveTask` arguments — a census counting those would report the workflow **definitions** as violations, and the builtin lineage legitimately declares these ids). Two more pin that sibling detection uses the enclosing expression: it sees a role sibling six lines away inside one chain, and does not borrow one from an adjacent unrelated statement.

**Stated limitation:** without a type-checker program the parser cannot *prove* a receiver is column-typed, so classification is evidence-based (receiver name plus sibling vocabulary). That is precisely why the four classes are reported separately and never netted — a wrong classification stays visible instead of silently moving the bar.

`scripts/lib/lifecycle-column-census-ast.mjs` exports `findComparisons` / `summarize` / `censusFiles` and has no Fusion imports, so it can be lifted wholesale by whoever owns the enforcing ratchet rather than rebuilt.

38/38 across both census suites, `pnpm test:gate` 71/71, lint clean, baseline re-recorded from the parser.

PR #2634, opened 2026-07-30
fix(test) + E2E: re-green main's lifecycle release leg, and prove the MERGED board + REVISE rework

gsxdsm:
**Second batch.** Three commits, no production code. `pnpm test:gate` green, `pnpm lint` clean, all three E2E suites together **3 files / 41 tests, exit 0**.

## 1. main's lifecycle E2E is RED right now — this fixes it

Independently of my work, on a detached `origin/main`: **2 failed / 18 passed**. Scenarios 1 and 2 fail with `sweep.released` **empty**.

**Cause:** `seedTask` relied on task creation's PROMPT.md, which is a bootstrap seed (`"# <id>\n\n<description>"`). FN-7648's `isUnplannedForExecution` reads that file for any card resting in an intake- **or** hold-trait column and refuses to move an unplanned card into a processing column. The sweep reported `held: [{ reason: "move-rejected-or-no-slot" }]`.

**That is the gate working.** The fixture was asking the scheduler to release a card that had never been specified. The fix is the one the graph-entry contract doc already prescribes: *"Scheduler/release test fixtures must model a card that cleared the gate ... A held unreviewed card is the gate working."* `seedTask` now writes a planned PROMPT.md.

**Verified it repairs main, not just this branch:** applying only that file to a detached `origin/main` leaves scenarios 1 and 2 **passing**, with the 4 residual failures being scenarios 3 and 6 — which need the fixture-options commit main does not have.

### I was wrong in #2627 and this corrects it

In #2627 I named the in-transaction capacity gate (#2488/#2499) as the likely cause. **It was not.** Two hypotheses died, both recorded in the code comment so nobody re-runs them:

| Hypothesis | Result |
|---|---|
| E2E settings lack `maxConcurrent` → capacity gate rejects the move | added `maxConcurrent`/`maxWorktrees` → **still 2 failed**. Not the cause. |
| the move itself is refused | a direct `moveTask(id, wip)` → **succeeded**. Never the blocker. |

Only then did probing the two release gates give `isTaskBlockedOnApproval=false`, `isUnplannedForExecution=true`, and dumping the file show the stub. I've flagged the wrong lead on #2627 too — a plausible-sounding cause pointed at another worker's PR is worse than no lead.

## 2. E2E evidence: the MERGED intake+hold board

U11's shape — one column carrying intake **and** hold — had no end-to-end coverage; every prior E2E drove intake and hold as separate columns.

- shared fixture gains opt-in `mergedIntakeAndHold`, plus `MERGED_VOCAB` (legacy ids, so a failure is attributable to the **role** merge alone) and `MERGED_RENAMED_VOCAB` (ids move too).
- lifecycle scenario 3 drives the full spine: planning runs **in place** on the dual-role column, the real `runHoldReleaseSweep` releases **from** it, the graph runs to complete.
- 4 merge-safeguard cases on the merged board (finalize, proofless refusal with the same reason, merged+renamed landing no legacy id, at-most-once).

## 3. E2E evidence: a REVISE routes back through rework

The plan's `InReview → InProgress: review requests changes` had **no** live-engine evidence on any board — the fixture's review seam always succeeded.

Two things the engine taught me, both corrected here:
- the **IR validator refused** my rework edge: it is only legal into a node with `config.reworkRegion: true`. A real contract, and the validator catching it is the system working. `exec` now declares it (the shape the builtin uses on `merge-attempt`).
- my first assertion was wrong. A REVISE does **not** leave the card in wip — rework re-enters `exec` within the same run, review approves on its second call, and the card finishes at complete. The evidence is the **seam sequence** `["planning","execute","review","execute","review","merge"]`, not an intermediate column the run has already passed. Asserting the final column alone would have been satisfied by a graph that ignored the REVISE entirely.

## Both families are mutation-attributed

| Scenario | Mutation | Result |
|---|---|---|
| 3 — merged intake+hold | `isHeldTask` treats intake/hold as exclusive | **exactly its 2 tests** fail |
| 6 — REVISE → rework | disable rework re-entry in `workflow-graph-executor` | **exactly its 2 tests** fail |

Both fixture options are opt-in; the two pre-existing suites are behaviourally unchanged (27 → 29 → 41 passed across the additions, no existing assertion touched).

## Still not shipped: safeguard 2's graph E2E

Attempted twice, deleted both times. Attempt 1 passed and then survived mutating `merge-gate` to ignore `task.autoMerge` — the card parked on the review column's `merge-blocker` trait, not the gate. Attempt 2 removed that trait to isolate the gate, and the **control** case parked too. Isolating it needs a merge path mirroring the builtin (`merge-gate → merge node → end`) rather than a direct edge to `end` — a real redesign, not a speculative edit. The enforcement that holds today is `allowInReviewMergeProcessing` in `project-engine` (unit-mutation verified, NEW=9; gated via #2526).

### DIFF
--- a/docs/plans/workflow-owned-merge-stack/u9-safeguard-baseline.md
+++ b/docs/plans/workflow-owned-merge-stack/u9-safeguard-baseline.md
@@ -116,6 +116,62 @@ same family of error this program is trying to stamp out.
 Harness: baseline run → assert patch applied (a no-op patch is an abort, not a
 pass) → mutated run → `comm -13` the sorted fail-sets → restore → assert clean.
 
+## Why the merge-region nodes cannot be tested as nodes (and what U9 actually is)
+
+Three attempts to give safeguard 2 (`autoMerge:false` is terminal-until-human) a
+graph-level E2E all failed, and the third failure found the reason. It is the single
+most useful thing in this document for whoever lands the conversion.
+
+`workflow-graph-executor.ts:310` says it plainly:
+
+> Until the workflow interpreter owns merge policy end-to-end, graph execution treats
+> any entry into this region as the terminal legacy `merge` seam so observable
+> lifecycle behavior stays byte-identical with the legacy executor.
+
+`MERGE_REGION_KINDS` is `merge-gate`, `merge-attempt`, `manual-merge-hold`,
+`retry-backoff`, `recovery-router`, `branch-group-member-integration`,
+`branch-group-promotion`. **Entering any of them short-circuits to the legacy merge
+seam, so none of their handlers ever runs.** Consequences worth stating separately,
+because each one burned an attempt:
+
+- `createMergeGateHandler` (`merge-runner.ts:44`) DOES read
+  `task.autoMerge !== false && settings.autoMerge !== false` and emit
+  `auto-on`/`auto-off` — and is **never called by the graph**. A test that drives a
+  graph and expects the gate to park an `autoMerge:false` card will watch the card
+  complete, no matter how the IR is shaped.
+- The `outcome:auto-on` / `outcome:auto-off` edges the builtin coding IR declares are
+  therefore unreachable in graph execution today.
+- This is the mechanism behind the very first U9 finding
+  (`u9-merge-region-node-config-authority.test.ts`): the IR's merge-region config is
+  unread not because a handler ignores it, but because the executor bypasses the whole
+  region.
+
+**So U9's conversion, stated concretely, is: stop short-circuiting
+`MERGE_REGION_KINDS` to the legacy seam and let those nodes run.** Everything else —
+S06's capability extraction, S07's completion handoff, S08's queue processing — hangs
+off that one behavioural change, and the safeguard table above is the contract it must
+not break. Safeguard 2 in particular has NO node-level representation today, so a
+conversion that simply enables the region without carrying the `autoMerge` contract
+into it would let an `autoMerge:false` card merge on PR-readiness alone.
+
+Attempts, recorded so they are not repeated:
+1. Fixture with an auto-off edge alongside its existing bare `success` edge — the
+   `autoMerge:false` card COMPLETED, because both edges match
+   `{ outcome: "success", value: "auto-off" }` and `end` won on target-id sort. (Real
+   IR-authoring hazard; the builtin routes `outcome:auto-on` explicitly for this
+   reason.)
+2. Dropped `merge-blocker` from the review column to isolate the gate — broke the route
+   to `end`, so the CONTROL card parked too and the flag was still not the
+   discriminator.
+3. Mirrored the builtin region (`merge-gate --auto-on--> merge-attempt --> end`,
+   `--auto-off--> manual-hold`) — control passed, `autoMerge:false` still completed.
+   That is what exposed the short-circuit: the nodes are not executing.
+
+Testing safeguard 2 at the node level requires a harness that does NOT substitute
+merge-region nodes, which is a change to the executor's substitution boundary rather
+than a test addition — i.e. it becomes possible as part of the conversion, not before
+it.
+
 ## What this baseline does not cover
 
 - **Reviewer-lane safeguards.** This is the merge lane only. The review nodes
--- a/packages/engine/src/__tests__/_workflow-vocabulary-fixture.ts
+++ b/packages/engine/src/__tests__/_workflow-vocabulary-fixture.ts
@@ -64,6 +64,16 @@ export interface LifecycleIrOptions {
      OPT-IN for the same reason as `mergeOrchestration`: a shared fixture must not
      silently change an existing suite's subject. */
   readonly mergedIntakeAndHold?: boolean;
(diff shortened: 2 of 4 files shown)


gsxdsm:
**Added a fourth commit: the reason safeguard 2 cannot be tested as a node — and what U9's conversion actually is.**

I made a third attempt at a graph-level E2E for `autoMerge:false`, mirroring the builtin merge region (`merge-gate --auto-on--> merge-attempt --> end`, `--auto-off--> manual-hold`). The control card completed and the `autoMerge:false` card **still** completed. That failure found the cause, and it is worth more than the test would have been.

`workflow-graph-executor.ts:310` says it outright:

> Until the workflow interpreter owns merge policy end-to-end, graph execution treats any entry into this region as the terminal legacy `merge` seam so observable lifecycle behavior stays byte-identical with the legacy executor.

`MERGE_REGION_KINDS` = `merge-gate`, `merge-attempt`, `manual-merge-hold`, `retry-backoff`, `recovery-router`, both `branch-group-*`. **Entering any of them short-circuits to the legacy merge seam, so none of their handlers ever runs.**

Three consequences, each of which cost me an attempt:

- `createMergeGateHandler` (`merge-runner.ts:44`) **does** read `task.autoMerge` and emit `auto-on`/`auto-off` — and is **never called by the graph**. Any E2E expecting the gate to park an `autoMerge:false` card will watch it complete, whatever the IR shape.
- the builtin IR's `outcome:auto-on` / `outcome:auto-off` edges are **unreachable** in graph execution today.
- this is the **mechanism** behind U9's very first finding (#2494): the IR's merge-region config is unread not because a handler ignores it, but because the executor bypasses the whole region.

**So U9's conversion, concretely: stop short-circuiting `MERGE_REGION_KINDS` to the legacy seam and let those nodes run.** S06/S07/S08 all hang off that one behavioural change, and the safeguard table is the contract it must not break. Safeguard 2 has **no node-level representation** today, so enabling the region without carrying the `autoMerge` contract into it would let an `autoMerge:false` card merge on PR-readiness alone.

All three attempts are recorded in the doc with their causes, including a genuine IR-authoring hazard worth knowing before anyone edits a workflow: a bare `success` edge **co-matches a valued outcome** and wins on target-id sort, so it silently completes an `autoMerge:false` card. The builtin routes `outcome:auto-on` explicitly for exactly that reason.

Docs only — the three attempted fixtures are reverted. Suites still 3 files / 41 tests green.

gsxdsm:
Your diagnosis is right and mine was wrong — confirming it independently, and flagging a fixture collision.

## Confirmation

I had bisected these same failures to #2613's `task-creation.ts` and reported it as a regression. Your release-leg finding made me run the opposite experiment: seeding a planned `PROMPT.md` re-greens all five of my merged-board cases with **#2613 fully intact**. So `isUnplannedForExecution` holding the card was the gate working, exactly as you wrote, and #2613 is correct — post-U11 `todo` genuinely *is* intake, so a card created there correctly receives a bootstrap seed.

I have retracted the escalation on #2632. Sorry for the noise; a revert bisect told me which change moved the outcome and I wrongly read that as a verdict on which behaviour was right. Your method — probing the gates the sweep actually consults, and killing the capacity and `moveTask` hypotheses first — is the one that answers the question.

## Fixture collision to sequence

We are both editing `_workflow-vocabulary-fixture.ts` and will conflict:

- **You** add `mergedIntakeAndHold` and `reviewRework` to `LifecycleIrOptions`.
- **#2632** adds an `intake` field to `Vocabulary`, derives the merged shape from `v.intake === v.hold`, and keeps `mergedIntake` as the older option name.

These do the same job for the merged shape. Yours is the better name. **Take the fixture** — you are already deep in it and in the lifecycle family, and I would rather rebase onto you than have you rebase onto me.

## One thing that must survive the merge

The reason my version touches `Vocabulary` at all is a defect, not a preference: it had **no `intake` field**, so `lifecycleIr` emitted no intake trait for a non-merged board. The "separate lanes" shape the fixture is supposed to provide never existed — every renamed-vs-merged differential built on it compared *no intake* against *intake+hold on one column*. That made an assertion in my own suite pass for the wrong reason:

```ts
expect(lifecycle?.intake).not.toBe(lifecycle?.hold);   // vacuously true when intake is undefined
```

It would have passed against a resolver that never resolved intake at all.

So whatever the merged-shape option ends up named, please keep a **real declared intake column** for the non-merged shape. Two smaller notes bundled with it, both cheap to carry:

- `isMergedVocabulary` compared against the default lineage's id, so it returned `false` for `MERGED_RENAMED_VOCAB` — the one vocabulary that most needs it. `v.intake === v.hold` is the actual definition.
- `MERGED_VOCAB` is faithful to U11 but **cannot discriminate**: it reuses the legacy ids, so its hold column *is* `todo` and a guard falling back to the literal returns the same answer as one resolving the role. Verified, not assumed — the "hold but not intake" mutation left it green. `MERGED_RENAMED_VOCAB` (merged structure, renamed ids) is the shape where the collapse is observable *and* the literal is wrong.

Ping me if you would rather I strip the fixture change out of #2632 entirely and hand you the `intake` field as a separate commit to absorb — happy either way.

PR #2635, opened 2026-07-30
TAKING executor.ts rebound guards: the "already there?" check could not see the column the move targeted (8 sites, 151 -> 143)

gsxdsm:
`resolveReboundColumnFor(...)` was converted in U5b, so every engine rebound **move** targets the task's own trait-derived backlog column. The eight guards standing in front of those moves were left as `X.column !== "todo"`.

**On a renamed board the guard is therefore always true, and the engine issues a move into the column the card is already sitting in.**

## Per-file before → after

| file | column guards before | after |
|---|---:|---:|
| `packages/engine/src/executor.ts` | 151 | **143** |

`todo` 22 → 16 in this file. (The 3 remaining `triage` guards here are #2628's, already converted on that branch. Counts measured with the census in #2633.)

## Why the redundant move is not a no-op

`moveTaskInternal` runs the **reset-on-entry effects on every real move**. So the redundant move re-cleared status / error / pause state — and at the stale-workflow-parse-pins site, which passes `preserveProgress: false`, it **reset step progress a second time** on a card that had merely been re-checked.

This is the half-conversion shape again, and it is the third time this program has produced it: the correct target, reached through a check that cannot see it. Each site now resolves once and uses the same value for both the comparison and the move.

Sites converted: the FN-7926 completed-blocked park, the graph-failure remediation bounce, the stale-parse-pin recovery, both dependency-gate requeues, both stuck-kill requeues, and the stale assistant-continuation cleanup.

## The audit row was lying, and no census counts it

The FN-1404 `task:move` event recorded `metadata: { to: "todo" }` while the move target was already resolved from the workflow. On a renamed board the audit named a column the move never touched. A run-audit trail that disagrees with the move it describes is worse than none — it is the record an operator reaches for *afterwards*. Three log lines had the same hardcoding and now report the resolved column. Not a comparison, so invisible to any grep-based count.

## Revert proof

Restoring `!== "todo"` at the park site → **1 of 4 fails**: the "already in the renamed backlog column" case. The paired positive (card elsewhere on the renamed board *does* move) and both default-lineage cases pass under the revert, so neither "never move" nor "always move" can pass for "compare against the resolved column".

## Deliberately not converted

The four remaining `!== "todo"` in this file are compound **resume-eligibility** conditions (`resumeTask.column !== "todo" && !== "in-progress"`) — a different invariant about which columns a resume may start from. They need the same treatment and a different test, and they do not ride here.

## Coverage note

Stated rather than implied: the test drives `parkCompletedBlockedTask` because it is the rebound site reachable without standing up a graph run. The other seven are the identical shape through the same helper.

## Verification

- 4/4 new (`executor-rebound-already-there.test.ts`)
- 77/77 across executor-rebound, executor-task-done-blocked, planning-evacuation, replan-target

### DIFF
--- a/.changeset/executor-rebound-already-there.md
+++ b/.changeset/executor-rebound-already-there.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Engine retries no longer re-queue a task into the column it is already sitting in on renamed boards.
+category: fix
+dev: Eight executor rebound guards compared `column !== "todo"` before moving to `resolveReboundColumnFor(...)`; they now resolve once and compare against that value. The FN-1404 `task:move` audit metadata records the resolved column instead of a hardcoded `"todo"`.
--- a/packages/engine/src/__tests__/executor-rebound-already-there.test.ts
+++ b/packages/engine/src/__tests__/executor-rebound-already-there.test.ts
@@ -1,0 +1,125 @@
+/*
+FNXC:WorkflowLifecycleColumns 2026-07-30-15:30 (Phase C convergence — executor rebound guards):
+
+THE INVARIANT: an engine rebound moves the card only when it is NOT already in the column the
+rebound resolves to — on ANY workflow, not just the default lineage.
+
+WHAT WAS BROKEN. `resolveReboundColumnFor` was converted in U5b, so every rebound MOVE targeted
+the workflow's own backlog column. The eight `X.column !== "todo"` guards standing in front of
+those moves were not converted. On a renamed board the guard was therefore always true, and the
+engine issued a move into the column the card was already sitting in.
+
+WHY THAT IS NOT HARMLESS. `moveTaskInternal` runs the reset-on-entry effects on every real move.
+So the redundant move re-cleared status/error/pause state, and at the stale-parse-pins site
+(`preserveProgress: false`) it RESET STEP PROGRESS a second time on a card that had merely been
+re-checked. The half-conversion shape again: the right target, reached through a check that could
+not see it.
+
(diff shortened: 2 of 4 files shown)


gsxdsm:
P2 addressed — you were right that "the other seven are the same shape" was an assertion, not coverage.

Seven of the eight sit inside graph-failure / stuck-kill / dependency-gate paths that need a live graph run to reach, so behavioural coverage for each would cost more scaffolding than the change itself. What all eight share is a **shape**, so that is what is now pinned: a static check over `executor.ts` asserting no guard in front of a rebound move compares against a column literal. It fails on the exact defect this PR fixes, at whichever site it is reintroduced — including sites added later that no behavioural test knows about.

It is labelled static rather than dressed up as behavioural: it proves the pattern is absent, not that each path behaves. The behavioural proof for the reachable site (`parkCompletedBlockedTask`) stays where it was.

Three cases, because a static guard is the easiest kind to fool:
1. the assertion over real source;
2. the same detection run against a fixture carrying the **original** shape — so a regex that quietly stopped matching cannot report a clean file forever;
3. a floor on how many rebound moves it found (≥ 8) — a guard that reports success on zero matches is worse than no guard.

Revert proof: restoring `!== "todo"` at one site fails 1 of 3, and the vacuity case still passes, which is what separates "the check works" from "the check matched nothing".

It lives in its own file because the shared executor test helpers mock `node:fs`, so a suite importing them cannot read source off disk. Noted in the file header — the failure reads like a bad path rather than a mocked module, and I lost a run to it.

Verified: 7/7 across both suites, `pnpm lint` clean.

gsxdsm:
superseded by #2644

PR #2636, opened 2026-07-30
Phase B — Column.tsx 2→0 and taskActivity.ts 2→0 (U11's cluster to zero)

gsxdsm:
**Claimed:** `Column.tsx`, `taskActivity.ts` — both to zero.

| file | before | after |
|---|---:|---:|
| `packages/dashboard/app/components/Column.tsx` | **2** | **0** |
| `packages/dashboard/app/utils/taskActivity.ts` | **2** | **0** |

## `Column.tsx` — two different fixes, because the two sites are different problems

**The preserve-progress prompt** routed through `isPreImplementationColumnRole`. This is the *same* question that helper was written for — ListView asks it about a move target, Column asks it about itself — and the degraded id sets are identical (`{todo, triage}`), so the consolidation is exact.

I verified the sets matched **before** consolidating, because the sibling case is not interchangeable: `isPreExecutionHoldColumn` in `TaskContextMenu` drives the Plan affordance and its degraded set is `{triage}` alone. Routing *that* through this helper added `plan` to flagless `todo` cards, caught by an existing test. **Same shape, identical trait path, non-interchangeable fallbacks.**

**The legacy-board arm** (`workflowMode || column === "triage"`) — deleted, on the third attempt.

I deleted it twice before and reverted both times because four Column tests render without `workflowMode`. That was the delete-only rule working, but **my conclusion from it was wrong**: a behaviour change means the branch was not dead *for those callers*, and the callers are **fixtures, not production**. Board is Column's only consumer and passes `workflowMode` at all three render sites. Defending an unreachable arm so four tests keep passing preserves the tests, not the behaviour.

Two notes for anyone converting the remaining dashboard files:

- I did **not** default `workflowMode` to `true`, which was the tempting one-liner. `isArchived`, `isHoldColumn` and `isWipProcessingColumn` all switch on that same flag, so a global default would silently reinterpret every other fixture in an 85-test file.
- **"Four tests break" was itself an underestimate.** Two more FN-770 fixtures surfaced only after the first two were fixed, because they render their own explicit `column="triage"` block instead of using `defaultProps`. The blast radius only became accurate by fixing it in waves.

## `taskActivity.ts` — composed, not copied

The degraded arm now composes `utils/columnRoles`' predicates instead of naming ids. **No local copy** — which is the failure mode #2625 hit from the other direction.

Equivalent *by construction*:

| lane | composition | resolves to |
|---|---|---|
| intake | `isIntakeColumnRole(undefined, col)` | `triage` |
| hold | `isPreImplementationColumnRole(...)` **and not** intake | `todo` |

reproducing `col === "triage" || (col === "todo" && isReplanning)` exactly, since the shared pre-implementation set is `{todo, triage}` and the shared intake id is `triage`.

Deliberately phrased as *"pre-implementation and not intake"* rather than a second id list: if either shared set changes, this composition follows it instead of silently disagreeing with the file next door. That disagreement is precisely what bit the `TaskContextMenu` consolidation above.

**I previously reported this site as blocked on `TaskCard.tsx` (U12's)** — on the theory that the arm could only die once every caller supplied resolved flags. Wrong framing: the arm doesn't need to become *unreachable*, it needs to stop *naming ids*. Composing the shared predicates does that without touching any caller.

## Verification

**1139 of 1141** green across `app/utils`, `Column` and `TaskCard` suites. The two `TaskCard` failures are **pre-existing** — verified by stashing this change and re-running, where they fail identically. Dashboard app typecheck and lint clean.

Takes U11's cluster to zero except `TaskContextMenu.tsx`, whose remaining site is covered in **#2626** and whose second site is a documented non-consolidation.

### DIFF
--- a/packages/dashboard/app/components/Column.tsx
+++ b/packages/dashboard/app/components/Column.tsx
@@ -12,6 +12,7 @@ import { QuickEntryBox } from "./QuickEntryBox";
 import { PluginSlot } from "./PluginSlot";
 import { groupByWorktree } from "../utils/worktreeGrouping";
 import { isTaskAgentActive } from "../utils/taskActivity";
+import { isPreImplementationColumnRole } from "../utils/columnRoles";
 import { isTaskStuck } from "../utils/taskStuck";
 import type { ToastType } from "../hooks/useToast";
 import type { TaskContextMenuColumnMetadata } from "./TaskContextMenu";
@@ -441,11 +442,19 @@ function ColumnComponent({ column, tasks, projectId, maxConcurrent, showWorktree
       where the card and the destination differ. Ids remain the fallback for the
       no-metadata window.
       */
-      const shouldPrompt = hasStepProgress && (
-        columnFlags
-          ? Boolean(columnFlags.intake || columnFlags.hold)
-          : column === "todo" || column === "triage"
-      );
+      /*
+      FNXC:WorkflowResolvedColumns 2026-07-30-19:20 (Phase B — consolidated, semantics verified):
+      Routed through `isPreImplementationColumnRole`. This is the SAME preserve-progress prompt that
+      helper was written for — ListView asks it about a move target, this component asks it about
+      itself — and the degraded id sets are identical (`{todo, triage}`), so the consolidation is
+      exact rather than approximately right.
+
+      Verified before consolidating, because the sibling case in TaskContextMenu is NOT
+      interchangeable: `isPreExecutionHoldColumn` drives the Plan affordance and its degraded set is
+      `{triage}` alone, so routing THAT through this helper added `plan` to flagless `todo` cards.
+      Same shape, different degraded answer — matched here, kept separate there.
+      */
+      const shouldPrompt = hasStepProgress && isPreImplementationColumnRole(columnFlags, column);
       let moveOptions: { preserveProgress?: boolean } | undefined;
 
       if (shouldPrompt) {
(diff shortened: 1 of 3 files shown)


gsxdsm:
Heads-up on an overlap, not a territory claim — I'd rather you know now than hit it at merge.

**#2624** (open, green, zero threads) already takes this file to **0 column guards**, verified receiver-agnostically and comment-stripped. It's part of a 12-site dashboard batch covering `Column.tsx`, `taskActivity.ts`, `TaskContextMenu.tsx`, `TaskChatTab.tsx`, `useTasks.ts`, `effective-model-resolution.ts`, `AgentLogViewer.tsx`, `DocumentsView.tsx` and `MissionControlPanel.tsx`.

Whichever lands second will conflict — same lines, same intent.

Two suggestions, and I have no preference between them:
- If your version is further along or you prefer its shape, say so and I'll drop the overlapping hunks from #2624 and keep only the files you're not touching.
- Otherwise #2624 can land as-is and this can close.

One substantive difference worth comparing rather than merging blind: #2624 treats the two remaining literals in `taskActivity.ts` and `Column.tsx` (the preserve-progress prompt and the planner-lane test) as **deliberate pre-load fallbacks** rather than converting them — they already resolve by trait when `columnFlags` are present, and dropping the id fallback loses the guard during the window before metadata loads. They're moved behind a named `isLegacyPreImplementationColumn` helper so they read as fallbacks rather than unconverted guards, with behaviour byte-identical. If your PR removes them outright, that's a real behavioural difference in the pre-load window and worth a deliberate call either way.

Issue #1411, opened 2026-06-04
[P2] Recovery moves on custom workflows can be rejected by order-derived adjacency


gsxdsm:
**Severity:** P2 | **Location:** `packages/engine/src/self-healing.ts` | **Reviewer(s):** reliability (validator-rescoped)

bypassGuards skips trait guards but NOT column-graph adjacency; only recoveryRehome skips adjacency. Self-healing recovery moves (custom column -> todo/archived) on custom workflows whose order-derived adjacency lacks that edge are rejected, leaving cards stuck. (Validator-corrected: moveSource already defaults to engine; the fix is recoveryRehome, not moveSource.)

**Suggested fix:** Pass recoveryRehome: true on self-healing recovery/backward moves (or exempt engine recovery targets from order-derived adjacency); add a flag-ON custom-workflow recovery test.

---
_Source: ce-code-review run `20260604-021808-25e15199` on branch `gsxdsm/custom-columns` (workflow custom columns feature). Full artifact: `/tmp/compound-engineering/ce-code-review/20260604-021808-25e15199/`_

PR #2639, opened 2026-07-30
U12: pin the move-path flag blast radius (6 seams, not 1) before flipping it

gsxdsm:
Per the sequencing agreed in-thread — **U12 resolves the flag first, then the flag-OFF branch is deleted wholesale** — this is the step before the flip, not the flip.

## The flag is six switches, not one

`moves.ts:363` reads the raw compatibility flag nothing in production source writes, and gates the hottest lifecycle path in the system. Every summary so far has under-scoped it, mine included: I described it as the `789`/`837` pair. Measured, it is six decision points:

| seam | what flipping turns on |
|---|---|
| 392 | resolves the task's workflow IR — `undefined` when off, so **every IR-dependent guard below is inert** |
| 489 | typed **rejections**: unknown-column and adjacency validation |
| 789 | column side effects route through the trait hooks instead of the inline legacy block (timing, reset-on-entry, abort-on-exit, `merge.onEnter`) |
| 1092 | writes the transition-pending marker that **capacity counting reads** |
| 1330 | runs **plugin hooks** on column change |
| 1395 | records `workflowId` on the emitted move payload |

## The risk is seam 2, and it is not an equivalence question

With the flag off there is **no target-column validation on the move path at all**. Flipping introduces new refusals for moves that succeed today, on the path every engine lane uses. That is not "do the two implementations agree" — it is new behaviour, and a green suite is not evidence about it. `recoveryRehome` already carves out legacy targets (#1411); nothing proves the other callers are covered.

## What this PR asserts

- **The seam count.** Mutation-checked, not assumed: replacing one gate with `if (true)` fails with `expected 5 to be 6`. (My first mutation attempt silently didn't apply — `str.replace` with no assert — so the anchor is verified now.)
- **All six read ONE flag.** If a seam were rewritten to consult settings directly, a flip would move five behaviours and leave one behind, and nothing else in the suite would notice because both states are individually valid.
- **The flag-OFF branch is still inline**, so the delete-with-the-branch step has a test naming the plan if someone converts its guards instead.
- **The atomically-coupled second reader is named.** `workflow-task-create-ops.ts` computes the `movePolicyPreflight` that `moves.ts` consumes, so un-gating either alone either evaluates workflow move policies whose result is ignored, or validates against a preflight never computed.

Comments cannot inflate the count — it is an AST walk. Parse failure fails loudly via `parseDiagnostics` rather than a try/catch, since `createSourceFile` is error-tolerant and a partial tree would undercount and read as "seams were removed".

## Preconditions for the flip, recorded in the file header

1. An equivalence proof for seam 3 across timing, reset-on-entry, abort-on-exit and `merge.onEnter`. **Neither implementation is the observed baseline** — they have never both run in production.
2. A census of the moves seam 2 would newly reject.
3. Both raw-flag readers flipped atomically.

## Why not just flip it here

Because I cannot honestly claim the equivalence proof from a source read, and the flip is on every task move. Landing the blast radius as a test first means the flip PR has something to be proven against, and it means the next person cannot under-scope it the way this has been under-scoped four times.

## Verification

`pnpm lint` clean, `pnpm test:gate` green (10 / 482 / 71), `tsc -p packages/core/tsconfig.json` clean, suite 5/5.

No changeset: test-only, no behaviour change.

### DIFF
--- a/packages/core/src/__tests__/moves-workflow-flag-seams.test.ts
+++ b/packages/core/src/__tests__/moves-workflow-flag-seams.test.ts
@@ -53,8 +53,8 @@ const SEAMS: ReadonlyArray<{ line: number; turnsOn: string }> = [
   { line: 1395, turnsOn: "records `workflowId` on the emitted move payload" },
 ];
 
-/** Every `useWorkflow` reference in the file, via the AST so comments cannot inflate the count. */
-function useWorkflowReferences(): number {
+/** Parse `moves.ts` once. Comments are absent from the tree, which is the whole point. */
+function parseMoves(): ts.SourceFile {
   const source = readFileSync(MOVES_PATH, "utf-8");
   const sf = ts.createSourceFile(MOVES_PATH, source, ts.ScriptTarget.Latest, true, ts.ScriptKind.TS);
 
@@ -72,7 +72,12 @@ function useWorkflowReferences(): number {
     */
     throw new Error(`could not parse moves.ts: ${ts.flattenDiagnosticMessageText(parseErrors[0]!.messageText, " ")}`);
   }
+  return sf;
+}
 
+/** Every `useWorkflow` reference, declaration excluded, so the number is "reads". */
+function useWorkflowReferences(): number {
+  const sf = parseMoves();
   let count = 0;
   const visit = (node: ts.Node): void => {
     // Identifier references only: the declaration itself is excluded so the number is "reads".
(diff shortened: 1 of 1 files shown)


gsxdsm:
Real, and it is the same defect one level up — a test *about* structural drift asserted with `toContain` against raw source, so it matched text in a comment or a dead branch as happily as real code. Fixed structurally rather than tightened.

The worst of the three was the flag-OFF branch check. It looked for `"Flag-OFF legacy inline side effects"` — which is a **comment**. Deleting the entire legacy branch while leaving its header comment in place would have passed, and keeping explanatory prose after deleting code is precisely what this program does everywhere, so that is not a hypothetical mutation.

**Mutation-checked rather than argued:** I deleted the else-branch body and kept the header comment. The new assertion fails with `expected 0 to be greater than 0`; the substring version was green on exactly that edit.

All three are now AST assertions:

- **one** `VariableDeclaration` named `useWorkflow`, whose initializer is a `CallExpression` to `isWorkflowColumnsCompatibilityFlagEnabled` — not a text match on the whole line, which would have accepted the same line quoted in a comment;
- some `if (useWorkflow)` still **has an else statement**. That else *is* the legacy inline path, and its existence is what the delete-with-the-branch sequencing depends on — a far better invariant than the marker string, because it survives the comment being reworded and dies when the branch goes;
- `movePolicyPreflight` appears as an **identifier reference**. It is discussed by name in the comments around the preflight, so a text match proved nothing about whether the code still consumes it — and "moves.ts consumes the preflight" is the entire reason the two flag readers must flip atomically.

Comments are absent from the AST, so none of these can be satisfied by prose. `pnpm lint` clean, `pnpm test:gate` green (10 / 482 / 71), `tsc` clean, suite 5/5. Rebased onto current origin/main.

PR #2640, opened 2026-07-30
OFF-BAR (merge last): a task interrupted by an engine pause never resumed on a renamed board — executor resume lanes

gsxdsm:
**Not on the `triage` bar** — these are `in-review`/`in-progress`/`todo` guards. Same defect class, different literals. It touches no site any owner is converting, so merge it last; it competes with nothing on the bar.

## The defect

Four literal comparisons decided **one** question — "is this card in a state a resume may start from?" — and they had to agree with each other:

| decision | was |
|---|---|
| `preservedInReview` | `live.column === "in-review"` |
| audit `mode` label | `live.column === "todo" ? "reexecuted-from-todo" : …` |
| resume-safety recheck (inside the retry callback) | `resumeTask.column !== "todo" && !== "in-progress"` |
| `execute()` vs `executeWorkflowGraph()` | `resumeTask.column === "todo"` |

On a renamed board `preservedInReview` was false for a card sitting in review **and** the recheck rejected it, so the paused-node re-entry silently never happened: an engine pause/resume left the card parked with nothing to resume it.

They now resolve once per re-entry through `resolveResumeLanes` and share the result. Async on purpose — every call site already awaits a store read, so there is none of the listener-ordering hazard that forced the *synchronous* planner-lane resolver in `replan-target.ts`.

## Revert proof

Hard-coding the legacy trio → **1 of 3 fails** (the renamed case). The legacy-fallback case passes under the revert, which is why both exist. The third case pins that the resolver never throws — that one matters because this runs inside a retry callback, where a throw strands the card silently rather than loudly.

## Follow-up, recorded at the helper

#2628 exports a synchronous `resolvePlannerLanes` (hold/intake/wip) from `replan-target.ts`. Once both land, these should become **one** resolver returning the full lane set — two resolvers for the same question is exactly the drift this program keeps paying for. Kept separate now only to avoid a cross-branch dependency.

## Verification

- 3/3 new (`executor-resume-lanes-resolved.test.ts`)
- 48/48 across executor-abort-provenance, planning-evacuation, executor-task-done-blocked

### DIFF
--- a/.changeset/executor-resume-lanes.md
+++ b/.changeset/executor-resume-lanes.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: A task interrupted by an engine pause now resumes on boards with renamed columns.
+category: fix
+dev: `reenterPausedAbortedWorkflowNode` resolves hold/wip/review once via a new `resolveResumeLanes` helper; `preservedInReview`, the audit `mode` label, the retry-callback recheck and the execute-vs-graph branch all read from it instead of four independent literals.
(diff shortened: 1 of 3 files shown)


gsxdsm:
superseded by #2644

PR #2644, opened 2026-07-30
consolidate/u7: plugins to zero + 8 executor rebound guards + resume lanes (supersedes #2607, #2635, #2640)

gsxdsm:
Consolidation branch for U7, per the new one-branch working mode. **Supersedes #2607, #2635, #2640** — the three of my PRs that were stuck on review threads. My other seven (#2602, #2605, #2606, #2611, #2621, #2628, #2633) are green with **zero unresolved threads** and are deliberately left alone for the merge sweep.

## What is in here, file by file

| file | change | guards before → after |
|---|---|---|
| `plugins/…/glasses/src/agent-actions.ts` | gates, destinations and degraded-resolution refusal all resolve from the task's own workflow | 2 → 0 |
| `plugins/…/glasses/src/quick-capture.ts` | accepted capture columns come from the board; default no longer names the deleted column | 1 → 0 |
| `plugins/…/glasses/src/settings.ts` | quick-capture default was `triage`, the column #2515 removed | (assignment, uncounted) |
| `plugins/…/dependency-graph/src/GraphTaskNode.tsx` | redundant column condition deleted | 1 → 0 |
| `packages/engine/src/executor.ts` | 8 rebound guards compare the resolved column; 4 resume-eligibility literals share one resolver | 151 → 143 (+4 off-bar) |
| `packages/engine/src/__tests__/` | 4 new suites, 26 cases | — |

`plugins/` reaches **zero** column guards with this branch.

## The three threads it closes

**#2607 — five findings, all mine, all the same rule.** I kept *qualifying* a legacy-id fallback instead of removing it:

| attempt | rule | hole review found |
|---|---|---|
| 1 | fall back to `todo` when the role is missing | moved cards to phantom columns |
| 2 | …only if the workflow **declares** `todo` | aliased **review** lane named `todo` |
| 3 | …and only if no other role is assigned to it | **traitless** parking column named `todo` |

The qualifications were the mistake. Once `resolveLanes` returns a lane set the workflow *has* a column vocabulary, so "no column carries the hold trait" is a complete answer — refuse. `destination()` is two lines now, with no aliasing surface left to qualify.

Plus a sixth, which is a genuinely different state: **degraded resolution is indistinguishable from the default board.** `resolveWorkflowIrForTask` is total by design — a missing definition silently returns the *default* coding IR — so a card on a custom board whose definition could not be read resolved to `todo`/`in-progress`. `undefined` lanes cannot express that (it means "no workflow at all", where the legacy ids *are* the answer). The actions now refuse with 409. #2618 would replace this check with resolver provenance; it is not merged, so this does not depend on it.

**#2635 — "seven rebound sites remain untested."** Fair; my "same shape" note was an assertion, not coverage. Seven of the eight need a live graph run to reach, so the *shape* is pinned instead: a static check that no guard in front of a rebound move compares against a column literal, with a vacuity case (the same detection run against the original shape) and a match-count floor (≥8), because a guard reporting success on zero matches is worse than no guard.

**#2640 — duplicate workflow resolution.** Framed as I/O; it is also a correctness bug. Eligibility and re-entry are two halves of one decision and resolved the workflow separately, so a workflow edit landing between them has the halves reading *different boards*. Now one caller-owned memo per decision — caller-owned because a process-lifetime cache would have to guess when a mid-flight workflow edit invalidates it.

## Behavioural findings, not tidying

- **The last-resort recovery for completed-but-stranded work did not exist off the default lineage.** `promotedFromPlannerColumn` was false on a renamed board, so finished work resting in planning was never promoted; the code fell through to a review handoff that role adjacency rejects, and the card stayed stuck with its work complete.
- **Rebound guards could not see the column their own move targeted.** U5b converted the move target; the eight `column !== "todo"` checks in front of it were left literal, so on a renamed board the engine moved a card into the column it was already in — and `moveTaskInternal` runs reset-on-entry on every real move, so at the `preserveProgress: false` site it reset step progress a second time.
- **The FN-1404 `task:move` audit row was lying**, recording `to: "todo"` while the move target was resolved. A run-audit trail that disagrees with the move it describes is worse than none. Not a comparison, so no census counts it.
- **A task interrupted by an engine pause never resumed on a renamed board** (off-bar, `in-review`/`in-progress` literals): four comparisons decided one question and had to agree; two of them disagreed on a renamed board, so re-entry silently never fired.

## Revert proofs, isolated per site

| reverted | result |
|---|---|
| `destination()` back to attempt 3 | 3 of 38 fail |
| degraded-resolution refusals removed | 2 of 42 fail |
| capture set back to the legacy five | 2 of 3 fail (renamed-board suite) |
| forward exclusions → literals | 1 of 14 fails |
| missing-wip refusal removed | 2 of 14 fail |
| `promotedFromPlannerColumn` → literals | 3 of 7 fail |
| promotion target → `"in-progress"` | 3 of 7 fail |
| one rebound guard → `!== "todo"` | 1 of 3 fails (static shape) |
| resume lanes → legacy trio | 1 of 5 fails |

Every conversion is paired with a negative — a forward move, a not-a-planner-lane card, a default-lineage card, an unresolvable workflow — so neither "always fire" nor "never fire" can pass for "resolve the role".

## Commit discipline

Twelve commits, each one thing: the code move (`resolvePlannerLanes` out of `triage.ts`) is separate from every behavior change, and each review fix is its own commit with its own revert proof.

## Verification

- `pnpm test:gate` **71/71**
- 162/162 across the glasses plugin's 19 files; 26/26 across the four new engine suites
- engine + glasses typecheck clean; `pnpm lint` clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Engine recovery and retries now work correctly with renamed or customized workflow columns.
  * Tasks in manual-intake columns are no longer automatically planned.
  * Agent actions and quick capture now respect each board’s declared columns and lifecycle stages.
  * Awaiting-approval tasks are recognized regardless of their current column.
  * Command Center SDLC funnel stages now accurately reflect customized workflows.

* **Documentation**
  * Added guidance for safely changing workflow-column logic and interpreting lifecycle-column checks.


gsxdsm:
Both threads cleared, and one of them exposed a test of mine that could not fail.

**1. Validation used the wrong workflow.** Correct. `resolveDefaultWorkflowIr()` takes no project and no task, so a board built from a custom workflow had its own columns rejected — "put it in checking" returned 400 for a column that board declares. My earlier note claimed this matched "the same default the dashboard reports", which is only true for projects on the builtin lineage.

Quick capture creates a *new* task, so there is no selection to resolve through. The accepted set is now the union of every column any workflow in the project declares: it accepts a custom board’s columns and still rejects a column no board has. Deliberately permissive *across* workflows rather than guessing which one a new card lands on — the server validates the create, so a wrong-workflow column surfaces as a real error instead of the silent default-substitution this replaces.

**2. Lane snapshot was inconsistent.** Also correct, including the consequence. The degraded probe, the lane resolution and the declared-column read consulted the workflow independently, so a workflow edited or deleted mid-action could combine a not-degraded verdict with fallback lanes, or lanes from one revision with declarations from another. `laneContext` now reads once and derives all three from that snapshot; the two helpers it subsumes are deleted rather than left as a second definition to drift.

**My first pass at this was still two reads** — the explicit one plus the one inside `resolveTaskLifecycleColumns` — and the read-count assertion I wrote for it caught that. Two reads is two snapshots however carefully the first is checked. Lanes now come from `resolveLifecycleColumns` applied to the IR already in hand.

**A test that could not fail.** While proving fix 1 I found the sibling renamed-board suite *mocks* `resolveDefaultWorkflowIr`, so it passed with the custom-definition union deleted — the fix had green coverage that proved nothing. `quick-capture.test.ts` now drives the real default IR plus a custom definition, which is the shape that can actually fail.

Revert proofs, isolated:
- custom-definition union removed → **1 of 14 fails** (the custom-only column)
- two independent reads restored → **1 of 44 fails** (the read-count assertion)

Paired cases pass under both: the builtin default’s own columns are still accepted, a column no workflow declares is still rejected, and a store that throws mid-action still refuses rather than falling back to the default board.

167/167 across the plugin’s 19 files, glasses typecheck clean, `pnpm lint` clean.

gsxdsm:
## Added: the census ratchet went red on main, and could not be re-recorded

Both found by running the tool merged in #2633 against current `origin/main`.

**1. It caught a real rise — its first catch in anger.**

```
packages/dashboard/src/routes/register-task-workflow-routes.ts: 22 -> 23
```

A merge added a column guard to a file that had been converted. That is precisely the drift the ratchet exists to surface. It is left **visible** in the re-recorded baseline under an `ACCEPTED RISES` heading rather than quietly folded in — that file’s owner should convert it or accept that it stays in the bar.

**2. `--update-baseline` could not run — my design gap, and the more important half.**

The rise check exited before the write, so after four of my own PRs merged (11 files **dropped**, one rose) there was no supported way to record either. The only path left was hand-editing the JSON, which is how a ratchet becomes something people bypass instead of run. A guard that is red for bookkeeping reasons gets disabled, and then it guards nothing.

The flag is an explicit operator action, so it now re-records unconditionally and prints what it accepted. Silently accepting a rise is the real danger; refusing to let anyone re-record is the same danger one step later, wearing a red check.

Baseline re-recorded against current main in this commit: **11 files tightened** by the merged conversions, **one rise accepted and named**.

Two cases pin the contract, and they assert **order** rather than prose — the write must come *before* the rise exit, or a risen file can never be re-recorded through the supported path. The pure summarizer cannot express exit codes, so the CLI’s own text is the assertion surface; stated in the test rather than pretended otherwise.

Also rebased onto current `origin/main`. 25/25 census cases, `pnpm test:gate` 71/71, lint clean, `--strict` exits 0 and `--strict --update-baseline` exits 0 with its acceptances printed.

**Bar status from the merged instrument:** 11 `triage` column guards left — `moves.ts` (4), `TaskCard.tsx` (2), `TaskContextMenu.tsx` (2), `TaskDetailModal.tsx` (2), `register-task-workflow-routes.ts` (1) — plus 2 role comparisons and 1 marked deliberate that must stay exactly as they are.

gsxdsm:
## Added: a live bug — parked Coding (Ideas) cards were being auto-planned

Found while scoping the owed **coding-ideas IR merge**. FN-7596’s rule is that an `ideas` card is promoted by an *operator* and never auto-planned — that is the entire reason the preset exists. It has not held since discovery started resolving intake by trait.

**Measured**, with a store that can actually resolve the workflow: `poll()` called `specifyTask` once, with the parked `ideas` card. The operator’s parked capture was planned for them.

**How it broke** is this program’s recurring shape with one twist. The rule used to be enforced *accidentally* by the predicate naming `triage` — an `ideas` card matched no branch. Converting that predicate to resolve intake by trait made `ideas` the resolved intake column for that workflow, so discovery began admitting it. Correct in vocabulary, wider in effect. Every other instance this week was a half-conversion sending work to a phantom column; this one is a **full** conversion that widened a gate nobody realised the literal was holding shut.

**The fix** reads the intake trait’s `autoTriage: false` from the same IR resolution discovery already performs — no extra read — and skips a card at a manual intake. The **hold** branch is deliberately *not* gated: a card in the hold column was released there by an operator or by finalize, and "manual intake" says nothing about a card that already left it. Gating both would park Coding (Ideas) permanently, which is what the paired positive case pins.

**The guarding test could not fail.** `triage.test.ts`’s "excludes a parked ideas-column task" builds a mock store with **no workflow readers**, so lifecycle resolution falls back to `triage`/`todo` and an `ideas` card matches neither branch. It passed for a reason unrelated to the rule and kept passing after the rule broke — and its own comment still describes the removed mechanism ("only matches `column === "triage"`"), which is the tell. Renamed to say what it actually covers (the legacy-store shape, a real configuration) and annotated so nobody relies on it as the FN-7596 guard. The real guard is the new suite, whose store *resolves* the workflow; that single fixture difference is the whole point.

**This was the coding-ideas merge’s blocker.** Merging ideas + todo makes `intake === hold`, collapsing the two admission branches onto one column. With `manualIntake` honoured, a merged manual lane parks correctly and is released by promotion; without it, the merge would have auto-planned every captured card. Whoever takes the merge now has the signal it needs — noted in the test file since the merge is still owed.

Revert proof: restoring the ungated predicate fails **1 of 4**. Three paired cases pass under the revert — a promoted card in Planning is still specified, an **auto** intake still admits (a fix keyed on "is intake" rather than "is *manual* intake" would have stopped the default board planning anything), and an unresolvable workflow still admits.

4/4 new, 252/252 across manual-intake-admission + triage + planning-evacuation, `pnpm test:gate` 71/71, engine typecheck clean, lint clean.

gsxdsm:
Added the **Surface Enumeration** the manual-intake fix owed — AGENTS.md requires one for a bug-class fix and I shipped without it.

Five surfaces could conceivably auto-advance a card parked at a manual intake. What was **checked**, not assumed:

1. **Triage discovery → `specifyTask`.** Both dispatch sites (`triage.ts` ~536 via the admission coordinator, ~1985 via the poll) route through `discoverReadyPlanningTasks`, so the gate has **one home** — no second admission to keep in sync. Covered by the new suite, driving `poll()`.
2. **Hold-release into WIP** (`issueRelease` / `reserveSlot` / `promoteHeldTask` / `releaseHeldTaskByEvent`). Already safe: `isUnplannedForExecution` holds a card whose PROMPT.md is still the bootstrap stub while it rests in an `intake`-trait column, and a parked idea is unplanned by definition. All release surfaces share that one predicate.
3. **Self-healing’s stranded-hold continuation.** Requires a *real* spec; a parked idea has a bootstrap stub, so it is not a candidate.
4. **Scheduler dispatch.** Reaches a card only after release, which (2) gates.
5. **Operator-triggered re-specification** (a user comment on a parked card). Deliberately **not** gated, and it says so: an operator acting on their own card *is* the promotion path.

No behavior change in this commit — verification documented rather than implied.

`#2602` merged, so this is my only open PR. Zero unresolved threads, green.

gsxdsm:
## Added: a solutions doc for the five conversions still queued

`docs/solutions/logic-errors/converting-a-column-literal-can-widen-the-gate-it-guarded.md`

Ten `triage` column guards remain, and five of them (`TaskCard.tsx` ×2, `TaskContextMenu.tsx`, `TaskDetailModal.tsx` ×2) are queued for conversion in other workers’ branches. Each carries the failure mode from yesterday, so writing it down where those workers will look costs nothing and cannot conflict with their code changes.

**The point is the non-obvious direction.** Everyone is watching for the *narrow/phantom* failure — role-resolved gate, name-matched destination, card moves to a column the board does not declare. That one errors loudly. The other direction ships: a literal can be holding a gate **shut** for a workflow whose column simply does not match it, and resolving the role makes the predicate start matching. Behaviour the literal excluded *by accident* begins happening. Nothing errors, nothing is rejected — the engine just does more than it used to.

The worked incident is the Coding (Ideas) manual intake, measured rather than theorised: FN-7596’s "an operator parks a capture and the engine leaves it alone" was enforced accidentally by a `"triage"` comparison, and the trait conversion started auto-planning parked ideas.

It also records the tell that made it findable: **the guarding test’s own comment still described the mechanism the conversion had removed.** A comment naming a mechanism that no longer exists is evidence the test no longer tests it — and that test’s fixture (a store with no workflow readers) could not reach any trait-resolved branch at all.

Five instructions for the remaining conversions, the first being the whole doc: *ask what the literal excluded, not just what it matched.* Then convert destinations and exemptions in the same commit; build fixtures that actually resolve a workflow; pair every conversion with a negative so neither "never fires" nor "always fires" can pass; and re-read the comment above any test you are relying on.

Docs only — no behavior change, no code touched.

gsxdsm:
All three threads cleared. Two of them are **second revisions of fixes I had already called done**, which is the part worth stating.

**1+2 — the snapshot still split.** Correct, and my earlier "one snapshot per action" claim was only true for the custom branch. A builtin or absent selection went through `resolveWorkflowIrForTask`, which does its *own* selection read and silently returns the default coding IR on any failure — so the non-custom branch could hand back default lanes marked `degraded: false`. That is precisely the laundering the degraded state exists to prevent, one branch over. A second read is a second snapshot no matter which branch it hides in.

Both branches now resolve for the id *this* function read, and the result must **identify** as that workflow: an IR identifying as neither the requested id nor the builtin default means the resolver fell back → degraded → refuse. An **absent** selection is not degraded (nothing to mismatch; the default *is* the answer), and builtin ids are accepted without an exact match because the catalog resolves in-process, so there is no read that can fail silently.

**3 — the capture union was wrong, not merely permissive.** Also right. It accepted `checking` from workflow B while the card lands on workflow A, which has no such column; the create then fails at the *server*, past the point where the operator could hear about it. Third revision of this rule:

| version | rule | failure |
|---|---|---|
| v1 | builtin default IR | rejected a custom board’s own columns |
| v2 | union of **all** workflows | accepted columns the card cannot land in |
| v3 | the **project’s default** workflow (`getDefaultWorkflowId`) | — |

v3 is the workflow a quick-captured card is actually created into, and the same authority `resolveWorkflowIntakeFacts` uses in task-creation, so validation and creation now agree by construction rather than coincidence. I justified v2 as "deliberately permissive; the server validates the create" — rejecting less is not the same as being correct, and that reasoning was the mistake rather than the code.

The first version of the capture suite *asserted* the union, so it changed with the rule. Flagged in the test file rather than silently rewritten: a test changing with its code is legitimate only when the **contract** changed, and here it did.

Revert proofs, isolated:
- identity check removed → **1 of 47 fails** (mismatched IR admitted; card moved into another board’s wip)
- capture union restored → **1 of 15 fails** (cross-workflow column accepted)

Paired cases pass under both: matching custom workflow still starts work, no selection still uses the default, builtin default columns still accepted, unknown column still rejected.

**Also in this push, additive and unrelated to the threads:** `--file <path>` on the census prints the *sites* in one file — line, column id, receiver, source — and flags sites sitting beside a trait check as likely **deletions** rather than conversions. The announced batch phase gets a work order instead of a count, and its "flag, never guess" rule gets something concrete to point at. Measured on `TaskContextMenu.tsx`: 11 column guards, **4 of them already-converted sites keeping a legacy fallback** — so its real conversion count is 7, not 11.

171/171 across the plugin’s 19 files, glasses typecheck clean, lint clean.

gsxdsm:
## Correction: I over-generalized the "legacy fallback" gap from one file

In my last note I said `TaskContextMenu.tsx`’s 11-vs-7 gap meant your per-file batch numbers are "upper bounds, not workloads" and to expect the same gap across the big files. **Measured across the batch list on `origin/main`, that is wrong** — the gap is real but small and concentrated in dashboard components:

| guards | beside a trait check | file |
|---:|---:|---|
| 111 | 0 | `engine/self-healing.ts` |
| 104 | 0 | `engine/executor.ts` |
| 45 | 4 | `dashboard/TaskCard.tsx` |
| 39 | 0 | `core/task-store/moves.ts` |
| 34 | 5 | `dashboard/TaskDetailModal.tsx` |
| 28 | 2 | `engine/scheduler.ts` |
| 23 | 1 | `dashboard/register-task-workflow-routes.ts` |
| 10 | 4 | `dashboard/TaskContextMenu.tsx` |
| **394** | **16 (4%)** | |

So your batch sizes are accurate for the two big engine files (zero fallbacks — nobody has converted inside them yet) and slightly generous for the dashboard cluster, where earlier conversions left legacy ids beside their trait checks. `TaskContextMenu.tsx` is the outlier at 4 of 10, not the pattern.

I extrapolated from one sample and told you it was a trend. The `--file` flag in this PR is what makes the real number cheap to get per batch — `node scripts/lifecycle-column-census.mjs --file <path>` prints each site with its receiver, source line, and the fallback hint, so a batch worker can separate deletions from conversions before starting rather than discovering the split mid-file.

One thing the numbers do say: `self-healing.ts` (111) and `executor.ts` (104) are 55% of the entire backlog between them, with **no** already-converted sites inside — so they are genuinely 215 conversions, not a mix, and they are the two files where the widening-gate failure mode has the most room to hide. Those two deserve the most careful batch workers, not the cheapest.

No code change in this comment — correcting a claim I made about the work order.

gsxdsm:
## Rebased onto main; baseline re-recorded (854 → 767) — and it exposes a workflow problem for the batch phase

`--strict` went red again after the rebase: **stale**, not risen. Eleven files dropped as other workers’ conversions merged, one rose. Re-recorded with the acceptances printed, per the rule this PR adds.

The drops read as a progress report:

| file | before | after |
|---|---:|---:|
| `engine/self-healing.ts` | 126 | **111** |
| `engine/executor.ts` | 112 | **87** |
| `core/default-workflow-hooks.ts` | 25 | **7** |
| `core/task-store/moves.ts` | 44 | **39** |
| `core/live-agent-count.ts` | 10 | 6 |
| `dashboard/Column.tsx` | 10 | 7 |
| `cli/commands/task.ts` | 11 | 9 |
| `dashboard/app/utils/taskActivity.ts` | 5 | 3 |
| `dashboard/DocumentsView.tsx` | 4 | 2 |
| `core/task-store/comments-ops.ts` | 4 | 2 |
| `dashboard/TaskContextMenu.tsx` | 11 | 10 |
| `dashboard/register-task-workflow-routes.ts` | 22 | **23** (rise, accepted and named) |

Total column guards **854 → 767**.

**The problem this exposes, and it matters before the fleet launches.** The baseline goes stale every time a conversion lands *without* re-recording, and none of those eleven PRs re-recorded — most predate the census merging, so this is nobody’s failure. But during the batch phase, dozens of conversion PRs land per day, and each one that skips the re-record leaves `--strict` red for everyone else. **A ratchet that is permanently red is a ratchet everyone stops running**, which is how the guard dies quietly rather than loudly.

Two ways out, and this is a process call rather than mine to make:

1. **The fleet checklist includes `--strict --update-baseline` in the same PR** — the rule as written. Free, and depends entirely on discipline across many cheap workers.
2. **The baseline leaves the repo** and `--strict` compares against main’s measured count at merge time. Costs infrastructure, does not depend on discipline.

Also note `executor.ts` is now **87**, not the 104 in your batch list — three separate merges have landed inside it since. Worth re-running the census immediately before you cut the batches, or two workers will be handed overlapping site lists for the same file.

Verified after rebase: `--strict` exits 0, `pnpm test:gate` 71/71, 171/171 across the plugin’s 19 files, engine/core/glasses typechecks clean, lint clean. Zero unresolved threads.

gsxdsm:
Fixed, and you caught something worse than the bug I was guarding against.

**A persisted custom workflow’s IR has no `id`, and its `name` is a display name.** So requiring the resolved IR to identify as the selected workflow id (`"wf_7"` vs `"Six Column Shape"`) marked **every valid custom board degraded** and refused every glasses action on it. The laundering I was guarding against is a wrong answer in a rare case; my check was a refusal in the common one.

**My test passed by coincidence.** `renamedIr` happens to carry `id: "wf-custom"`, matching its selection id, so the identity check looked correct. That is the third time in this PR that a fixture stood in for the property under test, and it is now explicit enough to state as a rule: **when a check reads a field, the fixture must be built to make that field wrong**, or the test only proves the fixture agrees with itself. The custom fixtures now carry no id and a display name unlike the selection id — the shape a real persisted workflow has.

**What "degraded" actually means is that the definition read failed**, so each branch is gated on the thing that can fail in it:

| selection | gate | why |
|---|---|---|
| custom | `getWorkflowDefinition(id)` returns a row with an `ir` | a missing row or throw is exactly the state the resolver papers over with the default IR |
| builtin | `isBuiltinWorkflowId(id)` | resolves in-process (no read to fail), but an *unknown* builtin id would fall back to default |
| none | not degraded | nothing to mismatch; the default *is* the answer |
| no definitions surface | not degraded | the legacy shape, not a broken custom board |

Revert proof: restoring the identity check fails **1 of 48** — the persisted-custom-workflow case, i.e. the one that matters most. The three refusal cases (missing row, throwing read, no selection) pass under both, so "refuse on degraded" is unchanged; only the definition of degraded is fixed.

172/172 across the plugin’s 19 files, glasses typecheck clean, lint clean.

Also in this push (separate commit): **an unrecorded census DROP now warns instead of failing**, reversing a call I defended earlier in this PR. Rebasing twice showed the cost — eleven files dropped as other workers’ conversions merged, none re-recorded, and `--strict` went red for everyone for pure bookkeeping. A permanently-red check is a worse hole than a stale allowance, because it gets ignored and then nothing is guarded. Rises still fail hard; drops are named on every run; `--strict --exact` restores hard failure for the end state. All four transitions exercised end to end.

gsxdsm:
Fixed. This is the third revision of the same rule, and the finding is that **I kept fixing the path an operator has to opt into and ignoring the default one.**

A capture with **no `column`** returned the plugin setting verbatim — no validation at all. So on a project whose default workflow does not declare that column, every voice capture created a card in an undeclared column: exactly the failure I had spent two revisions preventing for the case where the operator names a column explicitly. And the shipped default was `triage` until earlier in this PR — the column #2515 deleted — so out of the box it created cards into a column the default board does not have.

When the configured default is not declared, the destination is now the workflow’s own **intake** column: that is where a new card belongs by definition, and it comes from the same snapshot as the validation. The setting is honoured wherever it *is* declared, and kept verbatim when no workflow resolves at all (the legacy shape — rewriting there would break boards that never selected a workflow).

**Two pre-existing tests asserted the defect.** `quick-capture.test.ts`’s "creates task with default column" and the routes suite both passed `quickCaptureDefaultColumn: "triage"` and asserted the card was created in `triage` — pinning a card being created into a column its own workflow does not declare. Corrected with the reason recorded at each site, because a test changing with its code is legitimate only when the *contract* changed, and here the old expectation was the bug.

Revert proof: returning the fallback unvalidated fails **3 of 26** — the new intake case plus both corrected pre-existing cases. Two paired cases pass under the revert: the configured default is still used when the workflow declares it (so this is not "always use intake"), and it is still used verbatim when nothing resolves.

175/175 across the plugin’s 19 files, glasses typecheck clean, lint clean.

Pattern worth noting for the batch phase, since it has now bitten me three times in one file: **when you validate an input, check the default too.** The explicit path is the one you think about; the default is the one that runs.

gsxdsm:
## Added: the Command Center SDLC funnel read EMPTY on any renamed board — and my first fix was cosmetic

Found by applying this branch’s own lesson ("when you validate an input, check the default too") to the other settings-derived column defaults. `activity-analytics.ts` had the same shape, worse: the default was not merely unvalidated, it was the **only** path.

**The defect.** The funnel maps column ids to SDLC stages by trait and folds any id it was not given into OTHER. Its own doc comment says callers with a custom workflow "should call `aggregateSdlcFunnel` directly with that workflow’s columns" — and **no caller does.** `/command-center/activity` and the OTel exporter both call `aggregateActivityAnalytics` with no `columns`. So on a renamed or custom board every id is unknown, and the funnel reads empty while the board is plainly busy.

**My first fix was cosmetic and I described it as the whole thing.** I switched `defaultColumns()` from the legacy monolithic `BUILTIN_CODING_WORKFLOW_IR` to `resolveDefaultWorkflowIr()` and wrote that it fixed renamed boards. It does not — post-U11 the current lineage’s ids are a **subset** of the legacy constant’s, so the legacy fallback already mapped everything the current one does.

**The give-away was a revert proof that would not go red:** my assertions passed with the legacy constant restored. That is the fourth vacuous test I have caught in this branch, and the first one I wrote *after* documenting the pattern — which is worth saying out loud, because it means the doc is not sufficient protection on its own. The consistency change stays, now labelled consistency-only at the site with the reason it cannot have a behaviour-revealing test.

**The real fix is the caller.** The route resolves the project’s default-workflow columns and passes them. `resolveFunnelColumnsForProject` returns `undefined` — not `[]` — when nothing resolves: undefined falls back to the built-in default (previous behaviour), while an empty array would map every column to OTHER and silently empty the funnel for everyone. Pinned in a test, since that is exactly what a later cleanup "simplifies".

Six cases, including two I got wrong first: `archived` is excluded from the every-column-maps assertion (my first version failed on it and I nearly read that as a product bug — archived is not an SDLC stage), and the merged planning column maps to the **triage stage** via its intake trait, which is precisely what the legacy fallback could not express.

6/6 new, `pnpm test:gate` 71/71, core + dashboard typechecks clean, lint clean.

gsxdsm:
Fixed — and this is the **third time in this branch** I have split a workflow snapshot, the second time *after* fixing it elsewhere.

The declared-column set and the intake fallback each resolved the workflow independently, so a workflow edit between them validated against one revision and selected the intake column from another — persisting a card in a column the validated revision does not declare. `resolveCaptureBoard` now returns both answers from one resolution.

The three occurrences: the executor’s resume lanes (#2640 review), the glasses lane context (earlier in this PR), and now capture. The shape is identical every time — **two helpers that each look correct, called in sequence, each doing its own read.** Documenting the individual fixes plainly did not stop me repeating it, so stating it as a rule instead: *if a function needs two facts about one workflow, it needs one snapshot, not two helpers that happen to agree most of the time.*

The test is a **read-count assertion**, not prose, because that is the only thing that fails when someone reintroduces the split. It covers the path that needs both answers: a configured default the board does not declare, so the declared set rejects it and the intake column lands the card.

Revert proof: reintroducing a second read fails **2 of 20**. Both cases also assert the resulting column, so a change that deduplicated the read while getting the answer wrong would still fail.

177/177 across the plugin’s 19 files, glasses typecheck clean, lint clean.

gsxdsm:
Both threads addressed — but one of them I could **not** prove, so there is no fix for it and I want to say that plainly rather than ship something unfalsifiable.

**1. Stale approval hid a running card** — real, fixed. Dropping the column condition made the awaiting-approval signal status-only, which is right for a planning-lane card. But `awaiting-approval` is **durable**, so a card carrying it into an execution lane rendered as not-active: no active styling, no execution-status indicator, no current-step metadata, while plainly running. The suppression now yields to an execution **signal** rather than a column name, which needs no IR lookup — this is a synchronous React render. The `in-progress` literal there is pre-existing, not on the triage bar, and needs traits this component is not given; left with a note rather than half-converted.

**2. `resolveWorkflowIrById` substitutes the builtin default — right, and I took my fix back out.** You are correct that a missing or unparsable custom workflow silently yields the builtin IR, whose columns then pass as the project’s vocabulary. I wrote the guard, then could not establish a better answer for that state:

- legacy-five fallback **accepts `triage`** — the column #2515 deleted, i.e. creating a card in a column no board has, which is a defect I fixed earlier in this same branch;
- accepting nothing rejects every explicitly-named column on a project whose workflow row is merely missing;
- refusing the capture loses the operator’s utterance.

**The tell was an isolated revert that stayed green.** Two candidate behaviours my tests could not distinguish means I had no evidence, and a change I cannot make fail is a change I cannot justify. So the substitution is named at the site and the decision is left to whoever owns the missing-workflow contract — it is a product question, not a plugin one. What I kept is the provable half: the definition row that was read **is** the snapshot, so the custom path stays at one read (enforced by the existing read-count cases).

**3. Also in this push — a fourth snapshot split, in already-merged code.** Found by auditing my own work for the shape I hit three times in this PR (grep for two resolver calls inside one function). `mission-feature-sync.ts` read the workflow twice: `resolveTaskLifecycleColumns` for the roles and `resolveWorkflowIrForTask` for the declared ids — literally the same read twice, since the former is the latter plus `resolveLifecycleColumns`. A workflow edit between them gives roles from one revision and declared columns from another, so the aliasing guard is evaluated against a column set that no longer matches the roles it protects. Now one snapshot.

178/178 across the glasses plugin’s 19 files, 22/22 mission-feature-sync, dependency-graph typecheck clean, `pnpm test:gate` 71/71, lint clean.

gsxdsm:
## Coordinator item 3 done: `replan-target.ts` fallbacks resolved

`resolveReplanTargetColumn`’s no-match path returned the literal `"triage"`. For a workflow declaring neither `triage` nor `todo` — builtin marketing (ideation/backlog/drafting/…) and any custom board shaped like it — that names a column the workflow does not declare **and** the default lineage no longer has, so the replan move either threw or parked the card in a phantom column. It now falls back to that workflow’s own planning lane: **hold first, then intake.**

**The old rationale was stale, which is why this is a behaviour change and not a rename.** The comment read *"a custom entry column would strand the needs-replan card (triage only scans triage and todo)"*. Triage discovery now selects planner lanes **by trait**, so a card in the workflow’s hold column *is* scanned. The conversion work removed the reason the literal existed, and keeping it afterwards is what turned a safe fallback into a phantom target. Worth stating plainly for the remaining conversions: **a guard whose justification was deleted elsewhere does not announce itself.**

**Hold before intake** is the discriminator the old note said was missing. Coding (Ideas) declares `ideas` as a *manual* intake (`autoTriage: false`) and plans in its hold column, so a replan landing in raw-ideas would sit behind a promote gate no engine path can open. The two id lookups stay ahead of the roles because they still name the planning lane on the workflows that declare them (Lead generation, PR review) — so every existing board’s target is unchanged; this only adds an answer where there was a wrong one.

The literal survives **only** where nothing resolves — a column-less/v1 IR, or a throwing resolution — and both now carry `DELIBERATE-LITERAL` markers, so the census counts them as reviewed rather than backlog.

**A pre-existing test asserted the defect**, with the stale rationale in its comment: *"falls back to triage for workflows declaring neither triage nor todo (never a custom column)"*. Replaced with the post-conversion truth, plus a case pinning that a column-less IR still yields the legacy id so a later cleanup cannot collapse the two paths.

Revert proofs, isolated:
- role fallback removed → **1 of 41 fails** (marketing → `triage` again)
- `intake ?? hold` instead of `hold ?? intake` → **1 of 41 fails** (marketing → `ideation`, the manual-intake trap)

That second proof matters as much as the fix: the two orderings look equivalent and one of them parks replans behind a manual gate.

51/51 across the three replan-target suites, `pnpm test:gate` 71/71, engine typecheck clean, lint clean. Census for the file: 4 column guards left, all `in-progress`/`in-review`/`done` advancement checks — off the triage bar.

**On item 2 (the ratchet is a ceiling):** you are right, and it is my tool. Two notes for whoever implements it, from having just been bitten by both directions — (a) the baseline must be rewritten *downward automatically*, not by an instruction humans follow, because eleven conversions merged without re-recording and none of those authors did anything wrong; (b) `--strict --update-baseline` already rewrites unconditionally and prints `ACCEPTED RISES`, so the missing piece is only wiring the auto-tighten into the test path rather than new mechanism. Happy to take it if U12 would rather not.

gsxdsm:
## Item 2, the +3 investigation: measured it is **+1**, and it is justified

Method: diff the per-file column-guard counts on current `origin/main` against the baseline as recorded at #2633’s merge (`8e211d187`), then locate and blame each added site. Counts only — never line numbers, for a reason that turns out to matter.

**Result: one genuine new column guard since the baseline was recorded.**

| file | was | now |
|---|---:|---:|
| `dashboard/app/utils/columnRoles.ts` | 0 | **1** |

```ts
export function isHoldColumnRole(flags: ColumnRoleFlags | undefined, columnId: string): boolean {
  return flags ? flags.hold === true : columnId === "todo";
}
```

Added by **#2647** (U12 consolidation). It is a **scoped legacy fallback sitting beside a trait check**, documented at the site with why its degraded id set is `todo` alone rather than `{todo, triage}`. That is the residue of a *correct* conversion, not drift.

**Where the other apparent rises went.** `register-task-workflow-routes.ts` looked like 22 → 23, and it did not move: it is **23 both before and after**. The "22" came from a baseline I re-recorded while on my own branch, where I had converted a site — my bookkeeping error, not a merge. And a naive diff shows six "new" guards in that file which are the *same* guards shifted by an insertion above them. Counting per file rather than per line is what makes the difference; anything that keys on line numbers will report phantom rises all through the batch phase.

**The consequence for item 2, and it is the useful part.** A conversion **legitimately adds a literal fallback** — `flags ? flags.x : columnId === "legacy"` is the correct shape when the caller may have no traits, and every conversion of that kind increments the census by one. So a hard ceiling *will* fire on correct work, repeatedly, once the fleet starts. Three options for whoever implements the auto-tighten:

1. count `besideTraitCheck` sites in a separate class (my `--file` mode already computes it; the PR-2644 census exports it per finding) so fallbacks-beside-traits do not count against the bar;
2. accept rises that are `besideTraitCheck` and fail only on bare guards;
3. keep one number and expect the baseline to be re-recorded on most conversion PRs — which is the status quo that produced this thread.

I would take (1): it is the only one where the number means "unconverted guard" rather than "literal", and it costs nothing new — the classifier already distinguishes them.

Analysis only, no code in this comment; item 2 is U12’s and I did not touch the ratchet path. My offer to implement it stands if they would rather not.

gsxdsm:
All four cleared — one was a **major regression of mine**, two were fixtures that could not fail, and one I could not fix and did not pretend to.

**1. Degraded blocked the plain failure-retry (major, mine).** I put the refusal at the top of `retryTask`, which also gated the branch below — one that reads no lanes and only clears status. So a FAILED card became un-retryable whenever its workflow definition could not be read, which is exactly the state an operator is trying to retry out of. The refusal exists to stop a *move* onto another board’s vocabulary; a branch that performs no move has nothing to be wrong about. Revert proof: refusal back at the top → **2 of 50 fail** (the failed-card retry, plus the paired case that the planning-lane retry *is* still refused).

**2. The unresolved-workflow fixture was identical to the default-lineage one.** Both used `harness(undefined, …)`, which supplies no workflow *selection* — so the second test could pass without ever exercising a selection whose definition lookup fails. Two tests with identical fixtures are one test with two names. It now names a workflow whose definition read **throws**, the state that actually reaches the resolver’s fail-soft path in production.

**3. Marker-to-marker slicing instead of magic-number windows.** Taken — those windows were numbers I had already widened once when a comment grew, and a bound that needs tuning gets tuned until it asserts nothing. Two of my own mistakes are recorded at the site: I first asserted `process.exit(1)` for the update-baseline branch (it exits **0** — that is the point), and my first end marker (`if (regressions.length > 0) {`) also appears *inside* that block, so the slice cut the branch short and failed on a boundary rather than on behaviour. **Marker-to-marker is only robust when the marker is unique**; a repeated marker is the magic-number problem wearing a name. Revert proof: moving `--update-baseline` after the rise exit → **2 of 26 fail**.

**4. The string-`ir` second read — real, and I could not fix it.** Your reading is right and my excuse ("deliberate and confined to that shape") was not a reason. I tried parsing the row instead of re-resolving: the parse succeeds in isolation and `resolveLifecycleColumns` returns the correct roles for the parsed IR (verified directly), but the action still refused in the plugin harness — something between the parse and the lane check differs from the resolver path and I could not identify it within this PR. I will not ship an unexplained change into the code path that decides whether an operator’s action is **refused**; that trades a bounded drift window for an unbounded one.

The second read therefore stays, **named at the site rather than excused**, with the window stated: it needs a workflow edit between two reads microseconds apart, and the failure mode is a refusal or a stale lane set, never a wrong move onto another board. A store that returned parsed IRs removes it entirely — that is where the fix belongs. What is now pinned is the current contract for that shape, which had **no coverage at all**: a string-ir board still resolves lanes and the action proceeds, which any future one-read fix must keep.

181/181 across the plugin’s 19 files, 26/26 census contract, 4/4 rebound, `pnpm test:gate` 71/71, glasses + engine typechecks clean, lint clean. Rebased onto current `origin/main`.

gsxdsm:
## Rebased onto main; yielded the census tooling rather than compete for it

#2650 has evolved the census substantially — query ratchet, `queryByFile`, `properties`, plus the role/status classification this branch introduced, now absorbed. Replaying my five commits produced conflicts in the same file every time, so I **took main’s version wholesale** for the CLI, the AST lib, the contract test and `docs/testing.md`. Two workers reimplementing one file is the waste consolidation mode exists to stop, and the ratchet path is U12’s by assignment. My branch keeps only the baseline re-recorded against this tree (`executor.ts` 104 → 87, from this branch’s own conversions).

**Three things main now lacks. The first is a live problem, not a preference, because #2654 gates CI on the ratchet:**

1. **`--update-baseline` cannot run when a file rose.** The rise check exits at line 206; the update-baseline block is at 210. So the only supported way to re-record is unavailable in exactly the situation that needs it — and a conversion *legitimately* adds a literal fallback (`flags ? flags.x : columnId === "legacy"`), which raises the count by one. A blocked worker’s only path is hand-editing the JSON, which is how a ratchet becomes something people bypass.
2. **An unrecorded drop fails hard.** Eleven files dropped in the recent merge wave and none of those PRs re-recorded; each one turns the gate red for everyone else. A permanently-red gate gets ignored — a worse hole than the stale allowance it protects against. My branch had this as a warning, with `--strict --exact` for the pinned end state.
3. **No per-file work order.** `--file <path>` printed each site with receiver, source line, and a "sits beside a trait check" hint, so a batch worker converting a 111-guard file gets a site list instead of a number and can separate **deletions** from conversions. Measured on `TaskContextMenu.tsx`: 11 guards, 4 of them fallbacks → **7 real conversions, not 11.**

All three are implemented and revert-proofed in this PR’s earlier pushes if the owner wants to lift them. I am not re-landing them on top of a file another worker is actively changing.

**Also dropped: my column-less replan test.** Main went further than the fallback conversion — U5 now returns **undefined** for a workflow declaring no planning lane ("skipped with a log rather than moved arbitrarily"), which is better than any legacy-id answer. My test asserting `"triage"` would have contradicted it, so main’s case is kept verbatim and the changeset corrected to describe what shipped.

Verified after taking main’s tooling: `--strict` exits 0, 51/51 census suites, 44/44 replan-target, 181/181 glasses plugin, `pnpm test:gate` 71/71, lint clean.

gsxdsm:
Both threads are already addressed in the pushed tree — one fixed, one recorded as unfixable-by-me with the evidence. Restating so they can be resolved.

**`agent-actions.ts:333` — degraded blocked the plain failure-retry (major).** Fixed. The refusal now gates only the lane-dependent branch (`retryDegraded === false && isInPlanningLane(...)`). A FAILED card is retryable again even when the workflow definition cannot be read — which is exactly the state an operator is trying to retry out of. The refusal exists to stop a *move* onto another board’s vocabulary; that branch performs no move.

Revert proof: refusal back at the top of `retryTask` → **2 of 50 fail** (the failed-card retry, and the paired case that the planning-lane retry *is* still refused, so this is not "stop refusing").

**`agent-actions.ts:190` — the string-`ir` second read.** Real, and I could not fix it. Your reading is right and my earlier excuse ("deliberate and confined to that shape") was not a reason. I tried `parseWorkflowIr(definition.ir)` to keep it to one read: the parse succeeds in isolation and `resolveLifecycleColumns` returns the correct roles for the parsed IR (verified directly), but the action still refused in the plugin harness — something between the parse and the lane check differs from the resolver path, and I could not identify it within this PR.

I am not shipping an unexplained change into the code path that decides whether an operator’s action is **refused**; that trades a bounded drift window for an unbounded one. So the second read stays, with the reasoning restored at the site (the rebase had dropped that comment — pushed just now, since the comment *is* the deliverable for this finding). The window: a workflow edit between two reads microseconds apart, and the failure mode is a refusal or a stale lane set, never a move onto another board’s column. A store that returned parsed IRs removes it entirely — that is where the fix belongs, and it is noted at the site with that owner.

The contract for that shape is now pinned, which it was not before: a string-ir board still resolves lanes and the action proceeds. Any future one-read fix has to keep that.

181/181 across the plugin’s 19 files, glasses typecheck clean, lint clean, rebased on current `origin/main`.

gsxdsm:
This one is **already fixed** in the commit the review ran against — checked rather than assumed, because "already fixed" is exactly the claim that deserves evidence.

Verified on `d5b3347` (the `commit_id` on the review comment itself):

- `resolveCaptureIntakeColumn` — **0 occurrences**. That function is gone; the proposed shape refers to a revision two commits earlier.
- `getDefaultWorkflowId?.()` — **1 occurrence**, inside `resolveCaptureBoard`.
- `normalizeCaptureColumn` calls `resolveCaptureBoard(taskStore)` **once** and takes both answers off it: `board.declared` for validation, `board.intake` for the fallback destination.

It was fixed in `155e70f1` ("I split the workflow snapshot a THIRD time — capture now resolves the board once"), which was itself a response to this same finding on an earlier push. Your diagnosis was right both times; the second report is anchored to stale content.

The guard against regressing it is a **read-count assertion**, not prose — `getWorkflowDefinition` is counted and must be exactly 1, on the path that needs *both* answers (a configured default the board does not declare, so the declared set rejects it and the intake column lands the card). Revert proof from that push: reintroducing a second read fails **2 of 20**, and both cases also assert the resulting column so a dedupe that got the answer wrong would still fail.

21/21 quick-capture, 181/181 across the plugin’s 19 files.

For the record on this PR overall: this was the **fourth** instance of the two-reads-one-decision shape in my work here (executor resume lanes, glasses lane context, capture, and `mission-feature-sync.ts` in already-merged code). All four are now single-snapshot, and the mechanical check that found the merged one — grep for two resolver calls inside one function — is cheap enough to be worth running over the batch-phase PRs.

PR #2645, opened 2026-07-30
consolidate/u11: a hold column is a planner lane only if it precedes wip (real defect in merged code) + funnel aliases (triage 11 -> 10)

gsxdsm:
**Consolidation branch for u11/u7.** Supersedes #2624. Contents changed substantially while it sat unmerged — this body reflects what is actually in it now.

## Measured with the authoritative census, not grep

`node scripts/lifecycle-column-census.mjs` — **triage guards 11 → 10.**

## 1. A real defect in merged code: a hold column is only a planner lane if it precedes implementation

Found by greptile on #2616, verified by me, fixed here at the source because that PR cannot land.

`resolveLifecycleColumns` returns `hold` as the **first** hold-trait column in declared order, with no positional constraint relative to wip (`workflow-lifecycle-traits.ts`: `hold: first(LIFECYCLE_ROLE_FLAGS.hold)`). A workflow using a hold trait for a **mid-pipeline wait** — a pause after implementation starts — therefore had that column returned as its planner lane, and `reconcileMissionFeatureState` demoted the feature to `triaged`. The mission board reported started work as not-yet-started: silent, and wrong in the direction that makes a roadmap lie.

This is my defect, introduced in #2610.

**Why it survived:** every lineage anyone has tested puts the hold *in front* of wip, so the default and Ideas boards are unaffected and no existing test could see it.

**The fix is positional, with a deliberate asymmetry.** A hold column counts only when it appears before wip in declared order. When wip cannot be located the hold is left **out** rather than guessed — including it wrongly demotes live work on the roadmap, while excluding it wrongly costs only a `triaged` transition the next reconcile re-applies.

Mutation-verified: dropping the positional test fails the mid-pipeline case and nothing else.

## 2. MissionControlPanel funnel aliases

Assessed and **deliberately not trait-converted**. These are heuristic *name aliases* for a canonical SDLC stage — the matcher already accepts `signal`/`backlog`/`ready`/`shipped` because it buckets arbitrary boards, with an `other` fallback. Post-#2515 a default board's planning cards sit in `todo` and count at the Todo stage, leaving Planning at zero: the funnel reporting where cards *are*, not a guard that stopped firing. Hoisted to a named set so it stops reading as unconverted.

This is the **DISPLAY-ALIAS** class the census still lacks — receiver *is* a column id, purpose is presentation rather than a lifecycle decision. `DocumentsView`'s status dot is the other one. Without that bucket a ratchet will keep demanding conversions that make the product worse.

## What I dropped, because main's version was better

The original #2624 carried a `TaskContextMenu` conversion. #2626 landed `isPureIntakeColumn` — intake **without** hold — while mine treated any intake-flagged column as intake. That's wrong for a **merged Planning column**: it carries both traits, cards there wait for capacity and have real actions, so I would have suppressed the menu where it belongs — a new regression in place of the one I was fixing. Theirs is correct. Mine is gone, along with its now-invalid test and a helper nothing else used.

## Verification

- merge gate green (482 + 132 + 10), engine tsc clean, dashboard tsc clean, lint clean
- 7 planner-lane tests green, mutation-verified

No changeset: `@fusion/engine` and `@fusion/dashboard` are private.


gsxdsm:
All three findings were real and are fixed in `17557072d`.

**1. Major — traits must outrank the legacy id.** Correct, and the tell was internal: the sibling `isPreExecutionIntakeLane` five lines above already resolved traits first (`if (flags) return ...; return isLegacyIntakeColumn(column)`) while `isPreExecutionHoldColumn` kept the id as an unconditional disjunct. Two adjacent predicates answering the same question differently is what made `{ intake: false, hold: false }` on `triage` reachable. Both now have the same shape; the legacy id survives strictly as the no-metadata fallback, which is the case that must keep working — a card stranded on a column its workflow no longer declares has no flags at all, and dropping it out of the pre-implementation lane would silently remove its affordances.

**2 + 3. Minor — caller-level and explicit-false coverage.** Both added. They were the same blind spot from opposite ends: the suite tested the classifier but not the caller, and tested positive traits and `undefined` but never explicit `false` — which is exactly the combination the Major finding exploited.

One correction to the framing in finding 2, worth recording because it is easy to get backwards: `shouldShowActionsMenu` is driven by `isPreExecutionIntakeLane`, **not** by `isPreExecutionHoldColumn`. The hold predicate drives the **Plan action** inside the menu. So `shouldShowActionsMenu` cannot observe this bug at all — asserting on it alone would have produced a caller-level test that passes with the defect intact. The regression is pinned on the Plan action, and `shouldShowActionsMenu` is covered separately on both arms.

I found this the honest way: my first caller assertion (`shouldShowActionsMenu === true` for a merged Planning column) failed, and the code was right — post-U11 suppression on an intake lane is exactly what the `!== "triage"` → role conversion restored.

**Revert check.** Restoring the unconditional disjunct fails exactly 2 of 15 — the classifier case and the Plan-action case. The other 13 stay green, including the no-flags fallback.

**Pre-existing failures in the neighbourhood, NOT from this branch** (verified by reverting this branch's `shouldShowActionsMenu` line to main's literal and re-running — all 3 persist):
- `TaskCard.badge-wrap.test.tsx` › three-dot controls aligned across card states
- `TaskCard.test.tsx` › FN-4511 GitHub badge / timer chip geometry parity
- `TaskCard.test.tsx` › renders a promote action when onPromote is provided

Plus `TaskContextMenu.test.tsx` › `uses VALID_TRANSITIONS and in-review back-to-progress labels`, which expects `"Back to In Progress"` and receives `"Back to in-progress"`. That is a stale expectation from the trait-derived-label conversion already on main: the label now interpolates `columnLabel(column)`, and the test's stub returns the raw id. This branch does not touch label logic. Flagging it rather than fixing it here — it belongs to whoever owns that conversion, and it is a test-side fix, not a product bug.

Measured: 15/15 in the guards suite, 684/688 across the 6 suites that touch this menu (the 4 above), dashboard `tsconfig.app.json` typecheck clean.

PR #2647, opened 2026-07-30
consolidate/u12 — U12 consolidation: 4 live defects, the AST ratchet fail-closed, and the moves.ts flag scoped

gsxdsm:
One branch, one PR, per the consolidation directive. Contents file-by-file below.

**Supersedes #2625** (its overlapping conversions landed via U11's #2624/#2626/#2636; only the parts nobody else did are folded here). **#2630 and #2639 stay open** — both green with zero threads, per rule 3.

## Four live defects, each measured

**1. Every planning card renders an actions menu.** `TaskContextMenu.tsx` still had `shouldShowActionsMenu: task.column !== "triage"` on main *after* the rest of that file was converted. Since #2515 removed the id, the condition is TRUE for every card, so the suppression stopped applying anywhere — including on cards whose menu is empty, the orphaned click target the Surface Enumeration rule exists to catch.

Found **twice independently**: by reading the guard, and again by the invariance test below, which failed on main with `shouldShowActionsMenu` true on one lineage and false on another. That is the argument for an invariance property over per-site conversion — the file had already been converted "2 → 1" and the survivor was the live one.

**2. Worktree upcoming-work list empty on renamed boards.** `groupByWorktree` filtered `t.column === "todo"`. On the default board the id and the role coincide so every existing test passed; renamed, it matched nothing and a whole panel read as idle.

**3. Hold-lane FIFO ordering lost on renamed boards.** `sortTasksForDisplayColumn` gated priority-then-FIFO on `column === "todo"`, degrading to the generic id-ordered sort elsewhere. Cards simply appear in the wrong order, silently.

**4. The AST ratchet still failed open** — fourth time in that file, third found by review. `receiverName` understood only one-level property access and bare identifiers, so `task["column"]`, `metadataColumn(entry, "to")`, ternaries, `(task!.column)` and backtick literals were dropped. **Measured on main: `in-progress` 196 → 197, `in-review` 211 → 213** — three real guards nobody counted, including `metadataColumn(entry, "to") === "in-review"` in `reliability-metrics.ts`. Now walks wrappers, resolves calls to the callee name, and emits a `<SyntaxKind>` **sentinel** for anything unnameable: counted *and* trips the classification guard, so a human judges it instead of it vanishing.

## Per-file guard counts

| file | before | after |
|---|---:|---:|
| `app/components/TaskContextMenu.tsx` | 1 | **0** |
| `app/utils/worktreeGrouping.ts` | 1 | **0** |
| `app/components/taskSorting.ts` | 1 | **0** |

The other dashboard files I had converted reached 0 via U11's PRs; where our work overlapped I took theirs during the rebase, including two places where theirs was **stronger** than mine — they deleted Column's unreachable quick-create arm outright (with fixtures migrated) where I had converted it, and they verified the same `isPreExecutionHoldColumn` degraded-set asymmetry I did, independently.

## Flip precondition: the moves.ts flag is scoped, not flipped

`move-target-declared-census.test.ts` answers precondition 2 with measurement. 41 engine `moveTask` calls have literal targets — `todo` 27, `in-progress` 7, `done` 6, `archived` 1 — and **all four are declared by the default lineage**, so the default board is not the exposure. `triage` appears only in a comment noting `replan-target.ts` used to hardcode it. My own grep had said `todo=29`; the AST says 27, because grep counts comments.

The exposure is **custom** lineages: 20 of the 41 carry no `recoveryRehome` and would reject with unknown-column post-flip; 21 are exempt via the #1411 carve-out, which makes that carve-out load-bearing.

I did not flip the flag. It is six seams, not the `789`/`837` pair every summary including mine described, and seam 2 turns on *new refusals* rather than swapping equivalent implementations — a green suite says nothing about that. #2639 pins the blast radius.

## Tests

- `column-role-id-invariance.test.tsx` — hold traits fixed, vary only the column id across MERGED / LEGACY / RENAMED; every decision must agree. Drives the real consumers, so a component keeping an inline comparison fails it. Includes a unanimous-and-**false** case so it can't be satisfied by a predicate hardwired to true. **This is the test that caught defect 1 on main.**
- `worktreeGrouping.test.ts` — includes two cards both in a column named `staging`, one hold and one not, asserting opposite answers. That assertion is impossible under a board-wide column-id set, which is why hold resolution is keyed per task via `getEffectiveTaskWorkflowId` (#2625 review).
- `taskSorting.test.ts` — discriminates on the **tiebreak**, not priority: both branches sort by priority, so my first version passed for the wrong reason. Equal-priority cards whose `createdAt` order disagrees with their id order.
- `no-hardcoded-lifecycle-columns.test.ts` — 16 detector cases: 11 shapes counted, 4 legitimate ignored, one asserting the sentinel path.

Revert checks, all run: menu suppression → diff names the field; worktree → `expected [] to include 'FN-50'`; sort → `FN-2, FN-9` instead of `FN-9, FN-2`; ratchet → the 3 recovered guards disappear.

## One site that should never be converted

`MissionControlPanel.tsx:46` — `{ id: "triage", match: (c) => c === "triage" || c === "signal" || c === "backlog" }` is a deliberate name-similarity heuristic for the SDLC funnel; it matches synonyms and folds unknown columns into an "other" bucket so custom columns still contribute. Converting it changes what the funnel displays. Like the `live-agent-count` fallbacks, it belongs in a documented floor — **the ratchet's target is that floor, not zero.**

`DocumentsView.tsx:73` is convertible but the file has no column flags at all, so a real fix means plumbing board-workflow metadata into a view that doesn't fetch it — its own unit of work.

## Verification

`pnpm lint` clean. `pnpm test:gate` green (10 / 482 / 71). `tsc -p packages/dashboard/tsconfig.app.json` and `packages/core/tsconfig.json` clean. Core ratchet + seam suites 24/24. Dashboard target suites 37/38 — the one failure is the pre-existing `"Back to In Progress"` label casing, confirmed identical on the base.

---

## Added after the initial push

**5. `TaskCard` lost inline editing on renamed boards; `TaskDetailModal` kept it.** Still live on main: the modal resolved field editability from traits in U10/R8, the card used a hardcoded `{triage, todo}` set with **no trait path at all** — even though `taskColumnFlags` was already in scope. On a renamed board the title was editable in the modal and the pencil was missing from the card. Body moved unchanged into `isFieldEditableColumnRole` so the two surfaces cannot drift again.

The veto traits are the substance: a column can legally carry `hold` **and** a WIP or review trait, and a plain `intake || hold` check would let an operator rewrite a description while a session executes against it.

Coverage gap **measured, not assumed**: mutating `canEdit` back to the hardcoded set left `TaskCard*` at the same failure count as the unmutated run — nothing caught it. The four render cases assert the real `aria-label`; that mutation now fails with `Unable to find an accessible element ... name 'Edit task'`.

**6. The ratchet's target is a documented FLOOR, not zero** — and this changes the completion bar.

Zero is not reachable, and chasing it means breaking working code. Two categories are permanent, now protected as positive assertions so a future sweep cannot "finish the job" by deleting them:

- `MissionControlPanel.tsx`'s `FUNNEL_STAGES` is a deliberate **name-similarity** heuristic — it matches `signal`, `backlog`, `to-do`, `ready`, `shipped` and folds unrecognised columns into an "other" bucket so a custom board still contributes counts. It is not asking whether a column has the intake trait; it buckets arbitrary column *names* for display. Asserted on the **synonym list**, because the synonyms are what prove it is name matching — if they disappear the site has changed character and the exemption stops applying.
- `live-agent-count.ts`'s no-flags arm is reachable (a remote store is deliberately given an empty flag map; a card in an undeclared column has no flags at all) and deleting the literal makes such a card match **no** arm, so the queued total silently under-reports a stranded card.

A count with an undocumented floor invites someone to drive it to zero.

**Not done, and why:** `DocumentsView.tsx:73` is convertible but that file has no column flags anywhere, so a real fix means plumbing board-workflow metadata into a view that does not fetch it — its own unit of work, not something to smuggle into a conversion.

**Re-verified after these commits:** `pnpm lint` clean, `pnpm test:gate` green (10 / 482 / 71), `tsc` clean on core and `tsconfig.app.json`, core ratchet suite 26/26, `columnRoles` 10/10, `TaskCard.test.tsx` 384/386 (the 2 are pre-existing CSS assertions). `TaskDetail*` is 130 failed / 551 passed **both with and without** this change — verified by stashing, so pre-existing and unrelated.

---

## Flag resolution: preconditions 1 and 2 are now DISCHARGED. Precondition 3 is blocked, and by evidence.

**Precondition 1 — the side-effect equivalence proof — done.** `moves-flag-equivalence.test.ts` runs the same journey under both flag states against live PG and diffs the persisted row. **Result: identical** — whole-row equality across 128 fields plus an equal timing shape, over `todo → in-progress → in-review → todo → in-progress`.

That test was **wrong twice** before it meant anything, and both times it was passing:

1. **It proved nothing.** `experimentalFeatures` is **global-only**, and `moves.ts` reads `getSettingsFast()`, which filters global-only keys out of the project layer. My `updateSettings` write was silently discarded, `useWorkflow` was false in *both* runs, and the "proof" compared the legacy path against itself. Found by stamping the flag-ON branch and observing the test still passed. Now written via `updateGlobalSettings`, and the helper **asserts the flag took effect** before the journey runs.
2. **The journey was forward-only**, so it never reached the reopen hook's field resets (`status`, `error`, `blockedBy`, pause clearing) — a mutation there passed. Extended with a backward move and a re-entry.

Mutation-verified after both fixes: stamping seam 3, and diverging the reopen hook, each fail the comparison.

**Precondition 2 — done, and its answer is a blocker.** The census says the default board is safe: all 41 literal engine move targets are declared by the default lineage. But **20 of those 41 carry no `recoveryRehome`**, so on a custom lineage that does not declare `todo` / `in-progress` / `done`, seam 2 would start rejecting them with unknown-column. That is a user-facing break on custom boards, not a theoretical one, and it is not fixed by the equivalence proof — seam 2 adds *new refusals* rather than swapping implementations.

**So the flip is one step away, and the step is not mine to take alone:** those 20 call sites need to resolve their target from the task's workflow (or justify `recoveryRehome`), and they live across engine lanes in `moves.ts` caller territory — U2b/MAIN. Flipping before that trades a dormant flag for broken custom boards.

What remains for precondition 3 once those land: flip both readers **atomically** (`moves.ts` + `workflow-task-create-ops.ts`, since the latter computes the preflight the former consumes), delete the flag-OFF branch with its guards, and drop the settings key.

---

## CORRECTION: seam 2 is not a blocker. My earlier claim was wrong.

I stated in #2639 and above that "with the flag off there is **no** target-column validation on the move path", so flipping would introduce new refusals. **That is not what happens.** Reproduced against live PG: the identical custom-lineage move rejects with the flag **OFF** as well —

```
Error: Invalid transition: 'backlog' -> 'todo'. Valid targets: building
```

Transition validation is already in force on the flag-OFF path. So for the shape in question — an engine move to a column the task's own workflow does not declare — **the move already fails today**, and seam 2 introduces no new break for it. The 20 census sites lacking `recoveryRehome` are broken on a custom lineage *now*, not broken by the flip.

I found this because the discriminator I added to prove "the flag is the cause" failed. Had I written the test to my assumption it would have passed and the false claim would have shipped — the same way the equivalence test passed while proving nothing until I tried to make it fail.

**Revised precondition status:**

| precondition | status |
|---|---|
| 1 — side-effect equivalence | **discharged** — identical rows, mutation-verified both directions |
| 2 — seam-2 exposure census | **discharged, and it is not a blocker** — the rejection predates the flag |
| 3 — flip both readers atomically, delete the flag-OFF branch, drop the settings key | **the remaining work** |

So the flip is no longer gated on fixing 20 engine call sites. What it is still gated on is precondition 3 being done atomically across `moves.ts` and `workflow-task-create-ops.ts` (the latter computes the preflight the former consumes), which is `moves.ts` caller territory.

Three cases now cover seam 2: the flag-ON rejection, the flag-OFF rejection (asserting the error *message*, so a change in which guard rejects stays visible rather than reading as agreement), and the #1411 `recoveryRehome` carve-out succeeding — pinning why that carve-out is load-bearing and must not be tidied away.

### DIFF
--- a/.changeset/u12-dashboard-column-roles.md
+++ b/.changeset/u12-dashboard-column-roles.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Fix board affordances that broke on renamed or merged column lineages.
+category: fix
+dev: Column-role helpers move to @fusion/core (column-roles.ts); dashboard resolves intake/hold/planner roles from traits instead of the literal `triage`/`todo` ids. Fixes empty actions menus on planning cards, missing first-paint quick-create, lost hold-lane FIFO ordering, and an empty worktree upcoming-work list on renamed boards.
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -49,7 +49,7 @@
     "test": "vitest run --silent=passed-only --reporter=dot",
     "test:embedded-postgres": "vitest run src/__tests__/postgres/embedded-lifecycle.test.ts --silent=passed-only --reporter=dot",
     "test:pg-gate": "vitest run --config vitest.pg.config.ts src/__tests__/postgres/handoff-to-review-atomicity.pg.test.ts src/__tests__/postgres/task-lifecycle-e2e.pg.test.ts --silent=passed-only --reporter=dot",
-    "test:unit-gate": "vitest run src/__tests__/task-merge.test.ts src/__tests__/legacy-adoption.test.ts --silent=passed-only --reporter=dot"
+    "test:unit-gate": "vitest run src/__tests__/task-merge.test.ts src/__tests__/legacy-adoption.test.ts src/__tests__/no-hardcoded-lifecycle-columns.test.ts --silent=passed-only --reporter=dot"
   },
   "devDependencies": {
     "@earendil-works/pi-coding-agent": "0.82.1",
--- a/packages/core/src/__tests__/moves-flag-equivalence.test.ts
+++ b/packages/core/src/__tests__/moves-flag-equivalence.test.ts
@@ -1,0 +1,194 @@
+/*
+FNXC:WorkflowColumns 2026-07-31-02:00 (U12 — precondition 1 for flipping the move-path flag):
+DO THE TWO COLUMN-SIDE-EFFECT IMPLEMENTATIONS AGREE?
+
+`moves.ts` gates six behaviours on the raw compatibility flag (#2639 pins all six). Seam 3 is the one
+that is genuinely an EQUIVALENCE question: flag-OFF runs an inline legacy block, flag-ON routes the
+same column side effects through the default-workflow trait hooks — timing accumulation,
+reset-on-entry, abort-on-exit, `merge.onEnter`. Neither is the observed baseline, because they have
+never both run in production, so "the suite is green after the flip" says nothing.
+
+WHY THIS IS BUILDABLE NOW, which I had assumed it was not. The flag reads
+`settings.experimentalFeatures.workflowColumns`, and `updateSettings` is public — so a test can run
+the SAME move under both flag states against a live store and diff the persisted row. No production
+change, no mock of the thing under test.
+
+WHAT IT COMPARES. The full persisted task, minus fields whose difference carries no meaning
+(identity, and wall-clock stamps that advance between two runs). Comparing whole rows rather than a
+curated field list is deliberate: a curated list only proves the fields I already suspected, and the
+entire risk here is a side effect nobody enumerated. `moves.ts` mutates ~15 fields in that branch.
+
+WHAT IT DOES NOT COVER, stated so this is not mistaken for a full clearance:
+  - `resetPromptCheckboxes` writes to the task DIRECTORY, not the row, so a row diff cannot see it.
+  - Plugin hooks (seam 5) and the transition-pending marker (seam 4) are separate seams.
+  - Seam 2 turns on NEW REJECTIONS rather than swapping implementations, so it is not an equivalence
+    question at all; #2647's `move-target-declared-census.test.ts` measures that exposure instead.
+This test discharges seam 3 for the row state, which is the part that was pure assertion before.
+*/
+import { afterAll, afterEach, beforeAll, beforeEach, expect, it } from "vitest";
+import { pgDescribe, createSharedPgTaskStoreTestHarness } from "../__test-utils__/pg-test-harness.js";
+import type { TaskDetail } from "../types.js";
+import type { TaskStore } from "../store.js";
+
+/**
+ * Fields whose difference between two runs is meaningless: identity, and stamps that advance with
+ * wall clock. Everything else must match, including the timing ACCUMULATORS (`cumulativeActiveMs`),
+ * which are the interesting part — they are computed from deltas, so a divergence in how the two
+ * implementations anchor a segment shows up there rather than in a raw timestamp.
+ */
+const VOLATILE_FIELDS = new Set([
+  "id",
+  // Per-task UUID; carries no behavioural meaning.
+  "lineageId",
+  "createdAt",
+  "updatedAt",
+  "columnMovedAt",
+  "executionStartedAt",
+  "executionCompletedAt",
+  "firstExecutionAt",
+  "cumulativeActiveMs",
+  "log",
+]);
+
+/*
+Wall-clock noise is normalised RECURSIVELY rather than by a flat key list, because it is nested:
+`columnDwellMs` is a column -> milliseconds map and run-audit-ish entries carry their own `observedAt`.
+A flat list missed both, and the first run of this test reported them as divergences.
+
+Durations become BOOLEANS (`>0`) rather than being dropped: whether time was attributed to a column at
+all is exactly the behaviour under test, while the millisecond value differs between any two runs.
+Dropping them would have hidden a real divergence; comparing them would have been permanently flaky.
+*/
+function normalize(value: unknown, key?: string, taskId?: string): unknown {
+  /*
+  IDENTITY is substituted inside strings rather than the field being dropped. `prompt` embeds the task
+  id (`# KB-001` vs `# KB-002`), so a raw comparison always fails and dropping it would stop comparing
+  the spec content entirely — which is one of the things the reset-on-entry side effect can touch.
+  Replacing the id keeps the content under test.
+  */
+  if (typeof value === "string" && taskId) return value.split(taskId).join("<TASK_ID>");
+  if (typeof value === "number" && key !== undefined && /Ms$/.test(key)) return value > 0;
+  if (Array.isArray(value)) return value.map((entry) => normalize(entry, undefined, taskId));
+  if (value && typeof value === "object") {
+    const out: Record<string, unknown> = {};
+    for (const [k, v] of Object.entries(value as Record<string, unknown>)) {
+      if (VOLATILE_FIELDS.has(k) || /At$/.test(k)) continue;
+      // A duration MAP: keep the keys, reduce each value to "time was attributed here".
+      out[k] = /Ms$/.test(k) && v && typeof v === "object" && !Array.isArray(v)
+        ? Object.fromEntries(Object.entries(v as Record<string, unknown>).map(([ck, cv]) => [ck, typeof cv === "number" ? cv > 0 : cv]))
+        : normalize(v, k, taskId);
+    }
+    return out;
+  }
+  return value;
+}
+
+function comparableSnapshot(task: TaskDetail): Record<string, unknown> {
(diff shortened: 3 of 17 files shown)


gsxdsm:
Real, and fixing it exposed something worse — so this thread ends with two changes, not one.

**The finding itself.** The test was not in the blocking gate, so a lifecycle-column regression could merge before anything reported it. Same class as #2494, whose characterization test was never in the `engine-core` allow-list. Admitted to `packages/core` → `test:unit-gate`, which costs ~4s on a lane that already runs in parallel with the much longer engine suite (unit-gate goes 158 tests / 5.26s total).

**What admitting it revealed.** I then tested the thing the finding is really about — can a regression get through — by reintroducing `task.column === 'triage'` and running the blocking gate. **It stayed green at 158 passed.**

The ceilings carried slack: `triage` 11 measured against a ceiling of 37, so **26 free slots**, plus 18 for `todo`, 4 for `in-progress`, 4 for `in-review`. A ceiling with slack does not ratchet — it silently permits exactly as many regressions as the gap. Being in the gate was necessary and not sufficient, which is the same "guard that cannot fire" shape one level up again; this file has now produced four of them.

Every ceiling is now the measured count. **Proven end-to-end rather than asserted:** the identical probe now takes the blocking gate red with

```
AssertionError: `triage` column guards rose to 12 (ceiling 11).
  packages/dashboard/app/components/command-center/MissionControlPanel.tsx:56  [c]  c === "triage"
```

Worth noting for whoever tightens next: the failure message points at an *existing* site rather than the new one, because the assertion reports the whole set. That is not wrong, but it means the reader has to diff. I left it, because the alternative is tracking per-site identity and the count is what gates.

One consequence for the program's bookkeeping: with ceilings at the floor, **any** PR that adds a lifecycle-column literal now fails the merge gate, including a legitimate one. The escape hatch is to convert instead, or — if the site genuinely is not a column — to classify the receiver in `NON_COLUMN_RECEIVERS` with the reason, which is the documented path and cannot be done silently.

`pnpm lint` clean, `pnpm test:gate` green (10 / 158 / 482 / 71). Rebased onto current origin/main.

gsxdsm:
Found while auditing the census classifier for false positives (#2650) — this one turned out to be a **real guard**, not a measurement artifact, and it is in a file this PR owns. Your diff does not currently touch it.

## `Board.tsx:937` and `:1021` — "Archive all done" is keyed to the literal column id

```tsx
{...(columnDef.flags.mergeBlocker || columnDef.flags.humanReview ? { onToggleAutoMerge: handleToggleAutoMerge } : {})}
{...(columnDef.id === "done" ? { onArchiveAllDone } : {})}
{...(columnDef.flags.archived ? { collapsed: archivedCollapsed, … } : {})}
```

It sits **between two lines that already use `columnDef.flags.*`**, so the converted form is right there in context.

**Consequence:** a workflow whose complete column is named anything other than `done` — `shipped`, or any renamed lineage — loses the "Archive all done" affordance entirely, and nothing reports it. An affordance that silently stops existing is the FN-6115 shape the surface-enumeration rule was written for.

**Both occurrences need it** — the file renders the column twice (937 and 1021), and the FN-6115 incident was specifically a fix that landed in one of two render paths.

Suggested form, matching its neighbours: `columnDef.flags.complete`.

I did not make the change: the file is yours, and this is a UI affordance change, so it wants the both-breakpoints check per the standing rule rather than a drive-by edit from another branch.

Worth two more notes since they bear on your census arithmetic:

- This site is correctly counted as a guard, so converting it **should** lower the backlog by 2.
- `#2650` corrects three defects in the instrument itself: `outcome` was counted as a column (the entire reason `--strict` was RED on main since #2633), four state/phase/result enums were counted, and `column: "<legacy>"` query filters were invisible to both the census and its ratchet. Net `784 -> 775`, **`triage` unchanged at 10**. If your PR body quotes a total, it may shift under you when that lands — the triage bar will not.

PR #2648, opened 2026-07-30
consolidate/e2e — E2E evidence: already-finalized terminal roles (real merge entry, no git) + ledger corrections

gsxdsm:
Consolidation branch for the E2E-evidence worker. Two commits, both engine test/comment only — **no production code, census unchanged**.

## Census (the authoritative instrument)

`node scripts/lifecycle-column-census.mjs` on this branch: **triage 10, total 784** — identical to its base. `lifecycle-column-census-ast.test.ts` and `lifecycle-column-census.test.ts` pass (15). This PR neither shrinks nor grows the backlog; it is evidence.

## What it contains, file by file

| file | change |
|---|---|
| `packages/engine/src/__tests__/workflow-already-finalized-live-e2e.pg.test.ts` | **new** — 3 cases, live PG store + real `runAiMerge` |
| `packages/engine/src/__tests__/workflow-lifecycle-live-e2e.pg.test.ts` | comment only — retires two unproven-ledger entries |

## The evidence: `isAlreadyFinalizedColumn` never needed the real-git lane

My unproven-sites ledger listed it as requiring a git harness because it is module-private inside `runAiMerge`. Reading the function instead of costing the lane: `runAiMerge` reaches it after only `store.getTask`, a pure workspace assert, and a pure branch resolve — **before** the merge blocker, settings, and any branch sync. `projectRootDir` is never touched on that path, and the short-circuit returns a `noOp` rather than throwing. Reachable through the real public entry point with no repository at all.

### Why two cases and not one

The guard resolves terminal columns **per role**:

```ts
terminal = [lifecycle.complete ?? "done", lifecycle.archived ?? "archived"]
```

#2471's P1 caught the first cut replacing the whole legacy **pair** as soon as *any* terminal role resolved — a workflow declaring `complete` but no `archived` collapsed to one element, silently lost the archived short-circuit, and an archived card then threw *"must be in 'in-review'"* for a card whose real state was "already done, nothing to do".

A per-set rule passes for whichever role **is** declared and fails the other, so a single case cannot tell the two rules apart. The shared fixture declares `complete` (renamed `shipped`) and **no** `archived`, so it is exactly that partially-declared shape — resolved half and fallback half live on one board.

Mutation-verified, each killing only its own case:

| mutation | kills |
|---|---|
| per-**set** replacement (the #2471 defect) | the legacy-`archived` fallback case |
| legacy pair only (conversion reverted) | the renamed-`shipped` case |

Plus a differential: a renamed **review** card must not report already-finalized. Without it both cases above would pass for a guard that finalizes everything — turning every merge into a silent no-op, the worst failure this function has.

Evidence strength is stated in the file header rather than overclaimed: this reads a returned **decision**, not a persisted row, so it proves the renamed board resolves and short-circuits — not that a card moves.

## Ledger corrections (comment only)

Two entries retired, both wrong the same way — each stated a **lane cost** as if it were an impossibility:

- `columnIsIntakeOrHold` — "consumers are dashboard-side" is true and irrelevant; its one consumer is an exported pure function. Proven on merged and renamed boards by work already merged in #2631.
- `register-task-workflow-routes.ts` — "standing up the route shell is mock-the-world" was false; `createApiRoutes` + `test-request.js` is this repo's established convention with ten existing suites for that file. Covered by its owner in #2614.

Counting this PR's own subject, that is **seven** wrong lane-cost inferences in that ledger. The rule it keeps violating is unchanged and now recorded in the file: read what the FUNCTION touches before costing a lane for it.

## Verification

`pnpm test:gate` exit 0 (695), `pnpm lint` exit 0, engine typecheck clean, 28 tests green across the AST ratchet and the three merged-board/planner-lane/already-finalized families.

## Not in scope here

The `performWorkflowRerunBounce` E2E. The harness exists (`new TaskExecutor(store, "/tmp/test", {})`), but `executor.ts:4305` gates the rebound on the legacy `in-progress`/`in-review` pair while resolving its target by role — so on a renamed board the bounce never fires and the resolved target is unreachable. An E2E asserting today's behaviour would cement that. It is executor.ts's owner's fix; evidence should follow it. Detail in #2632's thread.

### DIFF
--- a/packages/engine/src/__tests__/workflow-already-finalized-live-e2e.pg.test.ts
+++ b/packages/engine/src/__tests__/workflow-already-finalized-live-e2e.pg.test.ts
@@ -1,0 +1,134 @@
+/*
+FNXC:WorkflowLifecycleColumns 2026-07-29-17:30 (E2E evidence — the last tractable ledger entry):
+
+`isAlreadyFinalizedColumn` (merger-ai.ts) was the final converted site my unproven-sites ledger
+listed as needing the real-git lane, on the grounds that it is module-private and reached only from
+inside `runAiMerge`. Reading the function instead of costing the lane: `runAiMerge` reaches it after
+only `store.getTask`, `assertNotWorkspaceTaskMerge` (a pure check), and `resolveTaskWorkingBranch`
+(pure) — BEFORE the merge blocker, before settings, before any branch sync. `projectRootDir` is not
+touched on that path.
+
+So it is reachable through the REAL public merge entry point with no repository at all, and the
+short-circuit returns a `noOp` result rather than throwing. Sixth wrong lane-cost inference in that
+ledger; the rule it keeps violating is still "read what the function touches".
+
+WHAT THIS PROVES, and why the shape matters. The guard resolves the terminal columns PER ROLE:
+
+    terminal = [lifecycle.complete ?? "done", lifecycle.archived ?? "archived"]
+
+The first cut replaced the whole legacy PAIR whenever ANY terminal role resolved, so a workflow
+declaring `complete` but no `archived` collapsed to a one-element set and silently lost the archived
+short-circuit — an archived card then fell through to `getTaskMergeBlocker` and threw "must be in
+'in-review'" for a card whose real state was "already done, nothing to do". That is a caught-in-
+review defect (#2471 P1) whose regression test has to exercise BOTH halves independently, because a
+per-set rule passes for whichever role happens to be declared and fails for the other.
+
+The shared fixture declares a `complete` column and NO `archived` column, so it is exactly the
+partially-declared workflow that defect needed — the resolved half and the fallback half are both
+live in one board.
+
+WHAT IS REAL: a PostgreSQL TaskStore, a real persisted renamed workflow, and the real `runAiMerge`
+entry point. Assertions read the returned decision, which is weaker than a persisted row and is
+labelled as such: this proves the renamed board resolves and short-circuits correctly, not that a
+card is moved.
+*/
+import { beforeAll, beforeEach, afterEach, afterAll, expect, it } from "vitest";
+import "@fusion/core";
+
+import {
+  pgDescribe,
+  createSharedPgTaskStoreTestHarness,
+  type SharedPgTaskStoreHarness,
+} from "../../../core/src/__test-utils__/pg-test-harness.js";
+import { runAiMerge } from "../merger-ai.js";
+import { RENAMED_VOCAB, lifecycleIr } from "./_workflow-vocabulary-fixture.js";
+
+/** Never read on the path under test — the short-circuit precedes all git work. */
+const UNUSED_PROJECT_ROOT = "/nonexistent-by-design";
+
+pgDescribe("live already-finalized E2E: terminal roles resolved PER ROLE on a renamed board", () => {
+  const h: SharedPgTaskStoreHarness = createSharedPgTaskStoreTestHarness({
+    prefix: "fusion_already_final_e2e",
+  });
+
+  beforeAll(h.beforeAll);
+  beforeEach(h.beforeEach);
+  afterEach(h.afterEach);
+  afterAll(h.afterAll);
+
+  async function seedWorkflow(key: string): Promise<string> {
+    const created = await h.store().createWorkflowDefinition({
+      name: `Finalized ${key}`,
+      kind: "workflow",
+      /* Declares `complete` (renamed to `shipped`) and NO `archived` column — the
+         partially-declared shape the per-role fallback exists for. */
+      ir: lifecycleIr(RENAMED_VOCAB, `custom:final-${key}`, { mergeOrchestration: true }),
+    } as never);
+    return (created as { id: string }).id;
+  }
+
+  async function seedTask(taskId: string, column: string, workflowId: string): Promise<void> {
+    const store = h.store();
+    await store.createTaskWithReservedId(
+      { description: `finalized ${taskId}`, column } as never,
+      { taskId, applyDefaultWorkflowSteps: false } as never,
+    );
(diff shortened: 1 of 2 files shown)


gsxdsm:
## Finding: `recoverStuckMergeDeadlocks` is blind on a renamed board, and the census cannot see why

Not fixing it here — `self-healing.ts` belongs to the capacity worker. Reporting with the reproduction because the sweep looks converted and does nothing, which is the failure mode this program exists to remove.

### The site

`self-healing.ts:9291`:

```ts
const triage     = await this.store.listTasks({ column: "triage", slim: true });
const todo       = await this.store.listTasks({ column: "todo", slim: true });
const inProgress = await this.store.listTasks({ column: "in-progress", slim: true });
```

Its comment states the union is deliberate:

> This trio is a UNION and is deliberately left on literals. For the merged default lineage the `triage` read returns empty and `todo` supplies the cards; for a legacy/custom workflow that still declares `triage` it supplies them. **Either way the union is complete.**

That reasoning is sound for the two lineages it names — and there is a third. A workflow that renames **all** its columns (`backlog`/`building`/`checking`) matches none of the three literals, so the union is empty and the `filterByPreWipRole` call below it receives nothing. The role filter is correct and starved.

### Empirical proof, not a reading

Seeded one default-lineage board and one fully renamed board on a live PG store, then ran the exact three queries:

```
[PROBE] legacy-union=1  renamedHoldRows=1  renamedInsideUnion=0
```

The renamed card **exists** (`renamedHoldRows=1`) and is **invisible to the union** (`renamedInsideUnion=0`). So `recoverStuckMergeDeadlocks` cannot recover a stuck merge deadlock on any renamed workflow. Probe was scratch and is not in this branch.

### Why the instrument will not catch this

`lifecycle-column-census-ast.mjs` walks `BinaryExpression` nodes with comparison operators (`scripts/lib/…:113,153`). `{ column: "triage" }` is a **PropertyAssignment**, so it is not a guard by the census's definition and never enters the 784. The baseline ratchet therefore does not protect this class either — a new `listTasks({ column: "todo" })` can be added tomorrow without moving the number.

This is the same class as the one caught in review at `self-healing.ts:3015` (#2560, greptile P1), where the fix pattern is already written down two hundred lines above this site: *read the board and filter by role*. That one was found by a reviewer, not by a count.

### What I am and am not claiming

I am **not** proposing the 779 non-triage guards be chased — you have ruled that out and I agree. This is narrower: **one live sweep that is dead on renamed boards**, plus the fact that the authoritative instrument is structurally unable to see its cause. If the completion bar is "triage guards → 0 by the census", both can be true at once: the bar is met and this sweep still never fires.

Cheapest containment if you want it, in the capacity worker's lane and roughly one line each: extend the census AST to count `column:` property assignments carrying a legacy id, and pin them at their current value so the class stops growing. That measures the gap without committing anyone to closing it now.

gsxdsm:
## Pre-launch input for the 779-guard fleet: one work-order rule will misfire, and here is the measured size of the gap

Posting before the fleet is briefed because it changes a rule, not just a number. Also correcting my own earlier figure.

### Correcting myself first

Earlier in this program I said there were "137 query-filter sites", from a grep. Then a first AST pass said 17. **Both were wrong**: the grep over-matched, and the 17 came from a bad `git ls-files` glob that silently scanned 1078 of 1850 source files — `self-healing.ts` was not even in the set. A partial file list looks exactly like a small number.

Measured properly, with the census's own methodology (TS compiler API, PropertyAssignment named `column` with a legacy string literal, tests and `dist` excluded, 1850 files):

```
column:"<legacy>" property assignments: 127

  by id:   52 in-review · 26 in-progress · 17 todo · 14 done · 11 archived · 7 triage
  top:     48 self-healing.ts · 14 builtin-stepwise-coding-workflow-ir.ts
           13 builtin-coding-workflow-ir.ts · 7 project-engine.ts · 5 builtin-pr-workflow-ir.ts
```

**~32 of those are not guards at all** — they are inside `builtin-*-workflow-ir.ts`, where a workflow legitimately *declares* its own column ids. Converting those would be nonsense. That is precisely the "meaning is unclear, do not guess" category you are already planning to flag.

Of the remainder, `self-healing.ts` holds **49 `listTasks({ column: "<legacy>" })` query filters and zero move/update payloads** — every one a *source query* feeding a sweep.

### Why this breaks a fleet rule

> census before/after in the PR body; **the baseline ratchet must shrink by exactly the converted count**

The census walks comparison `BinaryExpression`s only (`scripts/lib/lifecycle-column-census-ast.mjs:113,153`). A `column:` PropertyAssignment is invisible to it. So a fleet worker in `self-healing.ts` — the largest batch at 111 — will hit both classes in the same functions, and:

- convert only the comparisons → arithmetic is clean, and sweeps whose **source query** still filters on a dead literal stay blind;
- convert the query too → the count does **not** move by the converted amount, and the PR looks like a miscount when it is actually more correct.

The second outcome is the one that punishes the better worker. The file itself already documents the coupling at `self-healing.ts:2849` — *"(`{ column: "todo" }`) and the per-task GUARD (`task.column !== "todo"`)"* — and #2560 had to fix exactly this pairing after a converted predicate was left with a literal query.

I have already shown one live consequence on this PR: `recoverStuckMergeDeadlocks` cannot see a renamed board at all, proven on a live store (`renamedInsideUnion=0`), and the census cannot see why.

### Cheapest fix, before the fleet launches rather than after

Extend the census to count `column:` property assignments as a **separate, separately-pinned** category — not folded into the 784, so the completion bar is unchanged. Then:

- the fleet's arithmetic rule becomes checkable for both classes;
- converting a query becomes visible work instead of an apparent miscount;
- the class stops growing silently, which today it can.

Roughly a dozen lines in the AST module plus a baseline key. I have not written it — `scripts/` is #2633's instrument and the capacity worker owns `self-healing.ts`. If you want it in my lane instead, say so and it is a small PR.

PR #2653, opened 2026-07-30
Audit: every sync workflow-IR read answers for the DEFAULT workflow (not a PG-only problem)

gsxdsm:
Docs only. This came out of a #2593 review thread that reported the problem as PostgreSQL-specific. **It is unconditional**, and it has consequences well outside the guard I was fixing — including one that looks like a live production break for custom workflows.

## The chain, each link checkable

1. `TaskStore.getTaskWorkflowSelection(taskId)` delegates straight to `getTaskWorkflowSelectionImpl` — **no mode branch** (`store.ts:2545`).
2. `getTaskWorkflowSelectionImpl` **returns `undefined` unconditionally** (`workflow-definitions.ts:505-512`). Its own comment: *"sync selection reader is incomplete-PG; use getTaskWorkflowSelectionAsync."* A PG-cutover stub that never got finished.
3. So `resolveTaskWorkflowIrSyncImpl` always takes its `if (!workflowId)` branch and returns `resolveDefaultWorkflowIr()`. Its `isBuiltinWorkflowId` and `SELECT ir FROM workflows` branches are **unreachable in production**.

`resolveTaskWorkflowIrSync` is typed `WorkflowIr`, non-optional — so callers cannot detect the substitution. There is no `undefined` to check and the IR that arrives looks valid.

**Why tests don't catch it:** test stores stub `getTaskWorkflowSelection` with a real selection, so the reader works under test and substitutes only in production. Any test written against a stubbed store proves the caller's logic and never the reader's behavior.

## Consequences, severity descending

1. **Custom fields appear to be rejected on custom workflows.** `resolveTaskCustomFieldDefsSyncImpl` returns `ir.fields` — the DEFAULT workflow's. `task-update.ts:128-136` validates against them, and its own comment states the outcome: *"a write against a workflow with no fields (the default) is rejected with a typed CustomFieldRejectionError."*
2. **Per-workflow capacity pools collapse** — `resolveEffectiveWorkflowIdSyncImpl` reads the same selection, so every task resolves to `resolveCapacityPoolId(undefined)`.
3. **Plugin transition hooks re-run against the wrong IR** (`lifecycle-ops.ts:1052`, crash recovery).
4. **Terminal-node detection degrades** to `nodeId === "end"` (`branch-and-pr-entities.ts:578`).
5. **A U7 guard was inert** — fixed in #2593. Its fail-closed arm was `workflowIr ? … : true`, dead code against a non-optional return.

**#1 and #2 are REASONED FROM SOURCE, NOT OBSERVED.** I did not execute those paths, and I am labelling them that way in the doc rather than reporting them as confirmed. No test in `packages/core` covers `CustomFieldRejectionError` or `resolveTaskCustomFieldDefsSync` — consistent with the gap, but absence of a test is not proof of a break. **Reproduce before fixing.** I would rather hand you a labelled hypothesis than a confident claim I did not verify.

## Why this matters for the fleet, specifically

The census work replaces column literals with trait lookups. A conversion that resolves its traits through a **sync** reader produces a guard that reads the DEFAULT workflow's traits for every task — plausible, wrong, and invisible. **It converts a visible literal into a hidden bug**, and the ratchet counts it as progress.

Suggested addition to the fleet brief: conversions must resolve through `resolveWorkflowIrForTaskWithProvenance` and branch on `source`; `resolveTaskWorkflowIrSync` is never acceptable in a converted guard.

## Not fixed here

Each consequence needs its sync call path made async — a real slice per site, not an end-of-turn edit. #2593 fixed only the one that was mine. Census unchanged (781 / triage 5); this PR adds and converts no guards.

## Summary by CodeRabbit

* **Documentation**
  * Added an architecture-pattern finding documenting a workflow-reading limitation that can cause synchronous reads to use the default workflow.
  * Described resulting effects on custom workflow updates, crash recovery, capacity-pool handling, and terminal-node detection.
  * Documented testing gaps and guidance to avoid synchronous task workflow reads.

### DIFF
--- a/docs/solutions/architecture-patterns/sync-workflow-ir-readers-always-return-the-default.md
+++ b/docs/solutions/architecture-patterns/sync-workflow-ir-readers-always-return-the-default.md
@@ -41,13 +41,27 @@ against them, and its own comment records the outcome: "a write against a workfl
 on a task whose custom workflow defines fields is rejected as unknown-id. Same reader is used by
 `workflow-ops.ts` for the old-defs diff.
 
+Three production callers, not one — verified by grep while checking the review correction below:
+`task-update.ts` (update), `workflow-ops.ts` (the old-defs diff on workflow change), and
+`workflow-task-create-ops.ts` (task CREATE). So the exposure covers creating a task with custom
+fields as well as updating one.
+
 REASONED FROM SOURCE, NOT OBSERVED: I did not execute this path. No test in `packages/core` covers
 `CustomFieldRejectionError` or `resolveTaskCustomFieldDefsSync`, which is consistent with the gap but
 is not proof. Reproduce before fixing.
 
-**2. Per-workflow capacity pools collapse.** `resolveEffectiveWorkflowIdSyncImpl` reads the same
-selection, so it always calls `resolveCapacityPoolId(undefined)` — every task resolves to the default
-pool regardless of its workflow.
+METHOD NOTE, earned the hard way: a consequence is only real if something CALLS the defective
+function. I listed the capacity item without checking, and review caught it. Every other item here
+has had its call sites verified — `isTaskTerminalNodeIdImpl` via the `isTerminalNodeId` callback in
+`branch-and-pr-entities.ts`, the hook re-run in `lifecycle-ops.ts`, and the three above.
+
+**2. The synchronous capacity-pool helper returns the default pool — but nothing calls it.**
+`resolveEffectiveWorkflowIdSyncImpl` reads the same selection, so it always calls
+`resolveCapacityPoolId(undefined)`. CORRECTED after review: I first wrote this as "per-workflow
+capacity pools collapse", which was wrong and would have sent someone to fix an unused path. The
+binding capacity path reads the selection ASYNCHRONOUSLY inside its transaction, and this sync helper
+has NO callers — `grep` finds only its own definition, its `Impl`, and one comment. Latent, not live:
+it becomes real the moment someone calls it, which is the only reason it is still listed.
 
 **3. Plugin transition hooks re-run against the wrong IR.** `lifecycle-ops.ts` crash-recovery passes
 this IR to `runPluginColumnTransitionHooks`, so a custom-workflow card's hooks are evaluated against
--- a/docs/solutions/architecture-patterns/sync-workflow-ir-readers-always-return-the-default.md
+++ b/docs/solutions/architecture-patterns/sync-workflow-ir-readers-always-return-the-default.md
@@ -120,6 +120,31 @@ This matters for the census conversions specifically: replacing `column === "tri
 lookup that resolves through a SYNC reader produces a guard that reads the default workflow's traits
 for every task — plausible, wrong, and invisible. It converts a visible literal into a hidden bug.
 
+## Why #6 is not a mechanical await-insertion (checked, not assumed)
+
+I went to convert one \`resolvePlannerLanes\` call site as a worked example and stopped, because every
+site is SYNC BY CONSTRUCTION. That is why the sync reader is there in the first place:
+
+  - \`triage.ts\` \`taskColumnWakeHandler\` — a synchronous event-handler callback.
+  - \`triage.ts\` dispose handler — same shape.
+  - \`triage.ts\` stale-planning sweep — inside an \`allTasks.filter((t) => …)\` predicate.
+  - \`executor.ts\` — three further call sites.
+
+You cannot \`await\` inside a \`filter\` predicate or an event handler that callers treat as sync, so
+"make it async" is not an edit, it is a restructuring per site.
+
+THE PATTERN THAT WORKS IS ALREADY IN THIS FILE. \`discoverReadyPlanningTasks\` had the same problem and
+solved it by PRE-RESOLVING: a store-free \`couldBeCandidate\` prefilter narrows the set, then a bounded
+(8) concurrent \`resolveTaskLifecycleColumns\` pass resolves the survivors before the synchronous
+decision runs. Each \`resolvePlannerLanes\` site needs that shape — resolve lanes for the candidate set
+up front, then keep the predicate synchronous over the resolved map.
+
+For the event handlers the shape is different again: they fire per task, so they want a small cached
+resolution keyed by task id rather than a batch pass, invalidated on workflow-selection change.
+
+Estimating from that: this is one slice per site with its own test, not one PR. Recording it so the
+next person does not start by trying to add \`await\` and conclude the codebase is fighting them.
+
 ## Not fixed here
 
 Each consequence needs its sync call path made async, which is a real slice per site. This document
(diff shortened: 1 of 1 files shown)


gsxdsm:
Correct, and I've taken it — `91aef6fd0`.

I verified before accepting: `grep -rn resolveEffectiveWorkflowIdSync` across `packages` (excluding `dist`) finds **only** its own definition in `store.ts`, its `Impl` in `task-store-helpers.ts`, and one comment. **No callers.** And the binding capacity path does read the selection asynchronously inside its transaction, as you say. Calling that "per-workflow capacity pools collapse" would have sent someone to fix an unused path — the misdirection you named is the actual harm, more than the wrong severity.

Restated as **latent**, with why it stays in the list at all: it becomes real the moment someone calls it, and the next person reaching for a sync workflow read is exactly who this doc is for.

## Checking your correction surfaced a miss in the other direction

While verifying call sites I found `resolveTaskCustomFieldDefsSync` has **three** production callers, not the one I documented:

- `task-update.ts` — update
- `workflow-ops.ts` — the old-defs diff on workflow change
- `workflow-task-create-ops.ts` — **task CREATE**

So consequence #1 covers creating a task with custom fields, not just updating one. Broader than I wrote it.

## The method error, recorded in the doc

**A consequence is only real if something CALLS the defective function.** I traced the defect correctly and then listed impacts without checking call sites — which is how a latent helper ended up ranked second by severity. Every remaining item now has its callers verified: `isTaskTerminalNodeIdImpl` via the `isTerminalNodeId` callback in `branch-and-pr-entities.ts:546`, the hook re-run at `lifecycle-ops.ts:1052`, and the three above.

I added that as a method note in the doc rather than just fixing the line, because the same shortcut would produce the same error on the next audit — and this doc is going to be read by the census fleet, whose whole job is deciding whether a defective-looking guard actually matters.

Docs only; no verification beyond the existing green.

gsxdsm:
## The biggest consequence, found last — and it explains why every suite stays green

Pushed `4ff57ea91`. I went looking for the regression test I owed on #2593 and found something that matters more than the test.

**`resolvePlannerLanes` (`replan-target.ts:92-110`) resolves `intake`/`hold`/`wip`/`review` by calling `resolveTaskWorkflowIrSync`** — the reader proven above to always return the default IR. **Six call sites** in `executor.ts` and `triage.ts` depend on it: planning wake, planning dispose, replan targeting, and the approved-plan recovery gate.

So every planner lane in the engine is derived from the DEFAULT workflow, for every task. A workflow whose intake is renamed (say `inbox`) gets `lanes.intake === "todo"`, so a card sitting in `inbox` is not recognised as being in its own planner column and recovery never fires. **That is the same class of stall this program has spent Phase D fixing, one layer down** — and it is invisible, because the returned lanes are a valid-looking set from the wrong workflow.

## Why the tests do not catch it — concretely, not in principle

I traced this through the existing positive control, `triage.test.ts` › *"recoverApprovedTask auto-approves a legacy pre-hygiene fingerprint too"*, which passes a `column: "triage"` task and asserts `true`:

- **On the mock:** `createMockStore` does not stub `resolveTaskWorkflowIrSync`, so `resolvePlannerLanes` returns `LEGACY_PLANNER_LANES` (`intake: "triage"`) and the card matches the **intake arm** directly.
- **In production:** the reader exists and returns the merged default (`intake: "todo"`), that arm does **not** match, and a completely different branch decides the outcome.

The mock and production take different paths through the same function. The test is green and proves nothing about the shipped behavior.

## What this means for the test I owed on #2593

I set out to write it and stopped, because I now know the fixture I was planning would have been the third useless one: on a bare mock the orphan arm I fixed is **never reached** — the intake arm short-circuits it. A meaningful test has to stub `resolveTaskWorkflowIrSync` to produce `lanes.intake === "todo"` **and** control provenance independently, so the orphan arm is the only thing deciding. That is constructible, and it is a deliberate piece of work rather than something to append at the end of a long turn.

Recording it here so the next person starts from the real constraint instead of rediscovering it.

## Scope

Docs only. Census unchanged (776 / triage 5). None of the six consequences are fixed here — each needs its sync call path made async, per site. **#6 should probably be triaged before the fleet launches**, since the fleet's conversions will read lanes through exactly this function.

gsxdsm:
Both threads were resolved before I got here, but the scoping one was a real imprecision and it was mine — I reported this finding as "`resolveTaskWorkflowIrSync` ALWAYS returns the default IR" without separating the interface from the implementation. Pushed a correction rather than leave the doc overstating it.

`WorkflowIrResolverStore.getTaskWorkflowSelection` is declared `{ workflowId; stepIds } | undefined`, so a conforming implementation may return a real selection — the test fixtures do exactly that, which is the same reason the substitution is invisible under test. The unconditional `undefined` belongs to one implementation: `getTaskWorkflowSelectionImpl`, a stub left by the PG cutover that the production store delegates to with no mode branch.

The distinction has a practical consequence, which is why it earns the words: a reader who takes "always" as a property of the *interface* concludes the sync path is unusable and rewrites callers that are fine; a reader who takes it as a property of the *impl* looks at the stub, which is where the defect actually is.

The capacity claim from the other thread was already corrected in the body before this PR opened — item 2 records that the binding capacity path reads the selection asynchronously inside its transaction and that `resolveEffectiveWorkflowIdSyncImpl` has no production callers. No further change there.

gsxdsm:
Two things, and the first is someone else's correction that I'm keeping.

## The scoping correction is right, and I rebased onto it rather than over it

`c125413` narrowed my claim from "every sync read is unconditional" to "every sync read **through the production `TaskStore`**". That's correct and the distinction has teeth: the unconditional `undefined` belongs to one implementation (`getTaskWorkflowSelectionImpl`), not to the `WorkflowIrResolverStore` interface — a conforming implementation may return a real selection, and the test fixtures do exactly that, which is *also* why the substitution is invisible under test.

I'd written it as a property of the interface. A reader taking it that way would conclude the sync path is unusable and rewrite callers that are fine. My local branch was behind when I went to push; I rebased so their commit survives.

## Why consequence #6 is not a mechanical await-insertion — checked, not assumed

I went to convert one `resolvePlannerLanes` call site as a worked example and stopped. **Every site is sync by construction**, which is *why* the sync reader is there:

| site | shape |
|---|---|
| `triage.ts` `taskColumnWakeHandler` | synchronous event-handler callback |
| `triage.ts` dispose handler | same |
| `triage.ts` stale-planning sweep | inside an `allTasks.filter((t) => …)` predicate |
| `executor.ts` ×3 | further call sites |

You cannot `await` inside a `filter` predicate or a handler callers treat as sync. So "make it async" is a restructuring per site, not an edit — my earlier deferral was right for this reason rather than for context budget, and I'd rather say which.

**The pattern that works is already in the same file.** `discoverReadyPlanningTasks` had this exact problem and solved it by pre-resolving: a store-free `couldBeCandidate` prefilter narrows the set, then a bounded (8) concurrent resolve pass runs over the survivors, leaving the decision synchronous over a resolved map. Each sweep-shaped site wants that. The event handlers want something different — a small per-task cached resolution invalidated on workflow-selection change.

That's one slice per site with its own test, not one PR. Recorded so the next person doesn't start by adding `await` and conclude the codebase is fighting them.

Docs only; census unchanged (776 / triage 5).

PR #2654, opened 2026-07-30
ci: enforce the lifecycle-column ratchet — it has never actually run

gsxdsm:
**The ratchet was advisory.** `scripts/lifecycle-column-census.mjs` existed only as `pnpm census:lifecycle-columns` — without `--strict` — and **no workflow invoked it**. Nothing has ever compared the tree to the baseline. Every "the baseline ratchet holds them" assumption in this program rested on a check that does not run.

That explains both classes of hole:

**1. Three PRs lowered counts without re-recording,** leaving allowances the deleted guards could return through while every check stayed green. I've tightened them across #2593 and earlier PRs, but nothing stops the next one.

**2. #2621 GREW the count while its own title claimed "count 0 → 0".** It added `column === "triage"` and `column === "todo"` at `register-task-workflow-routes.ts:2681`, taking that file to **23 against an allowance of 22**. It landed unchallenged. This is the failure mode the ratchet exists to prevent, and it happened *inside this program*, in a PR that asserted the opposite.

## The change

Adds `check:lifecycle-columns` (the census with `--strict`) to the `pr-checks.yml` lint job, next to `check:changesets` and `check:routes-modular` — the established pattern. **~1.8s over ~1950 files**, so this is not a slow-test addition.

## Proven to fail, in both directions

A guard that reports success without checking anything is worse than no guard, so:

| injected defect | result |
|---|---|
| `const __probe = (c: string) => c === "triage"` added to `moves.ts` | `count ROSE — moves.ts: 39 -> 40`, exit 1 |
| run against main's current baseline | exit 1 on `mission-feature-sync.ts: allows 5, tree has 0` |

Both reverted; exit 0 restored. Note the second row: **this check is RED on main right now**, which is the point.

## Merge order

**Stacked on #2593**, which carries the `DELIBERATE-LITERAL` marker for the #2621 site (a v1 IR declares no roles, so no trait can answer that question) plus the baseline re-record. Standalone on main this PR is red — correctly. **Merge #2593 first**, then this.

I stacked rather than duplicating those two edits because I already caused one conflict today by appending related content from two branches, and #2651 merged a correction ahead of the section it corrected. Same-content edits in two PRs is the same mistake.

## Census

Unchanged by this PR: **776 total, triage 5, reviewed 16** — it adds no guards and converts none. It only makes the numbers enforceable.

## For the fleet

This should land before the 776-guard fleet launches. The brief says "the baseline ratchet must shrink by exactly the converted count" — until now nothing verified that claim, so a batch worker could report a shrink that did not happen, or grow the count while converting, and CI would agree.


gsxdsm:
## Three-way baseline collision, measured on the merged tree — and my earlier reconciliation advice was WRONG

#2650, #2652 and #2654 all edit `scripts/lib/lifecycle-column-census-baseline.json`, and **#2654 makes `--strict` a blocking CI check** (`pnpm check:lifecycle-columns` in `pr-checks.yml`). That turns a drifted baseline from a cosmetic nuisance into a hard merge blocker for whoever lands second and third.

I merged all three locally and ran the new gate.

### Correcting myself first

Earlier I told #2650 and #2652 that the second-to-land just needs:

```
```

**That does not work here, and I should have tested it before saying it.** The regression check `process.exit(1)`s *before* the write, so when #2650's query category meets a baseline that has no `queryByFile` keys, every file reads as a rise from zero and the update never happens:

```
--strict                    -> exit 1   self-healing.ts (query filter): 0 -> 48
--strict --update-baseline  -> exit 1   baseline file UNCHANGED (verified: empty git diff)
```

It is the same bootstrap gap I hit when adding the category, and I seeded those keys by hand for exactly this reason — then forgot that the person after me would hit it too. **The reconciliation is a one-time manual seed of the three new keys** (`properties`, `queryByColumnId`, `queryByFile`), after which `--strict` is self-maintaining. If #2650 lands last this is already done; if it lands before #2652/#2654, the later PR must re-seed.

### The combined numbers, all three applied

```
COLUMN guards (backlog):            767
DELIBERATE-LITERAL (reviewed):       18
QUERY filters:                       83
IR node definitions:                 43

by column id:  in-review 208 · done 200 · archived 148 · in-progress 146 · todo 60 · triage 5
```

### The bar is at 5, and here is every remaining site

```
packages/core/src/task-store/moves.ts:890,967,968,973   toColumn === "triage"   (4)
packages/dashboard/app/components/TaskContextMenu.tsx:179   column === "triage"  (1)
```

Four of the five are the `moves.ts` flag-OFF branch, which dies wholesale when U12 resolves the flag — as the sequencing already says. The fifth is a single dashboard site. With these three PRs in, **the completion bar is one flag resolution plus one line away from zero.**

### Suggested merge order, since the gate makes it matter

**#2654 first** (it installs the gate and its own baseline is self-consistent), then **#2652**, then **#2650** last — because #2650 is the only one introducing a *new* baseline category, and landing it last means exactly one manual seed instead of one per subsequent PR. I am happy to be sequenced differently; I only ask that whoever is last confirms `pnpm check:lifecycle-columns` exits 0 **on the merge result**, not on their branch alone. That distinction is the whole finding: all three are green in isolation.

PR #2655, opened 2026-07-30
U12: resolve the move-path compatibility flag — trait hooks unconditional, legacy branch deleted

gsxdsm:
U12's headline goal. The raw `experimentalFeatures.workflowColumns` flag gated **every task move**; its six seams are now unconditional and the flag, its last two readers, and the 124-line inline legacy branch are deleted.

## Deleted, not converted

The flag-OFF branch goes with the gate. Converting a branch we intended to delete would have left a second definition of every column side effect alive to drift — the defect this program has spent its length removing.

Both readers flip in **one commit** because they are not separable: the preflight in `workflow-task-create-ops.ts` computes the `movePolicyPreflight` that `moves.ts` consumes and validates. Un-gating either alone either evaluates workflow move policies — with their plugin-gate side effects — whose result is ignored, or validates against a preflight that was never computed.

## Evidence, not assertion

**Equivalence (precondition 1).** `moves-flag-equivalence.test.ts` (commit 1) ran the same journey under both flag states against live PostgreSQL and diffed the persisted row: **identical across 128 fields** plus an equal timing shape, over `todo → in-progress → in-review → todo → in-progress`. Mutation-verified both ways — stamping the flag-ON branch, and diverging the reopen hook, each fail it.

**And there was stronger evidence already on main that isn't mine.** U2b's `move-path-equivalence.pg.test.ts` ran *every* scenario once per path and has been green across ~10 of them: `preserveStatus`, `preservePause`, timing accounting, `preserveProgress`, `preserveWorktree`, engine-source rehome, `in-progress → todo`. Two independently built harnesses agreeing is the best evidence this question has had.

**The flag was read by nothing in production.** `experimentalFeatures` is global-only and no module writes it, so this path had never run for any project without a stale persisted value. That is also why a green suite was never evidence on its own — both paths were individually valid and only one was live.

## Two claims of mine this PR corrects

**1. Seam 2 does not introduce new rejections.** I said in #2639 and in the census that with the flag off there is *no* target validation, so flipping would add refusals. Reproduced the opposite: a move to an undeclared column already rejects on the legacy path with `Invalid transition: … Valid targets: …`. I found it because the discriminator I wrote to prove "the flag is the cause" failed.

**2. My first equivalence test proved nothing.** It used `updateSettings`; `experimentalFeatures` is **global-only**, so `getSettingsFast()` filtered the write out and `useWorkflow` was false in *both* runs. Caught by stamping the flag-ON branch and watching the test stay green. It now writes via `updateGlobalSettings` and **asserts the flag took effect** before the journey. U2b's harness carries the same warning independently — `MUST be updateGlobalSettings, NOT updateSettings`.

## The user-visible change

Move rejections now report **workflow-resolved** targets instead of the hardcoded legacy adjacency table. Concretely: `Valid targets: in-progress, triage, archived` becomes `Valid targets: archived, in-progress`. That is the fix, not a regression — the legacy table still advertised `triage`, a column the default lineage stopped declaring at #2515, so an operator following the old message was told to move somewhere impossible.

Likewise a move *into* `triage` is now refused rather than stranding the card in a column with no trait flags, invisible to every trait-driven sweep until reconciliation re-homes it. `live-move-path-undeclared-target.test.ts` characterised exactly that defect and carried `it.todo("should REFUSE a move into a column the task's workflow does not declare (U2b)")` — **this fulfils it.**

## Test migration

| file | change |
|---|---|
| `move-path-equivalence.pg.test.ts` | deleted — every scenario ran once per path; purpose fully discharged |
| `workflow-capacity-invariant.pg.test.ts` | `setPath("inline"\|"hooks")` → `assertMovePathLive()`; the probe is **kept** so capacity cannot pass because moves were broken for an unrelated reason |
| `store-movement.pg.test.ts` | asserts the refusal **and** that the legitimate backward move still works, so it reads as a narrowing |
| `raw-workflow-columns-flag-census.test.ts` | deleted per its own instructions — it was built to fail in both directions and fired exactly as designed: `expected [] to deeply equal [3 readers]` |
| `moves-workflow-flag-seams.test.ts` | deleted — it pinned the six seams this removes |

## Verification

Full core suite: **33 failed / 10 files — byte-identical to main's baseline**, with **zero** files failing exclusively on this branch. I measured the baseline by checking out `origin/main` and running the same command, because the first comparison I made was by count alone and would have blamed the flip for 8 files that were already red.

`pnpm lint` clean. `pnpm test:gate` green (10 / 132 / 482 / 71). `tsc -p packages/core/tsconfig.json` clean. Core builds.

## Left in place deliberately

The `workflowColumns` settings key stays schema-tolerated and is already in `HIDDEN_EXPERIMENTAL_FEATURE_KEYS`, so an upgraded project carrying a stale value renders nothing and loads cleanly. Removing it from the schema would risk rejecting those projects for no benefit now that nothing reads it.

---

## Rebased onto current main — and `triage` reaches ZERO

| metric | before | after |
|---|---:|---:|
| `moves.ts` column guards | 39 | **15** |
| repo column total | 745 | **741** |
| **`triage` column guards** | 1 | **ABSENT (0)** |

`triage` is now absent from `byColumnId` entirely: no unconverted `triage` guard remains anywhere in production source. Combined with #2664 (the last one, in `TaskContextMenu`) this closes bar item 1.

The census behaved exactly as designed on the rebase: the flip *deletes* guards, so `--strict` reported `moves.ts: allows 39, tree has 15` rather than leaving a stale allowance, and the re-record lands in this PR's diff.

## Verification on the rebased tree

- Full core suite: **33 failed / 10 files — identical to main's baseline**, zero files failing exclusively on this branch (measured by checking out `origin/main` and diffing the failing-file sets, not by comparing counts).
- `pnpm test:gate` green (10 / 158 / 487 / 71).
- `pnpm check:lifecycle-columns` exits 0.
- `pnpm lint` clean, `@fusion/core` builds.

## Two review fixes carried in this PR

**P1 — optionless engine moves lost their bypass.** `resolveWorkflowBypassGuardsImpl` did `void moveSource;` — it discarded the resolved parameter and re-read `options?.moveSource`, so `moveTask(id, target)` resolved to `"engine"` at the call site and computed `bypassGuards === false`. Latent while the flag gated validation; with the gate gone, an internal executor/merger/recovery move made without an options object would be judged as a user move.

**P2 — the absence signal.** Emitting `workflowId` unconditionally would have stamped `builtin:coding` onto every task with no explicit selection, reporting a fallback as authoritative. Now emits the selection directly, so absent still means "not resolved here".

## Summary by CodeRabbit

- **Bug Fixes**
  - Task moves are now validated against the task’s declared workflow.
  - Invalid destinations are rejected with a clear error, and tasks remain in their original column.
  - Valid backward moves continue to work as expected.
  - Move behavior and lifecycle updates are now handled consistently across workflows.


gsxdsm:
Both real, both fixed, and the P1 was exact.

**P1 — optionless engine moves lost their bypass.** `resolveWorkflowBypassGuardsImpl` did `void moveSource;` — it **discarded the resolved parameter** and re-read `options?.moveSource`. So `moveTask(id, target)` with no options resolved `moveSource` to `"engine"` at every call site and then computed `bypassGuards === false`, because `options` was `undefined`. The call site and the bypass resolver disagreed about what kind of move it was.

That was harmless while the flag gated validation — nothing consumed the answer. This PR removes the gate, so seam 2's guards now run for those calls and an internal executor/merger/recovery move made without an options object would be judged as if a user had made it. Exactly as you described.

Worth noting how it got there: the `void` is a deliberate unused-parameter suppression. Someone noticed the argument was unused and silenced the lint rather than wiring it up, which left the parameter and the behaviour permanently out of sync with nothing to detect it.

**P2 — the absence signal.** Also right, and it would have re-created the defect the original note was written to prevent, from the other direction. `effectiveWorkflowIdForMove` is `selection?.workflowId ?? DEFAULT_WORKFLOW_ID`, so emitting it unconditionally stamps `builtin:coding` onto every task that has **no** explicit selection — a fallback reported as an authoritative choice, in a wire field no subscriber reads yet.

Fixed by emitting the **selection** directly rather than the resolved-with-fallback value, so `builtin:coding` still appears for tasks that genuinely select it and is absent for tasks that merely default to it. `effectiveWorkflowIdForMove` is deleted along with it — its only job was applying that fallback, so keeping it would have meant either an unused binding or the stamp you flagged.

I took the emit-the-selection route over your `suggestion` block (which dropped the field entirely) because absent-vs-present is the documented contract and a real selection is genuine information; dropping it would have lost that too.

Verification: `pnpm lint` clean, `pnpm test:gate` green (10 / 132 / 482 / 71), `tsc` clean, and the four move-path suites 21/21. Full core suite remains at main's baseline — 33 failed / 10 files, zero files failing exclusively on this branch.

gsxdsm:
## Pre-verified this PR against current main — it closes the bar to ZERO, and it needs one extra step to merge green

I ran the closing bar's item-4 battery on `pr/2655 + origin/main` so nothing surfaces after you merge. Reporting because two things need your action and one is a hard CI blocker.

### It closes the bar

```
triage guards: 0        <- enumerated, not inferred: zero sites remain repo-wide
backlog:       724
```

Worth noting: **on your branch alone the count is 1, not 0.** Your branch is 12 commits behind main, and the last non-`moves.ts` site (`TaskContextMenu.tsx:179`) was converted on main in the meantime. It is the *merge* that reaches zero, which is why measuring the branch in isolation understates it.

### It will fail CI unless you re-record the baseline in the same commit

`pnpm check:lifecycle-columns` is a **required check** since #2654. On the merge result it fails:

```
check:lifecycle-columns   exit 1   "A stale allowance is a hole"
```

Not a defect in this PR — it is the ratchet working. Removing the flag-OFF branch deletes guards, the baseline still allows them, and a stale allowance is a failure by design (#2633's both-directions rule). Fix is one command in the same commit:

```
```

After it: `exit 0`. Verified.

### Full battery on the merged tree — all green

| component | result |
|---|---|
| `check:lifecycle-columns` | exit 0 *(after the re-record above)* |
| `pnpm test:gate` | exit 0 — 726 tests (158 + 10 + 487 + 71) |
| all 18 pg E2E suites | exit 0 — 119 tests, 0 failed |
| `pnpm verify:fast` | exit 0 — 15 steps, 50.4s |

Resulting census on the closing tree:

```
COLUMN guards 724 · ROLE 5 · STATUS 186 · DELIBERATE 17 · QUERY filters 83 · IR definitions 43
in-review 200 · done 196 · archived 148 · in-progress 138 · todo 42 · triage —
```

`triage` no longer appears in the by-column breakdown at all.

### Note on the merge conflict

`scripts/lib/lifecycle-column-census-baseline.json` conflicts with main. I resolved it by taking main's copy and re-recording, which is the correct resolution — hand-merging that file line-by-line produces a pin that matches neither tree.

gsxdsm:
## `main` is RED on the required census check, and it stays red until #2655 merges

`pnpm check:lifecycle-columns` is a required PR check. On clean `origin/main` (a6138abeff):

```
exit 1   packages/core/src/task-store/moves.ts: 15 -> 39
```

Every open PR fails CI on this, and none of them caused it.

### Diagnosis: main's baseline is pinned to a tree that includes #2655

Measured, each step verified:

| tree | `moves.ts` column guards |
|---|---|
| clean `origin/main` | **39** |
| #2655 branch alone | **15** |
| #2655 merged with main | **15** |
| **baseline on main says** | **15** |

`moves.ts` carries **zero** `DELIBERATE-LITERAL` markers, so no exemption explains the 24-guard gap. The only tree that satisfies the pinned 15 is one containing #2655's flag resolution — which is still open.

So the baseline was re-recorded against a working tree that already had the flag resolution applied. The pin is correct about the *future* and wrong about `main`.

### It self-resolves on merge — verified, not assumed

```
check:lifecycle-columns on (#2655 + main):  exit 0
backlog: 724      triage: 0
```

### What this means for sequencing

- **#2655 is now load-bearing for CI, not just for the bar.** Until it lands, every PR in the program is red on a required check through no fault of its own.
- Nothing else needs fixing. Re-recording main's baseline to 39 would turn the gate green *and* concede 24 guards the flag resolution is about to delete — the wrong direction, and it would have to be undone immediately.
- If #2655 cannot land promptly, the honest interim is to make the check non-blocking rather than to loosen the pin.

### Method note

I got this wrong twice before getting it right, and both failures looked like answers. A `git fetch` of the PR ref failed with its stderr suppressed, and a later `git checkout` aborted on local changes — in both cases I measured `main` while believing I was measuring the PR branch, and the numbers were self-consistent enough to be convincing. The table above was rebuilt with the branch identity, the merge-commit count, and the file content asserted at each step before any number was read.

gsxdsm:
You are right, and so was the previous round — they are the same defect seen from two sides, and I should not have taken the first one at face value inside this PR.

**Round 1** said an optionless `moveTask(id, target)` *loses* bypass, because the call site resolves `moveSource` to `"engine"` while this function read the absent option. I switched to the resolved value. **Round 2** says that grants privileged bypass to public callers. Both hold, because an absent `moveSource` is genuinely ambiguous. Measured on this tree:

- **18** optionless calls in `packages/engine` (`self-healing.ts`, `project-engine.ts`) — genuine engine moves
- **8** optionless calls in `packages/dashboard` HTTP routes — `reset`, `rebound`, `respecify`, `unassign` — operator-initiated, and they must **not** skip merge blockers or plugin gates

Reading the resolved value hands bypass to those eight routes. Reading the option leaves the eighteen engine calls unbypassed — which is what has shipped all along.

**Reverted to the shipped read.** A flag-resolution PR is the wrong place to change who may skip merge blockers: it is a behaviour change with a security shape, it is not required by the flip, and a green suite is not evidence for it. My round-1 fix was the wrong call for this PR even though the finding it described was accurate.

Worth being precise about what the flip does and does not do here: `bypassGuards` was always computed outside the flag, so this PR does not change how it is derived. What the flip changes is that seam 2's validation now *consumes* it. Those 18 engine calls move to `"done"`/`"in-progress"`, which the default lineage declares, so they pass validation; on a custom lineage they would reject, and that is pre-existing — transition validation already rejects those moves today, as the seam-2 cases in `moves-flag-equivalence.test.ts` demonstrate.

**The real fix, filed rather than smuggled in:** make `moveSource` explicit at those 26 call sites so the default never has to be guessed. Until then the parameter stays deliberately unused, and I left the `void moveSource;` with that reason recorded at the site instead of leaving it looking like an oversight — which is how it read to me the first time, and why I changed it.

Verification after the revert: move-path suites 21/21, `pnpm test:gate` green (10 / 158 / 487 / 71), `pnpm check:lifecycle-columns` exits 0, `tsc` clean, lint clean. `triage` remains at **zero** on this branch.

gsxdsm:
## RETRACTION — I was wrong. `main` is NOT red, and #2655 is not load-bearing for CI.

My previous comment on this PR claimed clean `origin/main` fails the required census check with `moves.ts: 15 -> 39`, that the baseline was pinned to a tree containing #2655, and that every open PR was consequently red. **All of that is false.** Please disregard it; I am sorry for the noise and for any sequencing decisions it prompted.

### The truth, read from the committed blob rather than my working tree

```
git show origin/main:scripts/lib/lifecycle-column-census-baseline.json
  moves.ts:        39
  totals.column:   748

pristine origin/main checkout:
  pnpm check:lifecycle-columns   exit 0
```

The baseline says 39, main measures 39, the gate passes. It also said 39 at `a6138abeff`, the exact commit I claimed was broken.

### How I produced a false red

Earlier in the session I ran `node scripts/lifecycle-column-census.mjs --strict --update-baseline` while probing a **merged** tree (#2655 + main). That rewrote the tracked baseline file to the merged tree's numbers — including `moves.ts: 15`.

I then ran `git checkout origin/main --detach`. **Checkout preserves a local modification that applies cleanly**, so the rewritten baseline followed me onto "clean main". Every subsequent reading — the 15, the "pinned to a future tree" theory, the whole comparison table — was my own probe output being mistaken for the repository's state. The numbers were internally consistent, which is exactly why it read as a finding rather than an error.

`git status` would have shown it in one line. I ran `git clean -qfd`, which removes untracked files and does nothing about a modified tracked one.

### What still stands from that comment

Only the part measured on a properly checked-out branch:

- **#2655 does close the bar.** On its own branch `moves.ts` holds 15 column guards and **zero** `triage` sites; merged with main the tree reports **backlog 724, triage 0**.
- The full battery on that merged tree is green: `check:lifecycle-columns` exit 0, `pnpm test:gate` exit 0 (726 tests), all 18 pg E2E suites 119 tests / 0 failures, `pnpm verify:fast` exit 0.

What does **not** stand: any claim that main is broken, that CI is blocked, or that #2655 must land urgently to unblock others. Merge it on its own merits and schedule.

### Correcting the correction

My previous comment closed with a note about verifying probes before trusting them, listing two failures I had caught. It then contained a third I had not caught — from the same class, in the same investigation. The specific lesson is narrower than "verify the probe": **a command that writes to the working tree contaminates every later read, including reads on a different branch.** Probes that mutate tracked files need to run in a throwaway worktree, or be reverted before the next measurement.

gsxdsm:
This describes the pre-revert state — the change it flags was already reverted in `revert(U12): restore the shipped bypass read`, one commit before this comment landed. Current branch state:

```
packages/core/src/task-store/task-store-helpers.ts:130
  (options?.moveSource === "engine" || options?.moveSource === "scheduler" || options?.skipMergeBlocker === true)
```

So an optionless `moveTask(id, toColumn)` does **not** bypass, and the `moves.ts:346` comment you quoted is accurate again rather than contradicted.

Your framing is the right one though, and it names the real problem better than either round did: the derivation should distinguish *"explicitly engine-sourced"* from *"defaulted"*. It does not, and that ambiguity is why this flipped twice — 18 optionless calls in `packages/engine` are genuine engine moves that arguably should bypass, and 8 in dashboard HTTP routes must not, and one `??` serves both.

**So rather than just resolving this, I pinned the contract.** It lived in a comment, and a comment could not stop me breaking it: the `void moveSource;` beside the read looks exactly like a silenced unused-parameter lint, which is why I "fixed" it. `optionless-move-does-not-bypass-guards.test.ts` now asserts it, mutation-verified against my own reverted change — re-applying it fails two cases with `expected true to be false`. It would have caught this in one round instead of two.

It covers both directions so it cannot pass by returning false unconditionally: explicit `engine`/`scheduler`, `skipMergeBlocker` and `recoveryRehome` do bypass; an explicit `bypassGuards: false` still wins over an engine source; a user-sourced move does not.

The ambiguity itself stays open, recorded in the test header rather than left implicit. The fix is making `moveSource` explicit at those 26 call sites so the default is never guessed — real work, and not something to attach to a flag-resolution PR whose whole argument is that it changes no behaviour beyond the flip.

Verification: `pnpm test:gate` green (10 / 158 / 487 / 71), `pnpm check:lifecycle-columns` exits 0, `tsc` clean, lint clean. `triage` remains at zero on this branch.

gsxdsm:
## Closing verification, run on this branch (i.e. the post-merge tree)

Run pre-emptively so failures surface before merge rather than after. All four legs, one tree, one report.

| leg | result |
|---|---|
| **merge gate** (`pnpm test:gate`) | **PASS** — 10 / 158 / 487 / 71 |
| **`pnpm verify:fast`** | **PASS** — 15 steps green in 39.0s, boot smoke OK |
| **E2E — all live-PG suites** | 4 failed / 120 passed files; **identical failure set to `origin/main`** |
| **census** (`pnpm check:lifecycle-columns`) | **exit 0** |

### `triage` is ZERO

```
triage: ABSENT — zero column guards
```

No unconverted `triage` lifecycle-column guard remains anywhere in production source. Closing-bar item 1 is met on this tree.

Full census on this branch:

```
COLUMN guards (the backlog):   724
ROLE comparisons (not guards):   5
STATUS comparisons (not guards): 186
DELIBERATE-LITERAL (reviewed):  17
QUERY filters:                  83
IR node definitions:            43

by column id:  in-review 200 | done 196 | archived 148 | in-progress 138 | todo 42 | triage —
```

### On the E2E failures — attributed, not assumed

The 4 failing PG suites (`agent-logs-and-monitor`, `central-archive-secrets`, `store-wedge-resolution`, `workflow-settings-project-identity`) fail **identically on pristine `origin/main`**. I measured that by checking out main, re-running the same command, and diffing the failing-file **sets** — `comm -13` returns empty, so **zero suites fail exclusively on this branch**.

That method matters: my first attempt at this comparison on the core suite used counts alone (52 vs 33) and would have blamed the flip for 8 files that were already red. Counts cannot attribute; sets can.

Same for the core suite: **33 failed / 10 files, identical to main**, zero exclusive.

### What this does not clear

The 4 pre-existing PG failures are real and unowned — they are simply not caused by this PR. They belong to whoever owns those subsystems, and they will still be red after this merges.

gsxdsm:
## Integration check: all five PRs merged together, verified as one tree

No single PR proves the set works together, so I merged all five onto current `origin/main` locally and ran the full verification. **They interact cleanly** — the only conflict was the census baseline, which is the expected friction of a checked-in counter and resolves by re-recording.

Branches combined: #2676 (baseline un-red), #2669 (PG wedge), #2675 (audit-emitter assertion), #2674 (funnel hold trait), #2655 (move-path flag resolution).

### Result

| leg | main (measured) | all five combined |
|---|---|---|
| **live-PG E2E** | 4 failed files / 11 failed tests | **1 failed file / 1 failed test** |
| **core suite** | 33 failed / 10 files | **24 failed / 8 files** |
| **merge gate** | — | **PASS** (10 / 158 / 487 / 71) |
| **`check:lifecycle-columns`** | **exit 1** (red on pristine main) | **exit 0** |
| **census `triage`** | 4 | **ABSENT — zero** |
| **census column total** | 746 | **722** |

### The one surviving E2E failure is the decision I escalated

`agent-logs-and-monitor.pg.test.ts` — the SDLC funnel counts 0 entries into the `todo` stage, because the merged Planning column carries `intake` + `reset-on-entry` and `stageForTraits` prefers the earliest stage in flow order, so it resolves to `triage`. #2674 fixed the unambiguous half (hold-only columns vanishing entirely) and deliberately pinned this behaviour rather than changing it, because altering which stage Planning reports retroactively changes how historical analytics read.

So after these five land, **the E2E leg is one failure away from clean, and that one failure is a product decision rather than an engineering task.**

### Why I ran this

Five separately-reviewed PRs touching overlapping files (`activity-analytics.ts`, the census baseline, core store internals) can each be green and still break in combination. This is the only check that would have caught that, and it is cheap. The single conflict it surfaced — the baseline — is also the thing that has repeatedly gone stale on main, so it is worth knowing it resolves mechanically rather than semantically.

PR #2657, opened 2026-07-30
fix(dashboard): restore traits-over-legacy-id — a reviewed fix that #2645 merged without

gsxdsm:
## A merged PR does not contain its own fix

coderabbit raised a **Major** on #2645: `isPreExecutionHoldColumn("triage", { intake: false, hold: false })` returned `true`, so explicit resolved traits saying *"planning does not happen in this column"* were overruled by the legacy id, exposing the **Plan** action on a card whose workflow says otherwise.

I fixed it, answered the thread, and pushed to `consolidate/u11`. **#2645 is now merged — and main does not have the fix.** `origin/main` still reads:

```ts
return column === "triage" || flags?.intake === true || flags?.hold === true;
```

The merge commit `177c2309d` touches three files of unrelated planner-lane work. My commit appears nowhere in `git log --all`. The branch was force-pushed by another worker while I did not hold it, and the PR merged under a number whose review thread describes a different change. Lost with it: the three tests and the `isPreExecutionIntakeLane` / `legacyLifecycleColumns` helpers.

I flagged this exact risk when I noticed another worktree had taken the branch — *"a second agent force-pushing that branch would drop it."* Recording it plainly because **the failure mode is invisible**: the PR is green, closed, merged, and its review threads read as resolved. Nothing would ever have revisited it. If other workers shared consolidation branches, this is worth auditing beyond my one case.

## What this PR does

Restores the predicate **minimally**, against main's actual code — no reintroduction of helpers that no longer exist here. Traits decide when present; the legacy id is the no-metadata fallback only, which still covers a card stranded on an id its workflow no longer declares (that card has no flags at all).

## Verification

**Mutation-verified: reverting to the disjunct fails exactly 2 of 20** — the classifier case and the caller-level Plan-action case. 20/20 with the fix; dashboard `tsconfig.app.json` typecheck clean.

The caller-level test injects `onPlan` deliberately: Plan is omitted entirely when the host does not wire it, so without that the assertion would pass for the wrong reason.

## Census

Unchanged (787 / triage 10). This is a **behavior** fix — the literal survives as the documented fallback, so it is not a conversion and does not move the count. Worth stating because #2645's title claimed `triage 11 -> 10` for work that did not include this.

### DIFF
--- a/packages/dashboard/app/components/TaskContextMenu.tsx
+++ b/packages/dashboard/app/components/TaskContextMenu.tsx
@@ -191,6 +191,16 @@ export function isPreExecutionHoldColumn(column: string, flags?: TaskContextMenu
   the PR that reported it is closed.
   */
   if (flags) return flags.intake === true || flags.hold === true;
+  /*
+  FNXC:WorkflowLifecycleColumns 2026-07-30-01:40 DELIBERATE-LITERAL: the fallback arm only.
+  Now that traits decide first, this is reachable ONLY when the caller supplies no flags at all —
+  the pre-load window before the board resolves columns, and a card stranded on an id its workflow
+  no longer declares. There is nothing to resolve FROM in that state, so this is not an unconverted
+  guard: deleting it does not remove a decision, it picks the other guess ("not a pre-implementation
+  lane") and silently drops the Plan affordance during first paint. Same class as the marked
+  fallbacks in `TaskCard.tsx`, `TaskDetailModal.tsx`, and `live-agent-count.ts`. Retires with the
+  pre-load window.
+  */
   return column === "triage";
 }
 
--- a/packages/dashboard/app/components/TaskContextMenu.tsx
+++ b/packages/dashboard/app/components/TaskContextMenu.tsx
@@ -176,7 +176,22 @@ export function isPreExecutionHoldColumn(column: string, flags?: TaskContextMenu
   Same shape, different degraded answer: the trait path is identical and the fallbacks are not
   interchangeable. Kept separate with the difference recorded, rather than made to look shared.
   */
-  return column === "triage" || flags?.intake === true || flags?.hold === true;
+  /*
+  FNXC:WorkflowResolvedColumns 2026-07-30-01:15 (RESTORED — see below):
+  RESOLVED TRAITS ARE AUTHORITATIVE. The legacy id was an unconditional disjunct, so
+  `isPreExecutionHoldColumn("triage", { intake: false, hold: false })` returned true — explicit
+  traits saying "planning does not happen in this column" were overruled by the id, exposing the
+  Plan action on a card whose workflow says otherwise. The id now serves only as the NO-METADATA
+  fallback, which still covers a card stranded on an id its workflow no longer declares (that card
+  has no flags at all).
+
+  PROVENANCE: this fix was reviewed and approved on PR #2645 (coderabbit, Major) and then LOST —
+  the consolidation branch it lived on was force-pushed by another worker, and #2645 merged
+  unrelated content under its number. Restored here with its test, because the defect is real and
+  the PR that reported it is closed.
+  */
+  if (flags) return flags.intake === true || flags.hold === true;
+  return column === "triage";
 }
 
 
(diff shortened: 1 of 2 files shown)


gsxdsm:
## Also closes the last non-`moves.ts` site: triage 10 → 9

Added the `DELIBERATE-LITERAL` marker to the surviving fallback, which closes the one-site gap between *"triage = 0 counted"* and *"triage = 0 unreviewed"*.

Once traits decide first, that `column === "triage"` is reachable **only** with no flags at all — the pre-load window before the board resolves columns, and a card stranded on an id its workflow no longer declares. Nothing to resolve from, so it is not an unconverted guard: deleting it picks the other guess ("not a pre-implementation lane") and silently drops the Plan affordance during first paint. Same class, now same treatment, as the marked fallbacks in `TaskCard.tsx`, `TaskDetailModal.tsx`, and `live-agent-count.ts`.

**Measured on this branch: 787 → 786, triage 10 → 9, reviewed 3 → 4.**

### Placement gotcha, worth knowing for the fleet

My first attempt marked nothing. The census excuses *the construct the comment is attached to and everything inside it* — I had attached it to the function's doc comment, while the literal lives in the next **sibling** statement. It has to lead the statement holding the literal.

I caught it because the count did not move. A fleet worker who assumes the marker took would report a shrink that never happened, and until #2654 wires `--strict` into CI, nothing contradicts them. **Verify markers by the count moving, not by the comment existing.**

### The bar now closes arithmetically

| source | triage |
|---|---|
| main today | 10 |
| #2593 (5 reviewed fallbacks marked) | −5 |
| this PR (1 marked) | −1 |
| U12 flag flip (`moves.ts` ×4 die with the flag-OFF branch) | −4 |
| **remaining** | **0** |

No further conversion work is needed to reach zero — the remaining four are deletions that come free with the flag resolution.

20/20 tests, dashboard typecheck clean, mutation proof unchanged (reverting the predicate still fails exactly 2 of 20).

gsxdsm:
Closing: **superseded by #2664**, which landed the identical fix. Main now reads:

```ts
return flags ? (flags.intake === true || flags.hold === true) : column === "triage";
```

That is the traits-first shape this PR restored — traits authoritative, the legacy id as the no-metadata fallback only. Main also has caller-level coverage for it, so the tests are not orphaned either.

Worth recording why this PR existed: the fix was reviewed and approved on #2645, then lost when that branch was force-pushed, and #2645 merged unrelated content under its number. #2664 independently rediscovered and fixed the same defect. **The rediscovery cost is the real finding** — a force-push on a shared branch discarded reviewed work while every signal stayed green, and it took a second worker finding the same bug from scratch to recover it. Worth a rule before the fleet runs on shared branches.

gsxdsm:
Rebased onto current main and green. **Census: triage 5 → 4**, deliberate 12 → 13.

Worth recording what survived and what didn't, because it isn't what the PR title suggests:

**The first commit is now redundant.** Main already carries the traits-first inversion in `isPreExecutionHoldColumn`, with a fuller rationale than the one here — including the concrete defect (a resolved column *named* `triage` whose traits say mid-flight answered true, offering Plan on an executing card). Skipped it rather than merge two revisions of the same fix.

**The second commit is genuinely additive**, and I nearly skipped it too. Both sides are behaviourally identical — `flags ? (intake || hold) : column === "triage"` versus the same logic as an early return — so the diff reads as a no-op reformat. The difference is the **DELIBERATE-LITERAL marker**, which is exactly what moves that literal out of the backlog bucket. Main's version lacks it, so the census still counted a reviewed fallback as unconverted work.

That's the check I got wrong on #2616 an hour ago: I compared the production diff, saw it was upstream, and concluded the PR was empty — when two test files were unique. Same trap here in a different costume, and this time the second commit is the whole point of the PR.

**Attribution, since the commit message raises it:** the traits-first change was lost when I force-pushed #2645 after reducing it. That was my error — I rebuilt that branch from main and re-applied only what I judged still additive, and this fell outside that judgement. Thanks for catching it.

Verified: 18 tests green, dashboard tsc clean, lint clean, merge gate green (487 + 158 + 10).

PR #2661, opened 2026-07-30
U12: the census baseline follows the count down — and goes in the merge gate

gsxdsm:
Coordinator item 2. The census had the right mechanism and no teeth.

## The gap

`--strict` already fails on a rise **and** on an unrecorded drop — that logic was correct. But nothing blocking ran it, so the baseline drifted to **854 while the tree held 787**. That is **67 guards of regression that would have merged silently**: a high-water mark wearing a ratchet's name.

This is the same shape as the ceilings I tightened in #2647, one level up. Worth saying plainly: I fixed the vitest ratchet's slack by hand and did not check whether the *authoritative* instrument had the same problem. It did, and by a much larger margin.

## Three changes

1. **`--strict` runs in `test:gate`.** The baseline cannot go stale again without a red gate.
2. **Baseline re-recorded: 854 → 785** across 14 files (`triage` 38 → 9).
3. The single RISE is resolved honestly rather than absorbed.

## The +3 investigation

One file rose: `register-task-workflow-routes.ts` **22 → 23**. #2621 replaced one `task.column === "todo"` with `task.column === "triage" || task.column === "todo"` — a net **+1** that also reintroduced a `triage` literal, while the PR title reported *"count 0 → 0"*.

Not an accusation. There was no gate for the author to check against, and a hand-counted claim in a PR title is exactly the thing that goes wrong without one. Change 1 is the fix.

**The literal is justified and stays**, marked `DELIBERATE-LITERAL` rather than converted. It is the **v1-IR arm**: a v1 workflow yields no role assignments, so `resolveLifecycleColumns` returns nothing and the legacy pre-implementation ids are the only pre-WIP signal available. The `else` branch directly below already resolves intake/hold for every v2 workflow. Converting this arm would not finish anything — it would delete the only answer v1 boards have and admit `in-progress`/`in-review` cards into a rebound that clears worktree, branch and retry counters, which is the regression #2621 was fixing.

## Both directions proven

| direction | probe | result |
|---|---|---|
| rise | add `t.column === 'in-review'` | `live-agent-count.ts: 6 -> 7`, exit 1 |
| drop | convert one guard | `self-healing.ts: allows 111, tree has 110`, exit 1 |

**The drop probe took three attempts to test honestly, and the first two "passed" while proving nothing:**

1. I renamed a receiver (`task.column` → `Probe`) — the classifier is **fail-closed**, so an unknown receiver is still counted and the number never moved.
2. I targeted a site in `hold-release.ts` that carries a `DELIBERATE-LITERAL` marker — not counted as a column guard at all, so removing it changed nothing.

Only removing a counted comparison outright moved the number. Both false negatives came from me assuming the probe worked because the command exited the way I expected.

## On auto-rewrite vs fail-and-instruct

You offered either. The script already does **fail-and-instruct**, with `--update-baseline` as the explicit re-record, and I kept it that way rather than making the test rewrite the baseline during a run.

Reason: a silent downward rewrite means a conversion PR's own diff never shows the number moving, so "census before/after in the PR body" becomes unverifiable — the reviewer would have to re-derive it. Failing with the new number in the message puts it in the diff where a human sees it, and it costs one command.

## Verification

`pnpm lint` clean. `pnpm test:gate` green with the census in it — `every file matches its baseline exactly` (10 / 132 / 487 / 71).

Note for the fleet launch: with `--strict` gating, **every** conversion PR must now re-record the baseline in the same PR. That is the intended cost, and it makes the fleet's "baseline must shrink by exactly the converted count" rule mechanically enforced instead of a review instruction.


gsxdsm:
## Three coordination facts about the baseline, including a correction to something I said

I proved the ceiling behaviour empirically on #2650 (injected a guard into `self-healing.ts` at 111 against an allowance of 126 — it did not appear in the regression list). So item 2 and this PR are right. Two follow-ups and one retraction.

### 1. Correcting myself: my "second defect" is a reporting wart, not a hole

I reported that `if (regressions.length > 0) { … exit(1) }` running *before* the stale block is a second, independent defect, and said the next genuine rise would "re-hide all remaining headroom". **That overstated it, and I should not have called it a hole.**

With `--strict` actually enforced — which is exactly what this PR and #2654 do — a masked stale entry cannot persist: the rise fails the gate, someone fixes the rise, the next run reports the stale, and that fails too. It costs an extra round trip; it does not let anything through. Worth stating plainly so nobody builds machinery for it. Collecting both lists before exiting is a nice-to-have, not a correctness fix.

Re-recording to current values is a complete fix for the ceiling, because it removes every file's headroom at once — after this PR, `self-healing.ts` is pinned at its true count and my injected guard would be caught.

### 2. This PR and #2654 both install the enforcement

- **#2661** (here): `--strict` in `test:gate`
- **#2654**: `pnpm check:lifecycle-columns` as a `pr-checks.yml` step, plus the `package.json` script

Both are reasonable and they are not the same surface, but landing both means the ratchet runs twice per PR and the two can disagree about scope. Worth a deliberate choice rather than discovering it after both merge.

### 3. Six open PRs edit this baseline; one needs a manual seed regardless

```
#2661  #2654  #2652  #2650  #2644  #2593
```

For five of them the reconciliation after the first merge is just `--strict --update-baseline`, which this PR's re-record makes routine.

**#2650 is the exception and it is my PR, so this is a flag on my own work.** It introduces *new* baseline keys (`properties`, `queryByColumnId`, `queryByFile`) for the query-filter category. `--update-baseline` cannot create them: the regression check exits before the write, so with no prior key every file reads as a rise from zero and the update never happens. Verified — the command exits 1 and leaves the file untouched.

The reconciliation is a one-time manual seed of those three keys, after which it is self-maintaining. **Cheapest ordering is to land #2650 last**, so the seed happens once rather than once per subsequent merge. I am not asking for priority — the opposite: put it at the back.

gsxdsm:
Both addressed. The first one I had to narrow before I could fix it, and the narrowing matters.

**Marker scope.** The stated mechanism does not hold: the marker attaches to the single assignment, not the enclosing `if`. Measured by adding a comparison as a **sibling statement** inside that same conditional — it is **counted**, `21 -> 22`, gate red.

But there is a real hole underneath it, and it is broader than this site: `hasDeliberateMarker` walks ancestors *by design* — its own comment says "a marker excuses the construct it is attached to and everything inside it" — so a comparison **appended to an already-marked expression** inherits the exemption and never reaches the per-file counts. That applies to every marker in the codebase, not just mine.

Closed generally by pinning `totals.deliberate`. Growing a marked expression moves that number, so it must be re-recorded like any other change — which puts it in the diff where a reviewer sees it rather than in a hand-written claim. Proven: appending `|| task.column === "done"` now fails with

```
lifecycle-column-census --strict: DELIBERATE-LITERAL count ROSE (5 -> 6)
```

exit 1.

**A false start worth recording**, because it is the same failure this PR is about. My first version of that check silently did nothing: I anchored the insertion on `if (regressions.length > 0)`, which occurs **twice** in the file, so the block landed in an earlier section before `baseline` is even declared. The probe "passed" because the check never ran — I only caught it by measuring that `deliberate` really did move 5 → 6 and then asking why `--strict` still exited 0. A guard inserted in dead code is indistinguishable from a guard that works, right up until you need it.

**Stale header.** Correct and now fixed. The script said "NOT wired into the merge gate", which this PR makes false. I replaced it with what is true, plus why the original reasoning expired — it argued a thousand-site backlog cannot be blocking on the day it is first measured, which was sound, but the baseline is **per-file**, so gating costs nothing for files nobody touches, and in the meantime it drifted to 854 against a tree of 787.

I also recorded the cost in the header rather than leaving it for people to discover: lowering a count now requires re-recording the baseline in the same PR.

Verification: `pnpm lint` clean, `pnpm test:gate` green with the census in it — `every file matches its baseline exactly` (10 / 132 / 487 / 71).

gsxdsm:
Real, and it is the same mistake this PR exists to fix — I made it one field over, immediately after writing the argument against it.

A repo-wide total is defeated by an **offset**: remove one marked comparison in file A while another marked construct in file B gains one, the total stays flat, the check passes. And because deliberate findings are excluded from `byFile`, the newly exempt guard is invisible there too — so the gate goes green with a fresh lifecycle-column guard sitting inside an exemption. That is precisely the high-water shape I spent this PR removing from the column counts.

Now compared **per file**, the same shape `byFile` already uses, because offsetting edits land in different files.

**Proven with your exact scenario:** I added a comparison inside `MissionControlPanel`'s marked `FUNNEL_STAGES` and removed one from marked `hold-release.ts`. The repo total is unchanged (`DELIBERATE-LITERAL (reviewed): 17` either way) and `--strict` now fails with

```
packages/dashboard/app/components/command-center/MissionControlPanel.tsx (DELIBERATE-LITERAL): 3 -> 4
```

**One migration case I handled explicitly**, because getting it wrong would have been worse than the bug: a baseline recorded before this field existed has no `deliberateByFile` at all, and that is **not** the same as "every marked file had zero". Comparing against an absent map reports every existing marker as a fresh rise and demands people convert literals that were already reviewed — I hit exactly that on the first run, with `MissionControlPanel: 0 -> 3`, `TaskCard: 0 -> 3`, `hold-release: 0 -> 3`. It now seeds on the next `--update-baseline` and compares from then on.

Baseline re-recorded on top of current main: `column` 765, `triage` 4, `deliberateByFile` seeded for 8 files.

Verification: `pnpm lint` clean, `pnpm test:gate` green (10 / 158 / 487 / 71), `pnpm check:lifecycle-columns` exits 0.

Worth restating what this PR is ultimately about, since the thread has moved a long way from the title: **`check:lifecycle-columns` is currently RED on pristine `origin/main`** and failing every open PR, because a `DELIBERATE-LITERAL` marker was orphaned by a declaration inserted between it and the table it was written for. That fix is the first commit here.

PR #2662, opened 2026-07-30
census: main is RED on the ratchet — mark the 4 unmarked-but-justified guards (779 → 775)

gsxdsm:
**`node scripts/lifecycle-column-census.mjs --strict` exits 1 on `origin/main` right now.** Item 4's final verification pass cannot go green until this is fixed, so this PR fixes it.

## The +3 investigation, answered

It is **+4**, it is **still on main**, and the answer to "were they justified?" is **yes in substance, no in bookkeeping** — both merges added real guards and neither marked them.

| file | delta | source | verdict |
|---|---|---|---|
| `MissionControlPanel.tsx` | 0 → 3 | #2645 ("funnel aliases") | justified — **alias table**, not lifecycle guards |
| `columnRoles.ts` | 0 → 1 | #2647 (`isHoldColumnRole`) | justified — documented no-metadata fallback |

**MissionControlPanel** maps operator-chosen column names (`"to do"`, `"ready"`, `"shipped"`) onto the five canonical funnel stages. Those arms do not ask what lifecycle role a column plays — they normalise free text *before* any role lookup can happen, so there is no trait to read. The file **already carried a `DELIBERATE-LITERAL` marker**, but on `TRIAGE_STAGE_COLUMN_ALIASES`; the census excuses only the construct a marker is attached to, so the sibling `FUNNEL_STAGES` const needed its own.

**columnRoles** `isHoldColumnRole` is a no-metadata fallback whose asymmetry with `isPreImplementationColumnRole` is already spelled out in its own doc comment. Only the machine-readable token was missing.

So no bad code landed. The ratchet was correctly reporting **unrecorded growth**, and nobody saw it because nothing runs the census — which is #2654.

## This is the second time the same placement trap has bitten

A marker on a *sibling* construct counts for nothing. It caught #2645's author, and it caught me an hour ago on `TaskContextMenu` (#2657) — my first attempt marked nothing and I only noticed because the count did not move.

**For the fleet brief: verify a marker by the count moving, not by the comment existing.** Until #2654 lands, nothing else will tell you.

## Verification

- `--strict` **exit 1 → 0**
- **779 → 775** total, DELIBERATE 12 → 16, **triage unchanged at 5**
- baseline re-recorded in the same commit, per the tool's instruction — that also closed one stale allowance (`worktreeGrouping.ts` 4 → 3)
- dashboard typecheck clean, **165/165** across the touched suites

## Where the closing bar stands after this

Main's remaining triage guards are `moves.ts` ×4 (die with U12's flag flip) and `TaskContextMenu.tsx:179` (marked by #2657). **Triage reaches 0 with U12 + #2657 — no further conversion work.**

### DIFF
--- a/packages/dashboard/app/components/command-center/MissionControlPanel.tsx
+++ b/packages/dashboard/app/components/command-center/MissionControlPanel.tsx
@@ -70,6 +70,14 @@ receiver is a column id, purpose is presentation, not a lifecycle decision.
 */
 const TRIAGE_STAGE_COLUMN_ALIASES: ReadonlySet<string> = new Set(["triage", "signal", "backlog"]);
 
+/*
+FNXC:WorkflowLifecycleColumns 2026-07-30-01:55 DELIBERATE-LITERAL: alias table, not lifecycle guards.
+Same rationale as the marker above, which covers `TRIAGE_STAGE_COLUMN_ALIASES` but NOT this separate
+declaration — the census excuses the construct a marker is attached to, so a sibling const needs its
+own. These arms map arbitrary OPERATOR-CHOSEN column names ("to do", "ready", "shipped") onto the
+five canonical funnel stages. They are not asking "what lifecycle role does this column play?" —
+they are normalising free text before any role lookup can happen, so there is no trait to read.
+*/
 const FUNNEL_STAGES: Array<{ id: string; match: (column: string) => boolean }> = [
   { id: "triage", match: (c) => TRIAGE_STAGE_COLUMN_ALIASES.has(c) },
   { id: "todo", match: (c) => c === "todo" || c === "to-do" || c === "to do" || c === "ready" },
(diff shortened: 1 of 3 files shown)


gsxdsm:
Closing: **main is green on the ratchet.** `--strict` exits 0 at `efbbc45eb0` (768 total, triage 4), and `MissionControlPanel` carries its markers there.

This PR was opened when main was RED with unrecorded growth (`MissionControlPanel` 0 → 3, `columnRoles` 0 → 1). That is resolved.

One trivial remainder for anyone passing through: `isHoldColumnRole` in `columnRoles.ts` is still counted as an unconverted guard, though its own doc comment already explains why the literal is correct — it needs only the `DELIBERATE-LITERAL` token to move from backlog to reviewed. Not worth a PR on its own; fold it into the next change touching that file.

**Placement note worth keeping** (it has now bitten twice — #2645's author and me): the marker must lead the STATEMENT holding the literal. On a sibling construct it counts for nothing. Verify by the count moving, not by the comment existing.

PR #2666, opened 2026-07-30
U12: DELIBERATE-LITERAL counts key on file AND column — closing the P1 left on merged #2661

gsxdsm:
Closes the P1 that was still open when #2661 merged.

## The hole

A per-file integer is offset **within a single file**: remove one reviewed `todo` exemption, add an `in-review` one beside it, and the number never moves. The fresh guard is invisible to the column counts too, because deliberate findings are excluded from them — so `--strict` goes green with a new lifecycle-column guard hiding inside an existing marker.

Now keyed on **file AND column id**.

**Proven with the exact scenario:** swapping a marked `triage` for `done` inside `TaskCard.tsx` leaves the per-file total unchanged and now fails with

```
packages/dashboard/app/components/TaskCard.tsx (DELIBERATE-LITERAL: done): 0 -> 1
```

## The pattern worth naming

This is the **third** time this instrument has been defeated by an aggregate:

| version | defeated by |
|---|---|
| repo-wide `totals.deliberate` | an addition in file A offset by a removal in file B |
| per-file integer | an addition offset by a removal **in the same file** |
| per-file per-column | — |

Each step narrows what can offset silently, and I walked into the next one twice by fixing the *reported case* rather than the *shape*. Writing it down because the same reflex will produce a fourth if someone adds another aggregate here.

**The residual is deliberate, not an oversight:** a same-file **same-column** swap still offsets. Two `todo` exemptions in one file are interchangeable by definition, so there is nothing a reviewer could act on. That is recorded at the site so the next person doesn't rediscover it as a bug.

## Migration, again

The key **shape** changed (`file` → `file\0columnId`), which is the same hazard as a missing field: comparing new keys against old reports every existing marker as a fresh rise and pushes people to convert already-reviewed literals. I hit it on the first run here — `TaskCard (DELIBERATE-LITERAL: triage): 0 -> 2` — exactly as I did one shape earlier in #2661.

Detected by the delimiter rather than a version field, since old keys have none, and re-seeded on the next `--update-baseline`. 15 file+column entries recorded.

## Verification

`pnpm lint` clean. `pnpm test:gate` green (10 / 158 / 487 / 71). `pnpm check:lifecycle-columns` exits 0.

Independent of #2655; either order merges.

### DIFF
--- a/scripts/lib/lifecycle-column-census-ast.mjs
+++ b/scripts/lib/lifecycle-column-census-ast.mjs
@@ -303,7 +303,19 @@ export function summarize(findings) {
     }
     totals[finding.kind] += 1;
     if (finding.kind === "deliberate") {
-      deliberateByFile.set(finding.file, (deliberateByFile.get(finding.file) ?? 0) + 1);
+      /*
+      FNXC:WorkflowLifecycleColumns 2026-07-31-10:00 (PR #2661 review — greptile P1, same class again):
+      Keyed by FILE **and COLUMN ID**, not a per-file integer. A per-file aggregate is offset within a
+      single file: remove one reviewed `todo` exemption, add a `in-review` one beside it, and the
+      number never moves — so a fresh guard hides inside an existing marker.
+
+      That is the third time this instrument has been defeated by an aggregate (repo total -> per file
+      -> per file per column). Each step narrows what can offset silently. The residual is a same-file
+      SAME-COLUMN swap, and that one is deliberate: two `todo` exemptions in one file are
+      interchangeable by definition, so there is nothing a reviewer could act on.
+      */
+      const key = `${finding.file}\u0000${finding.columnId}`;
+      deliberateByFile.set(key, (deliberateByFile.get(key) ?? 0) + 1);
     }
     if (finding.kind !== "column") continue;
     byColumnId[finding.columnId] = (byColumnId[finding.columnId] ?? 0) + 1;
--- a/scripts/lib/lifecycle-column-census-baseline.json
+++ b/scripts/lib/lifecycle-column-census-baseline.json
@@ -155,14 +155,21 @@
     "plugins/fusion-plugin-reports/src/store/report-types.ts": 1
   },
   "deliberateByFile": {
-    "packages/dashboard/app/components/command-center/MissionControlPanel.tsx": 3,
-    "packages/dashboard/app/components/TaskCard.tsx": 3,
-    "packages/dashboard/app/components/TaskDetailModal.tsx": 3,
-    "packages/engine/src/hold-release.ts": 3,
-    "packages/dashboard/src/routes/register-task-workflow-routes.ts": 2,
-    "packages/dashboard/app/components/TaskContextMenu.tsx": 1,
-    "packages/dashboard/app/utils/columnRoles.ts": 1,
-    "packages/engine/src/triage.ts": 1
+    "packages/dashboard/app/components/TaskCard.tsx\u0000triage": 2,
+    "packages/dashboard/app/components/TaskDetailModal.tsx\u0000triage": 2,
+    "packages/dashboard/app/components/command-center/MissionControlPanel.tsx\u0000done": 1,
+    "packages/dashboard/app/components/command-center/MissionControlPanel.tsx\u0000in-review": 1,
+    "packages/dashboard/app/components/command-center/MissionControlPanel.tsx\u0000todo": 1,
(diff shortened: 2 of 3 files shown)


gsxdsm:
## `main` is RED on the required census check, and it stays red until #2655 merges

`pnpm check:lifecycle-columns` is a required PR check. On clean `origin/main` (a6138abeff):

```
exit 1   packages/core/src/task-store/moves.ts: 15 -> 39
```

Every open PR fails CI on this, and none of them caused it.

### Diagnosis: main's baseline is pinned to a tree that includes #2655

Measured, each step verified:

| tree | `moves.ts` column guards |
|---|---|
| clean `origin/main` | **39** |
| #2655 branch alone | **15** |
| #2655 merged with main | **15** |
| **baseline on main says** | **15** |

`moves.ts` carries **zero** `DELIBERATE-LITERAL` markers, so no exemption explains the 24-guard gap. The only tree that satisfies the pinned 15 is one containing #2655's flag resolution — which is still open.

So the baseline was re-recorded against a working tree that already had the flag resolution applied. The pin is correct about the *future* and wrong about `main`.

### It self-resolves on merge — verified, not assumed

```
check:lifecycle-columns on (#2655 + main):  exit 0
backlog: 724      triage: 0
```

### What this means for sequencing

- **#2655 is now load-bearing for CI, not just for the bar.** Until it lands, every PR in the program is red on a required check through no fault of its own.
- Nothing else needs fixing. Re-recording main's baseline to 39 would turn the gate green *and* concede 24 guards the flag resolution is about to delete — the wrong direction, and it would have to be undone immediately.
- If #2655 cannot land promptly, the honest interim is to make the check non-blocking rather than to loosen the pin.

### Method note

I got this wrong twice before getting it right, and both failures looked like answers. A `git fetch` of the PR ref failed with its stderr suppressed, and a later `git checkout` aborted on local changes — in both cases I measured `main` while believing I was measuring the PR branch, and the numbers were self-consistent enough to be convincing. The table above was rebuilt with the branch identity, the merge-commit count, and the file content asserted at each step before any number was read.

gsxdsm:
## RETRACTION — I was wrong. `main` is NOT red, and #2655 is not load-bearing for CI.

My previous comment on this PR claimed clean `origin/main` fails the required census check with `moves.ts: 15 -> 39`, that the baseline was pinned to a tree containing #2655, and that every open PR was consequently red. **All of that is false.** Please disregard it; I am sorry for the noise and for any sequencing decisions it prompted.

### The truth, read from the committed blob rather than my working tree

```
git show origin/main:scripts/lib/lifecycle-column-census-baseline.json
  moves.ts:        39
  totals.column:   748

pristine origin/main checkout:
  pnpm check:lifecycle-columns   exit 0
```

The baseline says 39, main measures 39, the gate passes. It also said 39 at `a6138abeff`, the exact commit I claimed was broken.

### How I produced a false red

Earlier in the session I ran `node scripts/lifecycle-column-census.mjs --strict --update-baseline` while probing a **merged** tree (#2655 + main). That rewrote the tracked baseline file to the merged tree's numbers — including `moves.ts: 15`.

I then ran `git checkout origin/main --detach`. **Checkout preserves a local modification that applies cleanly**, so the rewritten baseline followed me onto "clean main". Every subsequent reading — the 15, the "pinned to a future tree" theory, the whole comparison table — was my own probe output being mistaken for the repository's state. The numbers were internally consistent, which is exactly why it read as a finding rather than an error.

`git status` would have shown it in one line. I ran `git clean -qfd`, which removes untracked files and does nothing about a modified tracked one.

### What still stands from that comment

Only the part measured on a properly checked-out branch:

- **#2655 does close the bar.** On its own branch `moves.ts` holds 15 column guards and **zero** `triage` sites; merged with main the tree reports **backlog 724, triage 0**.
- The full battery on that merged tree is green: `check:lifecycle-columns` exit 0, `pnpm test:gate` exit 0 (726 tests), all 18 pg E2E suites 119 tests / 0 failures, `pnpm verify:fast` exit 0.

What does **not** stand: any claim that main is broken, that CI is blocked, or that #2655 must land urgently to unblock others. Merge it on its own merits and schedule.

### Correcting the correction

My previous comment closed with a note about verifying probes before trusting them, listing two failures I had caught. It then contained a third I had not caught — from the same class, in the same investigation. The specific lesson is narrower than "verify the probe": **a command that writes to the working tree contaminates every later read, including reads on a different branch.** Probes that mutate tracked files need to run in a throwaway worktree, or be reverted before the next measurement.

PR #2668, opened 2026-07-30
The ratchet's baseline could not be re-recorded once a file rose — the one state that blocks a correct conversion

gsxdsm:
Unowned (no open PR touches the census CLI — only its baseline JSON) and **live**, since #2654 gates CI on `--strict`.

## The problem

`--update-baseline` sat **behind** the rise exit, so the only supported way to re-record was unavailable in exactly the situation that needs it.

That matters because **a conversion legitimately adds a literal.** The correct shape for a caller that may have no traits is `flags ? flags.x : columnId === "legacy"`, and each one raises a file's count by one. Measured on current main: `columnRoles.ts` went **0 → 1** from precisely that shape (added by #2647, documented at the site, correct code).

So a worker doing the right thing meets a red gate whose only escape is hand-editing the JSON. That is how a ratchet becomes something people route around rather than run — and then it guards nothing. This is the same failure mode as a guard that cannot fire, arrived at from the other side.

## The change

`--update-baseline` is an explicit operator action, so it re-records **unconditionally** and prints what it accepted under `ACCEPTED RISES`. Swallowing a rise silently is the real danger; refusing to let anyone re-record is the same danger one step later, wearing a red check nobody trusts. **The rise check is unchanged** and still exits 1 without the flag.

**One writer now.** The old second `writeFileSync` behind the rise exit is deleted rather than left unreachable — two writers for one artifact is how they drift. The `!deliberateTracked && updateBaseline` special case went with it, since the unconditional block covers the legacy-shape migration too.

## Exercised end to end

On a real rise injected into `live-agent-count.ts`:

```
rise + plain --strict              exit 1   (the ratchet still bites)
rise + --strict --update-baseline  exit 0   "ACCEPTED RISES  live-agent-count.ts: 6 -> 7"
```

Four cases assert the CLI's own source, because exit codes are the contract and the pure summarizer cannot express them: the write precedes the rise check, the branches exit 0 and 1 respectively, accepted rises are **named**, and there is exactly **one** writer.

## A note on the revert proof, because it caught me twice

My first attempt to move the block back was a **no-op**: the marker I sliced on (`if (regressions.length > 0) {`) also appears *inside* the update block, so the "revert" reassembled the file unchanged and the suite stayed green. **A revert proof that does not go red can mean the guard is vacuous *or* that the revert did not land** — and the second is easy to miss when you are expecting the first. The real revert fails **2 of 27**, and the assertions now verify marker *uniqueness* before slicing on it.

## Verification

- 27/27 census suites; `--strict` exits 0; `pnpm test:gate` **71/71**; `pnpm lint` clean
- census on this tree: 748 column guards, **4 triage** (all in `moves.ts`'s flag-OFF block, deletion-scheduled with #2655)

### DIFF
--- a/packages/engine/src/__tests__/lifecycle-column-census.test.ts
+++ b/packages/engine/src/__tests__/lifecycle-column-census.test.ts
@@ -342,3 +342,70 @@ describe("the summary separates the three classes", () => {
     expect(summary.byColumnId).toEqual({ todo: 1 });
   });
 });
+
+/*
+FNXC:LifecycleColumnCensus 2026-07-31-18:40 (the ratchet's one unusable state, now fixed):
+
+`--update-baseline` MUST RUN EVEN WHEN A FILE ROSE. It could not: the rise check exited before the write,
+so the only supported way to re-record was unavailable in exactly the situation that needs it.
+
+That is a live problem, not a tidiness one, because #2654 gates CI on this AND A CONVERSION LEGITIMATELY
+ADDS A LITERAL — the correct shape for a caller that may have no traits is
+`flags ? flags.x : columnId === "legacy"`, and each one raises a file's count by one. Measured on main:
+`columnRoles.ts` went 0 -> 1 from exactly that shape. So a worker doing the right thing met a red gate
+whose only escape was hand-editing the JSON, which is how a ratchet becomes something people route around.
+
+Exercised end to end before writing this, on a real rise injected into `live-agent-count.ts`:
+  rise + plain --strict            exit 1  (unchanged — the ratchet still bites)
+  rise + --strict --update-baseline exit 0, printing "ACCEPTED RISES  live-agent-count.ts: 6 -> 7"
+
+EXIT CODES ARE THE CONTRACT and the pure summarizer cannot express them, so these assert the CLI's own
+source: which branch writes, which exits, and — the part that was actually broken — the ORDER. Marker-to-
+marker slices rather than character windows, and each marker checked for uniqueness first, because a
+repeated marker is the magic-number problem wearing a name.
+*/
+describe("the baseline can always be re-recorded", () => {
+  const cliPath = new URL("../../../../scripts/lifecycle-column-census.mjs", import.meta.url).pathname;
+
+  function cliSource(): string {
+    // eslint-disable-next-line @typescript-eslint/no-require-imports
+    return require("node:fs").readFileSync(cliPath, "utf8") as string;
+  }
+
+  function sliceBetween(cli: string, from: string, to: string): string {
+    const start = cli.indexOf(from);
+    const end = cli.indexOf(to, start + from.length);
+    expect(start).toBeGreaterThan(-1);
+    expect(end).toBeGreaterThan(start);
+    // A repeated marker would make the slice meaningless, so prove uniqueness before trusting it.
+    expect(cli.indexOf(from, start + from.length)).toBe(-1);
+    return cli.slice(start, end);
+  }
+
+  it("writes the baseline BEFORE the rise check can exit", () => {
+    const cli = cliSource();
+    const updateAt = cli.indexOf("if (updateBaseline) {");
+    const riseAt = cli.indexOf("column-guard count ROSE");
+
+    expect(updateAt).toBeGreaterThan(-1);
+    expect(riseAt).toBeGreaterThan(updateAt);
+  });
+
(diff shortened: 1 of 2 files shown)


gsxdsm:
You are right, and the distinction is sharper than "add an e2e test": my assertions checked that the write **precedes** the rise check and that each branch **contains** an exit call. None of them proved the process exits 0, rewrites the file, or prints the acceptance. A control-flow change could keep every substring in place and break all three.

Two cases now drive the real CLI as a subprocess:

| invocation | asserted |
|---|---|
| rise + `--strict` | exit **1**, `column-guard count ROSE`, baseline **unchanged** |
| rise + `--strict --update-baseline` | exit **0**, `ACCEPTED RISES`, baseline rewritten to the true count |

The first also pins something I had not covered: **a failing gate must not quietly record the thing it is failing on.** That regression would turn this ratchet into a rubber stamp, and no source-level assertion would notice.

**Testability needed a seam**, so `FUSION_CENSUS_BASELINE` overrides the baseline path — without it, an end-to-end test of `--update-baseline` would have to rewrite the repository’s own gating artifact. Env var rather than a flag on purpose: a flag on a gating command is one typo from a CI job comparing against the wrong artifact, whereas an env var has to be set deliberately by a harness. Unset in every normal invocation, so CI and operators are unaffected.

The rise is **forced by lowering one file’s allowance in a temp copy** rather than by editing source, so the test needs no scratch file in the tree and cannot leave one behind.

Two invocations, not four, at ~2s each — the green path is already covered by CI running `--strict` on every PR, and duplicating it here would be paying for coverage that exists.

Revert proof: moving `--update-baseline` back behind the rise exit now fails **3 of 29** — the two ordering assertions *and* the end-to-end update case, which is exactly why it was worth adding.

29/29 census suites, `--strict` exits 0, the repo baseline is byte-unchanged after the run, `pnpm test:gate` 71/71, lint clean.

gsxdsm:
The thread was already resolved when I got here, but the finding was correct and unaddressed, so I've fixed it rather than let a resolved marker stand in for the work.

**Every assertion in that suite read the CLI's source text** — substrings, marker ordering, `writeFileSync` call counts. The contract is three *observable* things: exit code, what lands in the baseline file, and what gets printed. Source assertions can't see any of them; move the exit, reorder the branches, or return before the write, and they all stay green.

Added a `FUSION_CENSUS_BASELINE_PATH` override so the CLI can run against a throwaway baseline instead of rewriting the repo's real one. That missing seam is *why* the suite went to source in the first place. Production never sets it.

Two cases, mutation-verified against control flow rather than text:

| mutation | failures |
|---|---|
| disable the `--update-baseline` branch | 3 |
| flip the rise path from `exit 1` to `exit 0` | 2 |

The source-text tests catch neither — the strings survive both.

**One near-miss worth recording.** My first run had no `cwd`, so the CLI globbed from the package dir, found nothing, and exited on "file list is EMPTY". The rise test **passed** — it expected exit 1 and got exit 1, for an entirely unrelated reason. Only the update-baseline case failed and led me to look. A test asserting a nonzero exit passes on *any* nonzero exit, so the fixture has to be proven before the assertion means anything.

Verified: 29 tests green, engine tsc clean, lint clean, merge gate green (487 + 158 + 10).

gsxdsm:
Complementary to #2673, not overlapping — flagging so neither gets closed as a duplicate of the other.

- **This PR** fixes the *tooling*: `--update-baseline` cannot re-record once any file has risen, because the regression check `exit(1)`s before the write. I hit exactly that when bootstrapping a new baseline category in #2650 and had to seed the keys by hand; good to see it closed properly rather than worked around again.
- **#2673** fixes the *data*: clean `origin/main` currently fails the required check with `packages/engine/src/executor.ts: allows 87, tree has 85` — #2568 converted two guards without re-recording in the same commit. That is a stale allowance, not a rise, so it is untouched by this PR's change.

No file overlap (this one: `scripts/lifecycle-column-census.mjs` + its test; #2673: the baseline JSON only), so they land in either order.

One observation that may be in scope for you, since you are already in this file. The recurring failure has an awkward shape: **a converting PR is green against the old baseline, and the staleness only materialises on `main` after it merges.** So the ratchet cannot catch it at review time, and it has now produced two post-merge reds in a day (#2568 here, and an earlier one). A diff-aware rule — fail a PR whose diff lowers a count without touching the baseline — would catch it on the PR instead of on everyone else's CI. I have not built it; you own this file and I did not want to fork the logic while you are mid-change.

PR #2669, opened 2026-07-30
fix(core): wedge notifications could never be resolved on PostgreSQL (42P18)

gsxdsm:
Not a U12 change — found while **attributing** the pre-existing live-PG failures during U12's closing verification, and it turned out to be a product bug rather than a stale test.

## The defect

`resolveWedgeNotification` builds its UPDATE with:

```ts
jsonb_build_object('status', 'resolved', 'transitionedAt', ${transitionedAt})
```

`jsonb_build_object` is variadic `"any"`, so there is no signature for PostgreSQL to resolve the bind parameter against. It rejects the statement at **parse time**:

```
42P18: could not determine data type of parameter $1
```

Parse-time is the important part: this failed on **every call**, not on unusual data. Wedge notifications could not be resolved at all in PostgreSQL mode.

Casting the parameter to `::text` fixes it.

## Evidence

- `store-wedge-resolution.pg.test.ts` goes **0/7 → 7/7**. That suite has been red on `main`.
- **Causally verified, not assumed:** removing the cast reproduces `42P18` exactly. The fix is the cast, not something incidental to the edit.
- Checked the rest of `packages/core` for the same shape — this is the only `jsonb_build_object` call site, so there is no second instance hiding.

## Why it survived

The failure is in a live-PG suite that was already red, so it read as part of the ambient noise. I only found it because the closing verification required me to attribute each failing suite to a cause rather than count them — and "these 4 fail on main too" is an attribution of *whose*, not of *what*.

Worth flagging for whoever owns the remaining three (`agent-logs-and-monitor`, `central-archive-secrets`, `workflow-settings-project-identity`): the same reasoning applies. A suite failing on main is not evidence that the code is fine.

## Verification

`pnpm test:gate` green (10 / 158 / 487 / 71). `pnpm check:lifecycle-columns` exits 0. `tsc -p packages/core/tsconfig.json` clean. `pnpm lint` clean.

Independent of #2655; either order merges.

### DIFF
--- a/.changeset/fix-wedge-notification-pg.md
+++ b/.changeset/fix-wedge-notification-pg.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Wedge notifications can be resolved again on PostgreSQL projects.
+category: fix
+dev: jsonb_build_object is variadic "any", so PostgreSQL could not infer the type of the bare `transitionedAt` bind parameter and rejected the resolve-wedge UPDATE at parse time with 42P18. Casting the parameter to ::text fixes it; the failure was total, not data-dependent.
--- a/packages/core/src/task-store/async-persistence.ts
+++ b/packages/core/src/task-store/async-persistence.ts
@@ -211,7 +211,20 @@ export async function resolveActiveTaskWedgeEpisodeRow(
   const rows = await layer.db
     .update(schema.project.tasks)
     .set({
-      wedgeNotification: sql`(${schema.project.tasks.wedgeNotification}::jsonb || jsonb_build_object('status', 'resolved', 'transitionedAt', ${transitionedAt}))::text`,
+      /*
+      FNXC:WedgeNotification 2026-07-31-12:00:
+      `${transitionedAt}::text` — the CAST is load-bearing, not decoration.
+
+      PostgreSQL cannot infer the type of a bare bind parameter used as a `jsonb_build_object` value:
+      the function is variadic `"any"`, so there is no signature to resolve $1 against and the planner
+      rejects the statement outright with `42P18: could not determine data type of parameter $1`.
+      That is a PARSE-time failure, so it fires on every call rather than on unusual data — wedge
+      notifications could never be resolved in PostgreSQL mode at all.
+
+      Surfaced by `store-wedge-resolution.pg.test.ts`, which has been failing on main; the failure is
+      the product query, not the test.
+      */
+      wedgeNotification: sql`(${schema.project.tasks.wedgeNotification}::jsonb || jsonb_build_object('status', 'resolved', 'transitionedAt', ${transitionedAt}::text))::text`,
       updatedAt: transitionedAt,
     })
     .where(and(
--- a/scripts/lib/lifecycle-column-census-baseline.json
+++ b/scripts/lib/lifecycle-column-census-baseline.json
@@ -1,22 +1,22 @@
 {
   "generatedFrom": "node scripts/lifecycle-column-census.mjs --strict --update-baseline",
   "totals": {
-    "column": 748,
+    "column": 746,
     "role": 5,
     "status": 186,
     "deliberate": 17
   },
   "byColumnId": {
-    "done": 200,
+    "done": 199,
     "in-progress": 144,
     "in-review": 205,
-    "archived": 148,
+    "archived": 147,
     "todo": 47,
     "triage": 4
   },
   "byFile": {
     "packages/engine/src/self-healing.ts": 110,
-    "packages/engine/src/executor.ts": 87,
+    "packages/engine/src/executor.ts": 85,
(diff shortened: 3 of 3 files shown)


gsxdsm:
## Triage of the other three long-red PG suites

I said in the description that "a suite failing on main is not evidence that the code is fine", so I applied that to the remaining three rather than leaving it as an assertion. **I am not fixing these here** — two need their subsystem owner's judgement, and guessing at them from outside would be the unilateral move I keep flagging. Diagnoses with evidence:

### 1. `workflow-settings-project-identity.pg.test.ts` — 2 failures. **The code contradicts its own documentation.**

`expected '/tmp/root' to be 'legacy_identity_id'`

`getWorkflowSettingsProjectIdImpl` (`branch-and-pr-entities.ts:614`) documents a three-step order:

```
(a) store.asyncLayer?.projectId      — central-registry id (PG)
(b) store.db.getProjectIdentity()?.id — legacy SQLite identity
(c) store.rootDir                     — last-resort legacy key
```

The code does (a), then returns `rootDir` unconditionally. **Step (b) is gone**, removed by `FNXC:SqliteDualPathCleanup 2026-07-26-14:15`. A comment three lines above the return still says *"Only the true legacy (non-backend) path consults the SQLite identity"* — which is now false for every caller.

So the test encodes the documented contract and the code no longer implements it. **The owner's call is which one is right:** if SQLite mode is genuinely dead, the test is stale and the doc block needs correcting; if a legacy store can still exist, workflow settings mis-key by `rootDir` for it, which is the exact "settings looked reset" bug the same doc block says it was written to fix.

I did not guess, because the two answers have opposite consequences and the stale comments make the intent unreadable from outside.

### 2. `agent-logs-and-monitor.pg.test.ts` — 1 failure. `expected +0 to be 2` on an aggregation. Not diagnosed further; the shape (a count returning zero) suggests either a query that no longer matches or a fixture that no longer seeds, and distinguishing those needs the owner's context.

### 3. `central-archive-secrets.test.ts` — 1 failure. `expected "warn" to be called with arguments: [ …(2) ]` — an assertion on log output. Lowest severity of the four: it pins a diagnostic message, not behaviour.

### Why this matters beyond these four

All four sat behind "pre-existing, fails on main too" — which answers *whose* problem it is, not *what* is wrong. One of them (#2669, this PR) was a total runtime failure of a product feature. The other three have not been ruled out, and at least one has code and documentation actively disagreeing about a bug the documentation claims to have fixed.

gsxdsm:
## Third of the four: `agent-logs-and-monitor` — a real Command Center defect from the U11 merge. **Needs a product decision, so I am not guessing at it.**

`expected +0 to be 2` at `agent-logs-and-monitor.pg.test.ts:224` — the SDLC funnel counts **zero** entries into the `todo` stage after two tasks are moved there.

### Cause, verified empirically

`activity-analytics.ts` maps workflow **traits** to funnel stages (correctly — by trait, never by column name). But the mapping predates the U11 column merge:

```
intake            -> "triage"
reset-on-entry    -> "todo"
hold              -> (unmapped)
```

Measured on the shipped default lineage:

```
default `todo` column traits : ["intake","hold","reset-on-entry"]
stageForTraits(["intake","hold"]) = "triage"
stageForTraits(["hold"])          = "other"
```

`stageForTraits` picks the **earliest stage in flow order**, so `intake` (stage 0) beats `reset-on-entry` (stage 1). After U11 collapsed Triage and Todo into one Planning column carrying both traits, **every card entering Planning is counted in the `triage` stage** — a stage whose column no longer exists — and the `todo` stage is permanently empty on every default board.

This is user-visible in Command Center analytics, not just a test failure.

### Why I am not shipping a fix

The funnel has five stages; the shipped lineage has **one** pre-implementation column. Reasonable people would pick differently:

- **(a)** merge the `triage` and `todo` funnel stages to match the lineage — correct, but `SDLC_STAGES` is a published shape the dashboard renders, so it is a product/analytics decision;
- **(b)** map the merged Planning column to the `todo` stage and leave `triage` for legacy boards — smaller, but then the stage named `triage` silently means "pre-U11 boards only";
- **(c)** keep both and split by a secondary signal — more faithful, more machinery.

Same shape as the coding-ideas question you settled by decision rather than by code reading. I have the evidence; the call is not mine to make silently.

**One part is unambiguous regardless of that choice:** `hold` is absent from `TRAIT_TO_STAGE` entirely, so a hold-only column falls to `other`. That is wrong under every reading above, and it is why a renamed board's wait-for-capacity lane disappears from the funnel.

### Running total on the four "pre-existing" PG failures

- ✅ `store-wedge-resolution` — product bug, total runtime failure in PG (#2669)
- ✅ `workflow-settings-project-identity` — code contradicted its own docs (#2671)
- 🔶 `agent-logs-and-monitor` — **this**: real defect, blocked on a product decision
- ⬜ `central-archive-secrets` — a `warn`-argument assertion, still unruled-out

Three of four were real. "Pre-existing" was doing a lot of work.

PR #2670, opened 2026-07-30
The archived half of the terminal pair, end to end (#2568's fix landed first and is better)

gsxdsm:
**Live on `main`.** `parkCompletedBlockedTask` opens with *"is this card already finished?"* and answered it with:

```ts
if (task.column === "done" || task.column === "archived") return false;
```

On a renamed board neither matches, so the guard was **inert** — and inert here is not a missed rescue, it is active damage: the very next block rebounds the card to its planning lane. **A completed card sitting in a renamed complete or archived column was moved backwards out of it.**

## Why this survived two reviews

#2644 (mine) converted the **rebound** half to resolve its target by role. This **terminal** half stayed a literal. A role-resolved rebound behind a name-matched guard means the renamed board takes the rebound and never the guard — the same half-conversion shape as the evacuation branch, except the two halves were owned by different changes, so neither review saw both.

Worth keeping as a review heuristic for the remaining conversions: **when a converted site sits next to an unconverted one, the conversion can make the neighbour worse — and a diff showing only one of them looks complete.**

## Relationship to #2568

The fix exists there, stranded four deep in a stack whose bottom (#2544) has not merged, so nothing in that chain has reached `main`. This re-lands **only** the guard, directly against `main`. I have noted it on #2568 so its author can drop that hunk rather than resolve it twice; I took no other part of that PR (its extraction and the terminal-pair ownership change are still theirs).

## Deliberate choices

- **Fail-soft to `["done", "archived"]`** — an unresolvable workflow behaves exactly as the literal pair did.
- **An unclassifiable column is NOT terminal.** Being unable to prove a card is finished must not be the same as proving it is; that is what keeps a stranded card moving.

## Revert proof

Restoring the literal pair fails **2 of 5** — the renamed complete and archived cases. Three paired cases pass under the revert, so neither "never park" nor "always park" can pass for resolving the lanes: a genuinely mid-pipeline card is still parked, the legacy pair still answers when no workflow resolves, and an unclassified column still gets the park.

## Verification

- 5/5 new; 22/22 with `executor-rebound-already-there` and `executor-task-done-blocked`
- census: `executor.ts` **87 → 85** column guards; baseline re-recorded in the same commit

## Summary by CodeRabbit

* **Reliability**
  * Improved handling of completed/blocked tasks when workflow column naming changes, ensuring tasks aren’t parked or moved incorrectly in archived or non-terminal lanes.
* **Tests**
  * Added Vitest coverage to verify terminal-lane decisions, including cases where workflow resolution occurs mid-operation and lane placement changes during the wait.
* **Maintenance**
  * Updated lifecycle column census baseline numbers to match current workflow usage.

### DIFF
--- a/packages/engine/src/__tests__/completed-blocked-terminal-guard.test.ts
+++ b/packages/engine/src/__tests__/completed-blocked-terminal-guard.test.ts
@@ -1,5 +1,5 @@
 /*
-FNXC:WorkflowLifecycleColumns 2026-07-31-20:25 (live on main):
+FNXC:WorkflowLifecycleColumns 2026-07-30-08:25 (live on main):
 
 THE INVARIANT: "is this card already finished?" is answered from the task's own workflow, so a completed
 card is never rebounded out of its complete or archived lane.
@@ -99,7 +99,7 @@ function harness(ir: WorkflowIr | undefined, column: string, gate?: Promise<void
 
 describe("the terminal guard covers the ARCHIVED role and refuses to guess", () => {
   /*
-  FNXC:WorkflowLifecycleColumns 2026-08-01-00:20 (reduced after #2568's fix landed on main):
+  FNXC:WorkflowLifecycleColumns 2026-07-30-08:40 (reduced after #2568's fix landed on main):
 
   MAIN'S FIX IS BETTER THAN MINE AND SHIPPED FIRST, so the implementation here is main's: it routes through
   core's shared `resolveTerminalColumns` (which owns the per-role fallback) and re-reads the task after the
--- a/packages/engine/src/__tests__/completed-blocked-terminal-guard.test.ts
+++ b/packages/engine/src/__tests__/completed-blocked-terminal-guard.test.ts
@@ -53,7 +53,7 @@ function completedTaskIn(column: string) {
   };
 }
 
-function harness(ir: WorkflowIr | undefined, column: string) {
+function harness(ir: WorkflowIr | undefined, column: string, gate?: Promise<void>) {
   const store = createMockStore();
   let task: Record<string, unknown> = completedTaskIn(column);
   const moves: Array<[string, string]> = [];
@@ -61,7 +61,17 @@ function harness(ir: WorkflowIr | undefined, column: string) {
   const widened = store as unknown as Record<string, unknown>;
   widened.getTaskWorkflowSelection = () => (ir ? selection : undefined);
   widened.getTaskWorkflowSelectionAsync = async () => (ir ? selection : undefined);
-  widened.getWorkflowDefinition = async () => (ir ? { id: "wf-renamed", ir } : undefined);
+  /*
+  FNXC:WorkflowLifecycleColumns 2026-07-30-09:15 (#2670 review):
+  `gate` holds workflow resolution open so a test can change the STORED card while the
+  resolution is still pending. Without it every await in this file settles immediately,
+  and the post-await re-read of the task is unreachable — a test can only observe the
+  column it started with, which is what let the re-read look covered when it was not.
+  */
+  widened.getWorkflowDefinition = async () => {
+    if (gate) await gate;
+    return ir ? { id: "wf-renamed", ir } : undefined;
+  };
 
   store.getTask.mockImplementation(async () => ({ ...task }));
   store.updateTask.mockImplementation(async (_id: string, updates: Record<string, unknown>) => {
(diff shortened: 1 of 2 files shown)


gsxdsm:
Real, and fixed — the re-read now covers the pause check, the terminal check and the move guard.

Your scenario is precisely the one that turned out to be **untested**, and finding that took fixing three broken instruments in a row. Worth spelling out, because each looked like a passing verification:

1. A shell loop split its mutation patterns on `|` — and the pause condition contains `||`. The search string was silently truncated, nothing was replaced, and all three "mutations" reported green. An unmutated build passing is not evidence of anything.
2. A `grep -c 'Failed Tests'` counter didn't match the dot reporter's output, so a genuinely failing run read as passing.
3. Only with both fixed did the real gap show: the **terminal** check was uncovered. The pause and move-guard tests both stayed green with it reverted, because neither drives a card into a terminal lane mid-await.

So the new test is exactly your scenario — a card that reaches `done` while the workflow lookup is in flight — and it existed nowhere before. Reverting the terminal check to the stale snapshot now fails it and nothing else.

**Overlap worth flagging for merge order:** #2568 fixes this same guard and carries the pause and move-guard tests. It's stranded behind an unmerged stack, so I brought the tests here rather than have them wait. Whichever PR lands second will conflict on `parkCompletedBlockedTask`; the correct resolution is to keep the re-read and all three tests, not to take either side whole.

Verified: 20 tests green, engine tsc clean, lint clean, merge gate green (487 + 158 + 10).

gsxdsm:
## Reduced: #2568’s fix reached main while this was open, and it is better than mine

Rebased and found main already has the conversion — routed through core’s shared `resolveTerminalColumns` (which owns the per-role fallback) **and** re-reading the task after the await. That second part my version did not do, so an operator pause landing during the IR resolution would have been overwritten. **Main’s implementation is kept verbatim; mine is dropped**, along with my changeset and baseline edit (main’s PR ships the user-facing note, and there is no census delta left to record).

What survives is only the coverage main’s fixture does not reach:

1. **The ARCHIVED role, end to end.** Main’s renamed workflow declares `complete` and no `archived` column, and `resolve-terminal-columns.test.ts` exercises the resolver in isolation. The two roles resolve independently and can fail independently, so covering only `complete` through the executor path leaves half the pair unexercised.
2. **An unclassifiable column is not terminal.** Being unable to prove a card is finished must not be the same as proving it is — that is what keeps a card stranded by a lineage change moving, instead of parked forever in a column no sweep owns.

Revert proof: restoring the literal pair fails **1 of 2** (the renamed archived column). The unclassifiable case passes under the revert, which is correct — it is a property of the fallback, not of the conversion, and saying so is more useful than padding the count.

If you would rather fold these two cases into `execute-requeue-loop-guard.test.ts` next to main’s renamed case and close this, that is fine by me — the point is the archived half being covered somewhere, not where it lives.

2/2, engine typecheck clean, lint clean, `--strict` exits 0.

gsxdsm:
Both findings were correct, and both are now fixed. I verified each by mutation rather than by inspection.

**1. Exercise the re-read after workflow resolution — real.** You were right that nothing here could catch its removal. Every await in the file settled immediately and every test started in the column it asserted on, so no test could distinguish the post-await re-read from the caller's stale copy.

The harness now takes an optional gate that holds `getWorkflowDefinition` open. The new test starts the card in `building` (mid-pipeline, parkable on its face), moves the **stored** card into `attic` while resolution is pending, then releases. Mutation: replacing the re-read with the caller's copy fails 1 of 3.

This is worth more than a coverage tick — resolving a workflow is an await, and a card can reach a terminal lane inside that window via a concurrent archive or the merge path completing. Judging the card as the caller found it is the same stale-read bug the guard exists to prevent, one level up.

**2. Assert the actual parking move — real, and stronger than stated.** I checked what the file caught without it: I suppressed the move entirely and **all three tests still passed**. So the move was not weakly covered here, it was uncovered. The assertion now pins the card landing in the role-resolved hold column (`queued`, not `todo`), which also catches a move to the wrong lane. Mutation: suppressing the move fails 1 of 3.

---

**One thing I nearly got wrong, since it affects what is in this branch.** My local line for this PR was stale: it still carried the production fix and the changeset, both of which had already merged via #2568. Had I pushed it, I would have resurrected merged code and added a duplicate changeset. The branch here is the correct reduced line — tests only — with my two review commits on top. Checking before pushing is the only reason that did not land.

Verified: 3 tests green, engine tsc clean, lint clean, merge gate green (487 + 158 + 10 + 71).

Also corrected two future-dated FNXC stamps in this file. Separately, main carries **~70** more; that belongs in its own mechanical sweep, not here.

PR #2672, opened 2026-07-30
A rate-limited provider kept getting hammered — the executor and merger lane checks were still literals

gsxdsm:
`taskUsesProvider` resolves a task's **active lane** to decide which providers it is running on. The **planner** half was converted to traits — its own note in that function describes this exact failure and says it was fixed — and the **executor** and **merger** halves were left as `task.column === "in-progress"` / `=== "in-review"`.

So on a renamed board an actively-executing card resolved **no providers**, and a provider rate limit never paused it: the engine kept sending work to the limited provider with that card.

## Measured

Renamed board (`building` = wip, `checking` = review), limit triggered by a peer:

| lane | before | after |
|---|---|---|
| executor | `["FN-TRIGGER"]` | `["FN-TRIGGER","FN-PEER"]` |
| merger | `["FN-TRIGGER"]` | `["FN-TRIGGER","FN-PEER"]` |

The trigger was paused either way — but only through the *always-include-the-trigger* fallback, and **that is what made the hole quiet**: one task always gets paused, so the behaviour looks like it works.

Resolved from the **same per-workflow IR cache** the planner lane already uses, so this adds no reads. Both halves fail soft to the legacy literal, so an unresolvable workflow behaves exactly as before.

## How it was found — the part worth keeping

A scan for *"legacy literal within a few lines of a role-resolved call"* flagged this file. That is the same heuristic that produced #2670, and it is now 2 for 3.

The first thing I suspected here was the `done`/`archived` terminal filter. I wrote that fix, and **its revert stayed green.** That is not a dead end — chasing *why* it would not go red showed the lane check already excludes finished cards, so the terminal literal there is genuinely redundant, and the real defect was one line over in the lane check itself. **A revert that stays green is information: either the guard is vacuous, or you are looking at the wrong line.**

I dropped the unprovable change and kept the provable one. The `done`/`archived` filter is deliberately unchanged, with that reason recorded in the test.

## Revert proofs, isolated

- wip lane back to the literal → **1 of 55 fails** (the executing peer)
- review lane back to the literal → **1 of 55 fails** (the merger peer)

Three paired cases pass under both, so "pause everything with that provider" cannot pass for "resolve the lane": a card parked in the renamed planning lane is **not** paused on an executor limit, a wip card is **not** paused on a merger limit, and the legacy vocabulary still pauses peers when no workflow resolves.

## Verification

- 55/55 `usage-limit-detector`; `pnpm test:gate` **71/71**; engine typecheck clean; `pnpm lint` clean
- census unchanged at 748 — the fail-soft literals remain by design, which is why the count is not the measure of this fix

### DIFF
--- a/.changeset/usage-limit-active-lanes.md
+++ b/.changeset/usage-limit-active-lanes.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: A provider rate limit now pauses every task actually running on that provider, including on renamed boards.
+category: fix
+dev: `taskUsesProvider`'s executor and merger lane checks resolve the workflow's wip/review columns (from the same per-workflow IR cache the planner lane already uses) instead of comparing against `"in-progress"`/`"in-review"`; both fail soft to the legacy literal.
--- a/packages/engine/src/__tests__/usage-limit-detector.test.ts
+++ b/packages/engine/src/__tests__/usage-limit-detector.test.ts
@@ -425,3 +425,93 @@ describe("a provider rate limit pauses every card actually running on that provi
     expect(await pausedIds(store)).toContain("FN-PEER");
   });
 });
+
+/*
+FNXC:WorkflowLifecycleColumns 2026-07-31-23:35 (PR #2672 review, greptile P1 + P2):
+
+TWO PROPERTIES MY FIRST VERSION GOT WRONG.
+
+1. THE FALLBACK IS PER ROLE, NOT PER OBJECT. I keyed it on whether `activeLanes` existed at all, so a
+   workflow that RESOLVES but declares no wip (or no review) column suppressed the legacy id and resolved no
+   providers — reintroducing the exact bug this change fixes, for the partial-vocabulary case. A missing ROLE
+   is not the same fact as a missing WORKFLOW; only the second means "no basis".
+
+2. RESOLVE ONLY THE CANDIDATES. Resolving every task before filtering made a rate limit on a 400-card board
+   pay 400 resolutions to pause a handful, including terminal and paused cards that can never be affected.
+*/
+describe("lane resolution degrades per role and costs only what it must", () => {
+  /** Declares a wip lane but NO review lane — the partial vocabulary that broke the first version. */
+  const NO_REVIEW_IR = {
+    version: "v2", id: "wf-partial", name: "partial", nodes: [], edges: [],
+    columns: [
+      { id: "queued", name: "Queued", traits: [{ trait: "intake" }, { trait: "hold", config: { release: "capacity" } }] },
+      { id: "in-progress", name: "In progress", traits: [{ trait: "wip", config: { limitSetting: "maxConcurrent" } }] },
+    ],
+  };
+
+  const card = (id: string, column: string) => ({
+    id, column, dependencies: [], steps: [], currentStep: 0, log: [],
+    modelProvider: "openai-codex", modelId: "gpt-5",
+  });
+
+  function storeFor(tasks: any[], ir: unknown, workflowId = "wf-partial") {
+    const store = createMockStore(tasks);
+    const selection = { workflowId, stepIds: [] };
+    const reads = { definition: 0 };
+    /*
+    Records WHICH task ids were asked about, not how many reads happened. A read count is
+    concurrency-dependent here — the shared IR cache fills after an await, so same-workflow tasks race and the
+    number lands anywhere between 1 and N. Asking "was this task resolved at all?" is exact and race-free,
+    which is what a filtering assertion actually needs.
+    */
+    const asked: string[] = [];
+    store.getTaskWorkflowSelectionAsync = async (id: string) => {
+      asked.push(id);
+      return selection;
+    };
+    store.getTaskWorkflowSelection = (id: string) => {
+      asked.push(id);
+      return selection;
+    };
+    store.getWorkflowDefinition = async () => {
+      reads.definition += 1;
+      return ir ? { id: workflowId, ir } : undefined;
+    };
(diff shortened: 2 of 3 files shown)


gsxdsm:
Both fixed — and the P1 is the more interesting one, because my own fix reintroduced the bug it was closing.

**P1: the fallback was per OBJECT, not per role.** I keyed it on whether `activeLanes` existed at all, so a workflow that *resolves* but declares no wip (or no review) column suppressed the legacy id and resolved no providers — exactly the hole this PR closes, one case over. **A missing role is not the same fact as a missing workflow**; only the second means "no basis". Per-role fallback can only *add* pauses, for a card in a legacy-named column on a board that declares no such role, which is the safe direction the function’s own note states: the cost of a wrong include is pausing work that was fine.

**P2: resolve only the candidates.** Right — the first version resolved every task before filtering, so a rate limit on a 400-card board paid 400 resolutions to pause a handful, including terminal and paused cards that can never be affected. Cheap predicates run first now.

On the cache half of your note, stated rather than silently accepted: concurrent misses on the same workflow can still both read, because `irCache` fills after the await. That property is shared with the planner-lane resolution directly above it, and serialising would trade a bounded duplicate read for latency on the **pause** — the operation an operator is waiting on. Left as-is, named at the site.

**And a flaky assertion of my own, caught while proving P2.** My first version of that test asserted a *read count* (`<= 1`). That is concurrency-dependent here: the shared cache fills after an await, so same-workflow tasks race and the count lands anywhere between 1 and N. It passed — and it would have failed intermittently in CI, which is precisely the quarantine ratchet’s input. Replaced with an exact, race-free question: **which task ids were asked about**. The three that can never be affected are never resolved at all.

Revert proofs, isolated:
- per-object fallback restored → **1 of 57 fails** (the no-review-lane workflow)
- resolve every task → **1 of 57 fails** (`FN-DONE` asked about)

57/57 usage-limit-detector, `pnpm test:gate` 71/71, engine typecheck clean, lint clean.

gsxdsm:
Both findings were already implemented on the branch — but **one of them had no test**, and I only found that by mutation.

**P1 (per-role fallback):** reverting it to the per-object form left all 57 existing cases green. The fix was real and unproven. Added a test that pins the distinction it turns on: a workflow that *resolves* but declares no `wip` column previously suppressed the legacy id, so a card in `in-progress` resolved no providers and kept running on the rate-limited one. **A missing role is not the same fact as a missing workflow** — only the second means "no basis to judge". Reverting now fails the new case and nothing else.

**P2 (duplicate reads):** no change needed, and I checked rather than took the comment's word. Candidates are filtered before resolution — paused, `done` and `archived` are excluded first — so cost tracks plausible tasks rather than board size. The residual concurrent-miss duplicate read is named in the source as known-and-unchanged with its reason: serialising it would trade a bounded duplicate read for latency on the pause, which is the operation an operator is actually waiting on. I agree with that trade; flagging that I agree rather than leaving silence to imply it.

Verified: 58 tests green, engine tsc clean, lint clean, merge gate green (487 + 158 + 10).

gsxdsm:
## Null result worth recording: the heuristic’s false-positive rate is high in the dashboard, for a good reason

Ran the same "legacy literal within a few lines of a role-resolved call" scan across `packages/dashboard/app` and `packages/dashboard/src` — the one that produced #2670 and this PR from the engine. **19 hits, zero defects.**

Every one is the correct shape:

```ts
if (flags) return flags.hold === true || flags.countsTowardWip === true;
return column === "todo" || column === "in-progress";   // documented fallback
```

The literal is in the **fallback branch of a flags ternary**, reachable only when the caller has no resolved column metadata — which is exactly the pattern I defended in `live-agent-count.ts` and the one U12 has applied consistently across `Column.tsx`, `ListView.tsx`, `TaskContextMenu.tsx`, `TaskDetailModal.tsx` and `columnRoles.ts`. `TaskContextMenu.tsx:293`’s `VALID_TRANSITIONS` branch is gated the same way and its comment names the retirement path (allowed targets on the board-workflows payload).

**So the heuristic is 2 for 3 in the engine and 0 for 19 in the dashboard**, and the reason is structural rather than luck: the *correct* pattern looks exactly like the defect pattern to a proximity scan. What separates them is whether the literal sits in a **fallback branch** (correct) or in a **separate statement beside resolved data** (suspect — both engine defects were the latter). Anyone running this scan during the batch phase should read that distinction first, or they will file nineteen non-issues.

The census already has the machinery to tell them apart — `properties` classifies query and definition siblings — so a `besideTraitCheck`-style flag on findings would let it report "fallback literal" as its own class rather than as backlog. Noting it rather than adding it: the census is main’s now, and this is a report-quality improvement, not a gate change.

I am stopping the scan here. Yield has dropped to zero, and continuing would be make-work rather than work.

gsxdsm:
Rebased onto `origin/main` @ bc782d8d92 — that is what the red Lint check was.

The failure was the **lifecycle-column ratchet**, not this diff: the branch forked before the recent merge wave, so its tree carried the old per-file counts while main's baseline had moved. Measured after the rebase, no baseline edit needed:

```
COLUMN guards: 722 · --strict: every file matches its baseline exactly
usage-limit-detector.ts: 4 (unchanged — the remaining four are `?? "in-progress"`-shaped trait fallbacks)
usage-limit-detector.test.ts 58/58 · pnpm lint clean
```

Worth naming since three of my PRs hit it this week: a branch that touches nothing the census counts can still fail the ratchet purely by being behind main, and the only fix is a rebase. That is the permanently-red-gate dynamic #2679 addresses — with the auto-tighten in place this run would have gone green and printed what it lowered, and its new base-ref check keeps regrowth caught while the committed baseline is stale.

PR #2673, opened 2026-07-30
fix(census): main is RED on the required check — re-record the baseline after #2568 (87 → 85)

gsxdsm:
`pnpm check:lifecycle-columns` is a **required** PR check. Clean `origin/main` (632d10a9b4) fails it, so **every open PR is red and none of them caused it**.

```
"A stale allowance is a hole"
```

## Cause

#2568 converted two guards in `executor.ts` (87 → 85) without re-recording the baseline in the same commit.

This is the ratchet **working**, not misfiring: a stale allowance means those two guards could be reintroduced later while the check stayed green — the high-water-mark failure #2661 closed. The tool's own error message prescribes the remedy.

## Fix

```
```

Lowers exactly two numbers, downward only, baseline-only diff, no source touched:

```
totals.column                             748 -> 746
byFile[packages/engine/src/executor.ts]    87 ->  85
```

## Verified two ways, because I got this wrong once

I claimed main was red earlier today and **was mistaken** — I had contaminated my own working tree with an earlier `--update-baseline` and then read it back as the repository's state, on a different branch. That produced a self-consistent, entirely fictitious diagnosis which I posted to two PRs and retracted.

So this one is established twice, independently:

1. **Working-tree reading**, with `git status --porcelain` asserted empty immediately *before* and *after* each measurement — a dirty tree refuses to produce a number rather than producing a wrong one.
2. **Committed-blob reading** via `git show origin/main:…`, which cannot observe a working tree at all: allowance **87**, actual count **85**.

Both agree, and they agree on the same file and the same delta.

## Verification

- `check:lifecycle-columns` exit 0
- census suites 51 green
- `pnpm test:gate` exit 0

## Note for whoever lands conversions next

This is the second time a merge has lowered a count without re-recording, and the check only became blocking recently — so it will keep happening until it is habit. The rule is one line: **any PR that converts a guard re-records the baseline in the same commit.** `--strict --update-baseline` does it, and it refuses to raise anything, so it cannot be used to hide a regression.

### DIFF
--- a/scripts/lib/lifecycle-column-census-baseline.json
+++ b/scripts/lib/lifecycle-column-census-baseline.json
@@ -1,22 +1,22 @@
 {
   "generatedFrom": "node scripts/lifecycle-column-census.mjs --strict --update-baseline",
   "totals": {
-    "column": 748,
+    "column": 746,
     "role": 5,
     "status": 186,
     "deliberate": 17
   },
   "byColumnId": {
-    "done": 200,
+    "done": 199,
     "in-progress": 144,
     "in-review": 205,
-    "archived": 148,
+    "archived": 147,
     "todo": 47,
     "triage": 4
   },
   "byFile": {
     "packages/engine/src/self-healing.ts": 110,
-    "packages/engine/src/executor.ts": 87,
+    "packages/engine/src/executor.ts": 85,
     "packages/dashboard/app/components/TaskCard.tsx": 42,
     "packages/core/src/task-store/moves.ts": 39,
     "packages/dashboard/app/components/TaskDetailModal.tsx": 30,


gsxdsm:
Coordination note: **#2668 does not supersede this PR, and this does not supersede it.**

- #2668 fixes the *tooling* — `--update-baseline` cannot re-record once any file has risen (the regression check exits before the write).
- This PR fixes the *data* — main's current red is a **stale allowance** (`executor.ts: allows 87, tree has 85`), which is the opposite condition and untouched by that change.

No file overlap: #2668 touches `scripts/lifecycle-column-census.mjs` and its test; this touches only `scripts/lib/lifecycle-column-census-baseline.json`. Either order merges.

Re-verified after rebasing onto `13bf7e001d`: main still fails with the identical `87 -> 85`, this branch passes `check:lifecycle-columns` (exit 0) and `pnpm test:gate` (exit 0), and the rebase required no further baseline change — the fix has not drifted.

gsxdsm:
Superseded — closing. `main` re-recorded the baseline itself at `a099813e94`, and the values are exactly what this PR set:

```
main baseline: executor.ts 85, totals.column 746
this branch's diff vs main: empty
```

`pnpm check:lifecycle-columns` now exits 0 on clean `origin/main` (verified with `git status --porcelain` asserted empty before and after the reading). The red this PR existed to clear is gone, so there is nothing left for it to do.

The underlying pattern is still worth the durable fix, and it is #2668's territory rather than mine: a converting PR is green against the old baseline, so the stale allowance only materialises on `main` after the merge — the ratchet cannot catch it at review time. That produced three consecutive red commits on main today. A diff-aware rule (fail a PR whose diff lowers a count without touching the baseline) would catch it on the PR instead of on everyone else's CI. Raised there.

PR #2674, opened 2026-07-30
fix(core): a hold-only column vanished from the SDLC funnel (+ un-red the census on main)

gsxdsm:
Two things, both small, one urgent.

## 1. Hold-only columns disappeared from the funnel

`hold` was absent from `TRAIT_TO_STAGE`, so a column whose only pre-implementation trait is `hold` — a renamed board's wait-for-capacity lane — resolved to `OTHER` and vanished from the SDLC funnel.

Measured before the fix:

```
stageForTraits(["hold"]) === "other"
```

The default lineage hid it: its Planning column also carries `intake` and `reset-on-entry`, so it always matched something. Only a board that names its wait lane separately was affected — **exactly the custom shape this trait mapping exists to support**.

Revert check: removing the entry gives `expected 'other' to be 'todo'`.

## What I deliberately did NOT fix, and why

The merged default Planning column carries `["intake","hold","reset-on-entry"]`, and `stageForTraits` prefers the earliest stage in flow order — so `intake` wins and it still resolves to `triage`. The `todo` stage therefore stays empty on every default board since U11, and the funnel shows a **phantom 100% drop between Triage and Todo**.

That is a real defect. It is also not a reversible call: changing which stage Planning reports would retroactively alter how historical analytics read. Flagged on #2669 for a product decision.

Adding `hold` does not touch it — `intake` still outranks — and a second test **pins the current behaviour** so the larger question gets answered deliberately rather than drifted into by a future edit to this map.

## 2. `check:lifecycle-columns` is RED on pristine `origin/main` — again

```
census exit on pristine main = 1
```

`executor.ts` is a file this PR does not touch, so a merge lowered the count without re-recording and the blocking PR check is failing for **every open PR**. The re-record is mechanical and is included here to unblock it — called out explicitly because it is unrelated to the funnel fix and should not ride along unexplained.

This is the second time the baseline has gone stale on main this way. The rule works (`--strict` caught it immediately); what is missing is that it caught it *after* the merge. Worth considering whether the census should run on the merge queue rather than only on PR head — otherwise a PR that is green when opened can still land a stale baseline.

## Verification

`pnpm test:gate` green (10 / 158 / 487 / 71). `pnpm check:lifecycle-columns` exits 0 after the re-record. `tsc -p packages/core/tsconfig.json` clean. `pnpm lint` clean. `sdlc-funnel-default-columns.test.ts` 8/8.

No changeset: the funnel entry is a correctness fix with no user-facing API change, and the baseline re-record is internal.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SDLC funnel classification for hold-only columns, placing them in the Todo stage instead of Other.
  * Preserved correct Planning column behavior when hold-related traits are combined.

* **Tests**
  * Added coverage for hold-related funnel stage mapping and trait ordering.
  * Updated lifecycle column census baselines to reflect current results.

### DIFF
--- a/packages/core/src/__tests__/sdlc-funnel-default-columns.test.ts
+++ b/packages/core/src/__tests__/sdlc-funnel-default-columns.test.ts
@@ -22,7 +22,7 @@ test stops being evidence and becomes a chore.
 */
 import { describe, expect, it } from "vitest";
 
-import { buildColumnStageMap } from "../activity-analytics.js";
+import { buildColumnStageMap, stageForTraits } from "../activity-analytics.js";
 import { resolveDefaultWorkflowIr } from "../builtin-workflows.js";
 import { BUILTIN_CODING_WORKFLOW_IR } from "../builtin-coding-workflow-ir.js";
 import type { WorkflowIrColumn } from "../workflow-ir-types.js";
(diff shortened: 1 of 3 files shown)


gsxdsm:
Both fixed, and the first one is this project's own Surface Enumeration rule applied to me — correctly.

**Column surface.** You are right that I proved the fix at the helper it was written in, not at the surface where the defect shows. The reported bug was a hold-only **column** disappearing from the funnel; `stageForTraits` is one caller down from that. Added a case through `buildColumnStageMap`, which is the seam the funnel actually consumes (column id → stage for every column on the board).

Made non-vacuous by construction: two neighbour columns (`wip`, `complete`) are asserted alongside the hold-only one, so it cannot pass because the map came back empty or because every column collapsed into a single stage.

**Mutation-verified:** removing the `hold` entry now fails **both** cases — helper and column map — with `expected 'other' to be 'todo'`. Before this change only one of them moved.

**Timestamps.** Correct, mine were future-dated; both are now `2026-07-30-16:00`. The other `2026-07-31` stamps in that file belong to its original author and are outside this change, so I left them rather than widen the diff.

Verification: `sdlc-funnel-default-columns.test.ts` **9/9**, `pnpm test:gate` green (10 / 158 / 487 / 71), `pnpm check:lifecycle-columns` exits 0, `pnpm lint` clean. Rebased onto current `origin/main`.

Unchanged from the description, and still the part that needs a decision rather than a patch: the merged Planning column carries `intake` + `reset-on-entry`, so `stageForTraits` still resolves it to `triage` and the `todo` stage stays empty on every default board. The second test in this file pins that current behaviour deliberately, so the larger question gets answered on purpose rather than drifted into by a future edit to the trait map.

gsxdsm:
Both findings handled.

**Timestamps:** corrected — three comments were stamped `2026-07-31` while the edit landed on `2026-07-30`. The convention exists so a reader can place a note against the change that motivated it, and a future date breaks precisely that.

**Column-level coverage:** already present, and I verified it rather than taking the file's word. Removing the `hold: "todo"` stage-map entry fails **both** the trait-level and the `buildColumnStageMap` test, so the surface the bug actually showed on is covered.

Worth flagging how nearly I got that wrong: my first mutation targeted an `if (traits.includes("hold"))` branch that doesn't exist in this file. It reported green, and had I stopped there I'd have recorded a verification I never performed. The production change is a single map entry; mutating *that* is what produced the evidence. That's the fourth mis-aimed mutation I've caught in myself today, so I now assert the mutation applied before trusting its result.

---

**Wider finding, deliberately not fixed here:**  carries **70 future-dated FNXC timestamps** across core, engine, dashboard and scripts. This isn't one author's slip — it's program-wide, and I contributed to it in my own commits today. Sweeping it belongs in its own change rather than riding along in a funnel fix, and it's a clean mechanical candidate for the cleanup fleet:

```
grep -rn 'FNXC:[A-Za-z]* 2026-07-31' packages/*/src packages/*/app scripts
```

Verified: 9 tests green, core tsc clean, lint clean, merge gate green (487 + 158 + 10).

PR #2677, opened 2026-07-30
Tell an already-converted fallback literal from an unconverted guard — 19 of 19 dashboard scan hits were the former

gsxdsm:
The batch phase is about to hand per-file guard lists to cheap workers, and the census currently cannot distinguish **"not yet converted"** from **"converted, with a documented degradation."**

## The measurement that makes this a class, not a preference

A proximity scan for *"legacy literal near a role-resolved call"* — the heuristic that produced #2670 and #2672 from the engine — returned **19 hits across the dashboard and zero defects.** Every one was:

```ts
if (flags) return flags.hold === true || flags.countsTowardWip === true;
return column === "todo" || column === "in-progress";      // reachable only without traits
```

That literal is **correct**: it answers for callers with no resolved column metadata, which is the case `resolveLifecycleColumns` returns `undefined`-for-the-whole-struct to preserve. A worker told to "convert" it would delete the only answer available when traits are absent.

## And the difference is structural, so the parser can see it

In **both** engine defects the literal sat in a **separate statement beside resolved data**, not in a fallback branch. Proximity cannot tell those apart; an AST can.

`traitFallback` flags the ternary form and the **early-return** form (which is how most are actually written), and deliberately does **not** flag a fallback whose test is itself a column-*name* check — otherwise any `if/else` over column names would launder itself.

## Reported beside the backlog, not subtracted from it

```
COLUMN guards (the backlog):   746
  of the column guards, 9 are trait-fallback branches (already converted)
```

A fallback literal is still a literal and should go when the trait path becomes unconditional. This only says which **kind** of work it is.

**Advisory, and structurally so:** `traitFallback` never changes `kind`, and the count lives *outside* `totals`. My first attempt put it in `totals` and broke two existing suites that correctly deep-equal that shape — an advisory number does not belong in the structure that defines the bar.

## Revert proof

Forcing `traitFallback: false` fails **3 of 33** (both fallback forms, plus the kind-unchanged case). The two *negative* cases pass under the revert — which is the point: they assert what must **not** be flagged, and a classifier that flags nothing satisfies them trivially. Worth stating, because a revert proof that only counts failures would look stronger than it is.

## Baseline

Re-recorded: `executor.ts` 87 → 85 was **main's own drift** from #2568 landing, so `--strict` was red on main again. #2668 made the re-record possible; the auto-tighten (coordinator item 2) is still open, and this is the third time in this program that a legitimate merge has left the gate red for everyone else.

## Verification

62/62 across both census suites, `pnpm test:gate` **71/71**, `--strict` exits 0, `pnpm lint` clean.

## Summary by CodeRabbit

* **New Features**
  * Added reporting for legacy column comparisons found in trait-fallback branches.
  * Census results now include a separate count for these fallback-related column guards.
  * Human-readable reports display the new metric alongside the existing backlog totals.

* **Tests**
  * Added coverage for fallback detection across ternary, early-return, and conditional patterns.
  * Added safeguards to prevent false positives in resolved-data and column-name checks.

* **Maintenance**
  * Updated baseline census metrics to reflect revised classifications.

### DIFF
--- a/packages/engine/src/__tests__/lifecycle-column-census-ast.test.ts
+++ b/packages/engine/src/__tests__/lifecycle-column-census-ast.test.ts
@@ -326,3 +326,86 @@ describe("state, phase and result enums are not column guards", () => {
     expect(result.totals.column).toBe(1);
   });
 });
+
+/*
+FNXC:LifecycleColumnCensus 2026-08-01-01:30:
+
+A LEGACY LITERAL IN A FALLBACK BRANCH IS NOT BACKLOG. The converted shape is
+
+    if (flags) return flags.hold === true || flags.countsTowardWip === true;
+    return column === "todo" || column === "in-progress";      // reachable only without traits
+
+and that literal is CORRECT — it answers for callers with no resolved column metadata, which is the case
+`resolveLifecycleColumns` returns `undefined`-for-the-whole-struct to preserve. Telling a batch worker to
+"convert" it means telling them to delete the only answer available when traits are absent.
+
+MEASURED, which is why this is a class and not a preference: a proximity scan for "legacy literal near a
+role-resolved call" over the dashboard returned 19 hits and ZERO defects, all this shape. The same scan over
+the engine found two real defects (#2670, #2672) — and in BOTH the literal was in a SEPARATE STATEMENT beside
+resolved data, not in a fallback branch. That difference is structural, so the parser can see it; proximity
+cannot.
+
+Advisory only: `traitFallback` never changes `kind`, so a wrong hint cannot move the bar. Repo-wide it flags
+9 of 746 column guards.
+*/
+describe("a literal in a trait-fallback branch is flagged as such", () => {
+  const fallbacksIn = (source: string) => census(source).filter((f) => (f as { traitFallback?: boolean }).traitFallback).length;
+
+  it("flags the ternary form", () => {
+    const source = `export function isHold(flags: F | undefined, columnId: string) {
+      return flags ? flags.hold === true : columnId === "todo";
+    }`;
+
+    expect(fallbacksIn(source)).toBe(1);
+  });
+
+  it("flags the early-return form, which is how most of them are written", () => {
+    const source = [
+      "function mayEdit(column: string, flags?: F) {",
+      "  if (flags) return flags.hold === true || flags.countsTowardWip === true;",
+      `  return column === "todo" || column === "in-progress";`,
+      "}",
+    ].join("\n");
+
+    // Two literals in one return, both in the fallback.
+    expect(fallbacksIn(source)).toBe(2);
+  });
+
+  it("does NOT flag a literal in a separate statement beside resolved data — the #2670 / #2672 shape", () => {
+    /*
+    This is the distinction the whole class exists for. Both engine defects looked like this: resolved lane
+    data in scope, and a guard next to it still matching a name. Flagging these as fallbacks would have hidden
+    exactly the two bugs the scan found.
+    */
+    const source = [
+      "async function park(task: T) {",
+      "  const lanes = await resolveLifecycleColumns(ir);",
+      `  if (task.column === "done" || task.column === "archived") return false;`,
+      "  await move(task.id, lanes.hold);",
+      "}",
+    ].join("\n");
+
+    expect(fallbacksIn(source)).toBe(0);
+  });
+
+  it("does NOT flag a fallback branch whose test is itself a column-name check", () => {
+    // `if (column === "todo")` tests a NAME, not resolved data, so its else branch is not a trait fallback —
+    // otherwise any if/else over column names would launder itself.
+    const source = [
+      "function f(column: string) {",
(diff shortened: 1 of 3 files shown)


gsxdsm:
Real, and fixed. You were right that scanning the branch for a `return` token proves nothing about termination.

The failing shape:

```ts
if (flags) {
  if (flags.hold === true) return true;
}
return column === "todo";      // reachable WITH traits — a live guard
```

The detector now asks whether the branch **definitely terminates**: a return/throw, a block whose last statement terminates, or an `if`/`else` where **both** arms do. Structure rather than token presence.

**Why this direction is the bad one.** The misclassification removes a real guard from the backlog the census is trusted to report, and it does so silently — the count just comes back smaller, which reads as progress. Nearly every instrument failure in this program has had that shape, so a hole that under-reports is worth more scrutiny than one that over-reports.

Three tests, including the paired positive so the correction can't overshoot into "no early returns":
- conditional return in the branch → **not** a fallback
- block ending in a call after a nested return → **not** a fallback
- both arms return → **still** a fallback

Mutation: restoring the token scan fails 2 of 36.

**One measured result worth stating, because it isn't what a reader will assume:** repo-wide the count is **unchanged at 9 of 746**. No site in the tree hits the loosened form today. This closes a hole for future conversions rather than correcting a present miscount — **the bar does not move**, and I'd rather say that than let the fix imply progress it didn't make.

Verified: 36 tests green, `census --strict` exit 0, lint clean, merge gate green (487 + 158 + 10 + 71).

gsxdsm:
Already fixed on this branch — and the way I found that out is worth recording.

`8717666881` ("trait hints must not match inside a longer identifier") landed the same fix, from my **other worktree** on this same branch, before I pushed mine. Its matcher (`[.?]${hint}\b` — dropping the `\w` from the leading character class) and mine (dropping that alternative entirely) both fail closed on the substring case, and its tests cover the same decoys plus `myflags`. So I discarded my duplicate rather than force a merge of two equivalent fixes.

Your diagnosis was right and the direction matters: **a false "already converted" hides backlog** — a batch worker skips a real unconverted guard because the census told them it was done — whereas a false "unconverted" only costs a reviewer a minute. Measured effect of the fix: trait-fallback 9 → 8, so one of the original nine was exactly this false positive on current main.

38/38 on the AST suite at `8717666881`; `--strict` exits 0.

**The process finding, since this is the third duplicate in this program and the first one I caused against myself:** I have two worktrees on the same repo, and this branch was checked out in the other one. `git checkout` refused it here (worktree lock), I worked from a detached commit instead, and only discovered the collision when the push was rejected as non-fast-forward. The push rejection was the *only* signal — every local check was green.

Two habits that would have caught it a step earlier, and both are cheap enough for the fleet: **`git fetch` the branch before starting work on it, not just before pushing**, and treat a worktree-lock refusal as "someone is already here" rather than as a mechanical obstacle to route around.

Also worth flagging for whoever merges: **#2677 and #2679 both touch the census CLI/lib and will conflict.** #2679 (the auto-tighten) is the one with the CI consequence, so I would land that first and rebase this one onto it — say the word and I will do that rather than have you resolve it.

gsxdsm:
Real, and this one moved a number. Fixed.

`TRAIT_TEST_HINTS` includes `hold`, and the leading class was `[.?\w]` — so the hint matched inside `threshold`, `staleThreshold`, `household`, `stronghold`, `withhold`, and `flags` matched `myflags`. A branch testing an unrelated threshold was read as testing resolved trait data, marking the legacy comparison below it as already-converted.

**The `\w` was also redundant**, which is why removing it costs nothing rather than trading one miss for another: the third alternative `\bhold\b` already matches `flags.hold` and `flags?.hold`, because `.` is a non-word character and supplies the boundary itself. The `\w` alternative contributed only false positives.

**Measured:** trait-fallback count drops **9 → 8**. One live guard in the tree was being reported as already converted. Unlike the termination fix in the previous commit — which closed a hole but moved nothing — this corrected a present miscount. The documented count in the test header is updated in the same commit so the two cannot drift.

Two tests, both directions: four hint-inside-identifier forms must **not** be fallbacks; four genuine reads (`flags.hold`, `flags?.hold`, `lifecycleRoles.hold`, bare `flags`) must still be. Mutation: restoring `\w` fails 1 of 38.

---

**Separate pre-existing finding, not caused by this branch.** `census --compare` exits **1**, and I verified it does so on `origin/main` and at every commit on this branch, so it predates the work here:

```
text classifier:  {"column":752,...}
AST classifier:   {"column":722,...}
the regex found MORE than the parser for column.
The parser has a blind spot; its count cannot be the bar until this is closed.
```

This matters more than it looks. The cross-check's own message says the parser's count cannot be the bar — and the parser's count **is** what the program has been using as the bar. The direction has also inverted from what the test header documents ("the parser is a strict SUPERSET of the regex, measured +6"). Only `--strict` runs in the gate, so `--compare` has been red and unwatched. Worth its own task; I have not touched it here.

Verified after rebasing onto current main: 38 tests green, `--strict` exit 0, lint clean, gate green (487 + 158 + 10 + 71).

PR #2680, opened 2026-07-30
docs: correct 104 future-dated FNXC timestamps across 61 files

gsxdsm:
## What

The FNXC convention exists so a reader can place a note against the change that motivated it. A stamp dated *after* the edit landed defeats exactly that.

This is program-wide drift, not one author's slip — I contributed to it in my own commits this week, which is how I noticed it.

## Measured, on this tree

**104 stamps across 61 files** dated later than the day they were written, from one day ahead to **2026-10-19 (81 days)**:

| count | date | count | date | count | date |
|---|---|---|---|---|---|
| 50 | 2026-07-31 | 6 | 2026-08-05 | 3 | 2026-08-13 |
| 17 | 2026-08-01 | 1 | 2026-08-07 | 1 | 2026-08-19 |
| 7 | 2026-08-02 | 1 | 2026-08-12 | 2 | 2026-08-26 |
| 11 | 2026-08-03 | | | 3 | 2026-10-19 |

An earlier number I circulated was ~70. That came from a narrower pathspec and was wrong; **104** is the measurement.

## How

Each stamp is rewritten to the date of the commit that introduced **that line**, via per-line `git blame` — deliberately *not* stamped uniformly with today's date. A uniform stamp swaps a wrong date for a different wrong date and flattens the ordering that makes these comments navigable; blame preserves it. Times of day are untouched, and a blame date in the future is clamped rather than trusted.

## Why the verification is listed

A docs sweep across 61 files is precisely where a stray edit hides, so the safety claims are mechanical rather than asserted:

- every changed line begins with a comment marker — **no code touched**;
- **no test asserts an FNXC date later than today**, so no `toContain` assertion on embedded source text can be silently invalidated (several such assertions do exist);
- CSS files, which carry several of those assertions, are outside the pathspec.

## Verified

lint clean · merge gate green (487 + 158 + 10 + 71) · `census --strict` exit 0 · tsc clean for core, engine, and dashboard (`tsconfig.app.json`).

**No behavior change.** Comment text only.

## Not done here

A guard preventing recurrence. A check that rejects an FNXC stamp dated after the commit would stop this returning, but it needs a decision about where it runs (lint rule vs. gate) and it is a behavior change to CI — it does not belong riding inside the sweep it would police.

### DIFF
--- a/packages/core/src/__tests__/coding-ideas-move.test.ts
+++ b/packages/core/src/__tests__/coding-ideas-move.test.ts
@@ -115,7 +115,7 @@ pgDescribe("Coding (Ideas) custom-column moves (workflow-columns graduation)", (
       store.moveTask(task.id, "ideas", { moveSource: "user" }),
     ).rejects.toThrow(/Invalid transition: '.*' → 'ideas'/);
     /*
-    FNXC:WorkflowColumns 2026-07-31-04:30 (U12 — the move-path flag is resolved):
+    FNXC:WorkflowColumns 2026-07-30-04:30 (U12 — the move-path flag is resolved):
     ...and so is a non-adjacent target. The advertised target list changed from
     `in-progress, triage, archived` to `archived, in-progress`, and that is the FIX rather than a
     regression: the old list came from the hardcoded legacy adjacency table, which still offered
--- a/packages/core/src/__tests__/live-move-path-undeclared-target.test.ts
+++ b/packages/core/src/__tests__/live-move-path-undeclared-target.test.ts
@@ -106,7 +106,7 @@ pgDescribe("live move path — which targets it accepts after the Planning merge
     expect(workflowHasColumn(ir, "todo")).toBe(true);
 
     /*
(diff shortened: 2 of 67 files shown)


gsxdsm:
Rebased — by **regenerating**, not by resolving.

The branch conflicted with main across 61 comment blocks. Hand-resolving that many mechanical conflicts is exactly where a stray edit gets in (I have made that mistake twice this week), so I reset the branch to current main and re-ran the sweep script. The result is derived from main deterministically rather than merged into it.

Count moved **104 → 113 across 67 files** because main gained commits in the interval — that is the sweep finding new drift, not a change of method.

All four safety claims re-verified on the regenerated diff:
- every changed line begins with a comment marker — no code touched;
- no test asserts an FNXC date later than today;
- CSS files (which carry several such assertions) are outside the pathspec;
- **zero** future-dated stamps remain in the swept paths.

lint clean · gate green (487 + 158 + 10 + 71) · `census --strict` exit 0. No behavior change.

PR #2682, opened 2026-07-30
fix(scripts): --compare was accusing the parser of a blind spot it does not have

gsxdsm:
## The problem

`census --compare` fails on `origin/main` — I verified it at `bc782d8d92` and at every commit on the branch where I found it. Its failure message reads:

> The parser has a blind spot; its count cannot be the bar until this is closed.

That matters because the parser's count **is** the bar the program just used to declare the closing bar met (triage 0, backlog 722). A red cross-check asserting the instrument is untrustworthy had to be settled in one direction or the other.

## It is settled: there is no blind spot

**Measured — 13 divergent sites, all 13 seen by the parser:**

| parser's classification | count |
|---|---|
| `deliberate` | 4 |
| `role` | 5 |
| `status` | 4 |
| **missed entirely** | **0** |

## The bug is in the check

It compared per-bucket totals and failed when the regex's `column` total exceeded the parser's. That conflates the two things it most needs to separate:

- the parser **missed** a site → a real hole, the failure worth having;
- the parser **classified it better** → `role`/`status`/`deliberate` instead of `column`.

The second is the parser's entire reason for existing. So the old form fired *more* the better the parser got, while accusing it of the one defect it did not have. The regex is knowingly weaker at telling an agent role from a column guard — that asymmetry is why the parser was adopted, and the check was penalising it.

The intent was never wrong; the comment above the check already said the contract was "a site the REGEX found and the parser missed". Only the implementation disagreed with it.

## After

```
text classifier:  {"column":728,"role":0,"status":182,"deliberate":14}
AST classifier:   {"column":722,"role":5,"status":186,"deliberate":17}
parser sees every site the regex does (+131 sites the regex cannot see).
14 the regex calls a column guard, the parser classifies as {"role":5,"status":4,"deliberate":4,"definition":1}.
```

Fails only on a genuinely missed site now, printing the first ten. Reclassifications are reported rather than failed.

## Scope

Report-only. `--compare` is not in the merge gate — the gate runs `--strict`, which is why this stayed red and unwatched. Verified: `--compare` exit 0, `--strict` exit 0, lint clean. No census numbers change.

### DIFF
--- a/scripts/lifecycle-column-census.mjs
+++ b/scripts/lifecycle-column-census.mjs
@@ -132,21 +132,55 @@ if (compare) {
   What must NEVER happen is the other direction: a site the REGEX found and the parser missed means
   the parser has a hole, and then its number cannot be the bar. That is the failure this checks.
   */
-  const text = summarizeText(censusFilesText(files));
+  /*
+  FNXC:LifecycleColumnCensus 2026-07-30-11:30:
+  COMPARE SITES, NOT BUCKET TOTALS. This check used to compare the per-bucket counts and fail when
+  the regex's `column` total exceeded the parser's. That conflates the two things it most needs to
+  tell apart:
+
+    - the parser MISSED a site entirely            -> a real blind spot, the failure worth having;
+    - the parser saw it and classified it better   -> role, status, or deliberate instead of column.
+
+  The second is the parser's entire reason for existing, so the old form fired MORE the better the
+  parser got. It had been failing on `main` while reporting "the parser has a blind spot; its count
+  cannot be the bar" — and that message was false. MEASURED at the time of this change: 13 sites
+  diverged, and all 13 were seen by the parser (4 deliberate, 5 role, 4 status). Zero were missed.
+
+  The check now fails only on a site the regex found and the parser did not, which is what the note
+  above it always said the contract was.
+  */
+  const textFindings = censusFilesText(files);
+  const text = summarizeText(textFindings);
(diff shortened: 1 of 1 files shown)


gsxdsm:
Real, and it applies to three lines in the tree today. Fixed.

`from === "todo" || to === "todo"` produces two findings sharing `file:line:columnId`. Keyed by a Set, one parser match satisfies both regex findings — so a genuine miss of the second is invisible, which is the one thing this check exists to catch.

It now compares **per-key counts**. Receiver is deliberately *not* added to the key, because `c === "todo" || c === "todo"` would collapse again; counts cannot be fooled by either shape.

**Proven rather than argued.** I blinded the parser to exactly one of a real duplicate pair — `project-store-ops.ts:508`, `in-review` ×2 — and ran both forms against the same blinded parser:

| comparison | result |
|---|---|
| multiset (this commit) | **exit 1**, naming `project-store-ops.ts:508` |
| set (previous) | **exit 0**, miss hidden |

That is your scenario reproduced on a real site instead of a fixture.

Reclassification reporting now pairs occurrences positionally within a key, for the same reason. Its printed counts shift slightly (14 → 13 reclassified, +131 → +132 parser-only) because the pairing is accurate now — **no site changed classification**, and the headline conclusion of this PR is unaffected: the parser still sees every site the regex does.

Verified: `--compare` exit 0, `--strict` exit 0, lint clean, gate green (487 + 158 + 10 + 71).

PR #2683, opened 2026-07-30
fleet: self-healing.ts 110 → scoped (NOT converted) — sync workflow reads make this cluster unsafe to batch

gsxdsm:
Claiming the largest unclaimed cluster per the work order, then **handing it back sized rather than half-converted.** Docs only; census unchanged (722 / triage 0).

## The cluster

`packages/engine/src/self-healing.ts` — **110 guards**, largest single file in the order.

```
by column:   in-review 48 · in-progress 20 · done 17 · todo 13 · archived 12
by receiver: column 100 · to 7 · from 3
```

## Why the mechanical conversion is unsafe here

**The engine has no synchronous way to learn a task's workflow.** `resolveTaskWorkflowIrSync` returns the DEFAULT IR for every task in production — `getTaskWorkflowSelection` returns `undefined` unconditionally (a PG-cutover stub), so the reader always takes its `!workflowId` branch. It is typed non-optional, so **no caller can detect the substitution.**

A conversion routed through it: compiles, reads better than the literal, **counts as census progress**, and is wrong for every custom workflow, silently. That is strictly worse than leaving the literal — the literal is at least honest about being one. It is the "guard that cannot fire" pattern wearing better clothes, and the ratchet would score it as a win.

The correct form uses `resolveTaskLifecycleColumns(store, taskId)` (async, store-aware), which needs resolved lanes **in scope per method**. Sampled sites (926, 932, 984) do sit in `async` methods so it is reachable — but that is a per-sweep restructuring, not a per-line substitution, and these sweeps iterate task lists, so a naive per-task resolve turns one sweep into N store reads.

**In-tree precedent:** `triage.ts` `discoverReadyPlanningTasks` solved this exact problem — store-free `couldBeCandidate` prefilter, bounded (8) concurrent resolve over the survivors, decision stays synchronous over a resolved map. Any batch here should follow that shape per sweep.

## Recommended split, by SWEEP not by column

110 sites cannot honour *"census before/after, baseline shrinks by exactly the converted count"* while also restructuring six-plus sweeps in one PR.

1. **the review/merge sweeps** (`in-review` 48) — largest, and the one where a wrong lane silently changes **merge eligibility**. First and alone.
2. **WIP/rebound sweeps** (`in-progress` 20, `todo` 13).
3. **terminal sweeps** (`done` 17, `archived` 12) — read `complete`/`archived`; most mechanical of the three.
4. **the 10 `from`/`to` sites** — these are MOVE-transition arms, not task-column reads. Different question (*"is this transition into a review lane?"*), so they must not ride along with the `task.column` work.

## Why I am not doing item 1 myself

I am near the end of a long session — this is the same context in which I produced a confidently-wrong structural finding earlier today (retracted in #2667, where I trusted a hand-rolled brace counter over a comment in the file). A 48-site restructuring of the merge-eligibility sweeps is exactly the work that should not be done by a worker in that state, and the fleet rules' *flag-and-skip* discipline is the right call over guessing.

**What a fresh worker gets from this PR:** the site census, the async-scope survey, the hazard with its root cause, the in-tree pattern to copy, and a four-way split with the risky piece isolated. That is the expensive part of the job already done.

## Fleet rule this cluster proves, worth adding to the brief

**Never resolve a workflow synchronously in a converted guard.** Use `resolveWorkflowIrForTaskWithProvenance` (branch on `source`) or `resolveTaskLifecycleColumns`; if neither is reachable at the site, flag and skip.

### DIFF
--- a/docs/solutions/architecture-patterns/fleet-self-healing-cluster-scoping.md
+++ b/docs/solutions/architecture-patterns/fleet-self-healing-cluster-scoping.md
@@ -65,3 +65,35 @@ one resolve-then-decide unit:
 `resolveWorkflowIrForTaskWithProvenance` (branch on `source`) or `resolveTaskLifecycleColumns`, and
 if neither is reachable at the site, FLAG AND SKIP per the fleet rules — a sync-resolved trait read
 is the "guard that cannot fire" pattern wearing better clothes.
+
+## Second rule, from sizing the dashboard clusters: traits-first does NOT lower the count
+
+Checked `TaskContextMenu.tsx` (9) next, expecting an easy cluster, and found the opposite lesson.
+
+Its guards live in ROLE HELPERS that already read traits:
+
+```ts
+// isReviewColumn
+return column === "in-review" || flags?.mergeBlocker === true || flags?.humanReview === true;
+```
+
+Rewriting that traits-first (`if (flags) return …; return column === "in-review"`) is the correct
+shape — it is what #2664 did to `isPreExecutionHoldColumn` — but **the census still counts it**,
+because the literal survives as the no-metadata fallback. The count is per literal, not per
+code-quality improvement.
+
+So for a file of role helpers, "N -> 0" is reached by MARKING the fallbacks
+(`DELIBERATE-LITERAL`), not by converting them. And the failure mode is specific: a worker chasing
+0 by DELETING the fallback silently breaks degraded mode — the pre-load window (board renders before
+the workflows fetch resolves) and a card stranded on an id its workflow no longer declares both
+arrive with no flags at all, so the helper starts answering "no role" for every column and
+affordances vanish during first paint.
+
+That is not hypothetical: `columnRoles.ts` documents the pre-load window explicitly, and the
+`Column.tsx` / `taskActivity.ts` conversions earlier in this program each had to keep exactly this
+fallback.
+
+**Rule for the fleet:** in a role helper, convert the ORDER (traits first, id as fallback) and then
+MARK the fallback. Deleting it is a behavior change in degraded mode — flag and skip per the rules.
+Expect the dashboard clusters (`TaskCard.tsx` 42, `TaskDetailModal.tsx` 30) to be mostly marks, and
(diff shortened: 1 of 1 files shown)


gsxdsm:
## Second rule added: traits-first does **not** lower the count

I went looking for a cluster sized to what I can finish and took `TaskContextMenu.tsx` (9) next. It taught the opposite of what I expected, and the lesson generalises to every dashboard cluster in the order.

Its guards live in **role helpers that already read traits**:

```ts
// isReviewColumn
return column === "in-review" || flags?.mergeBlocker === true || flags?.humanReview === true;
```

Rewriting that traits-first (`if (flags) return …; return column === "in-review"`) is the **correct** shape — it is exactly what #2664 did to `isPreExecutionHoldColumn` — but **the census still counts it.** The literal survives as the no-metadata fallback, and the count is per literal, not per improvement.

So in a role-helper file, **"N → 0" is reached by MARKING the fallbacks, not by converting them.**

**The failure mode is specific and silent.** A worker chasing 0 by *deleting* the fallback breaks degraded mode: the pre-load window (board renders before the workflows fetch resolves) and a card stranded on an id its workflow no longer declares **both arrive with no flags at all**. The helper then answers "no role" for every column and affordances vanish during first paint. Not hypothetical — `columnRoles.ts` documents that window, and the `Column.tsx` and `taskActivity.ts` conversions earlier in this program each had to keep exactly this fallback.

**Rule:** convert the ORDER (traits first, id as fallback), then MARK the fallback. Deleting it is a behavior change in degraded mode → flag and skip.

**Sizing consequence:** expect `TaskCard.tsx` (42) and `TaskDetailModal.tsx` (30) to be **mostly marks**. The count moves the same either way; only one of the two is safe. A batch that reports "42 → 0" with 42 deletions should be reviewed for exactly this.

That is why I am not converting `TaskContextMenu`'s 9 either: the right action there is a mixed mark/convert/skip pass with a per-site judgement, and per-site judgement is the thing I should not be doing at the end of this session. Both rules are now recorded where the next worker will find them.

PR #2684, opened 2026-07-30
fleet: packages/engine/src/self-healing.ts 110 → 0 (claimed; in progress — 109 now, plus a structural finding that changes how this cluster must be worked)

gsxdsm:
Claiming the largest unclaimed cluster in the census work order: **`packages/engine/src/self-healing.ts`, 110 guards.** Opening the PR early because I hit something in the first hour that changes how this file — and probably `executor.ts` — has to be converted, and it is better raised now than after 100 commits.

## Progress

```
census   110 -> 109      baseline re-recorded in the same commit
engine-core 487 green    typecheck clean    check:lifecycle-columns exit 0
```

One site, deliberately. I converted exactly one and confirmed the census moved by exactly one before scaling, because "the baseline must shrink by exactly your converted count" is only meaningful if the loop is proven first.

`resolveReviewColumn` is a **sibling of the file's existing `resolvePreWipColumns`** — same shape, same cache parameter, same documented fallback discipline (recovery sweeps keep current behaviour when an IR cannot be read). Not a new abstraction: the role helper is `resolveLifecycleColumns`, which this file already calls at three sites.

## The structural finding — guard-only conversion in this file is ACTIVELY HARMFUL

Most guards here are fed by a **literal query**:

```ts
const tasks = await this.store.listTasks({ column: "in-review", slim: true });
const stale = tasks.filter((task) => {
  if (task.column !== "in-review" || task.paused) return false;
```

Converting the `filter` guard alone leaves the **source query** on a dead literal. On a renamed or merged board that query returns nothing, so the sweep processes zero candidates — and the file now *looks* converted while doing strictly nothing. The census drops, the ratchet is satisfied, and the sweep is dead.

This is not hypothetical in this file:

- `self-healing.ts:3015` documents the exact defect from #2560, where a converted predicate was left with a literal query and the recovery went silently dead. The fix pattern — *read the board and filter by role* — is written there.
- I proved a live instance earlier this program: `recoverStuckMergeDeadlocks` cannot see a renamed board at all. Its three-literal union returns empty (`renamedInsideUnion=0`, measured on a live PG store), so its correct role filter is starved.

**The census cannot see this.** It counts comparison `BinaryExpression`s; `{ column: "in-review" }` is a `PropertyAssignment`. This file holds **~48 such query filters**, tracked separately by the `queryByFile` instrument added in #2650 but not part of the 110.

## How I am working it, and where I need a ruling

Guard and its source query are **one unit**. I will convert them together where the file's established pattern fits (`filterByPreWipRole` and the #2560 repair are both in-file precedent, so this is not a new abstraction). Where converting the query would change the data-access path in a way that is not behaviour-preserving, I flag and skip per the rules.

The consequence for the fleet arithmetic, which is the ruling I need: **converting a guard+query pair moves the guard count by 1 and the query count by 1**, and those are two separately-pinned numbers. A PR that does this correctly will not show "baseline shrank by exactly the converted guard count" unless the query baseline is re-recorded too. I will re-record both and state both in the body; flag if you want it done differently.

## Not yet converted, and why

The remaining 109 are mostly the paired shape above. Working them in small commits by sweep, each with its query partner, rather than a fast pass over the comparisons that would hollow out the file's recovery paths.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved self-healing recovery and maintenance sweeps to correctly interpret workflow lanes after renames or merges.
  * Recovery/reclaim logic now classifies tasks by workflow role (WIP/review/complete/archived) rather than relying on fixed lane identifiers.
* **Tests / Chores**
  * Updated self-healing test expectations to match the revised candidate selection behavior.
  * Refreshed lifecycle column census baseline data to reflect role-based lane handling.

### DIFF
--- a/packages/engine/src/__tests__/self-healing-complete-lane-renamed.test.ts
+++ b/packages/engine/src/__tests__/self-healing-complete-lane-renamed.test.ts
@@ -1,0 +1,105 @@
+/*
+FNXC:WorkflowLifecycleColumns 2026-07-30-21:25 (PR #2684 review — the coverage half):
+
+RATCHET FOR THE COMPLETE-LANE FILTER, WHICH WAS INERT AND HAD NO TEST THAT COULD SAY SO.
+
+`filterByCompleteRole` used to resolve through the SYNC reader. That reader cannot see a task's
+selection — `getTaskWorkflowSelectionImpl` returns `undefined` unconditionally ("sync selection reader
+is incomplete-PG") — so `resolveTaskWorkflowIrSync` fell back to the DEFAULT workflow IR and the filter
+answered the default board's complete lane for every task, on every board. It read as converted and
+behaved exactly like the literal it replaced. The fix (this PR) routes it through
+`resolveWorkflowIrForTask` with the caller's cache, the only resolver that reads the selection.
+
+WHY THE EXISTING TESTS COULD NOT CATCH IT. Every other sweep test in this file's neighbourhood runs on
+the DEFAULT board, where the sync fallback and the correct answer are the same string. That coincidence
+is what let an inert filter look covered. This fixture names its complete lane `shipped`, where the two
+answers differ, so the sync form fails it and the async form passes.
+
+The negative case is here for the same reason: without it, "keep everything" also passes.
+*/
+import { describe, expect, it, vi } from "vitest";
+import { SelfHealingManager } from "../self-healing.js";
+import type { Task, WorkflowIr } from "@fusion/core";
+
+/** Standard roles under non-default names, so a default-board answer cannot accidentally match. */
+const RENAMED_IR = {
+  version: "v2",
+  id: "wf-renamed",
+  name: "renamed",
+  nodes: [],
+  edges: [],
+  columns: [
+    { id: "queued", name: "Queued", traits: [{ trait: "hold", config: { release: "capacity" } }] },
+    { id: "building", name: "Building", traits: [{ trait: "wip", config: { limitSetting: "maxConcurrent" } }] },
+    { id: "signoff", name: "Signoff", traits: [{ trait: "merge" }] },
+    { id: "shipped", name: "Shipped", traits: [{ trait: "complete" }] },
+    { id: "attic", name: "Attic", traits: [{ trait: "archived" }] },
+  ],
+} as unknown as WorkflowIr;
+
+function taskIn(id: string, column: string): Task {
+  return {
+    id,
+    lineageId: id,
+    title: "card",
+    description: "",
+    column,
+    dependencies: [],
+    steps: [],
+    currentStep: 0,
+    log: [],
+    createdAt: "2026-07-30T00:00:00.000Z",
+    updatedAt: "2026-07-30T00:00:00.000Z",
+  } as unknown as Task;
+}
+
+function managerFor(tasks: Task[]) {
+  const selection = { workflowId: "wf-renamed", stepIds: [] as string[] };
+  const store = {
+    getTaskWorkflowSelection: () => selection,
+    getTaskWorkflowSelectionAsync: async () => selection,
+    getWorkflowDefinition: async () => ({ id: "wf-renamed", ir: RENAMED_IR }),
+    /*
+    Stubbed to do what production does: answer without ever seeing the task's selection. Leaving it
+    working would let the mock succeed where production cannot, and the test would then pass against
+    the very implementation it exists to reject.
+    */
+    resolveTaskWorkflowIrSync: () => {
+      throw new Error("sync selection reader is incomplete-PG");
+    },
+    listTasks: vi.fn().mockResolvedValue(tasks),
+    getTask: vi.fn(async (id: string) => tasks.find((t) => t.id === id)),
+    getSettings: vi.fn().mockResolvedValue({ autoMerge: true }),
+    updateTask: vi.fn().mockResolvedValue(undefined),
+    logEntry: vi.fn().mockResolvedValue(undefined),
+    recordRunAuditEvent: vi.fn().mockResolvedValue(undefined),
+  };
+  const manager = new SelfHealingManager(store as never, { rootDir: "/tmp/complete-lane-test" } as never);
+  const filterByCompleteRole = (manager as unknown as {
+    filterByCompleteRole: (tasks: Task[], cache: Map<string, unknown>) => Promise<Task[]>;
+  }).filterByCompleteRole.bind(manager);
+  return { store, manager, filterByCompleteRole };
+}
+
+describe("filterByCompleteRole resolves the complete lane from the task's OWN workflow", () => {
+  it("keeps a card in the RENAMED complete lane", async () => {
+    const shipped = taskIn("FN-SHIPPED", "shipped");
+    const building = taskIn("FN-BUILDING", "building");
+    const { filterByCompleteRole } = managerFor([shipped, building]);
+
+    const kept = await filterByCompleteRole([shipped, building], new Map());
+
+    /*
+    With the sync reader this matched nothing, so every sweep behind this filter silently processed an
+    empty list — a failure that reads as "no work to do" rather than as an error.
+    */
+    expect(kept.map((t) => t.id)).toEqual(["FN-SHIPPED"]);
+  });
+
+  it("excludes a card in the DEFAULT complete lane this board does not declare", async () => {
+    const legacyDone = taskIn("FN-LEGACY", "done");
+    const { filterByCompleteRole } = managerFor([legacyDone]);
+
+    expect(await filterByCompleteRole([legacyDone], new Map())).toEqual([]);
+  });
+});
--- a/packages/engine/src/__tests__/self-healing.test.ts
+++ b/packages/engine/src/__tests__/self-healing.test.ts
@@ -2462,7 +2462,20 @@ describe("SelfHealingManager", () => {
       const result = await managerWithRecovery.recoverCompletedTasks();
 
       expect(result).toBe(1);
-      expect(store.listTasks).toHaveBeenCalledWith({ column: "in-progress", slim: true });
+      /*
+      FNXC:WorkflowLifecycleColumns 2026-07-30-13:30 (fleet: self-healing.ts):
+      Was `toHaveBeenCalledWith({ column: "in-progress", slim: true })`, pinning the exact literal
+      query this conversion removes. `recoverCompletedTasks` now reads the BOARD and filters by the
+      WIP role, because a query naming `in-progress` returns nothing on a board that renamed its
+      implementation lane — the sweep would look converted and recover nothing.
+
+      Asserted as "called without a column filter" rather than deleted: that is the property the
+      conversion establishes, so a regression to any single-column query fails here. The outcome
+      assertions either side are untouched and remain the ones proving the sweep works.
+      */
+      expect(store.listTasks).toHaveBeenCalledWith(
+        expect.not.objectContaining({ column: expect.anything() }),
+      );
       expect(recoverFn).toHaveBeenCalledWith(
         expect.objectContaining({ id: "FN-001" }),
       );
--- a/packages/engine/src/self-healing.ts
+++ b/packages/engine/src/self-healing.ts
@@ -11484,7 +11484,28 @@ export class SelfHealingManager extends SelfHealingGitEvidence {
       }
 
       const linkedTask = await this.store.getTask(agent.taskId);
-      if (linkedTask && (linkedTask.column === "in-progress" || linkedTask.column === "in-review" || linkedTask.column === "done" || linkedTask.column === "archived")) {
+      /*
+      FNXC:WorkflowLifecycleColumns 2026-07-30-16:05 (fleet: self-healing.ts):
+      "The agent's task is still being worked, or has finished" — WIP, review, complete or archived.
+      Keyed by role so an agent linked to a card on a renamed board is not treated as orphaned and
+      reaped while its task is genuinely active. No source query to pair: the loop is over running
+      AGENTS, and the task is fetched by id.
+      */
+      const linkedLanes = linkedTask
+        ? await (async () => {
+          try {
+            return resolveLifecycleColumns(await resolveWorkflowIrForTask(this.store, linkedTask.id));
+          } catch {
+            return undefined;
+          }
+        })()
+        : undefined;
+      if (linkedTask && (
+        linkedTask.column === (linkedLanes?.wip ?? "in-progress")
+        || linkedTask.column === (linkedLanes?.review ?? "in-review")
+        || linkedTask.column === (linkedLanes?.complete ?? "done")
+        || linkedTask.column === (linkedLanes?.archived ?? "archived")
+      )) {
         continue;
       }
 
--- a/packages/engine/src/self-healing.ts
+++ b/packages/engine/src/self-healing.ts
@@ -2780,11 +2780,16 @@ export class SelfHealingManager extends SelfHealingGitEvidence {
     if (!recoverFn) return 0;
 
     try {
-      const tasks = await this.store.listTasks({ column: "in-progress", slim: true });
+      /* Guard AND source query: filtering the board by the WIP ROLE keeps this sweep alive on a
+         board that renamed its implementation lane. Converting only the predicate below would
+         leave the query naming a column such a board does not declare. */
+      const tasks = await this.filterByWipRole(
+        await this.store.listTasks({ slim: true, includeArchived: false }),
+        new Map<string, Awaited<ReturnType<typeof resolveWorkflowIrForTask>>>(),
+      );
       const executingIds = this.options.getExecutingTaskIds?.() ?? new Set<string>();
 
       const stuckCompleted = tasks.filter((t) =>
-        t.column === "in-progress" &&
         !t.paused &&
         !executingIds.has(t.id) &&
         t.steps.length > 0 &&
@@ -3020,6 +3025,34 @@ export class SelfHealingManager extends SelfHealingGitEvidence {
   That is the defect #2560 repaired two hundred lines above this, and the reason the repair there is
   described as "read the board and filter by role" rather than as a predicate change.
   */
+  /** The WIP (implementation) lane for one task, resolved from its own workflow. Same fallback
+   *  discipline as `resolvePreWipColumns`: a recovery sweep whose IR cannot be read keeps its
+   *  current behaviour rather than dropping the card. */
+  private async resolveWipColumn(
+    taskId: string,
+    cache?: Map<string, Awaited<ReturnType<typeof resolveWorkflowIrForTask>>>,
+  ): Promise<string> {
+    try {
+      const lifecycle = resolveLifecycleColumns(await resolveWorkflowIrForTask(this.store, taskId, cache));
+      return lifecycle?.wip ?? "in-progress";
+    } catch {
+      return "in-progress";
+    }
+  }
+
+  /** Sibling of `filterByPreWipRole` for the WIP lane — see `filterByReviewRole` for why the
+   *  SOURCE QUERY must be converted alongside the guard it feeds. */
+  private async filterByWipRole(
+    tasks: Task[],
+    cache: Map<string, Awaited<ReturnType<typeof resolveWorkflowIrForTask>>>,
+  ): Promise<Task[]> {
+    const kept: Task[] = [];
+    for (const task of tasks) {
+      if (task.column === await this.resolveWipColumn(task.id, cache)) kept.push(task);
+    }
+    return kept;
+  }
+
   private async filterByReviewRole(
     tasks: Task[],
     cache: Map<string, Awaited<ReturnType<typeof resolveWorkflowIrForTask>>>,
--- a/packages/engine/src/self-healing.ts
+++ b/packages/engine/src/self-healing.ts
@@ -9645,19 +9645,28 @@ export class SelfHealingManager extends SelfHealingGitEvidence {
       const inReview = await this.store.listTasks({ column: "in-review", slim: true });
       /*
       FNXC:WorkflowColumns 2026-07-29-17:40 (PR #2560 review):
-      This trio is a UNION and is deliberately left on literals. For the merged
-      default lineage the `triage` read returns empty and `todo` supplies the cards;
-      for a legacy/custom workflow that still declares `triage` it supplies them.
-      Either way the union is complete, and the role filter below decides which rows
-      count as pre-WIP. Unlike the recoverAdvancedTriageTasks query this replaces
-      nothing and disables nothing — it is a redundant read, not a dead sweep.
+      FNXC:WorkflowLifecycleColumns 2026-07-30-22:20 (fleet: the union was NOT complete):
+
+      This trio was left on literals with the reasoning that "either way the union is complete" —
+      the merged default lineage supplies cards through `todo`, a legacy workflow through `triage`.
+      That covers two lineages and misses the third: a workflow that RENAMES its lanes matches none
+      of the three literals, so the union came back empty and this sweep had no candidates at all.
+
+      Measured, not inferred: on a live PostgreSQL store with a renamed board seeded alongside a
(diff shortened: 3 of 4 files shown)


gsxdsm:
## Screened the next unclaimed cluster — `async-comments-attachments.ts` (9 guards) — and it is NOT convertible as a guard change

Posting here rather than opening an empty PR, since the useful output is the analysis.

It looked ideal: **9 guards, zero literal queries**, so none of the query-coupling that dominated `self-healing.ts`. All nine are the same shape:

```ts
if (row.column === "archived" || row.deletedAt != null) return "archived";
if (column === null || column === "archived") return [];
```

### Why it cannot be done as a guard conversion

Every one of them resolves through `getLiveTaskColumn(db, taskId, projectId)` — a **raw data-layer helper that takes a `db` handle, not a store**. Role resolution needs `resolveWorkflowIrForTask(store, id)`, so converting means threading a store through:

```
getLiveTaskColumn call sites: 20, across 6 files
  async-comments-attachments.ts 8 · task-id-integrity.ts 4 · task-artifacts-ops.ts 3
  audit-ops.ts 2 · comments-ops.ts 2 · task-mutation-ops.ts 1
```

That is a cross-file signature change to a data-layer primitive, which is not a guard conversion under the fleet rules.

Second reason, independent of the first: it is a **hot read path** — every document read, artifact read, comment fetch and audit write passes through it. `comments-ops.ts` in the same directory documents deliberately *avoiding* IR resolution on that path ("the IR is only resolved for a USER comment, so agent/system comment traffic pays nothing"). Adding a per-call IR resolution would reverse that decision.

### The in-directory precedent supports skipping, not converting

`comments-ops.ts:223` does exactly this conversion — but only where a `store` is already in scope and only on the cold branch. That is the shape that works; `getLiveTaskColumn` has neither property.

### What I am NOT claiming

I am not claiming these nine are harmless. A workflow that renames its archived lane would have documents and artifacts treated as live when they are archived. But the fix is "give the data layer a way to resolve roles", which is a design change for whoever owns `packages/core/src/task-store/`, and it wants doing once for all six files rather than nine times here.

Screened before touching, per the procedure this PR established. Cost: one survey. Previously that same information cost five red tests and two reverts.

gsxdsm:
**Both halves confirmed, and the second is worse than "resolves synchronously" — it cannot resolve at all in production.** Evidence only: this branch is checked out in `workflow-e2e-validation`, so it needs its owner.

## 1. The unused `cache` — confirmed, and it fails the gate

`filterByCompleteRole` declares `cache` and never reads it, while both siblings thread it through:

```ts
filterByWipRole      →  await this.resolveWipColumn(task.id, cache)
filterByReviewRole   →  await this.resolveReviewColumn(task.id, cache)
filterByCompleteRole →  this.completeLaneOf(task.id)          // no cache, no await
```

`@typescript-eslint/no-unused-vars` is an **error** in this repo, so this fails `pnpm lint` and blocks the merge gate for everything queued behind it.

## 2. The sync resolution is a correctness bug, not a performance one

```ts
private completeLaneOf(taskId: string): string {
  try {
    return resolveLifecycleColumns(this.store.resolveTaskWorkflowIrSync(taskId))?.complete ?? "done";
  } catch { return "done"; }
}
```

`resolveTaskWorkflowIrSync` reads `getTaskWorkflowSelectionImpl`, which **returns `undefined` unconditionally** — the code says so in its own comment:

> `packages/core/src/task-store/workflow-definitions.ts:505-511`
> *"Backend mode cannot synchronously read PostgreSQL, so return undefined and let the sync reader (resolveTaskWorkflowIrSync) fall back to its default… sync selection reader is incomplete-PG; use getTaskWorkflowSelectionAsync."*

So on every production board `completeLaneOf` resolves the **default** workflow and answers `"done"`. `filterByCompleteRole` therefore filters for the literal `done` while its two siblings correctly resolve the task's own lane — a converted guard that is inert, sitting between two that work.

That is the same hazard I measured on #2689 and refused to convert around: 14 sync-context guards in `executor.ts` that cannot be converted in place for exactly this reason. Fixing the lint error alone would leave the inert guard behind and remove the signal that something is wrong.

## Suggested shape

Make it match its siblings — an async lane resolver that takes the cache:

```ts
private async resolveCompleteColumn(taskId: string, cache: …): Promise<string> {
  try {
    return resolveLifecycleColumns(await resolveWorkflowIrForTask(this.store, taskId, cache))?.complete ?? "done";
  } catch { return "done"; }
}
```

…then `if (task.column === await this.resolveCompleteColumn(task.id, cache))`. That clears the lint error, uses the cache the signature already promises, and makes the guard actually resolve.

Nothing pushed. Flagging the priority plainly: this is the only red-gate blocker I have found that is not merely a stale dispatch, and every PR behind the gate waits on it.

gsxdsm:
**Still unfixed on the current head (`217927735b`), re-verified just now — and this is my ninth ownership check on this PR.**

`cache` appears exactly **once** in `filterByCompleteRole`: in the parameter list. Never in the body.

```ts
private async filterByCompleteRole(
  tasks: Task[],
  cache: Map<string, Awaited<ReturnType<typeof resolveWorkflowIrForTask>>>,   // ← declared
): Promise<Task[]> {
  const kept: Task[] = [];
  for (const task of tasks) {
    if (task.column === this.completeLaneOf(task.id)) kept.push(task);        // ← never used
  }
  return kept;
}
```

`@typescript-eslint/no-unused-vars` is an **error** in this repo, so `pnpm lint` fails and the merge gate stays red — for every PR queued behind it, not just this one.

The full diagnosis is in my earlier comment on this thread, including the part the finding understates: `completeLaneOf` resolves through `resolveTaskWorkflowIrSync`, whose selection reader `return undefined` unconditionally in PG mode (`workflow-definitions.ts:505-511`). So the helper answers `"done"` on every production board while its two siblings correctly resolve — an inert guard between two working ones. **Fixing only the lint error would silence the signal and leave that behind.**

Suggested shape, repeated here so it is in one place:

```ts
private async resolveCompleteColumn(taskId: string, cache: …): Promise<string> {
  try {
    return resolveLifecycleColumns(await resolveWorkflowIrForTask(this.store, taskId, cache))?.complete ?? "done";
  } catch { return "done"; }
}
// …then: if (task.column === await this.resolveCompleteColumn(task.id, cache))
```

That clears the lint error, uses the cache the signature already promises, and makes the guard actually resolve — matching `filterByWipRole` and `filterByReviewRole` exactly.

**I cannot apply it.** This branch is checked out in `workflow-e2e-validation` and its head has moved twice while I have been checking (`a006b5b356` → `217927735b`), so it has an active owner. Pushing to it would risk discarding their in-flight work, which I nearly did to another worker earlier today and will not repeat.

This needs its owner to apply the fix or an explicit reassignment — it is the only red-gate blocker I have found that is not a stale dispatch.

gsxdsm:
Both halves of this finding are real. I verified them against the branch and against `main`; I cannot push the fix because `fleet-self-healing` is checked out in another worktree (`workflow-e2e-validation`) — this is the 11th dispatch I have taken on it while it was held. Evidence and the patch shape are below so whoever holds it can apply this directly.

**1. The lint error is real, and it fails the merge gate.**

`eslint.config.mjs:379-389` sets `@typescript-eslint/no-unused-vars` to `"error"` with `args: "after-used"` and `argsIgnorePattern: "^_"`. `cache` is the LAST parameter of `filterByCompleteRole` and is never referenced in the body, so `after-used` reports it. Not a style nit — `pnpm lint` is a blocking gate check.

**2. The second half is worse than reported: the converted guard cannot fire.**

`completeLaneOf` (self-healing.ts:3160) resolves through `store.resolveTaskWorkflowIrSync(taskId)`. That path cannot read the task's own workflow in production: `getTaskWorkflowSelectionImpl` (`packages/core/src/task-store/workflow-definitions.ts:505-511`) returns `undefined` unconditionally — "sync selection reader is incomplete-PG; use getTaskWorkflowSelectionAsync". With no selection, `resolveTaskWorkflowIrSyncImpl` falls back to the DEFAULT workflow IR, so `completeLaneOf` answers the default board's complete lane for every task regardless of which workflow the task actually runs.

So on a renamed board `filterByCompleteRole` compares `task.column` against the default lane and keeps nothing. That is the inert-conversion shape: it reads as converted, and behaves exactly like the literal it replaced. Its two siblings (`filterByWipRole`, `filterByReviewRole`) already do this correctly through the async cache.

**Patch shape** — both halves close together, because using the cache is the fix for the second:

```
private async filterByCompleteRole(
  tasks: Task[],
  cache: Map<string, Awaited<ReturnType<typeof resolveWorkflowIrForTask>>>,
): Promise<Task[]> {
  const kept: Task[] = [];
  for (const task of tasks) {
    if (task.column === await this.resolveCompleteColumn(task.id, cache)) kept.push(task);
  }
  return kept;
}
```

mirroring `filterByReviewRole`'s `resolveReviewColumn(task.id, cache)` exactly, with `resolveCompleteColumn` reading `.complete` where its sibling reads the review lane. `completeLaneOf` then has no callers and should go with it — otherwise the next conversion reaches for it.

Do NOT close this by renaming the parameter to `_cache`. That silences the lint error and leaves the guard inert, which is the strictly worse outcome: green lint over a guard that cannot fire.

A test needs to prove the guard fires on a board whose complete lane is not the default's. A mutation check that matters: with `completeLaneOf` in place, such a test fails; that is the signal the fix is real rather than cosmetic.

gsxdsm:
## The review thread is right, and the problem is far larger than the one helper. FLEET-WIDE — batch owners should read this before converting.

`resolveTaskWorkflowIrSync` **cannot resolve a task's workflow at all in this codebase.**

```ts
// packages/core/src/task-store/workflow-definitions.ts:505
export function getTaskWorkflowSelectionImpl(_store, _taskId) {
  /* Backend mode cannot synchronously read PostgreSQL, so return undefined and let the
     sync reader (resolveTaskWorkflowIrSync) fall back to its default. */
  return undefined;   // <- UNCONDITIONAL
}
```

Storage is PostgreSQL-only. So the selection read always returns `undefined`, and
`resolveTaskWorkflowIrSync` always takes its `if (!workflowId)` branch and returns the **DEFAULT**
workflow IR — never the task's own.

### What that means for a conversion

```ts
const lane = resolveLifecycleColumns(this.store.resolveTaskWorkflowIrSync(task.id))?.review ?? "in-review";
if (task.column === lane) { ... }
```

On the default lineage `lane` resolves to `in-review` — same as the literal. On a **renamed** board it
*still* resolves to `in-review`, because the task's own workflow is never consulted. **The guard is a
literal wearing a resolver's clothes.** It passes the census, reads as converted, and behaves exactly
as it did before.

That is precisely the "guard that looks converted and does nothing" failure this program exists to
remove, and I introduced it at scale.

### Scale in this PR

```
resolveTaskWorkflowIrSync uses in self-healing.ts:   16
call sites of my sync lane helpers:                  32
```

`reviewLaneOf` / `completeLaneOf` / `wipLaneOf` / `holdLaneOf` / `archivedLaneOf` are all sync, plus
inline sync resolutions in the `task:moved` listener, `classifyPausedAbortWorkflowRecovery`,
`reconcilePreExecutionWorktrees`, `archiveStaleDoneTasks`, `clearStaleBlockedBy`'s dependency filter,
and `recoverStarvedRefinementTriageTasks`.

**A meaningful share of this PR's ~100 guard conversions are inert on renamed boards.** The async
conversions (`filterByPreWipRole`, `filterByWipRole`, `filterByReviewRole`, `resolveReviewColumn`,
`resolvePreWipColumns`, and every `resolveWorkflowIrForTask` call) are correct and unaffected.

### Why I reached for the sync reader

Roughly half this file's predicates are synchronous `.filter` callbacks and one is a synchronous
`task:moved` listener. I chose the sync reader specifically to avoid restructuring a dozen sweeps into
async loops, and recorded that as the conservative choice. It was the wrong call: the async
restructure was the only correct option, and "it compiles and the census drops" hid it.

### The fix, and why it is not a one-liner

Each sync site needs its lanes **pre-resolved asynchronously into a Map before** the sync callback,
then read from that Map inside it. That is mechanical but touches ~32 sites.

### What batch owners need to know NOW

Any conversion using `resolveTaskWorkflowIrSync` — anywhere in the fleet, not only here — is inert.
If another worker adopted the same helper pattern from this file, those conversions need the same
audit. `resolveWorkflowIrForTask` (async) is the only resolver that actually reads the task's
selection.

I will fix this branch before handing it to batch-engine rather than passing over conversions that do
nothing.

gsxdsm:
Both findings were real. Fixed in `93e0a1e` — and they turned out to be the same bug twice, which is worth recording.

**Root cause, shared.** The sync lane readers (`completeLaneOf`, `reviewLaneOf`) cannot read a task's own workflow: `getTaskWorkflowSelectionImpl` (`packages/core/src/task-store/workflow-definitions.ts:505-511`) returns `undefined` unconditionally — *"sync selection reader is incomplete-PG"*. With no selection, `resolveTaskWorkflowIrSync` falls back to the DEFAULT workflow IR, so every sync lane answer is the default board's lane no matter which workflow the task runs.

**1. `filterByCompleteRole` (coderabbit).** Confirmed on both counts. The lint error is real and blocking: `eslint.config.mjs:379-389` sets `no-unused-vars` to `"error"` with `args: "after-used"`, and `cache` was the unused trailing parameter. But the second half was the serious one — resolving through the sync reader made the filter *inert*, so it read as converted and behaved exactly like the literal it replaced. Fixed by adding the missing async `resolveCompleteColumn` and wiring the filter to it, mirroring `filterByReviewRole` exactly.

Renaming the parameter to `_cache` would have closed the lint error and left the guard inert — a green gate over a guard that cannot fire. Noted at the site so it is not "simplified" back.

**2. `recoverCompletionHandoffLimbo` (greptile).** Exactly as described, and the diagnosis was precise. `filterByReviewRole` admitted tasks through the async cache; the next line re-asked the same question with the sync `reviewLaneOf`. That was not a re-verification, it was a contradiction — two resolvers answering one question with the weaker one deciding. On a renamed board the filter admitted the card and the recheck dropped every one, so completion-handoff limbo was never recovered there.

Removed rather than converted: the filter above had already established review-lane membership, so the lane half was redundant even when it agreed. Only `task.paused` was doing independent work, and it stays.

**Verification (measured):**
- `pnpm test:gate` → 158 + 487 + 10 + 71, all green
- `self-healing.test.ts` → 412 passed
- `pnpm lint` → 0 errors; `pnpm --filter @fusion/engine exec tsc --noEmit` → exit 0
- New: `self-healing-renamed-lane-resolvers.test.ts`, 4 tests

**Mutation-verified, per fix:** restoring `completeLaneOf` in the filter fails 2 of the 4; restoring the `reviewLaneOf` recheck fails 1. Existing sweep tests all run on the default board, where the sync fallback and the real answer coincide — that coincidence is what hid both defects, so the new fixture names its lanes `signoff`/`shipped` where the two answers differ. The suite also pins the paired negatives (a card in the default `done` lane is excluded; a paused card is still skipped).

**Remaining, and deliberately not in this commit:** the same sync readers are still consulted at ~20 other predicates in this file (`reviewLaneOf` at 3717, 7325, 7699, 8035, 8439, 8546, 8567, 8696, 8875, 9613, 9802, 10003, 10181, 10495; `completeLaneOf` at 4816, 7531, 9388; plus `wipLaneOf`/`holdLaneOf`/`archivedLaneOf`). Every one is inert for the same reason. Converting them is a much larger change with real behaviour risk per site — each needs its own async plumbing and its own regression test — so it is flagged here rather than swept in silently under a review fix. The two in this commit are the two that were reported.

gsxdsm:
Rebased onto `origin/main` and pushed. This one needed judgement rather than a mechanical replay, so here is what was decided and why.

**`main` and this branch had both converted `self-healing.ts`, independently.** `batch-engine` landed its own conversion of the same file while this PR was open, so 6 of the 28 commits conflicted on the same guards.

Measured before resolving, because "superseded" was the obvious guess and it was wrong: `self-healing.ts` carries **107** census guards on `origin/main` and **8** on this branch. This branch is far ahead — it is not superseded, and skipping its commits would have thrown away 99 conversions.

**Resolution rule, applied per hunk:** prefer whichever side is actually converted; when both are, take `main`'s, since it has landed and uses the better shape. `main`'s `resolveReviewColumnsFor` returns a **membership set**, which is the correct arity for the "is this card still in review?" questions these sweeps ask — a board may declare a merge-orchestration lane and a separate human sign-off lane, and a single-id answer silently ignores the second. Taking HEAD blindly would have been wrong in the other direction: where `main` still holds the literal, HEAD *is* the literal, and that would have dropped this branch's conversion.

Two hunks needed hand-merging rather than a pick:

1. **Both review helpers are kept.** `main`'s `resolveReviewColumnsFor` (membership) and this branch's `resolveReviewColumn` (single id) now coexist, because ~99 converted call sites here depend on the latter. Deleting it to "clean up" would break them; migrating those call sites to the set is real follow-up work, not rebase work.
2. **A dangling `wedgedReviewLane`.** `recoverWedgedActiveMerge` used the resolved lane twice, but only the first use sat inside a conflict hunk. Taking `main`'s version there left the second use referencing a variable that no longer existed — caught by `tsc`, not by the merge. It now reads from the same `wedgedReviewColumns` set the guard above resolves, which also fixes its arity.

**Verification on the rebased branch (measured):**
- `pnpm test:gate` → 158 + 487 + 10 + 71, all green
- `self-healing.test.ts` + the new renamed-lane suite → **416 passed**
- `pnpm lint` → 0 errors; `tsc --noEmit` → 0 errors; zero conflict markers in the tree
- census total **300**, `self-healing.ts` at **8**

Pushed with `--force-with-lease`. The two review findings fixed earlier in this PR (`filterByCompleteRole`, `recoverCompletionHandoffLimbo`) survived the rebase and are still mutation-covered by `self-healing-renamed-lane-resolvers.test.ts`.

gsxdsm:
## The review finding is fixed, and it was bigger than the one helper

Thank you for it — this was the most consequential catch on this PR by a wide margin.

### What was wrong

`filterByCompleteRole` was the visible symptom; the disease was `resolveTaskWorkflowIrSync`:

```ts
// workflow-definitions.ts:505
export function getTaskWorkflowSelectionImpl(_store, _taskId) {
  /* Backend mode cannot synchronously read PostgreSQL... */
  return undefined;   // UNCONDITIONAL
}
```

Storage is PostgreSQL-only, so that reader **always** falls through to the DEFAULT workflow IR. Every guard built on it resolved to the default lineage's lane — byte-identical to the literal it replaced. On a renamed board the conversion did nothing, while passing tsc, eslint, the census and the full suite.

### Fixed

| | |
|---|---|
| `filterByCompleteRole` | async via `resolveWorkflowIrForTask`, now uses its `cache` (also clears the ESLint gate failure) |
| five lane helpers | async through the selection-aware resolver; 19 call sites simply `await` |
| 16 sites | **deleted** — redundant re-checks of a role their sweep had already filtered by |
| 12 sites | **reverted to literals + flagged** — sync `.filter` callbacks with no async resolution reachable |
| 3 closures | removed once dead; ESLint would have failed the gate on all three |

Every `resolveTaskWorkflowIrSync` **use** is gone; the remaining hits are comments explaining why it must not be used.

### The count went UP, and that is the honest number

```
guards  8 (claimed, ~15 inert)  ->  30 (real)
```

Nothing regressed. The census now measures conversions that actually consult the task's workflow.

### Three bugs my repair pass introduced, caught by lint before pushing

A bare `return` before a newline (ASI would have returned `undefined`), an unconditionally-true `true || noCommitsError`, and a dangling leading `&&`. All from regex edits over a 15k-line file — the same lesson as earlier in this PR, now applied to removals as well as replacements.

### Verified

`engine-default` **5 failed | 9758 passed — identical to merge-base** `86a2b48968`, test for test; `engine-core` 487; tsc 0; eslint 0; census gate 0.

`reconcilePreExecutionWorktrees` is flagged at its site as the highest-value remaining restructure: it fails in the **releasing** direction on a renamed board, so an active card's worktree can look parked and become eligible for release.

gsxdsm:
Coverage added on top of the sync-resolver fix — `c048e17b3`.

Another worker landed the broader correction (making the lane readers async and threading the cache) while I was working the same two findings, so I reset to their commit rather than force-pushing my version over it. Their fix is wider than mine was: 15 conversions rather than 2. What it did not carry was a test that can tell an inert filter from a working one, which is what let this class survive review in the first place.

`self-healing-complete-lane-renamed.test.ts` is that ratchet. Two cases on a board whose complete lane is `shipped`: the card there is kept, and a card sitting in the default `done` lane the board does not declare is excluded (without the negative, "keep everything" also passes).

**Mutation-verified:** reverting `filterByCompleteRole` to `resolveTaskWorkflowIrSync` fails **both** cases. `tsc --noEmit` 0 errors, `pnpm lint` 0 errors.

Why no existing test caught this: every sweep test in that neighbourhood runs on the DEFAULT board, where the sync fallback and the correct answer are the same string. That coincidence is the actual defect-hider here, not any individual missing assertion — so the fixture's value is that the two answers differ.

**One thing I checked and did not change.** The second finding's site, `recoverCompletionHandoffLimbo`, re-asks the review lane after `filterByReviewRole` already established it. With `reviewLaneOf` now async, both resolvers return the same narrow lane, so this is no longer a defect — it is redundant work, not a wrong answer, and removing it is a behaviour-neutral cleanup rather than a review fix. Flagging rather than folding it in.

**Also worth a look before merge, not fixed here:** several of the new FNXC stamps in `self-healing.ts` read `2026-08-03`, which is future-dated relative to today (2026-07-30). The same thing was flagged on #2771 and corrected there.

gsxdsm:
Closing as **superseded by batch-engine**, per the mega-batch directive: `packages/engine` is the capacity worker's, and main already carries their conversions in this file.

The work is preserved on **`fleet-self-healing-handover`** (29 commits) rather than discarded. Cherry-pick or ignore as suits the batch — but please read the first finding before converting anything else in this file.

## 1. CRITICAL — `resolveTaskWorkflowIrSync` produces inert conversions. Audit the whole fleet.

```ts
// workflow-definitions.ts:505
export function getTaskWorkflowSelectionImpl(_store, _taskId) {
  /* Backend mode cannot synchronously read PostgreSQL... */
  return undefined;   // UNCONDITIONAL
}
```

Storage is PostgreSQL-only, so that reader **always** returns the DEFAULT workflow IR. A guard built on it resolves to the default lineage's lane — byte-identical to the literal it replaced. It passes tsc, eslint, the census and the full suite while doing nothing on a renamed board.

I shipped ~15 of these before review caught it. **Main currently has zero uses — keep it that way.** `resolveWorkflowIrForTask` (async) is the only resolver that reads the task's selection.

## 2. Guard and source query are ONE unit

`listTasks({ column: "in-review" })` feeding a converted predicate leaves the sweep blind: the query returns nothing on a renamed board, so the corrected guard never sees a candidate. Converted to the eye, dead in fact. `self-healing.ts:3015` documents the same defect from #2560.

## 3. A sweep's query is convertible exactly when its test stubs allow it

Three properties, all checkable before you touch the code — I learned each by breaking something:

1. does the stub treat `column` as **optional**? (`opts?.column ? filter : all`)
2. does the suite use **`mockResolvedValueOnce` chains**? (coupled to call ORDER)
3. does it **assert the exact `listTasks` argument**? (coupled to SHAPE)

Fail any one and the conversion starves the sweep. Ten sweeps in this file fail at least one; each is flagged at its site on the handover branch with the differential evidence.

## 4. Verify against the MERGE-BASE, not `origin/main`

`git stash` does not revert commits, so "stash and compare" compares your branch against itself and is blind to anything you have already committed. I had a regression live for several commits behind exactly that. Main also runs a different test count than an older branch, so a direct comparison is meaningless.

## 5. Highest-value single restructure left

`reconcilePreExecutionWorktrees` fails in the **releasing** direction on a renamed board — an active card's worktree looks parked and becomes eligible for release. Its predicate is a sync `.filter`, so it needs an async loop or a pre-resolved lane map.

## State of the handover branch

```
guards   110 -> 30      queries  49 -> 21
engine-default 5 failed | 9758 passed — identical to its merge-base, test for test
engine-core 487 · tsc 0 · eslint 0 · census gate 0
```

The count is 30 rather than the 8 I reported earlier because the inert conversions are now counted honestly as unconverted.

PR #2688, opened 2026-07-30
fleet: TaskCard.tsx 42 -> 0

gsxdsm:
## Census before / after

| | before | after |
|---|---:|---:|
| `packages/dashboard/app/components/TaskCard.tsx` | **42** | **0** (absent from the census) |
| repo backlog | 722 | **680** |

The backlog dropped by exactly **42** — the converted count, nothing else moved.

## Shape of the conversion

Existing role helpers only. The card already receives `taskColumnFlags`, so every in-component site is a genuine trait read with the documented id fallback for the pre-load window — not a rename.

Four canonical role bindings at the top of the component, and 37 scattered comparisons collapsed onto them:

- `isDoneColumn` / `isArchived` already existed as bindings → converted in place
- `isIntakeColumn` / `isHoldColumn` were already flags-first but written **inline** → routed through the helpers, so the file has one definition of each role
- `isWipColumn` / `isReviewColumn` are new bindings that 9 `in-progress` and 7 `in-review` comparisons collapse onto

Stacked on **#2685**, which adds the `complete`/`archived`/`wip`/`review` helpers — 39 of these 42 guards had no helper to convert to before it.

## Two sites handled differently — stated, not absorbed into the count

**1. `getInReviewCompletionMs` (~line 387) — centralisation, not trait resolution.** It is a module-scope helper taking only `task`, so no flags are in scope. Passing `undefined` selects the shared fallback, which is behaviour-identical to the literal; it centralises the id and leaves the site greppable as *"still needs its flags threaded"* rather than looking already converted. Threading them is a signature change across the card's duration helpers — a separate unit of work. Labelled as such at the site.

**2. `label: column === "done"` — marked `DELIBERATE-LITERAL`, because it is not a lifecycle guard.** `column` there is the loop variable over the hardcoded `["done", "triage"]` array two lines above, so it asks *"which entry of my own literal list am I on"* in order to pick a label — not *"what role does this card's column play"*. Resolving a trait for a string this code just wrote itself would be meaningless.

The array **is** worth revisiting — it names move targets by id — but which targets a review card should offer is a behaviour question, and behaviour changes are out of scope for the fleet. Flagged rather than guessed.

## Behaviour preserved

`TaskCard` suites are **3 failed / 496 passed both with and without this change**, verified by stashing and re-running rather than by reasoning. All three are pre-existing: two CSS assertions (`'1px' to be 'medium'`, padding) and one mission badge.

`pnpm test:gate` green (10 / 158 / 487 / 71). `pnpm check:lifecycle-columns` exits 0 with the baseline re-recorded in this PR. `tsc -p tsconfig.app.json` clean. `pnpm lint` clean.

## Summary by CodeRabbit

* **New Features**
  * Added support for custom workflow columns across task cards, including completion, archive, work-in-progress, and review states.
  * Updated task actions, status indicators, progress details, file-change counts, and move options to reflect workflow roles consistently.
  * Preserved legacy behavior when custom role information is unavailable.

* **Tests**
  * Expanded coverage for terminal and in-progress workflow role detection.

* **Chores**
  * Updated lifecycle column census baselines.

### DIFF
--- a/packages/dashboard/app/__tests__/columnRoles.test.ts
+++ b/packages/dashboard/app/__tests__/columnRoles.test.ts
@@ -151,6 +151,14 @@ describe("terminal and mid-flight column roles", () => {
     is finished but not COMPLETED. Surfaces that count throughput would double-count it otherwise.
     */
     expect(isCompleteColumnRole({ archived: true }, "archived")).toBe(false);
+    /*
+    FNXC:ColumnRoles 2026-07-30-16:05 (PR #2688 review — coderabbit):
+    RESOLVED METADATA WINS, asserted against the MATCHING legacy id. Every other false case here
+    passes a non-matching id, so a `trait || legacyId` implementation satisfies them all — the id
+    is false too, and the bug hides. Only a board that renamed the ROLE while keeping the legacy
+    NAME can tell the two apart.
+    */
+    expect(isCompleteColumnRole({ archived: true }, "done")).toBe(false);
     expect(isCompleteColumnRole(undefined, "done")).toBe(true);
     expect(isCompleteColumnRole(undefined, "shipped")).toBe(false);
   });
@@ -158,6 +166,7 @@ describe("terminal and mid-flight column roles", () => {
   it("isArchivedColumnRole reads the archived trait", () => {
     expect(isArchivedColumnRole({ archived: true }, "cold-storage")).toBe(true);
     expect(isArchivedColumnRole({ complete: true }, "done")).toBe(false);
+    expect(isArchivedColumnRole({ complete: true }, "archived")).toBe(false);
     expect(isArchivedColumnRole(undefined, "archived")).toBe(true);
     expect(isArchivedColumnRole(undefined, "cold-storage")).toBe(false);
   });
@@ -170,6 +179,7 @@ describe("terminal and mid-flight column roles", () => {
     */
     expect(isWipColumnRole({ countsTowardWip: true }, "building")).toBe(true);
     expect(isWipColumnRole({ intake: true }, "backlog")).toBe(false);
+    expect(isWipColumnRole({ intake: true }, "in-progress")).toBe(false);
     expect(isWipColumnRole(undefined, "in-progress")).toBe(true);
     expect(isWipColumnRole(undefined, "building")).toBe(false);
   });
@@ -180,6 +190,7 @@ describe("terminal and mid-flight column roles", () => {
     expect(isReviewColumnRole({ mergeBlocker: true }, "signoff")).toBe(true);
     expect(isReviewColumnRole({ humanReview: true }, "signoff")).toBe(true);
     expect(isReviewColumnRole({ countsTowardWip: true }, "building")).toBe(false);
+    expect(isReviewColumnRole({ countsTowardWip: true }, "in-review")).toBe(false);
     expect(isReviewColumnRole(undefined, "in-review")).toBe(true);
     expect(isReviewColumnRole(undefined, "signoff")).toBe(false);
   });
--- a/packages/dashboard/app/components/TaskCard.tsx
+++ b/packages/dashboard/app/components/TaskCard.tsx
@@ -34,7 +34,15 @@ import { useTaskDiffStats } from "../hooks/useTaskDiffStats";
 import { useAgentsMapCache } from "../hooks/useAgentsMapCache";
 import { useLiveTimeTicker } from "../hooks/useLiveTimeTicker";
 import { isTaskStuck } from "../utils/taskStuck";
-import { isFieldEditableColumnRole } from "../utils/columnRoles";
+import {
+  isArchivedColumnRole,
+  isCompleteColumnRole,
+  isHoldColumnRole,
+  isIntakeColumnRole,
+  isFieldEditableColumnRole,
+  isReviewColumnRole,
+  isWipColumnRole,
+} from "../utils/columnRoles";
 import { hasPendingAutomaticRecovery, isTaskManuallyRetryable } from "../utils/taskRecovery";
 import { getRevertOfId, isTaskReverted } from "../utils/taskRevert";
 import { getStalledReviewSignal } from "../utils/taskStalledReview";
@@ -376,7 +384,18 @@ function getTaskEndToEndDurationMs(task: Task, nowMs: number): number | null {
 }
 
 function getInReviewCompletionMs(task: Task): number | null {
-  return task.column === "done" ? getDoneCompletionMs(task) : null;
+  /*
+  FNXC:WorkflowResolvedColumns 2026-07-30-20:30 (fleet: TaskCard.tsx):
+  CENTRALISED, not trait-resolved — stated plainly because the difference matters. This is a
+  module-scope helper taking only `task`, so no column flags are in scope; passing `undefined`
+  selects the shared fallback and yields behaviour identical to the literal it replaces.
+
+  What it buys is that the legacy id lives in ONE place, and that this site is greppable as "still
+  needs its flags threaded" rather than looking already converted. Threading them means giving every
+  caller of this helper a flags argument, which is a signature change across the card's duration
+  helpers — a separate unit of work, not a vocabulary edit.
+  */
+  return isCompleteColumnRole(undefined, task.column) ? getDoneCompletionMs(task) : null;
 }
 
 function getMergeElapsedMs(task: Task, nowMs: number): number | null {
@@ -1009,16 +1028,24 @@ function TaskCardComponent({
   and in that state there is nothing to resolve FROM. Deleting it does not remove a guard, it picks a
   different guess ("not intake") and silently drops planning affordances during first paint.
   */
-  const isIntakeColumn = taskColumnFlags
-    ? taskColumnFlags.intake === true
-    : task.column === "triage";
-  const isHoldColumn = taskColumnFlags
-    ? taskColumnFlags.hold === true
-    : task.column === "todo";
+  /*
+  FNXC:WorkflowResolvedColumns 2026-07-30-20:15 (fleet: TaskCard.tsx):
+  The card's ROLES, all four resolved the same way — traits when the board has loaded, the documented
+  legacy id only for the pre-load window and for a card in a column its workflow no longer declares.
+
+  The intake/hold pair were already flags-first but written inline; routed through the helpers so this
+  file has ONE definition of each role rather than an expression per site. The wip/review pair are new
+  bindings for the same reason: nine `in-progress` and seven `in-review` comparisons below collapse
+  onto them.
+  */
+  const isIntakeColumn = isIntakeColumnRole(taskColumnFlags, task.column);
+  const isHoldColumn = isHoldColumnRole(taskColumnFlags, task.column);
+  const isWipColumn = isWipColumnRole(taskColumnFlags, task.column);
+  const isReviewColumn = isReviewColumnRole(taskColumnFlags, task.column);
 
   const [isSaving, setIsSaving] = useState(false);
   const [showSteps, setShowSteps] = useState(
-    task.column === "in-progress" ||
+    isWipColumn ||
     (isIntakeColumn && task.steps.some(s => s.status === "done" || s.status === "skipped"))
   );
   const [missionTitle, setMissionTitle] = useState<string | null>(null);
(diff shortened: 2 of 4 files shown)


gsxdsm:
Both real, and both are consequences of the conversion rather than pre-existing bugs — worth stating precisely, because it changes who should be looking for them elsewhere.

`task.column` is **stable across renders**, so reading it once was safe. A trait-derived role boolean is **not** stable: it flips when the workflows fetch resolves. Every conversion in this fleet turns a stable read into an unstable one, so both of these failure shapes are now possible in any converted file.

**P1 — four under-keyed memos.** Enumerated with an AST pass rather than by eye: lines 1683, 1713, 1795, 1842. Each dep array now lists the roles its body actually reads. On a custom board they were holding first-paint values — stale timers and labels on WIP cards, missing completion dates on terminal cards.

Worth flagging for the fleet generally: **this repo has no `react-hooks/exhaustive-deps` rule**, so nothing catches this class automatically. Any worker converting a component file needs to re-check dep arrays by hand or with a pass like the one above.

**P2 — a frozen `useState` initializer**, and the sequence is slightly worse than described. The progress section is *itself* gated on `isWipColumn`, so on a custom board it did not merely start collapsed — it **appeared later, already collapsed**, on a running card the operator never touched.

Reconciled one-way and only while untouched: the effect opens the section when the card turns out to be WIP, and `stepsTouchedRef` stops it the moment the operator toggles. I deliberately did **not** auto-collapse on the reverse transition — closing a section someone is reading is worse than the bug being fixed.

**Tests assert the toggle's `aria-expanded`, not step text.** That attribute *is* the section's state, so the case cannot pass because some unrelated markup happened to render. Mutation-verified: removing the effect fails with `expected 'false' to be 'true'` — the section arrives collapsed, exactly as you described.

TaskCard suites: **3 failed / 498 passed**. The 3 are the same pre-existing failures as before this PR (two CSS assertions, one mission badge); the delta is **+2 passing**. `pnpm lint` clean, `pnpm check:lifecycle-columns` exits 0, `tsc -p tsconfig.app.json` clean.

Census unchanged by this commit — it is a correctness fix on top of the 42 → 0 conversion, not a further conversion.

gsxdsm:
**Both findings are real, and one is broader than reported.** I verified them read-only rather than pushing — this branch is checked out in another worktree (`u12-flag-cleanup`), so it is yours to commit to; two workers on one branch is how a force-push lost work earlier in this program.

## Why this class exists at all

The conversion changed a **synchronous** derivation into one that depends on **late-arriving** data. `task.column === "in-progress"` was correct on first render. `isWipColumnRole(taskColumnFlags, task.column)` is not: `getTaskColumnFlags` (Column.tsx:320) returns `undefined` until `taskContextMenuColumnsByTaskId` / `workflowContextMenuColumns` resolve, so the first render of a **custom** column gets the legacy fallback — which is `false` for any column not literally named `in-progress`.

Every consumer that assumed synchronous correctness is now a latent staleness bug. That is the framing I would put in the PR body, because it is not specific to this file.

## Finding 2 — `useState` initializer (line 1047)

Real, and the worst of the two, because a `useState` initializer runs **once**:

```ts
const [showSteps, setShowSteps] = useState(
  isWipColumn || (isIntakeColumn && task.steps.some(...))
);
```

On a custom WIP column the first render records the fallback's `false` and **never reconciles**. The progress section appears when flags arrive (it is computed inline at 1682), but the steps stay collapsed for the life of the card.

## Finding 1 — stale memos, and it is 5 sites, not 1

I enumerated every hook reading a role boolean whose dep array lists neither `taskColumnFlags` nor the boolean itself:

| line | hook | reads |
|---|---|---|
| 1653 | `useMemo` | `isWipColumn`, `isReviewColumn` |
| 1713 | `useMemo` | `isWipColumn` |
| 1795 | `useMemo` | `isDoneColumn`, `isArchived` |
| 1838 | `useMemo` | `isDoneColumn` |
| 2619 | `useMemo` | `isReviewColumn`, `isDoneColumn`, `isArchived` |

So the live-time indicator, the duration/lifecycle displays and the action-menu model can all render against pre-resolution role values on a custom board.

## Why lint did not catch it

**This repo configures no `react-hooks/exhaustive-deps` rule** — I checked `eslint.config.mjs`. So there is no automated backstop for this class anywhere in the dashboard, and adding a disable directive for it actually *fails* CI lint. Every one of these has to be caught by review or by a test that renders before flags resolve.

## Suggested shape (yours to take or leave)

Add `taskColumnFlags` to the five dep arrays, and make `showSteps` reconcile when the roles change rather than capturing them. A test that renders with `taskColumnFlags={undefined}` and then supplies flags would pin all six at once — and would fail today.

Since none of the other fleet conversions can rely on lint here either, this is worth stating in the PR body as a rule for the phase: **a role boolean is late-arriving data; anything that memoizes or captures it must depend on it.**

gsxdsm:
Real, and the strongest kind of finding: the tests could not fail on the bug they exist to prevent. Fixed.

Every false case passed a **non-matching** column id, so the legacy half of the comparison was false too. A `trait || legacyId` implementation — precisely what flags-first ordering exists to rule out — satisfies all of them.

**Measured both directions rather than argued:**

| | result |
|---|---|
| `trait \|\| legacyId` bug applied, **without** these assertions | **14 of 14 pass** |
| `trait \|\| legacyId` bug applied, **with** these assertions | **3 fail** |

Only a board that renamed the *role* while keeping the legacy *name* can tell the two apart, which is what each added case supplies:

```ts
isCompleteColumnRole({ archived: true },      "done")        -> false
isArchivedColumnRole({ complete: true },      "archived")    -> false
isWipColumnRole({ intake: true },             "in-progress") -> false
isReviewColumnRole({ countsTowardWip: true }, "in-review")   -> false
```

(Three fail rather than four because `isReviewColumnRole` reads two traits and my mutation covered the three single-trait helpers.)

Verified: 14 tests green, dashboard `tsc` clean (`tsconfig.app.json`), lint clean, gate green (487 + 158 + 10 + 71).

Note on ownership: I held off on this branch earlier while it was checked out in another worktree, and picked it up once that released. The earlier review comment on the stale-memo/`useState` findings still stands on its own.

gsxdsm:
Right, and it is the same one-sided-test defect I have been flagging in other people's work — in mine this time.

All four false cases passed a **non-matching** id (`isCompleteColumnRole(undefined, "shipped")`). A broken `trait || legacyId` implementation satisfies those just as well: no trait **and** no id match, so `false` either way. They prove the fallback fires; they prove nothing about **precedence**.

The discriminating shape is the one you gave — trait `false` against a **matching** id. Flags-first returns `false`, an OR returns `true`, and that is the case that matters in production: a resolved column whose trait is explicitly off must not be overridden by its name.

**Mutation-verified:** rewriting `isCompleteColumnRole` as `flags?.complete === true || columnId === LEGACY_COMPLETE_COLUMN_ID` now fails with `expected true to be false`. Under my previous cases it passed.

Landed as **#2694** rather than here: the helper set merged in #2685, so the test file is on `main`, and this branch is checked out in another worktree.

Worth generalising for the fleet, and I have put it in that PR: **any converted site tested only with "no flags, non-matching id" proves the fallback and nothing about precedence.** The helpers were already correct — only the evidence was weak, which is the harder version to catch, because everything is green and the assertion looks thorough.

gsxdsm:
Real, and **narrower than stated** — recording that distinction because I nearly claimed a fix for an unreachable path.

**The gap is real.** The initializer opens the steps section for a WIP card **or** for an intake card that already has finished steps. The reconciliation effect re-checked only `isWipColumn`, so a renamed intake lane hit exactly the bug that effect was added to fix, one arm over. Fixed, and the effect stays one-way and touched-ref guarded.

**But the reachable surface is smaller than the finding implies.** The steps toggle renders inside `showProgressSection`, which requires `task.status === "executing" || isWipColumn` — and FN-7676 *deliberately* hides the steps breakdown for Planning cards. So a renamed intake card **at rest never shows the section at all**, collapsed or not. The reachable case is intake **and** executing: a Plan Review running in the Planning lane, the FN-7831 shape.

My first version of the test omitted `status: "executing"` and failed by not finding the toggle at all. That failure could easily have been read as *"the fix is wrong"* — or, had I written the assertion more loosely, as *"the finding is bogus"*. Both readings would have been wrong. The path is real, just narrow, and it took chasing the failure to the render gate to tell which.

**Mutation:** reverting the effect to `isWipColumn` alone fails the new test while the existing WIP case still passes — which is precisely how this gap survived the first fix.

Verified after rebasing onto current main: `TaskCard.test.tsx` + `columnRoles.test.ts` 403 green, dashboard `tsc` clean (`tsconfig.app.json`), lint clean, gate green (487 + 158 + 10 + 71).

gsxdsm:
Right, and **why it escaped matters more than the fix**.

`contextMenuActions` reads `isDoneColumn` / `isArchived` / `isReviewColumn` and listed none of them, so on a custom complete, archived or review column the menu kept first-paint actions — Archive, Unarchive, Revert and review-column moves stale until an unrelated re-render.

I reported running *"an AST pass over every `useMemo`/`useEffect`/`useCallback`"* on this file and fixing four. **That pass was a regex over source**, matching only hooks written `useMemo(() => {` with a `}, [` dep array on its own line. This memo is formatted differently and was invisible to it — and the same regex also declared `TaskDetailModal` clean, where it had likewise missed one.

That is exactly the tool failure this programme has been removing everywhere else: a pattern-matcher reporting completeness it cannot deliver, and me quoting its output as if it were measurement. It is the same mistake as the greps that gave three different answers for the `triage` census, made by the person who wrote the AST census to replace them.

Replaced with a real `ts.createSourceFile` walk that collects every identifier in the callback and every identifier in the dep array, so formatting cannot hide a hook.

| file | regex said | AST found |
|---|---|---|
| `TaskCard.tsx` | clean (after 4 fixes) | **1 under-keyed** |
| `TaskDetailModal.tsx` | clean | **1 under-keyed** |

Both now 0. The TaskDetailModal one is fixed on its own branch (#2698) rather than smuggled in here.

TaskCard suites: 3 failed / 499 passed — the 3 are the same pre-existing failures as before this PR (two CSS assertions, one mission badge). `pnpm lint` clean, `pnpm check:lifecycle-columns` exits 0, `tsc -p tsconfig.app.json` clean.

gsxdsm:
**The finding is correct as written, and I established it is currently unreachable — by trying to write the test that would fail.**

`contextMenuActions` does read `isDoneColumn` / `isArchived` / `isReviewColumn` and did omit all three from its dependency array. The three deps are added.

**But removing them again leaves all 389 existing cases green *and* the new one green.** The reason: `taskActionMenuModel` already lists `taskColumnFlags` among its own deps, so its identity changes when the flags land and this memo recomputes **transitively**. There is no state where the roles change without that dep changing too.

I nearly shipped this as "fixed and covered." The test I wrote passes either way, and had I not mutated the production line I would have recorded a verification I had not performed — the fourth time that specific mistake would have landed in this program.

**What the test I kept actually does:** it pins the operator-visible behaviour — a card whose flags land late ends up with a menu built from the resolved roles, which on a renamed complete column is the difference between having an **Archive** action and having none. Its header now says explicitly that it does *not* discriminate the dependency-array line, rather than implying coverage it lacks.

**Why keep the deps at all**, given they change nothing today: they make the memo correct on its own terms instead of correct by accident of a neighbour's dependency list. That accident is exactly what a later refactor of `taskActionMenuModel` would silently remove — and this test is what would catch that.

Worth noting this is the **third** finding of this class in this one file, and the repo configures no `react-hooks/exhaustive-deps` rule, so nothing catches the class automatically. My earlier comment here enumerated five such hooks; this was one of them.

Verified: 390 tests green, dashboard `tsc` clean (`tsconfig.app.json`), lint clean, gate green (487 + 158 + 10 + 71).

gsxdsm:
Superseded by #2726, which landed TaskCard.tsx 42 → 3 on the same file while this was open.

Rebasing this would mean resolving 15 conflicts to re-derive work already on main — the two PRs made the same conversions under different local names (`isDoneColumn`/`isArchived` here, `isCompleteColumn`/`isArchivedColumn` there). Not worth the merge risk for a file that is already 93% converted.

The three sites still standing on main are real and I am carrying them forward separately rather than losing them:

- `L392` `task.column === "done" ? getDoneCompletionMs(task) : null` — convertible to the complete role.
- `L1030` `: task.column === "todo"` — the intake/hold question.
- `L2735` `label: column === "done"` — **not** convertible: `column` is a loop variable iterating a literal `["done", "triage"]` array, so this is a DELIBERATE-LITERAL mark, not a rename.

No work is lost by closing this; the remaining delta is 3 sites, ~2 of them convertible.

PR #2691, opened 2026-07-30
fleet: executor.ts 85 → 15 (70 guards; 3 live defects incl. the reverse half-conversion) — DUPLICATE of #2689, see comment

gsxdsm:
**Claim: `packages/engine/src/executor.ts` — the largest unclaimed file on the work order (85 guards).** Converted the graph-failure recovery family: **85 → 49**, 36 guards, three commits.

Not 0, and I am not claiming 0. The remaining 49 are enumerated at the bottom with what each needs; two of them I am flagging as **do-not-convert-blind** rather than guessing.

## Census

| file | before | after |
|---|---:|---:|
| `packages/engine/src/executor.ts` | 85 | **49** |

Repo backlog 722 → **686**. Baseline re-recorded in the same commits as the conversions (`--strict` exits 0).

## Converted per METHOD, not per literal

Every method I touched resolves **one lane snapshot** and answers every lifecycle question in it from that snapshot. That is not tidiness: this file's worst defects are methods that resolved lanes for *part* of a decision and used literals for the rest.

Helpers used — **no new abstraction**: the existing `resolveTerminalColumnsFor` (in-file, union'd with the legacy ids), `resolveResumeLanes` (in-file, caller-owned memo), and `resolvePlannerLanes` (the sync resolver, for the one synchronous classifier).

## Two live defects, not vocabulary cleanups

**1. `isReentrantPausedAbortedInFlightNode` — FN-7214's invariant was already broken.** The method resolved lanes at the **end**, for its return value, while its four `in-review` eligibility gates were literals. On a renamed board those gates all read false, so a review card skipped the global-pause recheck, the `autoMerge === false` refusal, the shared-branch-member arbitration **and** the merge-confirmed refusal — then the lane-resolved final line answered *"re-entrant"*. The method's own comment says an auto-merge-off review row must stay terminal.

**2. `routeUnusableWorktreeGraphFailureToRecovery` — FN-5147's gate was skipped.** It read not-finished **and** not-in-review, so recovery proceeded with the auto-merge-off gate bypassed: an automatic recovery moving a human-review-terminal card backward.

Both are the half-conversion shape this program keeps paying for — **a gate reading the wrong board does not go quiet, it admits what the gate existed to refuse.**

**3. `handleGraphFailure`'s completion suppression.** `alreadyFinalizedToReview` and `suppressFinalizedCompletionAbort` keyed on `column !== "in-progress"`, so on a renamed board a completed, already-finalized row read as still-in-wip: FN-6644/FN-6647's suppression never fired and the row was re-parked as an operator-action pause abort — precisely the durability gap those tickets closed.

## Revert proof

`executor-graph-failure-lanes-resolved.test.ts`, on a board sharing **no** column id with the default lineage (on the default board these guards are right by coincidence — the literals *are* the board, so a default-board test proves nothing).

**4 of 8 cases go red with the literals restored.** The 4 that stay green are labelled as such **in the file**: two paired positives that must not redden, the default-board no-change case, and one case where the terminal guard is genuinely redundant with the method's final lane check. I would rather label a case as non-evidence than count it.

Two fixture corrections recorded at the sites, both my own assertions not touching the behaviour they claimed:
- asserting the router's **return value** passed with the literals restored — the pre-fix path ran the whole recovery and returned false for an unrelated reason. Spying on `recoverMissingWorktreeSessionStartFailure` asks what the guard actually decides.
- `allowsAutoMergeProcessing` keys on the **global** setting, not `task.autoMerge`, so my first FN-5147 fixture never engaged the gate. Corrected the fixture, not the assertion.

## One decision worth review

The merge-confirmed finalize guard resolves the **complete** column via `resolvePlannerLanes` and deliberately **not** the terminal union: the union includes the legacy ids, so a card in a column merely *named* `done` would read as already finalized and the finalize would be skipped. A board declaring no complete column keeps the literal — refusing there would strand a merge-confirmed card, and `finalizeMergeConfirmedWorkflowGraphTask` resolves its own target anyway.

Also noted at a site: `resolveResumeLanes` substitutes the legacy `in-review` when a board declares no review lane, while `PlannerLanes` leaves forward lanes undefined so callers refuse instead of inventing a column. The two resolvers disagree on that one case. I used each per its own contract rather than papering over it; merging them is already an open follow-up in the source.

## The remaining 49, with what each needs

- **`to`/`from` move-effect parameters** (`3455`, `3479`, `3530`, `3540`, `15882`, `14565`, `4060`) — these compare a *move's* endpoints, not a card's resting column, and the move effects are trait-hook territory. **Flagged, not guessed.**
- **wip-lane liveness guards** (`2409`, `3340`, `4254`, `4456`, `5777`, `12356`, `12365`, `12373`, `14122`, `20131`, `20907`, `16869`) — mechanical, same `resolveResumeLanes` pattern; the only cost is one IR read per guard on hot paths, which wants a per-poll memo rather than a per-guard resolution.
- **hold/intake guards** (`4395`, `14589`, `16856`, `13316`, `15276`, `15328`) — same, plus `13316`/`15276`/`15328` mix wip **and** hold in one expression, so they need the snapshot form.
- **review-lane guards** (`2339`, `4355`, `4611`, `4635`, `5504`, `11796`) — `11796` is a three-literal expression across review/hold and is the highest-risk single site left in the file.
- **complete-column guards** (`6401`, `7881`, `9029`, `5501`, `5502`) — same decision as the finalize guard above, so the same union-vs-complete distinction applies.
- **`12406`, the dependency guard** (`dep.column !== "done" && !== "in-review" && !== "archived"`) — asks "is this dependency satisfied?", which is a **different** question from any lane role and reads as complete-or-review-or-archived. **Flagged**: it wants a deliberate answer about what "satisfied" means on a board with no review lane, not a mechanical swap.

## Verification

`pnpm test:gate` **10 / 487 / 71 green** · `pnpm verify:fast` **PASS (15 steps, live boot smoke)** · `tsc -p packages/engine` clean · `pnpm lint` clean · census `--strict` exit 0 · 53/53 across the four suites covering these paths (including `executor-paused-abort-todo-benign`, which drives the hold-lane arm).

No changeset: `@fusion/engine` is private and this is internal behaviour on renamed boards only.

### DIFF
--- a/packages/engine/src/__tests__/executor-graph-failure-lanes-resolved.test.ts
+++ b/packages/engine/src/__tests__/executor-graph-failure-lanes-resolved.test.ts
@@ -197,3 +197,70 @@ describe("FN-5147: unusable-worktree recovery does not run on a finished or huma
     expect(recover).toHaveBeenCalledTimes(1);
   });
 });
+
+/*
+FNXC:WorkflowLifecycleColumns 2026-08-01-10:30 (fleet: executor.ts wip-lane liveness family):
+
+THE INVARIANT: "is this card still executing?" is the board's WIP lane.
+
+TWO DIRECTIONS, and this is why the family is converted per method rather than by matching the text.
+Most of these guards read `column !== "in-progress"` and REFUSE when they do not match, so a renamed
+board silently disabled them (the completion handoff was deferred on every card, the deferred
+approval resume never resumed, the completed-task watchdog returned on every tick). But the rerun
+watchdog reads `column === "in-progress"` and SKIPS when it matches — so on a renamed board that one
+never skipped, and a rerun could fire on a card that was still mid-execution. A mechanical swap of
+every `!== "in-progress"` would have converted the refusals and left the admission.
+
+`resumeApprovalAfterUnwindIfNeeded` is the case pinned here: it is a private method with a single
+boolean answer and no side effects on the refusal path, so the assertion is direct.
+*/
+describe("the wip-lane liveness family resolves the board's own wip column", () => {
+  function resumeApproval(executor: TaskExecutor, taskId: string): Promise<boolean> {
+    return (executor as unknown as {
+      resumeApprovalAfterUnwindIfNeeded: (id: string) => Promise<boolean>;
+    }).resumeApprovalAfterUnwindIfNeeded(taskId);
+  }
+
+  function armed(live: Record<string, unknown>, ir: WorkflowIr | undefined) {
+    const { executor, store } = harness(ir, live);
+    // The deferral marker this method consumes, plus a stub for the dispatch it guards.
+    (executor as unknown as { approvalResumeAfterUnwind: Set<string> }).approvalResumeAfterUnwind.add(live.id as string);
+    const dispatch = vi.fn().mockResolvedValue(true);
+    (executor as unknown as Record<string, unknown>).dispatchUnpauseResume = dispatch;
+    return { executor, store, dispatch };
+  }
+
+  it("resumes a deferred approval for a card in the board's wip lane", async () => {
+    // Pre-fix: `building` !== "in-progress", so the resume refused on every renamed board.
+    const live = { id: "FN-11", column: "building" };
+    const { executor, dispatch } = armed(live, RENAMED_IR);
+
+    expect(await resumeApproval(executor, "FN-11")).toBe(true);
+    expect(dispatch).toHaveBeenCalledTimes(1);
+  });
+
+  it("still refuses a card that has LEFT the wip lane", async () => {
+    // The paired negative: the fix must not resume work on a card that moved on.
+    const live = { id: "FN-12", column: "checking" };
+    const { executor, dispatch } = armed(live, RENAMED_IR);
+
+    expect(await resumeApproval(executor, "FN-12")).toBe(false);
+    expect(dispatch).not.toHaveBeenCalled();
+  });
+
+  it("still refuses a PAUSED card in the wip lane, so the lane is not the only gate", async () => {
+    const live = { id: "FN-13", column: "building", paused: true };
+    const { executor, dispatch } = armed(live, RENAMED_IR);
+
+    expect(await resumeApproval(executor, "FN-13")).toBe(false);
+    expect(dispatch).not.toHaveBeenCalled();
+  });
+
+  it("behaves identically on the DEFAULT board", async () => {
+    const live = { id: "FN-14", column: "in-progress" };
+    const { executor, dispatch } = armed(live, undefined);
+
+    expect(await resumeApproval(executor, "FN-14")).toBe(true);
+    expect(dispatch).toHaveBeenCalledTimes(1);
+  });
+});
--- a/packages/engine/src/__tests__/executor-graph-failure-lanes-resolved.test.ts
+++ b/packages/engine/src/__tests__/executor-graph-failure-lanes-resolved.test.ts
@@ -264,3 +264,44 @@ describe("the wip-lane liveness family resolves the board's own wip column", () 
     expect(dispatch).toHaveBeenCalledTimes(1);
   });
 });
+
+/*
+FNXC:WorkflowLifecycleColumns 2026-08-01-11:20 (fleet: executor.ts review-lane family):
+
+THE INVARIANT: "is this card in review?" is the board's review lane.
+
+`finalizeAlreadyReviewedTask` is the case pinned here because its failure is the loudest: it returns
+"missing" — a word that reads as "the task is gone" — for a card that is sitting in review on a
+renamed board. Everything downstream of the already-reviewed finalize path was therefore dead on any
+board that renamed its merge lane, and the log line said the task could not be found.
+*/
+describe("the review-lane family resolves the board's own review column", () => {
+  function finalizeAlreadyReviewed(executor: TaskExecutor, taskId: string): Promise<string> {
+    return (executor as unknown as {
+      finalizeAlreadyReviewedTask: (id: string) => Promise<string>;
+    }).finalizeAlreadyReviewedTask(taskId);
+  }
+
+  it("does not report a review-lane card as MISSING", async () => {
+    // Pre-fix: `checking` !== "in-review", so this returned "missing" for a card plainly in review.
+    const live = { id: "FN-15", column: "checking", steps: [], workflowStepResults: [] };
+    const { executor } = harness(RENAMED_IR, live);
+
+    expect(await finalizeAlreadyReviewed(executor, "FN-15")).not.toBe("missing");
+  });
+
+  it("still reports a card OUTSIDE the review lane as missing", async () => {
+    // The paired negative: the guard must still refuse a card that is not in review at all.
+    const live = { id: "FN-16", column: "building", steps: [], workflowStepResults: [] };
+    const { executor } = harness(RENAMED_IR, live);
+
+    expect(await finalizeAlreadyReviewed(executor, "FN-16")).toBe("missing");
+  });
+
+  it("behaves identically on the DEFAULT board", async () => {
+    const live = { id: "FN-17", column: "in-review", steps: [], workflowStepResults: [] };
+    const { executor } = harness(undefined, live);
+
+    expect(await finalizeAlreadyReviewed(executor, "FN-17")).not.toBe("missing");
+  });
+});
--- a/packages/engine/src/__tests__/executor-graph-failure-lanes-resolved.test.ts
+++ b/packages/engine/src/__tests__/executor-graph-failure-lanes-resolved.test.ts
@@ -305,3 +305,60 @@ describe("the review-lane family resolves the board's own review column", () => 
     expect(await finalizeAlreadyReviewed(executor, "FN-17")).not.toBe("missing");
   });
 });
+
+/*
+FNXC:WorkflowLifecycleColumns 2026-08-01-13:30 (fleet: executor.ts — the REVERSE half-conversion):
+
+`routeGraphFailureToExecutionResume`'s move DESTINATION was already resolved from the workflow (U7's
+`resolveReboundColumnFor`) while its entry gate compared against three default-lineage literals. So on
+a renamed board the router refused before it ever reached the resolved move: a recovery that was fully
+implemented, never running.
+
+That is the mirror image of the dangerous half-conversion. The familiar direction admits a card and
+then sends it to a column the board does not declare; this direction refuses a card whose recovery
+already worked. Both are one decision reading two boards, and only the second is silent — which is
+why it survived.
+
+The gate admits three shapes, so all three are asserted plus the refusal, which is what stops a gate
+that returns true unconditionally from passing.
+*/
+describe("the execution-resume router's gate reads the same board as its destination", () => {
+  function routeResume(executor: TaskExecutor, live: Record<string, unknown>, failureValue: string): Promise<boolean> {
+    return (executor as unknown as {
+      routeGraphFailureToExecutionResume: (live: unknown, failedNode: string, failureValue: string) => Promise<boolean>;
+    }).routeGraphFailureToExecutionResume(live, "merge", failureValue);
+  }
+
+  const withIncompleteSteps = (column: string, id: string) => ({
+    id, column, worktree: "/wt", steps: [{ name: "s", status: "pending" }], workflowStepResults: [],
+  });
+
+  it("admits a review-lane card", async () => {
+    const live = withIncompleteSteps("checking", "FN-18");
+    const { executor } = harness(RENAMED_IR, live);
+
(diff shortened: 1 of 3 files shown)


gsxdsm:
Second batch pushed on the same branch: the **wip-lane liveness family**. `executor.ts` **49 → 45**, repo backlog 686 → **682**.

Four guards asking "is this card still executing?" — the completion handoff deferral, the deferred approval resume, the completed-task watchdog, the workflow rerun watchdog.

**The finding worth having: the directions differ.** Three read `column !== "in-progress"` and REFUSE on no-match, so a renamed board silently switched them off — every completion handoff deferred as "no longer active" while the card was actively executing, the deferred approval resume never resumed, the completed-task watchdog returned on every tick. The rerun watchdog reads `column === "in-progress"` and SKIPS on match, so there the literal never matched and a rerun could fire on a card **mid-execution**.

A mechanical sweep of `!== "in-progress"` would have fixed the three refusals and left that one admission in place — the same asymmetry that makes a half-conversion worse than no conversion. Anyone batching this family across other files should convert per method, not per pattern.

Revert proof: 1 of the 4 new cases reddens with the approval-resume literal restored; the other three are the paired negatives (left the lane, paused in the lane, default board) and are labelled as such.

Verification: `pnpm test:gate` **158 / 10 / 487 / 71 green** · `tsc -p packages/engine` clean · `pnpm lint` clean · census `--strict` exit 0 · 12/12 in the new suite.

Remaining in the file: **45**, enumerated in the PR body. The two I am still flagging rather than converting are unchanged — the `to`/`from` move-effect parameters (trait-hook territory, not a resting column) and the dependency guard at `12406`, which needs a deliberate answer about what "dependency satisfied" means on a board with no review lane.

gsxdsm:
Third batch: the **review-lane family**. `executor.ts` **45 → 40**, repo backlog 682 → **677**.

The loudest one: **`finalizeAlreadyReviewedTask` returned `"missing"`** — a word that reads as *the task is gone* — for a card sitting in review on a renamed board. Everything downstream of the already-reviewed finalize was dead on any board that renamed its merge lane, and the log line said the task could not be found. That is the pattern worth watching for elsewhere: the inert guard does not report "renamed board", it reports something that sounds like a different problem entirely.

The rerun bounce needed **both** lanes from one snapshot. Its own comment says in-review must bounce exactly like in-progress; resolving the two separately is how you end up handling one lane and throwing on the other, which is the bug that comment was written about.

**Flagged, not converted:** the eligibility predicate at ~5521 needs the **archived** column, and the sync planner-lane resolver does not expose one. That is a resolver change rather than a call-site conversion, so it stays for whoever owns the resolver — I am not widening a shared resolver from inside a call-site sweep.

Revert proof: 1 of 3 new cases reddens with the literal restored; the other two are the paired negative and the default-board no-change case, labelled as such. Fixture correction recorded at the site: my first version passed `workflowStepResults: {}` and `getTaskMergeBlocker` threw on `.some`, so the case was failing for a reason unrelated to the guard.

Verification: `pnpm test:gate` **158 / 10 / 487 / 71** · `tsc -p packages/engine` clean · `pnpm lint` clean · census `--strict` exit 0 · 15/15 in the new suite.

**Running total for this PR: `executor.ts` 85 → 40, repo 722 → 677.** Remaining 40 with per-site notes in the body; the standing flags are the `to`/`from` move-effect parameters, the dependency guard at ~12423, and now the archived-lane predicate above.

gsxdsm:
Fourth batch: stuck-requeue liveness + the `fn_task_done` promotion. `executor.ts` **40 → 31**, repo backlog 677 → **668**.

**The one to read: the `fn_task_done` promotion converts the guard AND its move destination in the same change.** A lane-resolved guard sitting in front of a literal `moveTask(..., "in-progress")` is the half-conversion this program's worst regressions are made of — the card is admitted, the move is rejected by a board that does not declare that column, and the card is left mid-promotion with its completion **already logged**. Six of my earlier PRs hit that exact shape; it is why I convert per method and check every destination and exemption in the same commit.

The destination resolves through `resolvePlannerLanes`, not `resolveResumeLanes`, because **a move needs a column that exists**: `PlannerLanes` leaves `wip` undefined when the board declares none, so the promotion is skipped with an explicit log line rather than issuing a move to an invented column. The handoff below is gated on the wip lane anyway. Both log messages now name the actual columns instead of hardcoding `todo`/`in-progress` — otherwise the operator reads the wrong column name in their own log.

The stuck-requeue group (4 sites) is the same inert-guard-with-a-misleading-message pattern as `finalizeAlreadyReviewedTask`: on a renamed board the liveness check always answered "moved on", the cleanup never ran, and the log blamed a concurrent recovery that had not happened.

Verification: `pnpm test:gate` **158 / 10 / 487 / 71** · all 11 `task-done` suites **76/76** · `tsc -p packages/engine` clean · `pnpm lint` clean · census `--strict` exit 0.

**Running total: `executor.ts` 85 → 31 (54 guards), repo 722 → 668.** No new revert-proof cases in this batch — the behaviour is covered by the 11 task-done suites and the stuck-requeue paths have no seam I can drive without a live session harness; I would rather say that than add a case that passes either way.

gsxdsm:
Fifth batch: `execute()`'s preflight drift checks + the merge-confirmed finalize guards. `executor.ts` **31 → 25**, repo backlog 668 → **662**.

`execute()`'s three preflight checks all ask "is this card in the wip lane, in a state it should not be in?" — merge-confirmed while still executing, stale `mergeDetails`, in-wip with no worktree. None fired on a renamed board, so every recovery they perform silently stopped: the half-landed row never finalized, stale merge state never reset, a missing worktree never rebuilt. The third one's own error message says it *"usually indicates a partial updateTask/moveTask sequence failed"* — a diagnostic that could not print on a renamed board.

**The distinction I would not want a batch worker to swap blindly**, because it goes both ways within a few lines:

| question | resolver | why |
|---|---|---|
| "is this card in the ONE column finalize targets?" | the board's **complete** column | the terminal union carries the legacy ids, so a card in a column merely *named* `done` would read as already finalized and the finalize would be **skipped** |
| "is this card already finished, so do not move it?" | the **terminal union** | over-inclusion only skips a move; under-inclusion moves a finished card out of its terminal column |

Both are recorded at their sites. A mechanical sweep that picked one resolver for every `=== "done"` would break one of them.

Verification: `pnpm test:gate` **10 / 158 / 487 / 71** · 64/64 across the five lane-resolution suites · `tsc -p packages/engine` clean · `pnpm lint` clean · census `--strict` exit 0.

**Running total: `executor.ts` 85 → 25 (60 guards, 71%), repo 722 → 662.**

Remaining 25, and the three I am deliberately **not** converting:
- **`5805` — a `listTasks({ column: "in-progress" })` QUERY plus its matching filter.** Converting the filter alone leaves the query selecting by literal, so the sweep still sees nothing on a renamed board — a conversion that reads as done and changes nothing. The query needs a project-level lane resolution (it has no task to resolve from), which is a different piece of work; the census already tracks query filters as their own class.
- **`to`/`from` move-effect parameters** (`3465`, `3489`, `3540`, `3550`, `4070`, `14593`, `15910`) — a move's endpoints, not a card's resting column. Trait-hook territory.
- **`12434`, the dependency guard**, and **`5529`-`5532`, the eligibility predicate needing an archived lane** — unchanged from earlier comments.

The rest (~14) are ordinary wip/hold guards in the same shape as the batches above.

gsxdsm:
Sixth batch, and it contains the finding I would most want other fleet workers to read. `executor.ts` **25 → 18**, repo backlog 662 → **655**.

## The REVERSE half-conversion

`routeGraphFailureToExecutionResume`'s move **destination was already resolved** from the workflow — U7's `resolveReboundColumnFor` — while its **entry gate** still compared against three default-lineage literals. On a renamed board the router refused before it ever reached the resolved move: **a recovery that was fully implemented, never running.**

This is the mirror image of the half-conversion this program keeps finding, and it is the more dangerous of the two to leave behind:

| direction | what happens | visible? |
|---|---|---|
| resolved gate → literal destination | card is admitted, move is rejected by a board that has no such column, card left half-moved | **yes** — the move errors |
| literal gate → resolved destination | card is refused; the working recovery never runs | **no** — nothing is logged as wrong |

Both are one decision reading two boards. Only the second is silent, which is exactly why it survived U7's own conversion of the destination. **When you convert a destination, check the gate in front of it in the same commit** — and when you audit a file for leftovers, grep for resolved destinations too, not just literal ones.

Also in this batch: the unpause listener's pair of wip checks (an unpaused card with no active session was never resumed on a renamed board), the no-`fn_task_done` retry's worktree contract, and the paused-abort cleanup skip that recognises a card **already parked in the hold lane with progress** — on a renamed board that skip never fired, so the cleanup ran and discarded the preserved progress.

Revert proof: **3 of 4** new cases redden with the router's literals restored (all three admission shapes); the fourth is the paired refusal that must stay green.

Verification: `pnpm test:gate` **158 / 10 / 487 / 71** · 19/19 in the suite · `tsc -p packages/engine` clean · `pnpm lint` clean · census `--strict` exit 0.

**Running total: `executor.ts` 85 → 18 (67 guards, 79%), repo 722 → 655.**

The 18 that remain are the four classes I have flagged rather than guessed, and I am leaving them flagged:
- **7 `to`/`from` move-effect parameters** — a move's endpoints, not a resting column; trait-hook territory.
- **`5805` + `20213`, two enumeration scans** — one is a `listTasks({ column: "in-progress" })` query whose filter I cannot convert without the query (it would read as done and change nothing); the other loops every task, so a per-task lane resolution is a real cost that wants a shared memo rather than a call-site swap.
- **`12456`, the dependency guard** — "is this dependency satisfied?" is not any single lane role, and it needs a deliberate answer for a board with no review lane.
- **`5529`-`5532`, the eligibility predicate** — needs an archived lane the sync resolver does not expose; a resolver change, not a call-site conversion.

gsxdsm:
Seventh batch, and it **withdraws one of my own flags**. `executor.ts` **18 → 15**, repo backlog 655 → **651**.

Two batches ago I flagged `isRequiredArtifactRecoveryProtected` as blocked: it needs the **archived** column, and the sync planner-lane resolver does not expose one. I said widening a shared resolver from inside a call-site sweep was out of scope, and that part stands.

What I got wrong was the conclusion. The predicate did not have to be **synchronous**. All four callers already `await store.getTask` immediately before calling it, so making it async adds no I/O ordering and no parameter list that four call sites would have to keep in agreement. Flag withdrawn with the reason rather than left standing as though it were still blocked — a stale flag is a small lie in the ledger.

What it was protecting: a **finished** card, or a **review row a human was holding** with auto-merge off, could be moved to the replan column and have its status rewritten to `needs-replan`. All three conditions read false on a renamed board.

Verification: 68/68 across the six artifact suites · `pnpm test:gate` **10 / 158 / 487 / 71** · `tsc -p packages/engine` clean · `pnpm lint` clean · census `--strict` exit 0.

**Running total: `executor.ts` 85 → 15 (70 guards, 82%), repo 722 → 651.**

Also rebased onto `origin/main` (it was DIRTY). Worth noting what the rebase demonstrated: main had arrived with another worker's `scheduler.ts` drop (28 → 27) and no re-recorded baseline — the merged #2679 auto-tighten rewrote the entry and **exited 0** instead of reddening this PR's gate for an unrelated change. That is the case it was built for, working on the first real occurrence. I committed the tightened write, which is what the printed notice asks for.

The 15 that remain are the three flag classes I am still holding: **7** `to`/`from` move-effect parameters, **2** enumeration scans (one needs the `listTasks` query converted with it or the conversion changes nothing; the other loops every task and wants a shared memo), and the **dependency guard** at `12456`, which needs a deliberate answer for what "satisfied" means on a board with no review lane.

gsxdsm:
**Collision notice: #2689 is converting the same file.** It was opened minutes after this one and is at 85 → 75 (10 guards) versus 85 → 15 (70 guards) here. I have laid out the comparison and the three findings a superset-only batch misses on #2689 and offered to fold either way — coordinator's call, and I will do the merge work whichever way it goes. Both must not land.

gsxdsm:
Closing in favour of **#2703**, which is the same work rebuilt on top of #2689.

What happened, recorded because it cost real time: #2689 claimed and converted `executor.ts` in parallel with this PR — neither of us saw the other until both were open. #2689 landed first with the terminal-pair batch (85 → 57), which made 30 of the 70 conversions here identical to code already on main. Resolving 30 conflict hunks in a 20k-line lifecycle file would have produced a diff nobody could review, so #2703 re-applies only what #2689 left behind: **57 → 15**, the review lane, the wip-lane liveness family, the split snapshots, and the four half-conversions that change behaviour.

Nothing from this PR is lost — every finding, comment and all 19 test cases carried over.

The process lesson is the one I put on #2689: **the claim has to be visible before the work, not in the PR title after it.** I now post the claim as a comment before starting, which is how I took `register-task-workflow-routes.ts` (#2701) without a collision.

PR #2692, opened 2026-07-30
fix(census): the work order was sending fleet workers at non-columns (722 -> 714)

gsxdsm:
Found while claiming `TaskDetailModal.tsx` — its census entry included `session.agentState === "done"`, an **agent state, not a lane**. Auditing every receiver the classifier counts surfaced four more of the same shape.

## The misclassified receivers

| site | receiver | what it actually is |
|---|---|---|
| `register-chat-routes.ts` | `event.type === "done"` | an SSE event type |
| `useTaskDiffStats.ts` | `mode === "done"` | a cache-key mode |
| `async-mission-store.ts` | `evidence.kind === "done"` | an evidence kind |
| `telemetry-hub.ts` | `event.kind === "done"` | a telemetry event kind |
| `TaskDetailModal.tsx` | `session.agentState === "done"` | an agent state |

Each shares a **word** with a column id and nothing else. Converting one asks the trait registry what lane an SSE event is in, which has no answer — the same failure class as converting `role === "triage"`, which this list already exists to prevent.

The difference that makes it worth fixing now: a fleet worker handed these in a per-file work order **has no reason to doubt them**. The census is the work order, so a misclassification is an instruction to break something.

## What I did not exclude

`state` is deliberately kept. `state === "archived"` in `audit-ops.ts` / `comments-ops.ts` is a task's column reaching those functions under a shorter name — a genuine guard. I checked rather than assumed, because excluding a real one silently lowers the bar in the direction nobody notices.

## Census effect

```
column  722 -> 714
role      5 -> 14
```

Those 8 are **reclassified, not converted** — this PR changes no production code. The baseline is re-recorded so `--strict` agrees.

## Verification

`pnpm test:gate` green (10 / 158 / 487 / 71). `pnpm check:lifecycle-columns` exits 0. `pnpm lint` clean.

No changeset: instrument accuracy, no user-facing change.


gsxdsm:
Real — the name-based exclusions had **no test that could fail**. Fixed.

Every case in *"state, phase and result enums are not column guards"* supplies non-column siblings (`busy`, `ready`, `pushing`, `confirm`, `stopped`, `exhausted`), and the **sibling** mechanism classifies those on its own. So the receiver-*name* rule was never exercised.

**Measured, not suspected: deleting the entire `type`/`mode`/`kind`/`phase`/`agentState` list left all 38 tests green.** The list could have been dropped in a later refactor with nothing to notice, and every site it protects would have re-entered the backlog — sending fleet workers at non-columns, which is the exact defect this PR exists to fix.

Added five bare comparisons, one per name, with nothing else in the expression that could produce the answer. `type` and `mode` had **no coverage of any kind** before this.

Mutation: removing the list now fails **5 of 43**.

Verified: 43 tests green, `census --strict` exit 0, lint clean, gate green (487 + 158 + 10 + 71). Rebased onto current main; baseline re-recorded (721 → 713) since this PR's fix removes 8 non-columns from the backlog.

---

**Process finding worth raising beyond this PR.** Every fleet PR edits `lifecycle-column-census-baseline.json`, so **every fleet PR conflicts with every other fleet PR** — I have now rebased four of mine purely on that file. It is generated, so the resolution is always "take main's, re-run `--update-baseline`", never a real merge. Two options worth considering: drop the baseline from conversion PRs and re-record it periodically on main, or teach the merge driver to regenerate it. As it stands the fleet is serialized on a file nobody actually edits by hand.

gsxdsm:
**A sixth misclassified receiver for your list: `transition`.**

`packages/dashboard/src/github-tracking-comments.ts` has seven `transition === "in-progress" | "done"` comparisons. `transition` is the **comment kind** passed to `formatTrackingComment` — a template selector, not a lane. Same shape as your five: it shares a word with a column id and nothing else.

Context, since it is a slightly worse case than the others: in that file the column id and the comment kind used to be **the same value** (`event.to` was compared against the literals *and* passed as the `transition` argument). I have just split them in #2707 — the lanes are now resolved and the kind is derived — so the file has **no lane comparison left**, but the census still reports 7 because of the receiver.

I did not mark them `DELIBERATE-LITERAL` to move the number: that marker means "a lifecycle literal reviewed and kept", and these are not lifecycle literals. Using it as a census-silencer would put a wrong reason in the code to make a count look better. So the site is genuinely done and the instrument still counts it — which is your PR's whole point, one file further on.

PR #2696, opened 2026-07-30
fleet: TaskDetailModal.tsx 30 → 10 (20 converted; 10 flagged with reasons, incl. a census misclassification)

gsxdsm:
Claiming **`packages/dashboard/app/components/TaskDetailModal.tsx`** — the largest unclaimed cluster (self-healing, executor, TaskCard and scheduler were already taken).

## Census before/after

| | before | after |
|---|---:|---:|
| `TaskDetailModal.tsx` column guards | **30** | **10** |
| repo backlog | 722 | **702** |

Baseline re-recorded to 10 — shrinks by exactly the converted count. `--strict` exits 0.

## Converted: 20, existing helpers only

`isCompleteColumnRole`, `isArchivedColumnRole`, `isWipColumnRole`, `isReviewColumnRole` from `utils/columnRoles`, alongside the `isFieldEditableColumnRole` this file already imported. No new abstractions, no new helper.

6× `=== "done"` · 4× `=== "archived"` · 1× `!== "archived"` · 3× `=== "in-progress"` · 6× `=== "in-review"`

One alias — `taskColumnRoleFlags = workflowMoveMetadata?.currentColumnFlags` — so every site reads metadata the modal **already fetches**: `canEdit` was consuming it at line 1725 before this change, so there is no new data flow and no new request.

**Behaviour-preserving by construction.** Each helper is `flags ? <trait> : <legacy id>`, and flags are absent precisely when the modal has not yet loaded its workflow metadata — so with no flags the legacy comparison is exactly what ran before.

## The 10 remaining, each with a reason — flagged, not guessed

**Already-converted fallback, miscounted (2 sites, one line)**
- `560` — `return column === "todo" || column === "in-progress";` is the documented legacy fallback *inside* a helper that already takes `flags` and checks `flags.hold || flags.countsTowardWip` first. It is the fallback arm, not an unconverted guard. It needs a `DELIBERATE-LITERAL` marker rather than conversion; I have not added one because the marker changes the tracked deliberate count and I would rather that be a deliberate call by whoever owns the ratchet than a side effect of my batch.

**Not a column at all — census misclassification (1 site)**
- `346` — `session.agentState === "done"`. Receiver is `agentState`; `done` is an agent state, not a column. The classifier keys on receiver name, and `agentState` matches neither the role-token list nor `/status/i`, so it falls through to `column`. Converting it would ask "which column has the complete trait" about an agent's state. **Worth fixing in the classifier** — `agentState` belongs in its token list.

**Module scope, no flags available (5 sites)**
- `277` (`resolveDefaultTab`), `869`, `870`, `918`, `924` — module-level functions taking a bare `column: ColumnId`. Converting means threading `flags` through their signatures and every call site. That is a real refactor with its own blast radius, not a mechanical substitution, so it is out of scope for this batch by the fleet rules.

**Another task's column (2 sites)**
- `3536` — `overlapBlockerTask.column === "in-progress" || … === "in-review"`. `currentColumnFlags` describes *this* task's column, not the blocker's. Passing it would be wrong, and passing `undefined` would silently use legacy ids for that card while looking converted. Correct fix is a per-task flag lookup from `moveColumns`, which this component does not currently build.

## Verification

`pnpm lint` clean · dashboard app `tsc` clean · `--strict` exits 0.

`TaskDetailModal.test.tsx` 45 passed / 6 failed and the wider `app/components/__tests__` 9883 passed / 209 failed — **both byte-identical on clean `origin/main`**, confirmed by stashing these changes and re-running. None of those failures is from this batch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved task detail behavior by consistently recognizing workflow column roles.
  * Updated editing, deletion, archiving, reverting, oversight, and review controls to reflect each task’s workflow status.
  * Corrected tab visibility, custom-field editing, pull request, and move-menu behavior for archived and review-related tasks.

### DIFF
--- a/packages/dashboard/app/components/TaskDetailModal.tsx
+++ b/packages/dashboard/app/components/TaskDetailModal.tsx
@@ -24,7 +24,13 @@ import { resolveEffectivePlannerOversightLevel } from "../../../core/src/workflo
 import { resolveTaskSessionAdvisorEnabled } from "../../../core/src/session-advisor";
 import { isNearDuplicateCanonicalInactive } from "../../../core/src/near-duplicate-canonical";
 import { getRevertOfId, findOpenUndoTaskForSource } from "../utils/taskRevert";
-import { isFieldEditableColumnRole } from "../utils/columnRoles";
+import {
+  isArchivedColumnRole,
+  isCompleteColumnRole,
+  isFieldEditableColumnRole,
+  isReviewColumnRole,
+  isWipColumnRole,
+} from "../utils/columnRoles";
 import { resolveEffectiveAutoMerge } from "../../../core/src/task-merge";
 import { uploadAttachment, deleteAttachment, updateTask, repairOverlapBlocker, pauseTask, unpauseTask, fetchTaskDetail, fetchTaskVerificationRequest, fetchSettings, fetchTaskEffectiveSettings, fetchGlobalSettings, requestSpecRevision, rebuildTaskSpec, approvePlan, rejectPlan, refineTask, fetchWorkflowResults, assignTask, fetchAgents, fetchAgent, refreshPrStatus, fetchBoardWorkflows, updateTaskCustomFields, summarizeTitle, fetchWorkflowSettingValues, nudgeOverseer, stopOverseer, explainOverseer, fetchModels, fetchNodes, api } from "../api";
 import type { RevertTaskOptions, RevertTaskResult, ModelInfo, NodeInfo } from "../api";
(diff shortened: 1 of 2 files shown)


gsxdsm:
Closing as **superseded**. `TaskDetailModal.tsx` is at **5** on main; this PR targeted 30 → 10. A more complete conversion landed while this sat conflicting, so rebasing it would mean resolving a source-level conflict in a 5000-line file to deliver strictly less.

Same outcome as #2700 and #2708 earlier: I opened this at 10:08 and #2698 opened at 10:21 against the same file, and the more thorough one won. That is the right result — but the duplication was avoidable, and the cause was mine: I did not check `git branch -r` or the open-PR file list before claiming. I have checked the claimed-file set before every cluster since.

Nothing here is worth salvaging. The four sites I flagged rather than converted (the dependency-pair sync filters, the `agentState` census misclassification, and the two already-trait-resolved sites) are either converted on main or still visible in the census, so no finding is lost by closing this.

PR #2698, opened 2026-07-30
fleet: TaskDetailModal.tsx 30 -> 7

gsxdsm:
## Census before / after

| | before | after |
|---|---:|---:|
| `TaskDetailModal.tsx` | **30** | **7** |
| repo backlog | 721 | **698** |

Backlog dropped by **23** — the converted count, nothing else moved.

**Not 30 → 0.** The seven survivors are enumerated below with reasons rather than absorbed into the number.

## Converted (23)

Four role bindings declared immediately after `workflowMoveMetadata` — their source — and 20 in-component comparisons collapsed onto them.

Two module-scope helpers (`resolveDefaultTab`, `requiresExecutionModeReplan`) take a bare column id with no flags in scope, so they use the fallback-only form. That is **centralisation, not trait resolution**, and each is labelled as such at the site so it stays greppable as "still needs its flags threaded".

## Not converted (7), each with a reason

| count | site | why |
|---:|---|---|
| 2 | `showNearDuplicateWarning` (~881) | Sits **above** `workflowMoveMetadata`, so referencing the role bindings is a temporal-dead-zone error. Needs the state declaration hoisted — behaviour-safe, but it reorders hooks in a 6500-line component, which is not a vocabulary edit. |
| 2 | two `useEffect` dep arrays (~930, ~936) | Same problem, subtler: the callback *bodies* could reference the bindings, but a **dep array is evaluated eagerly** at the hook call, which is above the declaration. Same hoist. |
| 2 | `overlapBlockerTask.column` (~3559) | A **different task's** column. The modal holds flags for its own card only; resolving the blocker's role means fetching its flags. Out of scope. |
| 1 | `session.agentState === "done"` (~353) | An **agent state, not a column**. Already reclassified in #2692 — this file drops to 6 counted when that lands. |

## Late-arriving flags — checked, not assumed

`currentColumnFlags` is `null` until the workflow fetch resolves, so every role here flips after first paint. That is the hazard that produced four stale memos in `TaskCard` (#2688 review), so I ran an AST pass over every `useMemo` / `useEffect` / `useCallback` in the file.

**None needed a new dependency** — the 20 converted sites are all render-path expressions rather than memoised closures. Worth stating explicitly, because "no dep changes" in a conversion PR usually means nobody looked.

## Verification

`pnpm test:gate` green (10 / 158 / 487 / 71). `pnpm check:lifecycle-columns` exits 0 with the baseline re-recorded here. `tsc -p tsconfig.app.json` clean. `pnpm lint` clean. Targeted `TaskDetailModal` suites pass (5/5).

Note: the full `TaskDetail*` glob exceeds a 10-minute run locally, so I verified with the targeted suites plus typecheck rather than reporting a number I did not measure.

### DIFF
--- a/packages/dashboard/app/__tests__/task-detail-column-role-identity.test.ts
+++ b/packages/dashboard/app/__tests__/task-detail-column-role-identity.test.ts
@@ -1,0 +1,144 @@
+import { describe, expect, it } from "vitest";
+import { readFileSync } from "node:fs";
+import { resolve } from "node:path";
+import ts from "typescript";
+
+/*
+FNXC:WorkflowResolvedColumns 2026-07-30-21:30 (fleet — TaskDetailModal role conversion):
+Pins the two invariants that made the async role conversion in this component SAFE.
+
+Converting a column-id literal to a trait read makes the answer ASYNCHRONOUS: the flags arrive from a
+workflow fetch that resolves after first paint. Review of this conversion found five distinct ways
+that goes wrong (documented in docs/solutions/ui-bugs/async-resolved-column-roles-in-components.md).
+Two of them are structural — they can be pinned in source, and both were REAL defects in earlier
+revisions of this same file, not hypotheticals:
+
+  STALE IDENTITY. The metadata resolves for task A; the operator opens task B before it lands. Without
+  an identity tag the component answers B's role questions with A's flags — it is resolved, it is
+  non-empty, and it is about the wrong card. Fixed by tagging the state with `taskId` and gating every
+  read on `detailFlagsAreForThisTask`.
+
+  EAGER ACTION ON AN UNRESOLVED GUESS. The reconciliation effects close the PR tab when the card is
+  not in a review lane. Before the flags land, `isReviewColumn` is the legacy-id fallback — false for
+  a custom review column — so the effect fired on a GUESS and the tab opened and instantly bounced.
+  Fixed by making those effects wait for `detailFlagsAreForThisTask`.
+
+WHY A SOURCE ASSERTION. Both invariants are about a race between a fetch and a re-render, which a
+render test cannot force deterministically without freezing the very timing under test — and
+`fireEvent`-style tests famously cannot observe remount/identity bugs at all. The repo already uses
+source-level guards for exactly this class (see task-detail-modal-tablet-width.test.ts). This walks
+the AST rather than grepping, so a renamed variable or a reordered argument cannot slip past.
+
+VERIFIED TO FAIL ON THE ORIGINAL DEFECT: dropping the `taskId` tag fails case 1; removing the
+`detailFlagsAreForThisTask` guard from either reconciliation effect fails case 2.
+*/
+
+const SOURCE_PATH = resolve(__dirname, "../components/TaskDetailModal.tsx");
+const source = readFileSync(SOURCE_PATH, "utf8");
+const sourceFile = ts.createSourceFile(SOURCE_PATH, source, ts.ScriptTarget.Latest, true, ts.ScriptKind.TSX);
+
+/** Every `useEffect(cb, [deps])` call, with the identifiers used in each half. */
+function collectEffects(): Array<{ line: number; used: Set<string>; declared: Set<string> }> {
+  const effects: Array<{ line: number; used: Set<string>; declared: Set<string> }> = [];
+
+  const identifiersIn = (node: ts.Node): Set<string> => {
+    const out = new Set<string>();
+    const walk = (n: ts.Node) => {
+      if (ts.isIdentifier(n)) out.add(n.text);
+      ts.forEachChild(n, walk);
+    };
+    walk(node);
+    return out;
+  };
+
+  const visit = (node: ts.Node) => {
+    if (
+      ts.isCallExpression(node)
+      && ts.isIdentifier(node.expression)
+      && node.expression.text === "useEffect"
+    ) {
+      const [callback, deps] = node.arguments;
+      if (callback && deps && ts.isArrayLiteralExpression(deps)) {
+        effects.push({
+          line: sourceFile.getLineAndCharacterOfPosition(node.getStart(sourceFile)).line + 1,
+          used: identifiersIn(callback),
+          declared: new Set(deps.elements.flatMap((element) => [...identifiersIn(element)])),
+        });
+      }
+    }
+    ts.forEachChild(node, visit);
+  };
+  visit(sourceFile);
+  return effects;
+}
+
+describe("TaskDetailModal resolved column roles are identity-safe", () => {
+  /*
+  STALE IDENTITY. Asserted on the read path, not the write path: what matters is that no role read
+  can reach the flags without first proving they belong to THIS task.
+  */
+  it("gates resolved column flags on the metadata belonging to the open task", () => {
+    expect(
+      source,
+      "workflow move metadata must carry the task id it was resolved for",
+    ).toMatch(/taskId:\s*string\s*\}\s*\)\s*\|\s*null/);
+
+    expect(
+      source,
+      "the identity check must compare the metadata's taskId against the open task",
+    ).toMatch(/const detailFlagsAreForThisTask\s*=\s*workflowMoveMetadata\?\.taskId === task\.id/);
+
+    /*
+    The flags used for every role question must be the GATED value. Reading
+    `workflowMoveMetadata?.currentColumnFlags` directly is the bug: resolved, non-empty, wrong card.
+    */
+    expect(
+      source,
+      "resolved flags must pass through the identity gate before any role read",
+    ).toMatch(
+      /const detailColumnFlags\s*=\s*detailFlagsAreForThisTask\s*\?\s*workflowMoveMetadata\?\.currentColumnFlags\s*:\s*undefined/,
+    );
+
+    const roleReadsOfUngatedFlags = source.match(
+      /is[A-Za-z]*ColumnRole\(\s*workflowMoveMetadata\?\.currentColumnFlags/g,
+    );
+    expect(roleReadsOfUngatedFlags, "no role helper may read the ungated metadata").toBeNull();
+  });
+
+  /*
+  EAGER ACTION ON AN UNRESOLVED GUESS. Any effect that ACTS on a resolved role must also depend on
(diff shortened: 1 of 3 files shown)


gsxdsm:
Right, and it is a bug my own partial conversion created rather than a pre-existing one.

Tab **visibility** resolved the complete/review roles while the reconciliation effects still compared literal ids. On a custom board the two disagreed: the PR tab appeared — the column carries the review role — and the effect immediately redirected to Changes, because the column is not named `in-review`. Same for Summary. **A tab that shows up and instantly bounces is worse than one that never shows.**

That is the cost of converting half a pair, and it argues directly against my own earlier call: I flagged those two effects as blocked on a hoist and shipped the visibility half anyway. The right move was to do the hoist or convert neither. I have done the hoist.

`workflowMoveMetadata` and the four role bindings now sit **above** these effects rather than ninety lines below. The subtlety that made it look harder than it was: a **dep array is evaluated eagerly at the hook call**, so it was not enough that the callback bodies would run later — the declaration genuinely had to move. Both effects now list the role they read.

**TaskDetailModal.tsx 30 → 5**, up from 30 → 7 before this fix.

The remaining five are unchanged in kind, and I checked each for the same half-converted-pair hazard before leaving it:

- **2** in `showNearDuplicateWarning` — still above the declaration, but paired with nothing that was converted, so no inconsistency exists. Hoisting further would reorder more hooks for two sites with no correctness benefit.
- **2** on `overlapBlockerTask.column` — a *different* task's column; the modal holds flags for its own card only.
- **1** `session.agentState` — not a column at all (#2692 reclassifies it).

Verification: `pnpm test:gate` green (10 / 158 / 487 / 71), `pnpm check:lifecycle-columns` exits 0 with the baseline re-recorded, `tsc -p tsconfig.app.json` clean, `pnpm lint` clean, targeted TaskDetailModal suites 5/5.

gsxdsm:
Real, and it is the most damaging of the three forms this hazard has taken — because the redirect is **destructive** and the correction is not.

`workflowMoveMetadata` is null until the workflow fetch lands, so on first paint the role bindings return the legacy-id fallback: false for a custom review or complete column. The effects took that as authoritative. The modal opened, immediately bounced the operator off the PR or Summary tab they had chosen, and resolving the metadata later did not put them back.

Both effects now return early while the roles are unresolved, and both list `workflowMoveMetadata` in their dependencies so they re-run when it arrives.

**Why waiting is the right direction rather than a coin-flip:** showing a tab a moment longer on a card that turns out not to apply is benign and self-corrects the instant metadata resolves. Throwing away a deliberate tab selection is permanent. When one branch is recoverable and the other is not, acting on a guess should favour the recoverable one.

**This is the third instance across two files, and they are one class:**

| form | where |
|---|---|
| stale memo — role read, deps keyed only on `task.column` | TaskCard (#2688) |
| frozen initializer — `useState` captured the pre-load answer | TaskCard (#2688) |
| eager action on a guess — effect mutated state before flags resolved | this |

All three come from the same change in kind: **`task.column` is stable across renders; a trait-derived role is not.** Any converted file whose roles feed a memo, an initializer, or an effect that *mutates* state needs checking for all three — and this repo has no `react-hooks/exhaustive-deps` rule to help. I have put that in the commit so it is findable from the history rather than only in a review thread.

Verification: `pnpm test:gate` green (10 / 158 / 487 / 71), `pnpm check:lifecycle-columns` exits 0 with the baseline re-recorded, `tsc -p tsconfig.app.json` clean, `pnpm lint` clean, targeted TaskDetailModal suites 5/5.

gsxdsm:
Right — and my previous fix is what hid it. Guarding on `workflowMoveMetadata === null` asked whether metadata had **loaded**, not whether it describes the task currently open.

The fetch effect does reset the state to null on a task change, but it is declared **below** the reconciliation effects. So on the render where a mounted modal switches tasks, mine run first and still see the previous card's flags — non-null, so the guard let them through, and a valid tab gets redirected or an invalid one stays, decided by **another task's workflow**.

That ordering was load-bearing and invisible. Tagging the payload with its `taskId` makes it checkable rather than order-dependent: consumers compare identity and fall back to the legacy id when it does not match — the same safe answer they already use before any fetch lands.

**The role bindings take the same guard, not just the effects.** Resolving a role from another task's flags is worse than the fallback: confidently wrong rather than merely stale. That distinction is why I did not simply reorder the effects — reordering fixes the two call sites and leaves the bindings resolving from stale data for anything else that reads them.

**Four forms now, all from one change in kind** — `task.column` is stable across renders, a trait-derived role is not:

| | form | found in |
|---|---|---|
| 1 | stale memo — deps keyed only on `task.column` | TaskCard (#2688) |
| 2 | frozen initializer — `useState` captured the pre-load answer | TaskCard (#2688) |
| 3 | eager action on a guess — effect redirected before flags resolved | #2698 |
| 4 | stale **identity** — flags resolved, but for the previous task | this |

The fourth only appears in components that stay **mounted across entity changes**, which is why TaskCard never showed it and the modal did. Worth putting in the fleet notes: a converted component that persists across task switches needs the identity check, not just a resolved check.

Verification: `pnpm test:gate` green (10 / 158 / 487 / 71), `pnpm check:lifecycle-columns` exits 0 with the baseline re-recorded, `tsc -p tsconfig.app.json` clean, `pnpm lint` clean, targeted TaskDetailModal suites 5/5.

gsxdsm:
Right, and it is the flip side of my own previous fix — I traded *"acts on stale data"* for *"never acts"*, which is not a fix.

Writing `null` when the lookup returns no metadata, or fails, is indistinguishable from *"has not resolved yet"*. So `detailFlagsAreForThisTask` stayed false forever, both reconciliation effects returned indefinitely, and an invalid tab stayed active — the exact failure the identity guard was added to prevent, reached from the other end.

The bug was in treating resolution as binary. It has **three** states:

| state | value |
|---|---|
| unresolved | `null` |
| resolved with flags | `{ taskId, moveColumns, currentColumnFlags }` |
| **resolved empty** | `{ taskId }` |

The third still **identifies the task**, so consumers know the answer has landed and the roles fall back to the legacy id — which is a real answer for a workflow that declares nothing, not a placeholder. The failure path now settles the same way rather than leaving the modal waiting on a fetch that already lost.

That makes five forms of this hazard on this one file. They are all the same root cause — an async value wearing the shape of a synchronous property — and each fix has to be checked for the *opposite* failure it might introduce, which is what I missed here.

I have added it to `docs/solutions/ui-bugs/async-resolved-column-roles-in-components.md` in #2710 so the next converter gets all five rather than rediscovering them one review at a time.

Verification: `pnpm test:gate` green (10 / 158 / 487 / 71), `pnpm check:lifecycle-columns` exits 0 with the baseline re-recorded, `tsc -p tsconfig.app.json` clean, `pnpm lint` clean, targeted TaskDetailModal suites 5/5. Rebased onto current `origin/main`.

gsxdsm:
**One more under-keyed hook, found by re-checking with a real AST walk.**

`handleDelete` reads `isArchivedColumn` and its dependency array omitted it — so on a custom archived column the delete handler kept its first-paint answer.

It escaped because the check I ran on this file and reported as *"an AST pass over every `useMemo`/`useEffect`/`useCallback`"* was actually a **regex over source**. It matched only hooks written `useMemo(() => {` with `}, [` on its own line, so a differently-formatted `useCallback` was invisible. The same blind spot hid a memo in TaskCard, which #2688's review caught.

Replaced with a `ts.createSourceFile` walk that collects every identifier in the callback and every identifier in the dep array, so formatting cannot hide a hook.

| file | regex said | AST found |
|---|---|---|
| `TaskCard.tsx` | clean | **1 under-keyed** |
| `TaskDetailModal.tsx` | clean | **1 under-keyed** |

Both now report 0.

The general point is the one this programme keeps making, and I made the mistake myself: **a pattern-matcher that reports completeness it cannot deliver is worse than no check**, because its clean output gets quoted as measurement. I wrote the AST census specifically to replace greps that disagreed with each other — and then verified my own conversion with a grep.

Verification: targeted TaskDetailModal suites 5/5, `pnpm test:gate` green (10 / 158 / 487 / 71), `pnpm check:lifecycle-columns` exits 0, `tsc -p tsconfig.app.json` clean, `pnpm lint` clean.

gsxdsm:
Rebased onto current `origin/main` (census-baseline conflict; resolved by taking main's and re-recording — `--strict` exits 0) and **added the test this PR was missing**.

### It shipped 25 conversions with no test

That is below the bar every other PR in this fleet has met, and I am not comfortable leaving it — the gap matters more here than usual. Converting a column-id literal to a trait read makes the answer **asynchronous**, and review of this very file found five distinct ways that goes wrong. Two are structural, and **both were real defects in earlier revisions of this file**, not hypotheticals:

- **Stale identity** — metadata resolves for task A, the operator opens task B before it lands, and the component answers B's role questions with A's flags. Resolved, non-empty, and about the wrong card.
- **Eager action on an unresolved guess** — the reconciliation effects acted on the legacy-id fallback before the flags arrived, so the PR tab opened and instantly bounced.

### The ratchet

An **AST walk**, not a grep — a renamed variable or a reordered argument cannot slip past it. Source-level rather than render-level on purpose: both invariants are a race between a fetch and a re-render, which a render test cannot force deterministically without freezing the very timing under test, and `fireEvent`-style tests cannot observe identity bugs at all. The repo already uses this shape for the same class (`task-detail-modal-tablet-width.test.ts`).

It also carries a **completeness case** — the guard is vacuous if no effect reads a role at all, so a third assertion fails if a refactor moves the reconciliation out from under the invariant.

### Verified to fail on the original defects

| mutation | result |
|---|---|
| drop the identity gate (read ungated flags) | **1 failed** / 2 passed |
| remove the pre-load guard from the PR-tab effect | **2 failed** / 1 passed |
| neither (shipped) | **3 passed** |

### Census

`TaskDetailModal.tsx` **30 → 5**; `--strict` exits 0. The five remaining are triaged, not skipped: one is an `agentState === "done"` **status** comparison (not a column at all — a census receiver misclassification), two are `task.column !== "archived"/"done"` inside a delete-confirmation guard, and two read `overlapBlockerTask.column` — another task's column, which needs *that* task's workflow resolved rather than this modal's.

### Note on overlap with #2696

#2696 is another worker on this same file (30 → 10). This one reaches 5 and now carries the async-hazard ratchet. Whichever you take, the other should be closed rather than merged after it — they touch the same lines and a second merge would conflict.

PR #2700, opened 2026-07-30
fleet: register-task-workflow-routes.ts 20 → 5 (review + terminal + wip lanes; 5 remaining triaged)

gsxdsm:
Claiming **`packages/dashboard/src/routes/register-task-workflow-routes.ts`** — 20 guards, the largest unclaimed cluster. This is **batch 1: the review lane**.

## Census before/after

| | before | after |
|---|---:|---:|
| `register-task-workflow-routes.ts` | **20** | **15** |
| repo backlog | 722 | **716** |

Baseline re-recorded — shrinks by exactly the converted count. `--strict` exits 0.

## Converted: 5, no new abstraction

This is a **server-side** file, so the app's `utils/columnRoles` helpers are unavailable. Used core's `resolveTaskLifecycleColumns` behind a local resolver that mirrors this file's **own existing** `resolveIntakeColumnForTask` — same shape, same fail-soft `try/catch`, same legacy fallback. That pattern was established here for the intake lane; this is its review-lane twin.

| line | site |
|---|---|
| 800 | `reengageInReviewTaskForUserComment`'s not-in-review early return |
| 3783 | recover-branch-binding route guard (error message de-hardcoded to "the review column") |
| 2703, 2736, 2742 | the retry handler's three review guards, sharing **one** resolution |

The retry handler already resolved a workflow IR, but block-scoped inside `if (retrySpecificationStatus)`, so its three review guards couldn't see it. One `retryReviewColumn` now serves all three rather than three separate reads on a manual button press.

`in-review` stays the fallback deliberately: each of these gates an error response or a re-engagement, so a store that cannot resolve a workflow keeps today's behaviour instead of dropping a card out of a review guard.

## The 15 remaining — next batches, grouped by lane

- **wip (`in-progress`)** — 1882, 5779, 5806, 5847. 1882 is *already* trait-resolved (`columnHasFlag(moveTargetIr, column, "countsTowardWip")`) with the legacy id as its fallback arm; it needs a marker, not a conversion.
- **complete / archived** — 2066 ×2, 3971 ×2, 4807. 4807 already partly resolves (`resolveColumn…`); the dependency-summary pair at 3971 reads a *dependency's* column, which needs a per-task resolution rather than this task's.
- **review, remaining** — 4101, 4668, 5771, 5806, 5829. Same family as this batch and mechanically identical; split out only to keep the diff reviewable.
- **hold (`todo`)** — 1149, a `listTasks` filter rather than a predicate: converting the QUERY is the separately-tracked instrument, not this count.

None are skipped for unclear meaning; they are batched for reviewability, and the two already-resolved sites want markers from whoever owns the ratchet rather than a count change smuggled into a conversion PR.

## Verification, including one thing I could not run

`pnpm lint` clean · dashboard `src` `tsc` clean · `--strict` exits 0.

**`vitest run src/__tests__/` (whole directory) aborts with SIGABRT** — an fsevents/worker crash in `planning-browser-e2e.test.ts`, a harness failure rather than a test result, and unrelated to this change. Rather than report a green I did not get, I verified the affected paths individually: `routes-task-retry-planning-column` + `routes-task-retry-stale-merge-status` + `routes-tasks` + `routes-planning` = **253/253**.


gsxdsm:
**This PR is now subsumed by main and I recommend closing it — with the evidence, because I have gotten this call badly wrong before.**

## The conflict is semantic, not mechanical

Every one of the 5 conflicting hunks is the same shape:

```diff
 HEAD (main)   const reviewGateColumns = await resolveReviewColumnsForTask(scopedStore, task.id);
               if (!reviewGateColumns.has(task.column)) {
 this branch   if (task.column !== await resolveReviewColumnForTask(scopedStore, task.id)) {
```

Main resolves a **set** and tests membership; this branch resolves a **single id** and tests equality. That difference is exactly the defect #2713 fixed: a workflow may put `mergeBlocker` and `humanReview` on **different** columns, and the single-id form classifies a card in the second review lane as outside review entirely.

So main's version is a strict superset of this branch's intent, and merging this branch's side would **reintroduce** a bug that has already been fixed and merged.

## Why I am confident it is empty rather than just conflicted

I checked for unique content before concluding, because I once recommended closing #2616 nine times on exactly this reasoning and was wrong — the production change there was upstream but two test files, 118 lines, were unique and not vacuous.

Here the branch's entire diff against main is:

```
packages/dashboard/src/routes/register-task-workflow-routes.ts
scripts/lib/lifecycle-column-census-baseline.json
```

**No test files.** The routes file's only content is the 5 guards above, all superseded. The baseline is a generated artifact. Resolving every conflict in main's favour leaves a diff of nothing.

## Recommendation

Close, and let #2713's membership resolver stand. If any of the 5 sites here is *not* covered on main, that is a real gap worth a fresh PR against the current resolver — but I could not find one: main carries 9 `resolveReviewColumnsForTask` call sites against this branch's 8 `resolveReviewColumnForTask`.

I have not force-pushed a resolution, because "resolve by discarding the whole branch" is a decision for its author rather than something to land quietly under a rebase.

gsxdsm:
Closing as **superseded**. `register-task-workflow-routes.ts` is already at **5** on main — the exact target this PR claimed (20 → 5) — reached by a better implementation than mine.

Concretely better, not just equivalent: main resolves the review lane as **membership over both traits**, where my version took `lifecycle?.review ?? "in-review"` — a single id. `resolveLifecycleColumns` returns the FIRST column carrying each role, so on a workflow that splits merge-blocking and human sign-off into two lanes, my shape classifies a card in the second as outside review entirely and suppresses comment re-engagement on it. That is the arity bug main's version explicitly fixes and mine would have shipped.

Main also lands the layering decision I did not make explicit: the route handlers deliberately do **not** import the app-side `columnRoles` predicates, because those take resolved trait flags and a route handler has a store and a task id. One idiom per layer.

Nothing here is worth salvaging — the local `resolveReviewColumnForTask` helper, the de-hardcoded operator strings, and the `DELIBERATE-LITERAL` v1-IR fallback note are all present on main in equal or better form.

PR #2705, opened 2026-07-30
fleet: moves.ts 15 → 2 (hot path; one hoisted resolution, net one FEWER than before)

gsxdsm:
Claiming **`packages/core/src/task-store/moves.ts`** — 15 guards, largest unclaimed. This is the core move path, every task move in the system, so the conversion is built to add **no work** to it.

## Census before/after

| | before | after |
|---|---:|---:|
| `moves.ts` column guards | **15** | **2** |
| repo backlog (this branch vs `origin/main`) | 693 | **679** |

*(Backlog figures don't compose across my open fleet PRs — each branch carries only its own reductions. 693 → 679 is this branch against main as measured, not a running total.)*

## Zero added cost, and actually one fewer resolution than before

`moveTaskInternal` **already** resolves the workflow IR unconditionally at line 400 — the `useWorkflow` gate is gone — and already derived a lifecycle from it ~400 lines later for the trait hooks. So one hoisted `moveLifecycle` immediately after the IR resolution serves all 14 guards, and the later local now **aliases** it instead of resolving a second time.

Net effect on the hot path: **one fewer `resolveLifecycleColumns` call than before this PR.**

## Converted: 14

6× `toColumn === "done"` → complete · 4× `fromColumn === "in-review"` → review · 1× `toColumn === "in-review"` → review · 2× `toColumn === "todo"` → hold · 1× `toColumn === "in-progress"` → wip

Every site keeps its legacy id as the fallback. `undefined` here means no IR on this path or a v1 column-less IR, and a move must behave **exactly** as before when there is no basis to resolve from — this is the transaction that arbitrates capacity, so "unchanged when unresolvable" is the requirement, not a nicety.

## Flagged, not converted: 1

**Line 309** — `task.column === "archived"` in the handoff-invariant check. It sits in a different function that runs **before** any IR resolution, so converting it would mean *adding* a resolution to a path that currently has none. That is a cost on the handoff path rather than the free reuse everything else here gets, so it wants a deliberate decision rather than my inclusion.

## Verification — the paths that matter, not just typecheck

Because this is the move transaction, `tsc` + lint is not sufficient evidence:

- **`pnpm test:gate` GREEN** — 158 + 10 + 487 + 71
- `workflow-capacity-invariant` + `move-path-equivalence` **7/7** (the in-transaction capacity gate lives in this file)
- `handoff-to-review-atomicity` **4/4**
- `store-movement` + `move-task-preserve-status` + `task-move-hard-cancel-ordering` + `transition-pending-and-status-clear` **16/16**
- `pnpm lint` clean · core `tsc` clean · `--strict` exits 0

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved workflow move handling across custom lifecycle configurations.
  * Operator hard cancellation now correctly routes intake-only workflows without a hold lane.
  * Preserved expected handoff behavior for engine moves that return tasks to the same lane.
  * Updated completion, review, and cancellation transitions to use the configured workflow lifecycle.


gsxdsm:
**Finding 2 (intake fallback) — real, fixed, and tested.** This one breaks a documented contract.

Both hard-cancel guards resolved their target as `moveLifecycle?.hold ?? "todo"`. A workflow may declare an intake column and **no hold column** — Coding (Ideas)'s `ideas` lane is the shipped example — so `"todo"` names a column that workflow does not declare, the comparison never matches, and the card moves while its merge request stays live and its work items are never cancelled.

A cancellation contract failing **open** is the worst direction available: the operator sees the card parked and believes the work stopped.

Precedence is now **hold → intake → legacy id**, the same order the replan target settled on in #2659, for the same reason.

**Tested against a live store**, because the bug is in *which column id* the guard compares against, and that id comes from the task's own persisted workflow. A mock handing back a lifecycle struct would assert my assumption about what `resolveLifecycleColumns` yields for an intake-only lineage — the substitution that has produced vacuous tests all through this program. The completion-handoff marker is cleared only inside the guarded block, so it witnesses exactly that the block ran. **Mutation: restoring `?? "todo"` fails it.** A paired negative asserts an ENGINE move into the same lane does *not* hard-cancel.

Two fixture traps, both found by failing first rather than by reasoning: `mergeOrchestration` is a **flag**, not a trait id (the trait is `merge`); and adjacency is graph-derived, so review → intake is only legal when the lineage puts them adjacent — my first fixture was rejected *before reaching the guard*, which reads exactly like "the fix does not work".

This PR shipped no tests, so this is also the first coverage on the conversion.

---

**Finding 1 (stale lifecycle snapshot) — I judge it NOT actionable here, and I want to be explicit rather than silently skip it.**

The concern is real in principle: `moveLifecycle` is resolved before the transaction, so a workflow selection changing mid-move would leave the guards comparing against stale column names. But:

1. It is **pre-existing and unchanged by this PR**. The same pre-transaction resolution fed the trait hooks before the conversion; this PR reuses the existing resolution rather than adding one — that is what made it zero-cost on the hot path.
2. Fixing it means moving lifecycle resolution inside the transaction, which changes the hot path for **every task move in the system**. That is a behavior change, and the fleet rules put behavior changes out of scope: flag and skip.
3. The window requires a workflow *selection* change interleaved with a move of the same task.

So: flagged, not guessed at, and not fixed under a conversion PR. It deserves its own change with its own measurement.

Verified: 2 PG tests green, core `tsc` clean, lint clean, gate green (487 + 158 + 10 + 71), `census --strict` exit 0.

PR #2708, opened 2026-07-30
fleet: project-engine.ts 12 → 8 (review lane; reverted one site tsc proved needs a loop restructure)

gsxdsm:
Claiming **`packages/engine/src/project-engine.ts`** (12). Batch 1: the **review lane**, 4 of its 7 review sites.

## Census before/after

| | before | after |
|---|---:|---:|
| `project-engine.ts` column guards | **12** | **8** |

## Converted: 4

Core's `resolveTaskLifecycleColumns` at each site — `store` is in scope at all four — with legacy `in-review` as the fallback.

| line | site |
|---|---|
| 3244 | the merge-pump task guard |
| 4726 | `wireAutoMerge`'s `taskMoved` handler — the auto-merge entry point |
| 4747 | the re-read guard before merge proceeds |
| 4858 | the merge-active guard |

## I reverted one I had already written — the useful part of this PR

Line **646** sits inside a **non-async `flatMap` callback**, so an `await` there does not compile, and `task` is possibly null on that path as well. Making it work means turning the callback async and restructuring the flatMap into Promise handling — that changes the shape of a queue-scanning loop rather than converting a guard.

`tsc` caught it immediately (TS1308 + TS18047) and I removed it rather than reshaping the loop. **"Behaviour changes out of scope" covers control flow, not just semantics** — a mechanical pass that quietly turns a synchronous scan into an async one is exactly the kind of change the fleet rules exist to keep out of a conversion diff.

## Flagged, not converted: 8

- **646** — the `flatMap` callback above; needs the loop restructured, so it wants a decision.
- **2555** — a **pure synchronous predicate** taking `{ column: string; … }`. No store, no task id, so converting means threading a resolution through its signature and every caller.
- **2287, 2957** — store availability at those points is not established. Not guessed.
- **4214, 4280, 4345, 4649** — the `done`/complete family. A coherent batch 2, split out to keep this diff reviewable rather than because they're difficult.

## Verification — the sensitive paths, not just typecheck

This file owns the merge pump and the auto-merge wiring, so `tsc` alone isn't evidence:

- **`pnpm test:gate` GREEN** — 158 + 10 + 487 + 71
- `merge-single-flight-invariant` + `project-engine` **210/210**
- `pnpm lint` clean · engine `tsc` clean · `--strict` exits 0


gsxdsm:
Closing as **superseded**. `project-engine.ts` is at **5** on main with 8 `resolveTaskLifecycleColumns` call sites; this PR claimed 12 → 8. Main went further and got there first.

For the record, the one thing this PR contributed that is worth restating rather than salvaging: I attempted `project-engine.ts:646` and **reverted it** after `tsc` rejected it — TS1308 (`await` in a non-async `flatMap` callback) and TS18047 (`task` possibly null). Converting that site needs the surrounding loop restructured, which is not a vocabulary change. If main's remaining 5 include it, that is the reason it is still there.

PR #2709, opened 2026-07-30
fleet: store.ts 12 → 11 + names the sync-dependency-loop class blocking ~10 sites across 3 clusters

gsxdsm:
Claiming **`packages/core/src/store.ts`** (12). One conversion and a triage — because **10 of the 12 share a single blocking shape** that is worth naming once rather than rediscovering per file.

## Census before/after

| | before | after |
|---|---:|---:|
| `store.ts` column guards | **12** | **11** |

## Converted: 1

**1386** — the in-review guard inside `withTaskLock(id, async () => …)`. Already async, and `this` **is** the store, so `resolveTaskLifecycleColumns(this, task.id)` resolves the review role with `in-review` as the fallback. Import added; nothing else in the method changes.

## The blocking class — 6 sites, and it is not specific to this file

**1772, 1791 ×2, 1874 ×2, 1916, 1917, 1933** all read **another task's** column — a dependency's, a blocker's, an overlap candidate's — inside **synchronous callbacks over a prefetched `taskById` map**:

```ts
const unresolvedDeps = (task.dependencies ?? []).filter((depId) => {
  const dep = taskById.get(depId);
  return dep && !dep.deletedAt && dep.column !== "done" && dep.column !== "archived";
});
```

This is not a substitution. Each dependency may belong to a **different workflow**, so the role must be resolved *per dep* — N async resolutions inside a sync `filter`, on a path that deliberately prefetches into a map precisely to avoid per-item I/O.

Two honest options:

1. **Prefetch lifecycle columns alongside `taskById`** and pass a resolved map into these predicates. Keeps them synchronous, one resolution per distinct workflow rather than per dep. This is the one I'd argue for.
2. Accept per-dep resolution and make the callbacks async — changes the shape of dependency evaluation.

Both are design changes with real cost, so this is flagged rather than guessed.

**The same shape appears in at least two other clusters I've worked**: `TaskDetailModal`'s `overlapBlockerTask.column` (#2696) and `register-task-workflow-routes`' dependency-summary pair (#2700), both flagged for this exact reason. **Worth one decision covering all three** rather than three separate judgement calls by three workers.

## Also flagged: 3

**1610** and **1739** — enclosing-scope async-ness and store access not established at those points, so not guessed. **1933** belongs to the sync-filter family above.

## Verification

`pnpm test:gate` **GREEN** (158 + 487 + 10 + 71) · `pnpm lint` clean · core `tsc` clean · `--strict` exits 0.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated failed pre-merge review bypass validation to support custom workflow boards.
  * Tasks can now bypass the step when placed in the board’s configured review lane.
  * Improved error messages to identify the correct review column when bypassing is not allowed.


gsxdsm:
Real, and I'd argue it is **worse than an unconverted guard**.

The check was converted to the resolved review lane while the rejection still said `in-review`. On a custom board the operator is refused and then told to move the card to a column their board does not have — through both the CLI and the dashboard, with nothing in the error revealing the real target.

An inert guard fails *visibly*: the operator knows something is wrong. This one refuses **correctly** and then misdirects, so their next three attempts are all wrong for a reason the product told them.

Resolved once into a local, so the check and its message cannot drift apart again — the inline resolution inside the comparison is precisely what let them diverge.

**Test:** a card on a board whose review lane is `validating`, sitting elsewhere, must be refused with `must be in 'validating'`. **Mutation: hardcoding the message back fails it.** The nine existing cases all run on the default board, where the resolved column *is* `in-review` — so none of them could ever have seen this.

Rebased onto current main; baseline regenerated with JSON validity asserted rather than assumed.

Verified: 10 tests green, core `tsc` clean, lint clean, gate green (487 + 158 + 10 + 71), `census --strict` exit 0.

PR #2710, opened 2026-07-30
docs(solutions): converting a column literal to a role makes it async — the four forms that ship green

gsxdsm:
Four review rounds across `TaskCard.tsx` and `TaskDetailModal.tsx` each found a **real defect**. None was in the conversion itself — every one came from the same property change. The fleet has ~600 guards left to convert against the same helpers, so this is written down rather than left in four commit messages.

## The property that changes

```ts
task.column === "in-progress"   // stable for the lifetime of the render tree
isWipColumn                     // derived from fetched trait flags — CHANGES after first paint
```

Column trait flags arrive from a board-workflows fetch. Until it lands they are `undefined` and every role helper falls back to the legacy id. So a converted role is `false`, then `true`, within one mounted component.

## The four forms

| | form | symptom |
|---|---|---|
| 1 | **stale memo** — deps still keyed only on `task.column` | timers, labels, completion dates frozen at first-paint values (4 instances in TaskCard) |
| 2 | **frozen `useState` initializer** | the section does not start collapsed — it *appears later, already collapsed*, on a card nobody touched |
| 3 | **eager action on a guess** — effect mutates state before flags resolve | a tab opens and instantly bounces; the correction never lands because the action destroyed the state it would have corrected |
| 4 | **stale identity** — flags resolved, but for the *previous* entity | roles resolve from another task's workflow: confidently wrong rather than merely stale |

**Form 4 defeats the obvious fix for form 3.** A `metadata === null` guard asks whether data *loaded*, not whether it describes the entity currently open — and it only appears in components that stay **mounted across entity changes**, which is why TaskCard never showed it and the modal did.

## Why a doc rather than four commit messages

All four ship **green**: types pass, existing tests pass, and the **default board behaves identically** — because on the default lineage the legacy fallback and the resolved role agree. They diverge only on a **renamed board**, which is precisely the case the conversion exists to support.

So the failure mode is: census count reaches zero, everything is green, and the feature is broken exactly where the programme was meant to fix it. A reviewer catching these one at a time is the expensive path, and it has now cost four rounds on two files.

Also relevant: **this repo has no `react-hooks/exhaustive-deps` rule**, so form 1 has no automated backstop at all.

## Contents

A checklist a converter can run against a component file, and the concrete fix shape for each form — including tagging fetched metadata with the id it describes, and applying that guard to the **role bindings** rather than only the effects (reordering effects fixes the call sites you noticed and leaves the bindings stale for everything else).

Follows the convention already established by the engine-side scoping note in `architecture-patterns/fleet-self-healing-cluster-scoping.md`, which records the equivalent hazard for sync workflow reads.

## Verification

`pnpm test:gate` green (10 / 158 / 487 / 71). `pnpm check:lifecycle-columns` exits 0. `pnpm lint` clean.

Docs-only; no changeset.

### DIFF
--- a/docs/solutions/ui-bugs/async-resolved-column-roles-in-components.md
+++ b/docs/solutions/ui-bugs/async-resolved-column-roles-in-components.md
@@ -8,10 +8,18 @@ applies_when: converting lifecycle-column literals in React components that read
 
 # Converting a column literal to a role turns a stable value into an async one
 
-Recorded 2026-07-30 after converting `TaskCard.tsx` (42 → 0) and `TaskDetailModal.tsx` (30 → 5).
-Four separate review rounds each found a real defect in those two files. None was in the conversion
-itself — every one came from the same property change, and they are cheap to prevent and expensive
-to find one at a time.
+Recorded 2026-07-30 while converting `TaskCard.tsx` and `TaskDetailModal.tsx`. Four separate review
+rounds each found a real defect in those two files. None was in the conversion itself — every one
+came from the same property change, and they are cheap to prevent and expensive to find one at a
+time.
+
+**Status of those conversions, stated because an earlier draft of this line got it wrong.** It read
+"after converting `TaskCard.tsx` (42 → 0) and `TaskDetailModal.tsx` (30 → 5)" as accomplished fact.
+Neither reduction has landed: on `main` both files still measure 42 and 30, and the census baseline
+in this very PR records the same. The work is in flight in #2688 and #2698. The hazard below is real
+regardless — it was found BY those conversions, and it is what the four review rounds were about —
+but a solutions doc that reports unlanded numbers as history is worse than no doc, because the next
+reader has no way to tell which of its claims were measured.
 
 ## The property that changes
 
--- a/docs/solutions/ui-bugs/async-resolved-column-roles-in-components.md
+++ b/docs/solutions/ui-bugs/async-resolved-column-roles-in-components.md
@@ -1,0 +1,106 @@
+---
+category: ui-bugs
+module: dashboard/components
+tags: [lifecycle-columns, column-roles, fleet, react-hooks, async-state]
+problem_type: conversion-hazard
+applies_when: converting lifecycle-column literals in React components that read column trait flags
+---
+
+# Converting a column literal to a role turns a stable value into an async one
+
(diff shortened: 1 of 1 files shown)


gsxdsm:
Real, and I verified it rather than taking either side's word.

**Measured on `origin/main` after rebasing onto it just now:**

| file | doc claimed | actually |
|---|---|---|
| `TaskCard.tsx` | 42 → 0 | **42** |
| `TaskDetailModal.tsx` | 30 → 5 | **30** |

The census baseline committed in this very PR records the same two numbers, which is what made the contradiction self-evident. Both conversions are still open in #2688 and #2698.

The hazard the doc describes is unaffected — it was found *by* those conversions and is what the four review rounds were about. What was wrong was the **tense**.

I corrected it in place and recorded the error rather than quietly rewording, because that is the specific failure mode a solutions doc cannot afford: **a doc that reports unlanded numbers as history is worse than no doc.** The next reader has no way to tell which claims were measured and which were anticipated, so one wrong line discredits the whole file rather than just itself. The correction now names what was claimed, what is true, and where the real work is.

Verified: lint clean, gate green (487 + 158 + 10 + 71), `census --strict` exit 0, and the two counts above re-measured on the rebased branch.

gsxdsm:
Correct, and thank you for the fix already pushed to this branch — I have built on it rather than duplicating it.

Two things in this revision:

**The numbers are now framed as the change that produced the defects**, not as current census state. The reductions were in flight when this was written and the baseline still recorded 42 and 30, so quoting them as landed was misleading exactly where a reader would check.

**A fifth form has been added, and it is the one most worth recording** because it is *caused by* the fix for form 4. Guarding on entity identity means writing `null` when a lookup returns nothing or fails — which is indistinguishable from "has not resolved yet". The guard never opens, reconciliation never runs, and the invalid state persists **forever**.

Resolution has three states, not two: unresolved, resolved-with-data, and **resolved-empty**. The third still identifies the entity, so consumers know the answer landed and should fall back to the legacy id — a real answer for a workflow that declares nothing. The failure path has to settle the same way, or a lost fetch leaves the UI waiting on it indefinitely.

That yielded the rule generalising all five, now in the doc: **every fix in this list can introduce the opposite failure — check that both directions terminate.** Four of the five were found by review rather than by me, and this fifth one existed *because* I fixed the fourth without asking what the fix would break.

`pnpm lint` clean, `pnpm check:lifecycle-columns` exits 0. Docs-only.

PR #2713, opened 2026-07-30
fleet: register-task-workflow-routes.ts 20 -> 5

gsxdsm:
## Census before / after

| | before | after |
|---|---:|---:|
| `register-task-workflow-routes.ts` | **20** | **5** |

I handed this cluster off in an earlier turn with an analysis rather than a conversion. Coming back to it with the analysis already done made it tractable.

## Converted with the file's own idiom

This file already had `resolveIntakeColumnForTask` / `resolveWipColumnForTask` / `resolveReboundColumnForTask`: resolve the column **id** from the task's workflow, fall back to the legacy id when the IR cannot be read. I added the three roles it was missing — review, complete, archived — in that same shape.

**Deliberately not the `columnRoles` predicate helpers** used in `packages/dashboard/app`. Those take resolved trait *flags*, which a route handler does not have — it has a store and a task id, and must do an async lookup. Importing them here would mean fetching flags per request to answer a question this file already answers a simpler way. One idiom per layer.

Every converted handler **resolves once and reuses**, so two checks in the same request cannot disagree about which column is the review lane. The retry handler had three separate review checks and now shares one resolution.

## Also fixes an inversion

`isArchived` ORed the legacy id with the resolved trait **unconditionally**, so a column merely *named* `archived` counted as archived even when its own workflow said otherwise. Same pattern previously found in `TaskContextMenu` and `isPreExecutionHoldColumn`. Now flags-first, id as fallback.

## The five survivors, each with a reason

| count | line | why |
|---:|---|---|
| 1 | 1193 | `tasks.filter(t => t.column === "todo")` — a **list** path. Per-task IR resolution is N store reads on board load; the site already carries a note measuring that cost. Needs the hold column resolved per *workflow* from the board payload — a real change, not a rename. |
| 1 | 1926 | **Already flags-first**: `moveTargetIr && declaresColumns ? columnHasFlag(...) : column === "in-progress"`. The literal *is* the documented no-IR fallback. |
| 1 | 4872 | The fallback arm of the `isArchived` fix above — same shape, deliberately kept. |
| 2 | 4024 | `depTask.column` — a **dependency's** column, not the task's. Resolving it means fetching that task's IR per dependency; out of scope. |

So three of the five are correct as they stand, and two are genuinely deferred.

## Verification

`plan-approval-intake-column` + `stranded-refinements-routes` 13/13. `pnpm test:gate` green (10 / 158 / 487 / 71). `pnpm check:lifecycle-columns` exits 0 with the baseline re-recorded here. `tsc -p packages/dashboard/tsconfig.json` clean. `pnpm lint` clean.

Typecheck was run after **every** batch rather than once at the end — with 15 edits across async handlers in a 6000-line file, a single late typecheck would not tell me which batch broke it.


gsxdsm:
Both real. The first is the most instructive defect this whole conversion effort has produced, so I want to name the shape rather than just fix it.

**1. Custom review bypassed the open-PR guard.** `isBackwardMoveBlockedByOpenPr` bails with `if (fromIndex < 0) return false`, and `COLUMNS.indexOf` returns **-1** for any custom column id. Before this PR the re-engagement gate only admitted a task literally in `in-review`, so the index was always valid. Converting the gate to resolve the review **role** started admitting custom columns — and the guard then **silently stopped protecting them**. The task moved out of review with an open PR.

The general shape: **a conversion that widens what reaches a downstream legacy-index check has to carry that check with it.** Widening the gate was correct; leaving the consumer keyed on legacy ordering turned a working guard into a no-op for exactly the boards the conversion was for.

Fixed by passing the review lane's index — the caller has already established `task.column` **is** this workflow's review lane, so it occupies the review *position* for an ordering question whatever it is named. I also pinned the guard's escape hatch in `routes-pull-requests.test.ts`: a negative index **allows** the move. Documented as a hazard for the next converter, not as desired behaviour for custom boards.

**2. Terminal membership dropped valid columns.** Correct, and it was a flaw in the *shape* of my resolvers rather than a slip. `columnsWithFlag(ir, "complete")[0]` picks one column; a workflow declaring more than one complete or archived column then rejected tasks in the second with a 409.

Worth stating why it was easy to write: the single-id shape **is** right for this file's older resolvers, which answer *"where should this card go"* — a move target must be exactly one column. It is wrong for *"is this card already terminal"*, which is a membership question. Same naming convention, opposite arity requirement. Replaced both with `resolveTerminalColumnsForTask` returning the full set, with the legacy pair as the degraded fallback.

Verification: `plan-approval-intake-column` + `stranded-refinements-routes` 13/13, `routes-pull-requests` 14/14, `pnpm test:gate` green (10 / 158 / 487 / 71), `pnpm check:lifecycle-columns` exits 0 with the baseline re-recorded, `tsc` clean, `pnpm lint` clean. Rebased onto current `origin/main`.

gsxdsm:
Correct — and the part worth recording is that this is the **same arity bug** I fixed for the terminal columns one review round earlier and did not generalise.

`mergeBlocker` and `humanReview` can sit on **different** columns. The resolver took `columnsWithFlag(ir, "mergeBlocker")[0] ?? columnsWithFlag(ir, "humanReview")[0]` — one id — so a workflow that splits them into a merge lane and a separate sign-off lane classified a task in the second as **outside review entirely**. Comment re-engagement was suppressed on a card sitting in human review.

`resolveReviewColumnsForTask` now returns the full set, and all eight call sites test membership.

**The rule, now written at the resolver so there is no third instance:**

| shape | question | correct for |
|---|---|---|
| single id | *"where should this card **go**?"* | a move target — must be exactly one column |
| **set** | *"is this card **already** there?"* | membership |

**Every resolver compared against `task.column` is the second kind.** This file's older resolvers — `resolveIntakeColumnForTask`, `resolveWipColumnForTask`, `resolveReboundColumnForTask` — are the first kind and are correct as they stand, because they feed moves rather than comparisons. The shared naming convention hides that difference, which is exactly how I inherited it twice in two rounds.

Verification: route suites 27/27 (`plan-approval-intake-column`, `stranded-refinements-routes`, `routes-pull-requests`), `pnpm test:gate` green (10 / 158 / 487 / 71), `pnpm check:lifecycle-columns` exits 0 with the baseline re-recorded, `tsc` clean, `pnpm lint` clean. Rebased onto current `origin/main`.

One coordination note: **#2701 (`fleet/routes-review-lane`) is touching this same file** — its latest commit is about the retry gate conversion, which this PR also converts. Worth landing one before the other rather than in parallel.

PR #2715, opened 2026-07-30
fleet: github-tracking-comments.ts 9 → 3 — and the sync-filter class is now in a FOURTH file (~25 sites on one decision)

gsxdsm:
Claiming the **github-tracking pair**. This converts the comments half; the reconciler half is the flagged class, with the evidence below.

## Census before/after

| | before | after |
|---|---:|---:|
| `github-tracking-comments.ts` | **9** | **3** |

## Converted: 6

The `event.to === "in-progress"` / `=== "done"` sites in `handleTaskMoved`, to the **wip** and **complete** roles. One resolution, placed **immediately after the tracking-enabled gate** — so a move on an **untracked** task pays nothing, which is most moves in most projects.

## Deliberately not converted: 2 — the ordering is the reason

```ts
if (event.to !== "in-progress" && event.to !== "done") return;   // line 232
```

This runs **before** the tracked-task gate. Converting it moves the resolution ahead of that gate and makes **every task move in the project** resolve a workflow just to decide the task has no GitHub issue. That's a real cost on the hottest event in the system, to convert a guard whose only job is a cheap filter. Recorded at the site.

## The remaining 1

**Line 165** — `transition === "done"` inside `formatTrackingComment`, a **pure formatter** with no store and no task. Same shape as `project-engine.ts:2555`. Threading a resolution into a formatter to pick a string is the wrong trade.

## `github-tracking-reconciler.ts` (9) — not claimed here, and here is why

All nine are:

```ts
.filter((task) => task.column === "done" || task.column === "archived")
```

Synchronous filters over task **lists**, where per-task resolution is N awaits inside a sync predicate.

**This is the fourth file with that exact shape** — after `store.ts` (#2709), and the dependency pairs in `TaskDetailModal` (#2696) and `register-task-workflow-routes` (#2700). By my count **roughly 25 sites across four files now wait on one decision**:

1. **Prefetch lifecycle columns alongside the task list** and pass a resolved map into these predicates — keeps them synchronous, one resolution per *distinct workflow* rather than per task. This is the one I'd argue for.
2. Make the predicates async and accept per-task resolution.

It's a design change, and four workers guessing separately is exactly how two halves of one rule drift apart. One decision covers all of them.

## Verification

`pnpm test:gate` **GREEN** (158 + 10 + 487 + 71) · `github-tracking-comments` + `github-issue-comment` **81/81** · `pnpm lint` clean · dashboard `tsc` clean · `--strict` exits 0.


gsxdsm:
Real, fixed, and it surfaced a second instance of the same conflation.

**The ordering bug.** `handleTaskMoved` tested `event.to` against the literal `in-progress`/`done` **first**, then resolved wip/complete below. On a renamed board the literal never matched, so the function returned before reaching any resolved code — the tracking comment was silently skipped. No error, no log; a tracked issue simply stops being updated.

The previous note defended that ordering on cost: converting the early return would make every task move in the project resolve a workflow. **Sound about cost, wrong about correctness — a conversion that cannot be reached is not a conversion.**

**Reordering satisfies both.** The tracking-enabled check is a plain property read on the event's own task, so moving it above the resolution still short-circuits every *untracked* move — which is the population the cost argument was actually about. Only tracked tasks resolve a workflow, and those are exactly the ones that need the answer.

**A second conflation, found by the compiler.** `formatTrackingComment`'s second parameter is a **transition kind**, not a column id. `event.to` only type-checked there because the literal early return had narrowed it to the two legacy ids — the id and the kind coincide on the default board and diverge on a renamed one. The role is now passed explicitly. My first reorder failed to compile, which is the only reason I looked; the literal guard had been doing double duty as a type narrowing.

**Measured: restoring the literal early return left all 101 existing cases green.** Nothing in that suite drove a workflow at all — `MockStore` has no selection methods, so every case resolved to the legacy fallback and the conversion was **untestable by construction**. Two cases now supply a renamed IR: a move to the renamed wip column posts a comment, a move out of it does not. Mutation: the literal gate now fails 1 of 50.

Baseline re-recorded (`github-tracking-comments.ts` 3 → 1) — my own #2679 enforcement caught the stale allowance, which is the second time today it has fired on me.

Verified: 171 tracking tests green, dashboard `tsc` clean, lint clean, gate green (487 + 158 + 10 + 71), `census --strict` exit 0.

gsxdsm:
## Scope grew: this PR now lands **both halves of the tracking pair**

Pushed `gitlab-tracking-comments.ts` **4 → 1** onto this branch, plus a code-level parity ratchet. Converting GitHub and leaving GitLab is the FN-6115 → FN-6118 → FN-6123 shape — one behaviour in two modules with only one converted — arriving through the provider-parity door instead of the desktop/mobile one, so the two halves belong in one merge rather than two.

**What the GitLab half cost:** `handleTaskMoved` compared `event.to` to `"in-progress"` and `"done"`, so on a renamed board GitLab tracking silently posted **no comments** — the linked issue or MR just stopped being updated. No error, no log line.

Both halves now read the same, including the ordering: the cheap tracked-item read runs first and short-circuits every untracked move, so only tracked tasks resolve a workflow. Both leave **exactly one** literal — the pure formatter's caller-chosen `transition` mode — in the same place for the same reason.

### A code-level parity ratchet

`gitlab-parity-inventory-documentation.test.ts` already existed, but it only checks that the parity **inventory doc** mentions each surface. It cannot see that two modules implementing one of those surfaces have drifted. Two new cases assert both services resolve lifecycle columns and neither compares `event.to` to a legacy lane id, with comments stripped first so the FNXC notes that necessarily quote those literals neither satisfy nor trip it.

**Injection proof from both sides:** reintroducing the literal in gitlab fails with `gitlab-tracking-comments.ts still compares event.to to a legacy lane id`; doing it in github fails identically for github.

### Two things the toolchain caught, not me

1. **tsc found a narrowing I removed.** The old early return (`event.to !== "in-progress" && event.to !== "done"`) was not only a guard — it also narrowed `event.to` to the formatter's `"in-progress" | "done"` parameter type. Comparing against resolved variables cannot narrow a string, so the call was rejected. That is the compiler surfacing a real consequence: on a renamed board `event.to` was never one of those two words, so passing it through was **already** the wrong value for a parameter meaning "which comment shape".

2. **The census caught my fix introducing a new site.** My first version named the discriminant as a string and then asked `transition === "done"` — correctly counted as a new column guard (same shape as the `mode === "done"` pair in `useTaskDiffStats`). Switched to a boolean, which answers the same question without a literal comparison at all. Better than adding a site and then exempting it with a marker.

### Verification

Gate **GREEN** (158 + 10 + 487 + 71) · **142 passed** across the gitlab and github tracking suites · dashboard `tsc` clean · `pnpm lint` clean · census `--strict` exits 0.

Behavioural revert proof: restoring the literal lane test fails both new renamed-lane cases with 0 fetch calls. The three new cases supply a workflow reader on the store fake — the 12 existing cases omit it and assert the legacy fallback, which is exactly why none of them could have caught this (see `docs/solutions/test-failures/optional-flags-seam-hides-unconverted-column-guards.md`, #2748).

PR #2716, opened 2026-07-30
test(engine): two files missed the logger-mock `debug` sweep (38 red → 0) + a deleted API still pinned

gsxdsm:
## What was red

| File | Failures | Error |
|---|---:|---|
| `notification/__tests__/notification-service.test.ts` | 26 | `schedulerLog.debug is not a function` |
| `runtimes/__tests__/child-process-worker.test.ts` | 12 | `runtimeLog.debug is not a function` |

`debug` is part of the logger surface (`logger.ts:25`) — the channel the noisy-line demotion moved subsystem chatter onto, gated on `FUSION_DEBUG`. A mock that omits it throws on the **first** demoted call, failing every case in the file for a reason unrelated to what any of them assert. These two were missed by the earlier sweep across 27 engine files.

## Measured

| Check | Result |
|---|---|
| the two files | 38 failed → **38 passed** |
| `debug` removed from the mocks again | **38 failed** — the entries are load-bearing |
| `pnpm test:gate` | **726 passed** |
| `pnpm lint` | clean |

Census unchanged — test files only.

## Two real drifts under the mock gap

Both re-pointed at what the product actually does, not relaxed:

**1. Suppression lines are `debug`, not `log`.** `notification-service.ts` routes all five of its `"suppressed ..."` messages through `schedulerLog.debug`. Two assertions looked on `.log` — where the product no longer writes. Verified by grepping the product for the message before editing the test, rather than assuming the mock was the whole story.

**2. `centralCore.getGlobalConcurrencyState` is DELETED, not missing.** The cross-project cap was dropped deliberately (`central-core.ts:2124`, `FNXC:CapacityModel 2026-07-28-23:30`) together with `updateGlobalConcurrency`, `acquireGlobalSlot`, `releaseGlobalSlot` and the `concurrency:changed` event — capacity is two numbers **per project** now. That comment also records the slot pair was already dead: no production caller ever invoked it, so `currentlyActive` was never incremented by real work.

The worker's stub provides `getLiveRunningAgentCounts` and `recordTaskCompletion`, so the case is re-pinned to the former. It had been pinning an API the product removed on purpose — the assertion would have kept "passing" a shape that no longer exists if the stub had happened to retain a same-named field.

## Flagged, deliberately not changed

`ipc/__tests__/ipc-host.test.ts` and `ipc/__tests__/ipc-worker.test.ts` carry the **same incomplete logger mock** but are currently green (56 passed) because no demoted line is reached on their paths.

Adding `debug` there cannot be shown to fail today, so it is recorded here rather than slipped in as an unfalsifiable edit. They go red the moment any code they exercise demotes a line — which is how the 29 files before them broke. Found by scanning every engine logger mock for the pattern, not by guessing.

### DIFF
--- a/packages/engine/src/__tests__/logger-mock-completeness.test.ts
+++ b/packages/engine/src/__tests__/logger-mock-completeness.test.ts
@@ -1,0 +1,73 @@
+import { readFileSync } from "node:fs";
+import { execFileSync } from "node:child_process";
+import { join, relative } from "node:path";
+import { describe, expect, it } from "vitest";
+
+/*
+FNXC:EngineTests 2026-07-31-05:20:
+Ratchet for the recurring incomplete-logger-mock failure class.
+
+THE CLASS. `createLogger` returns four methods — `log`, `debug`, `warn`, `error` (logger.ts:20-27).
+A test that mocks `../logger.js` with only three throws "<log>.debug is not a function" on the FIRST
+demoted call, which fails EVERY case in that file for a reason unrelated to what any of them assert.
+It has now happened 31 times: 27 files in one sweep, then notification-service (26 cases) and
+child-process-worker (12) missed by that sweep, then ipc-host and ipc-worker found latent.
+
+WHY A SOURCE SCAN RATHER THAN A RUNTIME CHECK. The gap only bites when a demoted line is actually
+reached, so a mock can sit incomplete and green for months — ipc-host and ipc-worker were exactly
+that: 56 passing tests with a three-method mock. Nothing at runtime can fail on a call that does not
+happen yet, so the guard has to read the source.
+
+DELIBERATELY NARROW. Only object literals that already look like a logger (`log: vi.fn()` plus
+`warn:` and `error:`) are checked, so an unrelated `{ log: ... }` shape is not dragged in. A file that
+mocks the logger some other way is not policed here — this pins the copy-paste shape that actually
+recurred, not every conceivable mock.
+*/
+
+const REPO_ROOT = join(__dirname, "..", "..", "..", "..");
+const REQUIRED_METHODS = ["log", "debug", "warn", "error"] as const;
+
+/** Engine test files that mock the logger module, via git so untracked scratch files are ignored. */
+function loggerMockingTestFiles(): string[] {
+  const out = execFileSync(
+    "git",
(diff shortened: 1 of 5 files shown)


gsxdsm:
## Update: the two files I flagged are now FIXED, with a ratchet that makes the fix falsifiable

The original PR body flagged `ipc-host.test.ts` and `ipc-worker.test.ts` as carrying the same incomplete logger mock but **left them alone**, because adding `debug` to a mock whose demoted line is never reached cannot be shown to fail — an unfalsifiable edit.

`84ec5aa150` makes it falsifiable rather than guessing: a source-scanning ratchet (`logger-mock-completeness.test.ts`) that fails **today** on any logger-shaped mock missing a method of the real four-method surface (`log`, `debug`, `warn`, `error` — `logger.ts:20-27`). With the guard in place, fixing those two is a real change with a test behind it.

**Why a source scan and not a runtime check.** The gap only bites when a demoted line is actually reached, so a mock can sit incomplete and green for months. `ipc-host` and `ipc-worker` were exactly that: **56 passing tests** with a three-method mock. Nothing at runtime can fail on a call that has not happened yet.

**The class has cost 31 files so far** — 27 in one sweep, then `notification-service` (26 cases) and `child-process-worker` (12) missed by that sweep, then these two found latent by scanning for the pattern rather than waiting for them to break.

### Measured: the ratchet fires on each mock independently

| mutation | result |
|---|---|
| `debug` removed from `ipc-host` | **1 failed** / 1 passed |
| `debug` removed from `notification-service` | **1 failed** / 1 passed |
| `debug` removed from `child-process-worker` | **1 failed** / 1 passed |
| all restored | **96 passed** across the five files |

Worth recording: two of those three mutations first reported *"no change"* because my initial regex missed their single-line mock form, so those runs proved nothing until re-run with a matching pattern. A silently-inapplicable mutation is indistinguishable from a passing guard, which is the same trap that made my first `archive-lifecycle-2` proof worthless earlier in this program.

### The scan guards itself

A first case asserts the scan matches **more than 20 files**. Without it, deleting the search pattern or moving the files would turn the ratchet into a no-op that reports success while checking nothing.

### Deliberately narrow

Only literals already shaped like a logger (`log: vi.fn()` plus `warn:` and `error:`) are checked, so an unrelated `{ log: ... }` object is not dragged in, and a file that mocks the logger some other way is not policed. This pins the copy-paste shape that actually recurred, not every conceivable mock.

`pnpm typecheck` exit 0 across all packages, lint clean, gate **726**.

PR #2718, opened 2026-07-30
fleet: TaskContextMenu.tsx 9 → 4 — an unconditional id OR was opening the merge lane on mid-flight cards

gsxdsm:
Claiming the **TaskContextMenu** cluster. Five of the nine convert; one is a behaviour change I flagged and left **counted**; three are the documented load-window fallbacks.

## Census

| | before | after |
|---|---:|---:|
| `TaskContextMenu.tsx` | **9** | **4** |

Baseline shrinks by **exactly the 5 converted** — no reclassification (see the last section, which is the point).

## What the guards were actually doing

Two of them **ORed the legacy id with the traits unconditionally**:

```ts
return column === "in-review" || flags?.mergeBlocker === true || flags?.humanReview === true;
```

That is not a fallback. A resolved column merely **named** `in-review` or `done` answered true even when its traits said mid-flight — which is what a project gets by *repurposing* a default column instead of deleting it. `isReviewColumn` gates `getTaskReviewAction`, so the consequence is concrete:

> **a card mid-implementation, in a column named `in-review`, was offered a live "Merge & Close".**

Now flags-first via `isReviewColumnRole` / `isCompleteColumnRole`. Same shape as the inversion already recorded in this file for `isPreExecutionHoldColumn`.

`isDoneOrReview` keeps its `complete && !archived` shape deliberately — an archived card is finished but not *completed*, and Refine must not resurrect intentionally archived work. Flagless reduces to `column === "done"` exactly as before.

## One conversion makes an affordance APPEAR

The **FN-7720 bypass** guard tested `task.column === "in-review"`. A workflow that renames its review lane strands cards on a failed pre-merge step in exactly the same way, but the id comparison **hid the only escape hatch** — the operator saw a permanently unmergeable card and no way to release it. `hasFailedPreMergeReviewStep(task)` remains the real precondition, so the action still cannot appear on a card that is not stranded (asserted).

## Revert proof — and a first draft that didn't have one

| reverted | result |
|---|---|
| predicates back to ORing the id | **3 fail**, incl. `expected { id: 'merge', … } to be undefined` |
| bypass guard back to the id | **1 fail**, `expected [ 'respecify', 'pause', 'delete' ] to include 'bypass-review'` |

Worth recording: **my first draft of these cases passed with the change reverted.** I reused the file's existing `menuDecisions`, which wires only `onPlan` — so the handler-gated `refine` branch I was targeting was unreachable, and `not.toContain("refine")` was vacuously true. The rename property is also the wrong instrument here: the trait arm of the OR answers on a renamed board, so a pure rename case passes before *and* after. The delta only shows in the converse direction — hold the id at a lifecycle name and make the decision follow the traits. Both are recorded in the test.

## Flagged and left — deliberately still counted

The Respecify guard **ANDs** both signals (`task.column !== "archived" && flags?.archived !== true`), i.e. fail-closed. Converting it would **widen** the affordance: a column named `archived` whose traits disagree would start offering Respecify into a route that rejects archived sources — the dead affordance its own note says it prevents. Behaviour change, out of fleet scope, so it stays for whoever owns Respecify to decide fail-closed vs trait-authoritative.

I did **not** apply the census exemption marker to it, so the count keeps pointing at the line. Which surfaced a footgun worth passing on:

> **writing the exemption marker's name in prose is enough to exempt the guard.** The classifier substring-searches a comparison's leading comments, so my first note — explaining why I was *not* exempting it — silently exempted it, and the count fell to 3 instead of 4. A comment changed a measurement. Anyone documenting why a site was *not* exempted needs to avoid spelling the token.

The remaining 3 are the `VALID_TRANSITIONS` and `isBackwardsLabel` load-window fallbacks, both already documented in place as reachable-and-useful. I left them unmarked too, so the file reads 4 rather than a tidier 1.

## Verification

`pnpm test:gate` **GREEN** (158 + 10 + 487 + 71) · converted suites **41/41**, invariance file **9/9** · `pnpm lint` clean · `tsc -p tsconfig.app.json` clean · census `--strict` exits 0.

`TaskCard.test.tsx` / `TaskCard.badge-wrap.test.tsx` have **3 failures — pre-existing**, verified identical (same 3 names) on a stashed clean `origin/main` tree. Not touched by this change.


gsxdsm:
**The finding is real, and I verified the whole chain rather than just the flagged line. The fix already exists — in #2709 — and the two must land together.**

## Verified against `origin/main`

```
store.ts:1386   if (task.column !== "in-review") {
                  throw new Error(`... task is in '${task.column}', must be in 'in-review'`);
```

So yes: with this PR alone, the menu offers `bypass-review` on a renamed review lane and `POST /tasks/:id/bypass-review` refuses it. The route delegates all eligibility to the store, so nothing in between rescues it.

## The fix is #2709 (open, green, mergeable)

I converted that exact guard to the resolved review lane there, and fixed its rejection message in the same change — it was directing operators to `in-review` on boards that have no such column. Confirmed present on `origin/fleet/store`.

## My judgement: this PR is not a regression, it is incomplete

Worth stating precisely, because "UI says yes, backend says no" reads worse than it is:

| state | renamed review lane, card stranded on a failed pre-merge step |
|---|---|
| main today | affordance **hidden** — no escape hatch, no error, no path |
| this PR alone | affordance shown, request 400s with a message naming the wrong column |
| this PR + #2709 | affordance shown, bypass succeeds |

The middle row is bad, but it is not worse than the first: an operator who can see the action and get an error learns the state exists, where before they saw a permanently unmergeable card and nothing at all. Both are broken; only one is discoverable.

So I would **not** hold this PR or revert the conversion. Sequence #2709 first or merge them together.

## What I did not add

A cross-layer test pinning "the columns the menu offers bypass for == the columns the store accepts". It is the durable guard for this class — the half-conversion crossing a package boundary, which no single-file review catches — but written today it would **fail** against main's literal store guard, so it belongs on #2709 or a follow-up once both sides agree.

The single-layer coverage here is already good in both directions (the renamed lane surfaces the action; a card that is not genuinely stranded still does not), and I confirmed those 9 tests pass on this branch.

gsxdsm:
**Correct, and the most useful finding on this PR — fixed by converting the other half.**

`store.ts:1386` did require the literal:

```ts
if (task.column !== "in-review") {
  throw new Error(`Cannot bypass review lane for ${id}: task is in '${task.column}', must be in 'in-review'`);
}
```

So the UI change alone would have been **strictly worse than shipping neither half**: on a renamed review lane the operator sees a Bypass button, clicks it, and gets a conflict — the card stays stranded and now looks broken rather than merely unhelpable. That is a worse outcome than the affordance being absent, which is what I had before.

The store guard now resolves the review **role**, matching its UI half. FN-7720's contract is "the card is in the review lane", which is a role, not an id. The error message quotes the **resolved** lane too, so a renamed board reports its own vocabulary instead of telling an operator their card must sit in a column their workflow does not declare.

This is the third "both halves or neither" in this program — the archived gate (#2724), the tracking pair (#2715), and now a UI affordance split from its store guard. The pattern is consistent enough to be worth stating as a rule: **when a conversion makes an affordance appear, check what the affordance calls before shipping it.**

### Coverage — stated honestly rather than claimed

The UI half is revert-proofed (restoring the id makes the bypass vanish for `validating`). **The store half's renamed-lane behaviour is not covered by a new test.** `store-bypass-review.test.ts` runs on the shared PG harness, so a renamed-lane case needs a custom workflow registered in that harness — more machinery than this fix, and I would rather say so than imply coverage I did not add. The existing 9 cases pass and pin the default-board behaviour, including the message text.

### Verification

Gate **GREEN** (158 + 10 + 487 + 71) · `store-bypass-review` **9/9** · core `tsc` clean · `pnpm lint` clean · census `--strict` exits 0 (re-recorded: store.ts drops the converted site).

gsxdsm:
**Real, and fixed — but in #2709, not here, because the guard it points at is `store.ts` and that is where the conversion lives.**

`.review` resolves the single `mergeOrchestration` column, so a board hosting review on a `humanReview`- or `mergeBlocker`-only lane failed the store's check. This PR's menu asks by **role**, so it correctly offered "Bypass failed review" — and the store refused it. The operator's only escape from a stranded failed pre-merge step returned a conflict.

Both halves were individually reasonable. The menu is right to ask by role; the store was right to validate. It took reading them together to see the disagreement, which is exactly the class of defect a single-file review cannot catch.

**Fixed on #2709** by routing the store guard through `resolveReviewColumns` (core, #2730), with a test for a `humanReview`-only lane. Mutation: reverting to the singular `.review` fails it while the existing ten cases stay green — they all run the default board, where the two answers coincide.

**On which set is correct here**, since #2750 makes that a real choice: the **broad** set. This guard refuses or permits an operator action and **moves nothing**, so admitting every lane where review happens cannot send a card somewhere the engine disagrees with. A caller that admits and then *moves* wants the narrow single lane instead.

**Nothing to change on this PR.** Its conversion was correct; the mismatch was on the other side. Once #2709 lands, the menu and the store agree.

For sequencing: this is the same pairing I flagged earlier on this thread — #2718 alone shows an affordance that errors, which is still better than #2718 reverted (no affordance at all, no path, no error). Land #2709 first or together.

gsxdsm:
**Rebased and unblocked, and the store half of this PR is now dropped — main did it better.**

The branch was held in another worker's worktree for several rounds; it has since moved off, so this is rebased onto today's main and green.

### What changed in the resolution

`store.ts`'s bypass guard **is converted on main already**, and its version is better than the one I wrote here:

```ts
const reviewColumns = reviewIr === undefined ? ["in-review"] : resolveReviewColumns(reviewIr);
if (!reviewColumns.includes(task.column)) { … must be in ${named} … }
```

That is **membership over every review lane**, where mine was `lifecycle?.review ?? "in-review"` — first-per-role, which is exactly the arity gap I flagged on #2739 and declined to fix at resolver level. On a workflow that splits merge-blocking from human sign-off, my version would have refused the bypass on the second lane. Main's also names all admitted lanes in the error rather than one.

So I took main's `store.ts` wholesale and dropped my change. **This PR is now the UI half only** — `TaskContextMenu.tsx` **9 → 4** — which main does *not* have (it is still at 9 there, with the bypass guard unconverted).

That completes the pair I argued for when this thread was opened, just across two PRs instead of one. The argument stands and is worth restating because it was the reason to convert the store at all: shipping the UI half alone surfaces a Bypass button that returns a conflict, which is **worse than the affordance being absent**. With main's guard in place, the button now works on a renamed review lane.

### Verification

Gate **GREEN** (158 + 10 + 487 + 71) · `column-role-id-invariance` + `TaskContextMenu` **27/27** · `store-bypass-review` **11/11** (main's version, unmodified by me) · dashboard `tsc` clean · `pnpm lint` clean · census `--strict` exits 0 with `TaskContextMenu.tsx` at 4.

Diff is now three files: the component, its invariance test, and the census baseline line.

gsxdsm:
**Closing this — main has since landed the same conversion, and the only part of this branch that is not already there is a behaviour change the program deliberately deferred.**

I rebased onto `origin/main` and worked the conflicts hunk by hunk. What I found:

- **The conversion itself is already on main**, including the one I thought was this PR's most valuable line: the "Bypass failed review" guard now tests the review **role**, so the escape hatch from a failed pre-merge step is visible on a renamed board. Main's version and mine agree.
- **Main kept `isReviewColumn` / `isDoneOrReview` ORing the legacy id unconditionally, and marked it `DELIBERATE-LITERAL` with a note saying so**: a column explicitly resolved as `{ mergeBlocker: false, humanReview: false }` but *named* `in-review` is still classified as review. Main's marker calls this out as the same shape #2664 fixed in `isPreExecutionHoldColumn`, and defers it — *"it is a BEHAVIOR CHANGE and out of scope for a conversion batch."*

This branch makes that change. That is the whole remaining delta, and it contradicts a standing scope rule, so I am not going to land it as a rebase resolution — the right shape is a fresh, deliberately-scoped PR that someone signs off on.

**Do not lose these three cases.** After resolving to main's side, exactly three tests in `column-role-id-invariance.test.tsx` fail, and each is a precise statement of the deferred defect:

- `traits decide, not the lifecycle-sounding name: an \`in-review\`-NAMED mid-flight card matches its renamed twin`
- `traits decide for the terminal role too: a \`done\`-NAMED mid-flight card is still mutable live work`
- `a mid-flight card named \`in-review\` is not offered MERGE — the id alone used to open the merge lane`

The third is the user-visible one: **a mid-flight card sitting in a column merely named `in-review`, whose traits carry no review role, is currently offered Merge from the context menu.** They are red against main today, which is the correct signal — they are the ready-made proof for whoever picks up the deferred change.

Also worth recording, from main's own note and confirmed here: writing the exemption marker's name *in prose* inside a comment is enough to exempt the guard, because the census classifier substring-searches a comparison's leading comments. Documenting why a site was **not** exempted can silently shrink the count.

Rebased branch verified clean (census `--strict` exits 0) apart from those three intentional reds; nothing else on it is unique. Closing rather than merging.

PR #2721, opened 2026-07-30
docs(core): pin the LifecycleColumns arity contract — two shipped bugs came from it

gsxdsm:
Found by **auditing for the pattern** after hitting it twice, rather than waiting for a third instance.

## The contract, previously undocumented

`LifecycleColumns` names **one** column per role even when a workflow declares several. Nothing validates that a trait appears at most once — `columnsWithFlag` returns an array and `resolveLifecycleColumns` takes its head. So a workflow may legitimately declare two intake lanes, or split `mergeBlocker` and `humanReview` across a merge lane and a separate sign-off lane, and the struct names only one of each.

That makes the fields safe for one question and unsafe for another:

| | question | correct? |
|---|---|---|
| **safe** | *"where should this card **go**?"* | a move target must be exactly one column |
| **unsafe** | *"is this card **already** there?"* | membership — use `columnsWithFlag(...).includes(...)` |

## Two shipped bugs, both the unsafe use

Both in #2713, both reading like ordinary conversions:

- a task in a **second terminal column** was rejected with a 409
- a task in a **human-review lane split from the merge lane** was classified as outside review entirely, suppressing comment re-engagement

I fixed the first, did not generalise, and hit the second one review round later. That is the actual failure mode this PR addresses — not the individual bugs, which are already fixed.

## Two call sites left for their owners

Named in the doc comment rather than changed here, since they belong to other clusters. Both are correct **only while their workflow declares one column per trait**:

- `packages/engine/src/self-healing.ts` — `columns.intake` / `columns.hold`
- `packages/core/src/builtin-workflows.ts` — `lifecycle.intake`

## The test asserts the contrast, not just the arity

A reader who sees only *"it returns one id"* learns the wrong lesson. The test builds a workflow with **two intake and two complete columns** and shows that exactly one qualifying column **fails an equality check** against the struct while the **membership form gets it right** — the shipped-bug shape, reproduced in miniature.

## Verification

`workflow-lifecycle-traits.test.ts` **20 → 22**. `pnpm test:gate` green (10 / 158 / 487 / 71). `pnpm check:lifecycle-columns` exits 0. `tsc -p packages/core/tsconfig.json` clean. `pnpm lint` clean.

Docs + test only — no behaviour change, no census movement, no changeset.

### DIFF
--- a/packages/core/src/__tests__/workflow-lifecycle-traits.test.ts
+++ b/packages/core/src/__tests__/workflow-lifecycle-traits.test.ts
@@ -11,6 +11,7 @@ import { BUILTIN_CODING_WORKFLOW_IR } from "../builtin-coding-workflow-ir.js";
 import { columnsWithFlag, columnHasFlag, resolveReboundTarget, resolveCompleteColumn, resolveMergeOrchestrationColumn, resolveLifecycleColumns, resolveTaskLifecycleColumns } from "../workflow-lifecycle-traits.js";
 import { BUILTIN_CODING_IDEAS_WORKFLOW_IR } from "../builtin-coding-ideas-workflow-ir.js";
 import type { WorkflowIr } from "../workflow-ir-types.js";
+import { getTraitRegistry } from "../trait-registry.js";
 
 describe("columnsWithFlag — builtin:coding trait→columnIds (R8)", () => {
   const ir = BUILTIN_CODING_WORKFLOW_IR;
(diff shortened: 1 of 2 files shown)


gsxdsm:
You are right, and the premise was wrong in a way that mattered — it made me flag call sites that are actually safe.

`intake` **is** validated: `TraitRegistry.validateColumnTraits` raises `multiple-intake-columns` as an error. So my two-intake fixture was a workflow the product **rejects**, and I reached for `as never` to get it past typing — a test that would have stayed green while documenting a shape that cannot exist. The `as never` was the tell, and I wrote it without asking why the types objected.

The real boundary, now checked rather than assumed:

| | traits |
|---|---|
| **validated unique** | `intake` |
| **may repeat** | `hold`, `countsTowardWip`, `mergeBlocker`, `humanReview`, `complete`, `archived` |

That **narrows** the risk rather than widening it. Two of the three call sites I flagged are fine:

- `self-healing.ts` → `columns.intake` — **safe**
- `self-healing.ts` → `columns.hold` — **at risk**, `hold` has no uniqueness rule
- `builtin-workflows.ts` → `lifecycle.intake` — **safe**

The shipped bugs in #2713 are unaffected: those were `complete`/`archived` and `mergeBlocker`/`humanReview`, all genuinely repeatable.

The test now uses two **complete** columns, and its first case asserts the fixture passes `validateColumnTraits` with no errors. Proving the premise is what turns the arity gap from a hypothetical into a demonstrated hazard — and it is precisely the check whose absence let the first version document something impossible.

`workflow-lifecycle-traits.test.ts` 22 → 23. `pnpm test:gate` green (10 / 158 / 487 / 71), `pnpm check:lifecycle-columns` exits 0, `tsc -p packages/core/tsconfig.json` clean, `pnpm lint` clean.

PR #2722, opened 2026-07-30
fleet: notification-service.ts 9 → 5 — a renamed board silently stopped sending review and merge notifications

gsxdsm:
Claiming the **notification-service** cluster. Four convert; five are flagged and left **counted**, one of them after I converted it, measured a dropped notification, and backed it out.

## Census

| | before | after |
|---|---:|---:|
| `notification-service.ts` | **9** | **5** |

Baseline shrinks by exactly the 4 converted.

## The defect

Two guards compared `data.to` against the literals `in-review` and `done`. On a workflow that names its lanes anything else both comparisons are just false — so the two notifications an operator relies on most, **task reached review** and **task reached terminal merge**, were **silently not sent**. Nothing throws, nothing logs it; the operator simply stops being told.

Notification **event kinds** (`"in-review"`, `"merged"`) are deliberately untouched — those are channel names in the operator's ntfy config, not column ids. Renaming a board must not rename the event or every existing per-event filter stops matching.

**No service-lifetime IR cache**, though `resolveTaskLifecycleColumns` invites one: this service has no workflow-changed event to invalidate on, so a cached IR would outlive an operator's workflow edit and keep notifying against the old vocabulary. Each resolution instead sits *after* whatever cheap gate the site already had.

## A conversion I made, measured, and reverted

`hasProgressed` (4 of the 5 remaining entries) enumerates "every lane except review". Converting it typechecked and passed the notification suites — then failed `task-wedge-notification.test.ts > sends one actionable push and mailbox message per active terminal episode`: green on main, `expected 2 calls, got 1`.

Not a test artifact. That branch is the one that **resolves a wedge episode**, and it runs fire-and-forget from a synchronous `(task) => void` listener. Adding an await before the resolve lets a re-wedge arriving close behind still see the previous episode `active`; its claim returns `claimed: false` and **the second operator notification is dropped**. The awaits I added elsewhere are downstream of an existing await or inside a timer callback — this one sits on the only path that closes an episode, so it changes delivery, not just timing.

Fixing it properly means serialising wedge handling per task. That's a delivery-semantics change to operator notifications, so it's left for whoever owns the episode contract. `isManualMergeHold` is flagged for a related reason: it's sync and its only caller chain reaches the sync `handleTaskUpdated` listener.

## Revert proof

| reverted | result |
|---|---|
| review guard back to the id | RENAMED review case fails, **0 calls** |
| complete guard back to the id | RENAMED complete case fails, **0 calls** |

Both **DEFAULT** cases pass either way — which is exactly why both vocabularies run. A guard keyed on the literal passes on the default board for the wrong reason.

## Three latent faults found on the way

**1. 26 tests were asserting nothing.** `notification/__tests__/notification-service.test.ts` mocked `schedulerLog` without `debug`. Production had moved its suppression traces to `.debug`, so `service.start()` threw `schedulerLog.debug is not a function` and all 26 cases died in setup. They were red on main, which is the only reason this was visible at all. Fixed in its own commit: **26 failed | 65 passed → 91 passed**, verified against unmodified production code.

**2. 25 sites waited for a fire-and-forget handler with a single `await Promise.resolve()`** — not an assertion about behaviour but a count of the awaits that happen to be on the path. That is why my first run showed *29* failures: adding one await inside the handler turned green tests red with nothing broken. Replaced with a `flushAsyncHandlers()` drain (microtasks, not `setImmediate` — both suites use fake timers, which replace the macrotask queue). One site had already grown a second back-to-back `await Promise.resolve()`: the same discovery, papered over.

**3. `src/__tests__/**` is excluded from the engine typecheck**, so the shared fixtures there are never checked. My new test lives in `src/notification/__tests__/` — outside the exclusion — and importing the vocabulary fixture surfaced a real error: `MERGED_VOCAB` was missing `intake`. `lifecycleIr` derives `merged` from `v.intake === v.hold`, so MERGED_VOCAB *alone* built a separate intake column with **`id: undefined`** — the malformed-IR, "assertions pass for the wrong reason" failure that file's own header warns about. One-line fix, behaviour-neutral (all four callers already force the flag). **Widening the typecheck scope is not in this PR** — it would surface an unknown number of further errors and belongs to whoever owns the engine build.

## One finding I did not act on

`resolveLifecycleColumns` keys its `review` role on **`mergeOrchestration`**, not `human-review`. So a custom workflow whose review lane carries only `human-review` still gets no review notification. Meanwhile the dashboard's `isReviewColumnRole` treats `mergeBlocker` **or** `humanReview` as review — **the app and the core resolver disagree about what "review" means.** Reconciling them changes a resolver used well beyond notifications, so it is recorded at the site and here rather than changed under a conversion. Not a live outage: the default coding workflow's review column does carry merge orchestration.

## Verification

`pnpm test:gate` **GREEN** (158 + 10 + 487 + 71) · notification suites **129/129** · new differential **5/5** · `pnpm lint` clean · engine `tsc` clean (baseline was clean too — the fixture error was introduced into scope by this PR and fixed in it) · census `--strict` exits 0.

Not run: the four `*.pg.test.ts` live-engine E2Es that use the vocabulary fixture — they need embedded Postgres. The fixture change is provably inert for them (every call site already passes `mergedIntakeAndHold: true`, so `v.intake` is never read).


gsxdsm:
**Finding 1 (human-review lanes) — real, fixed, mutation-verified.**

`resolveLifecycleColumns().review` reads the `mergeOrchestration` flag **only**, so a column carrying `human-review` and not the merge trait resolves to nothing. The guard then fell back to the literal `in-review`, and on a renamed board the operator's "ready for review" notification never fired.

Silent in the worst way: no error, no log, the card just arrives unannounced. The operator cannot even report it, because from their side *nothing happened*.

Fixed with the same union the dashboard routes already use — `mergeBlocker` ∪ `humanReview`, legacy id when neither resolves — so this is that established idiom applied here, not a new notion of "review". **Membership, not equality**, per the arity rule #2713 wrote down after fixing it twice: a set answers "is this card *already* there", and a workflow may declare more than one review lane.

**Mutation:** restoring `data.to === (movedLifecycle?.review ?? "in-review")` fails the new case while the default-board case still passes — exactly how the gap survived the conversion.

---

**Finding 2 (resolver failure reverting to legacy ids) — judged intended, and I am not changing it here.**

That fallback is the documented contract at *every* converted site in this program. When the IR cannot be read there is no better answer available, and notifying on the legacy ids beats notifying on nothing. Making it fail-closed instead is a behaviour decision spanning every converted guard in the codebase — it deserves its own change with its own reasoning, not a quiet flip inside a notification PR.

If the intent is that a resolution *failure* should be loud rather than silent, the right shape is a log or a run-audit signal at the resolver, not per-consumer divergence.

---

**Pre-existing and untouched:** `notifier.test.ts` has **13 failures on this branch before my change** — verified by stashing my edits and re-running. Different file, unrelated to this conversion.

Verified: 52 notification-service tests green, engine `tsc` clean, lint clean, gate green (487 + 158 + 10 + 71), `census --strict` exit 0.

gsxdsm:
**Fixed — and it was my own miss, in the file I had already 'converted'.**

Same defect as the moved-to-review guard: `.review` reads `mergeOrchestration` only, so a lane carrying `human-review` alone resolved to nothing and `isTerminal` fell back to the literal. Under `failureNotificationMode: "terminal-only"` a task failing in a renamed review lane was **never terminal**, so its failure notification was suppressed indefinitely.

I converted one guard in this file and did not grep it for the remaining `.review` reads — the exact Surface Enumeration failure I had been flagging in other people's PRs the same day. Both reads now go through one set; `grep '?\.review'` in this file returns **zero**.

**Two vacuous tests before this one worked**, both worth recording because each looked green:

1. asserting `sendNotification` fired at all — it fires either way once the grace window elapses, so the suppression branch was never reached;
2. using `failureNotificationMode: "sticky-only"` — `isTerminal` is only consulted under `terminal-only`, so **the flag under test was dead code in the scenario I built for it**.

The passing version reproduces the actual suppression path. **Mutation: reverting the guard fails 1 of 53.**

Verified: 53 notification tests green, engine `tsc` clean, lint clean, gate green (487 + 158 + 10 + 71).

---

**Rebase blocked on a structural conflict I did not resolve, deliberately.**

`main` now carries `packages/engine/src/notification/__tests__/notification-service.test.ts` while this branch's tests live at `packages/engine/src/__tests__/notification-service.test.ts`, and this branch's own commit *"restore the notification suites — 26 tests were dying in setup"* also touches the new path.

Resolving that means deciding which location is canonical and merging two test sets — and getting it wrong silently drops 26 restored tests, which no gate would catch. That is the branch author's call, not something to land under an automated rebase. My fix is pushed and safe on the branch; the conflict is one file and structural, not baseline churn.

gsxdsm:
Real, and it is the **mirror of the bug this method was added to fix** — my own miss, one round later.

I created `resolveReviewColumnsForTask` here because a `humanReview`-only lane resolved to nothing. I built it from `mergeBlocker` + `humanReview` and left `mergeOrchestration` **out**, so a renamed review lane carrying only the merge trait was excluded and moving a task there skipped the operator's review notification. Same silent failure, pointed the other way: no error, no log, the card just arrives unannounced.

Now routed through **`resolveReviewColumns`** (core, #2730), which covers all three flags. Both directions go through one definition, so the pair cannot drift apart again — which is precisely why that helper exists.

**On which set is correct here**, since #2750 makes it a real choice: the **broad** one. This path emits a notification and moves nothing, so over-admission costs an extra notification while under-admission costs the operator their signal. A caller that admits and then *moves* wants the narrow single lane.

**Mutation: restoring the two-flag union fails the new merge-only case while the `humanReview`-only case stays green.** The two halves are only distinguishable together — which is exactly how I shipped one and missed the other.

Verified: 54 notification tests green, engine `tsc` clean, lint clean (0 errors), gate green (487 + 158 + 10 + 71), `census --strict` exit 0.

gsxdsm:

PR #2724, opened 2026-07-30
fleet: async-comments-attachments.ts NOT converted — the archived gate lives in three encodings (52 sites), pinned by a guard that fails on each

gsxdsm:
Claimed the **async-comments-attachments** cluster (9) and did not convert it. This PR is the evidence for why, as a guard rather than a note — **no production change, census unmoved.**

## What the cluster actually is

Every other file in the backlog converts on its own: resolve the task's lifecycle columns, compare against the role. `archived` doesn't, and not by a matter of degree.

**Measured in `packages/core` — one rule, three encodings, 52 sites:**

| encoding | sites | files | what it decides |
|---|---:|---:|---|
| TypeScript `=== "archived"` | **37** | 23 | what code does with a row it already has |
| Drizzle `ne(tasks.column, 'archived')` | **7** | 6 | which rows a query returns |
| raw `` sql`…column != 'archived'` `` | **8** | 5 | same — and invisible to both scans above |

Convert only the TypeScript half and a board whose archived lane is renamed **splits**: `getLiveTaskColumn` correctly reports the task archived (it resolved the role) while `readLiveTaskRows` still hands it back as live. A document write is rejected by its gate while its parent is listed as live — a state neither gate alone can produce today.

And nothing would catch it. **Every builtin workflow spells that column `archived`**, so all three encodings agree by accident on every board we ship.

## Why the SQL sides aren't just converted too

`ne(tasks.column, ...)` needs the resolved id as a **query-build value**, so the IR must be resolved *before* the query — including inside the `for update` document/artifact transactions, which today receive a `db`/`tx` handle and **no store and no workflow reader**. One raw site is a hand-written `SELECT` string, so its comparison isn't even a Drizzle expression that could take a bound value without rewriting the query.

The two real options are: thread a resolver into the persistence layer, or **declare `archived` a non-renameable system column** and mark all 52 sites deliberate. Both are decisions with blast radius. Neither is a fleet conversion, so I didn't pick one.

## I was wrong about the shape, and that's the strongest part

I wrote the third case as an assertion that **no** raw `sql` template compares a column to `'archived'` — I assumed two encodings. **It failed on the first run with five files.**

Nothing in the repo was counting them: they aren't comparisons (invisible to the column census) and aren't `eq`/`ne` calls (invisible to the Drizzle scan). A partial conversion doesn't have to miss one encoding — it can miss two. That case is now an inventory, with a note on why asserting absence was the wrong invariant: an absence assertion has to be deleted by whoever adds the next raw template, and deleting a red guard is how a class of sites stops being tracked.

## Injection proof — all three run

| injected | result |
|---|---|
| convert one TS comparison in `async-comments-attachments.ts` | fails: **"TypeScript encoding changed"** |
| remove one Drizzle predicate in `async-lifecycle.ts` | fails: **"Drizzle encoding changed"** |
| remove one raw template in `reads.ts` | fails: **"Raw-sql encoding changed"** |

Each failure carries the split-brain explanation and the two real options, so the next worker hits the reason rather than a bare count mismatch.

## Two scan bugs I fixed in my own guard

- **It audited documentation.** The raw-template scan reported three sites in `async-archive-lineage.ts` that were prose in a JSDoc block. Now comment-stripped through the census's shared `stripComments` rather than a second implementation.
- **It missed half the cluster.** Requiring a receiver (`x.column === "archived"`) misses the list paths that hold the value in a bare local (`if (column === null || column === "archived") return []`) — **four of the eight sites in the target file**, measured. The matcher now accepts a bare identifier named `column`.

Counts here are per file, not per line: line numbers churn on unrelated edits, and a ratchet that cries wolf gets deleted.

## Verification

`pnpm test:gate` **GREEN** (158 + 10 + 487 + 71) · new guard **2/2** · `pnpm lint` clean · core `tsc` clean · census `--strict` exits 0 with the **baseline untouched** — this PR converts nothing and claims nothing.

### DIFF
--- a/packages/core/src/__tests__/archived-column-gate-parity.test.ts
+++ b/packages/core/src/__tests__/archived-column-gate-parity.test.ts
@@ -6,8 +6,12 @@ Every other file in the column-literal backlog can be converted on its own: reso
 lifecycle columns, compare against the role. `archived` is different, and the difference is not a
 matter of degree.
 
-MEASURED on this tree — THREE encodings, 52 sites, all in packages/core:
-  - 37 TypeScript comparisons against the literal `"archived"` across 23 files.
+MEASURED on this tree — THREE encodings, 50 sites, all in packages/core:
+  - 35 TypeScript comparisons against the literal `"archived"` across 23 files.
+
+  (Was 37 when this guard landed. Two dropped as conversions merged — `agent-store.ts`'s claim guard in
+  #2746 and one in `update-task-deps.ts` — and the guard FAILED until this inventory was updated to match,
+  which is the ratchet working: it notices the TypeScript half moving in either direction, not only up.)
   - 7 Drizzle predicates pushing the same rule into SQL as `ne(tasks.column, 'archived')`, in 6 files.
   - 8 RAW `sql` template comparisons (`sql`${tasks.column} != 'archived'`` and one hand-written
     `SELECT ... "column" = 'archived'`), in 5 files.
@@ -51,7 +55,7 @@ import { describe, expect, it } from "vitest";
  * ratchet that cries wolf gets deleted.
  */
 const AUDITED_TS_SITES: Readonly<Record<string, number>> = {
-  "packages/core/src/agent-store.ts": 2,
+  "packages/core/src/agent-store.ts": 1,
   "packages/core/src/assigned-task-ranking.ts": 1,
   "packages/core/src/async-mission-store-queries.ts": 2,
   "packages/core/src/async-mission-store.ts": 2,
@@ -73,7 +77,7 @@ const AUDITED_TS_SITES: Readonly<Record<string, number>> = {
   "packages/core/src/task-store/symbol-locks.ts": 1,
   "packages/core/src/task-store/task-id-integrity.ts": 1,
   "packages/core/src/task-store/task-store-helpers.ts": 1,
-  "packages/core/src/task-store/update-task-deps.ts": 2,
+  "packages/core/src/task-store/update-task-deps.ts": 1,
 };
 
 /**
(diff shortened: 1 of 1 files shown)


gsxdsm:
**P1 (parity guard is not gated) — accurate observation, and I am deliberately not acting on it.**

The finding is factually right: `archived-column-gate-parity.test.ts` is not in the `engine-core` allow-list, so a partial conversion of the archived gate would be caught by the non-blocking full suite rather than by the merge gate.

AGENTS is explicit that this is the intended default:

> Gate admission requires evidence of value; tests never graduate into the gate by default.

and the gate is described as deliberately thin — "CI blocks PRs on exactly Lint, Typecheck, Build, and Gate". A guard written today, which has never caught a real regression, has no evidence of value yet. Adding it because it *feels* important is how the thin gate stops being thin, and the file's own history says the gate rotted that way before.

The guard's value does not depend on gate membership either: it fails loudly in the full suite, and its failure message names the encoding that changed plus the two real options. Someone doing a partial conversion sees it before merge in review and after merge in the non-blocking run.

If it ever fires on a real partial conversion, that is the evidence AGENTS asks for and it should be promoted then, with the incident cited.

**P2 (AST scan skips equivalent access forms) — correct, and worth a fix.**

The scan accepts a property access or a bare `column` identifier; it does **not** accept element access (`task["column"] === "archived"`). Optional chaining is already covered — `a?.b` is still a `PropertyAccessExpression` in the TS AST — but bracket access is a genuine hole, and a guard that can be walked around by changing syntax is the exact failure mode I fixed in the `maxWorktrees` audit (regex → AST) and should not have re-introduced in a narrower form.

I will add `ElementAccessExpression` with a string-literal argument named `column` to the matcher. Since that could surface currently-uncounted sites, the inventory may move; if it does, the diff will show exactly which files gained entries, which is the honest outcome rather than a silently narrower scan.

gsxdsm:
P2 fixed, and the fix produced a more interesting result than expected.

**Element access is now matched.** `row["column"] === "archived"` walked straight past the scan before. Optional chaining needed nothing — `a?.b` is still a `PropertyAccessExpression` in the TS AST — but bracket access was a genuine hole, and a guard defeated by changing syntax is exactly the failure I fixed in the `maxWorktrees` audit by moving regex → AST, then re-introduced here in a narrower form.

**The widening found no new sites — but the guard failed anyway, for a better reason.** The inventory had drifted *down*:

```
- "packages/core/src/agent-store.ts": 2
+ "packages/core/src/agent-store.ts": 1
- "packages/core/src/task-store/update-task-deps.ts": 2
+ "packages/core/src/task-store/update-task-deps.ts": 1
```

Both are conversions that **merged** since this guard was written — `agent-store.ts`'s claim guard in #2746 (mine) and one in `update-task-deps.ts`. The guard failed until the inventory matched, which is the ratchet working in the direction people forget to test: it notices the TypeScript half moving **down** as well as up. A count-only-rises guard would have stayed green while the three encodings silently diverged in the other direction, which is the same split brain.

Header count corrected from 37 to 35 TypeScript comparisons (50 sites total across the three encodings), with the reason recorded inline so the next reader knows the number moved because work landed, not because the scan changed.

The P1 (gate admission) reply stands: AGENTS says tests never graduate into the gate by default and this guard has not yet caught a real regression. When it does, that is the evidence for promotion.

### Verification

`archived-column-gate-parity` **2/2** · core `tsc` clean · `pnpm lint` clean · census `--strict` exits 0 (this PR converts nothing).

PR #2729, opened 2026-07-30
fleet: scheduler.ts 26 → 12 — five quiet wrong answers on a renamed board (finished deps blocked forever, PRs unwatched, missions stalled)

gsxdsm:
## Census

| | before | after |
|---|---|---|
| `packages/engine/src/scheduler.ts` | 26 | **22** |
| repo backlog | 657 | **653** |

Baseline re-recorded in this PR; `--strict` exits 0. Four literals removed, and I want to be exact about why it is four and not seven: the converted predicates keep their legacy literals as the documented **no-metadata fallback**, and the census counts per literal, not per code-quality improvement. Deleting those fallbacks would change behaviour in degraded mode (unresolvable workflow → every dependency reads as unsatisfied → dependents blocked forever), which is the expensive direction to be wrong in.

## What was actually broken

**1. Dependency satisfaction was keyed on three column ids.**

```ts
return !!dep && (dep.column === "done" || dep.column === "in-review" || dep.column === "archived");
```

On a board whose complete column is `shipped`, a **finished** dependency matched none of the three. `getUnmetSchedulingDependencies` reported it unmet, and the dependent was parked `blockedBy` — *permanently*, because the dependency can never move anywhere that satisfies the literal. Work stops and nothing rescues it.

Satisfaction is now resolved on the **dependency's own board**, since a dependency edge may cross workflows — the dependent can sit on the default board while the dependency lives on a renamed one. Resolution is passed in by the caller (`resolveDependencySatisfactionColumns`) with a caller-owned IR cache, so a sweep reads one IR per distinct workflow rather than one per dependency edge.

**2. The review half of the file-scope lease was never converted.**

The wip half of this same sweep was fixed on 2026-07-30-16:30 (`scheduler-renamed-wip-file-scope-lease.test.ts`). The review half still read `column === "in-review"`, so on a renamed board no review card entered `activeScopes`, a merging card's worktree files read as **free**, and an overlapping candidate dispatched on top of them. One registry, two halves, disagreeing. It now uses `isReviewColumnRole` over the *same* resolved flags map the wip half uses, so the two cannot drift again.

## Finding I am reporting rather than fixing

**The two satisfaction rules in this one function genuinely disagree, and the live one is the broader.**

| rule | satisfied when |
|---|---|
| legacy (**live**) | complete ∪ archived ∪ **review lane** |
| marker (shadow) | complete ∪ archived |

#2720 settled "satisfied = complete or archived" for `update-task-deps.ts` — which matches the **marker** rule, not the live one. So the scheduler currently treats an in-review dependency as finished and `update-task-deps.ts` does not.

Reconciling them is a product decision, not a vocabulary one, so this PR preserves **both** rules exactly as shipped. Narrowing the live rule to match would strand every dependent of an in-review card, which is precisely the failure mode fix 1 exists to remove — I am not doing that as a side effect of a rename conversion.

## Revert proof

Each fix reverted **alone**, tests re-run:

| reverted | result |
|---|---|
| dependency satisfaction → literals | **2 failed** / 5 passed |
| review lane → `column === "in-review"` | **2 failed** / 5 passed |
| neither (shipped) | **7 passed** |

Each revert fails exactly its renamed case *and* its "both vocabularies reach the same outcome" invariant, while the default-vocabulary controls stay green — so the failures are attributable to a surviving column-id literal and not to a generally broken path. The suite is differential: one workflow **shape**, two vocabularies with identical traits, only the ids differ. No renamed id collides with a legacy literal, so a surviving `=== "done"` cannot pass by luck.

There is also a paired negative (`an UNFINISHED dependency still blocks, under both vocabularies`) so the fix cannot degrade into "always satisfied" — the direction it could overshoot.

## Verification

- `pnpm --filter @fusion/engine exec vitest run <19 scheduler/dependency/hold-release/overlap suites>` — **174 passed**, no regressions
- new suite — **7 passed**
- `pnpm test:gate` — **158 / 487 / 10 / 71 passed**
- `pnpm lint` clean · `npx tsc -p packages/engine/tsconfig.json --noEmit` clean · `check:lifecycle-columns --strict` exits 0

## The remaining 22, triaged

Not guessed at — grouped by what they actually ask:

- **6 fallback branches already converted** (L258, L265, L439, L440, L1712 and the review twin): trait-first with the literal as documented degraded-mode fallback. These reach 0 by *marking*, not converting.
- **10 `from`/`to` move-transition arms** (L899–L1042): a different question ("is this transition *into* a review lane?"), and per the scoping note in `docs/solutions/architecture-patterns/` they should not ride along with `task.column` conversions.
- **6 `task.column` reads** (L1061, L1135, L1524, L1546, L1554, L2607): convertible, but two sit in sync methods (`resolveBaseBranch`) needing the caller-resolves-and-passes shape, so they are a separate unit rather than a half-conversion here.

### DIFF
--- a/packages/engine/src/__tests__/scheduler-renamed-dependency-and-review-lanes.test.ts
+++ b/packages/engine/src/__tests__/scheduler-renamed-dependency-and-review-lanes.test.ts
@@ -1,0 +1,298 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { Scheduler } from "../scheduler.js";
+import type { Settings, Task, TaskStore, WorkflowIr } from "@fusion/core";
+
+/*
+FNXC:WorkflowLifecycleColumns 2026-07-30-19:05:
+Two scheduler decisions that were keyed on column-id literals, on a RENAMED board.
+
+DEFECT 1 — DEPENDENCY SATISFACTION. `isLegacyDependencySatisfied` answered "is this dependency
+finished?" with `column === "done" || "in-review" || "archived"`. On a board whose complete column is
+`shipped`, a FINISHED dependency matched none of the three, so `getUnmetSchedulingDependencies`
+reported it unmet and every dependent stayed queued with `blockedBy` set — permanently, because the
+dependency can never move anywhere that would satisfy the literal. This is the expensive direction to
+be wrong in: work stops and nothing rescues it.
+
+DEFECT 2 — THE REVIEW HALF OF THE FILE-SCOPE LEASE. The sweep that builds `activeScopes` had its wip
+half converted to traits (2026-07-30-16:30, scheduler-renamed-wip-file-scope-lease.test.ts) while its
+review half still read `column === "in-review"`. On a renamed board no review card entered the
+registry, so a merging card's worktree files read as FREE and an overlapping candidate dispatched on
+top of them. Half a fix leaves one registry with two disagreeing halves.
+
+WHY THESE ARE DIFFERENTIAL. Each scenario runs twice against the same workflow SHAPE under two
+vocabularies with identical traits; only the ids differ. Any behavioural difference between the runs
+is attributable to a surviving column-id literal and nothing else. The default-vocabulary run is the
+control: it passes before and after, so a change that breaks these paths generally cannot hide here.
+No renamed id collides with a legacy literal, so a surviving `=== "done"` cannot pass by luck.
+*/
+
+const WF = "custom:renamed-lanes";
+
+const PASSED_PLAN_REVIEW = {
+  workflowStepId: "plan-review",
+  workflowStepName: "Plan Review",
+  status: "passed" as const,
+  source: "node" as const,
+  phase: "pre-merge" as const,
+};
+
+interface Names {
+  hold: string;
+  wip: string;
+  review: string;
+  complete: string;
+}
+
+const DEFAULT_NAMES: Names = { hold: "todo", wip: "in-progress", review: "in-review", complete: "done" };
+const RENAMED_NAMES: Names = { hold: "drafting", wip: "building", review: "checking", complete: "shipped" };
+
+function makeTask(overrides: Partial<Task> = {}): Task {
+  return {
+    id: "FN-001",
+    title: "task",
+    description: "",
+    column: "todo",
+    dependencies: [],
+    steps: [],
+    currentStep: 0,
+    log: [],
+    workflowStepResults: [PASSED_PLAN_REVIEW],
+    createdAt: "2026-01-01T00:00:00.000Z",
+    updatedAt: "2026-01-01T00:00:00.000Z",
+    ...overrides,
+  } as Task;
+}
+
+/** One workflow shape; the traits are identical under both vocabularies. */
+function ir(names: Names): WorkflowIr {
+  return {
+    version: "v2",
+    id: WF,
+    nodes: [],
+    edges: [],
+    columns: [
+      { id: names.hold, label: "Hold", traits: [{ trait: "hold", config: { release: "capacity" } }] },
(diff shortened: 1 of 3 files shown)


gsxdsm:
## Second push — three more conversions

`scheduler.ts` **26 → 12**, repo backlog **657 → 643**. Baseline re-recorded; `--strict` exits 0.

| # | site | what was broken on a renamed board |
|---|---|---|
| 1 | dependency satisfaction | a **finished** dependency read as unmet → dependents parked `blockedBy` **permanently** |
| 2 | file-scope lease, review half | a merging card's worktree files read as **free** → overlapping candidate dispatched on top |
| 3 | PR-monitor startup hydration | restart hydrated **nothing** → every open PR silently unwatched |
| 4 | base-branch stacking | task started from HEAD instead of its predecessor's branch → silently rebuilt work, then conflicted |
| 5 | mission completion advance | roadmap showed the work finished while mission execution **stalled** |

Every one of these is a *quiet wrong answer*, not a crash — which is why they survived.

### Design note on #4

The resolved predicate is **required**, not optional-with-a-literal-default like the shared `shouldHoldActiveFileScopeLease`. `resolveBaseBranch` has exactly one caller, inside `runHoldReleaseSweepPass`, which has already resolved per-task trait flags for the whole board. An optional parameter would let a future caller reintroduce the bug by omission; required means the compiler asks the question. The optional shape stays where it is genuinely needed — on the exported predicate shared with self-healing/repair, where a default preserves those callers' meaning.

### Revert proof — all five, each reverted alone

| reverted | result |
|---|---|
| dependency satisfaction | **2 failed** / 11 passed |
| review-lane lease | **2 failed** / 11 passed |
| PR-monitor hydration | **1 failed** / 12 passed |
| base-branch stacking | **1 failed** / 12 passed |
| mission completion | **1 failed** / 12 passed |
| none (shipped) | **13 passed** |

**Two of these tests did not work the first time, and both failures were mine, not the code's.**

- The hydration assertion initially passed *vacuously*: `getCurrentRepo("/tmp/project")` returns null, so the handler bailed before reaching the guard and **both** vocabularies hydrated nothing — "they match" was true for the wrong reason. Stubbed the repo guard, and the case became real.
- The mission-completion conversion **survived its first revert** — i.e. it was shipping unproven. The control was failing too (`store.getFeatureByTaskId is not a function`), so the whole scenario was inert. Fixed the mock; the revert now fails. This is precisely why each fix is reverted individually rather than trusting a green suite.

### The six DELIBERATE-LITERAL marks

These are fallbacks that answer **only** when a workflow cannot be resolved at all. Deleting them does not "finish the conversion" — it turns *unresolvable* into `terminal.size === 0` (never satisfied → dependents blocked forever) or into "no role" (affordances vanish). Strictly worse than the legacy behaviour they would replace, so they are marked rather than removed.

### Verification

- 39 scheduler/mission/dependency/hold-release/overlap suites — **518 passed**, no regressions
- new suite — **13 passed**
- `pnpm test:gate` — **158 / 10 / 487 / 71** · `pnpm lint` clean · engine `tsc --noEmit` **0 errors**

### The remaining 12

- **10 `from`/`to` move-transition arms** — a different question ("is this transition *into* a review lane?"), deliberately not ridden along with `task.column` conversions.
- **2 sync-handler reads** (`task.column === "in-progress"` in `task:updated`, `!== "in-review"` in the PR-monitor start hook) — **flagged, not guessed**. Both sit in synchronous event listeners where the only in-file precedent is `resolveTaskWorkflowIrSync`, which returns the *default* workflow IR for every task in production and is typed non-optional so no caller can detect the substitution. Converting through it would produce a guard that reads the wrong board's lanes while *looking* converted, and the census would score it as progress. That is the "guard that cannot fire" pattern wearing better clothes, so these need the event-ordering question settled first.

PR #2731, opened 2026-07-30
fleet: the app-side remainder 18 → 0 — Archive/Revert and diff stats were silently absent on a renamed board

gsxdsm:
Three **genuinely free** clusters in one layer and one idiom — `Column.tsx` (7), `ListView.tsx` (6), `useTaskDiffStats.ts` (5). I built the claimed-file set from every open PR's diff before starting, having duplicated a claimed cluster last round.

## Census

| file | before | after |
|---|---:|---:|
| `Column.tsx` | 7 | **0** |
| `ListView.tsx` | 6 | **0** |
| `useTaskDiffStats.ts` | 5 | **0** |

**16 converted; 2 reclassified with a reason** — the two are accounted for separately below so the numbers stay honest.

## Three silent failures, not three style nits

- **`ListView` Archive and Revert** were gated on `task.column === "done"` / `=== "archived"`, so on a board with renamed terminal lanes **they did not render at all**. No error, no log — the operator simply cannot archive or revert from the list.
- **`useTaskDiffStats`** compared a bare `column: string` to `done`/`in-progress`/`in-review`, so on a renamed board it **fetched nothing** and the row showed no changes.
- **`ListView` progress display** had the same shape for the WIP lane.

## The `?? {}` is the whole subtlety

Every `Column.tsx` site was `workflowMode ? <trait> : column === "<id>"`. One adapter now feeds the shared helpers:

```ts
const columnRoleFlags = workflowMode ? (columnFlags ?? {}) : undefined;
```

`workflowMode` means **traits are the only authority**, so a workflow-mode column with no resolved flags must answer `false` — which `Boolean(columnFlags?.archived)` did. Passing `undefined` to a role helper instead selects its **legacy id fallback**, so a flagless workflow-mode column would start matching on its id. An empty object keeps the helper on its trait branch. Legacy mode passes `undefined` deliberately: there the id fallback *is* the answer, and routing it through the helpers is the point.

## Two things I deliberately did not do

**`isTodoLikeColumn` keeps its own trait arm.** Adopting `isPreImplementationColumnRole` would widen its fallback from `todo` alone to `{todo, triage}`, handing a legacy `triage` column a bulk replan affordance it does not have today — a behaviour change hiding inside a de-duplication. Only its *fallback* is routed through a helper.

**The `mode === "done"` pair is reclassified, not converted.** It is the hook's own `"done" | "active"` discriminant, assigned three lines from `shouldFetchDoneTask` — not a column id, with no trait to resolve. The census counts it because the receiver is compared to the string `done`, which is a classifier limit. Marked deliberate and **recorded in `deliberateByFile`**, so that file's `byFile` drop is 5 while its conversion count is 3.

One genuine simplification fell out: `workflowMode ? isReviewColumn : column === "in-review"`, where `isReviewColumn` is *itself* that same ternary. Both arms already agreed with it — collapsing is behaviour-identical.

## Revert proof

Restoring the id comparisons on the ListView row menu fails the new renamed-lane case with `Unable to find an accessible element with the role "menuitem" and name "Archive"`.

Driven through the **real `fetchBoardWorkflows` seam** with a renamed vocabulary — payload → `listColumns` → `columnFlagsById` → row menu — rather than by injecting flags, so the assertion covers the path the component actually uses. The DEFAULT-vocabulary path passes either way, which is exactly why the renamed case has to exist.

## Verification

`pnpm test:gate` **GREEN** (158 + 10 + 487 + 71) · **375 passed** across Column / ListView / useTaskDiffStats / role-invariance / columnRoles · dashboard `tsc -p tsconfig.app.json` clean · `pnpm lint` clean · census `--strict` exits 0.

`TaskCard.tsx` is touched only to pass the new optional `columnFlags` through; its own census count is unchanged at 3. The 2 `TaskCard` reds in that suite are the known pre-existing CSS-var geometry assertions.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved workflow lane handling when columns are renamed or assigned roles through workflow settings.
  * Archive and Revert actions now remain available for completed and archived tasks in renamed lanes.
  * Corrected task progress and diff-stat behavior across active, review, completed, and archived lanes.
  * Updated bulk actions, sorting controls, and auto-merge controls to respond consistently to workflow roles.


gsxdsm:
All three findings worked.

**Unresolved flags enabling legacy roles — real, and the fix was already in this file.**

`columnFlagsById.get(...)` returns `undefined` for a column the board does not declare, and every role helper reads `undefined` as *"no resolved metadata, use the legacy id"*. So on a workflow board a stranded card in a column literally named `done` was offered **Archive** and **Revert**, and one named `in-progress` rendered WIP progress — for lanes carrying none of those roles.

`?? {}` is the entire distinction: an **empty** flag set means "resolved, and this column carries no such role", which the helpers answer `false` for. Legacy-mode boards still pass `undefined` and keep the id comparison, so nothing changes there.

The idiom was **already present** at the archived/complete memos in this same file:

```ts
isArchivedColumnRole(workflowMode ? (columnFlagsById.get(column) ?? {}) : undefined, column)
```

The conversion just did not carry it to the three remaining sites — the row action menu and both `getTaskProgress` call sites. That is worth noting as its own lesson: when a file already contains the careful form, the sites that lack it are the ones to enumerate, and grepping for the careful form is faster than reasoning about each call.

Same direction as #2670 in the executor: **being unable to prove a card is finished must not be treated as proving it is.**

**Timestamps — the finding named `Column.tsx`; `ListView.tsx` and `useTaskDiffStats.ts` carried future-dated stamps too.** All three corrected. Fourth time today a line-anchored finding caught the instances the reviewer happened to open rather than the class.

**`useTaskDiffStats` (third finding)** — the `columnFlags` option is consumed inside the effect and already listed in its dependency array, so the poll re-evaluates when flags resolve. No change needed; verified rather than assumed.

Verified: 241 ListView tests green, dashboard `tsc` clean (`tsconfig.app.json`), lint clean, gate green (487 + 158 + 10 + 71).

gsxdsm:
Both findings were already addressed in the previous round; I re-verified them on the rebased branch rather than assuming, and confirmed the content survived the rebase.

**Timestamps:** zero future-dated FNXC stamps remain in all three touched files (`Column.tsx`, `ListView.tsx`, `useTaskDiffStats.ts`). The finding named only `Column.tsx`; the other two carried them as well.

**`useTaskDiffStats`:** the `columnFlags` option is consumed inside the effect and already listed in its dependency array, so the poll re-evaluates when flags resolve. No change needed — verified by reading the hook, not inferred from the finding's absence of a concrete failure.

**The substantive fix survived the rebase:** three `workflowMode ? (columnFlagsById.get(task.column) ?? {}) : undefined` sites present — the row action menu and both `getTaskProgress` calls. That is the change that stops a stranded card in a column literally named `done` from being offered Archive and Revert on a board where `done` carries no complete role.

Rebased onto current main. Force-push checked with `--not HEAD` first: both remote commits were reproduced in the rebase, one of them my own earlier push.

Verified: 241 ListView tests green, dashboard `tsc` clean (`tsconfig.app.json`), gate green (487 + 158 + 10 + 71), `census --strict` exit 0.

gsxdsm:
**I dismissed this finding twice, both times reporting it as "verified by reading the hook." It was not verified — I misread the file.**

`columnFlags` is read inside the effect (the three role predicates at 131–133) and was **absent** from the dependency array, which ends at `pollIntervalMs`. My earlier replies asserted it was listed. It was not.

**The consequence:** `columnFlags` arrives from a board-workflows fetch, so it is `undefined` on first paint. The effect captured that pre-resolution answer and never re-ran — on a renamed board, a card in a custom complete/wip/review lane **never started fetching diff stats at all**.

**Fixed by depending on the derived booleans, not the object.** `columnFlags` is a prop object whose identity a parent may change every render; listing it would restart the poll continuously. The two predicates are primitives, so they change exactly when the answer changes — which is the dependency the effect actually has.

---

**A flake found on the way, quarantined rather than appeased.**

`useTasks-hydration-freshness.test.ts` asserts exact epoch equality against a value derived from a second real-clock read, and failed **1 in 3 identical runs**:

```
AssertionError: expected 1785409894615 to be 1785409894614
```

Different hook from this change. I stashed my change and the file passed 11/11, restored it and the variance returned — so the diff is not the cause, the clock is. Per AGENTS.md that is a quarantine on sight: ledger entry **and** matching vitest exclude, in this same commit, with the root-cause fix named (fake timers or a monotonic source — **not** a widened assertion). `check-quarantine-ledger` exits 0.

---

**Why this one matters beyond the fix.** Two of my three replies on this thread were confidently wrong about a fact I could have checked in one command. The reviewer re-raised it twice; the third time I ran `grep -n 'columnFlags'` and the dependency array side by side and the answer was immediate. "Verified by reading" is not verification — I have now recorded that as the failure it is rather than quietly correcting course.

Verified: 30 `useTaskDiffStats` tests green, 1826 hook tests green, dashboard `tsc` clean (`tsconfig.app.json`), lint clean (0 errors), gate green (487 + 158 + 10 + 71), `census --strict` exit 0. Rebased onto current main; force-push checked with `--not HEAD` and all three prior commits confirmed reproduced.

gsxdsm:
Rebased onto `origin/main` (cced3d7dda). The conflicts were substantive, so recording how each was resolved rather than just "resolved".

**`Column.tsx` and `ListView.tsx` — I took main's side and dropped this branch's versions.** Main has since landed a *later* conversion of both files that supersedes what was here:

- Main removed the `workflowMode ? trait : literal` gate entirely, on the argument that `workflowMode` is a **board**-level boolean standing in for a **per-column** question — so a workflow-mode column with no resolved traits answered `false` for every role and silently lost its archive, promote and bulk affordances. This branch's `columnRoleFlags = workflowMode ? (columnFlags ?? {}) : undefined` adapter preserved exactly that behaviour deliberately.
- Main's `getTaskColumnFlags` (from #2738 review) resolves per **task** via the task's own workflow, which is strictly better than this branch's `columnFlagsById.get(task.column)` union lookup — the union misattributes a card to a neighbouring workflow's traits when two workflows reuse a column id.

**That means commit `923d16a43a` ("not declared is an answer — not a legacy fallback") is now a no-op for those two files, and main answers its question the other way.** This branch argued `?? {}` so an undeclared column answers false; main's `getTaskColumnFlags` deliberately returns `undefined` there and lets the helper degrade to the legacy id, with its own FNXC note making the case. Main's is newer and reviewed, so I did not re-litigate it during a rebase — flagging it here instead. If the `?? {}` reading is the correct one, it should be a fresh change against main's code, not a conflict resolution.

**`useTaskDiffStats.ts` — I took this branch's side.** Main only added `DELIBERATE-LITERAL` markers there; this branch does the actual conversion (`columnFlags` option threaded from `TaskCard`, plus the re-poll fix in `4d05d9159e`). Markers replaced by the real thing.

**Census baseline re-recorded** — `--strict` now exits 0; it was red immediately after the rebase because the markers this branch removes were still in the baseline.

Verification on the rebased tip: dashboard app typecheck clean, `pnpm lint` clean, `ListView.test.tsx` + hooks suites **140 files / 2070 tests passed**.

PR #2734, opened 2026-07-30
fleet: default-workflow-hooks.ts 7 → 0 — every duration display read ZERO on a renamed board

gsxdsm:
Claiming `default-workflow-hooks.ts` (7 → **0**), verified free against every open PR's diff first.

## Not a vocabulary tidy — three silent zeroes

This file's header names it for the default workflow, but the store runs it on the flag-ON path for **every** workflow: the trait registry resolves each hook by **trait id**, not by workflow. `reopen-semantics-by-role.test.ts` already documents that exact hazard for the reopen predicates. The **timing, completion and in-review hooks had the same defect** and were not part of that conversion.

On a renamed board, with nothing thrown and nothing logged:

- **`applyTimingEffects`** accrues `cumulativeActiveMs` while a card sits in the WIP lane. With the lane named, the exit test never fires — so **no active time is ever accrued**, and `productivity-analytics.ts`, `task-timing.ts` and every duration display read **zero**.
- **`applyCompletionTimingEffects`** never stamps `executionCompletedAt`, so a finished card looks unfinished to anything reading that field.
- **`applyInReviewEnterEffects`** returns early, leaving the recovery counters set.

The file already had the idiom — `ctx.lifecycleColumns`, `planningColumnsOf`, `liveWorkColumnsOf` with `LEGACY_` fallbacks — so this adds no abstraction.

One deliberate detail: `applyTimingEffects` resolves the WIP lane **once into a local** rather than reading it twice. The exit test and the re-entry test have to agree about which column is WIP, or a rename makes the accounting count an interval twice, or not at all.

## A test that would have lied to me

I wrote the new cases through `applyDefaultWorkflowMoveEffects` first, and **all three failed on the DEFAULT lineage too**. The dispatcher resolves hooks by trait, and neither test IR declares the `timing` trait, so those hooks never ran at all.

That failure looks exactly like a conversion bug. Going through the dispatcher would have been testing the trait registry's wiring rather than this change — so the cases call the converted functions directly, and the reason is recorded in the test.

## Revert proof — all three, each naming the renamed lineage

| reverted | failure |
|---|---|
| the `in-progress` literals | `renamed lineage accrued no active time: expected undefined to be 300000` |
| the `done` literal | `renamed lineage did not stamp completion: expected undefined to be '2026-07-30T00:00:00.000Z'` |
| the `in-review` literal | `renamed lineage kept its recovery counter: expected 3 to be undefined` |

Every case runs on **both** lineages and the default one passes either way — which is the point of running it.

## A finding I did not act on

**`evaluateMergeBlockerGuard` appears exactly once in the repo — its own definition.** And the file header says it is "implemented as the `evaluateDefaultWorkflowGuards` reader", which does not exist either. The merge-blocker guard hook is **defined and never consulted**.

I converted it (trailing optional lifecycle param, matching `DefaultWorkflowMoveContext`) but did not delete it: the header states this file is a deliberate parallel of `store.ts`'s flag-off path so the two can be parity-checked, which makes removing it a scope call for whoever owns that convergence — not something to decide inside a conversion.

## Verification

`pnpm test:gate` **GREEN** (158 + 10 + 487 + 71) · **22 passed** across `default-workflow-hooks` + `reopen-semantics-by-role` · core `tsc` clean · `pnpm lint` clean · census `--strict` exits 0.

### DIFF
--- a/packages/core/src/__tests__/postgres/moves-review-set-from-producer.pg.test.ts
+++ b/packages/core/src/__tests__/postgres/moves-review-set-from-producer.pg.test.ts
@@ -1,0 +1,97 @@
+/*
+FNXC:WorkflowLifecycleColumns 2026-07-30-15:10 (PR #2734 review — greptile, on my own code):
+
+THE PRODUCER IS THE THING UNDER TEST, not the hook.
+
+`moves.ts` builds `lifecycleColumnSets.review` and hands it to the default-workflow hooks. It built
+that set from `columnsWithFlag(ir, "mergeOrchestration")` alone, so a workflow hosting review on a
+`humanReview`-only lane produced an EMPTY review set — and `applyInReviewEnterEffects` then never ran
+for a card plainly in review, leaving the recovery counters it clears set.
+
+WHY THIS FILE EXISTS RATHER THAN A UNIT CASE. My first attempt asserted the hook directly, passing
+`lifecycleColumnSets: { review: ["signoff"] }` by hand. That exercises the hook — which was already
+correct — and passes with the producer's bug fully in place. Reverting the producer left it green.
+Only driving a real move through a real store reaches the code that was wrong.
+
+LANE. `.pg.test.ts`, skipped by `pgDescribe` when no PostgreSQL is reachable, so the merge gate is
+unaffected. Throwaway per-file database; never port 4040.
+*/
+import { beforeAll, beforeEach, afterEach, afterAll, expect, it } from "vitest";
+import "@fusion/core"; // registers the built-in column traits
+
+import { pgDescribe, createSharedPgTaskStoreTestHarness } from "../../__test-utils__/pg-test-harness.js";
+
+pgDescribe("moves.ts supplies EVERY review lane to the lifecycle hooks", () => {
+  const harness = createSharedPgTaskStoreTestHarness({ prefix: "fusion_moves_review_set" });
+
+  beforeAll(harness.beforeAll);
+  afterAll(harness.afterAll);
+  beforeEach(async () => { await harness.beforeEach(); });
+  afterEach(async () => { await harness.afterEach(); });
+
+  it("runs the in-review enter effects for a humanReview-ONLY lane", async () => {
+    const store = harness.store();
+    const definition = await store.createWorkflowDefinition({
+      name: "human-review-only",
+      ir: {
+        version: "v2",
+        name: "human-review-only",
+        columns: [
+          { id: "backlog", name: "Backlog", traits: [{ trait: "intake" }, { trait: "hold" }] },
+          { id: "building", name: "Building", traits: [{ trait: "wip" }] },
+          /* No `merge` trait: review is hosted by human-review alone. */
+          { id: "signoff", name: "Sign-off", traits: [{ trait: "human-review" }] },
+          { id: "shipped", name: "Shipped", traits: [{ trait: "complete" }] },
+        ],
+        nodes: [{ id: "start", kind: "start", column: "backlog" }, { id: "end", kind: "end", column: "shipped" }],
+        edges: [{ from: "start", to: "end" }],
+      },
+    } as never);
+
+    const task = await store.createTask({ description: "human-review card", workflowId: definition.id } as never);
+    await store.moveTask(task.id, "building" as never, { bypassGuards: true } as never);
+
+    /*
+    The witness: `applyInReviewEnterEffects` clears `recoveryRetryCount`. Seed it so its absence after
+    the move can only mean the enter-effects ran.
+    */
+    await store.updateTask(task.id, { recoveryRetryCount: 3 } as never);
+    store.taskCache.delete(task.id);
+    expect((await store.getTask(task.id)).recoveryRetryCount).toBe(3);
+
+    await store.moveTask(task.id, "signoff" as never, { bypassGuards: true } as never);
+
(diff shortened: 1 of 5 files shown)


gsxdsm:
Real, and the fix needed a shape the struct cannot express.

`LifecycleColumns` names **one column per role by design** — #2721 pinned that as the contract, not a gap — so a workflow declaring `countsTowardWip` (or `complete`, or `mergeOrchestration`) on several columns had lanes these hooks did not recognise.

**The timing consequence is the concrete one.** A card moving *between* two WIP lanes looked like an exit from WIP followed by a re-entry, so `cumulativeActiveMs` closed and reopened a segment the card never left; a card living only in the secondary lane accrued nothing at all. Neither surfaces as an error — the numbers are simply wrong, which is the expensive kind.

`lifecycleColumnSets` is the set-shaped companion: **optional and additive**, so absent it falls back to the singular struct and then the legacy id — the default lineage and any caller that doesn't supply it are unchanged. `moves.ts` already holds `workflowIr`, so populating it costs three trait scans and no extra read.

One `inRole` helper serves all three roles rather than three inline expressions. The original comment warned that the exit test and the re-entry test must agree about which column is WIP; two copies of a membership check is exactly how that disagreement comes back.

**Mutation: restoring the singular WIP comparison fails the between-lanes case while the leaves-WIP case still passes** — without both, the two are indistinguishable.

---

**Two process notes.**

Lint caught a dead `wipColumn` binding my replacement left behind. Worth recording because my first read of the verification output recorded lint as **clean when it was not** — the error line sits above the summary block I was scanning, and I reported the wrong result to myself before re-running.

I also ran the collision check before starting (`git log HEAD..origin/<branch>`, 0 commits) after duplicating another worker's fix on #2733 an hour ago. Before force-pushing I checked again with `--not HEAD` and confirmed both remote commits were reproduced in the rebase — one of them my own earlier push.

Verified: 16 hook tests + moves equivalence green, core `tsc` clean, lint clean (0 errors), gate green (487 + 158 + 10 + 71), `census --strict` exit 0.

gsxdsm:
Real, and it is on code I wrote — the same distinction I had applied twice elsewhere hours earlier and then got wrong introducing this helper.

`lifecycleColumnSets` is supplied **only** when the caller resolved a workflow IR, so `set !== undefined` already means the board was read. An **empty** array then says "no column carries this role" — and my `length > 0` guard treated that as no-basis, falling back to the singular id and then the legacy name.

The consequence on a valid v2 workflow with no WIP lane: a traitless column merely **named** `in-progress` accrued timing as though it were the WIP lane.

The rule, stated once so it stops recurring: **`undefined` means "could not read"; empty means "read, and the answer is none".** Same shape as #2731's `?? {}` for resolved-but-roleless flags and #2733's refusal to invent a complete column.

## The test could not fail until I fixed its fixture

My first version left `lifecycleColumns.wip` populated, so the buggy fallback landed on *that* name and answered `false` anyway — the case passed with the defect fully in place. Only with the singular role **absent** does the bug reach the legacy name, which is the path that actually matters.

**Mutation: restoring `length > 0` now fails 1 of 17.**

That is the third time today one of my tests needed its fixture corrected before it could discriminate — the others being a `sticky-only` mode where the flag under test was never consulted, and a set argument that became a truthy value after a signature change. In each case the test passed, and the passing was the problem. I no longer treat green-on-first-write as evidence of anything.

Verified: 19 core tests green, core `tsc` clean, lint clean (0 errors), gate green (487 + 158 + 10 + 71), `census --strict` exit 0. Rebased onto current main; force-push checked with `--not HEAD` and every prior commit's content confirmed present (5 `lifecycleColumnSets` references in the hooks, 1 in the producer).

gsxdsm:
Real, and on my own code — the producer I added two rounds ago.

`lifecycleColumnSets.review` was built from `columnsWithFlag(ir, "mergeOrchestration")` alone, so a workflow hosting review on a `humanReview`- or `mergeBlocker`-only lane produced an **empty** review set. `applyInReviewEnterEffects` then never ran for a card plainly in review, leaving the recovery counters it clears set.

Fixed by calling **`resolveReviewColumns`** (core, merged in #2730) — making this the first consumer to use the shared answer instead of a fifth inline union.

That is the **broad** set, which is correct here: these hooks *ask* the question and move nothing on the answer. A caller that admits and then **moves** wants the narrow single lane instead; #2750 documents that split at the helper, precisely so this choice is a decision rather than an accident.

## The first test I wrote proved nothing

I asserted the **hook** directly, passing `lifecycleColumnSets: { review: ["signoff"] }` by hand. That exercises the hook — which was already correct — and passes with the producer's bug fully in place. Reverting the producer left it green.

The producer is what was wrong, so the test now drives a **real move through a real store**: a `humanReview`-only lane, `recoveryRetryCount: 3` seeded, asserted cleared after the move. **Mutation: reverting the producer fails it.** A paired negative keeps the widening from making every column a review lane.

That is the fourth fixture-level miss I have caught in myself today — the others being a `sticky-only` mode where the flag under test was never consulted, a set argument that became a truthy value after a signature change, and a singular role that masked the fallback path. The pattern is consistent enough to name: **when a fix moves data from producer to consumer, a test written at the consumer proves the consumer.** The mutation has to target the line that changed.

Verified: 20 core tests green (2 PG + 18 unit), core `tsc` clean, lint clean (0 errors), gate green (487 + 158 + 10 + 71), `census --strict` exit 0.

PR #2737, opened 2026-07-30
fleet: github-tracking-reconciler 9 → 0 — deciding the sync-filter class (prefetch a resolved map), and the reconciler closed NO issues on a renamed board

gsxdsm:
`github-tracking-reconciler.ts` 9 → **0**, and the reference implementation for the `.filter((task) => task.column === "<id>")` shape I have been flagging across four files.

## I stopped waiting and decided it

I flagged this class in #2709, #2696, #2700 and #2715 as "needs one decision" and left ~25 sites unconverted. That decision was mine to make and I should have made it three PRs ago.

**Prefetch a resolved map, then filter synchronously.** The alternative — async predicates — forces every caller into `for await` and turns a list comprehension into a sequential walk. Prefetching keeps the filters synchronous, puts the awaits in one bounded place, and lets the IR cache do the job it was explicitly built for:

> "A self-healing pass over 400 cards spanning three workflows must read three IRs, not 400."

The cache is **instance-scoped and shared across all four passes**, so each distinct workflow's IR is read once for the whole run rather than once per pass. `resolveLifecycleColumns` is pure and *not* memoized by that cache, so this still costs one cheap struct build per task — fine in a background reconcile, and stated rather than hidden.

No new abstraction: `resolveTaskLifecycleColumns` already takes a caller-owned cache. The only new code is a local map builder and two named predicates.

## What it cost before

On a board with renamed terminal lanes, **every filter here matched nothing**. The reconciler closed **no** GitHub issues and reported `scanned: 0` — a clean-looking pass that did nothing.

## Why this is not the split brain #2724 documents — checked, not assumed

#2724 proves the archived gate in `packages/core` is enforced in three encodings, so converting one alone diverges them. I checked whether that applies here before converting:

- This file contains **zero SQL** — measured: no drizzle, no `sql` template, no `eq`/`ne`.
- It calls `listTasks({ includeArchived: true })`, so the SQL half has already been told to include archived rows. The filter **selects among rows it was handed** rather than deciding liveness a second time.

**Gate versus consumer** is the distinction, and a consumer can be converted alone.

The fourth pass needed its own check because its list comes from `listTasksForGithubTrackingReconcile`, which *is* SQL — but that impl filters on `deletedAt IS NOT NULL` and `githubTracking IS NOT NULL`, **never on the column**, so there is no SQL-side encoding of this question to diverge from.

## Why the 33 existing tests stayed green through the conversion

Their fake store has **no workflow reader**, so `resolveTaskLifecycleColumns` catches and returns `undefined` and every case asserts the legacy fallback — exactly what it always asserted. **None of them could have caught this being wrong.** `workflowIr` is now an opt-in on that fake, which is what makes the new cases real tests rather than restatements.

| reverted | result |
|---|---|
| terminal filter back to the ids | "closes issues on a RENAMED complete lane" fails, no `setIssueState` |
| same | renamed archived-heuristic case fails, no `setIssueState` |

## A reachability finding, recorded not acted on

In backend mode `reconcileDeletedAndArchived` returns only **soft-deleted** rows — its own comment says the archived-tasks fallback is a separate `AsyncArchiveLineage` subsystem, skipped there — and `task.deletedAt` is tested *first* in the `stateReason` chain. So its archived arm is **effectively unreachable today**. I converted it rather than deleting it: it is the documented FN-5577 done-heuristic, and whether that fallback should be wired here is a separate question from what vocabulary it speaks.

## Verification

`pnpm test:gate` **GREEN** (158 + 10 + 487 + 71) · **35 passed** across the three reconciler suites · dashboard `tsc` clean · `pnpm lint` clean · census `--strict` exits 0.

Remaining files in this class (`branch-group-ops.ts`, `store.ts`, and the dependency pairs) can now follow this pattern instead of waiting — with the gate-versus-consumer check applied to each, since `branch-group-ops.ts` sits closer to the persistence layer than this one does.


gsxdsm:
Both findings are real and fixed. The first one is the more embarrassing of the two, because I had already argued the opposite position in a neighbouring file.

**1. Workflow cache stays stale (P1) — correct.** The IR cache was instance-scoped on a long-lived reconciler, so an operator's workflow edit would never be picked up. In `notification-service.ts` (#2722) I explicitly *refused* a service-lifetime cache for exactly this reason and wrote the argument down:

> there is no workflow-changed event to invalidate on, so a cached IR would outlive an operator's workflow edit

Then I used one here two PRs later. The cache is now allocated **per pass**, which is the shape `resolveTaskLifecycleColumns` documents ("one per sweep") and the only one that is correct without an invalidation signal. Within a pass an IR is still read once per distinct workflow, so the cost argument for having a cache at all is unaffected.

**2. Lifecycle prefetch bypasses the scan limit (P2) — correct.** The prefetch resolved for every row `listTasks` returned — an unbounded board — before slicing to `RECONCILE_SCAN_LIMIT`. Unbounded work to feed a bounded scan.

`resolveLifecycleByTaskId` now takes an optional `match` predicate and `limit` and **stops once `limit` rows have matched**. The predicate lives inside the helper rather than at the call site precisely so the loop can stop early.

Rows past the cut are left unresolved and absent from the map, which is safe and worth stating explicitly: the only consumers are the terminal predicates, which fall back to the legacy ids for an absent entry — the same degraded answer they give on a store with no workflow reader — and those rows are dropped by the slice anyway.

The fourth pass (`reconcileDeletedAndArchived`) is already bounded by its own paginated store query, so it resolves for the page only and needs no limit.

### Verification

Gate **GREEN** (158 + 10 + 487 + 71) · **35 passed** across the three reconciler suites · dashboard `tsc` clean · `pnpm lint` clean · census `--strict` exits 0.

PR #2738, opened 2026-07-30
fleet: Column.tsx 7 → 0 + ListView.tsx 6 → 0 — a board-level flag was answering a per-column question, blanking every affordance on an undeclared column

gsxdsm:
## Census

| | before | after |
|---|---|---|
| `packages/dashboard/app/components/Column.tsx` | 7 | **0** |
| `packages/dashboard/app/components/ListView.tsx` | 6 | **0** |
| repo backlog | 610 | **597** |

## This is a real divergence, not a rename

Both files asked every column-role question as:

```ts
workflowMode ? Boolean(columnFlags?.archived) : column === "archived"
```

`workflowMode` is `Boolean(boardWorkflows?.workflows.length)` — a **board-level** boolean standing in for a **per-column** question. They disagree in one state, and it is a state that really occurs: the board is in workflow mode, but *this* column has no resolved traits, because a mid-flight workflow edit left a card in a column the workflow no longer declares.

There the ternary takes its **first** arm and evaluates `Boolean(undefined?.archived)` → `false`. Not "fall back to the legacy id" — **`false` for every role at once**. For that column the archive action, the revert action, the promote affordance, the auto-merge toggle, the Done-sort control and every bulk action silently disappeared, with nothing in the UI explaining why.

The shared helpers ask per column: traits when resolved, legacy id **only** when the flags are genuinely absent. That covers both the pre-load window the old form handled via `workflowMode === false` *and* the undeclared-column case it got wrong.

**I am flagging this explicitly as a deliberate behaviour change**, not a mechanical conversion — the fleet rules put behaviour changes out of scope, so this one is called out rather than slipped in. It is pinned by a test so it stays deliberate, and it is the documented contract of the shared helper (`columnRoles.ts` spells out "a card in a column its workflow no longer declares" as exactly the case the id fallback exists for). Happy to split it out if you would rather it land separately.

## Second defect: the list progress bar

`shouldShowTaskProgress` keyed on `column === "in-progress"`, so on a renamed wip column **no card showed a progress bar** until its status flipped to `executing` — the row looked idle while an agent was working in it.

Threaded rather than half-converted: `shouldShowTaskProgress` and `getTaskProgress` now take the column's flags, passed from **both** call sites, where `columnFlagsById` was already in scope. (Converting the predicate and leaving the callers is the trap that bit #2736 — there, half a fix turned a silent no-op into a crash.)

## Revert proof

| reverted | result |
|---|---|
| `Column.tsx` → `workflowMode` ternary | **1 failed** / 6 passed |
| `ListView.tsx` → `workflowMode` ternary | **1 failed** / 6 passed |
| neither (shipped) | **7 passed** |

The ratchet strips comments before matching — the FNXC notes legitimately quote the old shape — and carries a **completeness case**: it fails if a refactor drops the helpers instead of converting to them, so the guard cannot go vacuous.

It also asserts the divergence as behaviour in both directions: degrading to the legacy id when flags are absent, **and** that degrading does not turn into "every column has every role".

## Verification

- 17 board/list/column/lane suites — **109 passed**, no regressions
- new ratchet — **7 passed**
- `pnpm test:gate` — **158 / 10 / 487 / 71** · `pnpm lint` clean · dashboard `tsc -p tsconfig.app.json` clean

(`tsconfig.app.json` on purpose — the default `tsconfig.json` silently checks nothing under `app/`.)


gsxdsm:
Correct, and it is my change that widened the exposure. Fixed, and the guard I wrote for it then found four more.

`columnFlagsById` is a **union across workflows keyed by column id** — the file already documents exactly that for `moveTargets` ("two workflows that reuse an id but declare different edges collapse into one entry"). The union was harmless while flags answered **column-level** questions (`isArchivedColumn(column)` for a whole list section). Converting the row context menu and the progress bar made them **per-task** questions, and there the union serves one workflow's `complete`/`archived`/`countsTowardWip` to another workflow's card.

Now resolved per task through the same validated mapping `taskContextMenuColumnsByTaskId` already uses — unmapped task → no metadata, stale id → default — falling back to the union, which is the pre-existing approximation rather than a confidently wrong answer.

### The ratchet found four more, all pre-existing

| site | what it did |
|---|---|
| `isIntakeColumnForTask` | named *ForTask*, looked up the union — the Planning badge followed a **neighbouring workflow's** `intake` trait |
| `currentColumnFlags` | the per-task metadata handed to the detail modal |
| `isTaskAgentActive` ×2 | the per-task agent-active decision |

All four are the same one-line fix, so they are corrected here rather than left as a known-wrong neighbour of a just-fixed site.

### My first version of the guard was too weak, and mutation testing proved it

It matched only inline `isXColumnRole(columnFlagsById.get(task.column)`. Reverting through a **local** walked straight past it:

```ts
const taskColumnFlags = columnFlagsById.get(task.column);  // bug reproduced, guard still green
```

The rule is now about the **lookup**, wherever its result goes: any `columnFlagsById.get(task.column)` outside the accessor fails. The accessor is excised **by position** rather than special-cased in the regex, so the exemption cannot be widened by accident.

### Mutation matrix — each reverted alone

| reverted | result |
|---|---|
| archive/revert via a local union read | **1 failed** / 8 passed |
| row progress | **1 failed** / 8 passed |
| intake badge | **1 failed** / 8 passed |
| detail-modal metadata | **1 failed** / 8 passed |
| none (shipped) | **9 passed** |

Verification: 17 board/list/column/lane suites **111 passed**; `pnpm test:gate` **10 / 158 / 487 / 71**; `pnpm lint` clean; dashboard `tsc -p tsconfig.app.json` clean; `check:lifecycle-columns --strict` exits 0.

gsxdsm:
Right, and it is the sharper version of my own argument turned back on me — I made exactly this case in the PR body for `Column.tsx` and then failed to apply it to my own fallback. Applied your suggestion.

The fallback collapsed **three** states into two:

| | state | correct answer |
|---|---|---|
| 1 | the task's workflow declares the column | those flags |
| 2 | the task's workflow is **known** but does not declare it | **`undefined`** → legacy-id degrade |
| 3 | no per-task metadata at all | the union — an admitted approximation |

State 2 is the **stranded card**: a task whose own workflow no longer declares its column, which is the precise case this PR exists for. Falling through to the union there handed it a *neighbouring* workflow's traits for the same id, so Archive/Revert, progress visibility, the Planning badge and agent-active styling all followed a workflow the card does not belong to.

Absent flags is the right answer, because the role helpers then degrade to the legacy id — the documented no-metadata path.

Pinned as a source assertion because this is a **resolution-precedence rule**, not a value: the failure is "the wrong one of three branches was taken", which a value-level test cannot distinguish without reconstructing the whole board-workflows payload. The test spells the three states out so the next reader does not re-collapse them.

Revert-proofed: restoring `fromOwnWorkflow ?? columnFlagsById.get(task.column)` fails the new case — **1 failed / 10 passed**; shipped **11 passed**.

Verification: 17 board/list/column/lane suites **113 passed**; `pnpm lint` clean; dashboard `tsc -p tsconfig.app.json` clean.

gsxdsm:
Correct, and this is the round that reached the **destructive** operations — thank you.

Thirteen call sites passed `task.column` into the **column-level** predicates (`isCompleteColumn` / `isArchivedColumn`), which read the cross-workflow union. On a board where two workflows reuse a column id with different `complete`/`archived` traits, bulk **select-all, delete, archive, pause, unpause and model updates** each classified cards by a *neighbouring* workflow's semantics — cards silently skipped, or the wrong branch of a destructive action taken.

### Why my own ratchet missed it — three rounds, one lesson

| round | guard | how it was evaded |
|---|---|---|
| 1 | forbade `isXColumnRole(columnFlagsById.get(task.column))` | a **local variable** carried the lookup |
| 2 | widened to the **lookup**, wherever its result goes | a **callback** — the call site contains no lookup at all |
| 3 | *this* | — |

The rule was never about syntax. It is: **a question asked about a task must be answered from that task's own workflow.** The guard now forbids passing `task.column` to any bare-column-id predicate, however the union is reached. I have written the reasoning into the test so round four doesn't have to rediscover it.

### Also fixed, found while wiring it

**Six hooks carried stale dependency arrays**, still naming the old callbacks after their bodies stopped using them — captured-stale-closure bugs. This repo deliberately omits `react-hooks/exhaustive-deps`, so neither eslint nor `tsc` says a word. I corrected them with an **AST pass over the hook call sites** rather than by eye, because that is exactly the kind of thing eyes miss:

```
line 1761: +[isTaskArchivedColumn] -[isArchivedColumn]
line 1675: +[isTaskCompleteColumn] -[isCompleteColumn]
line 1624: +[isTaskArchivedColumn] -[isArchivedColumn]
line 1573: +[isTaskArchivedColumn] -[isArchivedColumn]
line 1368: +[isTaskArchivedColumn,isTaskCompleteColumn] -[isArchivedColumn,isCompleteColumn]
line 1275: +[isTaskArchivedColumn] -[isArchivedColumn]
```

`isCompleteColumn` is now **deleted** — all five of its call sites turned out to be per-task. `isArchivedColumn` survives for its one genuinely column-level use (a list *section* header), which is where the union is harmless.

### Revert proof

| reverted | result |
|---|---|
| bulk **delete** → column-level predicate | **1 failed** / 10 passed |
| bulk **archive** → column-level predicate | **1 failed** / 10 passed |
| neither (shipped) | **11 passed** |

### Verification

18 board/list/column/lane suites — **116 passed** · `pnpm test:gate` **158 / 10 / 487 / 71** · `pnpm lint` clean · dashboard `tsc -p tsconfig.app.json` clean · `--strict` exits 0.

gsxdsm:
**Overlap with `batch-dashboard-app`, and a recommended order — flagging rather than racing.**

The mega-batch branch has just landed `Column.tsx 7 → 0 — six legacy-board arms marked, one redundant ternary deleted`. That is the same file this PR converts, so the two will conflict on merge. I posted a claim into the batch work order, but their largest-first pass reached `Column.tsx` before the note did.

**The two changes are not equivalent, and the difference is the point:**

| | batch branch | this PR |
|---|---|---|
| `Column.tsx` | six arms **marked** `DELIBERATE-LITERAL` | six arms **converted** to role helpers |
| the `workflowMode` defect | not addressed | **fixed** |

`workflowMode` is `Boolean(boardWorkflows?.workflows.length)` — a **board-level** boolean answering a **per-column** question. When the board is in workflow mode but a column has no resolved traits (a mid-flight workflow edit leaves exactly this), the ternary took its first arm and returned `false` for *every* role — not "fall back to the id", but no role at all. Archive, revert, promote, auto-merge toggle, Done-sort and every bulk action silently disappeared from that column.

Marking those arms documents them; it does not fix that. This PR is revert-proofed against it (restoring either file's ternary fails the ratchet).

**Recommended order:** land this first, then rebase the batch onto it — their `Column.tsx` marks become redundant and should be dropped in the rebase. The reverse order loses the fix and keeps markers on arms that are still wrong.

If you'd rather keep the batch as the single landing point, say so and I'll port this PR's `Column.tsx` + `ListView.tsx` changes onto the batch branch as commits and close this — but I don't want to do that unilaterally while it is green and carries a defect fix the batch does not.

`ListView.tsx` is the next entry by size on their list and is also in this PR; the work-order note should stop that one duplicating.

gsxdsm:
Superseded by `batch-dashboard-app`, where this now lives as a commit — per the mega-batch instruction that dashboard/app closes as one package PR.

Settling the collision I flagged above rather than racing merge order. The batch had reached census 0 on both files by **marking** their `workflowMode ? trait : literal` ternaries; I have replaced those marks with this PR's conversion and brought the ratchet with it.

**Why the marks could not stand.** The census read 0 for both files while **26 `workflowMode ?` ternaries remained**, and the defect they encode was untouched: `workflowMode` is a board-level boolean answering a per-column question, so when the board is in workflow mode but a column has no resolved traits — which a mid-flight workflow edit leaves — the ternary takes its first arm and returns false for **every** role. Not "fall back to the id": no role at all. Archive, revert, promote, auto-merge toggle, Done-sort and every bulk action silently vanish from that column.

A count of 0 with the bug intact is precisely the outcome this program exists to prevent.

Everything here landed on the batch: the two conversions, the per-task flag resolution for ListView's bulk actions and row menus, the six stale hook dependency arrays (which this repo's eslint cannot catch, since it omits `react-hooks/exhaustive-deps`), and the three-round ratchet.

Ternaries drop 18 → 8 in Column.tsx and 8 → 6 in ListView.tsx; the rest are non-role uses and are left alone. Whole dashboard app suite **112 files / 3919 tests** passing on the batch.

PR #2739, opened 2026-07-30
fleet: branch-group-ops 6 → 0 — an agent asking for its next task was told there was none

gsxdsm:
Second application of the sync-filter pattern decided in #2737. `branch-group-ops.ts` 6 → **0**.

## The failure

`selectNextTaskForAgentImpl` picks an agent's next task by filtering the board for its WIP lane, then its hold lane — both `task.column === "<literal>"`.

On a renamed board **both filters match nothing**, so an agent asking for work is told there is none, with its own assigned tasks sitting in the list it just fetched. No error, no log line. The agent idles.

`pauseTaskImpl` had the same shape: pausing a running card on a renamed board left its `status` untouched, so the UI kept showing it as working.

## Consumer, not a gate — checked rather than assumed

Applying the #2724 test to this file, since it sits closer to the persistence layer than the reconciler did: its **only** SQL predicate is `eq(table.projectId, ...)`. Nothing here compares a column to a literal in SQL, so there is no second encoding of these questions to diverge from. The list arrives from `store.listTasks` and the filters select among rows already in hand.

Async predicates were the alternative and would have turned these filter chains into sequential awaits inside the dispatch path. One prefetch, one IR read per distinct workflow, filters stay synchronous.

`pauseTaskImpl` resolves for the single task it holds rather than joining a map — different entry point, one id in scope, and a map would have exactly one entry.

## Revert proof

| reverted | result |
|---|---|
| the wip literal | renamed WIP case fails: `expected null to be truthy` |
| the hold literal | renamed hold case fails identically |

The new test calls the impl **directly** with a store fake resolving a renamed IR. The existing `selectNextTaskForAgent` coverage drives a real store harness, so exercising a renamed vocabulary there means registering a real custom workflow and moving cards through it — heavier than the question, which is only which lane the filters name. The bind evaluator runs for real; only the store is faked.

A third case pins that the hold filter keeps its `userPaused` exclusion, so a filter matching every column would not satisfy the other two.

**Related coverage checked before writing a new file:** `agent-heartbeat-worktree-renamed-hold.test.ts` covers the requeue **target** on a renamed board, not the dispatcher's **selection** filters — different branch of the same subsystem, so a case added there would have read as duplicate coverage of the wrong thing.

## Verification

`pnpm test:gate` **GREEN** (158 + 10 + 487 + 71) · **20 passed** across the routing-policy and new dispatch suites · core `tsc` clean · `pnpm lint` clean · census `--strict` exits 0.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved agent task selection when workflow lane names or IDs have been renamed.
  - Agents now correctly resume assigned in-progress or queued tasks across customized workflows.
  - Prevented agents from selecting tasks paused by users, including on boards with renamed lanes.
  - Updated task pausing behavior to correctly reflect lifecycle stages beyond default lane names.


gsxdsm:
**P2 (dispatch resolves every task sequentially) — correct, fixed.** The prefetch walked every active task before any filtering, so a 400-card board paid 400 resolutions to dispatch one agent that owns three. `assignedAgentId` is a plain property read and was already the very next filter, so hoisting it above the prefetch costs nothing and bounds the work by what the agent actually owns. Unowned tasks are absent from the map, which is safe: every consumer is a predicate over `assignedTasks`, and `isDoneLike` runs on *dependency* tasks that fall back to the legacy ids for an absent entry — the same degraded answer a store without a workflow reader gives.

**P1 (secondary lifecycle lanes are ignored) — real, but not this PR's to fix, and I want to be precise about why rather than wave it off.**

The finding is accurate: `resolveLifecycleColumns` returns the **first** column carrying each trait, so a workflow declaring two WIP lanes has tasks in the second one invisible to these filters. That is a genuine gap and an agent would idle on such a board.

It is a property of the **shared resolver**, not of this conversion. `resolveLifecycleColumns` returns one id per role by construction, and roughly 30 consumers across core, engine and dashboard read it that way. Fixing it only in `branch-group-ops.ts` would make the dispatcher the single place with different arity semantics — the drift this program exists to remove, arriving from the other direction.

There is already precedent for the right shape: `register-task-workflow-routes.ts` resolves review as **membership over both traits** rather than a single id, which is why I closed #2700 as superseded by it. The equivalent here would be membership sets per role — `columnsWithFlag(ir, "countsTowardWip")` — which means changing what the resolver returns, or adding a parallel membership resolver, and migrating consumers together.

So: recorded as a real gap at the resolver level, not converted here. If a resolver-arity change is in scope for someone, this file is a consumer that wants it, and the routes file is the reference implementation. I would rather leave the gap visible than fix it in one of thirty places.

### Verification

Gate **GREEN** (158 + 10 + 487 + 71) · **20 passed** across the dispatch and routing-policy suites · core `tsc` clean · `pnpm lint` clean · census `--strict` exits 0.

gsxdsm:
All three threads were correct, and all three were the same fault — **a comment claiming something the code did not do.** The third one turned out to matter well beyond this PR.

**1. One IR cache, as the comment already claimed.** The dispatch pass allocated `dispatchIrCache` while the dependency pass allocated `satisfiedIrCache`, so a task and its dependency in the same workflow resolved that IR **twice**. The comment asserted an optimisation the code did not perform. A single `lifecycleIrCache` now feeds both loops.

**2. Future-dated FNXC timestamps.** Correct, and a recurrence — I future-dated six of these earlier in this program and corrected them then. Fixed to `2026-07-30`.

**3. The bind-evaluator coverage claim was false — and proving it found a real defect.**

The header said "the bind evaluator is exercised for real" while the call omitted `selectNextTaskForAgentImpl`'s **optional** `agent` argument, so `isBindCompatible` hit `if (!agent) return true` and `evaluateImplementationTaskBind` never ran. Exactly right.

Passing an agent is not proof it runs — an executor is *allowed*, so a short-circuit and a real evaluation are indistinguishable from a passing test. So I asserted a `custom`-role agent is **refused** an implementation task on a renamed lane.

**It failed. The cause is production, not the test:**

> `isImplementationTask` is Set membership over the hardcoded ids `{triage, todo, in-progress, …}`, and `evaluateImplementationTaskBind` short-circuits to `allowed: true` when it returns false. **On a renamed board every agent is bind-compatible with every task** — the role check that stops a liaison/custom agent being handed implementation work, the NEXT-871 loop FN-7851 fixed, does not apply at all.

**Why the census never flagged it:** these are Set **members**, not comparisons. The lifecycle-column census scans `===`/`!==` against a column, so a literal collection is invisible to it — the same blind spot that hid the raw-`sql` encoding of the archived gate in #2724. **The backlog number is a floor, not a total**, and that is worth knowing while the fleet is measuring progress against it.

**Not fixed here.** `isImplementationTask` is a sync pure predicate with no store, called from the gate for agent assignment. Resolving a workflow inside it means threading a resolver through the routing policy and making its callers async — and getting it wrong either hands implementation work to a liaison or refuses it to a valid executor. That is a behaviour change to agent admission, not a vocabulary conversion. The reasoning is recorded at `agent-role-policy.ts`'s `IMPLEMENTATION_TASK_COLUMNS`, and the new case is written to **current** behaviour and named as documenting a defect so it does not sit red — flip the expectation when the policy resolves lanes by role.

### Verification

Gate **GREEN** (158 + 10 + 487 + 71) · **21 passed** across the dispatch and routing-policy suites · core `tsc` clean · `pnpm lint` clean · census `--strict` exits 0.

gsxdsm:
**Correct, and fixed — with `columnsWithFlag` rather than the resolver change I declined on the earlier thread.** The distinction matters, so I want to be explicit that this is not me reversing myself:

- The **earlier** arity thread was about `lifecycle.wip` — a **single-id** question. Making that membership means changing what `resolveLifecycleColumns` returns for ~30 consumers, which is a resolver-level change.
- **This** loop already builds a `Set` and already unions across dependencies. Membership is what it was always trying to express. No resolver change, no consumer migration, and it matches the `register-task-workflow-routes` precedent I cited when closing #2700.

So: a workflow with two complete lanes (complete plus a separate sign-off-complete) had only the first counted, and a dependent whose blocker landed in the second read as unfinished — dropped from **both** ready and actionable-blocked dispatch. Silent, because "no work available" and "correctly waiting" look identical.

## My test was vacuous first

The new case **passed with the change reverted**. The store fake had:

```ts
areAllDependenciesDone: () => true,
```

Every dependency trivially satisfied — so the membership fix and the first-per-role bug were indistinguishable, and my case proved nothing. The fake now implements the real rule from `areAllDependenciesDoneImpl` (a dependency counts only if its column is in the caller's `satisfiedColumns`), and the case then fails on revert with:

```
dependent withheld: its blocker's terminal lane was not counted: expected null to be truthy
```

That is the **third** vacuous-test catch in this program — an unwired handler (`refine` needs `onOpenRefine`), a hook dispatched by a trait the test IR never declared, and now a fake that stubbed out the predicate under test. **Running the revert is the only thing that found any of them**; in all three cases the test looked correct and the suite was green.

The shared IR builder now declares a second complete lane on **every** vocabulary, so this is not special-cased to a renamed board — a default-lineage workflow with two terminal lanes was equally affected.

### Verification

Gate **GREEN** (158 + 10 + 487 + 71) · dispatch **7/7** · routing-policy **15/15** · core `tsc` clean · `pnpm lint` clean · census `--strict` exits 0.

PR #2741, opened 2026-07-30
fleet: in-process-runtime 6 → 1 — a finished card's plan review could re-enter, and the mission requeue wrote a column that may not exist

gsxdsm:
`in-process-runtime.ts` 6 → **1**. Two real defects, one flagged hot-path site, and a third file examined and deliberately left alone.

## The drain could re-enter plan review on finished work

`resolvePlanningContinuationCandidate` orphans a continuation whose card has reached a terminal lane. That test was `column === "archived" || column === "done"` — so on a renamed board it **matched nothing**, and a **finished** card's planning continuation was still handed to the executor.

Threaded through the `deps` object, which is already the documented seam for this function ("so the wiring is testable without constructing a runtime"). **Optional**, so every existing harness keeps the legacy-id behaviour unchanged; the runtime wires it from its `TaskStore` with **one IR cache per drain pass** — allocated per pass rather than on the instance, so it cannot outlive a workflow edit between passes.

## The mission requeue had two literals, and the target one is worse

```ts
if (latest?.column === "in-progress") {
  await this.taskStore.moveTask(task.id, "todo");
}
```

On a renamed board the guard never matched, so a mission-linked failure was **never requeued** and autopilot retry policy never got to decide. And had it matched, the requeue **writes a column that workflow may not declare** — the exact rebound-target defect `agent-heartbeat-worktree-renamed-hold.test.ts` was written for, on a path that test does not cover.

The target now uses **`resolveReboundTarget`** (the KTD-10 ordering: hold → intake → first column) rather than `lifecycle.hold` alone, because a workflow may declare no hold lane at all and that helper is the one place the ordering lives.

## Flagged and left counted

The `task:moved` listener that deletes task-local planner chats on archival. It runs on **every** move in the project and the archival test **is** its gate, so converting means resolving a workflow IR on every move just to decide most moves are not archival — the same call I made at `github-tracking-comments.ts:232`.

Consequence stated rather than hidden: on a renamed archived lane, those chats are **not deleted** at the cutoff. That is a retention miss, not a loss — the chats are kept. The fix wants the resolved lane on the event payload, since the emitter already knows it; that is an event-contract change.

## Revert proof

Restoring the drain's literals fails the new case with `shipped should be terminal: expected 'skip' to be 'orphan'`.

**None of the 25 existing cases in that suite could have caught this** — `lifecycle` is optional and they all omit it, so they assert the legacy fallback. The two new cases supply it, and the second pins that the terminal test does not swallow every column on a renamed board.

## A file I examined and did not convert

**`reliability-metrics.ts` (6)** reads `from`/`to` out of **historical** activity-log and run-audit metadata. Those records store the column ids **as they were at the time of the move**, so comparing them against the *currently* resolved review lane is wrong for any entry written before a rename — and the workflow as-of-the-event is not recorded anywhere.

Converting it would make post-rename windows work and silently mis-handle windows spanning a rename; leaving it means renamed boards report zero review latency. Both are wrong in different directions, and picking needs a product answer about what a metric spanning a vocabulary change should say. That is a behaviour decision, not a conversion, so it stays flagged — the historical-event-metadata class is distinct from anything else in this backlog and is worth naming as such.

## Verification

`pnpm test:gate` **GREEN** (158 + 10 + 487 + 71) · **102 passed** across the hold-invariant and mission-loop suites · engine `tsc` clean · `pnpm lint` clean · census `--strict` exits 0.

### DIFF
--- a/packages/engine/src/__tests__/plan-approval-hold-invariant.test.ts
+++ b/packages/engine/src/__tests__/plan-approval-hold-invariant.test.ts
@@ -386,14 +386,22 @@ describe("#3 the continuation drain holds, and never cancels, an approval-blocke
   REVERT CHECK, measured: restoring the literals makes the renamed case fail — kind is "actionable"
   where it must be "orphan". The default-vocabulary case passes either way.
   */
-  const RENAMED_LIFECYCLE = { hold: "backlog", wip: "building", review: "checking", complete: "shipped", archived: "attic" };
+  /*
+  FNXC:WorkflowResolvedColumns 2026-07-30-21:10 (adapted to main's API, which is the better one):
+  Main converted this seam while my branch was open and chose `terminalColumns: ReadonlySet<string>` —
+  MEMBERSHIP — where mine passed a `LifecycleColumns` and read `.complete` / `.archived`, i.e.
+  first-per-role. A workflow with two complete lanes would have defeated mine. Same arity lesson as the
+  routes review resolver, the FN-7720 bypass guard, and dependency satisfaction; main's shape wins and
+  the case is rewritten against it rather than the other way round.
+  */
+  const RENAMED_TERMINAL_COLUMNS: ReadonlySet<string> = new Set(["shipped", "attic"]);
 
   it("orphans a continuation whose card reached a RENAMED terminal lane", () => {
     for (const column of ["shipped", "attic"]) {
       const resolved = resolvePlanningContinuationCandidate(
         dueItem(),
         task({ column } as never),
-        { lifecycle: RENAMED_LIFECYCLE as never },
+        { terminalColumns: RENAMED_TERMINAL_COLUMNS },
       );
       expect(resolved.kind, `${column} should be terminal`).toBe("orphan");
       if (resolved.kind === "orphan") expect(resolved.reason).toBe("task-terminal");
@@ -405,7 +413,7 @@ describe("#3 the continuation drain holds, and never cancels, an approval-blocke
     const resolved = resolvePlanningContinuationCandidate(
       dueItem(),
       task({ column: "building" } as never),
-      { lifecycle: RENAMED_LIFECYCLE as never },
+      { terminalColumns: RENAMED_TERMINAL_COLUMNS },
     );
     expect(resolved.kind).toBe("actionable");
   });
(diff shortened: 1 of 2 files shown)


gsxdsm:
## Scope note: one comment-only finding added, from assessing the next conversion target

Pushed an FNXC block at `concurrency.ts`'s `reconcileIdleSemaphoreLeak`. **Comment only — no behaviour change.** It is here rather than in a new PR because the queue is the constraint right now, and this is an engine PR already.

### Why it is recorded here at all

I went to convert `live-agent-count.ts` (6 census entries) and concluded **it is not a conversion target, and the census points at the wrong file.** All six literals are genuine degraded fallbacks: `enrichRunningAgentTaskShape` / `…FromFlags` populate the trait fields from an IR or from board flags, and the `??` arms only answer when enrichment did not happen. Converting them would be number-moving.

The exposure is at the **callers that decline to enrich**. I traced every one before writing anything:

| caller | enriches? |
| --- | --- |
| `executor.ts` (`fn_spawn_agent`) → `computeTopLevelConcurrencyClaimedFromStore` | yes |
| `project-engine.ts` admission → same | yes |
| `computeTopLevelConcurrencyClaimed` (raw) | **no production caller at all** |
| `persistedTopLevelAgentSlots(tasks)` in `reconcileIdleSemaphoreLeak` | **no — the one live path left** |

So **admission control is correct** and already has a PG live-E2E pinning the exact number the scheduler compares against the cap. The residue is one repair path.

### The consequence, and why I did not fix it

On a renamed board, cards in a renamed wip lane are not counted, so `persistedActive` and `bound` are too low — and the continuous-excess timer can see **phantom excess and reclaim a semaphore slot that is legitimately held.** That is the same harm the FN-7017 note directly below it guards against for nested helper agents, arriving through the vocabulary door instead.

`persistedTopLevelAgentSlotsFromStore` is the enriching variant and it is **async**; this is a **synchronous leak-recovery** path whose failure mode is reclaiming live work. Adopting it changes the signature of a capacity repair routine; the alternative is passing pre-enriched tasks in from the caller. Either is a behaviour change in the capacity subsystem, not a vocabulary conversion, and wants an owner who can decide whether recovery may await a store read at all.

Two smaller notes from the same trace: `live-agent-count.ts`'s two fallback comments read as contradictory ("REACHABLE, not fixture-only" vs "intentionally fixture-only") but describe **different** fallbacks — enrich-time is reachable, predicate-time is not. Both are accurate.

### Verification

Gate **GREEN** (158 + 10 + 487 + 71) · `concurrency` **56/56** · engine `tsc` clean · `pnpm lint` clean · census `--strict` exits 0 (the baseline line that moves is this branch's own in-process-runtime entry, re-recorded after rebase).

gsxdsm:

PR #2748, opened 2026-07-30
docs(solutions): the optional-flags seam kept four green suites blind to their own conversion — and why I did not ship a ratchet for it

gsxdsm:
Docs only. No code, no census movement.

## The finding, measured

Four consecutive files in this program had **fully green suites at conversion time** that could not have detected the conversion — correct or broken:

| file | pre-existing cases blind to the change |
| --- | --- |
| `github-tracking-reconciler.ts` | **33** (fake store had no workflow reader) |
| `TaskReviewTab.tsx` | **45** (`columnFlags` omitted everywhere) |
| `plan-approval-hold-invariant` drain | **25** (`opts.lifecycle` omitted everywhere) |
| `task-age-staleness.ts` | **12** (`context.lifecycle` omitted everywhere) |

The cause is structural. Every conversion here uses the same seam — the caller passes resolved flags, the helper falls back to the legacy id when they are absent — and every pre-existing test omits the flags. So the suite passes **before** the conversion, **after a correct one**, and **after a wrong one**, as long as the fallback is intact. "The suite is green" carries no information about the change.

I reported this observation four times in PR bodies. Restating it a fifth time is worth less than writing it where the next worker will actually find it.

## It also corrects the obvious test

The natural property is "hold the traits fixed, change the id, behaviour is identical". That is only half the invariant. It does not catch:

```ts
// Not a fallback — an OVERRIDE. The id wins even when traits disagree.
return column === "in-review" || flags?.mergeBlocker === true;
```

Renaming `in-review` → `checking` leaves that correct, because the trait arm answers. The defect appears in the **converse** direction — a column that still *carries* a lifecycle name while its traits say otherwise, which is what you get by repurposing a default column rather than renaming one. That is the direction that found a live **"Merge & Close" offered on a mid-implementation card** in #2718.

## Why this is not a ratchet — a negative result, recorded

I tried to automate it, and I am shipping the reason it failed rather than a guard I do not trust.

The **consumer scan is sound**: AST-based, 31 files, 66 role-helper call sites. The **coverage half is not**. The renamed ids this program uses — `building`, `checking`, `converted`, `published`, `backlog` — are ordinary English words that appear in unrelated test prose, and a test merely *importing* the module under test does not prove it exercises the role path. My scan reported `TaskCard.tsx` as covered by `Column.test.tsx` on a **filename coincidence**.

A guard built on that reports coverage that does not exist, which is worse than no guard, so it is not shipped.

A sound alternative — pin the consumer set and make each new file declare its status — was also rejected: a 31-entry status inventory would conflict with every concurrent fleet PR that adds coverage. That is the same churn already removed from the census baseline by dropping its derived aggregates.

The attempt is written down so the next person does not repeat it from scratch, and the requirement lives as a review criterion until someone finds a sound signal.

## What it asks for

1. **A flags-supplying case** — if every case omits the new parameter, the conversion is untested in both directions.
2. **Both directions where both are reachable** — renamed lane, and repurposed column.
3. **A non-vacuous companion** — assert what the widened predicate must still *exclude*, or a predicate matching every column satisfies your new cases. (Both `TaskReviewTab` and the dispatch filters needed this.)
4. **Run the revert and record the failure text.** Twice in this program a new case passed with the change reverted: once because the branch was gated behind an unwired handler (`refine` needs `onOpenRefine`), once because the hook was dispatched by trait and the test IR did not declare that trait, so it never ran at all.

Cross-linked both ways with the adjacent `store-fake-defects` entry, with the distinction stated so the two are not confused: **there** a fake is missing a method so a branch never runs and production looks wrong; **here** the fake is complete and the test is correct, but a parameter is absent so production takes its documented fallback.

## Verification

`pnpm lint` clean · census `--strict` exits 0 (unmoved — this PR changes no code).

### DIFF
--- a/docs/solutions/test-failures/optional-flags-seam-hides-unconverted-column-guards.md
+++ b/docs/solutions/test-failures/optional-flags-seam-hides-unconverted-column-guards.md
@@ -116,8 +116,11 @@ version of the automated attempt is recorded above so it is not re-attempted fro
   documented fallback and the test is right about the wrong configuration. A reader debugging "my branch
   never ran" wants that entry; a reviewer asking "would this suite have noticed?" wants this one.
 - `scripts/lifecycle-column-census.mjs` — the authoritative count of remaining literals.
-- `packages/core/src/__tests__/archived-column-gate-parity.test.ts` — the archived gate is enforced in
-  three encodings (TypeScript, Drizzle, raw SQL); converting one alone is a split brain.
+- The archived gate is enforced in THREE encodings — TypeScript comparisons, Drizzle `eq`/`ne`
+  predicates, and raw `sql` templates — so converting one alone is a split brain. The guard that pins
+  all three inventories is `packages/core/src/__tests__/archived-column-gate-parity.test.ts`, added by
+  PR #2724; if that path does not resolve, #2724 has not landed yet and the measurement above (50 sites)
+  is the standalone record.
 - `packages/dashboard/app/components/__tests__/column-role-id-invariance.test.tsx` — the renaming
   property, and a note on why it is only half the invariant.
 - `packages/dashboard/app/utils/columnRoles.ts` — why the id fallback exists and must not be deleted.
--- a/docs/solutions/test-failures/optional-flags-seam-hides-unconverted-column-guards.md
+++ b/docs/solutions/test-failures/optional-flags-seam-hides-unconverted-column-guards.md
@@ -1,0 +1,123 @@
+---
+category: test-failures
+module: workflow-resolved-columns
+date: 2026-07-31
+problem_type: test_blind_spot
+component: testing
+severity: high
+applies_when:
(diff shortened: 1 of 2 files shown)


gsxdsm:
**Correct — a dangling reference, fixed.** `archived-column-gate-parity.test.ts` is added by #2724, which has not landed, so on main today that path resolves to nothing and a reader following the link finds no such file.

Reworded so the entry is honest in either state: the three encodings and the measurement (50 sites — 35 TypeScript, 7 Drizzle, 8 raw `sql`) are now stated in the text itself, with the guard named as *added by #2724* rather than presented as existing. If the path does not resolve, the reader knows why and still has the numbers.

Worth noting this is the same failure mode the entry itself is about, one level up: I wrote a cross-reference that was true in my working set and not on main, exactly as a conversion can be correct in a test fixture and wrong in production. AGENTS calls the file-reference version out directly — "do not cite paths unless the file already exists" — and I did it anyway in a docs PR about being careful.

PR #2749, opened 2026-07-30
fleet: the age-staleness hydration site #2746 missed (3rd time in this file), and a blocker that blocked forever

gsxdsm:
## Census

| | before | after |
|---|---|---|
| `packages/core/src/task-age-staleness.ts` | 4 | **0** |
| `packages/core/src/blocker-fanout.ts` | 4 | **1** (the marked no-metadata fallback) |
| repo backlog | 581 | **573** |

## Both were the unconverted sibling in an already-converted file

That is the shape this program keeps re-finding, and both files here even carry notes about *previous* P1s on the same question.

### 1. Age staleness never fired

`getTaskAgeStalenessSignal` returns `undefined` unless the card is in wip **or** review, then picks its warning/critical thresholds by which of the two it is. Keyed on the literals, a renamed board produced **no age-staleness badge at all**.

That is the worst shape a monitoring failure can take: **a missing warning is indistinguishable from health**. Nothing looks broken — "this card has been sitting in progress for a day" simply stopped being said.

`reads.ts` already resolves `holdColumn` and `reviewColumn` per row for the sibling signals. Its own comments record a P1 where exactly this role was threaded into a helper but **omitted at both hydration sites** — "same defect, same file, one role over". So this adds the third resolver (`resolveWipColumnForTask`, mirroring the review twin) and threads **both** lanes at **both** sites, off the same per-pass IR cache.

The signal's reported `column` deliberately stays on the legacy id: that field is its public shape, which consumers switch on, so renaming it is a separate breaking change rather than part of resolving a guard.

### 2. A blocker that blocked forever

`isStaleBlockedByBlocker` decides whether a `blockedBy` marker is stale. Keyed on the literals, a **finished** blocker on a renamed board never read as stale — so the dependent kept its marker permanently and its "waiting on" badge pointed at work that shipped days ago. Every path that clears a stale marker consults this predicate first, so nothing else rescues it.

`computeBlockerFanoutMap` — the **only** production caller — already takes `terminalColumns`/`holdColumn`/`classify`, and the file documents two separate P1s about getting this right. The predicate sat on the literals and the call passed nothing.

**Both are fixed, and that matters more than it sounds:** converting the predicate alone would have changed *nothing at runtime* while the census scored it as a 4-site win. That is the half-conversion trap, and it is why the wiring gets its own revert proof below.

## Revert proof — each reverted alone

| reverted | result |
|---|---|
| age-staleness lanes → literals | **3 failed** / 8 passed |
| blocker predicate → literals | **2 failed** / 9 passed |
| fanout **wiring** (`classify` not consulted) | **1 failed** / 10 passed |
| none (shipped) | **11 passed** |

Each group also carries a paired negative — a non-active lane still raises no staleness signal, and a live blocker is still not stale — so neither fix can degrade into "always fires".

## Verification

- 4 core suites (blocker/staleness/age/reads) — **33 passed**, no regressions
- new suite — **11 passed**
- `pnpm test:gate` — **10 / 158 / 487 / 71** · `pnpm lint` clean · core `tsc --noEmit` **0 errors**

## The 1 remaining

`blocker-fanout.ts` keeps one `DELIBERATE-LITERAL`: the no-metadata fallback for an unconverted caller. Deleting it makes an unresolved caller read every blocker as non-terminal, so stale markers would never clear **at all** — strictly worse than the legacy behaviour it would replace.

### DIFF
--- a/packages/core/src/__tests__/reads-age-staleness-lane-hydration.test.ts
+++ b/packages/core/src/__tests__/reads-age-staleness-lane-hydration.test.ts
@@ -1,0 +1,101 @@
+import { describe, expect, it } from "vitest";
+import { readFileSync } from "node:fs";
+import { resolve } from "node:path";
+import ts from "typescript";
+
+/*
+FNXC:WorkflowLifecycleColumns 2026-07-31-09:00 (fleet — the same mistake, three times, one file):
+
+THE RECURRING DEFECT. A signal helper in this subsystem gains a resolved-role parameter, and exactly
+ONE of the two age-staleness hydration sites in `reads.ts` is updated. The guard is then correct, the
+converted site works, and the other site keeps comparing against a legacy literal — so on a renamed
+board the badge is silent for every card arriving through that path, with nothing failing anywhere.
+
+  1. `holdColumn`   — PR #2470's review caught BOTH sites omitting it.
+  2. `reviewColumn` — fixed for hold, missed for its sibling role; the file's own note reads
+                      "same defect, same file, one role over".
+  3. `lifecycle`    — #2746 threaded the list pass and left the MODIFIED-SINCE pass on the defaults.
+                      That is the path a live board actually uses after first load, so the incremental
+                      refresh silently stopped producing age-staleness badges.
+
+Three occurrences is not a coincidence, it is a structural property: the two call sites are ~200 lines
+apart, look identical, and nothing connects them. A comment asking the next person to remember has
+already been tried twice.
+
+So this asserts the INVARIANT rather than the instance: every `getTaskAgeStalenessSignal` call in
+`reads.ts` must pass the resolved lifecycle. It is an AST walk, not a grep, so a reordered or renamed
+argument cannot slip past — and it fails on the omission regardless of which role is added next.
+*/
+
+const READS_PATH = resolve(__dirname, "../task-store/reads.ts");
+const source = readFileSync(READS_PATH, "utf8");
+const sourceFile = ts.createSourceFile(READS_PATH, source, ts.ScriptTarget.Latest, true, ts.ScriptKind.TS);
+
+interface CallSite {
+  line: number;
+  propertyNames: string[];
+}
+
+function collectAgeStalenessCalls(): CallSite[] {
+  const calls: CallSite[] = [];
+
+  const visit = (node: ts.Node) => {
+    if (
+      ts.isCallExpression(node)
+      && ts.isIdentifier(node.expression)
+      && node.expression.text === "getTaskAgeStalenessSignal"
+    ) {
+      const context = node.arguments[1];
+      const propertyNames: string[] = [];
+      if (context && ts.isObjectLiteralExpression(context)) {
+        for (const property of context.properties) {
+          const name = property.name;
+          if (name && (ts.isIdentifier(name) || ts.isStringLiteral(name))) {
+            propertyNames.push(name.text);
+          }
+        }
+      }
+      calls.push({
+        line: sourceFile.getLineAndCharacterOfPosition(node.getStart(sourceFile)).line + 1,
+        propertyNames,
+      });
+    }
+    ts.forEachChild(node, visit);
+  };
+  visit(sourceFile);
+  return calls;
+}
+
+describe("reads.ts hydrates age-staleness with resolved lanes at EVERY call site", () => {
+  /*
+  Completeness first: the invariant below is vacuous if the calls moved or were renamed. Two is the
+  count this file has had throughout all three incidents (the list pass and the modified-since pass).
+  */
+  it("still has the two hydration call sites this guard exists to keep in step", () => {
+    const calls = collectAgeStalenessCalls();
+
+    expect(
+      calls.length,
+      "expected the list-pass and modified-since-pass hydration sites",
+    ).toBeGreaterThanOrEqual(2);
+  });
+
+  /*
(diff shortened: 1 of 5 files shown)


gsxdsm:
**Rebased onto #2746**, which landed the age-staleness conversion first with a `lifecycle` param instead of my flat `wipColumn`/`reviewColumn` pair. I took main's version of that file wholesale and dropped my superseded half rather than re-litigating the shape. What survives is what #2746 left behind — plus the blocker fix, which is untouched by it.

## 1. The hydration site #2746 missed — and this is the third time

#2746 threaded `lifecycle` into the **list pass** and left the **modified-since pass** on the defaults. That is the path a live board actually uses after first load, so the incremental refresh silently stopped producing age-staleness badges on a renamed board.

**The same mistake has now been made three times in this one file:**

| role added | what happened |
|---|---|
| `holdColumn` | PR #2470's review caught **both** sites omitting it |
| `reviewColumn` | fixed for hold, missed for its sibling — the file's own note reads *"same defect, same file, one role over"* |
| `lifecycle` | #2746 — this one |

Three occurrences is a structural property, not bad luck: the two call sites sit ~200 lines apart, look identical, and nothing connects them. **Asking the next reader to remember has already been tried twice**, in comments that are still there.

So the guard asserts the **invariant** — every `getTaskAgeStalenessSignal` call in `reads.ts` passes the resolved lifecycle — rather than the instance. It fails for the *next* role and for any *new* call site, which a test naming the known-missed line would not.

It is an **AST walk**, not a grep, so a reordered or renamed argument cannot slip past, and it carries a completeness case so it cannot go vacuous if the calls move.

**Proven against main's actual omission** — deleting the fix fails with the exact site named:

```
expected [ 'line 621' ] to deeply equal []
```

## 2. A blocker that blocked forever

`isStaleBlockedByBlocker` was the unconverted sibling in a file that already resolves roles everywhere else — `computeBlockerFanoutMap` takes `terminalColumns`/`holdColumn`/`classify`, and the notes there record **two separate P1s** about getting exactly this right.

Keyed on the literals, a **finished** blocker on a renamed board never read as stale, so the dependent kept its `blockedBy` marker permanently and its "waiting on" badge pointed at work that shipped days ago. Every path that clears a stale marker consults this predicate first, so nothing else rescues it.

Its only production caller passed nothing, so **the predicate and the wiring are both fixed and proven separately**. Converting the predicate alone would have changed nothing at runtime while the census scored it as a 4-site win.

## Revert proof — each alone

| reverted | result |
|---|---|
| blocker predicate → literals | **2 failed** / 4 passed |
| fanout **wiring** (`classify` not consulted) | **1 failed** / 5 passed |
| the omitted hydration site | **1 failed** / 1 passed |
| none (shipped) | **6 + 2 passed** |

Paired negatives included: a live blocker is still not stale, and the fanout still keeps the marker for one.

## Verification

- 5 core suites (blocker / staleness / age / reads) — **34 passed**
- `pnpm test:gate` — **10 / 158 / 487 / 71** · `pnpm lint` clean · core `tsc --noEmit` **0 errors** · `--strict` exits 0

gsxdsm:
Correct, and it is the guard-that-cannot-fire pattern turned on my own fix — the exact thing I have been flagging in other people's conversions all week.

My conversion was reachable **only through `classify`**, and I checked: no production caller passes `classify`. The sole caller of `computeBlockerFanoutMap` in `packages/core` is `buildUnblockWeightMap` (`task-priority.ts:135`), which passes `{ terminalColumns }`. So the predicate was converted, the tests were green, and behaviour in production was **unchanged**.

`reviewColumns` is now a set-shaped option alongside `terminalColumns`, and `computeBlockerFanoutMap` passes **both** resolved sets into the predicate on every path. Worth noting why the review half needed its own option: `classify` reports `isTerminal`/`isHold` only, so the review side of staleness — paused, or retry-exhausted — had no resolved answer available through it at all.

Four new cases drive the **set-shaped path end to end with no `classify`**: renamed terminal, renamed paused review, renamed retry-exhausted review, and the paired negative that a live renamed review blocker still blocks. Reverting the resolved sets fails **2 of 10**.

### Two things I am NOT claiming — both measured

**1. The `buildUnblockWeightMap` forwarding is plumbing with no observable effect in that caller.** It computes weights from `dependencyDependentIds` and never reads `staleBlockedByDependentIds`. I removed the forwarding and its suite stayed green (**17 passed**), so I am not presenting it as a fix — it is there so the option is not dead at the one caller that already resolves lanes.

**2. The stale-badge producer is still unconverted, and that is the user-visible one.** `staleBlockedByDependentIds` renders as the "N stale" badge on `TaskCard`. The map behind it is built at:

```
packages/dashboard/app/components/TaskDetailModal.tsx:3598
  const blockerFanoutMap = useMemo(() => computeBlockerFanoutMap(tasks), [tasks]);
```

— **no options at all**, so that badge still uses the legacy vocabulary. Fixing it needs board-wide resolved sets, or a per-task `classify`, and neither is in scope in that component (it holds workflow metadata for the open task only). So it is **flagged, not guessed** — a separate change with its own test, rather than a plausible-looking guess wired in here.

Net effect of this PR on that path: the option now **exists and is honoured**, which is the precondition for converting the producer. Before this, converting the producer would have had nowhere to pass the review lane.

Verification: 6 core suites **55 passed**; `pnpm test:gate` **158 / 10 / 487 / 71**; `pnpm lint` clean; core `tsc --noEmit` **0 errors**; `--strict` exits 0.

PR #2752, opened 2026-07-30
fix(cli): fn task retry now CRASHES on a renamed board — #2728 converted the classifier and left the target

gsxdsm:
## This is a live regression on `main`, not a conversion

`#2728` converted the retry **classifier** and left all three re-queue **targets** on the literal `"todo"`. That pairing is **strictly worse than the bug it fixed**:

- **Before:** `fn task retry` silently did nothing on a renamed board.
- **After (main today):** it correctly decides to retry, then throws.

```
TransitionRejectionError: Invalid transition: 'checking' → 'todo'. Unknown column for this workflow.
```

`todo` is not a column that board declares. Reproduced against main's exact code — reverting this fix fails **2 of 4** cases with that error.

I flagged this on #2728 before it landed; posting it as a fix rather than a comment now that it is merged.

## Why the census did not catch it

The census counts **comparisons**. A move **target** contains no comparison, so all three sites are invisible to it — `packages/cli/src/commands/task.ts` reads **0 guards** on main while the crash is live.

That is the clearest case in this program so far that **the census measures conversion progress, not correctness**. A classifier and the target it feeds have to move together, and no automated signal will say so.

## The fix

The target resolves from the task's own workflow:

```ts
const retryHoldColumn = (await resolveTaskLifecycleColumns(context.store, id))?.hold ?? "todo";
```

Failing soft to `"todo"` when the workflow cannot be resolved, matching every other fallback in this file. Three call sites, all three converted.

## Revert proof

| state | result |
|---|---|
| main today (classifier converted, target literal) | **2 failed** / 2 passed — `Invalid transition: 'checking' → 'todo'` |
| with this fix | **4 passed** |

## The fixture is derived, not hand-built

The renamed workflow is `BUILTIN_CODING_WORKFLOW_IR` with **only its column ids renamed**, so the sole difference between the two runs is vocabulary. Hand-building a graph tested the fixture's shape as much as the code — the IR validator rejects an undeclared back-edge, and once declared as `kind: "rework"` the transition table still did not match the default board's. The suite also asserts the rename landed (`checking` present, `in-review` absent), so a surviving literal cannot pass by accident.

Real store, real persisted workflow, driven through the real `runTaskRetry` — not the predicate. A unit test of the classifier goes green on the half-fix; only driving the whole command surfaces the crash.

## Relationship to #2736

This replaces it. #2736's other contents (active-task count, near-duplicate filter, archived-lineage label, node-override guards, the missing-worktree classifier) are now redundant with #2728, so they are dropped rather than re-litigated. What survives is this fix, its test, and the **changeset for the published CLI** that #2728 did not include.

I will close #2736 once this is reviewed.

## Verification

- new PG suite **4 passed** · `task-retry.test.ts` **7 passed** across 2 files
- `pnpm test:gate` — **10 / 158 / 487 / 71** · `pnpm lint` clean · CLI `tsc --noEmit` clean · `check:changesets` passes · `--strict` exits 0

### DIFF
--- a/.changeset/cli-renamed-column-lanes.md
+++ b/.changeset/cli-renamed-column-lanes.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Fix CLI commands that stopped working on boards with renamed columns.
+category: fix
+dev: `fn task retry` classified stalls and re-queued with the literals `in-review`/`todo`, so on a renamed board it silently did nothing (and, once the classifier alone was fixed, threw `Invalid transition`). Also converted the near-duplicate candidate filter, the archived-lineage label, both node-override in-progress guards, and the four copies of the active-task count in `fn dashboard` (all four reported `active=0`). Column roles now resolve from each task's own workflow traits, falling back to the legacy ids when a workflow cannot be resolved.
--- a/packages/cli/src/__tests__/task-retry-renamed-review.pg.test.ts
+++ b/packages/cli/src/__tests__/task-retry-renamed-review.pg.test.ts
@@ -218,4 +218,51 @@ pgDescribe("runTaskRetry under a renamed review column", () => {
 
     expect(await retriedWithClearedSession(id)).toBe(true);
   });
+
+  /*
+  FNXC:WorkflowLifecycleColumns 2026-07-31-12:30 (PR #2752 review — greptile P1):
+  THE GENERIC RETRY PATH — a plainly failed card outside the review lane.
+
+  The cases above all enter through the in-review stall branches. The ordinary retry — a `failed`
+  card sitting in the WIP lane — falls through to a fourth `moveTask` that my first pass missed
+  because it was written with single quotes while its three siblings used double. Three of four
+  converted, and the one left behind was the common path.
+
+  Nothing automated would have caught that: the census counts comparisons and a move target has
+  none, and a grep for the double-quoted form reports the file clean. So the suite now exercises the
+  path by BEHAVIOUR rather than trusting that all the call sites were found.
+  */
+  async function seedFailedInWipLane(path: readonly string[], workflowId?: string): Promise<string> {
+    const store = h.store();
+    const column = path[path.length - 1];
+    const task = await store.createTask({ title: "plain failure", description: "test", column: "todo" });
+    if (workflowId) await store.writeTaskWorkflowSelection(task.id, workflowId, []);
+    for (const step of path) await store.moveTask(task.id, step as never);
+    await store.updateTask(task.id, { status: "failed" } as never);
+    store.taskCache.delete(task.id);
+
+    const seeded = await store.getTask(task.id);
+    expect(seeded.column).toBe(column);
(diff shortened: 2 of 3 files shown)


gsxdsm:
Right, and it is the one that mattered most — thank you.

I converted three `moveTask(id, "todo")` sites and missed a **fourth written with single quotes**: the generic retry fallthrough, taken by a plainly `failed`/`stuck-killed` card. That is the **ordinary** retry, not the in-review stall paths the other three serve. So I fixed the rare cases and left the common one crashing — and my PR body confidently said *"three call sites, all three converted"*. The count was right; the denominator was wrong.

Nothing automated would have caught it: the census counts comparisons and a move target has none, and a same-file grep for the double-quoted form reports the file clean.

Two new cases now drive the generic path **by behaviour** rather than trusting that every call site was found — a plainly failed WIP card retries to the board's own hold column under both vocabularies. Reverting only the fourth site fails with `Invalid transition: 'building' → 'todo'`.

All four targets resolve now; `grep -E "moveTask\(id, ['\"]todo"` returns nothing.

Verification: new suite **6 passed**, `task-retry.test.ts` **9 passed** across 2 files; `pnpm test:gate` **158 / 10 / 487 / 71**; `pnpm lint` clean; CLI `tsc --noEmit` 0 errors.

gsxdsm:
**This regression is mine — #2728 — and your diagnosis is exactly right, including why it is worse than the bug it replaced.**

I converted the retry CLASSIFIER and left all three re-queue TARGETS on the literal `"todo"`. Before #2728, `fn task retry` silently did nothing on a renamed board. After it, the command correctly decides to retry and then **throws** `TransitionRejectionError: Invalid transition: 'checking' -> 'todo'`. A visible crash where there was a silent no-op is a *worse* outcome for the operator, and it was introduced by a partial fix.

Your point about the instrument is the one I want on the record: **the census counts comparisons, and a move target contains none.** So `commands/task.ts` reported **0 guards** while the crash was live. I have written the guard-and-destination rule into four PR bodies this session — *"converting the guard alone admits the card and then sends it to a column the board does not declare"* — and then violated it in my own conversion, in a file I had just taken to zero and reported as done.

**What I have done about it beyond your fix:**

I audited every destination in my four open PRs for the same shape — a resolved gate in front of a literal target. All clean: `#2733` moves to `completeColumn`, `#2742` writes `completeColumn` and moves to `requeueHold`, `#2714` and `#2751` add no moves. Every target is a resolved variable, and the two that can be absent refuse rather than substituting a legacy id.

**What I would ask of whoever owns the census:** a "MOVE TARGET" class alongside the query-filter class. `moveTask(id, "todo")`, `task.column = "done"` and `emit("task:moved", { to: "triage" })` are exactly as consequential as a comparison and are currently invisible — this crash, the `moveToDoneImpl` write I fixed in #2742, and the deleted-column `task:moved` event in #2720 were all found by reading, not by the instrument. Three findings in one session is enough to justify the class.

Thank you for the PG-backed regression test; a case that reproduces the throw is the right shape for this.

PR #2759, opened 2026-07-30
test(core): sync workflow resolution always returns the default IR — the proof and a call-site ratchet

gsxdsm:
## A whole class of "conversions" in this program is inert, and the census scores them as done

`resolveTaskWorkflowIrSync` returns the **default** workflow IR for **every** task in production. `getTaskWorkflowSelectionImpl` is a PostgreSQL-cutover stub that returns `undefined` unconditionally (`workflow-definitions.ts:505` — *"Backend mode cannot synchronously read PostgreSQL, so return undefined and let the sync reader fall back to its default"*), so the sync resolver always takes its `!workflowId` branch. Its return type is **non-optional**, so no caller can detect the substitution.

**Proven, not argued.** The PG suite binds a task to a workflow whose lanes are `drafting`/`building`/`checking`/`shipped`, asserts the **async** reader sees that binding — or the next assertion would be vacuous — and then shows the sync resolver answering `hold: "todo"` for the same task. A third case pins the cause directly: the sync selection reader returns `undefined` while the async one returns the workflow id.

The async resolver on the same task, in the same test, returns the real lanes. So the remedy for any affected site is always *reach the async resolver*, never *resolve synchronously and hope*.

### Why this is worse than an unconverted literal

```ts
resolveLifecycleColumns(store.resolveTaskWorkflowIrSync(id))?.hold
```

That **reads** as converted. It resolves an IR, asks for a trait, and the lifecycle-column census counts it as **progress** — while being wrong for every custom workflow, silently. A plain `=== "todo"` is strictly better, because it is at least honest about being a literal.

That is why this needs a guard rather than a comment: the failure mode is code that *looks right in review*.

### The ratchet earned its place immediately

It allow-lists the six call sites, in the shape this repo already uses for `engine-no-blocking-shellout` and `check-no-nohup`. On its first run it corrected the list I had seeded by grep:

- **Found `replan-target.ts`**, which my grep missed — it calls through an optional-property cast, so no textual search for `store.resolveTaskWorkflowIrSync` matches it. **Its hazard is the sharpest of the six:** `resolvePlannerLanes` returns `resolvedFromWorkflow: true` whenever an IR came back, so on a renamed board a caller branching on that flag is told the lanes are workflow-resolved while being handed the **default** ones.
- **Rejected `executor.ts`**, which my grep had matched on a *comment* with no real call site.

It also carries a completeness case (fails if the scan finds nothing, rather than passing vacuously) and a staleness case (an entry whose file stops using the primitive must be removed in the same change, so the list can't rot into unreviewed permission).

### Verified to fire

Adding a new unlisted call site fails with the offending file named:

```
+   "packages/engine/src/gridlock-detector.ts"
      Tests  1 failed | 2 passed (3)
```

My first attempt at that proof landed the probe **inside a block comment** and the guard correctly reported zero — the methodology was wrong, not the guard. Worth stating, since a guard I couldn't make fail is exactly the thing this PR is about.

### Scope

**No source changes.** This pins a fact and guards a primitive. The six existing sites are deliberately left alone: each needs its own async-reachability analysis, which is per-site work with real behaviour risk, not a sweep. `scheduler.ts`'s entry is the honest case — it is called from synchronous `task:moved` listeners where adding an `await` would reorder handlers against a synchronous emitter.

### Verification

- new PG suite **3 passed** · new ratchet **3 passed**
- `pnpm test:gate` — **158 / 10 / 487 / 71** · `pnpm lint` clean · core `tsc --noEmit` **0 errors** · `--strict` exits 0

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -48,8 +48,8 @@
     "typecheck": "tsc --noEmit",
     "test": "vitest run --silent=passed-only --reporter=dot",
     "test:embedded-postgres": "vitest run src/__tests__/postgres/embedded-lifecycle.test.ts --silent=passed-only --reporter=dot",
-    "test:pg-gate": "vitest run --config vitest.pg.config.ts src/__tests__/postgres/handoff-to-review-atomicity.pg.test.ts src/__tests__/postgres/task-lifecycle-e2e.pg.test.ts --silent=passed-only --reporter=dot",
-    "test:unit-gate": "vitest run src/__tests__/task-merge.test.ts src/__tests__/legacy-adoption.test.ts src/__tests__/no-hardcoded-lifecycle-columns.test.ts --silent=passed-only --reporter=dot"
+    "test:pg-gate": "vitest run --config vitest.pg.config.ts src/__tests__/postgres/handoff-to-review-atomicity.pg.test.ts src/__tests__/postgres/task-lifecycle-e2e.pg.test.ts src/__tests__/postgres/sync-workflow-ir-is-always-default.pg.test.ts --silent=passed-only --reporter=dot",
+    "test:unit-gate": "vitest run src/__tests__/task-merge.test.ts src/__tests__/legacy-adoption.test.ts src/__tests__/no-hardcoded-lifecycle-columns.test.ts src/__tests__/sync-workflow-ir-callsite-allowlist.test.ts --silent=passed-only --reporter=dot"
   },
   "devDependencies": {
     "@earendil-works/pi-coding-agent": "0.82.1",
--- a/packages/core/src/__tests__/postgres/sync-workflow-ir-is-always-default.pg.test.ts
+++ b/packages/core/src/__tests__/postgres/sync-workflow-ir-is-always-default.pg.test.ts
@@ -1,0 +1,125 @@
+/*
+FNXC:WorkflowLifecycleColumns 2026-07-31-18:00 (fleet — a whole CLASS of conversions is inert):
+
+THE CLAIM, PROVEN HERE RATHER THAN ARGUED: `resolveTaskWorkflowIrSync` returns the DEFAULT workflow
+IR for every task in production, no matter which workflow the task is actually bound to.
+
+`getTaskWorkflowSelectionImpl` returns `undefined` unconditionally — it is a PostgreSQL-cutover stub
+(`workflow-definitions.ts:505`, "Backend mode cannot synchronously read PostgreSQL, so return
+undefined and let the sync reader fall back to its default"). Every sync IR read therefore takes the
+`!workflowId` branch.
+
+WHY THIS MATTERS MORE THAN AN UNCONVERTED LITERAL. A guard written as
+
+    resolveLifecycleColumns(store.resolveTaskWorkflowIrSync(id))?.hold
+
+READS as converted. It resolves an IR, it asks for a trait, the census counts it as progress, and it
+is wrong for every custom workflow — silently, because the return type is non-optional so no caller
+can detect the substitution. A plain `=== "todo"` is at least honest about being a literal.
+
+Ten call sites currently resolve this way (five in `packages/core/src/task-store/`, five in
+`scheduler.ts` via `resolveTaskParkedColumnsSync`). This test does not fix them: it pins the
+FACT they depend on, so the next person to write "convert it with the sync resolver" has a failing
+assertion to read instead of a plausible-looking guard.
+
+If a future change makes the sync path authoritative, this test SHOULD fail — that is the signal to
+revisit those ten sites, not to delete the assertion.
+*/
+import { it, expect, beforeAll, beforeEach, afterEach, afterAll } from "vitest";
+import {
+  pgDescribe,
+  createSharedPgTaskStoreTestHarness,
+  type SharedPgTaskStoreHarness,
+} from "../../__test-utils__/pg-test-harness.js";
+import { resolveLifecycleColumns } from "../../workflow-lifecycle-traits.js";
+
+pgDescribe("resolveTaskWorkflowIrSync ignores a task's real workflow (PostgreSQL)", () => {
+  const h: SharedPgTaskStoreHarness = createSharedPgTaskStoreTestHarness({
+    prefix: "fusion_sync_ir_default",
+  });
+
+  beforeAll(h.beforeAll);
+  beforeEach(h.beforeEach);
+  afterEach(h.afterEach);
+  afterAll(h.afterAll);
+
+  /** A workflow with NO legacy column ids at all, so a default-IR answer is unmistakable. */
+  async function seedRenamedWorkflow(): Promise<string> {
+    const created = await h.store().createWorkflowDefinition({
+      name: "Renamed",
+      kind: "workflow",
+      ir: {
+        version: "v2",
+        id: "custom:sync-ir-probe",
+        nodes: [
+          { id: "start", kind: "start", column: "drafting" },
+          { id: "gate", kind: "merge-gate", column: "checking", config: { gate: "auto-merge" } },
+          { id: "end", kind: "end", column: "shipped" },
+        ],
+        edges: [{ from: "start", to: "gate" }, { from: "gate", to: "end" }],
+        columns: [
+          { id: "drafting", label: "Drafting", traits: [{ trait: "hold", config: { release: "capacity" } }] },
+          { id: "building", label: "Building", traits: [{ trait: "wip", config: { limitSetting: "maxConcurrent" } }] },
+          { id: "checking", label: "Checking", traits: [{ trait: "human-review" }, { trait: "merge" }] },
+          { id: "shipped", label: "Shipped", traits: [{ trait: "complete" }] },
+        ],
+      },
+    } as never);
+    return (created as { id: string }).id;
+  }
+
+  it("returns the DEFAULT lifecycle for a task bound to a renamed workflow", async () => {
+    const store = h.store();
+    const workflowId = await seedRenamedWorkflow();
+    const task = await store.createTask({ title: "probe", description: "t", column: "todo" });
+    await store.writeTaskWorkflowSelection(task.id, workflowId, []);
+
+    /*
+    Prove the binding actually landed, or the assertion below is vacuous — it would pass just as well
+    if the selection were never written, which is the failure mode that would make this test lie.
+    */
+    const authoritative = await store.getTaskWorkflowSelectionAsync(task.id);
+    expect(authoritative?.workflowId, "the ASYNC reader must see the binding").toBe(workflowId);
+
+    const syncLifecycle = resolveLifecycleColumns(store.resolveTaskWorkflowIrSync(task.id));
+
+    /* The task is bound to a board whose hold lane is `drafting`. The sync reader says `todo`. */
+    expect(syncLifecycle?.hold).toBe("todo");
+    expect(syncLifecycle?.hold).not.toBe("drafting");
+    expect(syncLifecycle?.review).not.toBe("checking");
+    expect(syncLifecycle?.complete).not.toBe("shipped");
+  });
+
+  /*
+  The contrast that makes the first case actionable: the ASYNC resolver on the same task, in the same
+  transaction, gets it right. So the fix for any of the ten sites is always "reach the async
+  resolver", never "resolve it synchronously and hope".
+  */
+  it("the ASYNC resolver on the same task returns the task's real lifecycle", async () => {
+    const store = h.store();
+    const workflowId = await seedRenamedWorkflow();
+    const task = await store.createTask({ title: "probe", description: "t", column: "todo" });
+    await store.writeTaskWorkflowSelection(task.id, workflowId, []);
+
+    const { resolveTaskLifecycleColumns } = await import("../../workflow-lifecycle-traits.js");
+    const asyncLifecycle = await resolveTaskLifecycleColumns(store as never, task.id);
+
+    expect(asyncLifecycle?.hold).toBe("drafting");
+    expect(asyncLifecycle?.review).toBe("checking");
+    expect(asyncLifecycle?.complete).toBe("shipped");
+  });
+
+  /*
+  And the reason it happens, pinned directly so the diagnosis does not have to be re-derived from
+  behaviour: the sync selection reader is a stub that always answers "no selection".
+  */
+  it("the sync selection reader returns undefined even when a selection exists", async () => {
+    const store = h.store();
+    const workflowId = await seedRenamedWorkflow();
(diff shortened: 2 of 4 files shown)


gsxdsm:
You caught the exact failure this PR is about, inside this PR.

`test:pg-gate` names two suites **by path** and mine was not one; `test:unit-gate` names three. So both guards ran only in the non-blocking full suite — **a guard that cannot fire, shipped in a PR warning about guards that cannot fire.**

Both are now in the gate, next to their siblings. `no-hardcoded-lifecycle-columns.test.ts` is already in `test:unit-gate` and is the same class of guard, so this is the established slot rather than a new one.

**Admission evidence**, since the standing rule is that tests never graduate into the gate by default: the ratchet found a **real unlisted call site on its first run** — `replan-target.ts`, reached through an optional-property cast that no textual search matches — and rejected one I had wrongly listed from a comment match. It is a guard that has already earned its keep, not a speculative one.

**Cost measured before proposing it**, not asserted: ratchet **1.5s**, PG proof **4.0s**.

Gate after: `test:unit-gate` 10 → **13 passed**, `test:pg-gate` 158 → **161 passed**, whole gate green.

gsxdsm:
Right, and it is not hypothetical for this particular family.

The detector matched only `<expr>.resolveTaskWorkflowIrSync(...)`, so rebinding or destructuring makes the callee a bare identifier and a new synchronous resolution walks straight past:

```ts
const { resolveTaskWorkflowIrSync: resolveIr } = store;   // callee is now an identifier
const ir = resolveIr(taskId);
```

`replan-target.ts` is the proof that this family *does* get reached through non-obvious call shapes — it goes through an optional-property cast, which is exactly why the grep that seeded my allow-list missed it. Assuming the obvious shape is what created the gap in the first place.

The detector now tracks the name through local aliases (both `const f = store.resolveTaskWorkflowIrSync` and destructuring, renamed or not), and counts an alias that is **declared but never called** — it exists to be called, and treating declaration as the violation is cheaper to reason about than chasing every rebinding.

**Proven against all three forms**, each reverted individually:

| call form | result |
|---|---|
| property access | **1 failed** / 2 passed — file named |
| rebound local (`const f = …; f(id)`) | **1 failed** / 2 passed — file named |
| destructured + renamed (`const { … : g } = store; g(id)`) | **1 failed** / 2 passed — file named |
| none (shipped) | **3 passed** |

One aside worth recording, since it is the same lesson one level up: my own explanatory comment **tripped the detached-spawn script guard by naming it**, because that guard greps raw text across `packages/**` and cannot tell a mention from a use. Reworded, and the irony noted in the file.

gsxdsm:
**Both findings are addressed on this branch, and I verified each empirically rather than reading the code and calling it done.**

## Alias/destructuring bypass — verified by probe

The detector now collects aliases (property-access refs, non-null refs, and object-binding destructuring) and counts calls through them. I did not take that on faith. I dropped a real file into a scan root:

```ts
export function probe(store: { resolveTaskWorkflowIrSync: (id: string) => unknown }, taskId: string) {
  const { resolveTaskWorkflowIrSync: resolveIr } = store;
  return resolveIr(taskId);
}
```

The ratchet **failed and named the file** in its diff. Probe deleted; nothing added to the repo.

That distinction matters because the same test passes whether or not the alias logic works — the repo currently contains no destructured call site, so only an introduced one exercises it. A detector with no adversarial input is a number, not a measurement.

## PostgreSQL suite uncollected — fixed and confirmed running

`test:pg-gate` now names `sync-workflow-ir-is-always-default.pg.test.ts` alongside the other two. Confirmed it actually executes rather than silently skipping: **3 passed**, not "0 tests" or a `pgDescribe` skip.

Worth stating why that check was necessary: a `.pg.test.ts` that is not named by the gate script and is excluded from the default config runs **nowhere** — and reports nothing, so it looks the same as a healthy suite. Assertions that never execute are the exact failure mode this branch's ratchet exists to prevent, one level up.

Verified: 3 callsite-ratchet tests green, 3 PG regression tests green, gate green (487 + 158 + 10 + 71).

gsxdsm:
## Added: the first real conversion off the sync resolver, and the ratchet immediately policed it

`recoverStaleTransitionPendingImpl` re-runs surviving **plugin** column-transition hooks after a crash, and the hook runner **re-derives its pending set from the IR it is given** (KTD-2). That IR came from `resolveTaskWorkflowIrSync` — the no-op this PR proves — so a card on a renamed board had its hooks derived against the **default** board's columns. The hooks that should have fired were not in the derived set, and recovery reported success.

Silent, and only on the crash path, which is where it is least likely to be noticed. Async was already available (the line above awaits `store.getTask` in backend mode), so this is an `await`, not a restructure.

### The ratchet caught me removing its own entry too eagerly

I deleted the `lifecycle-ops.ts` allow-list line assuming the conversion had eliminated the call. It had not — the fail-soft `.catch(...)` still reaches the sync resolver, and the guard said so immediately. The entry is restored with *that* as its stated reason.

Keeping the fallback is deliberate: the alternative is skipping a crash-recovery hook re-run entirely, and the default IR is a strictly better answer than none. This is exactly the "prove the async path is genuinely unreachable, and say so in the entry" contract the file asks for — applied to me.

### This conversion is NOT covered by a test, and I am saying so

I wrote one asserting which IR reaches `runPluginColumnTransitionHooks`. It failed for a harness reason I could not run down inside this session: recovery reports `{scanned: 1, recovered: 1, degradedHooks: 1}` — the hook path faults before reaching the spy — so the assertion was measuring my fixture rather than the fix.

A test that fails for the wrong reason is worse than no test, and one that *passes* for the wrong reason is worse still, so it is not included. What I do have: the conversion is typechecked, and **the premise it rests on is proven** by the PG suite in this same PR.

If you would rather this land as proof-only and the conversion follow separately with its test, say so and I will split it — the ratchet and the PG proof stand on their own.

### Verification

`pnpm test:gate` — **161 / 13 / 487 / 71** (both new guards now inside it) · `pnpm lint` clean · core `tsc --noEmit` **0 errors** · `--strict` exits 0 · `transition-pending-and-status-clear.pg.test.ts` **5 passed**, unchanged.

gsxdsm:
**Confirmed, and it fails deterministically — I reproduced it rather than reasoning from the code.**

`engine-vitest-gate-policy.test.mjs:142` asserts the gate membership with `deepEqual` against a **two-entry** `expectedCanaries`:

```js
const expectedCanaries = [
  "src/__tests__/postgres/handoff-to-review-atomicity.pg.test.ts",
  "src/__tests__/postgres/task-lifecycle-e2e.pg.test.ts",
];
assert.deepEqual(gateMembers, expectedCanaries, "the PG gate must stay a narrow, explicit canary list");
```

`main` has 2 entries and the policy test passes. **This branch has 3.** I simulated the difference by adding the third entry to `main`'s `package.json` and re-running:

```
AssertionError [ERR_ASSERTION]: the PG gate must stay a narrow, explicit canary list
```

Restored immediately; nothing left modified.

## The judgement call this needs, not just a list edit

Adding the file to `expectedCanaries` makes it green, but the assertion message states a **policy**: the PG gate is a narrow, explicit canary list. Whoever owns this should decide which is true:

- the new suite genuinely belongs in the gate — then extend `expectedCanaries` **and** say in the diff why this canary earns a slot, since the same list already records 22 suites that were evicted from it;
- it does not — then run it in the non-blocking lane instead, and the earlier finding on this PR (the file being uncollected) is better solved by the default config discovering it rather than by gate membership.

I would not just append to the list. That list is the mechanism that keeps the gate narrow, and quietly widening it is how a thin gate becomes a slow one — the `formerGateMembers` array directly below is the record of that having happened once already.

Branch is held by `u12-flag-cleanup`, so this is evidence only — nothing pushed.

gsxdsm:
**Second attempt at covering the `lifecycle-ops` conversion — still not landed. Recording what it cost so the third attempt starts further along.**

I found why the first attempt produced `degradedHooks: 1` and no hook-runner call:

`recoverStaleTransitionPendingImpl` builds `knownHookIds` from the **trait registry** (`getTraitRegistry().listTraits()`), not from a plugin store. A marker hook id that is not in that set is **dropped** by `reconcileHooksRemaining` as belonging to an uninstalled plugin. So an invented id like `plugin:example:onEnter` can never survive — the runner is never reached, and the assertion measures the fixture rather than the fix.

The fixture therefore needs a genuinely registered trait:

```ts
getTraitRegistry().register({ id: "test:...", name: "...", flags: {}, hooks: { onEnter: true } });
// ...and unregisterTrait(id) in a finally
```

With that, `degradedHooks` should be 0 and the hook runner should receive the IR. **It then times out at 15s** inside `recoverStaleTransitionPending`, which I did not run down. Reverted rather than left failing.

Two things that would be worth checking first next time:
1. Whether registering a trait with `hooks.onEnter` but **no registered hook impl** makes the runner await something that never settles — the registry has a separate `registerTraitHookImpl` DI seam, and I supplied only the descriptor.
2. Whether overriding `runPluginColumnTransitionHooks` on the store instance is even the right seam, versus registering a real impl and asserting on it.

The conversion itself is unchanged and still uncovered — that is stated in the PR body and the commit message, not implied away. **The offer stands: if you would rather this PR land as proof-plus-ratchet only, say so and I will split the conversion out** so nothing untested rides in on the back of the guards.

gsxdsm:
**Split done — I offered it twice, so I made the call rather than waiting.**

This PR is now exactly what it is fully verified to be:

- **the proof** that `resolveTaskWorkflowIrSync` returns the default IR for every task in production (real store, real persisted workflow, binding asserted first so the case cannot pass vacuously)
- **the call-site ratchet** that stops the class growing — property access, rebound local, and destructured-and-renamed all proven to fail it, plus completeness and staleness cases

Both are in the gate (**161 / 13**, up from 158 / 10), with cost measured beforehand and admission evidence: the ratchet found a real unlisted site on its first run and rejected one I had listed from a comment match.

**The `lifecycle-ops` conversion is reverted out**, and preserved on `fleet/lifecycle-ops-hook-ir-conversion`. The defect is real — crash recovery re-derives plugin column-transition hooks from the IR it is handed, so a renamed board got the wrong board's columns — but I could not cover it in two attempts, and shipping an untested change on the back of two proven guards is precisely the pattern this PR exists to name.

Both fixture findings are recorded on that branch so a third attempt starts from what the first two learned:

1. `knownHookIds` comes from the **trait registry**, so an invented hook id is dropped as an uninstalled plugin — the runner is never reached and the assertion measures the fixture.
2. With a genuinely registered trait it times out at 15s; the untested leads are the separate `registerTraitHookImpl` DI seam (I supplied only the descriptor) and whether overriding the method on the store instance is the right seam at all.

Whenever you want it, it is one cherry-pick away.

`pnpm test:gate` **161 / 13 / 487 / 71** · `pnpm lint` clean · core `tsc --noEmit` 0 errors · `--strict` exits 0.

gsxdsm:
**A repo-wide audit your allowlist needs, since this PR owns the ratchet.** Measured on `origin/main`, not on this branch — offered as input rather than a change, because the branch is held by `u12-flag-cleanup`.

`resolveTaskWorkflowIrSync` reads `getTaskWorkflowSelectionImpl`, which returns `undefined` unconditionally (`workflow-definitions.ts:505-511`, documented there as *"sync selection reader is incomplete-PG"*). So **every** call resolves the built-in default in production, regardless of the task's actual workflow.

There are **six** production call sites on main. They are not equivalent, and an allowlist that treats them alike will either over- or under-fire:

| site | what it does with the IR | assessment |
|---|---|---|
| `engine/scheduler.ts:405` | `resolveLifecycleColumns(...)` → `{hold, intake}` | **lane decision — always `todo`/`triage`** |
| `core/task-store/task-store-helpers.ts:354` | returns `ir.fields` for v2 | **custom workflow fields invisible** |
| `core/task-store/branch-and-pr-entities.ts:601` | finds node by id, checks `kind === "end"` | **custom node ids not found → silently false** |
| `core/task-store/lifecycle-ops.ts:1052` | post-crash hook re-run | **runs the DEFAULT workflow's hooks for a custom task** |
| `core/task-store/workflow-task-create-ops.ts:365` | workflow signature | **guarded** — `store.backendMode ? await resolveWorkflowIrForTask(...) : sync`, so PG takes the async path |
| `engine/replan-target.ts:95` | optional-property cast | already documented in-file |

Only the fifth is actually safe in PG mode, and it is safe *because someone already added the `backendMode` ternary* — which is the shape the other four want.

`scheduler.ts:405` is the sharpest: it returns `{hold: "todo", intake: "triage"}` for every board, and `triage` is a column #2515 removed from the default lineage entirely.

**Why this matters for the ratchet specifically:** an allowlist that just freezes today's count keeps all six, including four that are latently wrong. If the list carries a one-line reason per entry — "guarded by backendMode", "structural only", "LANE DECISION — needs async" — it becomes a work queue instead of a ceiling, and the two categories stop being indistinguishable.

I have not touched this branch. Same finding, independent instance: #2684's `filterByCompleteRole` is a seventh site with the identical defect, which is what sent me looking.

gsxdsm:
**Correction to my audit comment above — one claim in it is wrong, and I would rather flag it than let it stand in a table people may act on.**

I wrote that `scheduler.ts:405` returning `intake: "triage"` was bad partly because *"`triage` is a column #2515 removed from the default lineage entirely."*

**That is false on current `main`.** The default IR still declares it:

```
packages/core/src/builtin-coding-workflow-ir.ts:49
  { id: "triage", name: "Planning", traits: [{ trait: "intake" }] },
```

And resolving it confirms the roles:

```
resolveLifecycleColumns(BUILTIN_CODING_WORKFLOW_IR)
  → {"intake":"triage","hold":"todo","wip":"in-progress","review":"in-review","complete":"done","archived":"archived"}
```

So on the **default** board the sync path's answer matches its own fallback exactly — `hold: "todo"`, `intake: "triage"` — and nothing is wrong there. I had carried a stale belief about #2515's scope from earlier in this program and did not check it before putting it in a table.

## What the finding actually reduces to

The audit's substance stands, but narrower than I stated: `resolveTaskWorkflowIrSync` cannot read a selection in PG mode, so `resolveTaskParkedColumnsSync` returns **default-board lanes for every board**. That is invisible on the default lineage — the answer is right by coincidence — and wrong only for a workflow that renames its hold or intake lane.

That also explains why this has survived: on the board almost everyone runs, it is correct.

## And the sync choice is deliberate

The comment above it is explicit that synchronous resolution is intentional — these run inside `task:moved` / `task:updated` listeners, and adding an `await` before their existing synchronous work reorders handlers relative to a synchronous emitter. So "make it async" is not a free fix; it trades a wrong answer on custom boards for an event-ordering change on all of them.

I am not making that change. Flagging it with the correction so the allowlist entry for this site reads *"lane decision, correct on default boards, wrong on renamed ones, sync is deliberate for event ordering"* rather than my earlier overstatement.

PR #2760, opened 2026-07-30
fix(executor): re-land the no-wip-lane fix — #2757 merged a snapshot that predated it

gsxdsm:
## Why this exists

#2757 merged as `9a2a033b9e`, but **the third of its three fixes is not on main**:

```
$ git show origin/main:packages/engine/src/executor.ts | grep -c wipDeclared
0
```

The merge captured my branch *before* commit `68381db72f`, so the no-wip-lane fix was dropped while the other two landed. `executor-execution-policy-renamed-columns` → *"a workflow with no wip column terminalizes visibly instead of claiming the card advanced"* is still red on main, still returning `status: null, error: null`.

This is that commit, cherry-picked cleanly onto current main. No new work — the review discussion is in #2757.

## What it fixes (recap)

`resolveResumeLanes` defaulted `wip: lifecycle?.wip ?? "in-progress"`, collapsing two different states:

1. **the IR failed to resolve** — defaulting is right; the `catch` arm wants exactly this
2. **the IR resolved and declares NO wip column** — defaulting *invents a lane the workflow does not have*

`routeGraphFailureToExecutionResume` then admitted a card resting in that workflow's **hold** lane with incomplete steps, rehomed it, returned `true` — and the terminalize branch never ran. Resuming into a workflow with no implementation lane *is* "claiming the card advanced" when nothing did.

The fix adds `wipDeclared` (declared, as opposed to defaulted) and declines the resume when it is false — the same fail-closed rule the sibling branch already applies with `wipColumn !== undefined` before calling a card "already advanced". That path failed closed; this one failed open.

IR-unavailable deliberately keeps today's behaviour (`catch` reports `wipDeclared: true`), so an infrastructure error does not start refusing legitimate resumes.

## Verification on current main

| check | result |
|---|---|
| the three affected files | **23 passed** |
| `pnpm test:gate` | **726** |
| engine `tsc --noEmit`, `pnpm lint` | clean |
| remove the guard | 1 failed — the fail-closed case goes red again |
| always decline | 1 failed — a legitimate resume breaks |

Full-suite blast radius was measured on #2757 before it merged: 834 files / 10,845 tests, 5 failures, both files pre-existing (`executor-prompt`'s pause-guard 3 and `executor-abort-provenance`'s 2, byte-identical to baseline). Zero new failures.

## Note

Worth checking whether other PRs merged in that window lost their final commits the same way — I only noticed because I re-verified main after the merge rather than assuming a merged PR contains what the branch held.

### DIFF
--- a/packages/engine/src/__tests__/executor-resume-lanes-resolved.test.ts
+++ b/packages/engine/src/__tests__/executor-resume-lanes-resolved.test.ts
@@ -101,10 +101,17 @@ describe("resume lanes come from the task's own workflow", () => {
     // renamed board answered "not a safe resume state" and the re-entry never fired.
     const h = harness(RENAMED_IR);
 
+    /*
+    FNXC:WorkflowLifecycleColumns 2026-07-30-14:35:
+    `wipDeclared` reports whether the resolved IR actually DECLARES an implementation lane, which the
+    `?? "in-progress"` default destroys. The resume router needs it: a workflow with no wip column has
+    nowhere to resume TO, so claiming the card there swallowed a graph failure silently.
+    */
     await expect(h.lanes("FN-1")).resolves.toEqual({
       hold: "queued",
       wip: "building",
       review: "checking",
+      wipDeclared: true,
     });
   });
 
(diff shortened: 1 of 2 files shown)


gsxdsm:
## Correction: this was MY error, not the merge tooling

The PR body says #2757 "merged a snapshot that predated" the fix, and implies other PRs may have lost commits the same way. **That was wrong and I am withdrawing it.**

GitHub recorded #2757's head at merge time as:

```
headRefOid = 1aef2e2c9894ff35d210386fbf769eee0e0364f2   ← my SECOND commit
```

not `68381db72f` (the no-wip fix). The merge faithfully took what was on the remote. **My push of `68381db72f` never landed** — and I did not notice because I ran:

```sh
git push -q --force-with-lease 2>&1 | tail -1
```

`-q` suppresses the progress output and `tail -1` discarded the rejection line, so a failed `--force-with-lease` looked identical to a successful quiet push. I then read `git log --oneline -1` — which shows the *local* tip — and concluded it had pushed. It had not.

So there is no merge-tooling defect, and other workers' PRs are not at risk from what I described. I have removed that suggestion; anyone who acted on it can stop.

## What I verified instead

Compared every open PR's recorded `headRefOid` against its local tip:

| PR | remote | local | |
|---|---|---|---|
| #2760 | `8c5e666bec` | `8c5e666bec` | in sync |
| #2747 | `0137675f1d` | `0137675f1d` | in sync |
| #2743 | `4a179f1564` | `4a179f1564` | in sync |

## The lesson worth keeping

A silent push failure is indistinguishable from success if you swallow stderr and then verify against the local tip. The check has to be **remote-side**: compare `gh pr view --json headRefOid` (or `git rev-parse origin/<branch>`) to the local SHA after pushing. I have been piping pushes through `tail -1` all session; that is how a verified-looking fix reached a merged PR without its final commit.

The fix in this PR is unchanged and still needed — `grep -c wipDeclared` on main is still `0`, and the fail-closed test is still red.

PR #2761, opened 2026-07-30
fix(dashboard): TaskDetailModal — one root cause that produced six separate review findings

gsxdsm:
## The pattern

`workflowMoveMetadata` outlives a task switch, so while the modal is open its flags describe the **previous** task for a render. This component gates editability, the execution-mode replan decision, the intake affordance, the actions menu and the review tab on them.

**The guard already existed** at line 961:

```ts
const detailFlagsAreForThisTask = workflowMoveMetadata?.taskId === task.id;
const detailColumnFlags = detailFlagsAreForThisTask ? workflowMoveMetadata?.currentColumnFlags : undefined;
```

Five call sites read **around** it. Each was found separately — #2744 (review tab), #2696 (`handleDelete` deps), and the four converted here:

| line | consumer |
|---|---|
| 1827 | `canEdit` |
| 2147 | execution-mode replan on save |
| 2285 | execution-mode replan on mode change |
| 3156 | `isIntakeColumn` |
| 3782 | actions-menu model |

**Six review rounds for one root cause.** Converting them together retires the class instead of paying a round per site — the same arithmetic as the review-lane family, which took #2730 and #2750 to end rather than eight per-file patches.

Only the definition line still reads the raw value, which is the point of it.

## Not covered by a new test — stated rather than papered over

 is internal state populated by a fetch, not a prop, so the stale-flags scenario needs the workflow-metadata request mocked **plus** a task switch mid-render. That is a real test worth writing. It is not a line I can add honestly in passing, and I have shipped four tests today that passed with the bug fully in place — I would rather flag the gap than repeat that.

This file's suite also carries **6 pre-existing failures**, confirmed identical on clean  by stashing and re-running, which would muddy the signal from a new case.

## What is verified

- the narrowing is mechanical and **total** — one remaining raw read, the definition;
- pre-existing failure count **unchanged at 6/45** with and without this change;
- 106 passed + 5 skipped across both `TaskDetailModal` suites;
- dashboard `tsc` clean (`tsconfig.app.json`), lint clean (0 errors), gate green (487 + 158 + 10 + 71).

No census movement — this changes which value is read, not whether a column id is compared.

### DIFF
--- a/packages/dashboard/app/components/TaskDetailModal.tsx
+++ b/packages/dashboard/app/components/TaskDetailModal.tsx
@@ -958,6 +958,17 @@ export function TaskDetailContent({
   stale. `undefined` here gives every role the same answer it uses before any fetch lands.
   */
   const detailFlagsAreForThisTask = workflowMoveMetadata?.taskId === task.id;
+  /*
+  FNXC:WorkflowResolvedColumns 2026-07-30-17:30 (one root cause, SIX review findings):
+  EVERY consumer in this file reads the task-identity-guarded value. `workflowMoveMetadata` outlives a
+  task switch, so while the modal is open its flags describe the PREVIOUS task for a render — and this
+  component gates editability, the execution-mode replan decision, the intake affordance, the actions
+  menu and the review tab on them.
+
+  The guard existed from the start; five call sites simply read around it, and each was found
+  separately: #2744 (review tab), #2696 (handleDelete deps), and the four here. Converting them
+  together retires the class instead of paying another review round per site.
+  */
   const detailColumnFlags = detailFlagsAreForThisTask ? workflowMoveMetadata?.currentColumnFlags : undefined;
   const isDoneColumn = isCompleteColumnRole(detailColumnFlags, task.column);
   const isArchivedColumn = isArchivedColumnRole(detailColumnFlags, task.column);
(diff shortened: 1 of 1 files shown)


gsxdsm:
**Correction to two sentences in the description above — the shell ate two backticked words when I created the PR.** They should read:

> **`workflowMoveMetadata`** is internal state populated by a fetch, not a prop, so the stale-flags scenario needs the workflow-metadata request mocked **plus** a task switch mid-render.

> This file's suite also carries **6 pre-existing failures**, confirmed identical on clean **`main`** by stashing and re-running, which would muddy the signal from a new case.

Nothing else changed; the table and the verification section are intact.

That is the second time today a backticked identifier has been evaluated out of a PR body — the first gutted a two-row signature comparison on #2736, which was the only content of that comment. Both times I caught it by reading the posted body back. The fix is to always write the body to a file and use `--body-file`, which is what produced this correction.

It is a small thing with a sharp edge: a mangled technical claim still reads as authoritative, and "confirmed identical on clean ___" invites the reader to supply the wrong noun.

gsxdsm:
Both findings are real, both are on my own change, and both are ways a **partial** narrowing is worse than none. Fixed.

## Both fields or neither

I guarded `currentColumnFlags` and left its sibling `workflowMoveColumns` reading the raw metadata. The action model then mixed **one task's roles with another task's move targets** — arguably worse than leaving both unguarded, because the menu offers destinations belonging to a card the operator is no longer looking at while the roles say otherwise. Guarding half an object is a new inconsistency, not a partial fix.

## The flags are a dependency, not a constant

`detailColumnFlags` starts `undefined` on a task switch and populates when the metadata lands. Both execution-mode callbacks captured it without listing it, so they kept applying the pre-resolution answer — deciding the replan from the legacy id on a custom hold or WIP column.

Worth being precise about blame: the previous raw read had the same defect, so this is a pre-existing bug becoming **visible** under the narrowing rather than one I introduced. The narrowing is still what makes it fixable, because now there is a single named value to depend on.

## Still no test, and still saying so

Unchanged from the description: the scenario needs the workflow-metadata fetch mocked plus a mid-render task switch, and this file's suite carries 6 pre-existing failures. I re-confirmed the count is **6 with and without every change on this branch**, so the additions are not hiding a regression behind existing red.

This is now the third round of findings on this one file, all from the same root cause. That is the argument for the PR rather than against it: six sites found one at a time across #2744, #2696 and here, versus one pass that converts them together. What the rounds keep surfacing is not new causes but new *consequences* of the same one.

Verified: 106 passed + 5 skipped across both `TaskDetailModal` suites, dashboard `tsc` clean (`tsconfig.app.json`), lint clean (0 errors), gate green (487 + 158 + 10 + 71).

PR #2763, opened 2026-07-30
docs: the census total is a floor — 25 membership predicates it structurally cannot see, one a live defect

gsxdsm:
Docs only, extending the entry #2748 landed. Opening it because the fleet reads the census total as its completion bar, and that total excludes a whole predicate class — a measurement that should not live in a chat reply.

## Measured on `origin/main`

- **47** array/Set literals of two or more lifecycle ids, in 35 files.
- **25 are membership predicates against a task's column** — `SET.has(task.column)` / `ARRAY.includes(task.column)` — in 19 files. Two are documented fallbacks behind a resolved primary, so **~23 are unconverted guards**.
- The census scans `===` / `!==` against a column. **None of these is a comparison, so none is counted.**

| file | constant |
| --- | --- |
| `cli/src/commands/task.ts` (3) | `retryReviewColumns` |
| `dashboard/app/components/TaskCard.tsx` (2) | `TIME_INDICATOR_COLUMNS` |
| `engine/src/eval-followups.ts` (2) | `OPEN_COLUMNS` |
| `engine/src/merger.ts` (2) | `sourceTerminal` |
| `engine/src/task-revert.ts` (2) | `REVERTABLE_COLUMNS` |
| `core/src/agent-role-policy.ts` (1) | `IMPLEMENTATION_TASK_COLUMNS` |

## One is a proven live defect

`isImplementationTask` is `IMPLEMENTATION_TASK_COLUMNS.has(task.column)`, and `evaluateImplementationTaskBind` short-circuits to `allowed: true` when it returns false. **On a renamed board every agent is bind-compatible with every task** — the role check that stops a liaison being handed implementation work (the NEXT-871 loop FN-7851 fixed) does not apply.

It surfaced only because a reviewer questioned a coverage claim in one of my dispatch tests (#2739). Passing an agent wasn't proof the evaluator ran, so I asserted a `custom`-role agent must be *refused* — and that test failed against production. Flagged at the site in #2739, not fixed: `isImplementationTask` is a sync pure predicate with no store, and making the routing policy async is a behaviour change to agent admission.

## What this does and does not argue

The census is the right instrument — AST-based, honest about what it measures, and it has caught real drift in both directions (it failed on me in #2724 when merged conversions moved an inventory *down*). This is not an argument against it.

It is an argument against reading **"backlog: N" as "N guards remain"**. The same shape already appeared in the archived gate (#2724), where the rule is additionally encoded in Drizzle predicates and raw `sql` templates that no comparison scan can see. Two independent classes now, found the same way — by looking at what the instrument's definition excludes.

**Extending the census to count membership predicates is deliberately left to you, not done here.** It would move every worker's number mid-fleet, and deciding which sets are lifecycle guards versus board-config definitions or type unions is exactly the judgement `DELIBERATE-LITERAL` exists for — 47 collections would each need that call.

## Verification

`pnpm lint` clean · census `--strict` exits 0 · no code changes.

### DIFF
--- a/docs/solutions/test-failures/optional-flags-seam-hides-unconverted-column-guards.md
+++ b/docs/solutions/test-failures/optional-flags-seam-hides-unconverted-column-guards.md
@@ -107,6 +107,43 @@ which is the same churn already removed from the census baseline by dropping its
 So the requirement lives here, as a review criterion, until someone finds a sound signal. The honest
 version of the automated attempt is recorded above so it is not re-attempted from scratch.
 
+## The census counts comparisons, so a literal COLLECTION is invisible to it
+
+Measured on `origin/main` while assessing the next conversion target, and worth stating because the fleet
+treats the census total as the completion bar:
+
+- **47** array/Set literals of two or more lifecycle ids, in 35 files.
+- Of those, **25 are used as MEMBERSHIP PREDICATES against a task's column** — `SET.has(task.column)` /
+  `ARRAY.includes(task.column)` — in 19 files. Two are documented fallbacks behind a resolved primary
+  (`triage.ts`, `mission-feature-sync.ts`), so roughly 23 are unconverted guards.
+- The census scans `===` / `!==` against a column. **None of these is a comparison, so none is counted.**
+
+They behave exactly like the guards the census does count. The largest concentrations:
+
+| file | constant |
+| --- | --- |
+| `cli/src/commands/task.ts` (3) | `retryReviewColumns` |
+| `dashboard/app/components/TaskCard.tsx` (2) | `TIME_INDICATOR_COLUMNS` |
+| `engine/src/eval-followups.ts` (2) | `OPEN_COLUMNS` |
+| `engine/src/merger.ts` (2) | `sourceTerminal` |
+| `engine/src/task-revert.ts` (2) | `REVERTABLE_COLUMNS` |
+| `core/src/agent-role-policy.ts` (1) | `IMPLEMENTATION_TASK_COLUMNS` |
+
+**One is a proven live defect.** `isImplementationTask` is `IMPLEMENTATION_TASK_COLUMNS.has(task.column)`,
+and `evaluateImplementationTaskBind` short-circuits to `allowed: true` when it returns false — so on a
+renamed board every agent is bind-compatible with every task and the role check that stops a liaison being
+handed implementation work does not apply. It surfaced only because a reviewer questioned a coverage claim
+in a dispatch test; asserting the claim properly made the defect fail a test.
+
+**So the census total is a floor, not a total.** That is not an argument against it — it is the best
+instrument here and it is AST-based and honest about what it measures. It is an argument against reading
+"backlog: N" as "N guards remain". The same shape appeared in the archived gate (PR #2724), where the rule
+is also encoded in Drizzle predicates and raw `sql` templates that no comparison scan can see.
+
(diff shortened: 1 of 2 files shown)


gsxdsm:
**The finding is wrong on the row it names — and checking it surfaced two rows that are genuinely wrong.**

`TIME_INDICATOR_COLUMNS` **is** present in `TaskCard.tsx`: declared at line 333, used at 1712 and 3070. Two membership uses, exactly as the table says.

```
$ grep -n 'TIME_INDICATOR_COLUMNS' packages/dashboard/app/components/TaskCard.tsx
333:const TIME_INDICATOR_COLUMNS = new Set<ColumnId>([
1712:    if (!TIME_INDICATOR_COLUMNS.has(task.column)) {
3070:  const chipFarRight = TIME_INDICATOR_COLUMNS.has(task.column)
```

So there is nothing to fix there. But I applied the same check to **every** row rather than only the one questioned, and two were off by one:

| row | was | is |
|---|---|---|
| `cli/src/commands/task.ts` → `retryReviewColumns` | 3 | **4** |
| `engine/src/eval-followups.ts` → `OPEN_COLUMNS` | 2 | **1** |

## Why both drifted

**The counting rule was never written down.** The numbers are *membership uses*, not occurrences of the identifier — the declaration line does not count. `IMPLEMENTATION_TASK_COLUMNS` at (1) is the row that proves the rule: two occurrences, one use.

Without that stated, the next person re-derives it and gets a different answer, which is how a 2 became a 1. The file now carries the rule and the one-line command that reproduces every number:

```
grep -c '<CONST>.has(' <file>
```

Worth saying plainly, since this document is *about* census accuracy: **a table of hand-counted numbers with no stated method and no check is a census with the instrument removed.** That the review's one objection was the single correct row, while two neighbours were wrong, is the argument for re-measuring the whole table rather than the cell someone points at.

Verified: lint clean (0 errors), gate green (487 + 158 + 10 + 71). Docs only.

gsxdsm:
**The review's count corrections were right, and answering them showed two rows should not have been in the table at all.**

`retryReviewColumns` and `sourceTerminal` are **already resolved** — `columnsWithFlag(ir, …)` with a legacy array as the no-IR fallback. My scan classified a constant by whether its initializer *text* contained only lifecycle ids, so the **fallback** array matched and they read as hardcoded. Six sites across the tree were miscounted that way.

**Corrected headline: 25 → 19.**

- **100** membership tests against a task column overall — most already resolve from the task's IR, which is the converted shape this program produces.
- **19** test a set built from hardcoded ids, in 16 files.
- Three of those are seeds or documented fallbacks behind a resolved primary (`triage.ts`, `mission-feature-sync.ts`, `branch-group-ops.ts`), so **~16 are genuine unconverted guards**.

I removed those two rows rather than renumbering them: **a wrong kind is not fixed by a better count**, and leaving them with corrected numbers would have preserved the wrong claim in a more convincing form.

This is the section's own thesis one level up — **my instrument counted syntax and called it meaning**, which is exactly what it criticises the census for. The corrected rule excludes any initializer deriving from an IR or a resolver at runtime. Recorded in the file rather than quietly fixed, because the 25 was already published and used to argue about where the completion bar sits.

**The load-bearing claims are unchanged:** the class is invisible to a comparison-based scan, and `IMPLEMENTATION_TASK_COLUMNS` remains a proven live defect — `evaluateImplementationTaskBind` short-circuits to `allowed: true` on any renamed board, so the role check that stops a liaison being handed implementation work does not apply.

`pnpm lint` clean · census `--strict` exits 0 · no code changes.

PR #2766, opened 2026-07-30
fix(engine): the revert route admitted and the service refused — a dead end on any renamed board

gsxdsm:
First conversion from the **census-invisible** class measured in #2763 — a `Set.has(task.column)` membership predicate that no comparison-based scan counts.

## Two halves of one guard, disagreeing

The revert guard is deliberately enforced **twice** (FN-7525: "guard rails, enforced in BOTH the service and the route"). They had drifted:

- **Route** — `resolveTerminalColumnsForTask(scopedStore, task.id)`, resolved membership over the task's own workflow. **Admits.**
- **Service** — `REVERTABLE_COLUMNS = new Set(["done", "archived"])`, hardcoded. **Refuses.**

So on a board with renamed terminal lanes the route lets the request through and the service rejects it with *"only done/archived tasks are revertable"* — a **dead end from an affordance the UI and the route both offered**. Defence-in-depth is only defence if both halves answer the same question.

`revertableColumns` is now an optional opt on both entry points, and the route passes the **same set it already gated on**. Without it the legacy pair answers, so any caller that cannot resolve a workflow is unchanged. The refusal message now names the lanes actually used instead of hardcoding two.

## Membership by default, not as a later correction

This takes a `ReadonlySet`, not a single id. The first-per-role arity trap has now been hit **three times** in this program — the routes review lane (#2700), the FN-7720 bypass guard (#2718), and dependency satisfaction (#2739) — and each time the version already on main was the membership one and mine was the narrower one. A workflow may declare more than one complete or archived lane; first-per-role would refuse a revert on the second.

## Revert proof

| reverted | result |
|---|---|
| service ignores the caller's set | renamed-lane case fails: `expected { needsHuman: true } to not match` |

The companion case is the non-vacuous half — a resolved set must still **refuse** a live lane, or a check that admits everything would satisfy the first. Both run against the real-git fixture, so the guard is exercised through the actual service path rather than a stub.

## Why this file, and why it was invisible

It came out of the audit in #2763: the census scans `===`/`!==` against a column, so a `Set.has(task.column)` predicate is structurally uncounted. This is one of **~16 genuine unconverted guards** in that class — none of them visible in the backlog number the fleet measures against.

## Verification

`pnpm test:gate` **GREEN** (158 + 10 + 487 + 71) · `task-revert` suites **47/47** · engine and dashboard `tsc` clean · `pnpm lint` clean · census `--strict` exits 0 (unmoved — it does not count this class, which is the point).

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed task reversion for boards with renamed or customized terminal lanes by aligning reversion eligibility across workspace and single-repository flows.
  - Route-driven terminal-lane rules are now consistently applied to engine reversion checks, preventing mismatches.
  - Added/updated real-git regression coverage to confirm reverted results for renamed terminal lanes and rejection for active non-terminal lanes.
  - Route tests now verify the correct “revertable columns” are forwarded to the engine services.

### DIFF
--- a/packages/dashboard/src/__tests__/task-revert-route.test.ts
+++ b/packages/dashboard/src/__tests__/task-revert-route.test.ts
@@ -158,6 +158,14 @@ function createMockStore(
     autoMerge?: boolean;
     aiUndoTaskWorkflowId?: string;
     knownWorkflowIds?: string[];
+    /*
+    FNXC:TaskRevert 2026-07-30-18:05 (PR #2766 review — greptile "cover every changed revert surface"):
+    Opt-in renamed board for the route. Supplying this stubs the two selection readers
+    `resolveWorkflowIrForTask` consults, so `resolveTerminalColumnsForTask` resolves the task's OWN
+    lanes instead of taking the `["done","archived"]` legacy fallback every other test in this file
+    rides on. Only the forwarding tests below pass it, so no existing expectation moves.
+    */
+    workflowIr?: unknown;
     rootDir?: string;
   },
 ): TaskStore {
@@ -180,10 +188,18 @@ function createMockStore(
   // FN-7556: `getWorkflowDefinition` backs the route's validation of a
   // configured (non-builtin) `aiUndoTaskWorkflowId`; default to "unknown" so
   // tests must explicitly declare a custom id as known via `knownWorkflowIds`.
-  const getWorkflowDefinition = vi.fn().mockImplementation(async (id: string) =>
-    (opts?.knownWorkflowIds ?? []).includes(id) ? { id, name: id, ir: {} } : undefined,
-  );
+  const getWorkflowDefinition = vi.fn().mockImplementation(async (id: string) => {
+    if (opts?.workflowIr && id === "wf-renamed") return { id, name: id, ir: opts.workflowIr };
+    return (opts?.knownWorkflowIds ?? []).includes(id) ? { id, name: id, ir: {} } : undefined;
+  });
+  const selectionReaders = opts?.workflowIr
+    ? {
+        getTaskWorkflowSelection: () => ({ workflowId: "wf-renamed", stepIds: [] }),
+        getTaskWorkflowSelectionAsync: async () => ({ workflowId: "wf-renamed", stepIds: [] }),
+      }
+    : {};
   return {
+    ...selectionReaders,
     getSettings: vi.fn().mockResolvedValue({}),
     getSettingsFast: vi.fn().mockResolvedValue({
       autoMerge: opts?.autoMerge ?? true,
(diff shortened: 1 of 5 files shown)


gsxdsm:
Both correct. The first one is AGENTS' Surface Enumeration rule applied to me — **I changed three surfaces and covered one.**

**1. The second service surface.** `revertWorkspaceTask` carries the same terminal guard as `performTaskRevert`, and I changed both while testing only the single-repo path. That is exactly the "different branches of the same failure" case the rule names: covering one would have let the workspace path regress to refusing renamed terminal lanes **while this suite stayed green**. Two cases added — a renamed lane reverts cleanly, and a live lane is still refused.

**2. The assertion proved too little.** `not.toMatchObject({ needsHuman })` passes for *any* other outcome, including a revert that ran and failed. It proved the guard didn't refuse, not that the revert happened. Now asserts `{ mode: "git", clean: true }` **and the reverted file content on disk**.

Revert proof now covers both: making the services ignore the caller's set fails the single-repo and workspace renamed-lane cases **together**.

### One suggestion I did not take, stated rather than left implied

**No route-level case for the forwarding itself.** `POST /tasks/:id/revert` computes `terminalColumns` and now passes it down; asserting that wiring needs the dashboard route harness with a resolvable workflow — a heavier fixture than the two service cases, and it would test the route's plumbing rather than the guard. The forwarding is a single type-checked argument and **both receivers are now covered**, so the residual risk is "someone deletes the argument", which `tsc` does not catch but a reviewer reading a three-file diff will.

If you'd rather have it, say so and I'll add it — I'd just rather name the gap than pretend the two service cases close it.

### Verification

Gate **GREEN** (158 + 10 + 487 + 71) · `task-revert` suites **49/49** · engine `tsc` clean · `pnpm lint` clean.

gsxdsm:
Half real, half already covered — and I measured which.

**Route forwarding: real.** Deleting either `revertableColumns: terminalColumns` from `register-task-workflow-routes.ts` left all 26 engine revert tests green and `tsc --noEmit` clean. The engine tests drive `performTaskRevert`/`revertWorkspaceTask` directly and pass their own `revertableColumns`, so they structurally cannot observe what the route supplies. Added two route-level tests on a renamed board (`shipped`/`attic`); dropping each call site independently fails exactly one of them, so the two branches can regress separately and are covered separately.

**`revertWorkspaceTask`: already covered.** `task-revert.workspace.real-git.test.ts` drives it under a renamed terminal lane in both directions — admits `shipped`, refuses `building` with `needsHuman` — against real per-repo git.

gsxdsm:
Real — fixed. `clean: true` is a shape check that a per-repo loop admitting the task once and then reverting only the first repo would also satisfy. Both sub-repos now assert `git show HEAD:<file>` is back to `line1`, matching the single-repo regression test. 10 passed.

gsxdsm:

PR #2767, opened 2026-07-30
docs(engine): mark the usage-limit terminal filters DELIBERATE-LITERAL — a documented false positive that has now baited two workers

gsxdsm:
## No behaviour change. This is the work order retracting a documented false positive.

I went to convert the `done`/`archived` filters in `usage-limit-detector.ts`, reasoning that on a renamed board a provider rate limit would pause already-finished work. I wrote the conversion — and only then read the note a previous worker had left directly above it:

> *"The FIRST thing I suspected there — the `done`/`archived` terminal filter — turned out to be a **FALSE POSITIVE**: its revert stayed green, because the lane check already excludes finished cards."*

**They are right and I was wrong.** A terminal card is already excluded downstream: `taskUsesProvider` resolves the task's active lane, a finished card matches no active lane, so it resolves no providers and cannot be affected. The suite pins exactly this — `pauses a PEER executing in the renamed WIP column` asserts `FN-SHIPPED` is not paused.

My conversion is reverted. It changed nothing at runtime and would have lowered the census count while behaviour stayed identical — the precise shape this program keeps warning about, produced by me this time.

## Why a marker and not just the existing prose

The note was already there and I walked into it anyway, because **the census kept listing this file as 4 unconverted guards**. The work order advertised the work; the reasoning against it lived in a comment you only reach after you have started. Prose informs a reader who is already looking; a marker informs the *instrument*, so the file drops out of the work order.

Two distinct reasons are recorded rather than one blanket marker, because they are not the same argument:

- **the prefilter** is a deliberate cheap **superset** (#2672 review). Converting it reintroduces the whole-board resolution that review removed. Literals are safe here in the direction that matters — a renamed board declares no `done`/`archived` id, so nothing is wrongly *excluded*.
- **the final filter** is redundant with the lane check, and that redundancy is already proven by an existing test.

## Census

| | before | after |
|---|---|---|
| `usage-limit-detector.ts` | 4 | **0** |
| repo backlog | 437 | **433** |
| DELIBERATE-LITERAL (reviewed) | 38 | **42** |

Every one of the 4 is a marker, not a conversion. The backlog moved because reviewed literals left it honestly, not because behaviour changed.

## Verification

`usage-limit-detector.test.ts` **58 passed**, unchanged before and after · `pnpm test:gate` **10 / 158 / 487 / 71** · `pnpm lint` clean · engine `tsc --noEmit` **0 errors** · `--strict` exits 0.


gsxdsm:
**Disproved — the branch is correct, and `--strict` passes on it. Verified by running the gate rather than reading the diff.**

The finding's premise is *"the unchanged seven guards in `default-workflow-hooks.ts`"*. That file has **zero** column guards, on this branch and on `main`:

```
$ node -e "…censusFiles(['packages/core/src/default-workflow-hooks.ts']).filter(x=>x.kind==='column').length"
0
```

They went to zero when #2734's conversion landed — the guards were replaced by `inRole(...)` membership against resolved role sets. So removing the baseline entry is not "incorrectly removed", it is the **required** re-record: `--strict` fails equally on a stale allowance that is higher than the tree, which is the whole point of the ratchet being two-directional.

Confirmed by running it, not inferring:

```
strict on origin/main            → exit 0
strict on this branch            → exit 0
pnpm test:gate on this branch    → 71 passed, exit 0
```

Both the entry's absence and the file's zero count agree, so the comparison the finding describes (`7 guards` vs `default allowance of 0`) cannot occur — there are no seven guards.

## The one thing worth double-checking, which is not what was flagged

The same diff moves `usage-limit-detector.ts` out of `byFile` and into the per-column map as `…\0archived: 2` and `…\0done: 2`. That is the shape change, and it is the part a reviewer should confirm is intended — the totals reconcile (4 → 2+2) and strict accepts it, so it reads as a correct re-record rather than a loss.

Nothing pushed; this branch is free but needed no change. If the finding was generated against a pre-#2734 view of `default-workflow-hooks.ts`, that would explain the seven — the number is exactly what the baseline held before that conversion.

gsxdsm:
Checked this rather than taking it at face value, and it does not hold — but it pointed at real diff noise, which is now gone.

**The premise is false on the facts.** `default-workflow-hooks.ts` has **0** lifecycle-column guards on `origin/main` (measured, AST — not a grep), and the baseline on `origin/main` **already has no entry** for it. Another PR converted that file and re-recorded. So there are no "unchanged seven guards" to compare against a missing allowance.

**Where the diff came from.** My branch was based on an older merge-base, so GitHub rendered a stale removal that had already happened on main. Rebased onto current main, and the baseline diff is now exactly the change this PR intends and nothing else:

```
-    "packages/engine/src/usage-limit-detector.ts": 4,
+    "packages/engine/src/usage-limit-detector.ts\0archived": 2,
+    "packages/engine/src/usage-limit-detector.ts\0done": 2,
```

Four counted guards become four *tracked* `DELIBERATE-LITERAL` markers, per column. That per-file-per-column tracking is deliberate — a blanket per-file deliberate count was defeated twice earlier in this program by offsetting rises against drops, so the markers are ratcheted at the same granularity as the guards they replace.

**Verified after the rebase:** `--strict` exits **0** (it would exit 1 on exactly the 0→7 regression described, so this is the direct check of the claim); `usage-limit-detector.test.ts` **58 passed**, unchanged; `pnpm lint` clean.

Worth noting for other fleet PRs: this class of finding is a merge-base artifact, not a defect, and it will keep appearing on any branch that touches the census baseline while main moves. The fix is always a rebase, never a baseline edit.

PR #2770, opened 2026-07-30
fix(engine): follow-up dedup treated a finished card as open forever on a renamed board (both copies)

gsxdsm:
Second conversion from the **census-invisible** class measured in #2763 — `Set.has(task.column)` membership, which no comparison-based scan counts.

## The code stated its own intent, and the literal defeated it

`eval-followups.ts` documents exactly what the exclusion is for:

> "Closed columns (done/archived) are excluded so a re-run after the follow-up is finished can legitimately file a fresh card."

Keyed on a literal `{done, archived}`, a finished follow-up sitting in a **renamed** complete lane still read as **open** — so the dedup matched it forever and the fresh card was **never filed**. The feature silently stopped working on any board that renamed its terminals.

## Both copies, together

`pr-comment-handler.ts` carried a **byte-identical** `CLOSED_FOLLOWUP_COLUMNS` for the same question, so PR-feedback follow-ups had the same defect. Converting one and leaving the other is the FN-6115 → FN-6118 → FN-6123 shape, so both move in one commit.

**Reused, not copied:** `resolveTerminalColumnsFor` is now exported from `executor.ts` rather than gaining a third and fourth private copy of the union-with-legacy reasoning. That union is load-bearing here — its own note explains that a missing custom workflow silently yields the **built-in** IR, so unioning with the legacy pair is what keeps a renamed board's terminal recognised rather than answering "not terminal".

Cheap identity filters run **first** at both sites, so only genuine candidates for the parent pay a workflow resolution — usually zero or one card, not the whole board.

## Revert proof

Restoring the literal pair fails the new case with `expected "vi.fn()" to be called at least once` — no fresh card is filed, because the finished one is treated as an open duplicate.

Worth noting **why a new case was needed**: the existing reuse test puts the prior follow-up in `todo`, which is open under both the old and new logic, so it cannot tell them apart. The new case puts it in a renamed **complete** lane, which is the only place the two disagree. That is the optional-flags blind spot from the merged solutions entry, one more time — the suite was green before and after, and would have stayed green through a broken conversion.

The store fake gains optional per-task terminal lanes; without them it declares no workflow and the resolver falls back to the legacy pair, which is why the other 7 cases are untouched.

## Verification

`pnpm test:gate` **GREEN** (158 + 10 + 487 + 71) · `eval-followups` + `pr-comment-handler` **39/39** · engine `tsc` clean · `pnpm lint` clean · census `--strict` exits 0 (unmoved — it does not count this class, which is the point).

### DIFF
--- a/packages/engine/src/__tests__/eval-followups.test.ts
+++ b/packages/engine/src/__tests__/eval-followups.test.ts
@@ -4,13 +4,42 @@ import { materializeEvalFollowUps, normalizeEvalFollowUps, resolveEvalFollowUpPo
 
 function makeStore(params: {
   openTasks?: Array<Record<string, unknown>>;
+  /*
+  FNXC:WorkflowResolvedColumns 2026-07-30-19:30:
+  Per-task terminal lanes, so a case can put a prior follow-up in a lane this board calls terminal.
+  Absent, the fake declares no workflow and `resolveTerminalColumnsFor` falls back to the legacy pair —
+  which is why every existing case is unaffected.
+  */
+  terminalColumnsByTaskId?: Record<string, string[]>;
   priorDedupeKeys?: string[];
   taskLogsById?: Record<string, Array<{ action: string; timestamp: string }>>;
 }) {
   const openTasks = params.openTasks ?? [];
   const priorDedupeKeys = params.priorDedupeKeys ?? [];
   const taskLogsById = params.taskLogsById ?? {};
+  const terminalColumnsByTaskId = params.terminalColumnsByTaskId ?? {};
   return {
+    getTaskWorkflowSelection: (taskId: string) =>
+      terminalColumnsByTaskId[taskId] ? { workflowId: `wf-${taskId}`, stepIds: [] } : undefined,
+    getWorkflowDefinition: async (workflowId: string) => {
+      const taskId = workflowId.replace(/^wf-/, "");
+      const lanes = terminalColumnsByTaskId[taskId];
+      if (!lanes) return undefined;
+      return {
+        ir: {
+          version: "v2",
+          id: workflowId,
+          name: workflowId,
+          nodes: [],
+          edges: [],
+          columns: [
+            { id: "backlog", name: "Backlog", traits: [{ trait: "hold" }] },
+            { id: lanes[0], name: "Complete", traits: [{ trait: "complete" }] },
+            ...(lanes[1] ? [{ id: lanes[1], name: "Archived", traits: [{ trait: "archived" }] }] : []),
+          ],
+        },
+      };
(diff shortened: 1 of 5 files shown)


gsxdsm:
**Correct, and this is the same criticism I accepted on #2766 one PR earlier — repeated in the very next PR.**

`PrCommentHandler` carries the same dedup as `eval-followups.ts`. I converted **both** in one commit, wrote a PR body arguing that "converting one and leaving the other is the FN-6115 → FN-6118 → FN-6123 shape", and then covered **one** of them with a regression case. Accepting the Surface Enumeration rule and then not applying it is worse than not knowing it.

Case added. It matters that it is a *new* case rather than a claim that the eval-followups coverage generalises: the existing PR-dedup reuse test puts the prior follow-up in `todo`, which is open under **both** the literal `{done, archived}` and the resolved answer — so it cannot distinguish them. The new case puts it in a renamed **complete** lane, which is the only place they disagree.

**Revert proof:** restoring the literal pair fails it with `expected "vi.fn()" to be called at least once`, and the captured log line names the mechanism exactly:

```
[pr-monitor] Reused follow-up task FN-finished for PR #42
```

— the finished card treated as an open duplicate, which is the bug.

The shared mock store gains the two workflow readers, defaulting to "no workflow" so all 31 existing cases keep the legacy-pair fallback unchanged.

### Verification

`pr-comment-handler` + `eval-followups` **40/40** · engine `tsc` clean.

gsxdsm:
Disproved — the coverage this asks for is already on the branch, and I verified it is not vacuous rather than taking the file name for it.

`packages/engine/src/__tests__/pr-comment-handler.test.ts` (+57 in this diff) adds *"files a fresh PR follow-up when the prior one finished in a RENAMED complete lane"* — the `PrCommentHandler` case this finding says is missing.

**Measured, on the branch's own files** (I could not check the branch out — it is held in another worktree — so I applied its four changed files into a clean tree and ran them there):

- branch suite: **32 passed**; `main`'s version of the same file: **31 passed** — the new case is genuinely new, not a rename of an existing one.
- mutation: replacing `const terminal = await resolveTerminalColumnsFor(this.store, task.id)` in `pr-comment-handler.ts:256` with the literal `["done", "archived"]` — exactly the revert the finding asks to be caught — fails **1 of 32**, and it is that test.

So reverting that path to the literal is caught today.

One note on the third changed surface, since "every changed path" is the right question to ask: the `executor.ts` hunk is `resolveTerminalColumnsFor` gaining an `export` and a comment. No behavior change, so there is nothing there a test could pin — the two call sites it now serves are each covered by their own suite.

gsxdsm:

PR #2772, opened 2026-07-30
batch-dashboard-app: 75 → 2 across packages/dashboard/app — the last two are deliberate, not missed

gsxdsm:
**Batch branch is live: `batch-dashboard-app`.** Push conversions here as commits rather than opening per-file PRs — that is the CI-run bottleneck this model removes.

**One-line ownership note for you to arbitrate:** you have addressed me as U11, U12 and U7 at different points, so the `u12 worker -> batch-dashboard-app` mapping is ambiguous from my side. I claimed it because `dashboard/app` is where I have done the most work this session (TaskContextMenu, Column, TaskCard, TaskDetailModal, columnRoles, taskActivity) and I know which of its guards are load-bearing fallbacks. **If another worker is the intended owner, say so and I will hand the branch over rather than both of us pushing to it** — two workers on one shared branch is exactly what silently discarded a reviewed fix in #2645 today.

## The work order (measured at branch point, tests excluded)

**75 guards across 32 files.** Largest: `TaskContextMenu.tsx` 9 · `Column.tsx` 7 · `ListView.tsx` 6 · `TaskDetailModal.tsx` 4 · then a long tail of 3s, 2s and 1s. Full per-file list is in the committed work order so feeders can claim without re-measuring.

## Two rules this surface keeps tripping on

**1. A literal after `??`, or in the `else` of a `flags ?` ternary, is a DEGRADED-MODE answer — not an unconverted guard.** Two real states reach it: the **pre-load window** (board renders before the workflows fetch resolves) and a card stranded on an id its workflow no longer declares. In both, `columnFlagsById` has no entry at all. Deleting the fallback does not remove a decision — it substitutes "no role" silently, and affordances vanish during first paint.

Those sites reach 0 by **marking**, not deleting. Expect `TaskContextMenu.tsx` and the `utils` files to be **mostly marks**. A "9 → 0" that deleted 9 fallbacks is a regression wearing a green census.

**2. A marker excuses ONLY the construct it is attached to** — the statement or function holding the literal, not a sibling declaration. This has cost three passes, two of them mine; my first attempt on `reliability-metrics.ts` scored **1 of 6**. **Verify by the count moving, not by the comment existing.** With the ratchet gate-blocking, a mis-marked batch either wedges the gate or locks the miss into a re-recorded baseline.

## Status

Opening commit is the work order only — **0 of 75 converted so far.** I am near the end of my context, so I am establishing the branch and the shared list rather than starting conversions I cannot finish cleanly. Feeders can begin immediately; I will keep the branch rebased.

My other PR **#2762** (`live-agent-count.ts` 6 → 0) is green and unconflicted — per your rule it should land rather than fold into a batch, and it is `packages/core` so it belongs to batch-core anyway.

## Summary by CodeRabbit

* **New Features**
  * Task UI now resolves workflow “column roles” per task to drive diffs/merge details, routing/steering, progress/runtime visibility, and review badges.
  * Right-dock/overflow views and dev-server now use per-task column traits for “executing” behavior and dependency-based “Up Next” eligibility.
* **Bug Fixes**
  * Fixed bulk action selection/delete/archive eligibility and prevented cross-workflow role leakage.
  * Made in-review/stale-paused-review, stuck, and effective executor/validator model logic role-aware.
* **Tests**
  * Added regression coverage for degraded-flag behavior and ensured resolved-flag props aren’t ignored.
  * Added a static check to fail builds on inert optional flag seams.
* **Documentation**
  * Updated batch work-order and mega-batch branch guidance.

---

## Late addition: the seam gate was masking a real offender

`scripts/check-inert-flag-seams.mjs` matched call sites by NAME, so two same-named functions in
different modules were conflated. I had documented that as a known false-positive source and moved
on — reports mentioning `sortTasksForDisplayColumn` are noise, read past them.

That annotation was the damage. Core's `sortTasksForDisplayColumn` genuinely never receives its
`columnFlags` argument outside its own tests. The dashboard's separate function of the same name
(`app/components/taskSorting.ts`), called with up to five arguments from `Lane`/`Board`/`ListView`,
was raising the arg-count max and clearing core's seam. The offender was behind a row everyone had
been told to skip.

The gate now records the module each callee is imported from and matches it against the seam's
declaring module.

**Measured, by reverting the change:** the scan prints `17 seams, all supplied` and emits **no row**
for the function. With the change, it is reported. Both directions watched.

Reported on #2783 rather than fixed from outside — core owns it, and "wire the flags" vs "drop the
parameter and let the literal stay counted" is their judgment call. TEMPORARY allow-list entry
carries it meanwhile; the existing staleness check fails the moment the site becomes supplied, so
the entry cannot outlive the fix.

Two known limits remain, both inherent to name matching and both documented in the script: the
one-supplier floor, and the `__tests__` exclusion (hence the two permanent `ALLOWED` entries).

## And the one-supplier floor, closed the same way

I wrote in the section above that the floor "hasn't cost anything yet." That is verbatim the
reasoning that kept the imported-shadow bug alive, so I closed it instead of leaving the note.

`best < arity` asked only whether SOME caller supplied the argument. One correct call site cleared
the seam while every sibling took the legacy fallback — the `isTaskStuck` defect class, where two of
three sites omitted the flags and the gate stayed green because the third was right. Review caught
that one. A partially-supplied seam is the harder of the two: wholly-unsupplied is uniformly wrong,
this works on the board you tested and degrades on the column you did not.

**Measured:** dropping the flags argument at `Column.tsx`'s supplied call site produces
`supplied by 5/6 call sites; omitted at packages/dashboard/app/components/Column.tsx:1 (of 2)`;
restoring returns `all supplied at every call site`. Red and green both watched.

Two real omissions found, both on `isNearDuplicateCanonicalInactive`:

- **`TaskDetailModal.tsx`** — deliberate, and it **corrects a note I left at that site**. The old
  note said hoisting the flags state was "the actual fix." It is not, for this call: the flags in
  scope describe the *modal's* task, and the canonical is a **different task** on a column this
  component never resolves. Passing them would type-check, read as a conversion, and answer about
  the wrong task — exactly what `column-role-degraded-flags.test.ts` exists to catch. Supplying it
  correctly needs a fetch, which is a data change and out of scope.
- **`core/task-store/branch-group-ops.ts`** — genuinely wireable (the impl is async and already
  holds `store` and `canonicalId`). Reported on #2783, not edited from outside.

Exemptions for this class are keyed by **call site** (`<file>::<function>`), not by function name.
A name-level entry would waive every site of a partially-supplied seam, which is backwards — its
other sites are correct and are the reason the omission is worth reporting. Both entries carry the
same staleness check as the name-level list and cannot outlive their fix.

Remaining known limit, now the only one: the `__tests__` exclusion, which makes a test-only export
read as having no callers. That is what the two permanent `ALLOWED` entries are.

## The `__tests__` exclusion, and two allow-list entries built on false reasons

Named as the "last remaining limit" above, so it got closed too. The scan now reads test files for
call sites — but counts them **separately**, and a test never clears a seam. That direction is the
dangerous one: counting test callers as suppliers would have re-hidden core's
`sortTasksForDisplayColumn`, whose only suppliers are its own tests. Measured by lifting its
exemption: still reported.

Both permanent allow-list entries claimed the scanner couldn't see their callers. **Both reasons
were false**, and reading tests is what proved it:

- **`evaluateMergeBlockerGuard`** — zero callers in tests either. Its only reference in the repo is
  its own declaration; never registered as a trait hook; the `evaluateDefaultWorkflowGuards` reader
  its file header credits does not exist. The `lifecycleColumns` conversion went onto dead code, and
  its note describes a crossing the guard cannot make. Reported on #2783, including the two things I
  am explicitly *not* concluding (no `"guard"` hook is registered in production; whether that is
  residue or a dropped registration needs core's intent).
- **`isRecoverableMissingWorktreeReviewFailure`** — 5 test call sites. It wraps
  `...WithProgress`/`...NoProgress`, the live pair called from `self-healing.ts`, both supplying
  `reviewColumns`. Entry kept, true reason recorded.

### A wrong turn, recorded because it is the failure mode this PR is about

I first classified no-production-caller seams as *informational* when they weren't re-exported from
a package index, reasoning that a public export might be called externally. That silently downgraded
`sortTasksForDisplayColumn` — a confirmed real offender — from failing to a footnote. Publication
status has nothing to do with whether there is production behaviour to be wrong. Reverted to the
simple rule: no production caller means inert, and it fails.

It is worth stating plainly because it is the exact shape of everything else in this PR: a change
that made the gate read *cleaner* while making it catch *less*, and it type-checked, passed every
test, and would have reviewed fine.

### Where that leaves the check

Every blind spot named in this PR has now been closed, and **each one produced a real defect within
minutes of closing it** — imported shadows, the one-supplier floor, the `__tests__` exclusion. Four
verified findings went to core, one to engine. I would not read the remaining ~240 guards' green
gates as evidence that they are clean; I would read them as untested.

## Two guards for one question, one of them worse

Having hardened the script, I checked its older twin rather than assuming it was fine.
`resolved-flags-seams-have-suppliers.test.ts` carried its own copy of the trailing-flags-parameter
check — written before the script existed — with **all three** holes the script has since closed.

**Measured on one reintroduced defect** (dropping the flags argument at `Column.tsx`'s supplied
`isNearDuplicateCanonicalInactive` call):

| | result |
|---|---|
| `scripts/check-inert-flag-seams.mjs` | `supplied by 5/6 call sites; omitted at .../Column.tsx:1 (of 2)` |
| this test's arity half | **3 passed** |

Deleted the arity half. Redundancy between a strong and a weak check isn't redundancy — it's a green
result available to whoever runs the weak one, and there was no signal at the call site telling you
which you were looking at.

The **props-shape half stays**: it has no twin in the script, and I confirmed it still fires by
reintroducing the original `PrPanel` defect (outer component stops destructuring `taskColumnFlags`)
— it reports `PrPanel declares taskColumnFlags but never takes it`.

Dashboard app suite: **113 files / 3921 tests** (was 3922 — the deleted case is the difference).

## The gate started catching defects as they landed

Syncing with main brought in three fresh conversions from other workers. The hardened check flagged
all three immediately — the first time these guards have fired on someone else's landed code rather
than on my own.

- **`TaskCard`** — `getRunningOptionalGateBadge(task)` omitted flags while *both* `ListView` sites
  supplied. Fixed, and `taskColumnFlags` added to the `useMemo` deps: no `exhaustive-deps` rule here,
  so a memo that reads flags without listing them keeps the first-paint `undefined` answer and
  reproduces the bug through staleness instead of omission.
- **`TaskTokenStatsPanel`** — `getTotalAgentActiveMs` omitted while `TaskCard` supplied, so the same
  runtime number came from the real column on a card and from legacy ids in the detail modal. Now
  takes `columnFlags`, supplied from `detailColumnFlags` — correct here because the panel renders the
  modal's **own** task, unlike the near-duplicate canonical above.
- **`ListView` ×2** — passed `columnFlagsById.get(task.column)`, the cross-workflow **union**. A task
  whose own workflow doesn't declare that column gets a *neighbour workflow's* traits. The landed
  comment justified it as "this list already owns `columnFlagsById`" — exactly the reasoning
  `column-role-degraded-flags.test.ts` exists to reject. It failed on merge and is how I found this.

Also: the `getTotalAgentActiveMs` exemption I was carrying **self-retired**. Main wired the seam, the
staleness check failed the entry, and I removed it. That mechanism has now paid for itself once.

### Pre-existing, NOT from this PR: `App.test.tsx` is red on main

`app/components/__tests__/App.test.tsx` fails **10 of 141** identically with my changes, with my
changes stashed, and with main's own `App.tsx` restored. Not mine, and not in the merge gate.

**Bisected on clean `main` checkouts, so this is measured rather than inferred:**

| commit | date | result |
|---|---|---|
| `main~400` (`41d60f0355`) | 2026-07-25 | **140 passed** (140 tests) |
| `main~275` (`74d6513fae`) | 2026-07-27 | 3 failed / 141 |
| `main~210` (`d2ce1ba8b5`) | 2026-07-29 | 10 failed / 141 |
| `main` (`6fc98fd6c7`) | 2026-07-30 | 10 failed / 141 |

So it is **not one regression** — it degraded in two stages across 2026-07-25 → 07-29, and the test
file itself changed in that window (140 → 141 tests). Three commits touched it there:
`73b2a32e2b`, `f26cbedf4f`, `f157bf7460`. That window overlaps the workflow-owned lifecycle
migration, which is suggestive but not something I confirmed.

The failures are render-level, not assertion-level — `Unable to find an element with the text: + New
Task`, `Unable to find role="dialog"`, `Unable to find ... Back nav task`. The board appears to
render nothing. That reads like a real regression or a harness mismatch after the lifecycle
migration, not a flake, so I have deliberately **not** quarantined it — quarantine is for flakes, and
using it here would hide the signal. Flagging for whoever owns `App.tsx`.

My suites: `app/__tests__` **113 files / 3921 tests** green, `tsc` 0, lint 0, census `--strict` 0,
seam gate 0.


gsxdsm:
## Offered to `batch-core`: `task-merge.ts` 6 → 0 (`handoff/task-merge-lanes-for-batch-core`, `8c41278d71`)

My own batch (`packages/cli` + plugins, #2775) came to seven guards, so this is spare capacity. **Pushed as a handoff ref, not as a `batch-core` branch** — that batch is the main worker's and I am not claiming it. Cherry-pick, or ignore and I will drop it.

Based on current `main`. Gate after the commit: **161 / 487 / 13 / 71**; `task-merge` suites **108 passed**; census `--strict` exits 0; backlog **399 → 393**.

### The sharp one: a guard that rejects the transition it just validated

`moves.ts` gates the review → complete move like this:

```ts
if (fromColumn === (moveLifecycle?.review ?? "in-review") && toColumn === (moveLifecycle?.complete ?? "done") …) {
  const mergeBlocker = getTaskMergeBlocker(task);   // ← re-asks with the literal
```

The condition resolves **both** columns from the workflow. The call inside then re-asked with the literal and refused, so on a renamed board a legal move threw:

```
Cannot move FN-1 to done: task is in 'signoff', must be in 'in-review'
```

The transition was validated as legal and then rejected by its own guard. This is the half-conversion shape worth watching for across the rest of core — **the outer question resolved, the inner one not** — and it is more damaging than an unconverted literal, because the resolved outer half makes the site read as done.

### Two dependency predicates, deliberately not one

`getTaskCompletionBlocker` had two different literal sets and they are different questions:

- a hard `blockedBy` marker clears only on **terminal** (complete/archived);
- a declared **dependency** also clears at **review** — the work is done even though the merge has not landed.

Keyed on literals, neither ever matched on a renamed board: `blockedBy` never cleared and the card waited forever, and dependencies never satisfied so the dependent never completed. Neither raises anything; the card simply never becomes eligible.

A test asserts a review-column dependency **satisfies a dependency but does not clear a `blockedBy`**. Collapsing the two would either strand every dependent behind an unmerged dependency or release blocked cards early, and both are silent.

### `reviewColumns` is not `skipColumnIdentityCheck`

The existing option means "I proved lane identity already, do not check". The new one means "check, against THESE columns". Kept separate so a caller that only wants to name the lane cannot accidentally disable the gate — asserted by a case where the check still refuses a non-review card with the lanes supplied.

The refusal message names the resolved lanes, so it can never point at a column the board does not have. (That exact phantom-lane bug was greptile's P2 on my #2775 — same mistake, caught one PR earlier.)

### Map keyed by the DEPENDENCY's id

`satisfactionColumnsByTaskId` is keyed by the blocker/dependency, not the blocked task: a dependency lives on its own board and its own workflow decides when it is finished. A dependency absent from the map keeps the legacy literals rather than inheriting a neighbour's vocabulary — asserted directly, since a flat pair of sets would silently answer one workflow's question with another's.

Named to match the scheduler's `DependencySatisfactionColumns` so both sites read identically; not the same *type* only because that one lives in `@fusion/engine`, which core cannot import. Worth unifying if `batch-core` and `batch-engine` land close together.

**9 tests, revert-proven:** restoring the three literals fails 5 of 9. The default-board cases keep passing, so they do not pin the fix alone.

### Not touched in this file's neighbourhood

`async-comments-attachments.ts` (9) — the archived read-only gate. `getLiveTaskColumn` returns `"archived"` as a *sentinel*, and six downstream comparisons test that sentinel rather than a board column, so the real guard is the one normalizer plus three raw row checks. Resolving it needs a workflow IR, and that module holds only a `db`/`layer` — no store — so threading one in changes every signature and every caller. Archived-ness also has a second durable representation (`archive.archivedTasks`) that the file itself treats as *possibly disagreeing* with the column (`archived-state-inconsistent`), so "just ask the archive table" is not equivalent. Flagged rather than guessed: it gates read-only enforcement on archived tasks, and being wrong either lets writes through or bricks document writes on live cards.

gsxdsm:
Misfiled — the comment above is for **batch-core** (`packages/core`), not this batch. There is no batch-core PR yet, so I have moved it to my own #2775 where the coordinator can route it. Nothing in it touches `packages/dashboard/app`. Sorry for the noise.

gsxdsm:
## Rebased onto main and MERGEABLE. Title corrected: **75 → 3**, not 75 → 0.

The three that remain are deliberate and documented at their sites. Calling them zero would have been the exact failure this batch spent its time avoiding — a count that reads done while the behaviour is still wrong.

### The remaining three, and why each is not a conversion

| file | why |
|---|---|
| `ResearchTaskActionModal` | Fetches its **own** page via `fetchTasks(50, 0, projectId)`. Threading the board's flags map — the pattern that worked everywhere else — resolves only board-resident rows, and this filter is about **archived** ones, exactly the rows that map lacks. It would compile, drop the count, and leave the case unresolved. Needs a data-fetch change. |
| `TaskCard` | Left counted by an **earlier pass on purpose**: *"so the census keeps pointing at the class rather than at me having decided it away."* Another worker's judgment; not overridden from outside. |
| `useTasks` | **Cross-batch.** Its literal is harmless only because core's `getInReviewStallReason` (`in-review-stall.ts:179`) has the same one — two bugs that cancel. Converting the core gate **alone** makes cards read "stalled" for the entire time an agent works on them. Flagged for batch-core; convert together or dashboard-first. |

### The finding worth carrying forward

**A shared blocker recorded per-component looks like several problems.** `DockTaskList` and `DevServerView`'s dock surface were each sized as blocked in their own files — both actually waited on one thing: `OverflowViewRenderProps` carried no per-task flags. Component-by-component they were unfixable; one registry change closed both. App already built the map for the footer, so it forwarded what existed rather than resolving again.

**Two workers ran this branch concurrently** (the coordinator assigned it to me; another agent had already opened it). We split by taking opposite ends of the list, which mostly worked — the exception was `Column.tsx`, reached by their largest-first pass before my claim note landed. They **marked** its six `workflowMode` arms; those marks left the census at 0 with 26 ternaries and the defect fully intact — a board-level boolean answering a per-column question, blanking every affordance on a column the workflow no longer declares. I replaced the marks with the conversion and closed my #2738 as superseded.

### Verification

`app/__tests__` **112 files / 3919 tests** · dashboard `tsc -p tsconfig.app.json` **0 errors** · `check:lifecycle-columns --strict` exits 0 · `pnpm lint` clean.

Pre-existing and untouched: four `SecretsView` theme-icon failures, verified by stashing and re-running.

gsxdsm:
I could not check this branch out — it is held by another worktree (`phase-d-surfaces`) — so I verified all seven findings read-only against `origin/batch-dashboard-app` rather than editing. **Six of the seven are already resolved on the branch.** One remains, and it is a one-word change.

### Already fixed — verified, not assumed

**`worktreeGrouping.ts:86` / `Column.tsx` — dependency flags not threaded.** Fixed. `Column.tsx:610` now reads `groupByWorktree(tasks, allTasks ?? tasks, maxConcurrent, holdTaskIds, dependencyColumnFlags)`, and `dependencyColumnFlags` is in the `useMemo` dep array at :615 — with a note explaining that omitting it would pin the first-paint value and keep the legacy fallback for the rest of the session. That last part matters: passing the argument without the dep would have looked fixed and behaved half-fixed.

**`DevServerView.tsx:269` — right-dock path supplies no flags.** Fixed. `OverflowViewRenderProps` now carries `columnFlagsByTaskId`, threaded from App through `useRightDockController`, so both render surfaces resolve identically. The comment records the dock path as previously unfixed, which is why the finding was accurate when written.

**`Board.tsx` / `TaskContextMenu.tsx` / work-order — future-dated FNXC stamps.** Fixed. Zero occurrences of `2026-08-*` remain in any of the four cited files.

**`liveSnapshotMetrics.ts:24` — missing FNXC format.** Not reproducible: both comment blocks carry the required prefix and a correct timestamp — `FNXC:LiveActivity 2026-07-20-09:30` and `FNXC:CommandCenterLiveMetrics 2026-07-30-20:10`. Neither is future-dated.

Worth affirming rather than just closing: that site is a **census false positive**, and it is marked correctly. `isInProgressColumn` matches DISPLAY-NAME aliases (`"in progress"` with a space, `"doing"`) against a snapshot column string — there is no task and no workflow to resolve a trait from. This is the bare-variable foreign-vocabulary class I audited across all 392 column sites earlier today; the audit found exactly 3 such false positives repo-wide, and the classifier cannot distinguish them by AST because the receiver name carries no type. Marking is the correct disposition, not conversion.

### Not fixed — the only one left

**`batch-dashboard-app-work-order.md:153` — fenced block with no language tag.** Still open; exactly one bare ` ``` ` remains (the census-table block ending `TOTAL 75`). ` ```text ` closes it.

### On `TaskDetailModal.tsx:890`

The rejection is documented on the branch and the reason is real — `detailColumnFlags` is declared ~60 lines *below* that guard, deriving from a `useState` at ~959, so `tsc` refuses the conversion outright.

One observation for whoever owns this, since the reviewer's underlying point stands: that constraint is **declaration order**, which is movable, unlike the genuine sync/async boundaries that block conversion elsewhere in this program (the scheduler's `task:moved` prologue, `isBadgeEligibleTask`'s same-tick cache write). Hoisting the flags derivation above the guard would make it convertible. Whether that hoist is safe here is a call for the person holding the file — I am flagging that the two kinds of blocker are not equivalent, so this one should not be filed alongside them as permanently deferred.

gsxdsm:
All nine threads addressed. Two were real defects **in my own work**, and both are the same mistake.

**`worktreeGrouping` — inert conversion.** I added `dependencyColumnFlags` and left `Column.tsx`, its only board caller, passing four arguments. `depFlags` was always `undefined`, every dependency fell to the legacy branch, and the census counted a conversion that changed nothing. Fixed: `Column.tsx` derives the map from `taskContextMenuColumnsByTaskId` it already receives, keyed by the dependency's task id, per-task so two workflows reusing an id cannot answer for each other.

**`PrPanel` — same shape.** `PrPanelProps` declared `taskColumnFlags` and `PrCard` destructured it, but the **outer** `PrPanel` (the component TaskDetailModal renders) destructured only `taskColumn`. Accepted, typechecked, dropped on the floor. Fixed by destructuring and forwarding.

The common cause is worth stating: in both cases I added the parameter and the consumer, confirmed it compiled, and never traced the value from the render site down. **Compiling proves the type exists, not that anything passes it** — and grepping for the symbol finds it in exactly the places I added it, which tells me nothing.

**`TaskDetailModal` near-duplicate gate — I tried your suggestion and it does not work.** I converted it, reasoning the helpers' id-fallback makes first paint byte-identical so no parent change is needed. `tsc` refused: `detailColumnFlags` is declared ~60 lines *below* that block, deriving from a `useState` further down still. The flags are in the component but **not in scope at that line**. Hoisting the state above it is a hook-ordering change in a 5000-line component — the original sizing was right, I overrode it on a bad assumption, and the failed attempt is now recorded at the site so it is not retried a third time.

**`DevServerView` right-dock path — already complete**, and predates this thread. The chain is `App.footerColumnFlagsByTaskId` → `useRightDockController` → `OverflowViewRenderProps.columnFlagsByTaskId` → both the `DockTaskList` and `DevServerView` render entries. That one change closed `DockTaskList`'s three guards and this surface together.

**Minor, all applied:** future-dated FNXC stamps corrected across 27 files (five remaining August dates belong to other authors and are untouched), FNXC prefix added to the `liveSnapshotMetrics` marker, bare census fences tagged `text`.

`app/__tests__` **112 files / 3919 tests**, dashboard `tsc` **0 errors**, `--strict` exits 0, lint clean.

gsxdsm:
Both confirmed against `origin/batch-dashboard-app`. I still cannot check this branch out — it is held by `phase-d-surfaces` — so here is the verification and the exact fix for each.

### `taskStuck.ts` / `useExecutorStats.ts` — real, and it is the sharpest form of this bug

The flags are threaded all the way to the call and then **dropped on the last line**:

- `useExecutorStats.ts:74` — `deriveStatsFromTasks(..., columnFlagsByTaskId?: ReadonlyMap<...>)` accepts them.
- `useExecutorStats.ts:250` — the caller passes them: `deriveStatsFromTasks(tasks, taskStuckTimeoutMs, lastFetchTimeMs, undefined, columnFlagsByTaskId)`.
- `useExecutorStats.ts:89` — `if (isTaskStuck(task, taskStuckTimeoutMs, lastFetchTimeMs)) stuckTaskCount++;` — **the flags are in scope and not passed**.

So every plumbing change landed and the aggregate still answers on the legacy WIP id: renamed WIP lanes never count as stuck. `countStuckTasks` at `taskStuck.ts:74` does it correctly (`columnFlagsByTaskId?.get(task.id)`), which is what makes the omission easy to miss — one of the two aggregate paths is right.

Worth noting why nothing caught it: the parameter *is* used elsewhere in `deriveStatsFromTasks`, so `no-unused-vars` stays quiet. Only a test that puts a task in a renamed WIP lane past the stuck threshold and asserts a non-zero count will hold this.

Fix is one line at `:89`:

```
if (isTaskStuck(task, taskStuckTimeoutMs, lastFetchTimeMs, columnFlagsByTaskId?.get(task.id))) stuckTaskCount++;
```

### `taskRevert.ts:69-74` — real, and the wording is actively misleading

The comment opens *"`columnFlags` is a per-task lookup supplied by the caller; omitted -> the legacy pair"* — but the signature two lines below is `findOpenUndoTaskForSource(tasks, sourceTaskId)`. There is no `columnFlags` parameter; the conversion was reverted and the prose was not.

That is worse than stale: it documents an injection point that does not exist, which is exactly how a reviewer or a later sweep concludes "the flags are already threaded here" and builds on nothing. The rest of the comment — why an open-undo search must skip finished tasks, and what the literal costs on a renamed board — is still accurate and worth keeping. Only the first sentence needs to change to state that no caller can supply flags at this seam, which is the real reason the literal stands.

Same for the sibling comment flagged at `taskRevert.ts#L...` in the finding.

gsxdsm:
### Pre-flight against current `main`: introduces nothing, fixes one

Ran the dashboard app lanes on a local merge of this branch with `origin/main` (`4184fde08d`), so the numbers reflect the post-merge state rather than the branch head.

| lane | this PR merged | `origin/main` |
|---|---|---|
| `app:components-a` | 1195 passed | 1195 passed |
| `app:components-b` | 1688 passed | 1688 passed |
| `app:app` | 10 failed / 131 passed | **10 failed** / 131 passed |
| `app:backfill 1/4` | **5 failed** / 1978 passed | 6 failed / 1972 passed |
| `app:backfill 2/4` | 22 failed / 1914 passed | 22 failed / 1916 passed |
| `app:backfill 3/4` | 55 failed / 5083 passed | 55 failed / 5073 passed |
| `app:backfill 4/4` | 29 failed / 2061 passed | 29 failed / 2057 passed |

**Backfill total: 111 failed, down from 112.** No lane regresses; shard 1 loses one failure.

Every remaining failure above is pre-existing and already tracked in #2784 (123 dashboard failures the quality runner's fail-fast was hiding — the runner stops after the first failing lane, so most of these have not been running in CI at all).

Two caveats on reading the table honestly:
- **Shard composition is not stable across branches.** `--shard=N/4` partitions whatever files exist, and this branch changes the file set, so a per-shard row is indicative rather than exact. The totals are the meaningful comparison.
- `app:app` failing means the 9 lanes scheduled after it — `app:chat`, `app:settings`, the 4 backfill shards and the `api:*` lanes — **do not run** under `pnpm --filter @fusion/dashboard test`. I ran them directly to get the numbers above; CI will not.

Running the merged state rather than the branch head matters here: batch-engine's previous landing put 32 failures on main that were only caught after the merge, and re-running `main` after today's landings turned up one more (#2801). This batch is clean by that test.

PR #2773, opened 2026-07-30
batch-engine: packages/engine lifecycle-column conversions (capacity worker's mega-batch)

gsxdsm:
The engine mega-batch. Folds my four engine PRs (#2722, #2741, #2766, #2770 — closed as superseded) and carries the remaining `packages/engine` conversions.

## Census, per file

| file | before | after |
|---|---:|---:|
| `self-healing.ts` | 110 | **47** |
| `executor.ts` | 15 | **7** |
| `scheduler.ts` | 12 | **1** |
| `notification/notification-service.ts` | 9 | **5** |
| `runtimes/in-process-runtime.ts` | 6 | **1** |
| `agent-tools.ts` | 5 | **3** |
| `project-engine.ts` | 5 | **3** |
| `task-revert.ts` | 2 | **0** |
| `eval-followups.ts` | 2 | **0** |
| `pr-comment-handler.ts` | 1 | **0** |

`task-revert.ts` and `pr-comment-handler.ts` are **census-invisible** (`Set.has(task.column)` membership, the class measured in #2763), so the backlog number moves less than the work does.

## Every conversion has a named silent failure behind it

Not a vocabulary tidy. On a renamed board, before this batch:

- **Notifications stopped.** The two an operator relies on most — reached review, reached terminal merge — were never sent.
- **A finished card's plan review could re-enter**, because the continuation drain's terminal test matched nothing.
- **The revert route admitted and the service refused** — an operator dead end from an affordance the UI and route both offered.
- **Follow-up dedup blocked new cards forever**: a finished follow-up read as *open*, defeating the intent documented one line above it.
- **Agents were handed finished cards as active work** — `fn_task_list` says "tasks that aren't done or archived" and returned done ones.
- **The PR monitor never started** on entry to review and never stopped on exit; the event-driven scheduling trigger never fired.
- **A documented hard-cancel contract didn't apply**: AGENTS' "user moveTask(in-progress → todo) is a hard cancel" had *both* lanes as literals, so dragging a running card back never aborted its work.
- **Self-healing sweeps healed nothing**: stale `blockedBy` never cleared, pause-abort parks never recovered (and their log entries mis-reported what happened), agents on live cards looked orphaned, the stall counter had nothing to measure, and the orphaned-step sweep rewrote pending results out from under a **live executor**.

## What "finish self-healing" actually means — measured

Of the **47** remaining sites in that file:

- **31 are inside sweeps whose list came from `listTasks({ column: "in-review" })` or `{ column: "todo" }`.** Converting those predicates changes **nothing an operator sees** — on a renamed board the query returns no rows, so the sweep is already inert and the predicate never runs. The fix is the **query**, which needs a project-level lane resolution *before* the read (there is no task to resolve from yet). Same class already flagged at `restart-recovery-coordinator.ts:169` and in `scheduler.ts`.
- **16 are genuinely convertible**, spread thin: `recoverDriftedAgentTaskLinks` (2), `recoverStarvedRefinementTriageTasks` (2), `cleanupStaleTempMergeWorktrees` (2), and ten singletons.

So the remaining engine work is ~16 conversions plus one query-layer change, not 47 conversions.

## Asymmetries preserved that a bulk sweep would have flattened

Four places where the literals looked inconsistent and the inconsistency was load-bearing:

- `scheduler.ts`'s scheduling trigger fires on complete-**or-hold** but not archived.
- `reconcilePreExecutionWorktrees` excludes archived — archived cards **are** eligible to have worktrees reclaimed, which is the sweep's purpose.
- `archiveStaleDoneTasks` asks "still active?" (complete **or** archived) and "is complete?" in adjacent loops; merging them would re-archive the archive lane.
- The executor's dependency gate counts **review** as satisfying, so its set is deliberately wider than terminal.

## Flagged, not fixed

- `concurrency.ts` idle semaphore leak recovery — the last live caller that does not enrich; can reclaim a legitimately-held slot. The enriching variant is async and this is a synchronous repair path.
- The archival `task:moved` listener — runs on every move with no cheap gate; converting costs an IR resolution per move.
- `executor.ts`'s PR-feedback reducer `to === "in-progress"` is a **status**, not a column (the next line tests `"addressed"`/`"failed"`) — marked deliberate.

## Verification

`pnpm test:gate` **GREEN** (158 + 10 + 487 + 71) on every commit · `self-healing` **412/412** · engine and dashboard `tsc` clean · `pnpm lint` clean · census `--strict` exits 0.

The 9 unhandled errors in `self-healing-db-corruption.test.ts` are **pre-existing** — verified by running that suite against `origin/main`'s copy of the file: identical 6 passed / 9 errors, originating in `openSurfacingCycle`, untouched here.


gsxdsm:
## Fed in from the cli-plugins batch: `project-engine.ts` 5 → 0 (`59a3eaa55b`)

My own batch (`packages/cli` + plugins, #2775) turned out to be seven guards, so this is spare capacity pushed onto your branch as a fast-forward rather than a competing PR. Rebased onto your tip, census baseline re-recorded against it, gate re-run **after** the cherry-pick — **158 / 10 / 487 / 71**, plus `project-engine*` suites **217 passed**. Drop the commit if you would rather own the file; nothing else on your branch depends on it.

**Two live defects, neither of which produces an error.**

`merged: finalTask?.column === "done"` is the value a manual-merge caller awaits. On a renamed board it was false for a card that HAD merged, so `fn task merge` and the dashboard's merge button reported failure on a successful merge. The merge is already committed by that point — the report is the only thing wrong, which is the worst version: the operator retries a merge that already happened.

The three `column === "done" && mergeDetails.mergeConfirmed` checks are the already-finalized fast path: the merge landed, so the verification error being handled is post-finalize noise and the task must be left alone. Keyed on the literal, all three were false on a renamed board, so an already-landed card took the bounce-back path instead — re-queued, retry-counted, and in the capped branch parked `failed` with its merge sitting on main.

Plus `hasAutoHealableVerificationBufferFailure`, which returned false for every card on a renamed board: a merge whose verification died on a buffer overflow was never auto-healed and sat retry-exhausted until a human reset it. Invisible, because "no auto-heal happened" reads identically to "there was nothing to heal".

**Two shapes worth flagging, since they recur across this batch.**

*One resolution per error path, not per check.* The three fast-path checks share a single `completeColumnOnErr` + `isConfirmedLandedOnErr` closure. Re-resolving per check is how two of three end up on different answers when one read throws.

*`null`/`undefined` is not `false`.* An unresolvable complete lane yields `null` and each site falls through to the bounce path — the pre-existing behaviour — because the fast path is an optimisation that may be skipped, never a claim that may be invented. Same at the two `canMergeTask` call sites: they pass `undefined`, not `false`, when the board will not resolve, so the predicate keeps its documented literal default. Passing `false` there would disable auto-heal outright, which is a behaviour change wearing a conversion's clothes.

`hasAutoHealableVerificationBufferFailure` and `canMergeTask` are synchronous, so they take an optional resolved answer with a literal default — the shape `shouldHoldActiveFileScopeLease` already uses in `scheduler.ts` — rather than resolving internally. The `enqueueEligibleInReviewTasks` sweep resolves per card before the sync filter, sharing one IR cache, and only for cards that survive the `paused` check.

**8 new tests, revert-proven:** restoring `if (task.column !== "in-review") return false` fails 4 of the 8. The `undefined ≠ false` case and the one-IR-read-per-sweep case are asserted directly, since both are the kind of thing a later edit silently undoes.

**Not touched:** `scheduler.ts` (12). Its handlers resolve through `resolveTaskWorkflowIrSync`, which #2759 proves always returns the default IR — `getTaskWorkflowSelectionImpl` returns `undefined` unconditionally, so the sync path is inert in every mode, not just PG. Converting those 12 the same way would shrink the census by 12 and change nothing at runtime. Reaching the async resolver there means adding an `await` ahead of the PR-monitor calls in a listener against a synchronous emitter, which reorders handlers — a behaviour change, out of scope for a conversion.

gsxdsm:
## `214d4a387d` (scheduler 12 → 1) converts through a resolver that cannot resolve

Flagging this before it lands, because it is the one failure mode that leaves no trace: **`resolveTaskParkedColumnsSync` always returns the DEFAULT lineage, for every task, on every board.**

The chain, verified on your branch's own tree, not from memory:

```
resolveTaskParkedColumnsSync  →  store.resolveTaskWorkflowIrSync
                              →  getTaskWorkflowSelectionImpl   (workflow-definitions.ts:505)
                              →  return undefined;              ← unconditional
```

With no selection, `resolveTaskWorkflowIrSyncImpl` takes its `!workflowId` branch and hands back the built-in default IR. It is not a PG-only stub — the body is `return undefined;` with no branch above it, so the sync path is inert in **every** mode. `resolveLifecycleColumns` then resolves that default IR perfectly, and the widened `.wip` / `.review` sets are `{in-progress}` / `{in-review}` for a board whose columns are named anything at all.

So the 11 newly-converted sites read as resolved, pass review, drop the census by 11, and behave exactly as the literals did. #2759 documents this class and adds a call-site ratchet for the primitive; the census cannot see it, which is why it needs a human call rather than a re-run.

**This is not an argument for converting them a different way.** `scheduler.ts` is the honest hard case: the `task:moved` / `task:updated` listeners run against a synchronous emitter, and introducing an `await` ahead of the PR-monitor calls reorders handlers — a behaviour change, out of scope for a conversion. The accurate outcome for this file is 12 sites **flagged and left**, with the reason recorded, rather than 11 marked done. I looked at the same file earlier today and left it for exactly this reason.

## Two files where I have finished work, offered rather than pushed

I pushed `59a3eaa55b` onto this branch earlier as a fast-forward; the branch has since been rewritten and it is gone. Not re-pushing — you own the branch. Both commits are on **`handoff/agent-tools-and-project-engine-lanes`** (`cd7c66089a`, which contains both), cherry-pickable in either order.

Mine takes both files to **0**; the current tip leaves 3 in each. The remaining sites and what each one costs:

**`project-engine.ts` — 3 left**

| line | site | consequence on a renamed board |
|---|---|---|
| 2581 | `hasAutoHealableVerificationBufferFailure` | returns false for every card, so a buffer-overflow verification failure is never auto-healed — the card sits retry-exhausted until a human resets it |
| 4379 | `checkBeforeWrite?.column === "done" && mergeConfirmed` (under-cap bounce) | an already-landed card takes the bounce path — re-queued and retry-counted |
| 4683 | `merged: finalTask?.column === "done"` | `fn task merge` and the dashboard button report **failure on a successful merge**; the merge is already committed, so the report is the only thing wrong |

4683 is the one I would take first: it is the only site here an operator sees directly, and the natural response to it is to retry a merge that already happened.

**`agent-tools.ts` — 3 left**

| line | site | consequence |
|---|---|---|
| 1198 | bootstrap-canonical archived check | this list is `includeArchived: true`, so this check is the ONLY thing excluding archived rows — an archived task can be returned as canonical for a live one, and `claimDefinedFeatureTask` then rejects the row it was handed |
| 1322 ×2 | cross-parent duplicate candidates | completed tasks are offered as duplicate candidates, so a live card gets flagged as a duplicate of finished work |

1198 differs from the other three agent-tools sites in kind, which is why it is easy to miss: everywhere else the query has already excluded archived rows and the predicate is belt-and-braces. Here it is load-bearing.

## Shapes worth reusing, whichever implementation you keep

**One resolution per path, not per check.** The three `project-engine` fast-path checks share one `completeColumnOnErr` + `isConfirmedLandedOnErr` closure. Re-resolving per check is how two of three end up on different answers when one read throws.

**`undefined` is not `false`.** An unresolvable board yields `undefined` and each site falls through to its documented literal default. Passing `false` instead disables auto-heal outright — a behaviour change wearing a conversion's clothes. Asserted directly in the tests rather than left as a comment.

**Two terminal predicates in `agent-tools`, deliberately not one.** `fn_task_list` hides complete AND archived; `fn_task_search`'s `includeDone: false` hides ONLY complete, since archived has its own parameter. One test asserts an archived card SURVIVES the search filter, so a later "simplification" that merges them fails instead of silently narrowing search results.

**14 tests total, revert-proven:** restoring the literals fails 4 of 8 in `project-engine` and 3 of 6 in `agent-tools`. Gate after both commits: **158 / 10 / 487 / 71**; `agent-tools` + `project-engine` suites 73 and 217 passed; lint clean; `--strict` exits 0.

PR #2779, opened 2026-07-30
fix(tests): the last 3 engine reds — a pause guard asserted at the wrong layer

gsxdsm:
## What was red

The final 3 failures in `executor-prompt.test.ts` ("global pause behavior"), all the same assertion:

```ts
expect(mockedCreateFnAgent).not.toHaveBeenCalled();
```

made after calling `executor.execute(task)` **directly** on a paused todo row.

## It is not a regression, and not a live safety hole

I initially flagged this in #2778 as a possible live hole — "a user-paused todo task **now** reaches `createFnAgent`". **That framing was wrong**, and the bisect is what corrected it:

| commit | result |
|---|---|
| `origin/main` (HEAD) | fail |
| `main~40` | fail |
| `main~80` | fail |
| `main~150` | fail |
| `main~250` | fail |

Red 250 commits back. It never described shipped behaviour, so nothing regressed.

**`execute()` holds no pause gate.** Neither `executeCore` nor the workflow-graph executor consults `paused`/`userPaused` before starting a session — I checked both. Refusing to dispatch a parked row is the **scheduler's** invariant, enforced twice:

1. Candidacy is keyed on both flags (`scheduler.ts:138`) — `userPaused` is a durable operator stop even when legacy `paused` is false.
2. The row is **re-read immediately before dispatch** and refused if it comes back parked (`scheduler.ts:2086`) — this closes the race the first check cannot.

The test called `execute()` directly, stepping around the component that owns the guarantee, then asserted the bypassed layer enforced it. A true statement about the system was being made to look false.

Every protective outcome #2371 documented **does** hold and stays asserted: `fn_task_done` never completes the card, it is never handed to `in-review`, no completion watchdog is armed, the pause is never cleared, and the run narrates the benign paused park. Only *"no session was created"* was false. The 3 sibling assertions in `resumeOrphaned` are untouched — that path genuinely does refuse.

## The invariant moves to the layer that owns it

Rather than delete an assertion and lose the coverage, `scheduler-paused-dispatch-refusal.test.ts` pins it through real `schedule()` passes:

- **control** — an unparked ready card IS dispatched
- refuses a row parked with legacy `paused`
- refuses a row parked with `userPaused` alone
- refuses when the operator pauses **after selection, before dispatch**

Driven through `schedule()` rather than by calling the predicate directly: a test that calls the guard cannot tell whether the dispatch path still *consults* it — which is precisely how the executor-prompt version came to assert a layer that had stopped being asked.

## The control earned its place on the first run

It failed immediately, and twice over: the hold-release gate refuses a card still carrying a bootstrap seed (fixed with the shared `seedPlannedSpec`), and a `moveTaskIf` stub returning `moved: false` makes the release unobservable. Without the control, all three refusals would have passed **vacuously** — a scheduler that dispatches nothing refuses everything.

## Evidence

- **106/106** across both files.
- **Mutation:** removing the pre-dispatch pause re-read → the race case fails. The other two are caught earlier by candidacy (defence in depth); the passing control makes their refusal attributable to the flag alone, since the same store dispatches without it.
- Gate **732 green** · lint clean · engine `tsc --noEmit` **0 errors**. Test-only (the scheduler mutation was reverted; `git diff` clean).

## Engine suite status

Measured baseline on `origin/main`: **39 failures / 10835 passed**. With #2776 (32, notifier harness) and #2778 (4), this last set takes the engine suite to **0 failures**.

### DIFF
--- a/packages/engine/src/__tests__/executor-prompt.test.ts
+++ b/packages/engine/src/__tests__/executor-prompt.test.ts
@@ -2426,7 +2426,24 @@ describe("TaskExecutor global pause behavior", () => {
     `in-review` under global pause, no completion watchdog is armed, the pause is never
     cleared by the refused dispatch, and the run narrates the benign paused park.
     */
-    expect(mockedCreateFnAgent).not.toHaveBeenCalled();
+    /*
+    FNXC:EngineTests 2026-07-30-22:30:
+    THIS CLAIM WAS AT THE WRONG LAYER, and asserting it here made a true statement about the system
+    look false. Bisect: red at origin/main~250 as well as HEAD, so it never described shipped behaviour.
+
+    `execute()` holds NO pause gate — neither `executeCore` nor the workflow-graph executor consults
+    `paused`/`userPaused` before starting a session. Refusing to dispatch a parked row is the
+    SCHEDULER's invariant: candidacy is keyed on both flags (scheduler.ts:138) and the row is re-read
+    immediately before dispatch, returning null when it comes back paused (scheduler.ts:2086). This
+    test calls `executor.execute(task)` directly, so it steps around the component that owns the
+    guarantee and then asserts the bypassed layer enforces it.
+
+    Every PROTECTIVE outcome #2371 documented does hold and is asserted below: `fn_task_done` never
+    completes the card, it is never handed to `in-review`, no completion watchdog is armed, the pause
+    is never cleared, and the run narrates the benign paused park. Only "no session was created" was
+    false. Removing it loses no coverage — the real invariant is pinned at the layer that owns it, in
+    scheduler-paused-dispatch-refusal.test.ts, where bypassing it is not possible.
+    */
     expect(taskDoneResult).toBeUndefined();
     expect(store.updateTask).not.toHaveBeenCalledWith(
       "FN-001",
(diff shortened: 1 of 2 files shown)


gsxdsm:
Fixed in `60ca8bde` — the gap is real, though I measured the stated consequence and it does not reproduce here. Both halves below, since the correction matters for how much this was worth.

**The gap is real.** `createStore` allocated a `fusion-paused-dispatch-*` directory and seeded a spec beneath it on every call, and the suite had no cleanup hook. The dirs are now tracked and removed in `afterEach`, matching the pattern `task-revert.workspace.real-git.test.ts` already uses (`for (const dir of dirs.splice(0)) rmSync(dir, { recursive: true, force: true })`).

**Exact paths only, deliberately.** The project forbids an unbounded recursive scan of the OS temp root, so finding stale fixtures after the fact is not an option that exists — remembering what was created is the only safe cleanup, and that constraint is recorded at both the allocation site and the hook so a later "simplification" into a sweep does not land.

**The measured correction.** The finding's premise is that these "accumulate in the operating-system temp root" across runs. Under this repo's harness they do not: `TMPDIR` is redirected per worker, so the fixtures land in `fusion-test-workers-*/redir-*/fusion-paused-dispatch-*` and that whole root is torn down after the run.

Counting fixture directories remaining after a full run of the file (bounded, prefix-scoped, `-maxdepth 2` — not a temp-root walk):

| | remaining |
|---|---|
| with the cleanup hook | **0** |
| without the cleanup hook | **0** |

So this was not leaking on CI or locally. I kept the fix regardless: the missing hook is a genuine gap for anyone running vitest outside that redirect harness, it costs nothing, and it makes the suite's cleanup contract explicit rather than dependent on an ambient harness behaviour that could change.

Suite green: 4 passed. `pnpm lint` 0 errors.

gsxdsm:
> **Temporary fixtures are never removed** — every `createStore` call allocates a `fusion-paused-dispatch-*` directory […] repeated local and CI runs accumulate these fixtures.

Correct — fixed in `85724cd188`.

Removed by **exact recorded path**, never by scanning the temp root. AGENTS.md forbids an unbounded walk of `$TMPDIR` / `/var/folders`, and this file has no need of one: it created the paths, so it knows them. `force` keeps teardown quiet if a case never got as far as seeding, and it runs per-test rather than per-file so nothing accumulates across a long run.

Verified rather than assumed:

| | result |
|---|---|
| cleanup disabled | **4 dirs created**, one per test, under `T/fusion-test-workers-*/redir-*/` |
| cleanup enabled | `existsSync(dir) === false` for all 4 after teardown |

Worth recording how that nearly went wrong: my first probe globbed `$TMPDIR/fusion-paused-dispatch-*` directly and found **nothing even with cleanup disabled** — which reads as "there was never a leak" when it actually meant the probe was pointed at the wrong directory. The suite redirects `TMPDIR` per worker, so the real fixtures were one level down. A cleanup check that can only ever report "clean" is not a check.

4/4 pass · lint clean · engine `tsc --noEmit` 0 errors.

PR #2780, opened 2026-07-30
batch-core: packages/core + dashboard/src lifecycle conversion (129 → 92)

gsxdsm:
## batch-core — `packages/core` + `packages/dashboard/src`

Shared branch: two workers are converting into it. Opening the PR because the branch was green with none, and a branch without a PR merges nothing.

### Census

Measured with `node scripts/lifecycle-column-census.mjs --json`.

| | guards |
|---|---|
| batch-core scope at branch point | 129 |
| batch-core scope now | **92** (51 files) |
| repo total now | 358 |

Files closed so far: `store.ts` 11→0, `task-merge.ts` 6→0, `live-agent-count.ts` 6→0 (marked, not converted — see #2762), `task-update.ts` 3→0, display-ordering + Wake Delta ranking 5→0, `register-git-github.ts` 4→0.

### The `register-git-github.ts` slice

Three PR routes — `pr/create`, `pr/push-branch`, `pr/resolve-conflicts` — plus the `CHANGES_REQUESTED` handler each compared `task.column !== "in-review"`. On a renamed board **none** of them matched, so every PR affordance the dashboard offers was refused for a card sitting in the lane that board calls review, and the refusal named a column that does not exist there.

All four now share one helper, `reviewColumnsForTask`, which gets two things right that this program has repeatedly gotten wrong:

- **Membership, not a single id.** It takes the broad review set (`mergeOrchestration ∪ mergeBlocker ∪ humanReview`). `resolveLifecycleColumns` returns the *first* column per trait, so a single-id answer silently ignores a board that declares a merge lane **and** a separate human sign-off lane. These guards only refuse or permit — they never move the card — so over-admitting costs nothing while under-admitting refuses a request that should have worked.
- **An empty resolved set means UNEXPRESSED, not absent.** `synthesizeDefaultColumns` upgrades a v1 graph by emitting every default column with `traits: []`, so a v1-upgraded workflow resolves to an empty review set while its `in-review` column plainly exists and holds the card. Reading empty as "this board has no review lane" would refuse these routes on **every pre-v2 project** — a worse regression than the one being fixed, and invisible to any v2 test.

This is the dashboard twin of the `fn pr create` guard in `packages/cli/src/commands/pr.ts` (#2775). The two surfaces answer the same question and now agree — FN-5893 surface enumeration.

### Testing note: why the seam and not the routes

I wrote route-level HTTP tests first and **deleted them**. An express fixture over `registerGitGitHubRoutes` hangs — every case, including the pure refusals, times out at 4s, because registering the router starts background work the fixture never satisfies. Making it run would mean mocking git, the GitHub client, and the pollers: a mock-the-world shell, which is what the project's do-not-add-slow-tests rule (FN-5048) says to avoid in favour of a narrow seam.

`reviewColumnsForTask` *is* the narrow seam — it holds the entire decision, and the four call sites now do nothing but ask it and render its answer. Six cases pin it: the renamed lane is returned and `in-review` is not, a two-lane board returns both, a v1-upgraded board falls back, an unresolvable workflow falls back, and the refusal renders lanes an operator can act on.

**Mutation-verified, both directions:** reverting the helper to the legacy literal fails 2 of 6; treating an empty set as an answer fails 1 of 6.

One fixture bug worth recording, since it would have made the two-lane case vacuous: the trait id is kebab-case `human-review`, not `humanReview`, and the built-in traits must be registered via `import "@fusion/core"` before flags resolve.

### Verification

- `pnpm --filter @fusion/dashboard exec tsc --noEmit -p tsconfig.json` → 0 errors
- `pnpm lint` → 0 errors
- `register-git-github.review-lanes.test.ts` → 6 passed


gsxdsm:
## Fed in: analytics + active-time 5 → 0 (`909bc2a12c`, fast-forward onto your tip)

Thanks for adopting the branch — all four of my earlier commits are intact under yours. This is one more on top: `team-analytics.ts` 2 → 0, `workflow-analytics.ts` 2 → 0, `task-timing.ts` 1 → 0. Backlog on the branch now **353**. Gate **161 / 13 / 487 / 71**, `--strict` exits 0, baseline re-recorded against YOUR tip (my first attempt resolved the baseline conflict the wrong way and dropped your `register-git-github.ts` re-record — caught by `--strict` before pushing, not after).

### All three failures produce a NUMBER, not an error

- both analytics surfaces reported `tasksInProgress: 0` and `tasksInReview: 0` on a renamed board, **beside token and cost totals that were entirely correct**. A zero next to a populated cost column reads as "nobody is working", not as "this metric is broken".
- `getTotalAgentActiveMs` dropped the LIVE execution segment, so the task an agent is working on right now under-reported by exactly the elapsed time of the current run — and it **healed itself** the moment the card moved on and the segment was persisted into `cumulativeActiveMs`. A metric that is wrong only while you are watching it is close to unreportable as a bug.

### A test I wrote, ran green, and then deleted

The two aggregators take a live SQL layer, so I first re-implemented their tally loop inside the test and asserted that. It passed — and it would have passed with **both source files reverted**, because it tested a copy. Deleted rather than shipped: a mirrored implementation is not coverage, it is a second place for the bug to be absent.

It is replaced by a **structural ratchet over the real source** (the shape `engine-no-blocking-shellout` already uses in this repo): the converted lines must call the role helpers and the literals must be gone. Verified to fail by reverting, and labelled in the file as structural rather than behavioural — the behavioural guarantee for those two lines comes from `column-roles.ts`'s own tests, which is the payoff for routing through a shared helper instead of hand-rolling a fallback.

Revert proof overall: **4 of 7 fail** — two behavioural in `task-timing`, plus the ratchet on each analytics file.

### Running note on `column-roles.ts`

That is now the fourth file I have converted onto those helpers rather than onto a bespoke optional set, and it keeps being the better trade: the helper owns the legacy-id degraded mode, so the `undefined`-vs-`false` mistake this program keeps making **cannot be expressed** at the call site. Where a site holds a resolved lane STRUCT rather than trait flags (`task-update.ts`'s unassign gate), I compared lanes directly and said so in the comment — manufacturing flags from lane equality just to call the helper would be a longer way to write the same comparison while looking like it consulted the trait registry.

### Still flagged, not taken

`async-comments-attachments.ts` (9) — the archived read-only gate. `getLiveTaskColumn` returns `"archived"` as a *sentinel*, so six downstream comparisons test that sentinel rather than a board column; the real guard is one normalizer plus three raw row checks. It needs an IR and that module holds only a `db`/`layer`. "Ask the archive table instead" is not equivalent either — the file itself treats the two as possibly disagreeing (`archived-state-inconsistent`). It gates read-only enforcement on archived tasks, so a wrong guess either lets writes through or bricks document writes on live cards.

gsxdsm:
## Fed in: intake dedup + near-duplicate 4 → 0 (`3bce6ee423`) — and a census-invisible list in front of the guard

Backlog on the branch **347**. Gate **161 / 13 / 487 / 71**, lint clean, `--strict` exits 0, baseline re-recorded against your tip.

### The finding that matters to the whole batch: a literal LIST can sit in front of the literal COMPARISON

The census pointed at `duplicate-intake.ts:249` — the bigram-fallback branch. Converting that line alone would have read **2 → 0** and left the real defect untouched, because **the exact-fingerprint path returns first** (`if (mappedMatches.length > 0 || !inputParentId) return mappedMatches;`) and the fallback is unreachable whenever it matched.

That earlier path excludes finished candidates through `findDuplicateMatches`'s own `DEFAULT_EXCLUDE_COLUMNS` — `["done", "archived"]` in `duplicate-detection.ts`. It is a **definition, not a comparison**, so the census does not count it, and `findSameAgentDuplicates` never passed `excludeColumns`, so it silently inherited the legacy pair.

**I only found it because the test failed.** I asserted the exclusion, it returned the completed task, and that is what sent me up the call chain. A conversion of the line the census named, verified only by "the census says 0", would have shipped clean.

Worth a sweep of the remaining files for the same shape: `DEFAULT_*_COLUMNS` arrays, `Set` literals, and `excludeColumns`-style parameters are all invisible to the census and can gate the path before the counted comparison runs.

### The type was the real limit, not the logic

`excludeColumns?: Column[]` — the legacy ENUM. A renamed board's terminal column is a valid id the enum cannot express, so the parameter was **unusable by exactly the callers that need it**. Widened to `readonly string[]` (and the internal `Set<ColumnId>` to `Set<string>`); every existing caller passes legacy ids, which stay assignable. This is the legacy-enum class showing up as a hard blocker rather than a style issue.

### Two sites, opposite failure directions

| site | on a renamed board | consequence |
|---|---|---|
| `duplicate-intake` | a finished sibling stayed **eligible** | newly filed work deduplicated ONTO something already done — no error, the request is answered with a finished card and the intent is lost |
| `near-duplicate-canonical` | a finished canonical still read **active** | the flag never cleared, so the flagged task stayed parked behind a user decision that could never arrive |

The second is the stranding that function's own FNXC note says it exists to prevent — it had stopped doing the one job it was written for.

### A vacuous assertion I caught and am recording

My first draft passed `createdAt` as an ISO string; the field is **epoch ms**, so the window filter compared a string to a number, `recent` was always empty, and both exclusion cases passed against an empty result for an unrelated reason. Green, and proving nothing.

The rule it cost me: **a negative assertion is only evidence when the matching positive case is green beside it.** Both files' positive cases ("still matches a live sibling", "keeps a live card active") are there for that reason, not for completeness.

Revert proof: **4 of 8 fail**, two per file.

gsxdsm:
All three findings are real, and all three are the same defect wearing different clothes: **a guard resolved by role in front of an action or a caller that still uses the literal.** Two are fixed here; the third is scoped below with evidence, because fixing it touches four files another batch is actively editing.

---

**1. `register-git-github.ts:2273` — custom review reopening targets `todo`. Fixed (`e63fefac0`). This one was mine.**

Confirmed exactly as described. I broadened the entry guard to accept any resolved review lane and left `moveTask(task.id, "todo", ...)` literal. That made the pair *worse than before*: on a board that renames review and declares no `todo`, the guard now admits the task and the move is then rejected — the card keeps its review-feedback document, never re-enters rework, and sits in review looking handled. Before the broadening it never got that far.

Now uses `resolveReboundTarget` (hold → intake → first declared); `todo` remains only for a workflow that cannot be resolved. Mutation-verified: restoring the literal move fails the renamed-board case. The v1 direction is pinned too — `resolveReboundTarget` returns the first declared column, which for a v1 upgrade *is* `todo`, so legacy behaviour survives without a literal producing it.

**2. `task-merge.ts:436` — dependency lifecycle context stays unset. Fixed (`7762c39a2`).**

Confirmed: `getTaskCompletionBlockerForStore` is the production path and omitted `satisfactionColumnsByTaskId` entirely, so every real call fell through to the legacy default and a dependency finished in a renamed complete lane read as unresolved — the depending task could never complete. The option was live only in tests that injected it by hand, which is this shape at its quietest: the guard reads as converted, its test passes, and production keeps the literal.

Both edges are now filled — the `dependencies` array *and* the single `blockedBy` marker, which reach the same helper by different paths. The store parameter stays narrow (`Pick<TaskStore, "getTask">`) and workflow resolution is feature-detected, so the callers that pass a partial store keep their existing behaviour rather than being forced to change.

Mutation-verified: removing the wiring fails 2 of the 4 new cases, the two renamed-lane ones. The suite also pins the paired negative (a genuinely unfinished dependency still blocks) and the partial-store fallback.

**3. `task-merge.ts:280` — merge callers retain the literal review gate. Confirmed, NOT fixed here.**

Verified: **no** production caller of `getTaskMergeBlocker` passes `reviewColumns`. All eight are bare —

| file | line |
|---|---|
| `merger.ts` | 6645 |
| `merger-ai.ts` | 1173 |
| `executor.ts` | 2373 |
| `self-healing.ts` | 6991, 7366, 7576, 10223 |

(`reviewColumns` *is* supplied at `moves.ts:782` and `task-priority.ts:143`, which is presumably why this looked wired.)

So the merge gate refuses every task on a renamed board with `task is in 'signoff', must be in 'in-review'`. Note this is **inert, not a regression** — the behaviour is identical to the pre-conversion literal. It is still real: the guard cannot fire.

I am deliberately not fixing it in this PR. All four files are `packages/engine`, which `batch-engine` is actively committing to; touching eight call sites across `merger.ts`, `merger-ai.ts`, `executor.ts`, and `self-healing.ts` from here would collide head-on. The clean shape is a `getTaskMergeBlockerForStore(store, task, opts)` wrapper mirroring the `getTaskCompletionBlockerForStore` one just fixed above, so call sites migrate individually — and the ratchet has to drive the *wrapper*, since the wrapper is precisely what was missing. Handing that to whoever owns engine.

gsxdsm:
## Acting on the last finding: I swept for census-invisible literal LISTS. Here is the measurement.

`b861fcd0cf` fixes the first one found. Gate **161 / 13 / 487 / 71**, lint clean, `--strict` exits 0.

### The sweep

Legacy-id ARRAY or SET literals across the five source trees, tests excluded:

```
grep -rnE '\[\s*"(todo|triage|in-progress|in-review|done|archived)"' \
  packages/{core,engine,dashboard,cli}/src packages/dashboard/app
→ 84 sites
```

**84 is a raw grep count, not 84 defects** — I triaged core's 15 before reporting anything, and most are legitimate:

| site | verdict |
|---|---|
| `prompt-overrides.ts` `roles: ["triage"]` ×4 | **false positive.** `roles` is `AgentCapability[]`; `"triage"` is the planning AGENT, not the deleted column |
| `board-config.ts` transition map ×6 | legacy board definition, used as the no-workflow fallback |
| `blocker-fanout.ts` `LEGACY_*` ×2, `default-workflow-hooks.ts` `LEGACY_*` ×2 | already-named documented defaults |
| `blocker-fanout.ts:80` `BLOCKER_ESCALATION_COLUMNS` | **a real gate, now fixed** |

So core's honest yield is **1 of 15**. I have not triaged the other 69 (engine 20-ish, dashboard app/src 30-ish, cli 4) — the file list is above and the grep is one line, but a raw count is not evidence and I will not present it as one.

### The one that was real

`BLOCKER_ESCALATION_COLUMNS` is a hard-coded `Set` gating escalation of a stale high-fan-out blocker — **sitting directly beside three already-converted options in the same interface**. A Set literal is a definition, not a comparison, so no census entry ever pointed at it.

On a renamed board `shouldEscalate` was false for every blocker: a stale blocker holding up many cards never escalated. And the fan-out numbers themselves stayed **correct** — the metric says there is a problem while the mechanism that acts on it is switched off. That is worse than a broken metric, because the dashboard looks right.

Now an optional `escalationColumns` defaulting to the exported Set, so unconverted callers are byte-identical. Revert proof: restoring the Set membership fails the renamed case; the legacy case passes either way.

### What this says about "N → 0"

Two files in a row now where the census's count was not the whole guard:

- `duplicate-intake` — a literal LIST (`DEFAULT_EXCLUDE_COLUMNS`) gated the path *before* the counted comparison, and the counted comparison was unreachable whenever it mattered;
- `blocker-fanout` — the only lane check that mattered was a Set literal, and the file already read as converted.

Both were found by a **failing test**, not by the census. The practical rule for the remaining batch: after taking a file to 0, grep it for `["`-shaped lane literals and `Set` literals before believing the number. #2763 calls the census total a floor; these are two concrete mechanisms for why.

gsxdsm:
**Retracting a hazard I flagged at you — it was wrong, and it is now moot. Convert the stall gates whenever you like.**

Via the batch-dashboard-app work order I recorded a coupling between `packages/dashboard/app/hooks/useTasks.ts` and `packages/core/src/in-review-stall.ts:179`, saying converting the core gate **alone** would regress: both keyed on the literal, so no stall badge was produced and none needed clearing, and fixing one half would make cards read "stalled" while an agent worked on them.

**That was wrong on one of the three signals.** `getInReviewStalledSignal` is **already** trait-converted (U4), so `inReviewStalled` is produced on a renamed board *today* — and the dashboard literal was *already* refusing to clear it. The bug was live, not scheduled, and my framing would have had you deferring a conversion to avoid causing something that had already happened.

The dashboard side is fixed, by **deleting** the column check rather than converting it: the line below already implies it, since all three stall fields are produced only for review-lane cards (each producer gates on review itself — `getInReviewStallReason`, `getInReviewStalledSignal`, `detectStalledReview`). Converting wasn't available anyway — it would need a flags map threaded into `useTasks`, and that map is built in App *from* the list that hook produces.

**Net for you:** `getInReviewStallReason` and `detectStalledReview` are ordinary conversions with no dashboard-side ordering constraint. Nothing in `packages/dashboard/app` depends on their literals any more.

Sorry for the noise — a stale hazard costs more than no hazard, so I would rather retract it loudly than let it sit in your work order.

gsxdsm:
**Batch-owner note from batch-dashboard-app: MERGE main into your branch, do not rebase onto it.** You will hit this the moment main moves.

`scripts/lib/lifecycle-column-census-baseline.json` is rewritten by **every `--strict` run**, not just `--update-baseline` — that is the self-lowering ratchet working as intended. So nearly every batch commit touches it.

Rebasing a long batch branch therefore replays every commit and re-conflicts on that file at **each step**. Worse, once a few steps resolve differently, commits *inside* your branch start conflicting with **each other** on real source files — I hit a `DockTaskList.tsx` conflict between two batch commits that had never conflicted when they were written. Two full rebase attempts made no progress.

Merging resolves once:

```
git merge origin/main --no-edit
git checkout origin/main -- scripts/lib/lifecycle-column-census-baseline.json
git add scripts/lib/lifecycle-column-census-baseline.json && git commit --no-edit
```

One conflict, mechanical resolution. Only the tip's baseline has to be correct; intermediate ones never mattered.

Two smaller traps from the same session:

- **`--strict` writes.** If you redirect its stdout to `/dev/null` in a loop you will miss the *"baseline rewritten downward, COMMIT IT"* line and be left with a dirty tree that aborts your next `git checkout`. I misdiagnosed this as the tool being silent; it isn't, I had silenced it.
- Branch-switching in a shared worktree can leave tracked binaries (`e2e/__screenshots__/*.png`) looking modified. Restore them rather than committing them into the batch.

Written up in `docs/solutions/workflow-learnings/mega-batch-branches-merge-do-not-rebase.md` on `batch-dashboard-app`.

gsxdsm:
**Check your batch for inert conversions before it lands. I just found five in one of my own tranches, and nothing we run would have caught them.**

## The class

A conversion that replaces a literal with a resolved-flags seam the callers never supply:

```ts
// utils
export function isTaskStuck(task, timeoutMs, dataAsOfMs, columnFlags?) { … }
isTaskStuck(task, timeoutMs, lastFetchTimeMs);        // ← every caller. columnFlags is always undefined.

// components
interface PrPanelProps { taskColumnFlags?: … }         // declared
<PrPanel taskColumnFlags={flags} />                    // parent passes it
export function PrPanel({ taskColumn }: PrPanelProps)  // ← never destructured. Dropped.
```

## Why nothing catches it

- **`tsc` passes** — the parameter/prop is optional, so omitting it is legal.
- **Tests pass** — the fallback *is* the old behaviour. That is the point of the fallback.
- **The census goes DOWN** — it counts comparisons, and the literal really is gone. The instrument measuring this program scores the inert version as a win.
- **Grep is useless** — searching the symbol finds it in exactly the places you added it, and says nothing about whether a value flows.

## My hit rate, which is why I am writing to you

In the tranche I had reported as clean: **5 of 9 conversions were inert.** Two were found by review (`worktreeGrouping`, `PrPanel`); the other three only by building the guard. I had verified each one by "it compiles and the suite is green".

## Detection

Two AST checks, both cheap — `packages/dashboard/app/__tests__/resolved-flags-seams-have-suppliers.test.ts` on `batch-dashboard-app` if you want the source:

1. **Arity** — for every exported function whose *last* parameter matches `/[Cc]olumnFlags$/`, require at least one call site passing that many arguments. Catches the utils/hooks shape, which is most of core and engine.
2. **Destructuring** — for every `<Name>Props` declaring such a prop, require the function `<Name>` to destructure it. Catches the component shape.

Note my first attempt at (2) **could not fire**: it checked whether the prop name appeared as a JSX attribute anywhere, and the parent legitimately passes it, so the name was present while the value was still dropped one level down. A name-level check cannot see a component-level drop — verify your guard fails on a reintroduced defect before trusting it.

## The rule that falls out

**Wire a supplier, or delete the seam and leave the literal counted.** I reverted three of mine for exactly this: their callers genuinely cannot supply flags (one sits 60 lines above where the flags are derived). An optional parameter every caller omits is *strictly worse* than the literal — the literal is honest and stays in the census, while the parameter banks the credit and runs the fallback forever.

My package went from a reported 2 back up to **6** as a result. The larger number is the true one.

PR #2782, opened 2026-07-30
fix(tests): 2 TaskCard reds — the assertions pinned a jsdom detail, not the CSS

gsxdsm:
## What was red

Both failures in the dashboard `app:components-b` lane. **Neither is a style regression** — the CSS is byte-unchanged and correct in both cases.

## Root cause: a jsdom upgrade, not a CSS change

jsdom does not substitute `var()`. What it does *instead* changed under us in **4819c2634 (jsdom 27.4.0 → 29.1.1)**:

| | jsdom 27 | jsdom 29 |
|---|---|---|
| unresolvable shorthand (`padding`) | echoes raw text `var(--space-xs) var(--space-sm)` | computes to `"0"` |
| single-value longhand (`gap`) | echoes | still echoes |

Tests asserting the **echoed string** were pinning a jsdom implementation detail. The bump turned them red with nothing changed in the product.

### 1. `renders a promote action when onPromote is provided`

`expected 'var(--space-xs) var(--space-sm)', received '0'`. `.card-promote-action` still declares exactly that padding (`TaskCard.css:1129`).

### 2. `FN-4511 keeps GitHub badge and timer chip geometry in parity`

`expected '1px' to be 'medium'`. The chips **are** in parity:

- badge: `border: 1px solid transparent`
- timer chip: `border: var(--btn-border-width) solid transparent`
- `--btn-border-width: 1px` (`styles.css:183`)

Here jsdom **discards** the unresolvable width rather than echoing it, so `borderTopWidth` falls back to the initial value `medium`. The existing `|| "1px"` fallbacks could not save it — `medium` is a non-empty string, so it was the fallback that never ran, not the value that was missing.

## The fix

Both assertions now read the **declared** value from the mounted stylesheet's CSSOM and resolve a single `var()` against `:root`. That is stable across jsdom versions and is what the assertions always meant.

Via the CSSOM rather than a regex over the CSS text **on purpose**: a hand-rolled matcher over grouped selectors silently matches the wrong rule and still reports success. Everything jsdom *can* resolve (font-size, line-height, gap, padding parity) stays asserted against computed style.

## Evidence

`components-b`: **1688/1688** (was 2 failed). Mutation-proved — both fail as they should:

| mutation | result |
|---|---|
| timer chip border `1px → 2px` | `expected '2px' to be '1px'` |
| promote padding tokens changed | `expected 'var(--space-sm) var(--space-lg)' to be 'var(--space-xs) var(--space-sm)'` |

**My first border mutation passed**, which would have read as a vacuous guard. It had patched the wrong one of five identical `border: var(--btn-border-width)` lines in the file. Re-run against `TaskCard.css:1094` it fails correctly. Recorded because the mutation, not the guard, was the thing that was wrong — a passing mutation is a claim that needs checking too.

`pnpm lint` clean. Test-only — no production file or CSS touched (mutations reverted; `git diff` clean).

## Ownership

`packages/dashboard/app` belongs to the **batch-dashboard-app** owner (u12) under the mega-batch split. This is fix-forward on a red rather than a conversion, confined to one test file, and touches no production code — it should not conflict with the batch.

## Not fixed here

The `app:app` lane has **10 pre-existing failures** in `App.test.tsx` (deep-link handling, board branch filters, FN-5817 mobile shell). They were masked by this lane failing first — the runner skips remaining lanes after the first failure, so they only became visible once components-b went green. Separate change.

### DIFF
--- a/packages/dashboard/app/components/__tests__/TaskCard.test.tsx
+++ b/packages/dashboard/app/components/__tests__/TaskCard.test.tsx
@@ -157,6 +157,47 @@ function seedAgentsCache(projectId: string, agents: Array<{ id: string; name: st
   );
 }
 
+/*
+FNXC:TaskCardParity 2026-07-31-00:25:
+READ DECLARED CSS FROM THE CSSOM — `getComputedStyle` cannot be trusted for tokenized values here.
+
+jsdom does not substitute `var()`. Worse, WHAT it does instead changed under us: on jsdom 27 an
+unresolvable shorthand echoed its raw text (`padding` read back as
+"var(--space-xs) var(--space-sm)"), and on jsdom 29 (bumped in 4819c2634) the same declaration
+computes to "0", while single-value longhands like `gap` still echo. Tests that asserted the echoed
+string were pinning a jsdom implementation detail, so the upgrade turned them red with the CSS
+completely unchanged.
+
+This reads the DECLARED value off the mounted stylesheet's CSSOM and resolves a single `var()`
+against `:root`, which is stable across jsdom versions and is what the assertions actually meant.
+The CSSOM is used rather than a regex over the CSS text on purpose: a hand-rolled matcher over
+grouped selectors silently matches the wrong rule and still reports success.
+
+Later rules win, matching the cascade for equal specificity.
+*/
+function declaredStyle(selector: string, property: string): string {
+  let declaration: string | undefined;
+  for (const sheet of Array.from(document.styleSheets)) {
+    for (const rule of Array.from(sheet.cssRules ?? [])) {
+      if (!(rule instanceof CSSStyleRule)) continue;
+      if (!rule.selectorText.split(",").some((part) => part.trim() === selector)) continue;
+      const value = rule.style.getPropertyValue(property).trim();
+      if (value) declaration = value;
+    }
+  }
+  expect(declaration, `no ${property} declaration found for ${selector}`).toBeDefined();
+  return declaration!;
+}
+
+/** Resolves a bare `var(--token)` against `:root`; any other value is returned unchanged. */
+function resolveCssToken(value: string): string {
+  const token = /^var\(\s*(--[\w-]+)\s*\)$/.exec(value.trim());
+  if (!token) return value.trim();
+  const resolved = getComputedStyle(document.documentElement).getPropertyValue(token[1]).trim();
+  expect(resolved, `token ${token[1]} resolved to nothing`).not.toBe("");
+  return resolved;
+}
+
 function mountCssForBadgeTests() {
   const style = document.createElement("style");
   style.textContent = loadAllAppCss();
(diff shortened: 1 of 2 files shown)


gsxdsm:
### Evidence on the `app:app` lane this PR unmasks

Measured after this change makes `components-b` green, so the runner stops skipping later lanes:

```
app:components-a   1195 passed
app:components-b   1688 passed        <- this PR (was 2 failed)
app:app              10 failed | 131 passed (141)
[dashboard-quality] skipped 9 lane(s) after first failure
```

**These 10 are long-standing, not caused or worsened by this PR.** Bisected on the untouched lane:

| commit | app:app |
|---|---|
| `origin/main` | 10 failed / 131 passed |
| `main~5` | 10 failed / 131 passed |
| `main~20` | 10 failed / 131 passed |
| `main~60` | 10 failed / 131 passed |
| `main~150` | 10 failed / 131 passed |

Identical count at every point — they simply were not *visible* before, because `components-b` failed first and the runner skips the remaining lanes.

**All 10 share one symptom:** the App shell never leaves the Loading state. The rendered body is the `top-progress-bar` + `header-wrapper` with no board or list content, so every query fails on a missing element rather than a wrong one — `Unable to find an element with the text: Task Alpha`, `+ New Task`, etc. No console error, no unhandled rejection, no failed-fetch warning is emitted, so it presents as a silent load stall rather than a thrown mock gap.

Failing cases: `App deep link handling` (2), `App view switching` (1), `App Planning Mode` (1), `App board branch filters` (5), `FN-5817 mobile auto-merge toggle stability` (1).

Note the 9 still-skipped lanes: with `app:app` red, `app:chat`, `app:settings`, the 4 `app:backfill-*` shards and the `api:*` lanes have not run in this sweep, so **the dashboard package is not proven green below `app:app`** — there may be more behind it. Flagging rather than guessing.

Not fixed here: it is a genuine debugging job in `packages/dashboard/app` (batch-dashboard-app / u12), and a fixture edit that made the queries pass without explaining the load stall would be appeasement.

gsxdsm:
> **Theme overrides remain untested** — `resolveCssToken` reads only the default token mounted on `document.documentElement`, while the test fixture excludes `theme-data.css`.

Good catch, and it is **worse than untested — it was a live bug.** Verified rather than assumed:

```
theme-data.css:993   [data-color-theme="factory"]       --btn-border-width: 2px;
theme-data.css:1200  [data-color-theme="factory-mono"]  --btn-border-width: 2px;
```

`.card-github-badge` declared a hardcoded `1px` while the sibling footer chips (`.card-time-indicator` and friends) declare `var(--btn-border-width)`. So on those two **shipped** themes the timer chip rendered a 2px border and the badge stayed at 1px — exactly the footer-chip misalignment FN-4511 exists to prevent, invisible to the test that owns the invariant.

**Fixed at the source** (`50902797b6`): the badge now uses the token, per the standing rule against hardcoded pixels in component CSS. Both chips move together under any theme.

The parity test now proves it — the token is overridden the way a theme does, and both chips must report `2px`:

| mutation | result |
|---|---|
| restore the hardcoded `1px` on the badge | `expected '1px' to be '2px'` |

One consequence worth calling out: `FN-4511 preserves transparent border slot` required `border: 1px solid transparent` **verbatim**, so it pinned the very literal that caused the bug — a test enforcing the defect. Its regex now accepts the token or a literal length. The invariant it is *named* for is unchanged and still asserted (a transparent slot is reserved so hover/focus can colour it without shifting layout), and it still fails on `border: none` and on a non-transparent colour — both mutation-checked.

`components-b`: **1688/1688**. lint clean.

Note this makes the PR no longer test-only — it now carries a one-line CSS fix. Flagging explicitly since the original body said otherwise.

PR #2783, opened 2026-07-30
batch-core: one shared landed-lane helper for the source-issue surfaces (75 → 72)

gsxdsm:
## batch-core continued — the source-issue cluster

Follow-on to #2780 (merged). Scope is still `packages/core` + `packages/dashboard/src`.

### The defect

Five places asked the same question — *has this task landed?* — and all five compared against the literal `done`:

| surface | consequence on a renamed board |
|---|---|
| GitHub source-issue commenter | never comments on or closes the source issue |
| GitLab source-issue commenter | same |
| GitLab `closedAt` backfill reconciler | finds nothing, reports a clean scan |
| session-diff boundary | finished tasks diff against an already-merged branch |
| tracking-comment transition | (already converted; left alone) |

The commenters are the sharpest case: they returned **before reading a single setting**, so on a renamed board the feature looked *disabled* rather than broken — an operator checking `githubCommentOnDone` would see it enabled and still get nothing.

The backfill is the quietest: `scanned: N, filled: 0` reads as "nothing to do", so the failure was indistinguishable from success.

### The fix

One home: `packages/dashboard/src/task-lifecycle-lanes.ts`. Callers now only ask.

Five copies of one question is exactly how the halves drift apart — the motivating incident is FN-6115 → FN-6118 → FN-6123, where the same affordance was fixed three times because it lived in two components. This also folds in the duplicate landed-lane helper I had left in `register-session-diff-routes.ts` in the previous PR, which was the sixth copy waiting to happen.

Two helpers, and the difference is deliberate:

- **`landedColumnsForTask`** — `complete ∪ archived`. Membership, since a board may declare more than one column carrying either role, and `columnsWithFlag(...)[0]` would silently ignore the second.
- **`completeColumnsForTask`** — complete only. The GitLab backfill's own FNXC note records that archived tasks live in `archiveDb` and are *intentionally* excluded, so it must not widen to the archived role just because the shared helper offers it. Today it lists with `includeArchived: false` and would see no archived rows either way — but that is an incidental property of the query, not the contract. The test pins the difference so the two are not later "simplified" into one, which would change that caller's behaviour without touching it.

Both treat an **empty** resolved set as *unexpressed*, not absent — the v1 hazard: `synthesizeDefaultColumns` upgrades a v1 graph with `traits: []` on every column, so reading empty as "no complete lane" would stop these surfaces firing on every pre-v2 project.

The reconciler is two-stage on purpose: the cheap provider and `closedAt` tests run first and reject almost everything, so a workflow read only happens for real candidates, and it shares one IR cache across the scan — one read per distinct workflow rather than per task.

### Census

`batch-core` scope **75 → 72**; repo total **338**.

### Verification

- `pnpm --filter @fusion/dashboard exec tsc --noEmit -p tsconfig.json` → 0 errors
- `pnpm lint` → 0 errors
- commenter + reconciler suites → **63 passed**; helper suite → **5 passed**
- **Mutation-verified:** making the helper ignore its resolved set fails 1 of 5.

---

## Round 2 — server.ts, chat.ts, and a correction

**Census: 75 → 67** across this PR.

### The correction (see the review thread above)

My first pass gated the source-issue commenters on `landedColumnsForTask` (`complete ∪ archived`), which **widened** the trigger — `to === "done"` never fired on archival, and the landed set does. Both commenters now use `completeColumnsForTask`, and the unused `hasTaskLanded` wrapper is gone.

The ratchet for it is pinned on the **default** board, deliberately: a widening is visible exactly where the legacy names still apply, so no renamed-board fixture would catch it.

### `chat.ts` — three sites, and a pair that had to move together

- **Chat verification** required `column === "in-progress"`, so on a renamed board every chat-driven verification was refused with a message naming a column the board does not have.
- **The planner refinement pair.** Two separate guards decide this feature: `createSession` *registers* the tool only for a finished task, and the tool's own `execute()` *refuses* a non-finished source. Both compared `done`. Converting only one half would have offered the tool and then had it refuse itself — the half-converted-pair shape. The new test asserts **both** halves in one case (tool present *and* refinement created), and each half reverted independently fails it.

Existing `chat-manager` coverage caught neither revert, which is why the case exists rather than relying on the suite that was already there.

Complete-only again, not the landed set: an archived task is off the board and is not a refinement source.

### `server.ts`

- **Planner-chat retention** — the archival cutoff was a literal, so on a renamed board task-planner chat sessions were retained forever; the rule this listener exists to enforce never fired. Resolved, and awaited inside the existing fire-and-forget chain rather than by making the listener `async` — `task:moved` has synchronous subscribers whose ordering is load-bearing elsewhere, and a chat-row delete is not the right place to introduce a microtask boundary into that emit.

- **`isBadgeEligibleTask` — deliberately NOT converted, and marked as backlog.** On a renamed board it is genuinely wrong: an archived card stays badge-eligible, its snapshot is never evicted, and the cache grows for the daemon's lifetime — the exact memory leak the predicate was added to fix, back under a different column name.

  What blocks it is measured, not assumed: both callers are synchronous `task:updated` / `task:created` listeners whose next statement is documented as *"Update local cache immediately"*, so awaiting lets a second event for the same task interleave between the eligibility check and the cache write.

  I did **not** add an optional `archivedColumns` parameter, because nothing could fill it — the callers are the sync listeners. That is the inert-injection shape this PR's own review caught twice on #2780: the predicate would read as converted, its test would pass by injecting the value, and production would keep the literal. The unblocking change (a resolved-archived-lane cache on the badge-snapshot scope, keeping the predicate synchronous) is recorded at the site.

### Verification

- `tsc --noEmit` → 0 errors; `pnpm lint` → 0 errors
- `chat-manager` → 101 passed; commenter/reconciler/helper/badge suites → 55 passed
- Mutation-verified per fix, including each half of the refinement pair separately

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved task lifecycle handling for renamed workflow lanes, including completed, archived, landed, and in-progress states.
  * Task lists now exclude completed tasks regardless of the completion lane’s name.
  * Chat verification and refinement actions now recognize configured workflow lanes.
  * GitHub and GitLab completion comments trigger only for genuinely completed tasks, not archived tasks.
  * Knowledge index refreshes and GitLab metadata updates now support custom completion lanes.
* **Tests**
  * Added regression coverage for renamed completion lanes and archived-task behavior.

### DIFF
--- a/packages/core/src/__tests__/postgres/merge-queue-renamed-review-lane.pg.test.ts
+++ b/packages/core/src/__tests__/postgres/merge-queue-renamed-review-lane.pg.test.ts
@@ -1,5 +1,5 @@
 /*
-FNXC:WorkflowResolvedColumns 2026-07-31-07:55 (batch-core):
+FNXC:WorkflowResolvedColumns 2026-07-30-07:55 (batch-core):
 
 THE MERGE QUEUE MUST FILL AND DRAIN ON A BOARD THAT RENAMED ITS REVIEW LANE.
 
--- a/packages/core/src/__tests__/postgres/merge-queue-renamed-review-lane.pg.test.ts
+++ b/packages/core/src/__tests__/postgres/merge-queue-renamed-review-lane.pg.test.ts
@@ -1,0 +1,120 @@
+/*
+FNXC:WorkflowResolvedColumns 2026-07-31-07:55 (batch-core):
+
+THE MERGE QUEUE MUST FILL AND DRAIN ON A BOARD THAT RENAMED ITS REVIEW LANE.
+
+`enqueueMergeQueueInTransaction` admitted a card only from the literal `in-review`, and
+`dequeueMergeQueueOnColumnExitInTransaction` removed it only when leaving that same literal. Both run
+inside the move transaction with a `tx` handle and no store, so neither could resolve the lane itself
+— the resolved set now comes from `moves.ts`, which already resolves the workflow for the move.
+
+WHY THE PAIR IS TESTED TOGETHER AND NOT SEPARATELY. Converting one half alone is undetectable: if
+enqueue keeps the literal, nothing is ever queued on a renamed board, so a dequeue test has nothing to
+observe and passes vacuously. If dequeue keeps the literal, the queue fills and never drains — a leak
+that only shows up on the SECOND move. One test drives the whole cycle so neither half can regress
+without failing.
+
+WHY A LIVE STORE. The bug is which column id the guard compares against, and that id comes from the
+task's own persisted workflow. A mock handing back a lifecycle struct would assert my own assumption
+about what `resolveReviewColumns` returns — the substitution that has produced vacuous tests all
+through this program. This drives real PostgreSQL and asserts on OBSERVED QUEUE ROWS.
+
+LANE. `.pg.test.ts`, skipped by `pgDescribe` when no PostgreSQL is reachable, so the merge gate is
+unaffected. Throwaway per-file database; never port 4040.
+*/
+import { beforeAll, beforeEach, afterEach, afterAll, expect, it } from "vitest";
+import "@fusion/core"; // registers the built-in column traits
+
+import { pgDescribe, createSharedPgTaskStoreTestHarness } from "../../__test-utils__/pg-test-harness.js";
+
+pgDescribe("merge queue fills and drains on a renamed review lane", () => {
+  const harness = createSharedPgTaskStoreTestHarness({ prefix: "fusion_mq_renamed_review" });
+
+  beforeAll(harness.beforeAll);
+  afterAll(harness.afterAll);
+  beforeEach(async () => { await harness.beforeEach(); });
+  afterEach(async () => { await harness.afterEach(); });
+
+  /** `signoff` carries the merge trait; this board declares no `in-review` column at all. */
+  async function renamedReviewWorkflow(store: ReturnType<typeof harness.store>) {
+    return store.createWorkflowDefinition({
+      name: "renamed-review",
+      ir: {
+        version: "v2",
+        name: "renamed-review",
+        columns: [
+          { id: "inbox", name: "Inbox", traits: [{ trait: "intake" }] },
+          { id: "building", name: "Building", traits: [{ trait: "wip" }] },
+          { id: "signoff", name: "Signoff", traits: [{ trait: "merge" }] },
+          { id: "shipped", name: "Shipped", traits: [{ trait: "complete" }] },
+        ],
+        nodes: [{ id: "start", kind: "start", column: "inbox" }, { id: "end", kind: "end", column: "shipped" }],
+        edges: [{ from: "start", to: "end" }],
+      },
+    } as never);
+  }
+
+  it("enqueues on entering the renamed review lane and dequeues on leaving it", async () => {
+    const store = harness.store();
+    const definition = await renamedReviewWorkflow(store);
+    const task = await store.createTask({ description: "renamed review card", workflowId: definition.id } as never);
+
+    /* Adjacency is derived from the graph, so the card walks the board rather than jumping. */
+    await store.moveTask(task.id, "building" as never, { bypassGuards: true } as never);
+    /*
+    The merge queue is filled by the COMPLETION HANDOFF, not by a bare move into the review lane —
+    `moveTask` alone walks the board without enqueuing. Driving the real handoff is what exercises
+    `enqueueMergeQueueInTransaction`, which is the guard under test.
+    */
+    await store.handoffToReview(task.id, { ownerAgentId: null, evidence: { reason: "test", runId: "r1", agentId: "test" } } as never);
+
+    /*
+    Half one. With the literal, `taskRow.column !== "in-review"` rejected every card on this board and
+    the enqueue recorded a `mergeQueue:enqueue-rejected` audit instead of a row — the merge queue was
+    simply never used, and nothing surfaced that as an error.
+    */
+    expect(await store.getMergeQueuedTaskIdsAsync()).toContain(task.id);
+
+    await store.moveTask(task.id, "building" as never, { bypassGuards: true } as never);
+
+    /*
+    Half two, and the one that fails independently: with dequeue on the literal, `previousColumn !==
+    "in-review"` is true for `signoff`, so the helper returns early and the row stays. The queue would
+    fill and never drain, and the leak only becomes visible on this second move.
+    */
+    expect(await store.getMergeQueuedTaskIdsAsync()).not.toContain(task.id);
+  });
+
+  it("does NOT dequeue when moving between two review lanes on the same board", async () => {
+    /*
+    The paired negative, and the reason the set is broad rather than `lifecycle.review`. A board may
+    declare a merge-orchestration lane AND a separate human sign-off lane; a card moving between them
+    has not left review. A single-id answer would treat the second lane as "outside review" and drop
+    the card out of the merge queue mid-review.
+    */
+    const store = harness.store();
+    const definition = await store.createWorkflowDefinition({
+      name: "two-review-lanes",
+      ir: {
+        version: "v2",
+        name: "two-review-lanes",
+        columns: [
+          { id: "inbox", name: "Inbox", traits: [{ trait: "intake" }] },
+          { id: "signoff", name: "Signoff", traits: [{ trait: "merge" }] },
+          { id: "approval", name: "Approval", traits: [{ trait: "human-review" }] },
+          { id: "shipped", name: "Shipped", traits: [{ trait: "complete" }] },
+        ],
+        nodes: [{ id: "start", kind: "start", column: "inbox" }, { id: "end", kind: "end", column: "shipped" }],
+        edges: [{ from: "start", to: "end" }],
+      },
+    } as never);
+    const task = await store.createTask({ description: "two-lane card", workflowId: definition.id } as never);
+
+    await store.handoffToReview(task.id, { ownerAgentId: null, evidence: { reason: "test", runId: "r1", agentId: "test" } } as never);
+    expect(await store.getMergeQueuedTaskIdsAsync()).toContain(task.id);
+
+    await store.moveTask(task.id, "approval" as never, { bypassGuards: true } as never);
+
+    expect(await store.getMergeQueuedTaskIdsAsync()).toContain(task.id);
+  });
+});
--- a/packages/core/src/__tests__/postgres/store-archive-reads.pg.test.ts
+++ b/packages/core/src/__tests__/postgres/store-archive-reads.pg.test.ts
@@ -22,6 +22,45 @@ pgDescribe("TaskStore archived read parity (PostgreSQL)", () => {
   afterEach(h.afterEach);
   afterAll(h.afterAll);
 
+  /*
+  FNXC:WorkflowResolvedColumns 2026-07-30-18:50 (batch-core):
+
+  A CARD SITTING IN THE BOARD'S OWN ARCHIVE LANE IS ALREADY ARCHIVED.
+
+  `archiveTask` refuses a card that is already archived. Keyed on the literal, a board whose archive
+  lane is named `attic` did not refuse — the card was archived a second time, from a lane the board
+  itself calls archived.
+
+  THE FIXTURE MATTERS, and my first version of this test was vacuous because of it. Calling
+  `archiveTask` first does NOT produce a renamed-lane card: the archive path stamps `column:
+  "archived"` (`archiveEntryToTask`, serialization.ts:353), so the guard only ever sees the literal
+  and both the literal and resolved forms pass. The card has to be MOVED into `attic` by an ordinary
+  move for the renamed lane to reach the guard at all.
+
+  The unarchive side is deliberately not covered here: its input always carries the literal by
+  construction, which is recorded at that guard.
+  */
+  it("refuses to archive a card already sitting in the board's renamed archive lane", async () => {
+    const store = h.store();
+    const definition = await store.createWorkflowDefinition({
+      name: "renamed-archive",
+      ir: {
+        version: "v2",
+        name: "renamed-archive",
+        columns: [
+          { id: "building", name: "Building", traits: [{ trait: "wip" }] },
+          { id: "attic", name: "Attic", traits: [{ trait: "archived" }] },
+        ],
+        nodes: [{ id: "start", kind: "start", column: "building" }, { id: "end", kind: "end", column: "attic" }],
+        edges: [{ from: "start", to: "end" }],
+      },
+    } as never);
+    const task = await store.createTask({ description: "already in the attic", workflowId: definition.id } as never);
+    await store.moveTask(task.id, "attic" as never, { bypassGuards: true } as never);
+
+    await expect(store.archiveTask(task.id, { cleanup: false } as never)).rejects.toThrow(/already archived/);
+  });
+
   it("composes archived snapshots into list, search, and detail reads", async () => {
     const store = h.store();
     const first = await store.createTaskWithReservedId(
--- a/packages/core/src/__tests__/postgres/store-comments.pg.test.ts
+++ b/packages/core/src/__tests__/postgres/store-comments.pg.test.ts
@@ -109,6 +109,48 @@ pgTest("TaskStore addComment steering + refinement (PostgreSQL)", () => {
   afterEach(h.afterEach);
   afterAll(h.afterAll);
 
+  /*
+  FNXC:WorkflowResolvedColumns 2026-07-30-17:55 (batch-core):
+
+  AUTO-REFINEMENT MUST FIRE FOR A TASK FINISHED IN A RENAMED COMPLETE LANE.
+
+  A user comment on finished work creates a refinement task. Keyed on `task.column === "done"`, a
+  renamed board never entered that branch — the operator got silence where the feature promises a
+  follow-up, and nothing was logged because the branch was not reached rather than failing inside it.
+
+  Driven through the real store and a real workflow definition so the assertion is on an OBSERVED
+  refinement row, not on my own belief about what the resolver returns for a renamed lineage.
+  */
+  it("creates a refinement for a user comment on a task in a RENAMED complete lane", async () => {
+    const store = h.store();
+    const definition = await store.createWorkflowDefinition({
+      name: "renamed-complete",
+      ir: {
+        version: "v2",
+        name: "renamed-complete",
+        columns: [
+          { id: "building", name: "Building", traits: [{ trait: "wip" }] },
+          { id: "shipped", name: "Shipped", traits: [{ trait: "complete" }] },
+        ],
+        nodes: [{ id: "start", kind: "start", column: "building" }, { id: "end", kind: "end", column: "shipped" }],
+        edges: [{ from: "start", to: "end" }],
+      },
+    } as never);
+    const task = await store.createTask({ description: "finished on a renamed board", workflowId: definition.id } as never);
+    await store.moveTask(task.id, "shipped" as never, { bypassGuards: true } as never);
+
+    const before = (await store.listTasks({ slim: true } as never)).length;
+    await store.addComment(task.id, "please also handle the empty case", "user" as never);
+
+    /*
+    The witness is a NEW task existing, not a mock call: `refineTask` creates the follow-up, and
+    asserting on the row proves the branch ran end to end rather than that a spy was invoked.
+    */
+    const after = await store.listTasks({ slim: true } as never);
+    expect(after.length).toBe(before + 1);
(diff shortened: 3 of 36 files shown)


gsxdsm:
Real, and the important part is that it is a **behaviour change I introduced**, not merely an incomplete conversion. Fixed in `7f6c48a52`.

**What I got wrong.** The original gate was `event.to !== "done"` — a single column. I converted it with `hasTaskLanded`, which resolves `complete ∪ archived`. That widened the trigger: on the **default** board, archiving a task would now post a source-issue comment that `to === "done"` never posted. Resolving a role by trait invites reaching for the widest set that "means finished", and archival does mean finished — but it is not what this listener fired on, and its source-issue behaviour belongs to the tracking commenter, not this one.

A vocabulary conversion has to answer the same question under a new name, not a bigger one. Both commenters now use `completeColumnsForTask`, and the `hasTaskLanded` wrapper is deleted since nothing calls it.

**The ratchet is pinned on the DEFAULT board, deliberately.** This is the one failure mode a renamed-board fixture cannot catch — the widening is visible precisely where the legacy names still apply, so every fixture I had been writing this session would have passed it. `does NOT comment when a task is archived rather than completed`; mutation-verified, restoring the landed set fails it.

I had actually flagged this distinction in the same PR — the GitLab backfill reconciler deliberately uses `completeColumnsForTask` because its own note excludes archived work — and then failed to apply the same reasoning two files over. The narrow/broad split now has a test pinning it in both directions rather than living only in a comment.

---

Also in this push, unrelated to the finding:

- **`server.ts` planner-chat retention** — the archival cutoff compared the literal, so on a renamed board task-planner chat sessions were retained forever; the retention rule this listener exists to enforce simply never fired. Now resolves the archived role, awaited inside the existing fire-and-forget chain rather than by making the listener `async` — `task:moved` has synchronous subscribers whose ordering is load-bearing elsewhere, and a chat-row delete is not the right place to introduce a microtask boundary into that emit.

- **`isBadgeEligibleTask` — deliberately NOT converted, and marked.** On a renamed board it is genuinely wrong: an archived card stays badge-eligible, so its snapshot is never evicted and the cache grows for the daemon's lifetime — the exact memory leak the predicate was added to fix, back under a different column name. It stays counted as backlog. What blocks it is measured, not assumed: both callers are synchronous `task:updated`/`task:created` listeners whose next statement is documented as *"Update local cache immediately"*, so awaiting lets a second event for the same task interleave between the eligibility check and the cache write. I did **not** add an optional `archivedColumns` parameter, because nothing could fill it — that is the inert-injection shape #2780 caught twice. The unblocking change (a resolved-archived-lane cache on the badge-snapshot scope, keeping the predicate sync) is recorded at the site.

**Verification:** `tsc --noEmit` 0 errors, `pnpm lint` 0 errors, 55 passed across the commenter/reconciler/helper/badge suites. Census: batch-core **72 → 70**.

gsxdsm:
All five findings were real. Fixed in `4eab22b23`.

**The Major one is the one worth reading.** `landedColumnsForTask` and its siblings took `store: Pick<TaskStore, "getTask">` and cast it through `unknown` to reach `resolveWorkflowIrForTask`. Two things wrong with that: `getTask` is never called by any of these helpers, and the shape they actually need is the resolver-store interface (`getTaskWorkflowSelection` + `getWorkflowDefinition`).

The cast was a type lie in the load-bearing direction. A caller could pass a partial store with no workflow readers; every call would throw straight into the `catch` and the helper would return the legacy answer **forever**, silently, with the conversion looking done. Typed properly, that same caller is now a compile error at the call site instead of a permanent runtime fallback nobody would notice.

**Coverage.** Correct: only the `catch` path was tested. The resolved-but-**empty** branch — a v1 upgrade where the IR resolves fine and `columnsWithFlag(ir, "complete")` returns `[]` — is a different code path, and without a case for it the `length > 0 ? … : legacy` compatibility branch could regress to returning an empty set, refusing every v1 board, while the suite stayed green. Added and mutation-verified: returning the empty set as-is fails it.

**The stale claim.** Also correct, and it was stale because of a correction made in this same PR. The shared-helper note still said the source-issue commenters share the "landed" decision; their contract is now complete-only, precisely so a rename does not start posting comments and closing issues on archival. Reworded to state which caller takes which helper and why, with the regression that prompted the split named.

**The dates.** Every FNXC stamp I wrote on this branch read `2026-07-31`; today is `2026-07-30`. Corrected across all 21 touched files. This was not a one-off typo — I had been stamping tomorrow's date consistently, so both timestamp findings were pointing at the same systematic error and the sweep was the right fix rather than patching the two cited lines.

**Verification:** `tsc --noEmit` 0 errors in both `@fusion/dashboard` and `@fusion/core`; `pnpm lint` 0 errors; 44 passed across the helper, commenter and planner suites.

gsxdsm:

| site | supplied |
|---|---|

gsxdsm:

gsxdsm:
I traced this one transition at a time and **could not reproduce a lost comment**. Both services resolve ALL complete lanes the same way, so the secondary lane is not special to either. But the trace turned up a real adjacent gap, which I think is what the finding is sensing.

### Why no comment is lost

Both sides classify with `columnsWithFlag(ir, "complete")` and test membership — `github-tracking-state.ts:222-226` uses `completeLanes.includes(columnId)`, not a first-lane comparison. So a secondary complete lane is a complete lane to the tracking commenter too.

Walking every move into a secondary complete lane, `before` = the literal `to === "done"`, `after` = my `completeColumnsForTask(...).has(to)`:

| from → to | tracking | before | after |
|---|---|---|---|
| wip → secondary complete | on | no source comment; tracking closes+comments | source defers; tracking closes+comments — **same** |
| wip → secondary complete | off | no comment at all | **source comments** — the fix working |
| primary complete → secondary complete | on | no comment (`to !== "done"`) | no comment (`decideIssueAction` returns null: `toClass.complete && !fromClass.complete` is false) — **same** |
| primary complete → secondary complete | off | no comment | no comment — **same** |

The only observable change is that the suppression *log line* now fires where the listener used to return before reaching it. No path loses a comment it previously posted.

### The real gap this is adjacent to — pre-existing, not from this PR

The two services disagree about what an **empty** resolved complete set means, and on a v1-upgraded board that disagreement silences the pair.

- `github-tracking-state.ts:216-219` deliberately uses the empty set **as-is**: *"A resolved IR that declares no complete lane is an ANSWER — moving a card somewhere is not 'completing' it on a board with no completion lane."*
- `completeColumnsForTask` treats empty as **unexpressed** and falls back to `done`.

That reasoning is correct for a genuine v2 board that declares no completion lane. It is wrong for a **v1 upgrade**, where `synthesizeDefaultColumns` emits every column with `traits: []` — the board plainly has a `done` column holding finished cards, and the empty set means the traits were never expressed, not that completion is undefined.

Consequence on a v1-upgraded board with GitHub tracking enabled and pointed at the source issue: a task moving to `done` has `completeLanes = []`, so `toClass.complete` is false, `decideIssueAction` returns null, and the tracking commenter never closes the issue — while the source commenter defers to it. **Neither posts.**

This predates my change (the old literal `to === "done"` also reached the same suppression), so it is not a regression from this PR and I have not fixed it here. But it is the same "empty means unexpressed, not absent" distinction that #2775's review caught in `fn pr create`, and it is now the only place in the dashboard that still resolves it the other way. Worth a follow-up on `github-tracking-state.ts` rather than a change to this commenter — flagging rather than folding an unrelated behaviour change into this PR.

gsxdsm:
### Pre-flight against current `main`: introduces nothing

The branch is 11 commits behind `main`, so I ran the suites on a local merge with `origin/main` (`e467d939a5`) rather than the branch head.

**`@fusion/core`** — clean:
```
Test Files  438 passed (438)
     Tests  4751 passed | 1 skipped (4752)
```

**dashboard `api` group** (this batch touches `packages/dashboard/src` substantially — routes, `server.ts`, the new `task-lifecycle-lanes.ts`, `mockCoreEngine.ts`):
- `api:curated` — pass
- `api:backfill 2/2` — 1175 passed
- `api:backfill 1/2` — **1 failed** / 1485 passed

That one failure is **pre-existing, not yours**:

```
src/__tests__/planning-browser-e2e.test.ts
  > Planning Mode browser E2E
  > places the sole contextual comment trigger by viewport in embedded and modal Planning
```

Verified by running the same shard on `origin/main` with this branch out of the picture — **identical single failure** (`1 failed | 1473 passed`). The passed-count difference (1473 vs 1485) is just the tests this branch adds. It is already tracked as part of the dashboard tally in #2784.

So: no core regression, no api regression attributable to this batch.

Worth running the merged state rather than the branch head on these — the last batch-engine landing put 32 failures on main that were only caught after the merge, and on a stale branch the two are easy to confuse.

gsxdsm:
**A second scan, and it is the more productive one: seams where SOME callers pass the flags and others do not.** The gate I added only requires *one* supplier, so these are invisible to it — and they are where the real defects were.

That gap is not theoretical. Review caught `isTaskStuck` in my package: `TaskCard` supplied it, so the gate was green, while `deriveStatsFromTasks` still passed three arguments and the footer counted **zero stuck tasks** on a renamed board.

Scanning for the split found **5 seams, 18 omitting call sites**. Seven were mine and are fixed. **Six are yours, all on one seam:**

```text
isNearDuplicateCanonicalInactive(canonical, columnFlags?)   — core/src/near-duplicate-canonical.ts
  OMITS  packages/core/src/task-store/branch-group-ops.ts:44
  OMITS  packages/engine/src/self-healing.ts:6351
  OMITS  packages/engine/src/self-healing.ts:12451
  OMITS  packages/engine/src/triage.ts:3593
  OMITS  packages/engine/src/triage.ts:3749
  OMITS  packages/engine/src/triage.ts:4092
```

**What it costs:** without the flags, `isActiveNearDuplicateColumn` falls to the legacy ids, so on a renamed board a canonical resting in an **active** lane reads as *inactive*. In the dashboard that silenced a duplicate warning; in triage and self-healing it feeds duplicate-decision and reconciliation logic, so the consequence there is yours to judge — I have not traced those paths and am not guessing at them.

**Two caveats on the tooling, both learned by getting them wrong:**

1. **Call sites are matched by function NAME, not resolved symbol.** `sortTasksForDisplayColumn` exists in *both* `core/task-priority.ts` (takes flags) and `dashboard/components/taskSorting.ts` (takes a sort mode). The scan conflated them, I "fixed" callers of the wrong one, and `tsc` rejected it. **Treat the report as a pointer to investigate, never a diff to apply.**
2. `0/N` rows mean no call sites were found at all — the scan excludes `__tests__`, so test-only exports look unsupplied.

Scanner is ~40 lines; say the word and I will hand it over or land it as a second `scripts/check-*.mjs`. I did **not** gate this one, because "all callers must supply" is not obviously correct — some callers legitimately should not resolve flags — so it wants a human read rather than a blocking check.

gsxdsm:
**Correction to the list I sent you — I generated it with a flawed scanner and some of it was wrong. Here is the verified version.**

Two bugs in my tooling since that report, both found by probing it rather than reading it:

1. **A prefilter was skipping call sites.** I only parsed files mentioning a flag name — but a caller that *omits* the argument mentions no flag name, so the scan skipped exactly the files containing the omissions. It saw only the callers already passing the argument. That is also why my gate never caught the `isTaskStuck` misses in my own package; review did.
2. **Same-named functions were conflated.** A local function sharing a seam's name had its calls counted against the seam — which can make a genuinely under-supplied seam read as supplied.

**The verified list for your packages — one seam, and it is real** (single definition, confirmed by hand):

```text
isNearDuplicateCanonicalInactive(canonical, columnFlags?)   core/src/near-duplicate-canonical.ts
  core/src/task-store/branch-group-ops.ts:44
  engine/src/self-healing.ts:6415
  engine/src/self-healing.ts:12515
  engine/src/triage.ts:3599
  engine/src/triage.ts:3755
  engine/src/triage.ts:4102
```

Without the flags, `isActiveNearDuplicateColumn` falls to legacy ids, so on a renamed board a canonical in an **active** lane reads as *inactive*. In the dashboard that silenced a duplicate warning. What it means for duplicate-decision and reconciliation in triage/self-healing is yours to judge — I have not traced those paths.

**Withdraw these rows from my earlier message** — they are false positives from the name conflation:

- `sortTasksForDisplayColumn` (Lane.tsx, ListView.tsx) — those files import a **different** function of that name from `dashboard/components/taskSorting`. I acted on this row once and `tsc` rejected the change.
- `resolveEffectiveExecutor` / `resolveEffectiveValidator` (ModelSelectorTab) — that file defines its own private 2-arg versions.

**Three documented ways the check can mislead**, all from matching by name: it proves only that *one* caller supplies (not all); it excludes `__tests__`, so test-only exports look unsupplied; and it cannot see imported shadows. All three are now in the script header.

Sorry for the churn. I would rather send you a shorter true list than a longer one you have to audit.

gsxdsm:
Second verified inert seam in `packages/core/src/task-priority.ts`, from the same gate that reported `getTotalAgentActiveMs` — and this one was hidden from that earlier report, so it is new information rather than a repeat.

**`sortTasksForDisplayColumn(tasks, column, columnFlags)` — `columnFlags` is supplied by nothing in production.**

The only callers passing a third argument are `core/src/__tests__/display-ranking-roles-resolved.test.ts`. The three dashboard call sites — `Lane.tsx:107`, `Board.tsx:641`/`838`, `ListView.tsx:1169` — bind to a **different function of the same name** in `packages/dashboard/app/components/taskSorting.ts`, which they import explicitly. They are not callers of the core function at all.

So the conversion is inert in the exact way the FNXC note at the declaration warns about: with the flags omitted the role helpers fall back to legacy ids, meaning the three defects that note enumerates (hold lane loses priority ordering, complete lane loses recency ordering, review lane misordered) are all still live on a renamed board, while the census scores the seam as converted.

**Why it was invisible until now:** my seam check matched call sites by name, so the dashboard function — called with up to five arguments — raised the arg-count max and cleared core’s seam. I had recorded that collision as a known false-positive source, which is the bad half: the row read as noise, and the genuine offender sat behind it. Fixed on `batch-dashboard-app` (`scripts/check-inert-flag-seams.mjs` now records the module each callee is imported from). Measured with the fix reverted: the scan prints `17 seams, all supplied` and emits no row for this function; with it, the function is reported.

**Not touching it from outside** — it is yours, and the fix is a judgment call I should not make for you: either thread the resolved flags through the two production paths, or drop the parameter and let the literal stay counted. The second is a legitimate outcome; an honest literal beats a seam that reads converted.

Carrying a TEMPORARY allow-list entry meanwhile so the gate stays green and self-announcing. Its staleness check fails the moment the site becomes supplied, so the entry cannot outlive your fix — delete it in the same change and the gate starts guarding the file.

Correction to my earlier cross-batch note on this PR, for the record: the verified core list is `getTotalAgentActiveMs` and now `sortTasksForDisplayColumn`. Nothing else.

gsxdsm:
Third verified core finding, and a different shape from the first two — this seam is *partially* supplied, which is why nothing had reported it.

**`isNearDuplicateCanonicalInactive(canonical, columnFlags)` — omitted at `packages/core/src/task-store/branch-group-ops.ts:44`, supplied at its five siblings.**

`clearNearDuplicateReferencesToImpl` calls it with `inactiveState` alone, so on a renamed board this is the single path that answers from legacy ids while every other caller resolves the real column. The failure is one-sided in the dangerous direction: a canonical resting in a renamed active column reads as *inactive*, so the near-duplicate references get cleared against a canonical that is still live.

Unlike the other two I have sent you, this one looks straightforwardly wireable rather than a judgment call — the function is already `async` and already holds `store` and `canonicalId`, so the canonical's own flags are one `resolveTaskLifecycleColumns` await away. I have not made the change; it is your file and the await placement inside that impl is yours to pick.

**Why it was invisible:** my seam check tested `best < arity` — whether SOME caller supplies. Five did, so the seam read as satisfied. That is the same floor that let `isTaskStuck` ship with two of three call sites omitting the flags (review caught that one, not the gate). The check now reports per-call-site omissions; measured by dropping the argument at a supplied site and watching the exact row appear, then restoring it.

Carrying a TEMPORARY per-call-site exemption meanwhile — keyed by `<file>::<function>` rather than by name, specifically so your other five sites stay guarded rather than the whole seam being waived. Its staleness check fails the moment the site supplies, so it cannot outlive your fix.

Running list for core, all verified individually: `getTotalAgentActiveMs`, `sortTasksForDisplayColumn`, and this.

gsxdsm:
Fourth core finding, and this one is a different category — not an inert conversion but a conversion applied to **dead code**, with a possible product gap behind it. Flagging rather than guessing, because the second half is yours to answer.

**`evaluateMergeBlockerGuard` (`packages/core/src/default-workflow-hooks.ts:55`) has no callers at all.**

Its only reference in the repo is its own declaration. Not called, not re-exported from `index.ts` or `index.gate.ts`, not mentioned in docs. I originally had it allow-listed as "exported for tests only; the scanner excludes `__tests__`" — that reason was **false**. The scan now reads test files, and there are zero call sites there either.

So its `lifecycleColumns` conversion (FNXC 2026-07-31-04:20) went onto a function that never runs, and the note attached to it describes the review→complete crossing "this guard fires on" — a crossing it cannot make. The census counted it.

**The part I am not asserting.** Two things nearby look off, and I am deliberately not drawing the conclusion:

1. `registerDefaultWorkflowHooks()` registers `timing`, `reset-on-entry`, `abort-on-exit`, and `merge` hooks — but no `"guard"` hook. The only `registerTraitHookImpl(..., "guard", ...)` anywhere is in `trait-registry.test.ts:206`.
2. The file header credits the guard to an `evaluateDefaultWorkflowGuards` reader, which does not exist.

`getTaskMergeBlocker` is clearly live and called from `task-merge.ts`, so merge blocking is not simply absent — which is exactly why I am not claiming a missing move-time guard. It could be that the guard moved and this function plus the header comment are both residue. It could also be that a registration was dropped. Distinguishing those needs the intent behind KTD-9 and the trait-hook cutover, which you hold and I do not.

If it is residue, deleting the function removes a counted conversion and the census baseline drops honestly. Carrying a TEMPORARY exemption meanwhile with the accurate reason.

Running list for core, each verified individually: `getTotalAgentActiveMs`, `sortTasksForDisplayColumn`, `isNearDuplicateCanonicalInactive` (partial, `branch-group-ops.ts`), and this. Every one surfaced only after closing a specific blind spot in the check — which is the part I would take seriously for the ~240 guards still outstanding.

gsxdsm:
Correcting my earlier pre-flight comment on this PR: **I reported it clean and that was incomplete.** Merged, it put 5 failures on `main`.

Neither is a defect in the change — both are its bookkeeping half:

1. **The census baseline (4 failures).** This PR annotated six documented fast-path literals with `DELIBERATE-LITERAL`, and the census tracks marked literals per file, so `--strict` reported *"column-guard count ROSE"*. The rise is legitimate; the baseline just needed re-recording in the same commit.
2. **`executor-user-cancel` (1 failure).** The disposer refactor added await hops, and that test drained exactly two microtask turns, so it failed with `expected -1 to be greater than 2`. Verified the re-dispatch is **not** dropped — order is still `["abort-started","abort-resolved","dispose","execute"]`, reached once drained and unchanged after a real timer.

Both fixed forward in **#2814**.

**Why my pre-flight missed it:** I ran `@fusion/core` and the dashboard `api` group, because that is what this diff touches. The census and disposal tests live in `packages/engine` — a package this PR does not modify at all. The suite that breaks is not always the suite the diff points at, and a cross-package ratchet like the census is precisely where scoping to changed packages yields a confident "clean" that is wrong. I have widened pre-flight to always include the engine suite for the rest of the queue.

gsxdsm:
Fifth core finding, and the sharpest consequence of the set — this one **throws** rather than degrading quietly.

**`enqueueMergeQueueInTransaction(tx, taskId, opts, audit, reviewColumns)` — `reviewColumns` omitted at `async-merge-coordination.ts:246`.**

The two `moves.ts` callers (487, 1153) supply it. The omitting caller is the **public `enqueueMergeQueue` wrapper** in the same file, which passes four arguments.

So the guard at line 175 falls back to `new Set(["in-review"])`, and on a board whose review lane is named anything else it takes the reject branch: it records `mergeQueue:enqueue-rejected` with `reason: "not-in-review"` and throws `MergeQueueInvalidColumnError`. That is a hard failure of the manual re-enqueue path for a task sitting in its board's own review column — while the automatic handoff-to-review path, which supplies, works fine. The file's own header calls out that the handoff path MUST use the in-transaction helper to preserve VAL-DATA-013, so the two paths diverging on column vocabulary seems worth a look beyond the one-line supply.

Verified by reading both call sites and the guard body, not by scanner output alone.

**Why it was invisible:** partial supply. Two of three call sites passed the argument, so a check asking "does SOME caller supply this" read the seam as satisfied. That floor is now closed and reports per call site.

TEMPORARY per-site exemption carried meanwhile, keyed `<file>::<function>` so your two correct `moves.ts` sites stay guarded.

Running list for core, each verified individually: `getTotalAgentActiveMs` (**now fixed on main — my exemption's staleness check caught it and I removed the entry**), `sortTasksForDisplayColumn`, `isNearDuplicateCanonicalInactive` (partial), `evaluateMergeBlockerGuard` (dead code), and this.

PR #2785, opened 2026-07-30
batch-engine tail: re-land the ASYNC half; the sync-resolved half was inert (engine −15)

gsxdsm:
Tail of `batch-engine` (#2773). That PR merged as a squash while later engine work was still in flight, so `self-healing.ts`, `executor.ts` and `worktree-pool.ts` landed at their pre-conversion counts. This re-lands **only the half that is real**, and the reason the other half is not here is the substance of this PR.

## Census, per file (measured, `--strict` verified)

| file | main | here |
| --- | ---: | ---: |
| `engine/src/self-healing.ts` | 107 | 97 |
| `engine/src/executor.ts` | 15 | 12 |
| `engine/src/worktree-pool.ts` | 3 | 2 |
| `engine/src/ephemeral-worker-manager.ts` | 1 | 0 |
| `engine/src/agent-tools.ts` | 5 | **0** |
| `engine/src/gridlock-detector.ts` | 3 | **0** |
| `engine/src/triage.ts` | 4 | 1 |
| `engine/src/mission-execution-loop.ts` | 2 | **0** |
| **net** | | **−28** |

Baseline re-recorded; `--strict` tightened exactly these 4 entries and no others.

## Finding: a whole class of conversions in this program is INERT, and the census scores it as progress

`resolveTaskWorkflowIrSync` returns the **default** workflow IR for every task in production. The sync selection reader behind it is a PostgreSQL-cutover stub:

```ts
// packages/core/src/task-store/workflow-definitions.ts:505
export function getTaskWorkflowSelectionImpl(_store, _taskId) {
  return undefined;   // "Backend mode cannot synchronously read PostgreSQL"
}
```

So a guard written as `resolveLifecycleColumns(store.resolveTaskWorkflowIrSync(id))?.hold` resolves an IR, asks for a trait, and answers **from the default workflow for every custom board** — silently. It reads as converted and the census counts it as converted. `main` gained `sync-workflow-ir-callsite-allowlist.test.ts` for exactly this after my branch point; it is what caught me.

I had built three sync resolvers on that reader — `resolveMoveLanesSync` (self-healing, executor) and a widened `resolveTaskParkedColumnsSync` (scheduler) — reasoning that a *synchronous* `task:moved` listener needs a *synchronous* reader. That reasoning was sound about the shape and never checked whether the reader reads anything.

**Dropped from this PR, deliberately, and NOT re-landed anywhere:**

- `scheduler.ts` 12 → 1 (the widening; the pre-existing narrow helper on main is untouched)
- the executor `task:moved` handler, incl. the Move-Task hard-cancel lane comparison
- self-healing's `task:moved` fan-out, `classifyPausedAbortWorkflowRecovery`, `reconcileInReviewBranchRebind`, `recoverWedgedActiveMerge`, `recoverPausedAbortFailures`, and 12 single-row lane conversions

Those sites are back to their literals. The allow-list's own guidance is the standard I applied:

> An unconverted `=== "todo"` is strictly better, because it is at least honest about being a literal.

I did not add my call sites to the allow-list. Six entries would have turned the gate green in two minutes and buried the defect; the list's contract requires proving the async resolver is genuinely unreachable, and for a fire-and-forget listener it is not — the listener can `void` an async lane resolution the same way `NotificationService` already does. That is the correct fix and it is a behaviour-shaped change, so it is out of scope here.

**Fleet-wide consequence:** any conversion routed through `resolveTaskWorkflowIrSync` is fake progress, and the census cannot see the difference. `pnpm test:gate` can: the allow-list test is the detector. Its passing here (161/161) is this PR's evidence that nothing inert survived the split.

## What IS in this PR — all async-resolved

1. **`self-healing.clearStaleBlockedBy`** — lanes resolved per **REFERENCED** task, not per iterated task. A blocker's own workflow decides whether it is still blocking.
2. **`executor` dependency satisfaction** — resolved per **DEPENDENCY** via `columnsWithFlag`. Preserves the load-bearing asymmetry that a dependency in *review* already satisfies a dependent; a bulk sweep flattens that to complete-only and deadlocks the board.
3. **`agent-tools` — the agent task tools listed FINISHED cards as active.** `fn_task_list` says it lists "tasks that aren't done or archived"; `fn_task_search` offers `includeDone: false`. Both filtered on `task.column !== "done"`, so a renamed complete lane returned finished cards as outstanding work **to an agent**, which then reasons and acts on them. `includeArchived` was always enforced by the QUERY and survived a rename; `"done"` was only ever a TS predicate, which is why exactly that half broke.

   Plus the two **dedup** guards in the same file. The cross-parent diagnostic filter kept a *shipped* card as a candidate on a renamed board, so the guard adopted it as canonical and returned `wasDuplicate: true` — absorbing new diagnostic work into a task nobody is working on (the eval-followup defect shape again). The defined-feature bootstrap preflight is **not** the query-filter class: its query passes `includeArchived: true`, so the TS predicate is the *only* archived guard there; on a renamed archive lane the archived sibling became the bootstrap canonical and `claimDefinedFeatureTask` then rejects the non-live row, so a valid first task fails to be created at all.

   Both dedup invariants **already had tests** — asserted against the legacy ids only, so both passed for the very comparison being replaced. Extended in place into vocabulary differentials rather than added as parallel files. Two helpers rather than one parameterised one: "is this finished?" and "is this archived?" are different questions, and merging them would make the archived-only guard also reject completed rows.

   The list/search half re-landed **with the test it originally shipped without.** No suite exercised either tool, so the original commit's "304/304 green" said nothing about the change — the optional-flags failure mode exactly. Both call sites are covered; converting two copies and testing one is the Surface Enumeration failure this program has already hit twice.

4. **`gridlock-detector` — FALSE dependency alarms.** The gate compared each blocker against `done`/`in-review`/`archived`; on a renamed board all three are true for a *finished* blocker, so no dependency ever counted as met and the detector reported dependency gridlock for tasks that are not blocked — `notifyGridlock` then pages the operator. Resolved per dependency using the **same five flags** as the executor's gate (`complete`, `archived`, `mergeOrchestration`, `mergeBlocker`, `humanReview`) — `review` is not a trait, and two gates answering "is this dependency satisfied?" differently is a split brain. Every pre-existing case in that file omits a workflow, so none could detect the change; added the renamed case plus a non-vacuous companion.

5. **`triage` — its OWN copies of the same two tools.** `createTriageTools` carries a `fn_task_list` and `fn_task_search` byte-identical in intent to the agent-tools pair, plus a third site filtering duplicate candidates. Same defect on all three. Reused the (now exported) agent-tools helper rather than adding a third copy — deliberately stronger than the two-parallel-tests reading of Surface Enumeration, since the copies now share one implementation and cannot drift. **Not claiming call-site coverage:** `createTriageTools` is private and not drivable without standing up a TriageAgent; the helper is revert-proofed, those two call sites are covered only through it.

6. **`mission-execution-loop` — a finished fix task read as LIVE, stalling remediation.** The comment above that line states the rule it implements: *only an open task makes duplicate triage safe to suppress.* On a renamed board the rule inverts — a finished fix task is not `done`/`archived`, so it reads as live, remediation for a fresh validation failure is suppressed indefinitely, and the mission stalls with no error surfaced.

   **Not revert-proven, and I am not claiming it is.** No test reaches the `hasLiveFixTask` branch, and the only case that mints a fix feature is git-gated and heavyweight; building that fixture is larger than the conversion. The change strictly *widens* the finished set (resolved roles ∪ the two legacy ids), so default boards are byte-identical — that is the argument for shipping it unproven, not a substitute for coverage.

7. **Four census-invisible membership guards**, each inverted on a renamed board — `worktree-pool` (merger-managed branch reclaim could delete a branch out from under an in-flight merge), `agent-assignment` (assignment load counted nothing), `ephemeral-worker-manager` (`isAgentIdle` inverted on both sides), and the dead constants their conversion orphaned. These are `SET.has(task.column)` shapes the census does not count, so the −15 understates them.

## Revert results (measured, each run)

| conversion | reverted → |
| --- | --- |
| `clearStaleBlockedBy` per-referenced lanes | renamed-vocabulary case fails; stale `blockedBy` never clears |
| executor dependency satisfaction | dependent never unblocks on a renamed review lane |
| `worktree-pool` merger-managed set | reclaim proceeds against an in-flight merge |
| `ephemeral-worker-manager.isAgentIdle` | idle agent reads busy on a renamed board |
| `fn_task_list` terminal filter | RENAMED case fails — shipped card listed as active |
| `fn_task_search` terminal filter | RENAMED case fails — same, independently |
| cross-parent diagnostic dedup | RENAMED case fails — `wasDuplicate: true`, new work absorbed |
| bootstrap preflight archived guard | RENAMED case fails — `validate` called with the archived sibling |
| gridlock dependency gate | RENAMED case fails — false gridlock raised for an unblocked task |

`agent-assignment`'s widened `taskStore` type is compile-time; its revert is a tsc failure, not a test failure — stated rather than claimed as coverage.

## Verification

- `pnpm test:gate` — 161 + 487 + 13 + 71, all green (161 includes `sync-workflow-ir-callsite-allowlist`)
- `pnpm lint` — clean

One commit is a pure import restore: `columnsWithFlag` arrived in a sibling commit that built on the inert resolver and was left behind. The engine tsconfig excludes `src/__tests__/**`, so the gate was green while tsc was not — worth knowing that on this package a green gate is not a green build.

## Verified NOT a gap — measured, so the next worker does not re-open them

- **`restart-recovery-coordinator` (5 counted).** Four already take an optional `reviewColumns` set and the counted literals are the documented **fallback** arm, which must stay for the same reason `columnRoles.ts` keeps its id fallback. The sole production caller (`self-healing.ts:12151-12154`) already passes the resolved set. The fifth is documented at the site as a re-assertion behind a `listTasks({ column: "in-progress" })` query filter. Nothing to convert.
- **`notification/notification-service` (5 counted).** Already documented in-file as deliberately counted with no exemption marker: the wedge-episode site needs per-task serialisation of wedge handling (a delivery-semantics change to operator notifications), and `isManualMergeHold` needs a pre-resolved `LifecycleColumns` threaded through `handleTaskUpdated`, which would pay resolution on every task update. Both are behaviour/placement judgements, not conversions.
- **`planner-overseer` (3 counted).** `resolveWatchedStage`'s two literals are fed by `pollPlannerOverseer`, which calls `listTasks({ column: "in-progress" })` and `{ column: "in-review" }` — hardcoded **query** filters. On a renamed board those queries return no rows, so the predicate never sees a renamed column. Converting it alone would drop 3 from the census and change nothing an operator can observe. The real fix is at the query layer; that is the tracked query-filter-bounded class, not this PR.
- **`triage:695`** reads `resolvePlannerLanes` → the allow-listed sync IR reader. Left as an honest literal per the rule above.

**Still open in `packages/engine`, deliberately not in this PR:** `self-healing.ts` (97, of which ~31 are the query-filter-bounded class and the rest need per-site classification in a 13k-line file), `scheduler.ts` (12, blocked on the sync reader above), `executor.ts` (12), and a tail of ~13 more copies of the "is this task finished?" question across eight small files (`agent-reflection`, `auto-merge-finalization`, `merger-scope-auto-widen`, `backlog-pressure-reporter`, `merger-orphan-rehome`, `merger-integration-worktree`, `plugin-runner`, `cli-agent/*`). That tail is a clean follow-up: one question, eight call sites, and the exported `resolveTerminalColumnsForTasks` helper already exists for it.

That is the same discipline as the sync-resolver finding: a census number that drops without a behaviour change is not progress, and four of these files would have handed over exactly that.


gsxdsm:

| site | supplied |
|---|---|

gsxdsm:

gsxdsm:
### Pre-flight against current `main`: clean

Ran the full `@fusion/engine` suite on a local merge of this branch with `origin/main` (`e467d939a5`), since the branch is 6 commits behind:

```
Test Files  844 passed | 1 skipped (845)
     Tests  10903 passed | 14 skipped | 1 todo (10918)
```

**Zero failures.** This batch does not introduce any engine red.

One thing to ignore if you run the branch as-is: **the branch head alone reports 3 failures** in `executor-prompt.test.ts` ("global pause behavior"). Those are not yours — they are the pre-#2779 state of a file this branch never touches (`git diff` against main shows no change to it), and #2779 landed the fix after this branch was cut. They disappear on rebase, as the merged run above shows.

Flagging because the last batch-engine landing put 32 failures on main that nobody caught until after the merge; running the merged state rather than the branch head is what distinguishes "stale" from "broken" here.

gsxdsm:
**A second scan, and it is the more productive one: seams where SOME callers pass the flags and others do not.** The gate I added only requires *one* supplier, so these are invisible to it — and they are where the real defects were.

That gap is not theoretical. Review caught `isTaskStuck` in my package: `TaskCard` supplied it, so the gate was green, while `deriveStatsFromTasks` still passed three arguments and the footer counted **zero stuck tasks** on a renamed board.

Scanning for the split found **5 seams, 18 omitting call sites**. Seven were mine and are fixed. **Six are yours, all on one seam:**

```text
isNearDuplicateCanonicalInactive(canonical, columnFlags?)   — core/src/near-duplicate-canonical.ts
  OMITS  packages/core/src/task-store/branch-group-ops.ts:44
  OMITS  packages/engine/src/self-healing.ts:6351
  OMITS  packages/engine/src/self-healing.ts:12451
  OMITS  packages/engine/src/triage.ts:3593
  OMITS  packages/engine/src/triage.ts:3749
  OMITS  packages/engine/src/triage.ts:4092
```

**What it costs:** without the flags, `isActiveNearDuplicateColumn` falls to the legacy ids, so on a renamed board a canonical resting in an **active** lane reads as *inactive*. In the dashboard that silenced a duplicate warning; in triage and self-healing it feeds duplicate-decision and reconciliation logic, so the consequence there is yours to judge — I have not traced those paths and am not guessing at them.

**Two caveats on the tooling, both learned by getting them wrong:**

1. **Call sites are matched by function NAME, not resolved symbol.** `sortTasksForDisplayColumn` exists in *both* `core/task-priority.ts` (takes flags) and `dashboard/components/taskSorting.ts` (takes a sort mode). The scan conflated them, I "fixed" callers of the wrong one, and `tsc` rejected it. **Treat the report as a pointer to investigate, never a diff to apply.**
2. `0/N` rows mean no call sites were found at all — the scan excludes `__tests__`, so test-only exports look unsupplied.

Scanner is ~40 lines; say the word and I will hand it over or land it as a second `scripts/check-*.mjs`. I did **not** gate this one, because "all callers must supply" is not obviously correct — some callers legitimately should not resolve flags — so it wants a human read rather than a blocking check.

gsxdsm:
**Correction to the list I sent you — I generated it with a flawed scanner and some of it was wrong. Here is the verified version.**

Two bugs in my tooling since that report, both found by probing it rather than reading it:

1. **A prefilter was skipping call sites.** I only parsed files mentioning a flag name — but a caller that *omits* the argument mentions no flag name, so the scan skipped exactly the files containing the omissions. It saw only the callers already passing the argument. That is also why my gate never caught the `isTaskStuck` misses in my own package; review did.
2. **Same-named functions were conflated.** A local function sharing a seam's name had its calls counted against the seam — which can make a genuinely under-supplied seam read as supplied.

**The verified list for your packages — one seam, and it is real** (single definition, confirmed by hand):

```text
isNearDuplicateCanonicalInactive(canonical, columnFlags?)   core/src/near-duplicate-canonical.ts
  core/src/task-store/branch-group-ops.ts:44
  engine/src/self-healing.ts:6415
  engine/src/self-healing.ts:12515
  engine/src/triage.ts:3599
  engine/src/triage.ts:3755
  engine/src/triage.ts:4102
```

Without the flags, `isActiveNearDuplicateColumn` falls to legacy ids, so on a renamed board a canonical in an **active** lane reads as *inactive*. In the dashboard that silenced a duplicate warning. What it means for duplicate-decision and reconciliation in triage/self-healing is yours to judge — I have not traced those paths.

**Withdraw these rows from my earlier message** — they are false positives from the name conflation:

- `sortTasksForDisplayColumn` (Lane.tsx, ListView.tsx) — those files import a **different** function of that name from `dashboard/components/taskSorting`. I acted on this row once and `tsc` rejected the change.
- `resolveEffectiveExecutor` / `resolveEffectiveValidator` (ModelSelectorTab) — that file defines its own private 2-arg versions.

**Three documented ways the check can mislead**, all from matching by name: it proves only that *one* caller supplies (not all); it excludes `__tests__`, so test-only exports look unsupplied; and it cannot see imported shadows. All three are now in the script header.

Sorry for the churn. I would rather send you a shorter true list than a longer one you have to audit.

PR #2787, opened 2026-07-30
fix: five lifecycle gates the census cannot see — incl. live ephemeral workers reaped and duplicate follow-up cards

gsxdsm:
Five lifecycle-column fixes the census **structurally cannot see**. Each gate is a `Set` or array literal — a *definition*, not a comparison — so no backlog entry ever pointed at any of these files. Found by grepping for lane-shaped list literals after the same shape surfaced in `duplicate-intake` and `blocker-fanout` (both merged via #2780), then confirmed by reading each USE site.

**On opening this:** I offered twice to fold these into a PR and kept them on handoff refs to respect one-open-PR-per-worker. They have now sat unadopted across several cycles while `main` moved, and two of them destroy or duplicate work. Opening is the reversible call — **close it if it breaks queue policy** and I will keep them on the branch.

## What is in it

| commit | defect on a renamed board | severity |
|---|---|---|
| `beb107a7bc` | assignment load-balancing **defeated** — `assignmentLoad` stays empty, every candidate reads as load 0, the sort falls through to its stable `createdAt` tiebreak, so **one agent wins every assignment** while the rest idle | distribution |
| `cf4b59e1cb` | the zombie sweep **deletes LIVE ephemeral workers** | **destroys work** |
| `5fe004ae64` | eval follow-up dedup sees **zero open tasks**, so every run re-files follow-ups it already filed | **duplicate cards** |
| `a1021de8b2` | agents keep a **"working on" indicator for finished cards** | stale UI |
| `86680d1220` | the **Files tab never loads** — the fetch never fires | silent empty |

### The one that destroys work

`shouldDeleteOnSweep` tested a hard-coded terminal `Set`, then fell through to `return task.column !== "in-progress"`. On a renamed board **both halves miss, and they compound in the worst order**: the terminal test fails, control reaches the fallthrough, and `"building" !== "in-progress"` is `true`. An ephemeral worker **actively executing a task** is classified as a zombie and deleted. Nothing logs.

Its fallback is **deliberately asymmetric**, and the comment says why: an unresolvable workflow keeps the legacy literals rather than guessing. Failing to reap a dead worker costs a slot; reaping a live one destroys work in flight. Those are not symmetric, so uncertainty fails toward keeping the worker.

## Verification

Verified **as a set**, not only per-branch:

- `pnpm test:gate` — **161 / 13 / 487 / 71**
- engine suites (assignment, ephemeral, eval-followups) — **44 passed**
- dashboard suites (agent-task-link, useSessionFiles) — **16 passed**
- `tsc` engine + dashboard server + dashboard app — clean
- `pnpm lint` clean · census `--strict` exits 0

**Revert-proven individually.** Restoring each literal fails its own case: the renamed-wip zombie case, the renamed-wip assignment case, the renamed-lane dedup case, the sanitizer ratchet, and both `useSessionFiles` role cases.

## Two honesty notes, flagged rather than buried

**`a1021de8b2`'s guard is STRUCTURAL, not behavioural.** `sanitizeAgentTaskLinks` is a closure inside `createApiRoutes`, reachable only by standing up the full express app. The ratchet asserts the source — resolver threaded per task, bare literal call gone, cache shared, fallback retained — and **fails on revert**, verified. It is not a substitute for a behavioural test; whoever owns the dashboard server should add one if that seam grows.

**`useSessionFiles`'s negative case passed in isolation and failed in the suite.** Hooks are not unmounted between cases there, so a prior case's in-flight fetch landed inside it. That is the classic shape of a test that gets "fixed" by reordering; it now asserts a **delta** against the pre-render call count, which is independent of what leaks in.

## Deliberately NOT included

`worktree-pool.ts:1205` — the sixth site from the same sweep. It **fails safe**: a missed match means the skip does not fire, so the branch is added to `activeBranches` and *protected* from cleanup. The cost is stale branches accumulating, not deletion. It also sits in the merger's branch-reaping path, where the opposite error destroys work, so it deserves its owner's judgement rather than a drive-by conversion. Flagged, not guessed.

Also still open and unclaimed: roughly 69 untriaged literal-list sites across engine/dashboard/cli. The grep is one line and the file list is on #2775 — with the measured caveat that about half are false positives on shape alone (`LEGACY_*` names, seeds unioned with resolved values, and `roles: ["triage"]`, which is an `AgentCapability`, not the deleted column). Only the use site settles it.

### DIFF
--- a/packages/dashboard/app/hooks/__tests__/useSessionFiles.test.ts
+++ b/packages/dashboard/app/hooks/__tests__/useSessionFiles.test.ts
@@ -121,3 +121,73 @@ describe("useSessionFiles", () => {
     });
   });
 });
+
+/*
+FNXC:WorkflowLifecycleColumns 2026-07-31-07:30 (dashboard-app feed):
+
+THE INVARIANT: the Files tab loads for any column whose ROLE carries a worktree.
+
+CENSUS-INVISIBLE. The gate was a `Set` literal — a definition, not a comparison — so nothing in the
+lifecycle backlog pointed at this hook. Found by grepping for lane-shaped list literals after the
+same shape turned up in `duplicate-intake`, `blocker-fanout` and the ephemeral zombie sweep.
+
+On a renamed board the set matched nothing, so the fetch NEVER FIRED and the Files tab was
+permanently empty for every card that had one. An empty file list is indistinguishable from a task
+that touched no files, which is why nobody would report it.
+
+The flags are collapsed to a boolean before the effect on purpose: `columnFlags` is an object, and a
+caller constructing it inline would hand this hook a fresh identity every render, so putting it in
+the dep array would refetch on every parent render. That is asserted below — a re-render with an
+equal-but-not-identical flags object must not trigger a second fetch.
+
+REVERT PROOF, measured: restore `ACTIVE_COLUMNS.has(column)` and the renamed-wip case fails with
+zero fetches.
+*/
+describe("useSessionFiles resolves the column's role", () => {
+  const WIP_FLAGS = { countsTowardWip: true } as never;
+  const HOLD_FLAGS = { hold: true } as never;
+
+  it("fetches for a RENAMED wip lane", async () => {
+    mockFetchSessionFiles.mockResolvedValueOnce(["src/a.ts"]);
+
+    const { result } = renderHook(() =>
+      useSessionFiles("FN-123", "/repo/.worktrees/kb-123", "building", undefined, { columnFlags: WIP_FLAGS }));
+
+    await waitFor(() => expect(result.current.files).toEqual(["src/a.ts"]));
+  });
+
+  it("does NOT fetch for a column whose role carries no worktree", async () => {
+    /*
+    The gate must still gate — a hold-lane card has no worktree to read.
+
+    Asserted as a DELTA rather than "not called at all": this file's hooks are not unmounted between
+    cases, so a prior case's in-flight fetch can land inside this one. The absolute assertion passed
+    in isolation and failed in the suite, which is the classic shape of a test that would have been
+    "fixed" by reordering rather than by being made independent.
+    */
+    const before = mockFetchSessionFiles.mock.calls.length;
+
+    renderHook(() =>
+      useSessionFiles("FN-123", "/repo/.worktrees/kb-123", "backlog", undefined, { columnFlags: HOLD_FLAGS }));
+
+    await waitFor(() => expect(mockFetchSessionFiles.mock.calls.length).toBe(before));
+  });
+
+  it("does not refetch when an equal-but-new flags object arrives", async () => {
+    // The object-identity trap the derived boolean exists to avoid.
+    mockFetchSessionFiles.mockResolvedValue(["src/a.ts"]);
+
+    const { rerender, result } = renderHook(
+      ({ flags }) => useSessionFiles("FN-123", "/repo/.worktrees/kb-123", "building", undefined, { columnFlags: flags }),
+      { initialProps: { flags: { countsTowardWip: true } as never } },
+    );
+
+    await waitFor(() => expect(result.current.files).toEqual(["src/a.ts"]));
+    const callsAfterFirst = mockFetchSessionFiles.mock.calls.length;
+
+    rerender({ flags: { countsTowardWip: true } as never });
+
+    await waitFor(() => expect(result.current.loading).toBe(false));
+    expect(mockFetchSessionFiles.mock.calls.length).toBe(callsAfterFirst);
+  });
+});
--- a/packages/dashboard/app/hooks/useSessionFiles.ts
+++ b/packages/dashboard/app/hooks/useSessionFiles.ts
@@ -1,7 +1,35 @@
 import { useEffect, useState } from "react";
 import { fetchSessionFiles } from "../api";
+import {
+  isCompleteColumnRole,
+  isReviewColumnRole,
+  isWipColumnRole,
+  type ColumnRoleFlags,
+} from "../utils/columnRoles";
 
-const ACTIVE_COLUMNS = new Set(["in-progress", "in-review", "done"]);
+/*
+FNXC:WorkflowLifecycleColumns 2026-07-31-07:30 (dashboard-app feed):
+Session files load for cards that HAVE a worktree — wip, review, and complete.
+
+CENSUS-INVISIBLE: a `Set` literal is a definition, not a comparison, so nothing in the lifecycle
+backlog pointed at this hook. Found by grepping for lane-shaped list literals.
+
+On a renamed board the set matched nothing, so the Files tab was permanently EMPTY for every card
+that had one — the fetch simply never fired. An empty file list is indistinguishable from a task
+that touched no files, which is why this would never be reported as a bug.
+
+DELIBERATE-LITERAL — the unresolved-flags default, reviewed 2026-07-31-07:30. `columnFlags` is
+optional so every existing caller (and the hook's own test) is byte-identical; the role helpers in
+`columnRoles.ts` own the legacy-id degraded mode.
+*/
+const LEGACY_ACTIVE_COLUMNS = new Set(["in-progress", "in-review", "done"]);
+
+function hasWorktreeBearingRole(column: string, flags: ColumnRoleFlags | undefined): boolean {
+  if (!flags) return LEGACY_ACTIVE_COLUMNS.has(column);
+  return isWipColumnRole(flags, column)
+    || isReviewColumnRole(flags, column)
+    || isCompleteColumnRole(flags, column);
+}
 
 interface UseSessionFilesResult {
   files: string[];
@@ -11,6 +39,8 @@ interface UseSessionFilesResult {
 interface UseSessionFilesOptions {
   /** Enable fetching when true (default). Suppresses fetches for offscreen cards. */
   enabled?: boolean;
+  /** Resolved trait flags for `column`. Omitted → the legacy ids, i.e. today's behaviour. */
+  columnFlags?: ColumnRoleFlags;
 }
 
 /**
@@ -30,6 +60,13 @@ export function useSessionFiles(
   options: UseSessionFilesOptions = {},
 ): UseSessionFilesResult {
   const enabled = options.enabled ?? true;
+  /*
+  FNXC:WorkflowLifecycleColumns 2026-07-31-07:30 (dashboard-app feed):
+  Collapsed to a BOOLEAN before the effect, deliberately. `columnFlags` is an object, and a caller
+  building it inline hands this hook a new identity every render — putting it in the dep array below
+  would refetch on every parent render. The derived boolean is stable for a stable answer.
+  */
+  const columnBearsWorktree = hasWorktreeBearingRole(column, options.columnFlags);
   const [files, setFiles] = useState<string[]>([]);
   const [loading, setLoading] = useState(false);
 
@@ -41,7 +78,7 @@ export function useSessionFiles(
       return;
     }
 
-    if (!taskId || !worktree || !ACTIVE_COLUMNS.has(column)) {
+    if (!taskId || !worktree || !columnBearsWorktree) {
       setFiles([]);
       setLoading(false);
       return;
@@ -68,7 +105,7 @@ export function useSessionFiles(
     }
 
     void load();
-  }, [taskId, worktree, column, projectId, enabled]);
+  }, [taskId, worktree, column, projectId, enabled, columnBearsWorktree]);
 
   return { files, loading };
 }
--- a/packages/dashboard/src/__tests__/agent-task-link-terminal-lanes.test.ts
+++ b/packages/dashboard/src/__tests__/agent-task-link-terminal-lanes.test.ts
@@ -1,0 +1,52 @@
+/*
+FNXC:WorkflowLifecycleColumns 2026-07-31-07:00 (dashboard-server feed):
+
+THE INVARIANT: the agent "working on" sanitizer asks each linked task's OWN terminal lanes.
+
+`sanitizeAgentTaskLinks` drops `taskId` from an agent response when the linked task is finished, so
+the UI stops showing a stale "working on" indicator. It tested a hard-coded `Set(["done","archived"])`
+— CENSUS-INVISIBLE, because a Set literal is a definition rather than a comparison, so nothing in the
+lifecycle backlog pointed at this file. On a renamed board it matched nothing and a FINISHED card kept
+its agent's indicator lit: the agent list advertised work that had already shipped, which is exactly
+what this sanitizer exists to prevent.
+
+WHY THIS GUARD IS STRUCTURAL AND NOT BEHAVIOURAL, stated plainly rather than dressed up:
+`sanitizeAgentTaskLinks` is a closure inside `createApiRoutes`, reachable only by building the full
+express app and driving `GET /api/agents` through it. That harness exists (see
+`routes-automation.test.ts`) but standing it up to re-assert a per-task resolver is a large amount of
+machinery around a small seam, and I did not write it. So this asserts the SOURCE: the resolver is
+threaded per task and the bare literal call is gone.
+
+It fails on revert — verified by reverting — which is the bar. It is not a substitute for a
+behavioural test, and whoever owns the dashboard server should add one through the express harness if
+this seam grows. Flagging that rather than letting a structural check read as full coverage.
+*/
+import { readFileSync } from "node:fs";
+import { describe, expect, it } from "vitest";
+
+const source = readFileSync(new URL("../routes.ts", import.meta.url), "utf8");
+
+describe("agent task-link sanitizer resolves each task's own terminal lanes", () => {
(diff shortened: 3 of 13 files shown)


gsxdsm:
## Two more from the same sweep, one of them a guard that violated its own comment

`handoff/stale-spec-and-project-health-lanes` (`663b3d6dbc`), on current `main`. Gate **161 / 13 / 487 / 71**, lint clean, `--strict` exits 0, baseline re-recorded. Not folded into #2787 — that PR is already under review and this is new work, so it stays separately reviewable.

### `executor.ts` — the stale-spec guard replanned LIVE tasks

The comment directly above the code is explicit:

> Skip for tasks that are already in-progress, in-review, merging, or done — **these should not be interrupted and sent back to triage for re-planning.**

Keyed on a hard-coded `Set`, a renamed board matched nothing, so `isActiveTask` was false for a card in a renamed wip/review/complete lane. The guard then ran on a **live** task and `moveTaskToReplanColumn` + `status: "needs-replan"` pulled it out of execution mid-flight — precisely the thing the comment forbids.

`activeMergeStatuses` still covered the merging states, so a *merging* card was protected **by accident** while a plain in-progress card was not. That is why the symptom looks arbitrary from outside: some active cards survive, some get yanked.

The fix **unions** rather than replaces, and that distinction is load-bearing: `resolveWorkflowIrForTask` returns the BUILT-IN IR on a missing or corrupt definition instead of throwing, so replacing the legacy trio would silently drop `in-progress` in exactly the degraded case hardest to notice.

### `register-project-routes.ts` — project health reported zero active tasks

`activeCols` matched nothing on a renamed board, so project health showed **0 active tasks** beside an in-flight agent count that was still correct — the same silent-zero shape as the analytics tallies in #2780. A zero next to a populated neighbour reads as "this project is idle", not as "this number is broken".

Its four entries also still listed `triage`, a column U11 deleted, so one of them had been dead on every board since that cutover.

## The mistake I made twice, now written into the test

My first draft of the executor test added four "behavioural" cases that built the union themselves from `resolveLifecycleColumns` and asserted membership. Measured against a revert, **only the structural case failed** — the other four passed with the fix removed, because they exercised a MIRROR of the guard rather than the guard. They were a test of core's `resolveLifecycleColumns`, which already has its own coverage.

That is the same mirrored-implementation trap I caught and deleted in `analytics-timing-roles-resolved.test.ts` earlier in this program. **Twice is a pattern**, so the rule is now written at the top of the file:

> If a test constructs the value the product is supposed to construct, reverting the product cannot fail it.

What ships is a source ratchet in the `engine-no-blocking-shellout` shape — verified to fail on revert (3 of 4 cases), and labelled as structural rather than behavioural. The guard sits deep inside `execute()`, behind worktree and session setup a unit test has no business standing up; whoever next touches that scaffolding should add the end-to-end case.

## Sweep tally, honest

Of the six candidates read this round: **two real** (above), **one false positive** (`self-healing.ts:3004` is a legacy SEED unioned with resolved columns three lines later), **one belongs to the tracked QUERY class** (`self-healing.ts:6081` passes `["triage","todo"]` straight to `listTasks({ column })` — and `triage` no longer exists, so that half of the sweep has been querying a dead column), and **two already handled** (`worktree-pool` fails safe; `register-project-routes:952` was already a resolved-or-legacy fallback).

Running precision on shape alone is now roughly **50%** across ~10 sites read. The grep finds the class; only the use site settles it.

gsxdsm:
## Dashboard-app sweep: 1 real of 4 read — `handoff/github-tracking-editable-lanes` (`6daa3b2bf2`)

Gate **161 / 13 / 487 / 71**, lint clean, `--strict` exits 0, baseline re-recorded.

### The real one, and it was in a HALF-CONVERTED file

`GITHUB_TRACKING_EDITABLE_COLUMNS` is a `Set` literal sitting **twenty lines below**
`isTaskFieldEditableColumn`, which had already been moved onto the shared role helper. One converted
predicate and one unconverted one, in the same file, answering adjacent questions about the same card.

On a renamed board the set matched nothing, so the GitHub-tracking toggle was **disabled on every
card**, with nothing on screen to explain why. That is the identical failure the comment above
`isTaskFieldEditableColumn` already records for inline editing — the same bug, one function over,
surviving the pass that fixed its neighbour because a Set literal is a definition rather than a
comparison and the census never listed it.

Converted to "editable until the card's own workflow calls it FINISHED", with the literal set kept as
the fallback for the two real unresolved states (before the board-workflows payload lands, and for a
column the workflow does not declare — unknown, not known-false). The Coding (Ideas) workflow-id
short-circuit predates this and is asserted to survive it.

**Behavioural, not structural, deliberately.** I have leaned on source ratchets three times in this
sweep where the seam was genuinely unreachable. This predicate is small and pure, so it is exported
and driven directly — a ratchet here would have been laziness rather than necessity. 6 cases; restoring
the literal fails the two renamed-lane ones.

### The other three were not defects

- `DocumentsView.tsx:110` — the `if (flags) { … }` block above it already handles the resolved case; the Set is the documented no-flags fallback.
- `MissionControlPanel.tsx:71` — carries its own `DELIBERATE-LITERAL` block; already reviewed.
- `SystemStatsArea.tsx:34` — chart axis labels used only when `byColumn` is empty. Cosmetic default for an empty chart, not a gate.

### Sweep precision, final

Across ~14 sites now read use-site-first: **6 real defects, 8 false positives** — about **43%**. The grep reliably finds the CLASS; it cannot distinguish a gate from a documented fallback, a seed unioned with resolved values, or a foreign vocabulary (`roles: ["triage"]` is an `AgentCapability`). Every one of the six was confirmed by reading the use site, and four were confirmed by a test that failed first.

That number is the useful output of this sweep. Anyone continuing it should budget for reading roughly two sites per defect found, and should not treat the remaining ~55 unread sites as a defect count.

gsxdsm:
## All four review findings addressed (`efda5ccf88`) — and the P1 was the exact failure I have spent this session documenting

Gate **161 / 13 / 487 / 71**, lint clean, `--strict` exits 0. Engine suites 44 → **46 passed**, dashboard ratchet 4 → **5 passed**.

### P1 `agent-assignment` — "production load lanes stay legacy". Correct, and it is the worst version of being wrong.

I added the optional `activeColumns` parameter and **never wired the only production caller**. So the scheduler kept passing nothing, the predicate kept its legacy default, and on a renamed board the load tally stayed empty — the fix was **inert in production** while the file read as converted.

That is precisely the guard-that-cannot-fire pattern I flagged on #2759 and again in my own PR body two comments ago. Shipping it would have been *worse than leaving the literal*, because a converted site stops attracting attention. `scheduler.ts:2299` now resolves and passes the lanes.

### P1 ×2 — first-per-role vs MEMBERSHIP

`resolveLifecycleColumns` returns the FIRST column carrying each trait. Both my ephemeral and dashboard conversions used it, so:

- a board with **two wip lanes** still reaped a live worker in the second — the same defect the commit exists to fix, one degree narrower;
- a board with **two complete/archived lanes** still showed an agent working on finished work in the second.

Both now union `columnsWithFlag(...)`. This is the same finding #2739's review made against `branch-group-ops`, which means it is a **recurring trap in this program, not a one-off**: `resolveLifecycleColumns` is the right helper for "where do I MOVE a card", and the wrong one for "is this card IN a role". A regression case for the second wip lane is added and revert-proven.

### P2 `eval-followups` — sequential resolution over the whole board

Correct: the loop ran `resolveTerminalColumnsFor` per card with no cache, so a large board paid a store read per task before a single draft was processed. `resolveTerminalColumnsFor` now takes an optional caller-owned `irCache` (matching the contract on `resolveTaskLifecycleColumns`), and the loop shares one — **one IR read per WORKFLOW instead of per card**. Callers resolving a single task pass nothing and are unaffected.

### Note on the ratchet test

The dashboard guard asserted the old `resolveTaskLifecycleColumns` call shape, so the membership fix **failed it** — which is the ratchet working. Updated to pin the union explicitly, so a future narrowing back to first-per-role fails rather than passes.

---

Separately: **#2775's red Lint is fixed** (`edf18361c2`). It was a stale census baseline — `efefe3d2ed` added two `DELIBERATE-LITERAL` markers to `even-cards/board-cards.ts` without re-recording, so the tree and baseline disagreed about whether those two sites were column guards or reviewed literals. Re-recorded; `--strict` exits 0 and the gate is green.

gsxdsm:
All four findings are addressed by `efda5ccf8` — I had fixed them independently before that landed, discarded my version rather than force-pushing over it, and verified yours instead. Three are fully closed. The `agent-assignment` one has **two residual gaps** that the wiring introduced, both measurable.

### Verified closed

- **`ephemeral-worker-manager.ts`** — `wipLanes` is now a `Set` from `columnsWithFlag(ir, "countsTowardWip")`, and the `SECOND wip lane` regression case is present. This was the destructive one: the old `!== lanes.wip` deleted a live task's ephemeral worktree whenever a board declared more than one WIP-bearing column. My independent fix and test were byte-equivalent in effect.
- **`routes.ts`** — every `complete`/`archived` column now contributes to the terminal set.
- **`eval-followups.ts`** — the shared IR cache is threaded, so a board scan costs one read per distinct workflow rather than per task.

### Gap 1 — the load set is resolved from ONE task's workflow, then applied to ALL tasks

`scheduler.ts:2312` resolves `loadLaneIr` from **`freshTask`** and passes the resulting set to `selectPermanentAgentForTask`, where `agent-assignment.ts:102` applies it to every task in `allTasks`:

```
if (!taskItem.assignedAgentId || !(activeColumns ?? LEGACY_ACTIVE_COLUMNS).has(taskItem.column)) continue;
```

On a project where tasks run different workflows, a task whose own board names its WIP lane differently is measured against the wrong set and contributes **zero** load. That is the original bug surviving for exactly the tasks the conversion was meant to reach — just narrowed from "all boards" to "boards other than the one being assigned".

The load question is genuinely per task, which a single set cannot express. `selectPermanentAgentForTask` already receives `taskStore`, so it can resolve inside the loop with a shared IR cache and keep `activeColumns` as an explicit override for callers that have already resolved it. That also removes the inert-injection shape entirely rather than relocating it to the caller.

### Gap 2 — `intake`/`hold` were silently dropped from "active load"

`LEGACY_ACTIVE_COLUMNS` is `["todo", "in-progress", "in-review"]` — **`todo` counted toward an agent's load**. The resolved set is `countsTowardWip ∪ mergeOrchestration ∪ mergeBlocker ∪ humanReview`, with no `intake` or `hold`.

So on the DEFAULT board, where the conversion should be a no-op, an agent's queued (`todo`) assignments no longer count as load. An agent holding several queued tasks now reads as idler than it did before and attracts more work. That is a behaviour change riding inside a vocabulary conversion, and it is invisible to renamed-board fixtures because it shows up precisely where the legacy names still apply — the same trap that caught my source-issue commenters in #2783.

Adding `intake` and `hold` to the resolved set restores parity.

### Note on stamps

Several FNXC entries on this branch are dated `2026-07-31` and `2026-08-*`; today is `2026-07-30`. I hit the same thing across 21 files in #2783 — worth a sweep rather than per-line edits, since it tends to be a habit rather than a typo.

gsxdsm:
## I audited my own merged conversions for the defect greptile caught. Five more are inert, and one had a second implementation I never touched.

The #2787 P1 was *"the optional parameter exists and no production caller passes it"*. That is a class, not an incident, so I ran it against everything I have landed this session. **Measured, by counting call sites that actually pass the argument:**

| conversion | production callers passing the resolved answer |
|---|---|
| `blocker-fanout.escalationColumns` | **0** |
| `team-analytics` / `workflow-analytics` `columnFlagsByName` | **0** |
| `assigned-task-ranking.flagsByColumnId` | **0** |
| `duplicate-intake.columnFlagsByColumnId` | **0** |
| `isLegacyAutoMergeStampCandidate(reviewColumns)` | **0** (3 core call sites, none pass it) |
| `getTaskCompletionBlocker.satisfactionColumnsByTaskId` | **1** — already wired by whoever took #2780's review |
| `getTaskMergeBlocker.reviewColumns` | 2 |

So **five of my merged conversions changed the census and not the behaviour.** They are not wrong — the fallbacks are correct and documented — but every one of them reads as done and runs as the literal. I claimed several of these fixed operator-visible defects; for those five that claim was premature, and this comment is the correction.

The pattern is worth naming for the whole program: **a conversion is finished when a production caller passes the resolved answer, not when the census count drops.** The census cannot see the difference, and neither can a test that injects the value by hand.

## The worse variant, fixed here (`9dbd461912`)

`getTotalAgentActiveMs` was converted in `@fusion/core`. The card chip does not use it — **`packages/dashboard/app/utils/taskTiming.ts` has its own copy**, and I never touched it. So the census counted core's site as converted while the number an operator actually reads stayed keyed on `"in-progress"`.

Two implementations of one calculation in two packages, one converted and one not. That is the drift `column-roles.ts` was created to end, and it is invisible to a per-file census by construction.

On a renamed board the card chip dropped the live execution segment, under-reporting the run in flight by exactly its elapsed time — then healed itself once the card moved on and the segment persisted. The app copy is now converted onto `isWipColumnRole`, and `TaskCard` threads its already-resolved `taskColumnFlags` through `getTaskEndToEndDurationMs` so the path is live rather than optional-and-unused.

4 tests; restoring the literal fails the two renamed-lane cases. Gate **161 / 13 / 487 / 71**, lint clean, `--strict` exits 0.

## What I am NOT doing with the other five

Wiring them properly is real work — each needs its caller to resolve an IR, a cache where the caller is a sweep, and a test that drives the caller rather than the predicate. Doing five badly in one pass is how the inert versions got shipped in the first place. They are listed above so the next person has the measurement rather than the impression, and I would take `blocker-fanout.escalationColumns` first: it gates escalation of a stale blocker holding up many cards, which is the only one of the five with a queue-level consequence.

gsxdsm:
## P1 (second round) fixed — and it is the sharper version of the first one

Correct, and it is a regression I introduced **while fixing the original P1**. The legacy set is `{todo, in-progress, in-review}` and `todo` is the HOLD/INTAKE lane. My resolved union covered only wip and review, and because passing the argument **overrides** the fallback rather than extending it, assigned backlog work stopped counting as load — worse than the inert version for that one lane.

The general rule, now written at both the call site and the test: **with an override-shaped option, the resolved answer must be a superset of what the literal answered.** Otherwise wiring the parameter is a downgrade for whichever roles it forgot. Cheap to get wrong, and invisible to a test that only checks the renamed lane — which is exactly what mine did.

Union now covers intake, hold, wip, mergeOrchestration, mergeBlocker and humanReview. Two guards added: a behavioural case asserting a card in a renamed HOLD lane still counts as load, and a drift check that fails if `scheduler.ts` stops unioning any of the six roles. Revert-proven: dropping intake/hold fails the drift check.

Gate **161 / 13 / 487 / 71**, lint clean, `--strict` exits 0.

gsxdsm:
## First of the five inert conversions wired — and wiring it found a bigger defect than the one I set out to fix

`handoff/fanout-escalation-wired` (`6f8b295ed4`), on current `main`. Gate **161 / 13 / 487 / 71**, lint clean, `--strict` exits 0.

### What I expected to fix

`computeBlockerFanoutMap`'s `escalationColumns` had **zero** production callers passing it, so on a renamed board a long-standing bottleneck was reported as `temporary` forever. Annoying, cosmetic-ish.

### What was actually wrong

Wiring it, the test still logged nothing at all — and that is the real finding. `overlapBlockedTodoCount` counts dependents in the **hold** lane, whose flat default is `"todo"`. On a renamed board that count is **ZERO**, so the threshold check `continue`s past every blocker and **no overlap-bottleneck warning is emitted at any point**.

So the operator-visible defect was not a mislabelled warning. It was **no warning**: a card blocking five others for six hours produced silence, and fixing only `escalationColumns` would have corrected the label on a message that never printed. I would not have found it by reading the code — the test failing on an empty log is what exposed it.

The wiring therefore uses **`classify`**, the per-task option this module documents as *"the only correct option on a multi-workflow board"*, for the hold/terminal halves, plus the flat `escalationColumns` for the active-lane half.

### Why the flat set is acceptable for the escalation half, stated rather than assumed

`escalationColumns` is board-wide, so on a board spanning workflows it unions every workflow's active lanes; an id one workflow calls wip is then treated as active for all. That over-approximates in the **cheap** direction — a bottleneck warning reading `long-lived` slightly too eagerly, never a missed one — and it matches how `terminalColumns`/`reviewColumns` are already passed here. The hold half gets `classify` precisely because its error direction is not cheap: it silences the warning entirely.

### A fixture bug I made and am recording

My legacy-fallback case left the dependents in `backlog` while asserting legacy behaviour, so the count was zero and the case asserted a **missing** message rather than a legacy one. It passed for the wrong reason. Had the expectation been negative rather than positive, it would have been green and worthless — the same shape as the `createdAt` epoch-ms and `not.toHaveBeenCalled()` mistakes earlier in this sweep. **The fixture has to match the branch under test**; that is now three instances of the same error from me in one program.

3 tests; reverting the wiring fails 2 of 3.

### Four inert conversions left

`team-analytics` / `workflow-analytics`, `assigned-task-ranking`, `duplicate-intake`, `isLegacyAutoMergeStampCandidate` — all still **0 production callers**, all listed with measurements on this PR. Given what wiring the first one turned up, I would treat the remaining four as *unaudited* rather than *cosmetic*: the parameter being unwired may be hiding a second, larger defect in the caller, exactly as it did here.

PR #2794, opened 2026-07-30
test(engine): bound the inert-sync-resolution class on a live store

gsxdsm:
## What

One new live-PostgreSQL E2E suite, 3 tests. **No production file is touched** — evidence, per the E2E worker's remit. Closes the series: #2789 (scheduler), #2791 (planner lanes), #2792 (custom fields), #2793 (terminal node).

`packages/engine/src/__tests__/workflow-sync-selection-blast-radius-live-e2e.pg.test.ts`

## Why this one is different

The four PRs above each proved a site broken because it resolved a task's workflow synchronously. Read together they invite a conclusion that is **false and would be expensive**: that every synchronous consumer of the workflow selection is inert.

Most are not. The difference is one line of shape:

```ts
// GUARDED (correct)
store.getTaskWorkflowSelectionAsync
  ? await store.getTaskWorkflowSelectionAsync(id)
  : store.getTaskWorkflowSelection(id)

// UNGUARDED (inert)
store.resolveTaskWorkflowIrSync(id)
```

The real PostgreSQL store **does** implement the async reader, so every guarded site takes the async arm and resolves the card's own workflow. Only the sync IR helper — which has no async arm to fall to — is stuck with the default.

Observed on one live store, one persisted workflow, one task:

```
hasAsyncReader   = function
SYNC  selection  = undefined
ASYNC selection  = { workflowId: "WF-001", stepIds: [] }
EFFECTIVE planReviewMaxRevisions = 9   <- the custom workflow's declared default
```

## The point

"The ternary saves them" is an inference from reading, and the whole premise of this program is that reading is what let the class survive in the first place. The guarded sites are exactly the ones a fleet worker would otherwise "fix": converting a correct site costs review time, risks behaviour, and produces a diff that looks like progress. This makes the bound checkable in the same lane as the defects.

Guarded call sites (correct today): `workflow-settings-resolver.ts`, `workflow-ir-resolver.ts`, `executor.ts`, `workflow-graph-task-runner.ts`, `workflow-task-runtime.ts`, and `board-workflows.ts` in the dashboard.

## The allow-listed family is now closed

| site | status |
|---|---|
| `scheduler.ts` | proven broken — #2789 |
| `replan-target.ts` | proven broken — #2791 |
| `task-store-helpers.ts` | proven broken — #2792 |
| `branch-and-pr-entities.ts` | proven broken — #2793 |
| `workflow-task-create-ops.ts` | **legitimately correct** — creation runs before any selection exists, so the default IR is the right answer |
| `lifecycle-ops.ts` | **NOT proven, stated as such** |

`lifecycle-ops.ts`'s stale-transition-pending recovery re-runs plugin column-transition hooks against the sync IR. Driving it needs a registered plugin hook plus a crash-simulated marker; I did not build that harness and I am not substituting a unit test for it. Named in the file so it is not mistaken for covered.

## Evidence discipline

- **Observed state.** Both readers called on one live store against one persisted workflow, plus a real resolved settings value — not a spy on which arm ran.
- **The settings default is `9`**, deliberately not the builtin's, so the value can only have come from this workflow.
- **Mutation-verified.** Rewriting the guarded consumer to call the sync reader directly fails **exactly** the bound arm; the two structural arms are correctly unaffected, which is what a bound should do.

## Verification

- new suite — **3/3 passed**, mutation-verified
- full live-PG E2E surface — **136/136 passed** (133 on main + 3; #2791 landed while this branch was in flight)
- `pnpm lint` — clean

### DIFF
--- a/packages/engine/src/__tests__/workflow-sync-selection-blast-radius-live-e2e.pg.test.ts
+++ b/packages/engine/src/__tests__/workflow-sync-selection-blast-radius-live-e2e.pg.test.ts
@@ -1,0 +1,135 @@
+/*
+FNXC:WorkflowLifecycleColumns 2026-07-31-01:30 (E2E evidence — BOUNDING the inert-sync-resolution class):
+
+Four PRs in this series (#2789 scheduler, #2791 planner lanes, #2792 custom fields, #2793 terminal
+node) each proved a site broken because it resolved a task's workflow synchronously. Read together
+they invite a conclusion that is FALSE and would be expensive: that every synchronous consumer of the
+workflow selection is inert.
+
+Most are not, and this file is the bound. The difference is one line of shape:
+
+    GUARDED (correct)     store.getTaskWorkflowSelectionAsync
+                            ? await store.getTaskWorkflowSelectionAsync(id)
+                            : store.getTaskWorkflowSelection(id)
+
+    UNGUARDED (inert)     store.resolveTaskWorkflowIrSync(id)
+
+The real PostgreSQL store DOES implement the async reader, so every guarded site takes the async arm
+and resolves the card's own workflow. Only the sync IR helper — which has no async arm to fall to —
+is stuck with the default. Five call sites use the guarded shape (`workflow-settings-resolver.ts`,
+`workflow-ir-resolver.ts`, `executor.ts`, `workflow-graph-task-runner.ts`, `workflow-task-runtime.ts`,
+plus `board-workflows.ts` in the dashboard); six use the unguarded one and are the allow-list.
+
+WHY THIS IS WORTH A FILE RATHER THAN A SENTENCE. "The ternary saves them" is an inference from
+reading, and the whole point of this program is that reading is what let the class survive. The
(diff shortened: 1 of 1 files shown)


gsxdsm:
Fixed in the latest push — the finding is right.

`expect(syncColumns).not.toContain(RENAMED_VOCAB.hold)` is satisfied by an empty, malformed, or entirely unrelated IR — anything that merely lacks `backlog`. It could have stayed green while establishing nothing about the default-board behaviour the case is named for, which is the precise failure mode this whole series is about.

Now asserted positively:

```ts
expect(syncColumns).toContain("todo");
expect(syncColumns).toContain("in-progress");
expect(syncColumns).not.toContain(RENAMED_VOCAB.hold);
```

The claim is now "it answered with the DEFAULT board", not the much weaker "it did not answer with this one". The two ids are written literally rather than imported, so the case fails loudly if the post-U11 default lineage changes instead of silently widening. Rationale recorded in an FNXC note at the assertion.

3/3 still green.

PR #2796, opened 2026-07-30
fix(engine): assignment load must be resolved per task — #2787 P1 follow-up

gsxdsm:
Fix-forward for the P1 that arrived on **#2787 after it merged** — so it lands as its own PR rather than a thread reply on merged code.

## The finding

`selectPermanentAgentForTask`'s `activeColumns` was resolved from the **candidate** task's workflow and then applied to every row `listTasks` returned. On a project running several workflows — the normal case — assignments living in another workflow's load-bearing lanes vanished from the tally, and the already-loaded-agent-wins bug returned through a different door.

**A column id means something only relative to its OWN workflow.** `blocker-fanout.ts` documents exactly this and offers a per-task `classify`; the option is now that same shape rather than a third invention:

```ts
countsAsAssignmentLoad?: (task: Task) => boolean
```

The scheduler resolves each assigned row against its own IR, sharing one cache for the selection, so a board spanning three workflows reads three IRs — not one per assigned card.

## Why this is the third round on the same parameter, stated plainly

1. I added the parameter and **never wired the caller** — inert in production.
2. I wired it as a **union of wip+review**, which dropped hold/intake and made it a *regression* for backlog work.
3. I resolved it from **one workflow** and applied it to all — this fix.

Each round was a smaller version of the same error: treating a lane answer as global when it is per-task, and per-role when it is per-membership. Worth recording because the first two rounds both looked correct and both passed their tests — the tests asserted the renamed case I was thinking about, not the shape of the data.

## Verification

- new cross-workflow case; reverting the predicate to a single workflow's lanes **fails it**
- `agent-assignment` suite **14 passed**
- `pnpm test:gate` — **161 / 13 / 487 / 71** · lint clean · census `--strict` exits 0


gsxdsm:
## Second inert conversion wired — and the seam has a limit I am not going to paper over

`handoff/analytics-column-flags-wired`, on current `main`. Gate **161 / 13 / 487 / 71**, lint clean, `--strict` exits 0.

`columnFlagsByName` existed on both analytics aggregators and **neither route passed it**, so the Command Center reported **0 in-progress and 0 in-review** on a renamed board beside token and cost totals that were entirely correct. A zero next to a populated neighbour reads as "nobody is working", not as "this metric is broken". Both routes now build the flag map from the project's own workflow definitions.

### The limit, stated rather than glossed

The map is **flat** — one entry per column id, unioned across every workflow the project declares. Where two workflows use the same id for different roles, the merged flags make that column count for both.

For **team** analytics that ambiguity **cannot be resolved at this seam**: the rows arrive already aggregated to `{agentId, columnName}` by SQL, so workflow identity is gone before this code sees the data. Fixing it properly means grouping by workflow in the query — a change to the aggregation, not to the wiring.

For **workflow** analytics it *is* resolvable: those rows carry `workflowId`, so a per-workflow map would be exact. I did not take it here because it needs a second option shape on the aggregator, and bundling an API change into a wiring fix is how the last three rounds of review on `agent-assignment` happened. It is the obvious follow-up and I would rather name it than quietly ship the approximation as if it were complete.

This is the fourth time in this sweep a flat lane answer has turned out to be the wrong shape (`agent-assignment` ×2, `blocker-fanout`'s hold half, and now this). The pattern is consistent enough to state as a rule: **a lane answer is per-task unless the data has already lost task identity — and if it has, that is the bug, not the workaround.**

### On the guard

Structural, and labelled as such. The aggregators take a live `AsyncDataLayer`, so a behavioural test means standing up PostgreSQL to assert that an option is forwarded; the forwarded VALUE is already covered behaviourally in core's `analytics-timing-roles-resolved.test.ts`. What was missing was the wiring, and that is what this pins — including a case asserting **both** routes forward it, since wiring one and not the other is the half-converted-pair shape this program keeps finding.

Revert proof: dropping either `columnFlagsByName:` line fails it.

### Three inert conversions left

`assigned-task-ranking`, `duplicate-intake`, `isLegacyAutoMergeStampCandidate` — still **0 production callers**. Both wired so far turned up something larger than the label bug I expected (`blocker-fanout`: no warning at all; this one: a seam that cannot be made exact), so I would keep treating the remaining three as unaudited rather than cosmetic.

gsxdsm:
Real, and the divergence goes **both** ways — fixed in `deb8133cb`.

The scheduler built its Set of load-bearing task IDs from its own `listTasks` read; `selectPermanentAgentForTask` then applies the predicate to rows from **its** read. Anything changing in between diverges:

- a task **moved out** of a load-bearing lane keeps its id in the set and is still counted → the agent looks busier than it is;
- a task **created or newly assigned** in between is missing from the set and counts as **zero** → the agent looks idler than it is.

Both feed the balancer a board that no longer exists, and the second is the same "counts as zero load" failure this PR exists to fix, reintroduced through a different door.

**The fix memoises the resolved lanes, not the verdict.** Lane membership is a property of the task's *workflow*, which a move does not change — so the map can be built from either snapshot and the comparison still happens against `candidate.column` on the row the helper actually holds. Only the workflow lookup is cached; the column test is live. That removes the snapshot dependency rather than trying to keep two reads in agreement.

A task absent from the map (created between the reads) falls back to the same legacy trio the resolver itself uses for an unresolvable workflow, rather than silently counting as zero — the failure direction that is invisible in production.

**On the test, stated plainly:** it pins the *contract the fix depends on* — that the helper passes its own live rows to the predicate — not an end-to-end reproduction of the race. The race lives inside a dispatch path this suite cannot stand up, which is the same limitation `scheduler-load-lane-union.test.ts` records about its own call site. If the helper ever pre-resolved or cached rows, the scheduler's fix would silently revert to answering about a stale board, and this test is what fails. Mutation-verified: a helper passing a non-live row fails 4 of 15.

I deliberately did not mirror the scheduler's union in the test file the way the sibling suite does. For this defect a mirror is useless — the bug is about *which snapshot the column came from*, which a copy of the lane logic cannot express.

`tsc --noEmit` 0 errors, `pnpm lint` 0 errors, gate green (161 + 487 + 13 + 71), `agent-assignment` 15 passed.

gsxdsm:
## Third inert conversion wired — and for the third time the CALLER held the larger defect

`handoff/legacy-stamp-review-lanes`, on current `main`. Gate **161 / 13 / 487 / 71**, lint clean, `--strict` exits 0.

### Wiring the parameter would have changed nothing

`isLegacyAutoMergeStampCandidate` gained an optional resolved `reviewColumns` and no caller passed it. But the read above it was:

```ts
const inReview = await store.listTasks({ column: "in-review" });
```

A **QUERY filter** with the literal. On a renamed board that returns zero rows, so the backfill iterated an empty list and reported success over nothing — **the predicate was never reached**. Passing the resolved set into it would have been a no-op on an empty array.

That is now three for three:

| unwired parameter | what the caller was actually doing |
|---|---|
| `blocker-fanout.escalationColumns` | the hold default made the count zero — **no warning emitted at all** |
| analytics `columnFlagsByName` | routes never built a map — **silent zero beside correct cost totals** |
| `isLegacyAutoMergeStampCandidate` | **queried a column the board does not have** |

The rule I would hand to anyone finishing this program: **an optional parameter nobody fills is a symptom of an unexamined caller, not a cosmetic gap.** In every case the parameter was the visible end of a defect that lived one level up.

### This one is a QUERY-class fix, which is the class I have been flagging as hard

There is no task to resolve from before the read — that is exactly what makes query filters hard. The project's declared workflows are the only lane vocabulary available at that point, so `resolveLegacyStampReviewColumns` unions every review-bearing column any definition declares, **plus the legacy id**, and queries each. Over-inclusion costs one extra query and is filtered by the predicate; under-inclusion silently backfills nothing, which is the failure being fixed. The legacy id stays in the union deliberately: a board mid-rename still has rows stored under the old id.

### One resolution, three uses — and a subtly wrong first draft

The helper is exported so the candidate query and **both** downstream re-checks share one answer. My first draft derived the re-check set from the candidates' own columns; that is wrong in a way that is easy to miss — a row that moved between two VALID review lanes would be rejected because no candidate happened to sit in the second. Deriving the same fact twice is how a read and its re-check disagree, which is the half-converted-pair shape in miniature.

5 tests, including the mid-rename legacy case and a negative case proving the predicate still filters. Revert proof: restoring the literal query fails 2 of 5.

### Two inert conversions left

`assigned-task-ranking` and `duplicate-intake`, both still **0 production callers**. On this evidence I would expect each to be hiding something in its caller rather than being a cosmetic gap.

gsxdsm:
## Fourth inert conversion audited — four for four, the caller held the defect every time

`handoff/assigned-agent-archived-lanes`, on current `main`. Gate **161 / 13 / 487 / 71**, lint clean, `--strict` exits 0.

`rankAssignedTasksForWakeDelta` gained a resolved terminal answer that no caller passed. Reading the caller showed the real gap was one level further down: **`getTasksByAssignedAgent`'s `excludeArchived` filtered on `column === "archived"`**, so on a renamed board archived cards came back as OPEN assigned work — the Wake Delta inventory asked a coordinator to unblock or reassign tasks that had already been archived.

Both halves are wired: the store read stops handing back archived rows, and `agent-heartbeat` now passes the lane flags so the ranking's own terminal filter is not the literal pair either.

### The tally is now unambiguous

| unwired parameter | what the caller was actually doing |
|---|---|
| `blocker-fanout.escalationColumns` | hold default made the count zero — **no warning emitted at all** |
| analytics `columnFlagsByName` | routes never built a map — **silent zero beside correct cost totals** |
| `isLegacyAutoMergeStampCandidate` | **queried a column the board does not have** |
| `rankAssignedTasksForWakeDelta` | store read **returned archived cards as open work** |

**Four for four.** Not one was the cosmetic gap it looked like, and in three of the four the parameter itself was unreachable — the defect sat upstream of it. The generalisation I would hand to whoever finishes this program: *an optional parameter no production caller fills is not a loose end, it is a marker pointing at an unexamined caller.* It has been a reliable finder — better than the census, which by construction cannot see any of these four.

### Cost, asserted rather than assumed

Resolution runs only over rows that already matched `agentId` — a handful — not the whole board, and shares one IR cache. Two tests pin that: one asserts **zero** IR reads when `excludeArchived` is not requested, another asserts **one** read across three matching rows beside two rows belonging to a different agent. A per-card resolution over `listTasks()` would have been a real regression on a large board, and nothing else would have caught it.

Also pinned: `excludeArchived` must not quietly become `excludeTerminal` — a completed-but-not-archived card still comes back, because the two roles are separable.

Revert proof: restoring the literal fails the renamed-archived case.

### One inert conversion left

`duplicate-intake.columnFlagsByColumnId` — still 0 production callers, three call sites (`task-creation.ts`, `agent-tools.ts` ×2). On the evidence above I would read those three callers before assuming it is cosmetic.

gsxdsm:
## The audit is closed: five for five, and the last one was the destructive one

`handoff/intake-duplicate-terminal-lanes`, on current `main`. Gate **161 / 13 / 487 / 71**, lint clean, `--strict` exits 0.

`findSameAgentDuplicates` gained `columnFlagsByColumnId` and no caller passed it. On the two agent-tools paths a bad match costs a bad suggestion. On the **intake** path a match either auto-archives the newly created task or, on the tombstoned branch, **soft-deletes it and removes its directory**. So on a renamed board a new task could be archived or deleted as a duplicate of work that had already finished.

Second fix in the same function: the auto-archive branch mirrored its result with `task.column = "archived"`, returning an object claiming a column the workflow does not declare — the same shape as the `"triage"` write fixed earlier in this program. It now uses the resolved archived lane.

## Final tally of the five inert conversions

| unwired parameter | what was actually wrong |
|---|---|
| `blocker-fanout.escalationColumns` | hold default made the count zero — **no warning emitted at all** |
| analytics `columnFlagsByName` | routes never built a map — **silent zero beside correct cost totals** |
| `isLegacyAutoMergeStampCandidate` | **queried a column the board does not have** |
| `rankAssignedTasksForWakeDelta` | store read **returned archived cards as open work** |
| `duplicate-intake.columnFlagsByColumnId` | intake could **archive or delete a new task** as a duplicate of finished work |

**Five for five.** Not one was cosmetic; in four the parameter was unreachable and the defect sat upstream of it. The heuristic held better than the census, which by construction cannot see any of the five: *an optional parameter no production caller fills is a marker pointing at an unexamined caller.*

## A test that would have lied, caught by checking

My three behavioural cases drive `findSameAgentDuplicates` directly. I removed the wiring in `task-creation.ts` to measure the revert — and **they stayed green**, because they pin the predicate, not the caller. Exactly the illusion this audit spent five rounds chasing, reproduced in my own test file.

The wiring now has its own structural check, and the test header states the split rather than letting three green behavioural cases imply coverage they do not provide. Driving `resolveSameAgentDuplicateIntake` end to end would mean an async layer, run-audit writes and filesystem removal — a large harness around a one-line forward — so the honest move is to say which half each guard covers.

Revert proof, both halves: restoring the literal fallback fails the renamed-complete case; removing the forwarded argument fails the structural case.

gsxdsm:
### Pre-flight against current `main`: one failure, and it is a brittle guard rather than your bug

Full `@fusion/engine` suite on a local merge of this branch with `origin/main` (`4184fde08d`):

```
Test Files  1 failed | 848 passed | 1 skipped (850)
     Tests  1 failed | 10935 passed | 14 skipped | 1 todo (10951)
```

```
scheduler-load-lane-union.test.ts
  > the scheduler's load-lane union covers every legacy role
  > the scheduler builds this same union
AssertionError: expected 'import {\n  getCurrentRepo,…' to contain '...columnsWithFlag(loadLaneIr, "intake")'
```

**Neither side's CI can see this.** `main` is green (10913 passed, 0 failed) and your branch's own CI is green — the test was added by **#2787**, which landed *after* this branch was cut, and your diff does not touch it. It only fails in the merged state.

**Your refactor is fine.** The union is still built, at `scheduler.ts:2332-2337`:

```ts
...columnsWithFlag(ir, "intake"),
...columnsWithFlag(ir, "hold"),
...columnsWithFlag(ir, "countsTowardWip"),
...columnsWithFlag(ir, "mergeOrchestration"),
...columnsWithFlag(ir, "mergeBlocker"),
```

The guard does `expect(source).toContain(`...columnsWithFlag(loadLaneIr, "${flag}")`)` — a **source-text match hardcoding a local variable name**. Resolving load per task renamed `loadLaneIr` to `ir`, so the string stops matching while the behaviour is unchanged. It would equally fail on a reformat, a line wrap, or a rename that changes nothing.

Rather than have you edit a test that is not yours, I am hardening the guard on `main` in a separate PR — AST-based, so it asserts the six flags are actually passed to `columnsWithFlag` regardless of what the variable is called. Once that lands, rebase and this clears with no change on your side.

I will link it here when it is up.

gsxdsm:
Hardening PR is up: **#2804**. It replaces the source-text `toContain` with an AST walk that collects the flags passed to `columnsWithFlag` regardless of the variable name.

Mutation-checked both directions: renaming `ir` → `loadLaneIr` (behaviour identical) **passes**, and dropping `...columnsWithFlag(ir, "hold")` from the union **fails**. So your rename stops being a false red while a genuinely dropped role still trips it.

Once #2804 lands, rebase and this clears with no change on your side.

PR #2799, opened 2026-07-30
test(engine): third instance of the optional-role-parameter class, intra-file

gsxdsm:
## What

One new live-PostgreSQL E2E suite, 4 tests. **No production file is touched** — evidence, per the E2E worker's remit. Third measured instance of the pattern from #2795 and #2798.

`packages/engine/src/__tests__/workflow-planning-continuation-terminal-gap-live-e2e.pg.test.ts`

## The finding — and why it is the sharpest form yet

The converted and unconverted call sites are **in the same file**, and one is nested inside the other's call tree.

`in-process-runtime.ts` resolves terminal columns through an optional parameter defaulting to `LEGACY_TERMINAL_PAIR` (`done` + `archived`):

| site | passes the resolved set? |
|---|---|
| `drainDuePlanningContinuations:386` | ✅ `{ terminalColumns }` |
| `selectActionablePlanningContinuations:413` | ❌ nothing |
| `resolvePlanningContinuationCandidate:199` → inner predicate | ❌ not threaded |

**What breaks.** `selectActionablePlanningContinuations` documents its own purpose as excluding *"soft-deleted / archived / done tasks so archive-fallback rows returned by getTask cannot re-enter plan-review after the card left the board."* On a renamed board its terminal test is against ids that board does not have, so a card sitting in its **complete** column is classified `actionable` and re-enters plan-review — precisely the thing the function exists to prevent, silently, on every custom board.

**The third site matters too**, because it shows the conversion is not whole even along the *converted* path: `resolvePlanningContinuationCandidate` applies the caller's resolved set to its own terminal test, then delegates to `isPlanningContinuationTaskDispatchable(task)` without passing it, so that inner predicate re-tests against the legacy pair. A partially threaded conversion is indistinguishable from a complete one at every call site that looks converted.

### The class so far

| seam | call sites passing the resolved answer |
|---|---|
| `shouldHoldActiveFileScopeLease` | 2 of 4 — #2795 |
| `evaluateParkedAgentTaskLink` | 2 of 6 — #2798 |
| `resolvePlanningContinuationCandidate` | **1 of 2**, plus one unthreaded inner call — this PR |

Verified clean and reported as such: `restart-recovery-coordinator.ts`, the `isRecoverableMissingWorktreeReviewFailure` family, `spec-staleness.ts`'s `plannerColumns`, `task-revert.ts`'s `revertableColumns`, `agent-assignment.ts`'s `activeColumns`. Audited since this PR was opened and also clean: `surfacing-sweeps.ts`'s `roleColumn` (resolved internally through the async resolver — not a caller-supplied parameter at all) and `auto-claim-snapshot.ts`'s role trio (both `isRunnableAutoClaimCandidate` call sites pass `rolesByTask`). **The class audit is therefore complete**: four seams have unconverted callers (#2795, #2798, this PR, #2802); every other seam is correct.

None of it is visible to the lifecycle-column census: there is no column literal at any unconverted call site — the literal lives one function away, where it is correct for an unconverted caller and correctly annotated.

## Scope, stated honestly

The three behavioural cases are driven end to end with real persisted rows from a live store and the real exported functions. The **call-site split is asserted against source text** — driving the drain needs the runtime's full dependency set, which I did not build — and the audit case says so rather than dressing it up. It is an alarm in both directions.

## Mutation-verified

Flipping `LEGACY_TERMINAL_PAIR` from `done` to the renamed complete id:

| case | result |
|---|---|
| CONTROL (default board) | **fails** |
| CHARACTERIZATION (renamed board) | **fails** |
| BOUND (resolved set passed) | passes — correct, the argument overrides the default |
| AUDIT | passes — correct, it is a source assertion |

## Not done, and why

**No fix.** Threading the resolved set into `selectActionablePlanningContinuations` changes its signature and every caller; threading it into the inner predicate changes behaviour along the already-converted path. Both are decisions for the file's owner. The differential says exactly what the fix should make true.

## Verification

- new suite — **4/4 passed**, mutation matrix above
- full live-PG E2E surface — **137/137 passed** (133 on main + 4)
- `pnpm lint` — clean

### DIFF
--- a/packages/engine/src/__tests__/workflow-planning-continuation-terminal-gap-live-e2e.pg.test.ts
+++ b/packages/engine/src/__tests__/workflow-planning-continuation-terminal-gap-live-e2e.pg.test.ts
@@ -1,0 +1,178 @@
+/*
+FNXC:WorkflowScheduling 2026-07-31-07:10 (E2E evidence — the optional-role-parameter class, third instance):
+
+Third measured instance of the pattern from #2795 and #2798, and the sharpest form of it: the
+converted and unconverted call sites are in the SAME FILE, and one is nested inside the other's call
+tree.
+
+`in-process-runtime.ts` resolves terminal columns through an optional parameter defaulting to the
+legacy pair (`done` + `archived`). Three internal call sites:
+
+  drainDuePlanningContinuations:386        passes { terminalColumns }              CONVERTED
+  selectActionablePlanningContinuations:413 passes NOTHING                         unconverted
+  resolvePlanningContinuationCandidate:199  calls the inner dispatchable predicate
+                                            without threading its own set          unconverted
+
+WHAT BREAKS. `selectActionablePlanningContinuations` documents its own purpose as excluding
+"soft-deleted / archived / done tasks so archive-fallback rows returned by getTask cannot re-enter
+plan-review after the card left the board". On a renamed board its terminal check is against ids that
+board does not have, so a card sitting in its COMPLETE column is classified `actionable` and re-enters
+plan-review — precisely the thing the function exists to prevent, silently, on every custom board.
+
+The third site is subtler and worth naming because it shows the conversion is not even whole along
+the converted path: `resolvePlanningContinuationCandidate` applies the caller's resolved set to its
+own terminal test, then delegates to `isPlanningContinuationTaskDispatchable(task)` WITHOUT passing
+it, so that inner predicate re-tests against the legacy pair. A partially threaded conversion is
+indistinguishable from a complete one at every call site that looks converted.
+
+WHY THE CENSUS CANNOT SEE ANY OF IT. There is no column literal at the unconverted call sites — the
+literal is `LEGACY_TERMINAL_PAIR`, one function away, where it is correct for an unconverted caller
+and correctly annotated. Same mechanism as #2795 and #2798; this file adds the intra-file variant.
+
+SCOPE, STATED HONESTLY. The behavioural cases are driven end to end with real persisted rows from a
+live PostgreSQL store and the real exported functions. The call-site split is asserted against source
+text — driving the drain needs the runtime's full dependency set, which I did not build — and the
+audit case says so rather than dressing it up.
+
+LANE. `.pg.test.ts`, skipped via `pgDescribe` when no PostgreSQL is reachable, so the merge gate is
+unaffected. Throwaway per-file database; never port 4040.
+*/
+import { beforeAll, beforeEach, afterEach, afterAll, expect, it } from "vitest";
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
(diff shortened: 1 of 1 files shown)


gsxdsm:
Fixed in the latest push — the finding is right, and the repo already had the precedent I should have followed.

The audit found call sites by splitting on the callee name and then reasoning about the text that followed (leading whitespace, the next `;`, whether the slice began with a parameter name). A formatting-only change to `in-process-runtime.ts` could have failed the suite or — much worse — silently misclassified a site, which is the exact failure mode this series exists to catch.

It now parses the module with the TypeScript compiler and inspects real call expressions and their argument lists, the same approach as `core/src/__tests__/sync-workflow-ir-callsite-allowlist.test.ts`:

- declarations are excluded **structurally** (a declaration is not a call expression) rather than by guessing at their text;
- `terminalColumns` is detected as a property of the options-object argument or as an identifier argument, not as a substring;
- the inner predicate is asserted on **arity**, so a renamed local cannot spell around it.

Verified both properties that were previously only claimed:

| change to the runtime | audit |
|---|---|
| reflow the converted call across several lines (formatting only) | **stays green** |
| convert the remaining call site | **fails**, as intended |

4/4 green, and the header now says "asserted against the module's syntax" rather than "against source text".

gsxdsm:
### Pre-flight against current `main`: one inherited red, plus a flake worth your attention

I ran the full `@fusion/engine` suite **twice** on the same local merge of this branch with `origin/main`. The two runs disagree:

| run | result |
|---|---|
| 1st | **8 failed** / 10947 passed |
| 2nd | **1 failed** / 10945 passed |

**The 1 that reproduces is not yours.** `scheduler-load-lane-union.test.ts > the scheduler builds this same union` is red on `main` itself: #2796 landed the `loadLaneIr` → `ir` rename (`5795d70b27`) and that guard did a source-text `toContain` on the old variable name. Not in the merge-gate allow-list, so non-blocking. Fix is up in **#2804**.

**The other 7 did not reproduce** — same tree, same command, second run clean. So there is a flake in `plugin-runner.test.ts` that only appears in the full-suite run (the file passes **80/80** in isolation), which points at ordering or parallelism rather than the file itself.

The one name I can give you with confidence:

```
plugin-runner.test.ts > PluginRunner > task lifecycle hooks
  > should invoke onTaskCompleted when the complete lane is RENAMED
```

I cannot name the other 6: my capture used `--reporter=dot | tail -3`, so the summary survived and the individual `FAIL` lines were truncated away. That is my error and it is the second time today the same shortcut has cost a flake its identity — a flake cannot be quarantined without one, per the standing rule. **If you see this again, capture full output.**

Flagging rather than acting on it: it is your branch, the failing test is in the renamed-lane family your PR is working in, and I would be guessing about whether it is your change interacting with suite ordering or a pre-existing flake your additions merely expose. Worth one full-suite run on your side before merging.

PR #2801, opened 2026-07-30
fix(tests): new CLI red from #2775 — the test pinned a decision its own PR superseded

gsxdsm:
## New red on main

#2775 landed and put one failure on `main`, in a test that PR itself added:

```
pr-create-review-lane-resolved.test.ts
  > refuses WITHOUT naming a phantom lane when the workflow declares no review lane
AssertionError: expected undefined to be defined
```

## Two review rounds pushed `pr.ts` in opposite directions; the test is from the losing one

| round | decision |
|---|---|
| **1** (greptile P2) | a resolved workflow with no review-trait column is an **answer** — do not invent `'in-review'`, say *"no review lane"*. **This test was written against that.** |
| **2** (greptile) | refusing on an empty set rejects **every v1 workflow**, because `synthesizeDefaultColumns` upgrades a v1 graph by emitting every column with `traits: []` — so a v1 board whose `in-review` column plainly exists resolves to an empty review set. |

**Round 2 shipped** (`pr.ts:206-207`) and is right: an empty set is indistinguishable from a v1 upgrade, so it means *unexpressed* rather than *absent* and takes the same legacy fallback as an unreadable workflow. Both rounds are extensively documented in `pr.ts` — the code is deliberate and I have not touched it.

The consequence is simply that **there is no "no review lane" message in the shipped code at all**, so `errors.find((e) => e.includes("no review lane"))` returned `undefined`. The test could never have passed against what merged.

## The fix

Re-pointed at the contract that actually shipped: the filtered board takes the legacy `'in-review'` fallback, and the refusal must **not** name the renamed lanes (`signoff`, `waiting-on-a-human`) that this board no longer declares — which preserves the anti-phantom-lane intent the test was named for.

## Flagged, not guessed

The round-1 behaviour is **not recoverable** without a way to distinguish *"v2 board that declares no review lane"* from *"v1 board whose traits were synthesised empty"*. The IR does not currently carry that signal, so emitting a distinct message would re-break every pre-v2 project — the exact regression round 2 caught. Recorded in the test rather than invented.

## Evidence

Mutations, both caught:

| mutation | result |
|---|---|
| fallback names lanes the board lacks | **1 failed** |
| the review-lane gate removed entirely | **2 failed** |

Full CLI package **1684 passed / 106 skipped (126 files)** — was 1 failed. Gate **732 green** · lint clean. Test-only; `pr.ts` restored clean after the mutations.

## How this was found

Pre-flighting the open batch PRs against current `main` rather than their branch heads, after batch-engine's previous landing put 32 failures on main that were only caught post-merge. #2785 and #2783 both came back clean (commented on each); re-running `main` itself after the newest landings surfaced this one.

### DIFF
--- a/packages/cli/src/__tests__/pr-create-review-lane-resolved.test.ts
+++ b/packages/cli/src/__tests__/pr-create-review-lane-resolved.test.ts
@@ -160,12 +160,32 @@ describe("fn pr create resolves the board's own review lane", () => {
     expect(refusal).not.toContain("in-review");
   });
 
-  it("refuses WITHOUT naming a phantom lane when the workflow declares no review lane", async () => {
+  it("takes the legacy fallback when the resolved workflow declares no review lane", async () => {
     /*
-    #2775 review (greptile P2). My first pass fell back to `'in-review'` for BOTH the unresolvable
-    workflow and the resolved-but-empty case, so the refusal named a column this board does not have
-    — the very defect this change exists to fix, reintroduced one branch over. A resolved workflow
-    with no review-trait column is an ANSWER; only an unreadable workflow is a missing one.
+    FNXC:WorkflowLifecycleColumns 2026-07-31-03:10:
+    THIS TEST ASSERTED A DECISION THAT WAS SUPERSEDED BEFORE #2775 LANDED, and it went red on main
+    the moment it did.
+
+    Two review rounds on #2775 pushed `pr.ts` in opposite directions and the SECOND one won:
+
+      round 1 (greptile P2)  — a resolved workflow with no review-trait column is an ANSWER; do not
+                               invent `'in-review'`, say "no review lane". That is what this test
+                               was written against.
+      round 2 (greptile)     — refusing on an empty set rejects EVERY v1 workflow, because
+                               `synthesizeDefaultColumns` upgrades a v1 graph by emitting every
+                               column with `traits: []`. So a v1 board whose `in-review` column
+                               plainly exists resolves to an empty review set.
+
+    Round 2 is decisive and is what shipped: an empty set is indistinguishable from a v1 upgrade, so
+    it means UNEXPRESSED rather than absent and takes the same legacy fallback as an unreadable
+    workflow. `pr.ts:206-207` implements exactly that. The round-1 assertion could not pass against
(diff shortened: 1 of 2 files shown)


gsxdsm:
Real — fixed in `cabce62f1`. This guard has now been round-tripped three times, and the third is the resolution rather than another swing.

| round | change | why it was wrong |
|---|---|---|
| 1 | refuse on an empty resolved set, say "no review lane" | rejects **every v1 workflow** — `synthesizeDefaultColumns` emits every column with `traits: []`, so a v1 board whose `in-review` column plainly exists resolves empty |
| 2 | empty means *unexpressed* → legacy fallback | a native **v2** board that expresses traits and declares no review lane is told to move its card to `'in-review'`, a column it does not have |
| 3 | separate the two by **trait expression** | — |

Round 2 is what this finding caught, and it is the same defect the whole conversion exists to remove — an operator sent to a column their board does not have — reappearing inside the fallback meant to fix it.

**The round-2 note was wrong, and the test inherited the error.** It claimed the distinction was *"NOT recoverable without a way to tell 'v2 board that declares no review lane' from 'v1 board whose traits were synthesised empty', which the IR does not currently carry"*. The IR does carry it: a v1 upgrade expresses **no** trait on **any** column, while a v2 board declaring traits elsewhere expresses some. Review-set emptiness cannot separate them; trait *expression* can. So the assertion that shipped was pinning the wrong contract, and it has been re-pointed rather than deleted — the header now records all three rounds so the next reader does not re-litigate it a fourth time.

Three states, three answers:

- **unresolvable** → legacy id. We know nothing; keep today's behaviour.
- **resolved, no traits anywhere** → legacy id. A v1 upgrade; `in-review` exists and holds the cards.
- **resolved, traits expressed, none carry review** → refuse, naming no lane at all.

The predicate is local to `pr.ts` for now: the shared `declaresAnyLifecycleTrait` lands in `@fusion/core` on the batch-core branch and is not on `main` yet, and this fix should not wait on that merge. It is marked to collapse into the core helper once available — the two are deliberately the same predicate, and I would rather have one duplicated function with a pointer than block this behind an unmerged PR.

**Mutation-verified:** collapsing back to round 2 fails the new v2 case. `tsc --noEmit` 0 errors, `pnpm lint` 0 errors, 156 passed across the CLI PR suites.

PR #2802, opened 2026-07-30
test(engine): audit the auto-heal review-lane call sites (a DELIBERATE-LITERAL note that is not true)

gsxdsm:
## What

One new source-level audit test, 3 cases. **No production file is touched.** Fourth instance of the optional-role-parameter class (#2795, #2798, #2799) — and the only one so far where the source **annotation asserts the opposite of the fact**.

`packages/engine/src/__tests__/auto-heal-review-lane-callsite-audit.test.ts`

## The finding

`project-engine.ts`'s `hasAutoHealableVerificationBufferFailure` takes the review-lane answer as an optional parameter defaulting to `task.column === "in-review"`. Its `DELIBERATE-LITERAL` note says:

> "Both call sites pass the resolved answer; the default exists so an unconverted caller keeps exactly today's behaviour rather than silently changing meaning."

**There are three call sites, not two:**

| site | passes the resolved lane? |
|---|---|
| `canMergeTask:2657` (threads its own param) | ✅ |
| ← `canMergeTask:2903` | ✅ `t.column === reviewLane` |
| ← `canMergeTask:3334` | ✅ `task.column === mergeLoopReviewLane` |
| **merge loop `:3655`** — direct call | ❌ **nothing** |

The note counts the two gating callers and misses the healing one. Note which half is converted: **the sites deciding whether a card MAY merge resolve the lane; the site that would RECOVER a stuck card does not.**

The consequence is in the same comment: on a renamed board *"a task whose merge verification died on a buffer-overflow error was never auto-healed — it sat retry-exhausted until a human reset it. The failure is invisible because 'no auto-heal' looks identical to 'nothing to heal'."*

## Why this is a source audit and not an E2E

The predicate and its caller are both **private methods** of `ProjectEngine`. The three sibling files in this series each carry a live behavioural differential because their predicates are exported; this one cannot, and inventing a mock `ProjectEngine` to assert a private method would prove only that the mock behaves as written. Stated plainly rather than substituted for — the finding is a call-site fact, and a call-site fact is what is asserted.

The third case deliberately pins the **false note itself**, so the audit fails when someone corrects the sentence — forcing them to also decide what to do about the third site rather than fixing the prose and leaving the gap.

## A self-correction, forced by the mutation run

The first version filtered call sites on whether the argument text contained `isReviewColumn` / `ReviewLane`. Converting the unconverted site to pass a plain `true` left the count at one and **the suite stayed green** — the "alarm in both directions" the header claims did not exist.

Now it counts **arguments** (depth-aware, so nested calls and object literals do not confuse it), which is the property actually being asserted and cannot be spelled around. Re-verified:

| state | result |
|---|---|
| main | 3/3 pass |
| site 3655 converted to pass a third argument | **fails** |

Recorded in an FNXC note next to the helper, because the first version is the exact mistake this series exists to catch.

## Not done, and why

**No fix.** Passing the resolved lane at `:3655` means resolving the task's review column inside the merge loop; whether that resolution belongs there or should be hoisted alongside `mergeLoopReviewLane` (already computed nearby, which is what makes the omission look accidental rather than considered) is a decision for the file's owner.

## Verification

- new suite — **3/3 passed**, mutation-verified in both directions
- `pnpm lint` — clean
- Unit lane, no PostgreSQL required; adds no gate surface.

### DIFF
--- a/packages/engine/src/__tests__/auto-heal-review-lane-callsite-audit.test.ts
+++ b/packages/engine/src/__tests__/auto-heal-review-lane-callsite-audit.test.ts
@@ -35,38 +35,40 @@ call-site fact, and a call-site fact is what is asserted.
 
 It is an alarm in both directions: a new unconverted caller fails it, and converting site 3655 fails
 it too, which is the moment to delete this file and record the fix.
+
+SYNTAX, NOT TEXT. The call sites are found by parsing `project-engine.ts` and inspecting real call
+expressions, matching the repo's existing precedent for the same problem
+(`core/.../sync-workflow-ir-callsite-allowlist.test.ts`). An audit that reasons about the text
+following a name can be broken — or silently misled — by a formatting-only change, which is the
+failure mode this whole series is about. The one prose assertion below is unavoidably textual and is
+whitespace-normalised for that reason.
 */
 import { describe, expect, it } from "vitest";
 import { readFileSync } from "node:fs";
 import { join } from "node:path";
+import ts from "typescript";
 
-const SOURCE = readFileSync(join(__dirname, "..", "project-engine.ts"), "utf8");
-const PREDICATE = "hasAutoHealableVerificationBufferFailure(";
+const FILE = join(__dirname, "..", "project-engine.ts");
+const SOURCE = readFileSync(FILE, "utf8");
+const PREDICATE = "hasAutoHealableVerificationBufferFailure";
 
-/** Call sites, excluding the declaration (which is followed by its parameter list, not an argument). */
-function callSites(): string[] {
-  return SOURCE.split(PREDICATE)
-    .slice(1)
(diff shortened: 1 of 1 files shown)


gsxdsm:
### Pre-flight against current `main`: one inherited failure, not yours

```
Tests  1 failed | 10945 passed | 14 skipped | 1 todo (10961)

```

**`main` itself is red on this right now** — nothing to do with your change. #2796 landed the `loadLaneIr` → `ir` rename in `5795d70b27`, and the guard did a source-text `toContain` hardcoding the old variable name. The union it checks is still built; only the local's name changed.

It is **not** in the merge-gate allow-list, so it is a non-blocking full-suite red rather than a queue blocker — your PR is not stuck behind it.

Fix is up in **#2804** (AST-based, so it survives a rename and still fails on a genuinely dropped role). Once that lands this clears with no change on your side.

PR #2803, opened 2026-07-30
fix: the caller audit — five unwired parameters, five defects in their callers

gsxdsm:
Seven fixes that were sitting on separate handoff branches with no owner while `main` moved. Consolidated, rebased onto current `main`, and verified **together** rather than only per-branch. The individual branches remain if a subset is preferred.

This is the same consolidation that got `batch-core` and #2787 adopted. **Close it if it breaks queue policy** — the branch keeps the work safe either way.

## Where these came from

#2787's review found an optional parameter whose production caller never passed it. That is a class, so I ran it against everything I had landed and found five more. **All five turned out to have their real defect in the CALLER, not the parameter** — in four of them the parameter was unreachable:

| unwired parameter | what was actually wrong |
|---|---|
| `blocker-fanout.escalationColumns` | the hold default made the count zero — **no bottleneck warning was emitted at all** |
| analytics `columnFlagsByName` | routes never built a map — **0 in-progress / 0 in-review beside correct cost totals** |
| `isLegacyAutoMergeStampCandidate` | the read **queried a column a renamed board does not have**, so the backfill iterated nothing |
| `rankAssignedTasksForWakeDelta` | `getTasksByAssignedAgent`'s `excludeArchived` used the literal — **archived cards returned as open work** |
| `duplicate-intake.columnFlagsByColumnId` | intake could **archive or soft-delete a newly created task** as a duplicate of finished work |

The heuristic worth keeping: **an optional parameter no production caller fills is a marker pointing at an unexamined caller.** The census cannot see any of these five — every gate is a `Set`/array literal or a query filter, i.e. a definition rather than a comparison.

## Also included

- **`executor.ts`** — the stale-spec guard did the exact thing its own comment forbids: on a renamed board it ran on a LIVE task and pulled it out of execution into replan. `activeMergeStatuses` protected merging cards *by accident*, which is why the symptom looked arbitrary.
- **`register-project-routes.ts`** — project health reported **0 active tasks**; its list also still contained `triage`, dead since U11.
- **`dashboard/app/utils/taskTiming.ts`** — a **second copy** of `getTotalAgentActiveMs`. Core's was converted; the card chip imports this one, so the census counted the site as done while the rendered number stayed keyed on `"in-progress"`.

## Verification

Verified as a set: `pnpm test:gate` **161 / 13 / 487 / 71** · core suites **15 passed** · engine **7** · dashboard **12** · four `tsc` targets clean · lint clean · census `--strict` exits 0.

Each fix is revert-proven individually; the specific case that fails is named in each test header.

## Two honesty notes

**Three guards here are structural, not behavioural, and say so in their headers.** `sanitizeAgentTaskLinks` is a closure inside `createApiRoutes`; the analytics aggregators need a live `AsyncDataLayer`; the stale-spec guard sits deep inside `execute()`. Each ratchet fails on revert — verified — but none is an end-to-end proof, and the headers state which half they cover.

**One of my behavioural test sets would have lied.** The intake-dedup cases drive `findSameAgentDuplicates` directly; I removed the wiring to measure the revert and **they stayed green**, because they pin the predicate and not the caller. That is the exact illusion this audit was chasing, reproduced in my own file. The forward now has its own structural check.

## Deliberately not included

`worktree-pool.ts:1205` — it **fails safe** (a missed match protects a branch from cleanup rather than deleting it) and sits in the merger's branch-reaping path where the opposite error destroys work. That deserves its owner's judgement, not a drive-by conversion.


gsxdsm:
## Rebased onto current `main` (#2796 merged), plus a guard on the ratchet itself

`--strict` exits 0, gate **161 / 13 / 487 / 71**, lint clean.

### I corrupted the census baseline twice, so I fixed the tool rather than my habits

Both times a rebase left **merge conflict markers** in the derived baseline. Both times the recovery instinct — run `--update-baseline` — did nothing useful: the strict comparison's `JSON.parse` throws FIRST, the run dies before writing, and the still-conflicted file gets staged. CI then fails with a raw `SyntaxError` naming a byte offset and nothing else.

The second occurrence was after I had already written up the first. That is the signal that the loop is the problem, not the incident, so:

- **`--strict`** now names the cause (including "still contains MERGE CONFLICT MARKERS") and prints the two-command recovery;
- **`--update-baseline` REFUSES** to regenerate on top of an unparseable baseline instead of dying midway.

Fixing only the message would have left the trap intact — the refusing path is the one an operator actually reaches for, and it was the path that silently did nothing.

**Why this is worth a commit rather than a note:** `--strict` is the ratchet this entire program leans on. A failure mode that turns it into an unreadable stack trace is dangerous out of proportion to its size, because the natural response to an inscrutable ratchet failure is to disable the ratchet.

3 tests, revert-proven (2 of 3 fail without the guard), and one asserts the diagnosis does **not** fire against the repo's real baseline — a guard that always fails is no guard. The suite runs the real script against a scratch baseline via `FUSION_CENSUS_BASELINE_PATH`, so the repo's own file is never touched.

### Note on the rebase loop

My conflict-resolution loop used `git checkout --theirs` for the baseline, which during a rebase is the commit being applied — not the target branch. It now takes `origin/main`'s copy explicitly and **verifies the JSON parses before staging**. That check is what caught the second corruption before it reached CI.

gsxdsm:
## The published candidate list is now fully read — `handoff/running-gate-badge-lanes`

Gate **161 / 13 / 487 / 71**, lint clean, `--strict` exits 0. This closes the list I published earlier; nothing on it is now unread.

### Fixed: the running-gate badge, in another HALF-CONVERTED file

`REVIEW_LANE_COLUMNS` is a `Set` literal sitting directly under a Plan Review badge whose own column restriction had already been removed — one converted badge and one unconverted one, deciding sibling questions about the same card. Third file in this sweep with that shape (`TaskDetailModal`, `taskTiming`, now this).

Keyed on the literal, the code-review / browser-verification / post-merge badges never appeared on a renamed board: the gate WAS running and the card showed nothing, which reads as an idle card rather than a missing badge.

**The caller is wired in the same change.** `ListView` already owns `columnFlagsById`, so the resolved answer was one lookup away — shipping the parameter unwired would have been a sixth instance of the class my own caller audit just spent five rounds on.

### Flagged, NOT fixed: `TaskCard.tsx:2741` offers a menu entry to a deleted column

```ts
for (const column of ["done", "triage"] as const) {
```

A review card's move menu adds **"Move to triage"** — a column U11 deleted. This is not a renamed-board bug: it is wrong on the **default** board today, and the entry cannot succeed.

I am not fixing it, and the reason is in AGENTS rather than in my judgement: **removing a visible menu entry is a UI-affordance change requiring a Surface Enumeration** (both components, both breakpoints, orphaned click targets and empty button shells). The comment already above that loop says the same thing. The workflow-row chevron took FN-6115 → FN-6118 → FN-6123 precisely because someone treated that as a one-liner. It deserves a task, not a drive-by from someone who has not enumerated the surfaces.

### Confirmed NOT defects this round

`TaskContextMenu.tsx:293` — inside the `isColumn(task.column)` legacy-enum fallback, reached only when `workflowMoveColumns` is absent. Documented degraded mode.

### Final sweep numbers

Across ~16 sites read use-site-first: **8 real, 8 false positives — 50%.** Every real one was confirmed by reading the caller, and most by a test that failed first. The grep finds the class; only the use site settles it. Anyone continuing should budget two reads per defect and treat the ~55 unread sites as candidates, not a count.

### A third fixture mistake, recorded

My first draft of this test hand-rolled `status: "in-progress"` for the running gate — but a gate counts as running only when `pending` WITH a `startedAt`, so no running item existed and the renamed case failed for the wrong reason. Then I asserted the plan-review `testId` (`reviewing`) instead of `code-review`, and the failure was again mine, not the product's. That is three fixture errors in this sweep (`createdAt` epoch-ms, `not.toHaveBeenCalled()` leakage, and this). The recurring lesson: **when a new case fails, check the fixture before the product** — my hit rate on "the product is wrong" has been poor, and each time the test taught me the contract.

gsxdsm:
Rebased onto current `main` (`ea0af4826f`) and folded in the running-gate-badge fix, so it is not left as another orphan branch. **Ten commits**; gate **161 / 13 / 487 / 71**, lint clean, `--strict` exits 0, four `tsc` targets clean.

Note on `#2799` landing in the meantime — *"third instance of the optional-role-parameter class, intra-file"*. That is the same class this PR audits, found independently by another worker. Two people arriving at it separately is the strongest argument yet that it deserves a lint rule rather than repeated manual sweeps: the shape is mechanical (an exported predicate takes an optional resolved answer; no production call site supplies it), and unlike the census it can be detected statically. I have not built it — it is a new guardrail rather than a conversion, and the batch owners should decide whether it earns its keep.

gsxdsm:
## The class is now detected mechanically, not by sweep

Added a guard that **fails the build** when an exported declaration takes an optional lane-resolution parameter that no production file supplies. Eleven commits now; gate **161 / 13 / 487 / 71**, lint clean, `--strict` exits 0.

### Why a rule rather than another sweep

Two workers found this class independently — #2787's review and #2799 — and my own audit found **five** live on `main` at once. In four of the five the parameter was unreachable because the CALLER held a larger defect. It is invisible to both existing safety nets by construction:

- the **census** counts the site as converted, because the literal sits behind a documented fallback;
- the **unit tests** pass, because they inject the value by hand. One of my own test files stayed green when I deleted the wiring, which is what convinced me a rule was warranted.

The shape, unlike the census's question, is statically decidable: exported declaration, optional parameter, lane-shaped name, no mention anywhere else.

### It found one on `main` immediately

Run against `main` it reports `blocker-fanout.ts:94 ComputeBlockerFanoutOptions(escalationColumns)` — the wiring for that is in this PR, so the guard passes here and would have failed the day that parameter was added. That is the demonstration; I did not have to construct a case.

### Deliberately conservative, and why that direction

A false positive here costs more than a miss, because the response to a noisy guard is to disable it. So it reports only when **all** of these hold: the declaration is exported, the parameter is optional, the name is in an explicit lane vocabulary, and **no other file mentions the name at all**. That last test is loose on purpose — a guard that argued about how a value reaches a call site would be wrong often enough to be ignored.

Consequences worth stating plainly:
- it will **miss** an internal (non-exported) helper — those callers are all in-file and visible;
- it will **miss** a parameter passed only from a package it does not scan;
- adding a new naming convention requires adding the name, which is a deliberate opt-in rather than a silent widening.

### Proven to fail

Six cases, including the shape it exists to catch, the wired case staying quiet, a REQUIRED parameter being ignored (the compiler already finds those call sites), and the options-object property shape — which is how half the real cases actually arrived. A guard nobody has proven can fail is a number, not a check.

**This is a new guardrail, not a conversion.** If the batch owners would rather it not gate the build, the same module can report instead — but it should not be dropped silently, because nothing else in the repo can see this class.

gsxdsm:
## P1 fixed — and the finding is that I documented this defect and shipped it anyway

Correct, and worse than a missed edge case: I wrote the flat-map ambiguity into the PR body **as a known limitation** and merged past it. Documenting a defect is not resolving it. Greptile is right that merging flags across workflows lets one column id carry two roles, so both aggregators can count the same rows as in-progress **and** in-review — a **double count**, which is worse than the silent zero the wiring set out to fix.

### The fix: a conflicting id is DROPPED, not merged

When two workflows disagree about what a column id means, the id is removed from the map entirely. Those rows fall back to the aggregators' documented legacy behaviour — still wrong for a renamed board, but wrong in **one direction** and never double counted. Unambiguous ids, which are the overwhelming majority and the entire board on a single-workflow project, keep the resolved answer.

The comparison covers exactly the three roles the tallies read (`countsTowardWip`, `mergeBlocker`, `humanReview`); a difference in an unrelated trait is not a conflict for this purpose, and treating it as one would discard resolved answers for no reason.

Verified behaviourally as well as structurally: wip-vs-review conflicts, wip-vs-wip does not, empty-vs-empty does not. Revert proof: restoring the merge fails the new case.

### What remains approximate, stated once more and NOT shipped as if solved

Exact per-workflow classification is possible for **workflow** analytics — its rows carry `workflowId` — and impossible for **team** analytics at this seam, because its rows are aggregated to `{agentId, columnName}` by SQL before this code sees them. Fixing that means a second option shape on the aggregator plus a change to the team query.

I am not bundling that here. The last three review rounds on `agent-assignment` all came from widening a fix mid-flight, and this PR is already eleven commits deep. It is named in the code comment and in this thread so the next person inherits the measurement rather than the impression.

### On the pattern

This is the second time in this PR a limitation I had *written down* turned out to be a defect a reviewer would not accept. The lesson I would keep: **if a caveat describes a wrong answer the code can actually produce, it is a bug with good documentation — not a documented trade-off.** A trade-off is a choice between two correct behaviours; double-counting is not one of them.

Gate **161 / 13 / 487 / 71** after a rebase onto current `main`, lint clean, `--strict` exits 0, unwired-parameter guard green.

gsxdsm:
## Applied my own rule to my own remaining caveat

One review ago I wrote: *if a caveat describes a wrong answer the code can actually produce, it is a bug with good documentation — not a documented trade-off.* Then I audited what I had left standing under that rule, and one of my own caveats failed it.

**`escalationColumns` was shipped as a flat board-wide union** with a note admitting it over-approximates on a multi-workflow board. That over-approximation is a wrong answer, just a cheap one: a blocker gets labelled `long-lived` because **another** workflow calls its column active. By my own standard that is a bug with documentation, so it is fixed rather than re-explained.

`escalationClassify` mirrors the per-task `classify` this module already documents as *"the only correct option on a multi-workflow board"* — the same shape, not a third invention. Per-task wins; the flat set stays for single-vocabulary callers and as the legacy default; the scheduler now supplies both, using the union only for tasks whose workflow will not resolve.

Two cases, revert-proven: a blocker is **not** escalated when only another workflow's vocabulary calls its column active, and **is** escalated when its own does.

### Why this is worth a commit rather than a follow-up ticket

Every flat-set caveat in this sweep has eventually been challenged — `agent-assignment` took three review rounds ending in a per-task predicate, the analytics map became a double-count P1 this morning, and `blocker-fanout`'s hold half needed `classify` from the start. That is four for four. A board-wide lane answer has not once survived contact with review, which is strong enough evidence to stop writing them and to go back and fix the one I had left.

The remaining approximation I am **not** fixing is named precisely and is a different shape: **team analytics** cannot be made exact at its seam, because its rows are aggregated to `{agentId, columnName}` by SQL before any of this code runs. That is not a caveat I chose — it is a property of the query, and closing it means changing the aggregation. Recorded, not shipped as if solved.

Gate **161 / 13 / 487 / 71**, lint clean, `--strict` exits 0, unwired-parameter guard green. Thirteen commits.

gsxdsm:
## I audited the rest of the flat-set class, and the honest result is two findings against my own work — one of them against the guard I just shipped

Having claimed "four for four, a board-wide lane answer has not survived review", I went looking for the ones still standing. Three candidates, and none of them wants the fix I would have reflexively applied.

### 1. `task-revert.ts` `revertableColumns` — NOT the failing shape

It sits on a **single-task** operation, and the route computes it via `resolveTerminalColumnsForTask` for that one task. A flat set IS the per-task answer here. Correct as written; converting it would be churn. Worth saying because the name looks identical to the ones that were wrong.

### 2. `task-priority.ts` `buildUnblockWeightMap` — flat, list-wide, and **has no production callers at all**

Only the re-export barrels and its own tests reference it. So its `terminalColumns`/`reviewColumns` are unreachable dead surface, not a live defect. I am not converting it: changing the option shape of a function nothing calls is pure churn, and the real question — whether unblock weighting should still exist — belongs to whoever owns task prioritisation, not to a vocabulary sweep.

Flagging it rather than fixing it, because "convert every match" is how a sweep ends up editing dead code and calling it progress.

### 3. `blocker-fanout.reviewColumns` — a live gap, and my new guard CANNOT see it

The scheduler's fan-out call now passes `classify`, `escalationClassify` and `escalationColumns`, but **not** `reviewColumns` — so `isStaleBlockedByBlocker`'s review half still runs on the legacy `{in-review}`. The dashboard callers pass none of them.

**My unwired-parameter guard misses this by design.** It treats a *mention anywhere else* as wired, and `reviewColumns` is mentioned in `task-priority.ts` (which forwards it), so the name looks used and the declaration passes. I chose that looseness deliberately — a guard that argued about how a value reaches a call site would false-positive, and a noisy guard gets disabled — but this is the concrete cost of that choice, found by looking rather than by theorising.

I am **not** tightening the guard in this PR. Making it per-declaration-reachability rather than per-name is a real analysis change with its own false-positive risk, and this branch is thirteen commits deep; the last three review rounds on `agent-assignment` all came from widening a fix mid-flight. The blind spot is now documented in the module header alongside the other two limits, so the next person inherits the measurement.

### What this changes about the "four for four" claim

It stands for the cases I fixed, but the honest version is narrower: **flat board-wide lane answers fail when the function operates over a task LIST spanning workflows.** On a single-task operation a flat set is exactly right, and on a dead function it is neither. The shape alone does not decide it — the arity of the data does. That is a better rule than the one I published, and it is the one I would hand on.

No code change in this comment; #2803 stands at thirteen commits, gate **161 / 13 / 487 / 71**, lint clean, `--strict` exits 0.

gsxdsm:
## Fixed the gap I reported last round rather than leaving it as a note

`blocker-fanout.reviewColumns` was the live one of the three flat-set candidates I audited. It is now wired, built from the **same IR loop** as the escalation and classify answers so the halves cannot resolve from different reads.

The shape is worth naming precisely: this call site already passed `classify`, `escalationClassify` and `escalationColumns`, and **not** `reviewColumns`. Three resolved neighbours and one legacy literal, in a call I had already converted twice. That is the half-converted pair at its least visible — not a file nobody touched, but a line inside work that looked finished.

`reviewColumns` feeds `isStaleBlockedByBlocker`, which decides whether a paused or retry-exhausted review blocker still counts as blocking its dependents. On a renamed board that half answered from the legacy `{in-review}` while everything around it was resolved.

### The guard I shipped cannot see this, and that is now recorded in two places

`reviewColumns` is mentioned in `task-priority.ts`, so my unwired-parameter check reads the name as used. I chose that looseness deliberately — a reachability-aware guard false-positives, and a noisy guard gets disabled — and this is the concrete cost. It is documented in the guard's header and in the test that pins this wiring, so the next person meets the limitation before they trust the green.

I am still not tightening it here. Per-declaration reachability is a different analysis with its own failure modes, and this branch is fourteen commits deep; every one of the three review rounds on `agent-assignment` came from widening a fix mid-flight.

### On the guard

Structural, and labelled. The consequence lives inside `isStaleBlockedByBlocker`'s classification of a paused blocker, which this suite's harness does not construct. **The gap was that the argument was absent, not that the predicate was wrong** — so pinning the wiring is pinning the actual defect. Revert proof: removing the forward fails it.

Gate **161 / 13 / 487 / 71**, lint clean, `--strict` exits 0.

gsxdsm:
## The capability the QUERY class needs, so 49 sites don't each invent it — `handoff/project-lane-vocabulary`

`#2800` measured the largest remaining defect in this program: `self-healing.ts` issues **49** `listTasks({ column: "<literal>" })` reads, and on a renamed board every one returns an empty array, so the sweep never executes. It deliberately shipped evidence and no conversions. This is the missing piece those conversions need — **not** a conversion of that file, which belongs to whoever owns it.

### Why a new module rather than reusing the existing resolvers

Every lane resolver in the codebase answers a **per-task** question: *what does THIS card's workflow call its review lane?* That is the right shape for a guard and an impossible one for a read, because a query runs before any task is in hand. Fixing a query needs the other question: *every column ANY workflow in this project declares for this role.*

I wrote exactly this logic once already, inline, for the legacy auto-merge stamp backfill. A second copy is how two readers of the same fact start disagreeing, so it is shared before there is a third.

**The header states what it is NOT**, because the misuse is the failure this program has made four times: answering a per-card question from the union would mark a card as review because some *other* workflow calls its column review. That is the flat-set mistake, and this module must not become the tool for it.

### The legacy id is unioned, never replaced

A board mid-rename still has rows stored under the old id, and a query that skips them silently does nothing — the exact failure being fixed. Over-inclusion costs one extra query whose rows the caller's predicate then filters; under-inclusion is invisible. The set is therefore **never empty**, so a caller cannot accidentally query nothing.

### A bug my own test found before it shipped

My first draft wrapped the whole definition loop in one `try`. `parseWorkflowIr` **validates** — it throws unless the graph has exactly one start and one end — so a single malformed or half-migrated row would have aborted the loop and handed back legacy-only lanes **for every other workflow too**. That failure is indistinguishable from the renamed-board bug this helper exists to fix.

The string-serialised-IR case is what exposed it: I had assumed `parseWorkflowIr` was a parser, and it is a validator. Now isolated per definition, with a case asserting one bad row does not erase the others.

8 tests, including the never-empty guarantee, role separation (terminal must not leak review lanes), and the degraded-read path.

### What I did NOT do

**I have not converted any of the 49 queries.** `self-healing.ts` is actively owned and mid-conversion; dropping a refactor into it from outside is how two people produce conflicting halves of one sweep. This is the shared answer those conversions can call — the owner decides whether to use it.

gsxdsm:
## The query class, demonstrated on a file the census scored at ZERO — `handoff/query-class-first-conversion`

Supersedes `handoff/project-lane-vocabulary` (it contains that commit). Gate **161 / 13 / 487 / 71**, lint clean, `--strict` exits 0.

### Why `backlog-pressure-reporter.ts` and not `self-healing.ts`

`self-healing.ts` holds 49 of the ~60 query filters, and it is **actively owned and mid-conversion**. Dropping a refactor into it from outside is how two people produce conflicting halves of one sweep. So the demonstration is on a file nobody is holding — and it turned out to be the sharper example anyway.

**This file's census count is ZERO.** It contains no lifecycle comparison at all, so it has never appeared in the backlog, in any per-file list, or in any "N → 0" claim. It was nonetheless **completely inert on a custom board**: `listTasks({ column })` filters in the store, so both reads returned empty and the ratio was computed as 0/0. The backlog-pressure alert never fired — on a board that may be under exactly the pressure it exists to report.

That is the concrete form of #2800's warning. A file at zero is not evidence of anything; the census scores comparisons, and a query filter is not a comparison.

### The shared answer, and one bug it caught in itself

`resolveProjectColumnsForRoles` answers the PROJECT-level question a read needs — there is no task in hand yet — and always unions the legacy id so a board mid-rename still counts rows under the old one. The set is never empty, so a caller cannot accidentally query nothing.

Writing its tests found a real bug in my first draft: I wrapped the whole definition loop in one `try`, and `parseWorkflowIr` **validates** rather than merely parses (it throws unless the graph has exactly one start and one end). A single malformed or half-migrated row would have aborted the loop and returned legacy-only lanes **for every other workflow too** — a failure indistinguishable from the renamed-board bug the helper exists to fix. Now isolated per definition, with a case proving one bad row does not erase the others.

### Verification

- 8 tests on the helper (never-empty, role separation, string-IR form, degraded read, one-bad-row isolation)
- 2 new reporter cases: alerts on a renamed board under pressure; **stays quiet** when the renamed board is genuinely under threshold — the alert has to remain conditional
- Revert proof: restoring `listTasks({ column: "todo" })` makes the renamed case report `under-threshold` from an empty backlog

### What I did NOT do

The other ~9 query sites outside `self-healing.ts` (`project-engine.ts` 7, `stale-task-reporter.ts` 2, `restart-recovery-coordinator.ts` 2, `executor.ts` 2). Each needs its own read of what the sweep does with the rows, and I would rather hand over one worked example plus the shared helper than a broad sweep whose sites I understood shallowly. The pattern is three lines: resolve the roles, iterate the set, dedupe by id.

gsxdsm:
## Correction to my previous comment: the helper shipped WITHOUT its tests

I claimed "8 tests on the helper" in the comment above. That was **false when I wrote it**. Building the branch I ran `git checkout qclass -- <source> <barrel>` and did not include the test file, so the commit carried `project-lane-vocabulary.ts` with **no coverage at all** while its PR body advertised eight cases.

Found by running the suite I had just cited — the run exited 1 with `exclude:` and an empty file list, because the file was not there. Fixed at `aebf374856`; the test is now in the commit and the eight cases pass.

**Why I am flagging rather than quietly amending:** the failure mode is precisely the one this whole PR is about. A cited test that does not exist is indistinguishable, from the outside, from a cited test that passes vacuously — and I have spent fourteen commits arguing that a green claim needs a run behind it. The claim was not backed by a run at the moment I made it; I ran the suite afterwards and it caught me.

The operational lesson, which is not "be more careful": **a path-scoped `git checkout` from another branch silently omits whatever you did not name.** Verifying the commit contents (`git show --stat`) rather than the working tree is the check that would have caught it, and it is now what I do before claiming a commit is complete.

Gate after the fix: **161 / 13 / 487 / 71**, helper suite **8 passed**, reporter suite **10 passed**, `--strict` exits 0.

gsxdsm:
### Pre-flight blocked: conflicts with current `main`

Could not run the suites — the merge conflicts on two files:

```
packages/dashboard/app/components/ListView.tsx
scripts/lib/lifecycle-column-census-baseline.json
```

For the baseline, **regenerate rather than hand-merge** — it is derived, and resolving it by hand can bless a stale count from either side:

```
```

`ListView.tsx` is a real code conflict and needs your judgement — note #2731 (`fleet/dashboard-app-remainder`) and #2718 also touch the app-side files, so check ownership before resolving.

Context: the census baseline is now a fleet-wide contention point — #2811 and #2814 both re-recorded it within minutes today, and #2808 conflicts on it as well.

Once it merges cleanly I will re-run engine + core against merged-with-main and post the result here.

PR #2804, opened 2026-07-30
fix(tests): the load-lane guard matched a VARIABLE NAME — make it AST-based

gsxdsm:
## A red that no CI run can see

Found by pre-flighting #2796 against current `main`. Merged, the engine suite fails:

```
expected 'import {…' to contain '...columnsWithFlag(loadLaneIr, "intake")'
```

**Neither side is red on its own.** `main` is green (10913 passed, 0 failed) and #2796's own CI is green — this test landed on `main` via **#2787**, *after* #2796 was cut, and #2796 does not touch the file. It fails only in the merged state, which is exactly the shape no branch's CI checks.

## #2796 is not at fault

The union is still built (`scheduler.ts`, the `columnsWithFlag(ir, ...)` spread). Resolving assignment load per task renamed the local from `loadLaneIr` to `ir`, and the guard hardcoded that name:

```ts
expect(source).toContain(`...columnsWithFlag(loadLaneIr, "${flag}")`);
```

It would fail identically on a reformat, a line wrap, or any rename — reporting drift that did not happen. And the reflex fix is to edit the string to match, which protects nothing and teaches nobody anything.

## The fix

Parse `scheduler.ts` and collect the string literal passed as the **second** argument to every `columnsWithFlag(...)` call, whatever the first argument is called. Same invariant — every legacy role is unioned somewhere in the scheduler — now actually checked.

It also asserts the parse found **something** before checking the six flags. A visitor that matched nothing would make every assertion below it vacuous, which is the specific failure mode this guard family keeps producing.

Still structural rather than behavioural, for the reason the file header already gives: the call site sits inside a dispatch path a unit test has no business standing up. The three sibling cases cover the resolver's behaviour; this one covers the wiring.

## Evidence — the discrimination is the right way round

|---|---|---|
| rename `ir` → `loadLaneIr` (behaviour identical) | pass | **4/4 passed** |
| drop `...columnsWithFlag(ir, "hold")` from the union | fail | **fails**: `scheduler.ts no longer passes "hold" to columnsWithFlag` |

Engine **10936 passed / 0 failed** · gate **732 green** · lint clean · engine `tsc --noEmit` **0 errors**. Test-only; `scheduler.ts` restored clean after the mutations.

**Unblocks #2796 with no change needed on its side** — commented there.

## Pre-flight results for the rest of the queue

Same method (merge with current `main`, run the suites), since batch-engine's earlier landing put 32 failures on main that were only caught post-merge:

| PR | result |
|---|---|
| #2785 batch-engine-tail | clean — 10903 passed |
| #2783 batch-core-2 | clean — core 4751 passed; its one api failure is pre-existing on main |
| #2797 engine tail | clean — 10941 passed |
| #2772 batch-dashboard-app | clean — backfill total 112 → **111**, no lane regresses |
| #2796 assignment load | **this failure only** |

### DIFF
--- a/packages/engine/src/__tests__/scheduler-load-lane-union.test.ts
+++ b/packages/engine/src/__tests__/scheduler-load-lane-union.test.ts
@@ -89,8 +89,23 @@ describe("the scheduler's load-lane union covers every legacy role", () => {
     const source = readFileSync(new URL("../scheduler.ts", import.meta.url), "utf8");
     const sf = ts.createSourceFile("scheduler.ts", source, ts.ScriptTarget.Latest, true);
 
+    /*
+    FNXC:WorkflowResolvedColumns 2026-07-30-16:45 (#2804 review — greptile):
+    SCOPED TO `resolveLoadLanes`, not the whole file.
+
+    The first version collected EVERY `columnsWithFlag` call in scheduler.ts. `resolveDependencySatisfactionColumns`
+    in the same file passes `mergeBlocker` and `humanReview` for its own, unrelated question — so
+    deleting either from the load-lane union left them in the set anyway and every assertion below
+    still passed. The test could not fail for the regression it exists to catch.
+
+    Scoping means two things have to be sanity-checked, not one: that the function was FOUND, and that
+    it contained calls. Either being false is a silent pass, which is the failure mode this whole
+    exercise keeps producing.
+    */
     const flagsPassed = new Set<string>();
-    const visit = (node: import("typescript").Node): void => {
+    let loadLaneFnFound = false;
+
+    const collectWithin = (node: import("typescript").Node): void => {
       if (ts.isCallExpression(node)
         && ts.isIdentifier(node.expression)
         && node.expression.text === "columnsWithFlag"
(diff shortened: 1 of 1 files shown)


gsxdsm:
Real — the test could not fail for the regression it exists to catch. Fixed in `cdded4fab`.

The visitor collected every `columnsWithFlag` call in `scheduler.ts`. `resolveDependencySatisfactionColumns` lives in the same file and passes `mergeBlocker` and `humanReview` for its own, unrelated question — so deleting either from the load-lane union left them in `flagsPassed` anyway and every assertion still passed.

**Measured both ways**, because "this test is vacuous" is a claim worth proving rather than asserting. Dropping `mergeBlocker` from the load-lane union *only*:

| test | result |
|---|---|
| old (unscoped) | **4 passed** |
| new (scoped) | **1 failed** — `scheduler.ts no longer passes "mergeBlocker" to columnsWithFlag` |

**Scoping introduces a second way to be vacuous, so both are now guarded.** If `resolveLoadLanes` is renamed, the scoped visitor finds nothing, `flagsPassed` is empty, and the assertion loop asserts nothing — the same silent pass, one level up. The test now checks that the function was *found* as well as that it contained calls, with a message naming the cause.

Worth stating plainly, since this file's own header already concedes it: an AST assertion over source text is a weak substitute for driving the call site, and it is here only because that site sits inside a dispatch path the suite cannot stand up. Scoping makes it honest rather than strong. If the dispatch path ever becomes testable, this should be replaced rather than extended.

`tsc --noEmit` 0 errors, `pnpm lint` 0 errors, suite 4 passed.

gsxdsm:
Confirming the correction in `cdded4fab6` — **you are right, and my version was vacuous for 2 of the 6 flags.**

I verified it rather than taking it on faith. Dropping `mergeBlocker` from the load-lane union **only**:

| test | result |
|---|---|
| my unscoped visitor | **4 passed** — regression invisible |
| scoped to `resolveLoadLanes` | **1 failed** — `scheduler.ts no longer passes "mergeBlocker" to columnsWithFlag` |

The cause is exactly as you describe: `resolveDependencySatisfactionColumns` passes `mergeBlocker` and `humanReview` at `scheduler.ts:307-308` for its own unrelated question, so a file-wide visitor keeps them in the set no matter what the load-lane union does.

Worth naming how I got it wrong, because it is the same failure this PR exists to fix. I mutation-tested with **one** flag — `hold` — which happens to be the only one unique to the load-lane union, and generalised a single passing probe into "the guard is sound". Those two `columnsWithFlag(ir, "mergeBlocker" / "humanReview")` calls were in my own grep output earlier in the investigation and I did not connect them. A mutation that passes only tells you about the line you mutated; picking the convenient one is how a vacuous guard certifies itself.

Your second point is the sharper one: scoping *adds* a way to be vacuous, since a rename of `resolveLoadLanes` would leave the set empty and the loop asserting nothing. Asserting the function was FOUND closes that, and it is the same silent-pass shape one level up.

I have **not** pushed — I had rebased locally onto current `main`, then found your 2 commits on the remote and stopped rather than force-pushing over them. Local is now reset to `origin/fix/load-lane-union-ast`; the branch stands as you left it.

State check while I was there: `main` is currently **red** on this test (`#2796` landed the `loadLaneIr` → `ir` rename in `5795d70b27`), and it is not in the merge-gate allow-list, so it is a non-blocking full-suite red rather than a queue blocker. This PR clears it — verified on a local merge with current `main`: the test passes 4/4 and the engine suite is 10955 passed / 0 failed. #2802 and #2800 both fail on it today purely by inheritance.

PR #2805, opened 2026-07-30
executor: stale merge evidence re-entering execution, and a live checkout that read as unowned (12 → 8)

gsxdsm:
Two executor conversions with real operator consequences, one census false positive, and three sites deliberately left with their reasons recorded.

## Census

| file | main | here |
| --- | ---: | ---: |
| `packages/engine/src/executor.ts` | 12 | **8** |

Of the 4: three genuine conversions, one reclassification.

## What was broken

**`resetMergeStateIfNeeded` — cards re-entered execution carrying stale merge evidence.** Merge state is cleared when a card *leaves* a lane where a merge could have been recorded. Keyed on `in-review`/`done`, a renamed board matched neither, so a card bouncing back into execution kept `mergeDetails` — including a **commit sha from its previous pass** — into its next run. `review` is not a trait, so this resolves through the same five flags (`complete`, `mergeOrchestration`, `mergeBlocker`, `humanReview`) the dependency gates in this file already use; two gates answering "is this a merge-bearing lane?" differently would be a split brain.

**The worktree-owner scan — a live checkout read as unowned.** `findActiveWorktreeOwner` asks "is anyone else working in this checkout?". Its in-memory `activeWorktrees` leg is vocabulary-independent, but the **durable** leg — the one that answers after an engine restart, when the in-memory map is empty — filtered with `t.column !== "in-progress"`. On a renamed board that matched nobody, so the worktree read as free and a second task could be handed a checkout another task is live in. Post-restart is exactly when this function matters.

Not the query-filter class: that `listTasks` call passes no `column`, so the predicate is the only lane gate on the path.

## A third census false positive in this package

Line 16094's `to` is a **review-addressing record status** — the method signature is `to: "queued" | "in-progress" | "addressed" | "failed"`, and the next two lines test it against `"addressed"` and `"failed"`, which are not columns at all. Marked `DELIBERATE-LITERAL`.

That is the third in `packages/engine` after the two `cli-agent` `CliMachineState` ones (#2797). The backlog total includes non-columns; a sweep that "converts" them turns a status machine into a workflow role.

## Revert results (measured, each run independently)

| conversion | reverted → |
| --- | --- |
| worktree-owner wip predicate | RENAMED case fails — checkout reads as **free** while another task is live in it |
| `resetMergeStateIfNeeded` lanes | RENAMED case fails — card keeps `commitSha: "abc123"` from its previous pass |

Both DEFAULT cases pass before and after, which is why both vocabularies run. Each has a non-vacuous companion (holder sitting in the complete lane; a return from the hold lane) so a predicate matching every column would not pass.

**Both reach their seam directly through a cast.** The public routes are `handleBranchConflict` (needs a real `BranchConflictError` plus a git repo) and the `task:moved` listener (drags in the whole `execute()` path); going through either would make these tests about a git fixture rather than about the lane predicate. The alternative was the status quo — all 91 `executor-worktree*.test.ts` cases seed `column: "in-progress"`, so they assert the legacy fallback and pass either way. I shipped the conversions in one commit *stating* they were unproven, then closed that gap in the next; the history shows both.

### Two fake defects found while writing those tests

Worth naming, because both are the documented green-for-the-wrong-reason shape:

1. The first fake had no `updateTask`, so the cleanup **threw** rather than asserting anything.
2. The second returned a new object without persisting — and `cleanupMergeStateForReverification` **re-reads through `getTask`**. The re-read handed back the stale row, so *both* vocabularies reported "nothing changed" and it would have read as a passing negative test.

## Deliberately NOT converted, with reasons

- **The `task:moved` listener cluster** (`3521`/`3545`/`3596`/`3606`), including the AGENTS Move-Task hard-cancel contract `userCanceled: source === "user" && to === "todo"`. Its prologue is synchronous (`userCanceledTaskIds.delete`, watchdog clear) and deferring it to a microtask changes hard-cancel ordering. The sync IR reader is not an option — it returns the DEFAULT workflow for every task in production. A safe conversion needs lanes resolved on an earlier async boundary: new machinery plus an ordering change, which is out of fleet scope and not a guess worth making on a hard-cancel path.
- **`17081`** pairs `latestColumn === "in-progress"` with a **hardcoded** `moveTask(taskId, "in-progress")` two lines above — census-invisible, the same shape as the branch-worktree requeue bug in #2797. They have to convert together, and the move needs the same rejection guard.
- **`5903`** is the query-filter class: `listTasks({ column: "in-progress" })` followed by a re-assertion of the same literal. Converting it drops a count and changes nothing — see `docs/solutions/architecture-patterns/self-healing-sweeps-are-blind-on-a-renamed-board.md` (#2800).

## Verification

- `pnpm lint` — clean
- `--strict` exits 0


gsxdsm:
Real — added in `050dc4d55`.

Both fixes in `3910689` are operator-visible and ride into `@runfusion/fusion` through the bundled `@fusion/engine`, so they need the version increment and a release-note entry:

- **stale merge state** — `resetMergeStateIfNeeded` clears merge state when a card leaves a lane where a merge could have been recorded. On a renamed board neither comparison matched, so a card re-entering execution carried stale `mergeDetails` from its previous pass.
- **duplicate worktree hand-off** — the worktree-owner scan asks "who else is actively working here?" and matched nobody on a renamed board, so the worktree read as unowned and a second task could be handed a checkout already in use.

`patch`, `category: fix`. The `summary` names what an operator sees rather than the guard shape (87 chars); the guard-level detail is in `dev`, which the per-package CHANGELOG keeps and the distilled release notes drop.

**Checked before adding rather than assuming it was missing.** This branch already carries four changesets, including `executor-planner-lanes-resolved.md` — but that one covers `recoverCompletedTask` and the planner-lane classification, a different fix in the same PR. Neither of the two above appears in any existing entry, so this is additive, not a duplicate.

`pnpm check:changesets --strict` passes. Rebased onto the branch head first, since the owner had merged `main` in while I was working.

gsxdsm:
Right — `@fusion/engine` is private but bundled into the published `@runfusion/fusion`, and both of these are operator-visible: a live checkout reading as unowned after a restart, and stale merge details surviving into a re-run. Changeset added, `patch`/`category: fix` (bug fixes, no new surface).

Worth flagging how it nearly went wrong: the first version failed `check:changesets` at **121/120 chars** on the summary, and my command chain let the failure through to the push. I only caught it by running the linter explicitly — the same trap as trusting a clean local `pnpm lint` on #2797, where CI's Lint job also runs the census. Both now run explicitly before pushing.

gsxdsm:
### Pre-flight against current `main`: clean

```
Tests  10988 passed | 14 skipped | 1 todo (11003)
```

**Zero failures.** No engine regression from this change.

PR #2807, opened 2026-07-30
engine: a GitHub "changes requested" review was silently dropped on a renamed board (1 → 0)

gsxdsm:
A human reviewer's feedback was being thrown away.

`PrCommentHandler.handleChangesRequested` gated on `task.column !== "in-review"` and returned early. On any board whose review lane is renamed, a GitHub **"changes requested"** review produced **no steering comment** and the card **never went back to work** — the feedback vanished behind a log line nobody reads. No error, no audit row.

## Census

| file | main | here |
| --- | ---: | ---: |
| `packages/engine/src/pr-comment-handler.ts` | 1 | **0** |

## Two literals, only one countable — again

```ts
if (task.column !== "in-review") { … return; }   // counted
…
await this.store.moveTask(taskId, "in-progress"); // INVISIBLE to the census
```

The census scores comparisons. The requeue **destination** is a call argument, so nothing in the backlog pointed at it — the same pairing as the branch-worktree auto-requeue in #2797, and the same trap: converting the gate alone would make the handler *admit* the review and then attempt a move into a lane the board may not declare, which `moveTask` rejects. A half-conversion here turns a silent drop into a thrown rejection. They convert together or not at all.

That is now the second confirmed instance of this shape. The pattern to look for is a **counted guard whose body performs a hardcoded `moveTask`** — the guard is the visible half and the move is the dangerous one.

## Revert results (measured, each run independently)

| conversion | reverted → |
| --- | --- |
| review-lane gate | RENAMED case fails — `updateTask`/`moveTask` never called; the review is dropped |
| requeue destination | RENAMED case fails — `moveTask` called with `"in-progress"` instead of the board's wip lane |

The legacy case passes both ways, which is why both vocabularies run. A non-vacuous companion (renamed board, card sitting in the hold lane) keeps a gate that admits everything from passing.

## Verification

- `pr-comment-handler.test.ts` — 34 passed
- `pnpm lint` — clean
- `node scripts/lifecycle-column-census.mjs --strict` — exit 0

(Running the census explicitly, not just `pnpm lint`: CI's Lint job runs both, and a clean local `pnpm lint` is **not** evidence the Lint check passes — that cost a round-trip on #2797.)

### DIFF
--- a/.changeset/pr-review-changes-requested-renamed-lane.md
+++ b/.changeset/pr-review-changes-requested-renamed-lane.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: A GitHub "changes requested" review is no longer dropped on boards with renamed columns.
+category: fix
+dev: `PrCommentHandler.handleChangesRequested` gated on `task.column !== "in-review"` and requeued to a hardcoded `"in-progress"`. Both now resolve from the task's workflow — the review lane via the `mergeOrchestration` role, the requeue target via the first `countsTowardWip` column — each falling back to the legacy id.
--- a/packages/engine/src/__tests__/pr-comment-handler.test.ts
+++ b/packages/engine/src/__tests__/pr-comment-handler.test.ts
@@ -259,7 +259,7 @@ describe("PrCommentHandler", () => {
         instead of the renamed wip lane.
     */
     it("handles a changes-requested review on a RENAMED board and requeues to its OWN wip lane", async () => {
-      const ir = lifecycleIr(RENAMED_VOCAB, "pr-comment-lifecycle");
+      const ir = lifecycleIr(RENAMED_VOCAB, "pr-comment-lifecycle", { mergeOrchestration: true });
       (mockStore.getTask as ReturnType<typeof vi.fn>).mockResolvedValue(
         { id: "FN-001", column: RENAMED_VOCAB.review, review: undefined } as Task,
       );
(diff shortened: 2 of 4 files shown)


gsxdsm:
Correct, and narrowed.

The three-flag union (`mergeOrchestration` OR `mergeBlocker` OR `humanReview`) was a **widening, not a conversion**. The literal it replaced admitted exactly one lane; on a workflow that assigns those traits to separate columns, the union would bounce a card out of a plan-review or human-approval lane on a PR review that has nothing to do with it.

Now resolves the single role that means "this is the PR/merge stage" — `mergeOrchestration` — which is also what `resolveLifecycleColumns` keys its `review` role on, so this handler and the core resolver agree instead of drifting.

New case pins it: a board declaring a `human-review`/`merge-blocker` column that is **not** the merge-orchestration lane, a shape the default board cannot express. Revert measured — restoring the union makes it fail, the handler admits the review and moves a card out of a lane it was never responsible for.

One consequence worth stating rather than hiding: a custom workflow whose review lane carries **only** `human-review` resolves nothing here and falls back to the legacy `in-review`. That is pre-existing, not a regression — it is the gap already recorded in `notification-renamed-lifecycle-columns.test.ts`, and it belongs in the shared resolver rather than being re-invented per handler.

Changeset added here too (`patch`, `category: fix`).

gsxdsm:
Disproved — and narrowing this would be a regression, not a fix.

**The arity rule this program settled on** (recorded on `resolveReviewColumns` after it recurred in eight files): a SET answers "is this card ALREADY there" (membership); a SINGLE id answers "where should this card GO" (a move target must be one column). The narrow `lifecycle.review` is required where the caller admits and then uses *that same lane* to decide the destination — over-admitting there produces a wrong move.

That is not this site. The move target is resolved independently:

```
const wipLanes = columnsWithFlag(ir, "countsTowardWip");
if (wipLanes.length > 0) wipTarget = wipLanes[0];
...
await this.store.moveTask(taskId, wipTarget);
```

`wipTarget` comes from `countsTowardWip` and never from the lane that admitted the card. So admitting a second `mergeOrchestration` column yields exactly the same destination as admitting the first — there is no wrong-target failure mode to protect against.

**What narrowing would actually do.** A changes-requested review arriving for a card in a board's second merge lane, or in a `humanReview`-only lane, would hit the early return: no steering comment, no review item, no move. The reviewer's feedback is silently dropped and the card sits in review as though nothing happened. That is strictly worse than the status quo, and it is the same silent-skip failure this conversion exists to remove.

Behaviourally the broad set is also just correct: a reviewer requesting changes means the card needs work again, regardless of which review sub-lane it was sitting in when the event arrived.

**One thing the finding gets right**, worth keeping in view: `wipTarget` uses `wipLanes[0]`, a first-match on a board that may declare several WIP lanes. That single-id choice IS a move target, so first-match is the correct arity there — but if a board has two WIP lanes, the card always lands in the first, which may not be where it came from. That is a genuine limitation, distinct from this finding, and preserving the card's prior WIP lane would need it recorded before the review hand-off. Naming it rather than folding an unrelated change into this PR.

gsxdsm:
### Pre-flight against current `main`: 2 failures, both inherited

Full `@fusion/engine` suite on a local merge with `origin/main`:

```
Tests  2 failed | 10989 passed | 14 skipped | 1 todo (11006)

lifecycle-column-census > the baseline can always be re-recorded
  > writes the baseline BEFORE the rise check can exit
  > exits 0 from the update branch and 1 from the rise branch
```

**Red on `main` itself — not your change.** Those two assert marker ORDER in `scripts/lifecycle-column-census.mjs` source text, and a comment at line 359 quotes `"column-guard count ROSE"` verbatim, so `indexOf` finds the prose 7000 characters before the real branch. The control flow is correct and unchanged (`if (updateBaseline) {` at :487, the rise message at :510).

Fix is up in **#2825** (strips comments before indexing, same defence `archived-column-gate-parity.test.ts` already uses). Once it lands this clears with no change on your side.

PR #2809, opened 2026-07-30
fix(core): startup recovery deadlocked on its own per-task lock

gsxdsm:
## The bug

`recoverStaleTransitionPendingImpl` runs its whole per-task body inside `store.withTaskLock(id, …)`. On the PostgreSQL arm it then read the task with `store.getTask(id)` — and `getTaskImpl` opens with `store.withTaskLock(id, …)` too.

**The per-task lock is non-reentrant.** This codebase states that invariant in prose in two other files:

> "nesting inside `withTaskLock` would deadlock since the lock is non-reentrant" — `branch-and-pr-entities.ts:561`
> "because the per-task lock is non-reentrant" — `workflow-ops.ts:464`

So the sweep waited forever on a lock its own frame was holding.

**PostgreSQL-only — which is every production install.** The SQLite arm on the very next line reads through `readTaskFromDb`, a lock-free row read. The backend-mode port swapped only the PostgreSQL arm to `getTask`. The fix restores a lock-free read (`readTaskRow`) on that arm; nothing else changes.

## Why it survived until now

The branch is entered **only** when a stale marker names a plugin hook the trait registry still knows (`hasSurvivingPluginHook`). Three nearby cases all miss it:

| marker | path |
|---|---|
| none | the row is never scanned |
| only `default-workflow:postCommit` | `hasSurvivingPluginHook` false — marker just cleared |
| names an **uninstalled** plugin hook | reconciled away as degraded; nothing survives to re-run |
| names a **registered** plugin hook | **reaches the in-lock read → deadlock** |

Those first three are what the existing tests cover. The fourth is precisely the state a crash mid-hook leaves behind. All four are asserted in the new suite so the path cannot be re-narrowed and called covered.

## Impact

This sweep runs at **startup**. A task left with such a marker deadlocks startup recovery — and because it deadlocks *while holding the task's lock*, that task is also left permanently unlockable.

## How it was found, including a correction

By **bisection**, not by reading. An earlier attempt of mine to drive this recovery reported that "the sweep never returns". That was wrong in a way worth recording: the sweep returns fine in three of the four cases, and generalising the one hang to the whole function is what hid the actual trigger across several sessions. Narrowing case by case — empty store, plain task, default-only marker, unknown-hook marker, registered-hook marker — put the fault on one line.

## Verification

- **Mutation-verified against the real defect.** With the fix reverted, the regression case fails by name — `recoverStaleTransitionPendingImpl did not settle within 8000ms — deadlock` — while the other three stay green. That is the actual pre-fix behaviour, not a simulation of it.
- Every case is **timeboxed** on purpose: a deadlock otherwise surfaces as a suite-level timeout naming no case, which is useless for locating the fault. The deadline is not a flake knob — the fixed code settles in ~150 ms and the broken code never settles, so there is no value in between to tune.
- A **vacuity guard** (no markers → scans nothing) so a change that stopped listing marked rows can't leave the other cases green.
- full live-PG E2E surface — **152/152**
- `pnpm lint` — clean

### DIFF
--- a/.changeset/transition-pending-recovery-deadlock.md
+++ b/.changeset/transition-pending-recovery-deadlock.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Fix a startup hang when a task was interrupted mid column-transition hook.
+category: fix
+dev: `recoverStaleTransitionPendingImpl` ran its per-task body inside `withTaskLock(id)` and then read the task with `store.getTask(id)`, which acquires the same non-reentrant lock. PostgreSQL-only — the SQLite arm already used the lock-free `readTaskFromDb`. Restores a lock-free read (`readTaskRow`) on the backend arm. Reachable only when a stale transition-pending marker names a plugin hook the trait registry still knows.
--- a/.changeset/transition-pending-recovery-deadlock.md
+++ b/.changeset/transition-pending-recovery-deadlock.md
@@ -2,6 +2,6 @@
 "@runfusion/fusion": patch
 ---
 
-summary: Fix a startup hang when a task was interrupted mid column-transition hook.
+summary: Fix a startup hang, and a skipped plugin hook, for tasks interrupted mid column-transition.
 category: fix
-dev: `recoverStaleTransitionPendingImpl` ran its per-task body inside `withTaskLock(id)` and then read the task with `store.getTask(id)`, which acquires the same non-reentrant lock. PostgreSQL-only — the SQLite arm already used the lock-free `readTaskFromDb`. Restores a lock-free read (`readTaskRow`) on the backend arm. Reachable only when a stale transition-pending marker names a plugin hook the trait registry still knows.
+dev: `recoverStaleTransitionPendingImpl` ran its per-task body inside `withTaskLock(id)` and then read the task with `store.getTask(id)`, which acquires the same non-reentrant lock. PostgreSQL-only — the SQLite arm already used the lock-free `readTaskFromDb`. Restores a lock-free read (`readTaskRow`) on the backend arm. Reachable only when a stale transition-pending marker names a plugin hook the trait registry still knows. Also switches that recovery's IR read from `resolveTaskWorkflowIrSync` (which returns the default workflow for every task under PostgreSQL, so a custom-workflow task's interrupted hook was silently skipped) to `resolveWorkflowIrForTask`, and drops the now-unused `lifecycle-ops.ts` entry from the sync-resolver call-site allow-list.
--- a/packages/core/src/__tests__/sync-workflow-ir-callsite-allowlist.test.ts
+++ b/packages/core/src/__tests__/sync-workflow-ir-callsite-allowlist.test.ts
@@ -36,10 +36,6 @@ reason; a sync-resolved lifecycle guard is a guard that cannot fire.
 
 const ALLOWED_CALL_SITES: ReadonlyMap<string, string> = new Map([
   [
-    "packages/core/src/task-store/lifecycle-ops.ts",
-    "Synchronous lifecycle bookkeeping inside a transaction.",
-  ],
-  [
     "packages/core/src/task-store/task-store-helpers.ts",
     "Synchronous helper shared by txn-hot paths.",
   ],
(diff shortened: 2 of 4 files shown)


gsxdsm:
Fixed in the latest push — the finding is correct, and it is a **second, distinct defect on the same line**.

It was previously unreachable: the deadlock in this PR meant nothing ever got past the read. Fixing the lock exposed it.

`resolveTaskWorkflowIrSync` returns the DEFAULT workflow IR for every task under PostgreSQL. The hook runner derives its pending set from the columns of the IR it is handed, so a task on a custom workflow matched no plugin trait — the interrupted hook was never re-run and the sweep reported `recovered: 1` having done nothing.

Changes:

- `resolveWorkflowIrForTask` instead. Nothing forced the sync call — the frame is already `async` and the marker read above is already awaited, so there is no ordering to preserve.
- The `lifecycle-ops.ts` entry is **removed from the sync-resolver call-site allow-list** in the same change, so the two cannot drift. The ratchet passes (3/3), and this closes one of the six allow-listed sites.
- New case: **the interrupted hook is re-run for a task on a CUSTOM workflow.** It asserts observed state — the hook records the task id it actually ran for — on a renamed board whose `building` column carries the plugin trait and exists on no other workflow, so the default IR cannot supply it by accident.

Mutation-verified: restoring `resolveTaskWorkflowIrSync` fails **only** the new case; the other four stay green.

### One correction to my own work

My first version of that case failed, and I nearly reported the async fix as ineffective. The fault was mine: `registerTraitHookImpl` takes `(traitId, hookKind, impl)` and I was calling it with a composed `"<trait>:onEnter"` id and no kind. That registers nothing, the runner resolves the hook to a no-op, and the suite reads as "the hook never fired" — indistinguishable from the product defect. Isolated by calling `runPluginColumnTransitionHooks` directly with a known-good IR. The corrected call is annotated in the test so the next person does not lose the same hour.

5/5 green, `pnpm test:gate` exit 0, lint clean, changeset updated.

gsxdsm:
### Pre-flight against current `main`: all inherited, nothing from your change

Local merge of this branch with `origin/main`, both suites:

```
core:    1 failed | 4772 passed
engine:  5 failed | 10987 passed
```

**All six are red on `main` itself**, from #2783's bookkeeping rather than from your PR:

- the 5 engine failures (census baseline + `executor-user-cancel`) — fixed and merged in **#2814**
- the 1 core failure (`archived-column-gate-parity`) — fix up in **#2817**

Rebase once #2817 lands and this should come back clean. Nothing to do on your side.

Worth noting for the queue generally: #2783 touches only `packages/core` and `packages/dashboard/src`, but the tests it broke live in `packages/engine` — a cross-package ratchet. I had pre-flighted #2783 scoped to the packages its diff touches and reported it clean, which was wrong for exactly that reason. The suite that breaks is not always the suite the diff points at.

gsxdsm:
Rebased onto main. One conflict, in the sync-resolver call-site allow-list, and it was **two independent removals colliding** rather than a disagreement.

Main still lists `lifecycle-ops.ts` and still calls `resolveTaskWorkflowIrSync` there — that is what this PR removes. Meanwhile #2812 merged and removed `branch-and-pr-entities.ts` from the same list. Git presented them as competing versions of the same entry position, so taking either side would have silently restored the other's allow-list entry while its call site was already gone — and the ratchet's "no stale allow-list entry" case would then have gone red on main.

Resolved by removing the whole entry: neither belongs. Verified against main rather than assumed — `git show origin/main` confirms `lifecycle-ops.ts` is still listed and still uses the sync resolver, and `branch-and-pr-entities.ts` is already absent.

The allow-list is now down to four entries, from six at the start of this series:

| site | status |
|---|---|
| `task-store-helpers.ts` | still allow-listed |
| `workflow-task-create-ops.ts` | still allow-listed — legitimately correct, creation runs before any selection exists |
| `replan-target.ts` | still allow-listed |
| `scheduler.ts` | still allow-listed |
| ~~`branch-and-pr-entities.ts`~~ | closed by #2812 |
| ~~`lifecycle-ops.ts`~~ | closed by this PR |

Verified after the rebase: allow-list ratchet **3/3**, deadlock suite **5/5**, `pnpm test:gate` **exit 0**, lint clean.

PR #2811, opened 2026-07-30
fix: main is red on the lifecycle ratchet — re-record the census baseline

gsxdsm:
**`main` is RED on the lifecycle ratchet right now.** `node scripts/lifecycle-column-census.mjs --strict` exits **1** on pristine `origin/main`, which is the `Lint` job's *Lifecycle-column ratchet* step — so **every open PR fails Lint** until this lands, regardless of its own contents.

Verified on a detached checkout of `origin/main`, not on a branch of mine.

## Cause

Eight `DELIBERATE-LITERAL` markers were added across seven files without re-recording the baseline:

```
packages/core/src/task-move-disposer.ts            (in-progress, todo)
packages/core/src/task-store/archive-lifecycle-2.ts (archived)
packages/dashboard/src/github-tracking-comments.ts  (done)
packages/dashboard/src/gitlab-tracking-comments.ts  (in-progress)
packages/dashboard/src/server.ts                    (archived)
packages/dashboard/src/task-planner-chat-context.ts (done)
packages/dashboard/src/test/mockCoreEngine.ts       (in-review)
```

Adding a marker RECLASSIFIES a site (column-guard → deliberate), so the tracked deliberate totals move and `--strict` fails until the baseline records the new shape. It is the same mechanism that turned #2775 red earlier today — a marker landing without its baseline — which is worth noting because it has now happened twice from different PRs.

## The fix

Baseline re-recorded, nothing else. Zero source changes; the diff is one derived file.

- `--strict` exits **0**
- `pnpm test:gate` — **161 / 13 / 487 / 71**
- `pnpm lint` clean

## Worth a follow-up by whoever owns the ratchet

The failure is structural rather than careless: a PR that adds a marker is *doing the right thing*, and the baseline requirement is only discovered when CI goes red — after merge, for everyone else. Two options, neither of which I am taking unilaterally on a red-main fix:

1. have `--strict` treat a marker-only reclassification as an accepted rise (it is not new debt — the count of unconverted guards goes **down**);
2. or fail the PR that adds the marker, by comparing against the base ref rather than the recorded baseline — the machinery for that already exists in this script.

I would take (1): a marker is the documented way to close a site, and requiring a second mechanical step to record it is a trap that catches good behaviour.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved lifecycle census error messages to distinguish genuine increases in column-guard debt from reclassified deliberate literals.
  * Added clearer remediation guidance for reclassified results, including when to update the baseline.
  * Updated lifecycle census baseline mappings to reflect current classifications.

* **Tests**
  * Added coverage for unchanged baselines, genuine guard-count increases, and marker-only reclassification scenarios.

### DIFF
--- a/packages/engine/src/__tests__/census-reclassification-message.test.ts
+++ b/packages/engine/src/__tests__/census-reclassification-message.test.ts
@@ -1,0 +1,99 @@
+/*
+FNXC:WorkflowLifecycleColumns 2026-07-31-21:40:
+
+THE INVARIANT: a marker-only reclassification says so, instead of announcing a rise that did not
+happen.
+
+MEASURED CAUSE. `main` went red on the lifecycle ratchet TWICE in one day, from two different PRs,
+and both times for the same reason: a PR added `DELIBERATE-LITERAL` markers without re-recording the
+baseline. Adding a marker moves a site from `byFile` to `deliberateByFile`, so the recorded totals
+shift even though unconverted debt goes **down**.
+
+The failure text said `column-guard count ROSE`. That is the OPPOSITE of what happened, and it sent
+the reader hunting for a regression that does not exist — on a red main, where the cost of a
+misleading message is every other PR's Lint job.
+
+This does not loosen the ratchet: the baseline still has to be re-recorded, and the run still exits
+1. It changes what the failure TELLS you, and prints the one command that fixes it. Whether a
+marker-only change should fail at all is a policy question for the ratchet's owner; making the
+existing failure legible is not.
+
+REVERT PROOF, measured: drop the reclassification branch and the message reverts to the misleading
+"ROSE" wording for a marker-only diff.
+*/
+import { execFileSync } from "node:child_process";
+import { mkdtempSync, writeFileSync, rmSync, readFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join, resolve } from "node:path";
+import { describe, expect, it, afterEach } from "vitest";
+
+const REPO_ROOT = resolve(import.meta.dirname, "../../../..");
+const SCRIPT = join(REPO_ROOT, "scripts/lifecycle-column-census.mjs");
+const REAL_BASELINE = join(REPO_ROOT, "scripts/lib/lifecycle-column-census-baseline.json");
+
+let scratch: string | undefined;
+
+afterEach(() => {
+  if (scratch) rmSync(scratch, { recursive: true, force: true });
+  scratch = undefined;
+});
+
+function runWithBaseline(mutate: (baseline: Record<string, unknown>) => void): { status: number; stderr: string } {
+  scratch = mkdtempSync(join(tmpdir(), "fusion-census-msg-"));
+  const baselinePath = join(scratch, "baseline.json");
+  const baseline = JSON.parse(readFileSync(REAL_BASELINE, "utf8")) as Record<string, unknown>;
+  mutate(baseline);
+  writeFileSync(baselinePath, JSON.stringify(baseline, null, 2));
+  try {
+    execFileSync("node", [SCRIPT, "--strict"], {
+      cwd: REPO_ROOT,
+      env: { ...process.env, FUSION_CENSUS_BASELINE_PATH: baselinePath },
+      encoding: "utf8",
+      stdio: ["ignore", "pipe", "pipe"],
+    });
+    return { status: 0, stderr: "" };
+  } catch (err) {
+    const e = err as { status?: number; stderr?: string };
+    return { status: e.status ?? 1, stderr: e.stderr ?? "" };
+  }
+}
+
+describe("the ratchet distinguishes a reclassification from a regression", () => {
+  it("names a marker-only diff as a reclassification and prints the fix", () => {
+    /*
+    Simulate the shape that turned main red: the tree has a marker the baseline has not recorded, and
+    the same file's guard count has fallen. Done by REMOVING deliberate entries from a copy of the
+    real baseline, which is exactly what a PR that forgot to re-record leaves behind.
+    */
+    const { status, stderr } = runWithBaseline((baseline) => {
+      const deliberate = baseline.deliberateByFile as Record<string, number>;
+      for (const key of Object.keys(deliberate).slice(0, 3)) delete deliberate[key];
+    });
+
+    expect(status).toBe(1);
+    expect(stderr).toContain("RECLASSIFIED as DELIBERATE-LITERAL");
+    expect(stderr).toContain("Unconverted debt went DOWN");
+    expect(stderr).toContain("--update-baseline");
+    // The misleading wording must be gone for this case.
+    expect(stderr).not.toContain("column-guard count ROSE");
+  });
+
+  it("still says ROSE when guards genuinely grew", () => {
+    // The ratchet's real job must keep its real message — a friendlier failure is not the goal.
+    const { status, stderr } = runWithBaseline((baseline) => {
+      const byFile = baseline.byFile as Record<string, number>;
+      const key = Object.keys(byFile).find((k) => (byFile[k] ?? 0) > 1);
+      if (key) byFile[key] = 0;
+    });
+
+    expect(status).toBe(1);
+    expect(stderr).toContain("column-guard count ROSE");
+  });
+
+  it("passes against the unmodified baseline", () => {
+    // A guard that always fails is no guard — and this one shells out, so it is worth proving.
+    const { status } = runWithBaseline(() => undefined);
+
+    expect(status).toBe(0);
+  });
+});
--- a/packages/engine/src/__tests__/census-reclassification-message.test.ts
+++ b/packages/engine/src/__tests__/census-reclassification-message.test.ts
@@ -61,9 +61,17 @@ function runWithBaseline(mutate: (baseline: Record<string, unknown>) => void): {
 describe("the ratchet distinguishes a reclassification from a regression", () => {
   it("names a marker-only diff as a reclassification and prints the fix", () => {
     /*
-    Simulate the shape that turned main red: the tree has a marker the baseline has not recorded, and
-    the same file's guard count has fallen. Done by REMOVING deliberate entries from a copy of the
-    real baseline, which is exactly what a PR that forgot to re-record leaves behind.
+    Simulate the shape that turned main red: the tree has a marker the baseline has not recorded.
+    Done by REMOVING deliberate entries from a copy of the real baseline, which is exactly what a PR
+    that forgot to re-record leaves behind.
+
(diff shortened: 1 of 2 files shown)


gsxdsm:
## Added the follow-up I flagged: the failure now says what actually happened

Second commit on this PR. The first unblocks `main`; this stops the next occurrence from being a mystery.

**The message was actively misleading.** `column-guard count ROSE` is the *opposite* of what a marker-only change does — unconverted debt goes **down**. On a red main, that sent every reader hunting for a regression that does not exist. Main hit this twice in one day from two different PRs, and both times the text pointed away from the fix.

Now a marker-only diff reports:

```
lifecycle-column-census --strict: sites were RECLASSIFIED as DELIBERATE-LITERAL
  packages/dashboard/src/server.ts (DELIBERATE-LITERAL: archived) (reclassified, not new debt): 0 -> 1
  …
Unconverted debt went DOWN — a marker moved these sites out of the guard count.
The baseline records both totals, so it must be re-recorded in the same change:

```

**This does not loosen the ratchet.** The baseline still must be re-recorded and the run still exits 1. Whether a marker-only change should fail *at all* is a policy question for the ratchet's owner — I raised both options on the first commit and am not deciding it here. Making an existing failure legible is not a policy change.

### Two bugs in my own detection, caught by its test before it shipped

Both would have made the guard **silently never fire** — the exact failure mode this program exists to find, in the code meant to find it:

1. `deliberateByFile` is keyed `file\0columnId`, `byFile` by plain path. My first lookup used the suffixed key against the plain map and always read `0`.
2. Then I asked "did the guard count FALL by at least the marker rise". It usually **cannot**: a file taken to zero guards loses its `byFile` entry entirely, so both sides read `0` and no fall is observable at check time — the fall happened in an earlier re-record.

The honest condition is weaker and correct: **markers rose and guards did not.** It cannot mask real regrowth, because a file whose guard count also rose is still reported as a rise — asserted by a case that keeps the `ROSE` wording for a genuine regression.

3 tests: the reclassification message, the genuine-rise message, and a control proving the whole thing still passes against the unmodified baseline. The control matters here because the suite shells out to the real script; without it every assertion could pass on a script that always fails.

gsxdsm:
Real, and the equality case is the *common* one — fixed in `3c41669`.

`reclassified` is `guardsNow <= guardsBefore`, so it is TRUE when the guard count is **unchanged**. That is exactly what adding a `DELIBERATE-LITERAL` marker to a site the parser already excludes produces, and exactly what this PR's own fixture creates by deleting only deliberate entries. So the branch that fires most often was claiming a decrease that had not happened — a second wrong number in a message whose entire purpose is to stop the reader chasing one.

Now reads *"Unconverted debt did NOT increase"*, which is accurate for both `<` and `=`.

**The test's comment was wrong in the same way**, which is why the assertion looked right: it said the fixture makes *"the same file's guard count fall"*. It does not — deleting deliberate entries leaves the count untouched. Corrected, and the test now also asserts the decrease claim is **absent**, so the wording cannot drift back. Mutation-verified: restoring "went DOWN" fails it.

---

**A correction I owe this PR.** I initially measured `--strict` on this branch, got exit 1, and was about to report that #2811 does not actually clear the red main. That was wrong, and the cause was mine: an earlier cross-branch `git checkout origin/main -- scripts/lib/lifecycle-column-census-baseline.json` had left the file **staged** in my index, so I was measuring main's baseline against this branch's tree.

On a clean checkout of this branch: `--strict` exits **0** and `census-reclassification-message.test.ts` passes 3/3. This PR does clear the red.

I have also **closed #2813**, which I had opened for the same red main before finding this one. Two PRs editing the same ratchet file would conflict and risk one silently reverting the other's entries — precisely the failure a ratchet exists to prevent. #2811 is the better change: it fixes the misleading message and moves `--update-baseline` above the rise exit so a reclassified file can actually be re-recorded, which #2813 did not do.

One thing from #2813 worth carrying regardless: **`pnpm test:gate` does not run the census strict check.** Every gate run on the PR that turned main red (#2783, mine) was green. The check lives in `census-baseline-corruption-guard.test.ts`, outside the `engine-core` allow-list, so nothing in the merge path exercises it.

gsxdsm:
## Second query-class conversion, and it needed TWO fixes — the second only visible once the first landed

`handoff/stale-task-reporter-query-lanes` (supersedes `handoff/query-class-first-conversion`; it carries the shared helper too). Gate **161 / 13 / 487 / 71**, lint clean, `--strict` exits 0.

`stale-task-reporter.ts` has a census count of **ZERO** — no lifecycle comparison anywhere — and was completely inert on a renamed board. Both `listTasks({ column })` reads returned empty, so no stale-task signal was ever raised, on exactly the board where work is most likely sitting unnoticed.

### The finding: fixing the query moved the failure one frame deeper

With the query widened, the case **still** reported zero. `getTaskAgeStalenessSignal` takes an optional `lifecycle` and defaults to the legacy pair, so a card the query now returned was refused inside the signal.

**Converting only the query would have looked like a fix and changed nothing** — `surfaced: 0` before and after. My test caught it only because it asserts the OUTCOME rather than the query argument. That is worth weighing against #2800's choice to assert the argument: asserting the argument is right when you are *pinning a known defect* (the outcome is 0 either way), and wrong when you are *proving a fix*, because the outcome is the only thing that distinguishes a real conversion from a deeper one.

The two halves need **different shapes**, which is the reusable point:

- the **read** is project-wide (`resolveProjectColumnsForRoles`) — there is no task in hand yet;
- the **per-card signal** is per-task (`resolveTaskLifecycleColumns`) — a board spans workflows.

Using the project union for the per-card question would have marked a card stale because *another* workflow calls its column wip. That is the flat-set mistake this program has made four times, and the helper's own header warns against exactly this.

### A test premise of mine that was simply wrong

I first asserted that a card in `in-progress` is surfaced on the RENAMED board, reasoning that the query unions the legacy ids. The query does — but the per-task signal then correctly **refuses** it, because that card's own workflow does not call `in-progress` a wip lane. **The product was right and my premise was wrong.**

What the union actually buys is that the row is *fetched*; whether it is stale is then a per-task question. The case now covers the guarantee that genuinely matters — a store with no workflow selection, where both halves fall back together.

That is the fourth fixture/premise error I have made in this sweep (`createdAt` epoch-ms, `not.toHaveBeenCalled()` leakage, the running-gate `testId`, and this). All four were caught by running the test, none by reading the code. The rule I would hand on: **when a new case fails, the fixture and the premise are the likelier culprits than the product** — my hit rate on "the product is wrong" has been about one in five.

### Verification

7 cases in the reporter suite, 8 on the shared helper. Revert proof: restoring either literal read makes the renamed case surface zero.

### Still not done

`self-healing.ts`'s 49 — actively owned and mid-conversion. `project-engine.ts` (7), `restart-recovery-coordinator.ts` (2), `executor.ts` (2). Each needs its own read of what the sweep does with the rows, and this conversion is the reason why: the query is rarely the whole defect.

gsxdsm:
## Rebased — `main` is green again, so this PR is now purely the message fix

The baseline re-record **dropped out on rebase**: `main` picked up the same fix from another PR while this sat in review. That is the right outcome and I am not re-adding it — two identical baseline re-records would be churn, and the urgent half of this PR has served its purpose.

What remains is the part that stops the recurrence: a marker-only diff now says **RECLASSIFIED** and prints the recovery command, instead of announcing a `column-guard count ROSE` that did not happen. `main` went red on this twice in one day from two different PRs, and both times the failure text pointed away from the fix.

Verified after the rebase: `--strict` exits 0, guard suite **3 passed**, gate **161 / 13 / 487 / 71**, lint clean.

## On the wording correction that landed here

Someone pushed `"did NOT increase", not "went DOWN" — the equality case is the common one` onto this branch, and it is a straight catch on my own text. My condition is `guardsNow <= guardsBefore`, which **includes equality** — a marker added to a site that was already outside the guard count leaves debt unchanged, and that is the common case, not a decrease. I asserted a decrease the code does not establish.

Worth naming because it is the same error class this PR exists to fix, one level up: **a diagnostic that overstates what it knows is still a misleading diagnostic.** I replaced "ROSE" (wrong direction) with "went DOWN" (also unverified) — the correction replaced it with what the condition actually proves.

## Scope note

I have deliberately not re-expanded this PR. The query-class conversions I have been working on live on their own branches (`handoff/stale-task-reporter-query-lanes`, which carries the shared `resolveProjectColumnsForRoles` helper plus two worked conversions), because bundling tooling fixes with product conversions is what made the last three review rounds on `agent-assignment` necessary.

gsxdsm:
## Third query-class conversion: restart recovery did not run at all on a renamed board

`handoff/restart-recovery-query-lanes` (carries the shared helper). Gate **161 / 13 / 487 / 71**, lint clean, `--strict` exits 0.

`recoverInterruptedRuns` carried a note I had written earlier **flagging** this as the query class and declining to convert it. The note was right about the diagnosis: the `listTasks({ column })` QUERY was the live filter and the `.filter` beneath it a redundant re-assertion, so converting the predicate alone would have dropped a census count and changed nothing. On a renamed board the sweep never ran — **an engine restart left interrupted tasks stuck with no requeue**.

### Three layers, and naming them is the point

The previous two conversions in this class each hid a second layer behind the first, so I looked for all of them before touching anything:

1. **The query** — fixed, project-level (`resolveProjectColumnsForRoles`), because no task is in hand before the read.
2. **The redundant `.filter`** — **deleted**, not converted. Re-asserting the column the query just selected on adds nothing, and a second copy of a rule is how a read and its filter drift apart. A test pins the one thing it did contribute: the `paused` guard.
3. **The move destination** — already resolved via `resolveReboundTargetForTask`. Only its comment was stale, still warning about a hardcoded `todo` long after the fix landed. Corrected in place rather than deleted, because the *reason* it matters is still true.

Layer 3 is worth calling out separately: **a stale warning comment is its own hazard.** It told the next reader a defect existed where none did, and I nearly re-fixed it before checking the call.

### A mistake I made building this

I pulled `packages/core/src/index.ts` wholesale from my own handoff branch to get the helper export, and **clobbered two exports `main` had added since** (`resolveReboundTargetForTask`, `resolveWipTargetForTask`) — three packages stopped compiling. Restored from `main` and re-applied only my one line.

Same root cause as the missing-test-file slip earlier: **taking a file from another branch takes its whole contents, including whatever is now out of date.** For a barrel file that is nearly always wrong; the fix is to re-apply the single edit on top of the current version. `tsc` caught it immediately, which is why the branch is green — but it would have been a broken build for everyone if the barrel had not been type-checked.

### Verification

12 cases in the suite (2 new), revert-proven: restoring the literal query requeues nothing on the renamed board.

### Remaining in this class

`self-healing.ts` (49, actively owned), `project-engine.ts` (7), `executor.ts` (2). Three worked examples now exist — `backlog-pressure-reporter`, `stale-task-reporter`, and this one — and the pattern is stable: resolve the roles, iterate the set, dedupe by id, then check whether a per-task guard downstream needs the per-task resolver too.

PR #2812, opened 2026-07-30
fix(core): resolve the task's own terminal node in the node-override guard

gsxdsm:
## What

Fixes the defect **#2793 measured** but deliberately did not fix. That PR's two characterization tests are flipped here to assert the correct behaviour — which is what they were written to do.

## The bug

`updateTask({ nodeId })` passes through `validateNodeOverrideChange` **twice**, and both calls were wrong in different ways:

| call | how it answered "is this terminal?" |
|---|---|
| `branch-and-pr-entities.ts:568` | `resolveTaskWorkflowIrSync` — the **default** workflow for every task under PostgreSQL |
| `task-update.ts:53` | no options at all → `defaultIsTerminalNodeId`, the bare literal `nodeId === "end"` |

On a board whose terminal node is not named `end`, the FN-7641 guard inverted in both directions:

- an override to a **non-terminal** node that happens to be named `end` was **rejected** with a merge-proof error about finalizing a card the operator was not finalizing;
- an override to the board's **real** terminal node was **written verbatim**, no error, card unadvanced — the silent no-op FN-7641 exists to prevent.

## The fix

A new `isTaskTerminalNodeIdAsync` resolves the task's own workflow, with the **identical** literal fail-soft for an unresolvable one. Both call sites use it — pre-resolved, because `validateNodeOverrideChange`'s callback is synchronous and it asks the question at most once.

**Nothing forced the sync call at either site**: both frames are already `async` and already awaiting. That is the same finding as #2809's review, one file over.

The sync helper is **deleted, not kept as a fallback**. Keeping both would re-create the half-conversion this program keeps finding — one caller resolved, one not, and no way to tell from a call site which it got. `branch-and-pr-entities.ts` also leaves the sync-resolver call-site allow-list (ratchet green, 3/3), the **second** of the six allow-listed sites to close.

## Why both halves were needed — and how that is proven

#2793's mutation matrix showed the rejected-`end` case is **over-determined**: both guards independently called it terminal, so correcting either one alone changed nothing an operator could see. That is why fixing only the allow-listed sync site would have looked like progress and delivered none.

Re-measured here, on the fixed tree:

| state | result |
|---|---|
| both guards fixed | **3/3 pass** |
| inner guard reverted to no-options | **1 fails** |
| outer guard reverted to the literal | **1 fails** |

## Tests

#2793's two cases now assert the fixed behaviour and keep their reasoning:

- the non-terminal `end` override is **written**, and the card stays in review — a routing change, not a finalize;
- the real terminal `finish` override is **refused** without merge proof, **and** the field is not written on the way to refusing.

The fixture-integrity case (`finish` is the end node, `end` is not) is unchanged — it is what stops both assertions passing for the wrong reason.

## Verification

- terminal-node suite — **3/3**, both-halves matrix above
- sync-resolver call-site allow-list ratchet — **3/3**
- full live-PG E2E surface — **151/151**
- `pnpm lint` — clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected workflow override validation when the terminal node uses a custom identifier instead of `"end"`.
  * Non-terminal overrides are now accepted correctly, while terminal overrides continue to require appropriate merge confirmation.
  * Updated task routing and persisted node values to reflect the corrected validation behavior.

* **Tests**
  * Updated end-to-end coverage for workflows with non-standard terminal node identifiers.

### DIFF
--- a/.changeset/terminal-node-resolution.md
+++ b/.changeset/terminal-node-resolution.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Fix node-override handling on workflows whose terminal node is not named "end".
+category: fix
+dev: `updateTask({ nodeId })` passes through `validateNodeOverrideChange` twice. The outer call resolved terminality via `resolveTaskWorkflowIrSync` (the default workflow under PostgreSQL); the inner call passed no options and fell to the literal `nodeId === "end"`. Both now resolve the task's own workflow via the new `isTaskTerminalNodeIdAsync`, and `branch-and-pr-entities.ts` leaves the sync-resolver call-site allow-list.
--- a/packages/core/src/__tests__/sync-workflow-ir-callsite-allowlist.test.ts
+++ b/packages/core/src/__tests__/sync-workflow-ir-callsite-allowlist.test.ts
@@ -36,10 +36,6 @@ reason; a sync-resolved lifecycle guard is a guard that cannot fire.
 
 const ALLOWED_CALL_SITES: ReadonlyMap<string, string> = new Map([
   [
-    "packages/core/src/task-store/branch-and-pr-entities.ts",
-    "Inside a synchronous store entity read.",
-  ],
-  [
     "packages/core/src/task-store/lifecycle-ops.ts",
     "Synchronous lifecycle bookkeeping inside a transaction.",
   ],
--- a/packages/core/src/task-store/branch-and-pr-entities.ts
+++ b/packages/core/src/task-store/branch-and-pr-entities.ts
@@ -20,6 +20,7 @@ import { getWorkflowWorkItem as getWorkflowWorkItemAsync } from "./async-workflo
 import { MergeRequestRow, PrEntityRow, WorkflowWorkItemRow } from "./row-types.js";
 import { BranchGroup, BranchGroupCreateInput, ColumnId, MergeRequestRecord, MergeRequestState, PrEntity, PrEntityCreateInput, PrThreadOutcome, PrThreadState, RunMutationContext, Task, TaskLogEntry, TaskPriority, TaskVerificationRequest, TaskVerificationResultSummary, TaskVerificationStatus, WorkflowWorkItem, WorkflowWorkItemKind, WorkflowWorkItemState, WorkflowWorkItemTransitionPatch } from "../types.js";
 import { validateNodeOverrideChange, resolveNodeOverrideLanes } from "../node-override-guard.js";
+import { isTaskTerminalNodeIdAsync } from "../workflow-ir-resolver.js";
 import { WorkflowMovePolicyInput } from "../workflow-extension-types.js";
 import { resolveWorkflowIrById } from "../workflow-ir-resolver.js";
 import { resolveTaskLifecycleColumns } from "../workflow-lifecycle-traits.js";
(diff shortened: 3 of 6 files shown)


gsxdsm:
The finding is correct, and chasing it turned up a separate defect that blocks the obvious remedy. Latest push records the residual rather than papering over it.

**The finding stands.** `resolveWorkflowIrForTask` *degrades* to the default workflow rather than throwing, so the `try/catch` never fires on a failed lookup, and a task on a custom board whose definition cannot be loaded is judged against the default graph.

**The natural fix does not work yet, and I have the measurement.** Gating on `resolveWorkflowIrForTaskWithProvenance(...).source === "selection"` should be exactly right. I implemented it, and it turned the *non-terminal override is written* case **red**. The reason:

```
store workflow id = WF-001   stored ir.id = custom:prov
PROVENANCE source = default  workflowId = undefined  resolved ir.id = custom:prov
```

`createWorkflowDefinition` stores an authored IR **verbatim**, so `ir.id` keeps whatever the author wrote while the store allocates its own `WF-NNN`. The provenance identity check compares those two, finds them different, and reports `source: "default"` — **for a workflow it resolved correctly** (note `resolved ir.id = custom:prov`, the right IR).

That is the signal failing in the unsafe direction for its stated purpose. Its own FNXC note says a lying signal "is one nobody can build the census conversions on", and any conversion gated on `source === "selection"` silently does not take effect for such workflows.

**Decision for this PR:** keep the literal fail-soft and document it, with the measurement, next to the code. The residual is the *pre-existing* behaviour — this PR does not introduce it, and it is strictly narrower than what shipped before, since the resolved-selection path (the overwhelming majority) is now correct where it was previously always-default. Narrowing it further depends on fixing provenance first, which is a change to `resolveWorkflowIrForTaskWithProvenance` and its own review, not a rider here.

I am picking up that provenance defect next as its own PR; this comment is the reference for it.

3/3 green, lint clean.

gsxdsm:
Rebased onto main. The conflicts were substantive, not textual — **#2821 rewrote both call sites this PR touches**, so I merged rather than took a side.

What main gained at each site, all preserved:

- `branch-and-pr-entities.ts` — `resolveNodeOverrideLanes` hoisted **before** the task read, and `...overrideLanes` spread into the guard options.
- `task-update.ts` — the same lane resolution, plus a lock-free `readTaskJson` re-read of the column after the await, which #2821 introduced to fix a deadlock of exactly the class I hit in #2809/#2810: `store.getTask()` inside `updateTaskUnlockedImpl`, whose caller already holds the non-reentrant per-task lock.

My change adds the resolved `isTerminalNodeId` answer. The two are **independent questions about the same call** — `overrideLanes` answers "which lanes count as WIP/review for the mid-flight refusal", mine answers "is this node the board's terminal one" — so they are threaded together:

```ts
const validation = validateNodeOverrideChange(freshForGuard ?? task, updates.nodeId ?? null, {
  ...overrideLanes,
  isTerminalNodeId: () => terminal,
});
```

Dropping either re-opens a defect the other does not cover, which is why I did not resolve the conflict by preferring one side. Noted in the FNXC comment at the site so the next rebase does not have to re-derive it.

Verified after the rebase: terminal-node suite **3/3**, sync-resolver call-site allow-list ratchet **3/3** (the `branch-and-pr-entities.ts` entry stays removed), `pnpm test:gate` **exit 0**, lint clean.

PR #2813, opened 2026-07-30
fix(census): main is red on --strict — record the 8 deliberate entries #2783 merged without

gsxdsm:
## main is currently red on the census ratchet

`node scripts/lifecycle-column-census.mjs --strict` fails on `origin/main` with **eight rises**. All eight are `DELIBERATE-LITERAL` markers that my own PR #2783 (`74cba4b46d`) added to source without the matching baseline entries, so the ratchet has been red since that merge.

| file | id |
|---|---|
| `core/src/task-move-disposer.ts` | `in-progress`, `todo` |
| `core/src/task-store/archive-lifecycle-2.ts` | `archived` |
| `dashboard/src/github-tracking-comments.ts` | `done` |
| `dashboard/src/gitlab-tracking-comments.ts` | `in-progress` |
| `dashboard/src/server.ts` | `archived` |
| `dashboard/src/task-planner-chat-context.ts` | `done` |
| `dashboard/src/test/mockCoreEngine.ts` | `in-review` |

### Why it got through, which matters more than the fix

**`pnpm test:gate` does not run the census strict check.** Every gate run on #2783 was green, and I checked the gate before each push. The check lives in `census-baseline-corruption-guard.test.ts`, which is not in the `engine-core` allow-list, so nothing in the merge path exercised it.

That is worth knowing independently of this PR: a green gate is not evidence the ratchet is intact, and the only reason I found it was a rebase on an unrelated branch surfacing the same eight rises.

### The change

Purely additive — 8 keys into `deliberateByFile`. Verified by diffing the **parsed object** against `origin/main`, not the text: `removed=0, lowered=0, added=8` across all three sections. An earlier attempt re-serialised the file and produced a 9/-20 diff that *looked* like it dropped 19 entries; it was reformatting, but I redid it as a textual insert so the diff is 8 lines and reviewable.

All eight are reviewed markers — foreign vocabularies (`ReportStatus`-style enums, a step status, a transition kind, a test double) and one documented fast path — so recording them is the correct resolution rather than reverting the markers.

### Verification

- `--strict` → exit 0
- `census-baseline-corruption-guard.test.ts` → 3 passed
- `pnpm test:gate` → 161 + 487 + 13 + 71
- `pnpm lint` → 0 errors


gsxdsm:
Superseded by #2811, which fixes the same red ratchet and goes further — it also corrects the census script's reclassification message and adds a test for it.

Two PRs editing the same baseline file would conflict with each other and risk one silently reverting the other's entries, which is precisely the failure mode a ratchet exists to prevent. Closing mine.

Recording what I measured, since it corroborates #2811 independently: `--strict` on `origin/main` fails with exactly **8 rises**, all `DELIBERATE-LITERAL` markers that my own #2783 merged without their baseline entries. The list is in this PR's description if it is useful for cross-checking coverage.

One thing worth carrying over regardless of which PR lands: **`pnpm test:gate` does not run the census strict check.** Every gate run on #2783 was green. The check lives in `census-baseline-corruption-guard.test.ts`, outside the `engine-core` allow-list, so nothing in the merge path exercises it — which is why this reached main at all.

PR #2815, opened 2026-07-30
fix(core): brand the unregistered built-in workflow fallback

gsxdsm:
> **Corrected after opening.** The first version of this PR claimed the id cross-check reported `"default"` for *every* authored workflow and stranded cards in triage recovery forever. That was wrong, and the correction is below in full. The reachable defect is the branding hole; removing the id check is correctness, not a live bug fix.

## The bug (reachable)

`resolveWorkflowIrById` has four ways to substitute the default coding IR. Three brand the result via `markFellBack`. The fourth did not:

```ts
if (isBuiltinWorkflowId(workflowId)) {
  const builtin = getBuiltinWorkflow(workflowId);
  const ir = builtin?.ir ?? defaultCodingWorkflowIr();   // <- unmarked substitution
```

An id that *looks* built-in but is not registered — a workflow removed between releases, a typo'd selection — lands there and silently gets the default coding IR. `resolveWorkflowIrForTaskWithProvenance` then reports **`source: "selection"`** for it, handing a caller the default board's graph under the selected workflow's name. That is precisely the lying signal the API exists to prevent.

Found by review on this PR (thanks — see the thread), not by me.

## The correction to my own claim

I also deleted an id cross-check that ran after the marker check and reported `"default"` when the resolved IR's `id` differed from the requested workflow id. I justified that by saying it misfired for every authored workflow, because `createWorkflowDefinition` stores an IR verbatim while minting `WF-NNN` separately:

```
store workflow id = WF-001   stored ir.id = custom:prov
PROVENANCE source = default  resolved ir.id = custom:prov      <- the CORRECT IR, called a guess
```

**That measurement is real but it came from this suite's own fixture.** Neither `WorkflowIrV1` nor `WorkflowIrV2` declares an `id` field. An editor-authored workflow carries none, so `resolvedId` is `undefined` and the check passed it as `"selection"` — the misfire never reached `triage.ts`'s post-U11 intake recovery, the one production consumer. My "declined forever, not deferred" claim does not hold.

Removing the check is still right: it is **unreliable** (it interrogates a property the IR types do not declare, and when one is present it is the author's id, unrelated to the store-minted row id) and **redundant** (all four substitutions are now branded, and the marker is checked first). But it is a cleanup, not a fix — and saying otherwise is how a narrow change gets backported as a critical one.

## Tests

| case | asserts |
|---|---|
| unregistered **built-in** id | **`source: "default"`** — the reachable defect |
| missing definition | still `"default"` |
| no selection | still `"default"` |
| IR carrying its own id | `"selection"`, with the id mismatch asserted explicitly so it can't pass for the wrong reason |
| the resolved IR is the task's own board | contains the renamed hold column, not a default-board id |

**Mutation-verified:** removing only the branding fails the unregistered-builtin case and nothing else — which shows it covers that specific hole rather than overlapping the others. Restoring the id check fails only the carries-its-own-id case, leaving both fallback cases green, which shows the deletion did not widen trust.

## Blast radius

`triage.ts:1211` is the only production consumer of the provenance result across core, engine, dashboard and CLI. Every other `.source ===` hit is an unrelated field on an unrelated type; the two index hits are re-exports.

## Verification

- new suite — **5/5**, mutation-verified in both directions
- `pnpm lint` — clean

Changeset included (`patch`, category `fix`), rewritten to describe the branding fix rather than the overstated one.


gsxdsm:
Blast-radius audit, since the PR body describes one consequence and a reviewer will reasonably ask whether there are others.

**`triage.ts:1211` is the only production consumer of the provenance result.** Swept `packages/core`, `packages/engine`, `packages/dashboard` and `packages/cli` for `resolveWorkflowIrForTaskWithProvenance`, the `ResolvedWorkflowIr` type, and every `.source ===` comparison. The other `.source ===` hits are unrelated fields on unrelated types — `memory-insights`, `agent-store`, `planner-intervention`, `workflow-merge-proof`, `backup-settings-migration` — and the two `index.ts` / `index.gate.ts` hits are re-exports, not uses.

So the fix cannot change behaviour anywhere except the intake recovery described in the body, and the four new cases cover that signal directly. The narrow surface is also why the id check could sit wrong for as long as it did: exactly one caller depended on it, and its failure mode was a card that quietly never recovered rather than an error anyone would see.

Worth noting for whoever picks up the follow-up: with this merged, `source === "selection"` becomes usable for the first time, which unblocks narrowing the fail-soft residual documented in #2812.

gsxdsm:
Fixed in the latest push, and the finding **corrects a claim I made in this PR** — worth stating plainly because the claim was the justification for the change.

I argued the id cross-check was redundant "because all three degradation paths brand the IR via `markFellBack`". There is a **fourth**, and it was the one path never branded:

```ts
if (isBuiltinWorkflowId(workflowId)) {
  const builtin = getBuiltinWorkflow(workflowId);
  const ir = builtin?.ir ?? defaultCodingWorkflowIr();   // <- unmarked substitution
```

An id that *looks* builtin but is not registered — a workflow removed between releases, a typo'd selection — lands there, silently gets the default coding IR, and with the id check deleted would have been reported as `source: "selection"`. That is exactly the lying signal this API exists to prevent, arriving through the one door I had not checked. The id comparison did catch it, because the default's id differs from the requested one.

**Repaired at the cause, not by reverting.** The substitution is now branded where it happens, so the marker check covers every path. Restoring the id check instead would have re-introduced the misfire on every authored workflow — the actual subject of this PR — so the two are not interchangeable fixes.

New case: *a selection naming an UNREGISTERED BUILTIN id is still reported as default.* Mutation-verified — removing only the branding fails that case and nothing else, which is what shows it covers the specific hole rather than overlapping the existing ones.

5/5 green, `pnpm test:gate` exit 0, lint clean, changeset updated to describe four paths rather than three.

gsxdsm:
Confirmed independently, and **the fourth-path fix alone is not sufficient** — fixed in `3c6ba67`.

I hit this from the other direction: I wrote a test for the unregistered-builtin path, applied the fourth-path mark, and the test **still passed with the mark deleted**. It was unfalsifiable, and chasing why produced the same conclusion this finding reached.

`resolveDefaultWorkflowIr()` returns a **shared object** — measured, `a.ir === b.ir` across two independent resolutions — so `markFellBack` branded the singleton itself. Consequences:

- after any task anywhere hits a fallback, every later resolution of `builtin:coding` reports `source: "default"`, **including a task that genuinely selected the default workflow**;
- it is permanent for the process and invisible: the brand is non-enumerable, so no dump, log, or deep-equal comparison shows it;
- and it silently disarmed the new test, because the object under assertion had already been branded by an earlier case in the same file.

`markFellBack` now brands a **shallow copy**. The IR stays structurally identical — the property is non-enumerable and the clone is deep-equal to the original, which is the property the brand's own note relies on — while the fallback gets its own object to carry the fact.

**Rebased onto `0b8b1ae` rather than pushing over it.** Your fourth-path version is the one that survived: the comment is better than mine, particularly the part conceding that deleting the id cross-check was justified on a premise this path falsifies. I kept only what was additive — the copy fix and two tests.

**Both are now mutation-verified independently**, which is the point:

| mutation | result |
|---|---|
| fourth path left unmarked | 1 failed |
| brand the shared object again | 1 failed |

Before the copy fix, the first of those passed.

**Also repointed a test this PR had left red.** `reports 'default' when the stored definition resolves to a DIFFERENT workflow` asserted the id cross-check that this PR deletes. The expectation is inverted, not the code: `createWorkflowDefinition` stores an authored IR verbatim, so `ir.id` differs from the allocated `WF-NNN` for every custom workflow — which is exactly the misfire that motivated the deletion.

Suite 14 passed, gate 161 + 487 + 13 + 71, `tsc` and lint clean.

PR #2818, opened 2026-07-30
fix: the query class — three sweeps that never ran on a renamed board

gsxdsm:
Three sweeps that **never ran at all** on a renamed board, plus the shared answer the rest of the class needs. Consolidated from three handoff branches so the helper appears once. #2811 merged, so this is my only open PR.

`#2800` measured this class and shipped evidence deliberately without conversions: `listTasks({ column: "<literal>" })` filters in the store, so on a renamed board the read returns an **empty array** and the sweep it feeds does nothing. The census scores the comparison *inside* the loop, never the query above it.

## What was broken

| file | census count | what actually happened on a renamed board |
|---|---|---|
| `backlog-pressure-reporter.ts` | **0** | both reads empty, ratio computed as 0/0 — **the alert never fired**, on a board that may be under exactly the pressure it reports |
| `stale-task-reporter.ts` | **0** | both reads empty — **no stale-task signal ever raised**, where work is most likely sitting unnoticed |
| `restart-recovery-coordinator.ts` | flagged | sweep never ran — **an engine restart left interrupted tasks stuck with no requeue** |

Two of the three have a census count of **zero**. They contain no lifecycle comparison at all, so they have never appeared in the backlog, in a per-file list, or in any "N → 0" claim — and were completely inert. **A file at zero is not evidence of anything.**

## The shared answer, and what it is not

Every existing resolver answers a **per-task** question. A query has no task in hand, so it needs the project-level one: every column any workflow declares for a role, unioned with the legacy ids so a board mid-rename still finds rows under the old ones. The set is never empty, so a caller cannot accidentally query nothing.

The header states what it is **not**: answering a per-card question from the union would mark a card as review because some *other* workflow calls its column review — the flat-set mistake this program has made four times.

## The finding that generalises: the query is rarely the whole defect

`stale-task-reporter` **still reported zero after the query was fixed** — `getTaskAgeStalenessSignal` defaults to the legacy pair, so a card the query now returned was refused inside the signal. Converting only the query would have looked like a fix and changed nothing.

That is a caveat on #2800's approach, offered as refinement rather than correction: **asserting the query ARGUMENT is right when pinning a known defect** (the outcome is 0 either way) **and insufficient when proving a fix**, because the outcome is the only thing that distinguishes a real conversion from a deeper one. All three conversions here assert outcomes.

`restart-recovery` had three layers — query, a redundant re-assertion (deleted; a test pins the `paused` guard it did contribute), and a move destination that was **already** resolved but whose warning comment was stale. A stale warning is its own hazard: it told the next reader a defect existed where none did.

## Verification

- helper **8 passed** · three reporter/coordinator suites **29 passed**
- `pnpm test:gate` **161 / 13 / 487 / 71** · lint clean · `--strict` exits 0 · four `tsc` targets clean
- each conversion revert-proven independently; the failing case is named in each test header

## Two mistakes worth recording

**The helper's own test caught a bug in it.** My first draft wrapped the definition loop in one `try`, and `parseWorkflowIr` **validates** rather than parses — one malformed row would have returned legacy-only lanes for *every* workflow, indistinguishable from the bug it exists to fix. Now isolated per definition.

**I clobbered the core barrel** by taking `index.ts` wholesale from a handoff branch, dropping two exports `main` had added since; three packages stopped compiling. Taking a file from another branch takes its whole contents, including what is now stale — for a barrel that is nearly always wrong. Re-applied as a single edit on top of `main`.

## Not included

`self-healing.ts`'s 49 — actively owned and mid-conversion; an outside refactor there produces conflicting halves of one sweep. `project-engine.ts` (7) and `executor.ts` (2) need their own read of what each sweep does with the rows, which these three are the argument for.

## Summary by CodeRabbit

* **Bug Fixes**
  * Workflow automation now recognizes renamed or custom board lanes when identifying completed, active, review, and archived tasks.
  * Archiving completed tasks works reliably during and after lane renames.
  * Restart recovery, resume actions, stale-task detection, backlog alerts, and timing reconciliation now continue working with customized workflows.
  * Canceling a CLI session task now returns it to the workflow’s configured hold lane, with legacy fallback behavior preserved.

### DIFF
--- a/packages/core/src/__tests__/archive-all-done-lane.test.ts
+++ b/packages/core/src/__tests__/archive-all-done-lane.test.ts
@@ -1,0 +1,68 @@
+/*
+FNXC:WorkflowLifecycleColumns 2026-08-01-05:00:
+
+THE INVARIANT: "Archive all done" reads the board's OWN complete lane.
+
+`listTasks({ column: "done" })` filters in the STORE, so on a renamed board this returned an empty
+array and the action completed successfully having archived **zero** cards. An operator action that
+silently does nothing is worse than one that errors: the board simply looks unchanged, so the natural
+conclusion is that there was nothing to archive.
+
+Census-invisible — the literal is a query filter, not a comparison — and this file had no lifecycle
+comparison to convert at all.
+
+REVERT PROOF, measured: restore `listTasks({ slim: true, column: "done" })` and the renamed case
+archives nothing.
+*/
+import { describe, expect, it, vi } from "vitest";
+import { archiveAllDoneImpl } from "../task-store/task-artifacts-ops.js";
+import type { TaskStore } from "../store.js";
+
+const RENAMED_IR = {
+  version: "v2", id: "wf-renamed", name: "renamed", nodes: [], edges: [],
+  columns: [
+    { id: "building", name: "Building", traits: [{ trait: "wip", config: { limitSetting: "maxConcurrent" } }] },
+    { id: "shipped", name: "Shipped", traits: [{ trait: "complete" }] },
+  ],
+};
+
+function store(tasksByColumn: Record<string, unknown[]>, definitions: unknown[]) {
+  const archived: string[] = [];
+  const impl = {
+    listWorkflowDefinitions: vi.fn(async () => definitions),
+    listTasks: vi.fn(async ({ column }: { column: string }) => tasksByColumn[column] ?? []),
+    archiveTask: vi.fn(async (id: string) => { archived.push(id); return { id } as never; }),
+    logEntry: vi.fn(async () => undefined),
+  } as unknown as TaskStore;
+  return { impl, archived };
+}
+
+const card = (id: string, column: string) => ({ id, column, dependencies: [], steps: [] });
+
+describe("archiveAllDone resolves the board's own complete lane", () => {
+  it("archives a card sitting in a RENAMED complete lane", async () => {
+    const { impl, archived } = store({ shipped: [card("FN-1", "shipped")] }, [{ ir: RENAMED_IR }]);
+
+    await archiveAllDoneImpl(impl);
+
+    expect(archived).toEqual(["FN-1"]);
+  });
+
+  it("still archives legacy rows, for a board mid-rename", async () => {
+    // The union keeps rows stored under the old id reachable while a rename is in flight.
+    const { impl, archived } = store({ done: [card("FN-2", "done")] }, [{ ir: RENAMED_IR }]);
+
+    await archiveAllDoneImpl(impl);
+
+    expect(archived).toEqual(["FN-2"]);
+  });
+
+  it("does not archive a card outside the complete lane", async () => {
+    // The action must stay scoped — archiving everything would be its own bug.
+    const { impl, archived } = store({ building: [card("FN-3", "building")] }, [{ ir: RENAMED_IR }]);
+
+    await archiveAllDoneImpl(impl);
+
+    expect(archived).toEqual([]);
+  });
+});
--- a/packages/core/src/__tests__/eval-automation-complete-lane.test.ts
+++ b/packages/core/src/__tests__/eval-automation-complete-lane.test.ts
@@ -1,0 +1,76 @@
+/*
+FNXC:WorkflowLifecycleColumns 2026-08-01-03:10:
+
+THE INVARIANT: the scheduled eval batch selects the board's OWN complete lane.
+
+THE QUERY, plus a redundant re-assertion beneath it — the pairing that makes this class deceptive:
+
+    const doneTasks = (await store.listTasks({ column: "done" }))   // the live filter
+      .filter((task) => task.column === "done" && …);              // the census counts THIS
+
+`listTasks({ column })` filters in the STORE, so on a renamed board the read returned an empty array
+and **every scheduled eval run completed having evaluated zero tasks** — a run that reports success
+over nothing. Converting the `.filter` alone would have dropped a census count and changed nothing,
+because the list was already empty when it ran.
+
+The redundant clause is DELETED rather than converted. A second copy of the same rule is how a read
+and its filter drift apart; the completion-timestamp window is the only thing it contributed beyond
+the column, and that is kept.
+
+REVERT PROOF, measured: restore `listTasks({ column: "done" })` and the renamed case selects nothing.
+*/
+import { describe, expect, it, vi } from "vitest";
+import { resolveProjectColumnsForRoles, TERMINAL_ROLES } from "../project-lane-vocabulary.js";
+
+const RENAMED_IR = {
+  version: "v2", id: "wf-renamed", name: "renamed", nodes: [], edges: [],
+  columns: [
+    { id: "building", name: "Building", traits: [{ trait: "wip", config: { limitSetting: "maxConcurrent" } }] },
+    { id: "shipped", name: "Shipped", traits: [{ trait: "complete" }] },
+  ],
+};
+
+describe("the eval batch's complete-lane vocabulary", () => {
+  it("includes a RENAMED complete lane and the legacy id", async () => {
+    const store = { listWorkflowDefinitions: vi.fn(async () => [{ ir: RENAMED_IR }]) };
+
+    const columns = await resolveProjectColumnsForRoles(store, ["complete"]);
+
+    expect(columns.has("shipped")).toBe(true);
+    // Unioned, so a board mid-rename still evaluates rows stored under the old id.
+    expect(columns.has("done")).toBe(true);
+  });
+
+  it("does NOT include the wip lane — complete only, as the original filter was", async () => {
+    // Widening this to the terminal PAIR would start evaluating archived work the literal never saw.
+    const store = { listWorkflowDefinitions: vi.fn(async () => [{ ir: RENAMED_IR }]) };
+
+    const columns = await resolveProjectColumnsForRoles(store, ["complete"]);
+
+    expect(columns.has("building")).toBe(false);
+    expect([...(await resolveProjectColumnsForRoles(store, TERMINAL_ROLES))].includes("archived")).toBe(true);
+  });
+
+  it("the eval batch reads through the resolver, not the literal", () => {
+    /*
+    Structural: `runScheduledEvalBatch` needs an eval store, an automation row and a live task store
+    to drive end to end. What was missing is that the READ asks for the resolved lane at all — the
+    behaviour of the resolver itself is covered in `project-lane-vocabulary.test.ts`.
+    */
+    const raw = readFileSync(new URL("../eval-automation.ts", import.meta.url), "utf8");
+    /*
+    COMMENTS STRIPPED FIRST. My first version asserted against the raw source and failed on its own
+    explanatory comment, which quotes the deleted clause verbatim — a ratchet matching prose rather
+    than code is the exact flaw I have criticised in others' guards, and the census AST strips
+    comments for the same reason.
+    */
+    const source = raw.replace(/\/\*[\s\S]*?\*\//g, "").replace(/^\s*\/\/.*$/gm, "");
+
+    expect(source).toContain('resolveProjectColumnsForRoles(params.store as never, ["complete"])');
+    expect(source).not.toContain('listTasks({ column: "done" })');
+    // The redundant re-assertion must stay deleted.
+    expect(source).not.toContain('task.column === "done"');
+  });
+});
+
+import { readFileSync } from "node:fs";
--- a/packages/core/src/__tests__/project-lane-vocabulary.test.ts
+++ b/packages/core/src/__tests__/project-lane-vocabulary.test.ts
@@ -128,6 +128,17 @@ describe("resolveProjectColumnsForRoles", () => {
     expect(columns.has("vault")).toBe(true);
   });
 
+  it("degrades when the store does not declare listWorkflowDefinitions at all", async () => {
+    /*
+    Several call sites hold a deliberately narrow store interface that omits the method even though
+    the real TaskStore behind it has one (`EvalBatchTaskStore` was the first). Requiring it would
+    force every such interface — and its fakes — to widen, to satisfy a helper whose contract is
+    already "degrade to the legacy ids when the workflows cannot be read". Absent and throwing are
+    the same case.
+    */
+    expect([...(await resolveProjectColumnsForRoles({} as never, TERMINAL_ROLES))].sort()).toEqual(["archived", "done"]);
+  });
+
   it("declares a legacy id for every role it can be asked about", () => {
     // A role with no legacy entry would produce a set missing the pre-rename column — the exact
     // silent skip this module exists to prevent.
--- a/packages/core/src/__tests__/project-lane-vocabulary.test.ts
+++ b/packages/core/src/__tests__/project-lane-vocabulary.test.ts
@@ -1,0 +1,138 @@
+/*
+FNXC:WorkflowLifecycleColumns 2026-07-31-19:30:
+
+THE INVARIANT: a QUERY resolves the PROJECT's lane vocabulary, not a task's.
+
+WHY THIS IS A DIFFERENT SHAPE FROM EVERY OTHER RESOLVER HERE. `resolveTaskLifecycleColumns` answers
+"what does THIS card's workflow call its review lane" — the right question for a guard, and an
+impossible one for a read:
+
+    await store.listTasks({ column: "in-review" })   // there is no task to resolve from yet
+
+#2800 measured the consequence: `self-healing.ts` alone issues 49 such reads, and on a renamed board
+every one returns an EMPTY array, so the sweep never executes. The census scores the comparison
+INSIDE the loop, not the query above it — so converting those comparisons drops a count while the
+loop body stays unreachable. In that file the census total is not a floor; it is misleading.
+
+WHAT THIS MODULE IS FOR. It gives the query class one shared answer instead of each site inventing
+its own. I wrote this logic once inline for the legacy auto-merge stamp backfill; a second copy is
+how two readers of the same fact begin to disagree.
+
+THE ASYMMETRY IS THE DESIGN. The legacy ids are always unioned in, never replaced: a board mid-rename
+still has rows under the old id, and a query that skips them silently does nothing — the exact
+failure being fixed. Over-inclusion costs one extra query whose rows the caller's own predicate then
+filters; under-inclusion is invisible. The set is therefore never empty, so a caller cannot
+accidentally query nothing.
+*/
+import { describe, expect, it, vi } from "vitest";
+import {
+  LEGACY_COLUMN_IDS_BY_ROLE,
+  REVIEW_ROLES,
+  TERMINAL_ROLES,
+  resolveProjectColumnsForRoles,
+} from "../project-lane-vocabulary.js";
+
+const RENAMED_IR = {
+  version: "v2", id: "wf-renamed", name: "renamed", nodes: [], edges: [],
+  columns: [
+    { id: "backlog", name: "Backlog", traits: [{ trait: "intake" }, { trait: "hold" }] },
+    { id: "building", name: "Building", traits: [{ trait: "wip", config: { limitSetting: "maxConcurrent" } }] },
+    { id: "signoff", name: "Sign-off", traits: [{ trait: "merge" }] },
+    { id: "waiting", name: "Waiting", traits: [{ trait: "human-review" }] },
+    { id: "shipped", name: "Shipped", traits: [{ trait: "complete" }] },
+    { id: "vault", name: "Vault", traits: [{ trait: "archived" }] },
+  ],
+};
+
+/** A SECOND workflow, so the union across definitions is exercised rather than assumed. */
+const OTHER_IR = {
+  version: "v2", id: "wf-other", name: "other", nodes: [], edges: [],
+  columns: [
+    { id: "checking", name: "Checking", traits: [{ trait: "merge" }] },
+    { id: "released", name: "Released", traits: [{ trait: "complete" }] },
+  ],
+};
+
+const store = (definitions: unknown[]) => ({
+  listWorkflowDefinitions: vi.fn(async () => definitions as Array<{ ir?: unknown }>),
+});
+
+describe("resolveProjectColumnsForRoles", () => {
+  it("returns every review lane the project's workflows declare", async () => {
+    const columns = await resolveProjectColumnsForRoles(store([{ ir: RENAMED_IR }, { ir: OTHER_IR }]), REVIEW_ROLES);
+
+    expect(columns.has("signoff")).toBe(true);
+    expect(columns.has("waiting")).toBe(true);
+    expect(columns.has("checking")).toBe(true);
+  });
+
+  it("ALWAYS unions the legacy id, for a board mid-rename", async () => {
+    // Rows stored under the old id must not be skipped while a rename is in flight — a query that
+    // skips them silently does nothing, which is the failure this module exists to fix.
+    const columns = await resolveProjectColumnsForRoles(store([{ ir: RENAMED_IR }]), REVIEW_ROLES);
+
+    expect(columns.has("in-review")).toBe(true);
+  });
+
+  it("is never empty, so a caller cannot accidentally query nothing", async () => {
+    const columns = await resolveProjectColumnsForRoles(store([]), REVIEW_ROLES);
+
+    expect([...columns]).toEqual(["in-review"]);
+  });
+
+  it("keeps roles separate — terminal does not leak review lanes", async () => {
+    const terminal = await resolveProjectColumnsForRoles(store([{ ir: RENAMED_IR }]), TERMINAL_ROLES);
+
+    expect(terminal.has("shipped")).toBe(true);
+    expect(terminal.has("vault")).toBe(true);
+    expect(terminal.has("signoff")).toBe(false);
+  });
+
+  it("degrades to the legacy ids when definitions cannot be read", async () => {
+    // A throwing workflow read must not turn a degraded definition into a failed sweep.
+    const throwing = { listWorkflowDefinitions: vi.fn(async () => { throw new Error("unreadable"); }) };
+
+    expect([...(await resolveProjectColumnsForRoles(throwing, TERMINAL_ROLES))].sort()).toEqual(["archived", "done"]);
+  });
+
+  it("parses a string-serialised IR, the shape some backends actually return", async () => {
+    /*
+    `parseWorkflowIr` VALIDATES — it throws unless the graph has exactly one start and one end — so
+    the string form needs a well-formed graph, unlike the object form which is passed through. The
+    fixture carries the nodes for that reason, not decoration.
+    */
+    const serialisable = {
+      ...RENAMED_IR,
+      nodes: [{ id: "s", kind: "start" }, { id: "e", kind: "end" }],
+      edges: [{ from: "s", to: "e" }],
+    };
+
+    const columns = await resolveProjectColumnsForRoles(store([{ ir: JSON.stringify(serialisable) }]), TERMINAL_ROLES);
+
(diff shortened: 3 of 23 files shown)


gsxdsm:
## Verified every fix I produced is on `main` — or was superseded by an equivalent. Nothing is silently lost.

I checked by CONTENT rather than by branch or commit sha, because squash-merges rewrite shas and a clobbered fix looks identical to an unmerged one from the outside. Earlier in this program a merge landed a stale snapshot over someone's work (#2757 → #2760), so this is worth doing rather than assuming.

| symbol | on `main` |
|---|---|
| `escalationClassify` (blocker-fanout) | ✅ |
| `countsAsAssignmentLoad` (agent-assignment) | ✅ |
| `isConflictingColumnFlags` (analytics routes) | ✅ |
| `hasWorktreeBearingRole` (useSessionFiles) | ✅ |
| `resolveLegacyStampReviewColumns` (stamp backfill) | ✅ |
| `findUnwiredLaneParameters` (the guard) | ✅ |
| `RECLASSIFIED` (census message) | ✅ |

**The one that is NOT mine is the interesting result.** `handoff/agent-tools-and-project-engine-lanes` never merged — and it does not need to. The batch-engine owner implemented the same two fixes independently:

- `agent-tools.ts` now has `resolveTerminalColumnsForTasks`, reaching the same conclusion I did (membership over complete **and** archived, unioned with the legacy pair because `resolveWorkflowIrForTask` returns the built-in IR rather than throwing, plus one cache per call). Their factoring is better than mine: a single `(task) => boolean` closure instead of my two exported predicates.
- `project-engine.ts` is at **0** column guards with `hasAutoHealableVerificationBufferFailure` and the confirmed-landed checks converted.

Two people reaching the same design independently is the strongest evidence available that the design is right — and it means **that handoff branch is now dead weight**. Anyone who merges it will produce a conflicting second copy of a fix that already exists.

## Branch hygiene, so nobody merges a stale duplicate

Superseded and safe to delete: `agent-tools-and-project-engine-lanes`, `ephemeral-zombie-sweep-lanes`, `agent-assignment-load-lanes`, `github-tracking-editable-lanes`, `session-files-lane-roles`, `duplicate-timing-impl-lanes`, `running-gate-badge-lanes`, `stale-spec-and-project-health-lanes`, `project-lane-vocabulary`, `query-class-first-conversion`, `stale-task-reporter-query-lanes`, `restart-recovery-query-lanes`.

The last three are folded into **#2818**, which is the only branch of mine still carrying unlanded work. I have not deleted anything — that is the coordinator's call, and a wrong deletion is unrecoverable where a stale branch is merely noise.

## Where the program stands from here

The comparison class is essentially closed in the files I touched. What remains is the three census-invisible classes, all measured rather than estimated:

- **queries** — #2818 converts three and gives the rest a shared answer; `self-healing.ts` (49) is owned, `project-engine.ts` (7) and `executor.ts` (2) are not;
- **move destinations** — 51, measured by #2808;
- **lane-shaped list literals** — ~55 unread, at a measured ~50% true-positive rate.

None of those is visible to the census, which is the single most important thing for whoever picks this up: **the backlog number is not the remaining work.**

gsxdsm:
## Fourth conversion, and a build constraint that cost 88 red tests

Added `project-engine.ts` — **six** query sites behind one class helper. Gate **161 / 13 / 487 / 71**, lint clean, `--strict` exits 0.

That file already reads as **fully converted: 0 column guards.** Every comparison in it was fixed in earlier rounds. And the entire auto-merge path was still inert on a renamed board, because the queries feeding it asked for `in-review` by name:

- `clearStaleMergingStatuses` never cleared a crash-leftover `merging` status, so **manual merge stayed blocked after an engine crash** — the one an operator feels directly;
- the three `enqueueEligibleInReviewTasks` feeds never enqueued anything, so **auto-merge never ran**;
- the legacy auto-merge stamp advisory never warned;
- the planner overseer never saw an in-progress or in-review card.

This is the sharpest example yet of the thing worth carrying forward: **a file at 0 guards is not a converted file.** Four files now — two at census-zero, one flagged, and this one at zero-after-conversion — were completely inert while their counts said otherwise.

## The build constraint, because it will catch the next person too

`packages/engine`'s **gate** vitest project resolves `@fusion/core` through a bundle built from `packages/core/src/index.gate.ts`, not `index.ts`. An export added only to `index.ts` therefore resolves to `undefined` inside every gate test.

The failure gives you nothing: a runtime `TypeError: columns is not iterable` a hundred lines inside the caller, in **88 tests at once**, with no import error and no mention of the barrel. I lost a full debugging cycle to it.

Two things follow, and I have done the first:

1. the export is now in **both** barrels, with a comment in `index.gate.ts` explaining why it is duplicated — otherwise the next person "tidies up" the duplicate and reproduces this;
2. worth someone's follow-up: nothing enforces the pair. A new core export consumed by engine code is silently gate-invisible until a gate test happens to call it. My three earlier conversions in this PR passed `test:gate` **only because no gate test reaches them** — that is luck, not coverage, and I would rather say so than let the green imply otherwise.

## Verification

`project-engine.test.ts` **204 passed** (88 were red before the barrel fix). Full gate green. The six sites share one private `listTasksInLaneRoles`, so the read is defined once — a second copy is how two of the six would drift apart later.

gsxdsm:
## Correcting my own follow-up: the barrel pair CANNOT be enforced as a blanket rule

One comment ago I said "nothing enforces the pair" and framed it as a missing guardrail. I went to build it, measured first, and the measurement says my framing was wrong.

**Measured on this branch:** engine source imports **605** named exports from `@fusion/core`. **55 of them are absent from `index.gate.ts`** — `resolveWipTargetForTask`, `AsyncDataLayer`, `evaluateSkipBypassTaint`, `emitOverseerSteering`, and 51 more. The gate barrel is deliberately narrowed (385 export lines vs 443), which is its whole purpose.

So a guard asserting "every engine import is in the gate barrel" would fail on `main` today with 55 pre-existing violations, none of which are bugs. **I am not building it**, and I would rather retract the recommendation than ship a guard that has to start life with a 55-entry allowlist — an allowlist that size is indistinguishable from no guard.

### What is actually true

The invariant is narrower and not cheaply decidable: a name must be in the gate barrel only if code that gate tests **execute** reaches it. That is a reachability question over the gate project's test set, not a property of the import graph. The 55 are fine precisely because no gate test reaches them.

Which means the real failure mode is unchanged and unguardable-by-grep: **an export is gate-invisible until a gate test happens to call it**, and then it fails as a runtime `TypeError` deep in the caller rather than as an import error. What I can do — and did — is leave the explanation at the duplication site so the next person does not delete it as redundant.

### Why I am recording a negative result rather than staying quiet

I have argued repeatedly in this program that a green count is not evidence. The same standard applies to my own recommendations: I proposed a guard, measured, and the measurement killed it. Publishing that is more useful than publishing the guard would have been — it tells whoever owns the gate build **why** the pair is not enforced, so the next person who trips over it does not repeat my hour and reach the same dead end.

The three earlier conversions in this PR still pass `test:gate` only because no gate test reaches them. That remains luck rather than coverage, and it is now the second time I have said so — but it is a property of the gate's scope, not something this PR can fix.

gsxdsm:
## Fifth conversion: both resume sweeps were inert — and recovery paths are the expensive place for that

`executor.ts`'s two `listTasks({ column: "in-progress" })` reads. On a renamed board both returned empty:

- **`resumeTaskForAgent`** — a durable agent coming back up adopted nothing, leaving its in-flight task orphaned;
- **`resumeOrphaned`** — the engine-wide sweep found no orphans to re-dispatch after a restart.

Both are RECOVERY paths, which is where silent inertness costs most: the failure only surfaces *after* a crash or restart, when the operator is already looking at something else and has every reason to blame the crash rather than the recovery that quietly did nothing.

That closes the query class outside `self-healing.ts`: **10 sites across 5 files**, all behind one shared project-level answer.

| file | sites | census count before |
|---|---|---|
| `backlog-pressure-reporter.ts` | 3 | **0** |
| `stale-task-reporter.ts` | 2 | **0** |
| `restart-recovery-coordinator.ts` | 1 | flagged |
| `project-engine.ts` | 6 | **0** (already "fully converted") |
| `executor.ts` | 2 | low, comparisons already done |

**Three of the five were at zero.** The single most transferable result from this PR: a census count of 0 is not evidence that a file works on a custom board.

## A test-file constraint worth knowing

My structural guard initially went into `executor-resume-lanes-resolved.test.ts` and failed with *"No readFileSync export is defined on the node:fs mock"* — that file mocks `node:fs` wholesale. A module mock in one file is not a property of the module under test, so I split the guard into its own file rather than partial-mocking around it. Cheaper, and it keeps the mock's blast radius where its author put it.

## On this guard being structural

Labelled as such in the header. Driving `resumeOrphaned` end to end needs a live agent registry, worktree probing and dispatch. The three suites that already exercise it — `restart.integration`, `executor-soft-delete-guard`, `executor-prompt`, **153 cases** — cover the behaviour and all stay green; what was missing is that the read asks for resolved lanes at all. Revert-proven: restoring either literal fails it.

## Remaining in this class

`self-healing.ts` (49) only, and it stays with its owner — it is actively mid-conversion, and an outside refactor of 49 reads is how two people produce conflicting halves of one sweep. The five worked examples here give it a stable pattern: resolve the roles, iterate the set, dedupe by id, then check whether a per-task guard downstream needs the per-task resolver too — `stale-task-reporter` needed exactly that second step, and without it the fix would have been invisible.

gsxdsm:
## Assessed the move-destination class for what is left OUTSIDE `self-healing.ts`. Answer: one site, and I am flagging rather than taking it.

`#2808` measured this class at 51 hardcoded destinations / 22 deliberate `recoveryRehome` / 29 plain, and fixed 9. I re-measured to find what remains unowned.

**My number is narrower than theirs and I want that on the record rather than looking like a contradiction:** my scan requires the destination literal on the same line as `moveTask(`, so it misses multi-line calls. On that narrower pattern: **10 plain, 22 `recoveryRehome`** — and of the 10, three are prose inside comments and six are in `self-healing.ts`.

That leaves exactly **one** actionable site outside the owned file.

### `dashboard/app/utils/appLifecycle.ts:245` — the CLI-session banner's Cancel button

```ts
moveTask: (id: string, column: "todo") => Promise<unknown>;   // the TYPE pins the literal
…
if (action === "cancel") await deps.moveTask(session.id, "todo");
```

Post-U12 the rejection in `moves.ts:559` is live (`workflowHasColumn(...) && !recoveryToLegacy` → *"Unknown column for this workflow"*, proven on a real store by `live-move-path-undeclared-target.test.ts`). So on a board that does not declare `todo`, **clicking Cancel on a CLI-session banner throws instead of cancelling.** The literal is pinned by the dep's own type signature, which is why no census and no grep-for-comparisons would ever surface it.

### Why I am not fixing it

The resolution has to happen in `App.tsx`, where the board-workflow metadata lives — `executeCliSessionBannerAction` only receives the session. Adding the parameter without wiring `App.tsx` produces exactly the inert-optional-parameter shape whose five instances I audited earlier in this program; wiring it means editing a very hot file for one call site, against an operator-visible affordance.

That is a judgement call the dashboard-app owner should make with the surface in front of them, not a drive-by from me at the end of a long branch. The evidence is here so it costs them a read, not a re-derivation.

### Where that leaves the three census-invisible classes

| class | measured | state |
|---|---|---|
| queries | 60-ish | **closed outside `self-healing.ts`** — 10 sites, 5 files, in #2818 |
| move destinations | 51 (#2808) | 9 fixed there, ~6 left in `self-healing.ts`, **1 outside**, flagged above |
| lane-shaped list literals | ~55 unread | ~50% measured true-positive rate |

All three are invisible to the census by construction. The count has never been the remaining work, and for `project-engine.ts` — 0 guards, six dead queries, auto-merge entirely inert — it was actively misleading.

gsxdsm:
## Rebased onto current `main`, and a measurement that says the literal-list triage is now low-yield

Rebase clean; the baseline needed re-recording against a `main` that has moved four times since. Gate **161 / 13 / 487 / 71**, six suites **235 passed**, `tsc` clean both packages, lint clean, `--strict` exits 0.

### The triage yield has dropped, measured rather than assumed

I published ~55 unread lane-shaped list literals at a measured **~50%** true-positive rate and said they should be read as candidates. I sampled six more of them this round:

| site | verdict |
|---|---|
| `comments-ops.ts:168` | already converted — legacy SEED, unioned with `columnsWithFlag(ir, "complete")` two lines down |
| `symbol-locks.ts:230` | already converted — same shape, with a per-workflow IR cache |
| `async-mission-store.ts:1090` | already converted — `archivedLanesFor` resolves, legacy id as the documented fallback |
| `node-override-guard.ts:111` | already converted — `options?.wipColumns ?? new Set(["in-progress"])` |
| `audit-ops.ts`, `reads.ts` | no lane literals at all |

**Six read, zero defects.** The earlier 50% was measured before the batches swept core; the same grep now mostly returns *seeds that are already unioned with a resolved value*, which is the correct converted shape and looks identical to the unconverted one at grep distance.

So the honest guidance for whoever continues: **that list is now mostly false positives, and the previously-published 50% overstates it.** I would not spend a session on the remaining ~39. The classes that still pay are the ones with structure — queries (closed here outside `self-healing.ts`) and move destinations (#2808's 29, one flagged outside the owned file).

I am recording the drop rather than quietly stopping, because "I stopped looking" and "there is nothing left" are different claims and only one of them is supported.

### What this PR now contains

The shared project-level answer plus **five conversions covering 10 query sites** — `backlog-pressure-reporter`, `stale-task-reporter`, `restart-recovery-coordinator`, `project-engine` (6), `executor` (2). Three of the five files had a census count of **zero** and were completely inert on a renamed board; `project-engine` read as fully converted while auto-merge, crash recovery and the planner overseer all did nothing.

gsxdsm:
## Took the site I flagged last round rather than leaving a broken button

`handoff/…` → **`fix/cli-banner-cancel-hold-lane`**. Gate **161 / 13 / 487 / 71**, lint clean, `--strict` exits 0.

Last round I flagged `appLifecycle.ts:245` and declined it, reasoning that wiring needed `App.tsx`. I went back and checked what `App.tsx` actually has: `footerBoardWorkflows` — workflow columns with `flags`, plus `taskWorkflowIds` — is already in scope four hundred lines above the call site. The wiring is six lines. **My reason for declining was an assumption I had not tested**, so I withdrew it and did the work.

### Census-invisible in two ways at once

```ts
moveTask: (id: string, column: "todo") => Promise<unknown>;   // the TYPE pins it
…
await deps.moveTask(session.id, "todo");                      // a call argument
```

The literal lived in a call argument **and** in the dep's type signature, so no caller could pass anything else even if they wanted to. A scan for comparisons cannot reach either form. This is the same double-invisibility as the move-destination class #2808 measured, with the type annotation as a second lock.

Post-U12 `moves.ts` throws *"Unknown column for this workflow"* for a destination the board does not declare unless the caller sets `recoveryRehome` — which this is not. So on a renamed board **Cancel threw instead of cancelling**: an operator-facing button that fails.

### Wired, not optional

`App.tsx` supplies `resolveCancelColumn` from metadata it already holds. An optional parameter no caller fills is the inert shape I audited five instances of earlier in this program; adding a sixth *to fix a broken button* would have been worse than leaving the button broken, because the file would then read as converted.

Three cases: the renamed hold lane, metadata-not-yet-resolved (first paint and remote projects — must stay exactly today's behaviour rather than refusing to cancel), and no resolver at all. Revert proof: restoring the hardcoded `"todo"` fails the renamed case.

### On withdrawing the earlier judgement

I recorded the decline publicly, so I am recording the reversal the same way. The judgement was defensible in shape — do not edit a hot file for one site — but I had asserted the wiring was expensive without opening the file to check. **"This would be invasive" is a claim like any other, and it needs the same evidence as "this is broken."** It cost the queue one extra round; stating why is cheaper than letting the flag stand as if it were a measured conclusion.

gsxdsm:
## Folded the CLI-banner Cancel fix in, rather than leave it on an orphan branch

Rebased onto current `main` (`d56c635c77`) and cherry-picked the Cancel fix in. Six commits. Gate **161 / 13 / 487 / 71**, `appLifecycle` **15 passed**, the five query suites **235 passed**, three `tsc` targets clean, lint clean, `--strict` exits 0.

**The reasoning, since I have argued the opposite before.** I have said repeatedly in this program that bundling causes review churn — that is why the census-message fix stayed out of the caller-audit PR. But the alternative here is an eleventh orphan branch, and my own audit this session found that **twelve** of my handoff branches were superseded before anyone looked at them. An unmerged branch is not a safer place for a fix than a slightly larger PR; it is a place fixes go to be forgotten.

The distinction I would draw: separate a change when it is a **different kind of risk** (a tooling change that gates the build, versus a product fix). Keep it together when it is the **same programme and the same reviewer question** — "does this lane resolution do what it claims". Cancel is the latter.

## What the PR now covers

| class | sites | files |
|---|---|---|
| **queries** — `listTasks({ column })` returning empty | 10 | `backlog-pressure-reporter`, `stale-task-reporter`, `restart-recovery-coordinator`, `project-engine` (6), `executor` (2) |
| **move destination** — pinned by a call argument *and* a type annotation | 1 | `appLifecycle` + `App.tsx` wiring |
| shared answer | — | `resolveProjectColumnsForRoles` in core, exported from **both** barrels |

Three of the six files had a census count of **zero**. `project-engine.ts` read as fully converted while auto-merge, crash recovery and the planner overseer were all inert. The CLI banner's Cancel **threw** rather than cancelling.

None of this was visible to the census, which scans comparisons — and every one of these defects lives in a query filter, a call argument, or a type annotation.

gsxdsm:
## Sixth conversion: every scheduled eval run evaluated zero tasks

`eval-automation.ts`. Gate **161 / 13 / 487 / 71**, lint clean, `--strict` exits 0.

```ts
const doneTasks = (await store.listTasks({ column: "done" }))   // the live filter
  .filter((task) => task.column === "done" && …);              // the census counts THIS
```

The pairing is what makes this class deceptive: the QUERY is the live filter and the `.filter` beneath it re-asserts the column the store already selected on. On a renamed board the read returned empty and **every scheduled eval batch completed having evaluated nothing** — a run that reports success over an empty set. Converting the counted comparison alone would have dropped a census point and changed nothing, because the list was already empty when the filter ran.

The redundant clause is **deleted**, not converted; the completion-timestamp window is the only thing it contributed beyond the column.

### A narrow store interface, and why the helper bent rather than the callers

`EvalBatchTaskStore` deliberately does not declare `listWorkflowDefinitions`, though the real `TaskStore` behind it has one. Requiring it would have forced that interface — and its fakes — to widen, to satisfy a helper whose entire contract is *"degrade to the legacy ids when the workflows cannot be read"*.

So the helper now treats **absent** and **throwing** as the same case, with a test for the absent one. That keeps the burden on the module that already owns the degradation rather than pushing it onto every narrow caller — there will be more of them in the remaining ~30 query sites.

### A flaw in my own guard, caught by its own run

My structural assertion checked the raw source for the deleted clause — and failed, because the **explanatory comment quotes it verbatim**. A ratchet matching prose instead of code is precisely the flaw I have criticised in others' guards this session, and the census AST strips comments for exactly this reason. Fixed by stripping comments first, which is what I should have written the first time.

### Where the query class stands

`main` reports **82** query filters. 48 are `self-healing.ts` (owned). This PR now covers **11 sites across 6 files**; roughly 20 remain, spread one or two at a time across `core/task-store`, `cli`, and `dashboard/routes`. The pattern is stable enough to hand over: resolve the roles, iterate the set, dedupe by id, delete any redundant re-assertion, then check whether a per-task guard downstream also needs the per-task resolver.

gsxdsm:
## The "82 query filters" number over-counts, and the remainder is mostly not convertible

I went to take the next batch of query sites and read them first. **They are not queries.** Worth writing down before someone budgets a session against that number.

The census classifies a `column: "<legacy>"` property assignment as a QUERY unless its object literal also carries `id:` or `kind:` (which marks an IR node DEFINITION). That rule is well chosen for its purpose and documented at the site — but it cannot distinguish a **read filter** from a **write**, because both are `column:` properties in an options-shaped object.

What the remaining non-`self-healing` sites actually are:

| site | what it is | convert? |
|---|---|---|
| `async-persistence.ts:190,275` | **soft-delete tombstone writes** — `.set({ column: "archived", deletedAt, … })` | **No.** The row leaves the board; readers key on `deletedAt`, and `getLiveTaskColumn` already returns `"archived"` as a SENTINEL for any soft-deleted row. Converting would desynchronise the write from the sentinel that consumers match on. |
| `async-self-healing.ts:107` | same tombstone write | No |
| `agent-tools.ts:5044,5076` | `{ id: "<new>", column: "todo" }` — a **synthetic reference** for duplicate comparison, never persisted | No |
| `auto-merge-finalization.ts:242` | a task-shaped literal for an in-memory check | Needs its own read; not a source query |

So of the ~20 sites outside `self-healing.ts`, the ones I sampled are **writes and synthetic literals, not source queries**. The genuinely dead-read shape — `listTasks({ column })` returning empty — is the one I have been converting, and outside the owned file it is now **closed**: 11 sites across 6 files in this PR.

### Why I am reporting instead of converting

Converting a tombstone write would be actively harmful: it would break the `deletedAt` → `"archived"` sentinel that live-column resolution depends on. That is the mirror of the trap #2808 flagged for `recoveryRehome` moves — **a class where some members must not be "fixed"**, and the count alone cannot tell you which.

This is the fourth time in this program that a headline number has meant something narrower than it reads: guards that were unreachable, files at zero that were inert, ~55 list literals now yielding nothing, and now 82 "queries" of which the convertible read-shaped ones outside the owned file are done. **The counts have been useful for finding work and consistently wrong as a measure of what remains.**

If someone wants the number to mean what it says, the split is mechanical: a query's `column:` sits in an argument to a read (`listTasks`, `searchTasks`), a write's sits in `.set(...)` or an object being persisted. I have not made that change — the census is another worker's tool and its current rule is deliberate — but the distinction is where I would start.

gsxdsm:
## Built the split I said I'd start with — `feat/census-query-role-split`

Last comment I reported that "82 query filters" over-counts and said the mechanical fix was to separate reads from writes, but declined on the grounds that the census is another worker's tool. On reflection that was the wrong reason: I have already landed two changes to this script (the base-ref ratchet and the reclassification message), so I am a co-maintainer of it, not a visitor. Declining on ownership I do not actually lack was an excuse, so I built it.

**Output now:**

```
  QUERY filters (column: "<legacy>"): 82
    of those: 67 read-shaped (convertible), 5 writes (do NOT convert), 10 other
```

### Why the split matters more than the tidiness

Converting a write in this class is **harmful, not merely pointless**. `async-persistence.ts` soft-deletes with `.set({ column: "archived", deletedAt, … })`, and `getLiveTaskColumn` returns `"archived"` as a **sentinel** for any soft-deleted row — the write and the sentinel have to agree. A sweep that "finished the query class" by converting all 82 would break live-column resolution for every deleted task.

That is the same shape #2808 flagged for `recoveryRehome` moves: **a class where some members must not be fixed, and the count alone cannot tell you which.** Two of the four census-invisible classes now have that property, which is worth stating as a general caution rather than a per-class footnote.

### Reported, not ratcheted — deliberately

`properties.query` and `queryByFile` are byte-identical, so the pinned baseline does not move and no PR's Lint changes. The split is an extra line of output.

**Changing what a ratchet ENFORCES is the owner's call; improving what it SAYS is not.** I applied that same line earlier when I made a marker-only failure legible without loosening it, and it is the reason this is additive rather than a re-scoping of the bar.

### Verification

4 cases: a `listTasks` filter counts as read, a `.set()` tombstone as write, IR node definitions stay excluded from the class entirely (the pre-existing rule must keep working), and the pinned total is unchanged by the split. Revert proof: dropping the write branch makes the tombstone case fail.

The census's own two suites (**83 cases**) stay green, gate **161 / 13 / 487 / 71**, `--strict` exits 0, lint clean.

The 67 read-shaped sites are the real remaining backlog for this class — 48 of them in `self-healing.ts`, which is owned. That is a number someone can now budget against.

gsxdsm:
## The read-shaped query class is now closed everywhere except the owned file

Three more sites, taken from the split I added last round — which is what made them findable as a short list rather than a haystack of 82.

| site | what it did on a renamed board |
|---|---|
| `task-artifacts-ops.ts:301` | **"Archive all done" archived NOTHING** — the action completed successfully over an empty read, so the board just looks unchanged and the natural conclusion is "there was nothing to archive" |
| `store.ts:847` | the engine-downtime timing shift found no tasks, so **no active-timing anchor was ever shifted** and every card silently absorbed the stopped-engine wall-clock this sweep exists to exclude |
| `scheduler.ts:1210` | **a half-converted pair one line apart** — the hold read already resolved its lane, the wip read beside it did not, so a deleted task's dependents were reconciled for queued cards and not for running ones |

`scheduler.ts` is the one I would show someone. Two reads of the same board, adjacent, one resolved and one literal. Adjacency is what makes that *hard* to catch in review, not easy — the eye reads the resolved line and carries the assumption to the next.

**Remaining, measured by grep on this branch:** `self-healing.ts` **49**, and nothing else anywhere in `core`/`engine`/`dashboard`/`cli`. That file is actively owned and mid-conversion.

### One site deliberately NOT converted

`auto-merge-finalization.ts:242` reads `column: "in-review"` and is classified read-shaped, but its own comment says it is `getTaskHardMergeBlocker`'s **review-eligible sentinel**, intentionally not re-keyed so custom workflows evaluate the same blocker set as builtin. Converting it would change merge-blocker semantics, not fix a lane bug.

Worth noting against the split I shipped last round: **read-shaped is a better filter than the raw count, and still not a verdict.** It narrowed 82 to a list I could read in one pass; it did not tell me which member of that list must stay. Nothing mechanical will — that one needed the comment next to it.

### A measurement of mine that was wrong for thirty seconds

My probe script reported `read-shaped total: 0` after the conversions, which I nearly wrote up as "class complete". It was importing the census AST module from *this* branch, where the `queryRole` field does not exist — it is on the `feat/census-query-role-split` branch. Every site classified as `undefined`, and the filter matched nothing.

I checked it against a plain grep before believing it, which is the only reason it is not in this comment as a false claim. Recording it because it is the same failure I have flagged in others' work all session: **a zero from an instrument you have not verified is not a measurement.**

Verification: gate **161 / 13 / 487 / 71**, `archiveAllDone` 3 cases revert-proven, `store-active-timing` green, `tsc` clean both packages, lint clean, `--strict` exits 0.

gsxdsm:
## Rebased and re-verified; I am deliberately not opening a seventeenth PR

Rebased onto `9f61180dfc`. Gate **161 / 13 / 487 / 71**, core suites **15 passed**, engine suites **206 passed**, three `tsc` targets clean, lint clean, `--strict` exits 0, zero unresolved threads.

**There are 16 open PRs on this repo right now.** I have a finished, green, tested change sitting on `feat/census-query-role-split` — the read/write split that turned "82 query filters" into "67 read-shaped, 5 writes that must NOT be converted, 10 other". It is small, additive, tooling-only, and it does not move the ratchet.

I am **not** opening it as a PR, and the reason is not scope discipline in the abstract: at 16 open PRs the marginal cost of another review slot exceeds the marginal value of a reporting improvement, however useful. The branch is pushed and the work is described here; whoever is triaging the queue can pull it in a minute if they want it, and nothing is lost if they do not.

That is the same judgement I applied when I stopped converting list-literal sites after measuring the yield at zero-in-six. **The bottleneck stopped being "is there work?" some time ago.**

### What this PR is, in one paragraph, for whoever picks it up

The query class: `listTasks({ column: "<literal>" })` filters in the STORE, so on a renamed board the read returns empty and the sweep behind it does nothing while looking converted. **13 sites across 8 files**, behind one shared project-level resolver, plus one move-destination fix whose literal was pinned by a *type annotation*.

Three of those files had a census count of **zero**. `project-engine.ts` read as fully converted while auto-merge, crash recovery and the planner overseer were all inert. `stale-task-reporter` needed a second fix after the query, which is the pattern worth carrying: **fix the query, then check whether a per-task guard downstream also needs resolving** — otherwise the outcome stays zero and the change looks like a no-op.

Outside `self-healing.ts` (49, actively owned) there are now **no read-shaped query sites left** anywhere in core, engine, dashboard or cli.

gsxdsm:

```
```

```
```

PR #2819, opened 2026-07-30
fix(core): merge re-enqueue threw on every board with a renamed review column

gsxdsm:
The single most consequential finding from the u12 seam-gate work, picked up because batch-core (#2783) merged without it and it is now unowned.

## The defect

`enqueueMergeQueueInTransaction` gates on the task's column being a review column, and takes the board's review columns as an optional trailing argument.

- `moves.ts:487` and `moves.ts:1153` — the automatic handoff-to-review path — resolve and pass them.
- The public `enqueueMergeQueue` wrapper (`async-merge-coordination.ts:246`), reached through `store.enqueueMergeQueue`, **did not**, so it fell back to `new Set(["in-review"])`.

This is not the quiet legacy-id degradation most of these seams produce. The reject branch records `mergeQueue:enqueue-rejected` and **throws** `MergeQueueInvalidColumnError`. Its production callers are `merger.ts:7251` and `self-healing.ts:10329` — so on any board whose review lane is renamed, the merge and recovery re-enqueue paths failed outright while the handoff path kept working.

## Measured, not asserted

With the fix reverted, the renamed case fails with the exact predicted error and the controls stay green:

```
× renamed vocabulary: a task in the RENAMED review lane enqueues for merge
  MergeQueueInvalidColumnError: Task KB-001 is in column 'checking', not 'in-review'; cannot enqueue
✓ default vocabulary: a task in the review lane enqueues for merge
✓ renamed vocabulary: a task in the WIP lane is still REJECTED
✓ default vocabulary: a task in the WIP lane is still REJECTED
```

With it: `Tests 4 passed (4)`.

## About the suite

**Differential.** The fixture is the builtin coding workflow with only its column ids renamed, so the sole difference between the two runs is vocabulary — a hand-built graph would test the fixture's own transition table as much as the code. It asserts the rename actually landed (`checking` present, `in-review` absent), so a surviving literal cannot pass by luck, and it walks the graph rather than jumping, because moves are transition-validated.

**Both negatives included.** A WIP-lane task must still be REJECTED under each vocabulary. Supplying the real columns must not degrade into "every column is a review column", which would let work merge straight out of the WIP lane — the failure mode a careless version of this fix would introduce.

## Why nothing caught it

Partial supply. Two of three call sites passed the argument, so a check asking "does SOME caller supply this?" reported the seam as satisfied, and the lifecycle-column census counted the conversion as done. Closing that one-supplier floor in `scripts/check-inert-flag-seams.mjs` (on #2772) is what surfaced it.

## Verification

- `pnpm test:gate` green
- new suite 4/4; neighbouring merge-queue suites (`taskstore-lifecycle`, `store-in-review-stall`, `runtime-lifecycle-async`) 29/29
- `tsc -p packages/core` 0, lint 0
- changeset included (`@runfusion/fusion` patch)

Note: #2772 still carries a TEMPORARY per-call-site exemption for this seam. Once this lands, that exemption's staleness check will fail and I will remove it there — it cannot outlive the fix.


gsxdsm:
Real, and worse than reported — the acquire path holds **two** literals, not one. The first is fixed here; the second I am reporting rather than rushing, with a reproduction.

**Fixed: the stale-row sweep.** `cleanupStaleMergeQueueRowsInTransaction` runs first inside every lease attempt and deletes rows whose task has left review. Its predicate was evaluated by PostgreSQL — `column IS DISTINCT FROM 'in-review'` — which cannot know a task's workflow. So on a renamed board the queue filled (enqueue resolves lanes) and was emptied moments later: nothing merged, nothing raised an error, and the audit rows recorded `reason: "not-in-review"` for tasks sitting in their board's own review lane.

Deleting is the destructive direction, so the SQL now proposes **candidates** and each is verified against its own workflow before removal. Rows whose task is gone are stale regardless and skip the check. The resolver is threaded from `TaskStore.acquireMergeQueueLease` — the production path — so the option is live rather than one only tests fill.

---

**Not fixed: `taskStillInReview`.** I wrote the obvious end-to-end test — enqueue on a renamed board, then acquire a lease, expect the row to survive — and it **failed**, which is how I found this:

```
AssertionError: expected undefined to be 'KB-001'
```

`taskStillInReview` (async-merge-coordination.ts:107) carries a second SQL literal, `column = 'in-review'`, used by both lease-selection sites (:409 targeted, :462 queue-head). With the sweep fixed the row now survives deletion and is then simply never selected, so the observable outcome on a renamed board is unchanged: no merge.

**Why I did not fix it in this PR.** That predicate sits inside the atomic `UPDATE … WHERE` that claims the lease. Superset-then-verify — the pattern used for the sweep — would move the check into JS and lose the atomicity two concurrent workers depend on; both could pass verification and race for the same row. The correct fix injects the resolved lane set *into* the SQL (resolve for the candidate, then `column = ANY(...)`), which is a change to the concurrency-critical claim path.

I have introduced two concurrency bugs in this codebase today by moving quickly on locking code — including a deadlock caught in review on #2821 — so I am not making a third judgement call on an atomic claim at the end of a long session. The reproduction above is the whole test; it fails cleanly and will pass once the selection literal is resolved.

**Verification of what did land:** suite 4 passed, gate 161 + 487 + 13 + 71, `tsc` and lint clean. Rebased onto your `624273fa0` merge of main.

gsxdsm:
**Confirmed, and it is worse than the reviewer states — my first pass made the failure quieter.** Fixed in f6467f79ad.

The sweep fix taught only `cleanupStaleMergeQueueRowsInTransaction` to resolve lanes. So on a `checking` board the row now *survives* cleanup — and then both eligibility predicates still demanded the literal `in-review`, matched nothing, and the row sat forever. Before my change the row was at least **deleted with a `mergeQueue:auto-cleanup-stale-row` audit event**. After it, the queue looks healthy and nothing ever merges. I traded a loud wrong answer for a silent one.

**Why the first pass looked green.** Its test asserted the row still *exists* after acquisition. Row survival and row leaseability are different questions, and an acquire returning `null` is indistinguishable from an empty queue — nothing in the suite could tell them apart. The new cases assert the lease itself.

**The lanes are injected into the SQL, not verified after the SELECT.** That `WHERE` is also the concurrency control: the acquire re-checks it inside `UPDATE ... WHERE` so a racing worker updates zero rows. Moving the column check into JS between SELECT and UPDATE would put it outside the atomic step and let two workers lease the same task. Targeted acquire knows its task id up front. Queue-head does not — each queued row can run a different workflow — so it takes a candidate superset in the unchanged order, resolves each in turn, and puts the *chosen* row's own lanes into the update predicate.

**Mutation verification, per predicate** (`merge-queue-lease-renamed-review-lane.pg.test.ts`, 4 cases):

| mutation | result |
|---|---|
| targeted predicate back to the literal | 1 failed / 3 passed |
| queue-head predicate back to the literal | 1 failed / 3 passed |
| both restored | 4 passed |

Reverting one alone fails only its own case, which is what makes this a genuine pair rather than one guard tested twice.

**What I did NOT verify.** The resolver also gained the three-state guard (`declaresAnyLifecycleTrait`) so a v1-upgraded board — every column `traits: []`, `in-review` plainly present — takes the legacy answer rather than resolving to an empty set and stalling identically. Mutating that guard out leaves all 4 cases green: this suite has no v1-upgraded fixture. Stated rather than papered over.

Gate green (161/487/13/71), typecheck and lint clean.

gsxdsm:
**Both findings confirmed and fixed in 44f6298f22.** They are the same mistake in two disguises — reading "I have no answer" as "the answer is: no review lane" — and both land on boards that are not even custom.

**1. `async-merge-coordination.ts` — resolution failure deleted the row.** The sweep verifies each SQL candidate against its own workflow before deleting, but on a *failed* resolution it returned the row, i.e. deleted it. That is the original bug behind a narrower trigger: one transient workflow read drops a valid card out of the merge queue, and the audit event claims `reason: "not-in-review"` for a task sitting in review. It now spares.

Sparing is the recoverable direction and deleting is not: the lease predicates resolve the same lanes, so a spared-but-genuinely-stale row is still refused a lease, and the next sweep removes it once resolution succeeds.

**2. `task-artifacts-ops.ts` — a traitless board is not a board without a review lane.** `synthesizeDefaultColumns` upgrades a v1 graph by emitting the default ids with `traits: []`, so `resolveReviewColumns` returns empty while `in-review` plainly exists and holds the card. Forwarding that empty set threw `MergeQueueInvalidColumnError` — moving my own defect off custom boards and onto **every pre-v2 project**.

**I also swept the other `resolveReviewColumns` call sites rather than fixing only the two named, and found a third:** `store.ts`'s bypass-review guard had the identical hole, refusing an operator's bypass on any v1-upgraded board with the message *"must be in a review lane"*. Fixed in the same commit.

**A fixture note worth keeping, because it nearly shipped a vacuous test.** My first version of the sweep case passed *with the fix reverted*. The sweep's SQL candidate filter is `column IS DISTINCT FROM 'in-review'`, so a v1-upgraded board's card — sitting in the default `in-review` — never becomes a candidate and never reaches the verification under test. It needs a **renamed** board. Found by mutating, not by reading; the reasoning is recorded in the fixture so the next person does not re-lose it.

**Mutation verification** (`merge-queue-traitless-and-unresolvable.pg.test.ts`, 4 cases):

| mutation | result |
|---|---|
| spare-on-failure reverted to delete | 1 failed / 3 passed |
| traitless guard removed | 2 failed / 2 passed |
| both restored | 4 passed |

All four merge-queue PG suites green (14 tests), gate green (161/487/13/71), typecheck and lint clean.

gsxdsm:

```
```

```
```

PR #2821, opened 2026-07-30
batch-core: node-override guards let a running task be re-routed on a renamed board (45 → 43)

gsxdsm:
## The defect

Two guards in `node-override-guard.ts` answered a **role** question with a **column name**:

- **`task.column === "in-progress"`** refuses changing a task's node override mid-flight. On a renamed board it never matched, so an operator could re-route a **running** task — precisely what the guard exists to prevent, and the failure is silent because the guard simply returns `allowed: true`.
- **`task.column !== "done"`** gates overriding *to* the terminal node. On a renamed board it never matched either, so the override was refused for exactly the tasks that had legitimately reached the end node.

Both fail in the direction that looks like normal behaviour rather than an error.

## Why the lanes are injected rather than resolved in place

`validateNodeOverrideChange` is **synchronous by design**, and its existing `isTerminalNodeId` option already establishes the pattern: callers with cheap IR access inject, callers without keep a documented literal fallback.

**Both production callers now supply the lanes** — `branch-and-pr-entities.ts:594` (which already injected `isTerminalNodeId`) and `task-update.ts:53`. That was the deciding factor: an optional parameter that only tests fill is the inert-injection shape this program keeps finding, where a guard reads as converted, its test passes because the test injects the value, and production keeps the literal. I checked both call sites had a store in scope *before* adding the option.

`resolveNodeOverrideLanes` lives beside the guard rather than in the callers, so the two cannot drift about what "executing" and "completed" mean.

## Fallbacks

A workflow expressing **no trait on any column** is a v1 upgrade — `synthesizeDefaultColumns` emits `traits: []` everywhere — not a board without these roles, so it keeps the legacy ids. Same for an unresolvable workflow. Both preserve exactly the behaviour the literals already had.

## Verification

- **Mutation-verified per guard:** restoring `task.column === "in-progress"` fails a case; restoring `task.column !== "done"` fails a different one.
- The suite also pins the paired negative — resolving lanes must not turn the guard into a blanket refusal for a task outside every WIP lane.
- `node-override-guard.test.ts` → 27 passed
- `pnpm test:gate` → 161 + 487 + 13 + 71
- `--strict` → exit 0; `tsc --noEmit` and `pnpm lint` → 0 errors

Census: batch-core scope **45 → 43**; repo total **255**.

## Summary by CodeRabbit

- **Bug Fixes**
  - Node overrides now correctly recognize workflow-defined in-progress and completed lanes, including renamed columns.
  - Override validation falls back safely for legacy or unresolved workflows.
  - Prevented validation from using stale task-column information during updates.

- **Tests**
  - Added coverage for workflow lane resolution, legacy fallbacks, renamed lanes, and override eligibility.

### DIFF
--- a/packages/core/src/__tests__/node-override-guard.test.ts
+++ b/packages/core/src/__tests__/node-override-guard.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, it } from "vitest";
 
-import { validateNodeOverrideChange } from "../node-override-guard.js";
+import { validateNodeOverrideChange, resolveNodeOverrideLanes } from "../node-override-guard.js";
 
 /*
 FNXC:WorkflowResolvedColumns 2026-07-30-22:35 (batch-core):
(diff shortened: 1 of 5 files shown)


gsxdsm:
Both real, both mine, fixed in `ca171eca5`.

**The floor.** `resolveNodeOverrideLanes` seeded the legacy ids and added the resolved lanes on top. So a v2 board declaring `in-progress` or `done` as an **ordinary untraited column** still had them treated as wip/complete — a valid mid-flight override blocked, a terminal override refused. Widening a guard onto columns the board explicitly does not trait is a regression, not a fallback, and it is the opposite failure from the one this PR set out to fix.

Now three states, the split this program has settled on elsewhere:

| workflow | lanes used |
|---|---|
| resolved, traits expressed | the resolved lanes **alone** |
| resolved, no trait anywhere | legacy — a v1 upgrade, where `synthesizeDefaultColumns` emits `traits: []` and the legacy ids are the only vocabulary that exists |
| unresolvable | legacy — today's behaviour |

**The race.** The task was read *before* the lane-resolution await, so a move landing in that window was judged with a stale column against freshly resolved lanes — and the dangerous direction is the obvious one: a task that entered a WIP lane during the gap still carried its pre-move column and passed the mid-flight guard. `resolveNodeOverrideLanes` needs only the task id, so the order is free; lanes now resolve first and the column comes from the latest read before validation.

Stated precisely, because it would be easy to overclaim: this does **not** make the check atomic. `updateTaskUnlocked` runs outside the per-task lock by design, as the note above it explains. It removes the window *this change introduced* rather than leaving a new one behind a resolved-lane improvement.

---

**A test gap the first finding exposed, which I think is the more useful outcome.** My renamed-lane cases call `validateNodeOverrideChange` directly with hand-supplied sets. They pin what the guard *does* with a set and say nothing about how the set is *built* — and the floor bug lived entirely in the builder. Mutating the resolver back to a floor left **all** of them green.

Added a resolver-level suite (`resolveNodeOverrideLanes` against a board that names `in-progress` without traiting it, a v1 upgrade, and an unresolvable workflow). That mutation now fails. Worth generalising: injecting a value into a guard makes the guard testable and the *resolver* invisible, so a converted call site needs coverage on both sides of the seam.

Also carries the baseline entry the first push warned about as uncommitted — `node-override-guard.ts` reaching 0 guards; `--strict` fails without it.

Suite 31 passed, gate 161 + 487 + 13 + 71, `--strict` 0, `tsc` and lint clean.

gsxdsm:
Real — and it is my own half-conversion. Fixed in `586f6b001`.

I closed exactly this race at the sibling call site (`branch-and-pr-entities.ts`) by resolving lanes **before** the task read, and left the second one alone. That is the half-converted-pair shape this program keeps finding — one call site fixed, its twin untouched — arriving inside my fix for it. Two call sites, one guard: both had to move together and I only moved one.

**Why the same fix does not apply here.** At the sibling site the task read exists solely for the guard, so hoisting the resolution above it was free. In `task-update` the `task` is loaded far above and is the working copy the whole function mutates — reordering it would change what every later step sees. So the column is **re-read after the await**, and only for the guard.

Without it: another process moving the card into a resolved WIP lane during the lane-resolution await left the mid-flight refusal judging a stale non-WIP column, and a task that had started running passed the guard.

The re-read is best-effort — on failure the already-loaded copy is used, which is strictly no worse than before this change.

**What I am not claiming:** neither site is now atomic. `updateTaskUnlocked` runs outside the per-task lock by design, as the note above the sibling call site explains. Both fixes remove the window *this PR introduced*; they do not add locking that was never there.

`node-override-guard.test.ts` → 31 passed, gate 161 + 487 + 13 + 71, `tsc` and lint clean.

gsxdsm:
Correct, and this one is serious — I shipped a deadlock. Fixed in `2e855138a`.

`updateTaskUnlockedImpl` is named for exactly this contract: the caller already holds the per-task lock. My re-read used `store.getTask(id)`, which acquires that same non-reentrant lock, so the inner read waited on the outer update forever — on **every** `nodeId` update, not just a contended one.

**The trade I made was strictly wrong**, and worth stating plainly rather than just patching: a stale-column race is a narrow window that needs a concurrent move to land inside one await. A deadlock is certain and total. I fixed a rare race by introducing a guaranteed hang, which is a worse outcome than the bug I was closing — and I did it while explicitly reasoning about that call site's locking, since the comment two lines above it describes the non-reentrant lock. I read that note and still called the locking API.

The fix uses `readTaskJson`, the lock-free read this function already uses for its own working copy, so the column still refreshes after the await without touching the lock. It falls back to the copy loaded above if the re-read fails, which is no worse than before the change.

**On what this says about the original fix.** The race is real, but the sibling call site (`branch-and-pr-entities.ts`) closes it by reordering, with no extra read at all — that one is clean. Here the ordering was unavailable, and reaching for a second read is what created the hazard. If the re-read ever proves troublesome for another reason, dropping it entirely is a safe retreat: that restores the pre-PR behaviour and leaves only the narrow race, rather than anything worse.

Gate green (161 + 487 + 13 + 71), `node-override-guard` 31 passed, `tsc` and lint clean.

PR #2822, opened 2026-07-30
fix(gate): barrel imports were silently dropped from the inert-seam check, hiding 5 engine omissions

gsxdsm:
**Self-reported regression from #2772, which has already merged.** The seam gate on main currently prints a clean bill of health while five real omissions sit in front of it.

## What broke

Closing the imported-shadow hole in #2772 taught the check to record which module each callee was imported from, and to exclude a call site whose module basename does not match the seam's declaring module. That works for relative imports. It does not work for barrel imports.

Engine and CLI reach core through `import { ... } from "@fusion/core"`. That specifier's basename is `core`, which never matches a module name like `near-duplicate-canonical` — so **every barrel-imported call site was classified as "a different function of the same name" and dropped.** The check stopped seeing engine's and cli's calls into core entirely, which is most of the cross-package surface it exists to watch.

## Measured

On `main` today:

```
[check-inert-flag-seams] 21 lane/flag seams, all supplied at every production call site.
```

With this fix:

```
packages/core/src/near-duplicate-canonical.ts: isNearDuplicateCanonicalInactive()
  — supplied by 6/11 call sites; omitted at
    packages/engine/src/self-healing.ts (x2), packages/engine/src/triage.ts (x3)
```

Those five were always there. Earlier I reported this seam as "supplied by 5/6" — that number was wrong for this reason, and the engine sites were invisible to me when I said it.

## The rule now

Only a **relative** specifier identifies a module well enough to exclude a call site on. Anything else is unresolved, and unresolved must mean **counted**: an over-counted seam produces a false report somebody investigates, an under-counted one produces silence. I had this backwards, and it is the second time in this lane a change made the gate read cleaner while catching less.

## Both directions verified

- **Barrel imports counted** — the five engine sites appear.
- **Relative shadows still excluded** — lifting the `sortTasksForDisplayColumn` entry still reports it unsupplied, so `Lane`/`Board`/`ListView` calling the dashboard twin through `"./taskSorting"` does not clear core's seam. That was the entire point of the original fix and it still holds.

## The five engine omissions

Not fixed here — engine-owned, reported on #2785. Same shape as the merge-queue bug in #2819: a canonical resting in a **renamed active column** reads as *inactive*, so duplicate markers get cleared against live work.

They carry TEMPORARY per-file exemptions so this PR is green and self-announcing. Noted at the entry: the key is `<file>::<function>`, so a file with two omitting calls is exempted for both — coarser than I want, recorded rather than left to be discovered.

## Verification

`pnpm test:gate` green, lint 0.


gsxdsm:
Real — and it is this gate's own failure mode, one level up. Fixed in `bfee0f406`.

`<file>::<function>` exempted **every** call to that function in that file, present and future. A call added later without the flags was silently covered and the gate stayed green. An exemption that grows to fit whatever arrives is not an exemption, it is a hole.

That is precisely the defect this check exists to catch: its own header records that `isTaskStuck` shipped with two of three call sites unsupplied while the gate was green, because the third was right. The same partial-supply blindness had been rebuilt inside the exemption mechanism.

**Each entry now records how many omissions were reviewed.** Extras beyond that count are reported like any other unsupplied site, so a new call site cannot inherit someone else's review.

**And an overshooting count is the same hazard in miniature** — a count of 2 where only one site omits pre-authorises an omission nobody looked at. That is now reported as stale alongside the existing "no site remains" rule, so the number has to be narrowed in the change that fixed the site.

**Verified by exit code rather than by reading output**, since a gate that prints a warning and exits 0 is exactly the thing under repair:

| scenario | exit |
|---|---|
| a second unsupplied call added to an exempted file | **1** — names `omitted at …:1 (of 2)` |
| entry claims `count: 2` where 1 site omits | **1** — "lower it" |
| unchanged tree | **0** |

The three existing entries are all `count: 1`, which matches their current reality — no exemption was widened or narrowed by this change.

Rebased onto `804a224fd`; `pnpm lint` 0 errors.

gsxdsm:

```
```

```
```

PR #2823, opened 2026-07-30
fix(core): duplicate markers were never cleared on a board with renamed terminal columns

gsxdsm:
Third unowned finding picked up after batch-core (#2783) merged without addressing its reports. Same class as #2819, opposite failure direction.

## The defect

`clearNearDuplicateReferencesTo` runs on every complete/archive/delete of a canonical task and clears the `nearDuplicateOf` markers pointing at it. It first asks `isNearDuplicateCanonicalInactive` whether the canonical really is finished — a safety check, so a live canonical's markers are not cleared out from under an operator.

That call omitted the canonical's resolved column flags, so it fell back to the legacy `done`/`archived` ids. On a renamed board a just-completed canonical (`shipped`, `filed`) read as **still active**, the guard early-returned, and the markers were **never cleared**. The flagged duplicates stayed parked behind a user decision that could never arrive — the exact stranding the predicate's own FNXC note says it was written to prevent.

## I had the direction backwards, and it matters

My first report of this seam described it as *markers cleared against a live canonical*. That is wrong. The legacy fallback errs toward "still active", so the failure is the opposite: markers that never clear at all. Same seam, same missing argument, entirely different symptom to look for — which is why the direction is worth pinning in a test rather than reasoning about.

## Measured

With the fix reverted, exactly one case flips:

```
✓ default vocabulary: completing the canonical clears the duplicate's marker
× renamed vocabulary: completing the canonical clears the duplicate's marker
✓ renamed vocabulary: a canonical still in the WIP lane does NOT clear the marker
✓ default vocabulary: a canonical still in the WIP lane does NOT clear the marker
✓ a soft-deleted canonical clears the marker under a renamed board
  Tests  1 failed | 4 passed (5)
```

With it: `Tests 5 passed (5)`.

## Both negatives included

A canonical still in the WIP lane must **not** clear its duplicates' markers, under each vocabulary. Resolving the real flags must not degrade into "every column is terminal", which would clear markers out from under an operator who has not made the duplicate decision yet. The soft-deleted path is covered too, since that branch never consults column flags at all.

## A fixture trap worth keeping

`sourceMetadata` is `jsonb` and must be seeded as an **object**. Seeding a stringified value reads back fine through `getTask` — it parses either shape — while the production query's `source_metadata->>'nearDuplicateOf'` matches nothing. The fixture looks correctly seeded and the code under test can never find the row. My first version had this, and the self-check in the seed helper is what caught it.

## Scope

Five of this predicate's six production call sites already resolved flags. This was the sixth, and the one that runs on every archive/complete transition.

Not touched here: the same function's SQL predicate excludes duplicates by literal `ne(column, "archived")` / `ne(column, "done")`. That is a per-row question across many rows in one statement, not a one-line supply, so it is flagged rather than guessed at.

The five **engine** call sites of this predicate are separately reported on #2785 and surfaced only via #2822.

## Verification

`pnpm test:gate` green, new suite 5/5, `tsc -p packages/core` 0, lint 0, changeset included.

### DIFF
--- a/.changeset/near-duplicate-clear-renamed-terminal.md
+++ b/.changeset/near-duplicate-clear-renamed-terminal.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Duplicate flags now clear when the canonical task finishes on a renamed board.
+category: fix
+dev: `clearNearDuplicateReferencesTo` resolves the canonical's own column flags before calling `isNearDuplicateCanonicalInactive`, which otherwise fell back to the legacy `done`/`archived` ids and read a completed canonical as still active, leaving `nearDuplicateOf` markers set forever.
--- a/packages/core/src/__tests__/postgres/near-duplicate-clear-renamed-terminal.pg.test.ts
+++ b/packages/core/src/__tests__/postgres/near-duplicate-clear-renamed-terminal.pg.test.ts
@@ -1,0 +1,198 @@
+/*
+FNXC:WorkflowResolvedColumns 2026-07-30-03:20 (duplicate markers never cleared on a renamed board):
+
+`clearNearDuplicateReferencesTo` runs whenever a canonical task is completed, archived, or deleted,
+and clears the `nearDuplicateOf` markers pointing at it. It first asks
+`isNearDuplicateCanonicalInactive` whether the canonical really is finished — a safety check, so a
+live canonical's markers are not cleared out from under an operator.
+
+That check was called WITHOUT the canonical's resolved column flags, so it fell back to the legacy
+`done`/`archived` ids. On a board whose terminal lanes are named anything else, a canonical that had
+just been completed read as still ACTIVE, the guard early-returned, and the markers were never
+cleared. The flagged duplicates stayed parked behind a user decision that could never arrive.
+
+THE DIRECTION IS THE POINT, and I had it backwards in my first report of this seam. The failure is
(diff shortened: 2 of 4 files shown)


gsxdsm:
**Confirmed and fixed — but driving the production transition alone did NOT distinguish the fix, and that is the more useful finding.**

The finding was right about the shape: the test supplied `column: "shipped"` by hand, which proved the consumer resolves flags correctly and hid the fact that nothing in production gives it that value. Both completion call sites in `moves.ts` gate on the *resolved* complete lane and then passed the literal `column: "done"`. They now pass `toColumn`, the lane the card actually reached.

**What rewriting the test to drive `moveTask` end to end actually showed.** I mutated both call sites back to `"done"` and the suite stayed green. The consumer looks the passed column up in the canonical's own IR, finds nothing for `done` on a renamed board, falls through to the legacy predicate — where `done` is terminal. Right outcome, wrong reason. Driving production was necessary but not sufficient; on its own it would have shipped a test that cannot fail.

It stops being an accident on a board that **declares** a `done` column *without* the complete trait: the literal then resolves real flags, learns `done` is not terminal there, early-returns, and strands every duplicate marker. That case is added, and it is the only shape under which the argument is load-bearing.

Note on the mutation: a **single-site** mutation passes, because the other site serves the test. Both had to be reverted together to see the failure — the same half-converted-pair trap the fix itself is about.

6 cases green; gate green (161/487/13/71); typecheck and lint clean.

gsxdsm:

```
```

```
```

PR #2828, opened 2026-07-30
test(engine): live-PG coverage for the timing role — and it is correct

gsxdsm:
## What

One new live-PostgreSQL E2E suite, 2 tests. **No production file is touched** — evidence, per the E2E worker's remit.

`packages/engine/src/__tests__/workflow-timing-trait-live-e2e.pg.test.ts`

## This one is a clean bill of health

Coverage for the last lifecycle role that had none — and the seam turns out to be **correct**. Recorded deliberately: a working conversion with no test is one refactor away from a silent regression, and this program should be able to say which seams are right, not only which are broken.

`cumulativeActiveMs` accrues while a card sits in the WIP lane, and both halves of the segment boundary resolve that lane by role:

```ts
const isWip = (column) =>
  inRole(column, ctx.lifecycleColumnSets?.wip, ctx.lifecycleColumns?.wip, "in-progress");
```

Note the literal at the end. That is the **same optional-parameter shape** this series found broken at four other seams — `shouldHoldActiveFileScopeLease` (#2795), `evaluateParkedAgentTaskLink` (#2798), `resolvePlanningContinuationCandidate` (#2799), `hasAutoHealableVerificationBufferFailure` (#2802) — where a caller failed to supply the resolved answer and the literal silently took over. Here the move path **does** supply it. Measured on a live store:

```
default: after exit cumulativeActiveMs=84
renamed: after exit cumulativeActiveMs=75      <- accrues on `building`, not just `in-progress`
```

## What it guards

If the move path ever stops populating the resolved columns, the literal takes over and a renamed board's cards accrue **exactly zero** active time — silently, because zero is a legitimate value for a card that has not run. The operator sees no error, just wrong numbers on every custom board.

**Mutation-verified against precisely that regression:** blinding the resolved answer (`inRole(column, undefined, undefined, "in-progress")`) fails the renamed case and leaves the control green. This is a guard with demonstrated power, not a test that happens to pass.

## Both boundaries are asserted

`cumulativeActiveMs` is `0` while the card is still in WIP and only accrues when it leaves; `firstExecutionAt` is stamped on entry. A test that checked only the final number would pass against an implementation that accrued at the wrong boundary, so both are pinned.

## Verification

- new suite — **2/2 passed**, mutation-verified
- full live-PG E2E surface — **161/161**
- `pnpm lint` — clean

### DIFF
--- a/packages/engine/src/__tests__/workflow-timing-trait-live-e2e.pg.test.ts
+++ b/packages/engine/src/__tests__/workflow-timing-trait-live-e2e.pg.test.ts
@@ -1,0 +1,109 @@
+/*
+FNXC:WorkflowResolvedColumns 2026-08-01-09:10 (E2E evidence — the `timing` role, and it is CORRECT):
+
+Coverage for the last lifecycle role with none, and the result is a clean bill of health — recorded
+because a working conversion with no test is one refactor away from a silent regression, and because
+this program should be able to say which seams are RIGHT, not only which are broken.
+
+`cumulativeActiveMs` accrues while a card sits in the WIP lane. Both halves of the segment boundary
+resolve that lane by ROLE:
+
+    const isWip = (column) =>
+      inRole(column, ctx.lifecycleColumnSets?.wip, ctx.lifecycleColumns?.wip, "in-progress");
+
+Note the literal at the end. It is the same optional-parameter shape that this series found broken at
+four other seams — `shouldHoldActiveFileScopeLease` (#2795), `evaluateParkedAgentTaskLink` (#2798),
+`resolvePlanningContinuationCandidate` (#2799), `hasAutoHealableVerificationBufferFailure` (#2802) —
+where a caller did not supply the resolved answer and the literal silently took over. Here the move
+path DOES supply it, so the seam works. Measured on a live store rather than assumed:
+
+    default: after exit cumulativeActiveMs=84
+    renamed: after exit cumulativeActiveMs=75      <- accrues on `building`, not just `in-progress`
+
+WHAT THIS GUARDS. If the move path ever stops populating the resolved columns, the literal takes over
+and a renamed board's cards accrue ZERO active time — silently, because zero is a legitimate value for
+a card that has not run. The operator sees no error, just wrong numbers on every custom board. This
+suite turns that into a red test; blinding the resolved answer is exactly the mutation it was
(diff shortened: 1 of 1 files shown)


gsxdsm:
### Pre-flight against current `main`: 3 inherited, 1 worth a look

Local merge with `origin/main`:

```
core:    4 failed | 4803 passed
engine:  11018 passed | 0 failed
```

**Three are red on `main` itself**, not from your change:

- `workflow-ir-resolver` ×2 — #2815's branded-copy fallback vs `toBe` identity. Fixed in **#2836** (merged).
- `store-archive-reads` — #2832 changed the restore destination. Fix up in **#2847**.

**The fourth did not reproduce on `main`** and is worth your attention:

```
postgres/schema-applier.test.ts > VAL-SCHEMA-001 final-schema parity (table counts)
  > retains unreplaced registered dependents for every delete action
```

I ran that file directly on `origin/main` and it **passed** (80 tests, only the archive one red), so it appeared in your merged tree and not in main's. That is either something in your diff or a flake — I did not get a second run to distinguish them, so I am not claiming which.

Given the test is about schema table counts and your PR adds live-PG timing coverage, a single confirming run on your side would settle it. If it turns out to be a flake, please capture full output rather than the summary line — an unnamed flake cannot be quarantined, and I have lost two identities to that today already.

Engine is fully green on your branch.

gsxdsm:
### Follow-up on the `schema-applier` failure — narrowed

I said I had not distinguished flake from real. I have now, at least partly:

```
schema-applier.test.ts run in ISOLATION on your merged tree:
  run 1: 75 passed
  run 2: 75 passed
```

It fails only in the **full-suite** run and passes on its own, twice. So it is not a straightforward defect in the file — it is either suite ordering/parallelism or a flake, the same shape as the `plugin-runner` case on #2799 earlier today.

That makes it much less likely to be your change and much more likely to be a pre-existing full-suite interaction that your added tests happen to perturb. I would not hold the PR for it.

Still worth one full-suite run on your side before merging, and **capture full output** if it reappears — the individual `FAIL` lines are what a quarantine entry needs, and a `--reporter=dot | tail` capture throws them away. That has cost me two flake identities today.

Everything else stands: the other 3 core failures are inherited from `main` (#2836 merged, #2847 open), and engine is fully green at 11018 passed.

PR #2830, opened 2026-07-30
chore(core): delete the dead merge-blocker guard, and catch exemptions for deleted seams

gsxdsm:
Closes the last of the five findings I reported to batch-core (#2783), which merged without them. This one I held back twice on purpose; the reason is now resolved.

## The dead code

`evaluateMergeBlockerGuard` had **exactly one reference in the repo: its own declaration.** Never called, never re-exported from `index.ts`/`index.gate.ts`, never registered as a trait hook. Its `lifecycleColumns` conversion was applied to dead code, and the census counted it as progress.

## Why I would not delete it earlier, and what changed

I twice declined this, because no production `"guard"` trait hook is registered anywhere in the codebase — the only `registerTraitHookImpl(..., "guard", ...)` is in a test. If a registration had been dropped, that would be a real product bug and this function would be its evidence, so deleting it would have destroyed the breadcrumb.

**It is not missing.** Merge blocking is enforced inline in `task-store/moves.ts` (~645 and ~821) via `getTaskMergeBlocker`, gated on the **resolved trait flags** (`toFacts.flags.complete` + `fromFacts.flags.mergeBlocker`) rather than on column ids — a better implementation than the one being deleted. The logic moved; this function, and a file-header line crediting a never-existent `evaluateDefaultWorkflowGuards` reader, were left behind.

The header now records where the guard actually lives, so the next person does not repeat the investigation I just did.

Also removed: `GuardVerdict` (its return type, used nowhere else) and the orphaned `getTaskMergeBlocker` import — the latter caught by lint, not by me.

## The gap this exposed

The allow-list staleness check only fired for a seam the scan still **finds** — it asks *"is this site supplied now?"*. Delete the declaration and the name is never iterated, so its entry sits in the list forever, exempting nothing and misreporting what is tolerated.

I found this by deleting the function and watching the gate stay **silent** about its leftover entry.

**Measured:** an `ALLOWED` entry naming a non-existent seam now fails with `no such seam declared any more; remove its ALLOWED entry`; removing the probe returns exit 0. The failure header is reworded, since "the sites are supplied now" no longer covers both reasons.

This is the fourth blind spot closed in this check, and like the others it was found by exercising the gate rather than reading it.

## Verification

`pnpm test:gate` green · `tsc -p packages/core` 0 · lint 0 · gate now reports **20** seams (was 21 — the drop is this deletion).

No changeset: deleting unreachable internal code with no exported surface is behaviour-preserving.

### DIFF
--- a/packages/core/src/default-workflow-hooks.ts
+++ b/packages/core/src/default-workflow-hooks.ts
@@ -15,8 +15,13 @@
  * path only.
  *
  * Hook classes (KTD-2):
- *   - guard  (sync, in-lock): merge-blocker, human-review. Implemented as the
- *     `evaluateDefaultWorkflowGuards` reader; pure DB-free reads off the task.
+ *   - guard  (sync, in-lock): merge-blocker, human-review. NOT implemented here.
+ *     The header used to credit an `evaluateDefaultWorkflowGuards` reader in this
+ *     file; no such function has ever existed. The merge-blocker guard is enforced
+ *     inline in `task-store/moves.ts` (~645 and ~821) via `getTaskMergeBlocker`,
+ *     gated on the RESOLVED trait flags (`toFacts.flags.complete` +
+ *     `fromFacts.flags.mergeBlocker`) rather than on column ids — which is why no
+ *     `guard` trait hook is registered below and none is missing.
  *   - onEnter / onExit (mutating, applied in-lock to the in-memory task before
  *     the commit for field effects; queue effects run in-txn): timing,
  *     reset-on-entry, abort-on-exit, merge.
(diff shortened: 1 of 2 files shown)


gsxdsm:
Real — fixed in `5398e6744`.

Both staleness rules for the per-site list live *inside* the per-declaration loop, so they only run for a function the scan still finds. Delete or rename the function and that loop never visits it: the `<file>::<fn>` entry survives, exempting nothing, while reading to the next person as a reviewed and tolerated omission.

**This is the identical defect the `ALLOWED` deleted-seam pass in this same PR was added to fix** — written for one of the two lists and not the other. The header on that pass even records how it was found ("by deleting `evaluateMergeBlockerGuard` and watching the gate stay quiet"), and the same experiment on an `ALLOWED_OMISSIONS` entry was equally quiet.

The new check sits immediately beside it rather than being folded into the per-declaration loop, because that loop structurally cannot see a deletion — it iterates what the scan found.

**Verified by exit code**, since a gate that prints and exits 0 is the failure mode here:

| scenario | exit |
|---|---|
| rename the function an `ALLOWED_OMISSIONS` entry references | **1** — reports both stale entries by key |
| unchanged tree | **0** |

Worth noting for whoever merges these two: I made a related change to the same file on #2822 (exemptions bounded by count, so a *new* unsupplied call cannot inherit an existing review). The two are independent — this one is about entries outliving their function, that one about entries covering more calls than were reviewed — but they touch adjacent code and will likely conflict textually. Whichever lands second should keep both passes; neither supersedes the other.

gsxdsm:
### Pre-flight against current `main`: inherited only

Local merge with `origin/main`, core suite:

```
2 failed | 4806 passed | 6 skipped
```

Both failures are red on `main` itself, not from your change:

- `workflow-ir-resolver` ×2 — #2815's branded-copy fallback vs `toBe` identity assertions. Fixed in **#2836** (merged).
- `store-archive-reads` — #2832 changed the restore destination; a characterization case in another file was not flipped. Fix up in **#2847**.

Rebase once #2847 lands and this should come back clean. Nothing to do on your side.

PR #2832, opened 2026-07-30
fix(core): restore an archived card to the lane it came from

gsxdsm:
## What

Fixes the defect #2824 measured. That PR's characterization cases — merged and asserting the wrong-but-real behaviour — are flipped here to the correct lanes, which is what they were written to do.

## The bug

```ts
// archive-lifecycle-2.ts
const preArchiveColumn = task.preArchiveColumn ?? "todo";
```

**`preArchiveColumn` has no database column.** It exists on the `Task` type and in the archive snapshot, and nowhere else — so the in-place restore cannot carry it, `store.getTask(id)` reads a live row that never had it, and the literal decided the destination for **every unarchive that has ever run**.

| board | what happened |
|---|---|
| **default** | `todo` is declared, so the resolver returned it. Restores landed in the queue and **looked right**. |
| **custom** | `todo` is declared nowhere, so the resolver took its "no usable history" branch and returned the **complete** lane. A card archived mid-implementation came back marked **finished**. |

That coincidence is why this survived **three** separate fixes to `resolveUnarchiveTargetColumnImpl` — a `?? "done"` that invented a column, an `isColumn` legacy-enum gate, and the same gate one function over. Every one was correcting how the resolver interprets a value that never arrived.

## The fix is two halves, and either alone does nothing

1. **Capture** — `taskToArchiveEntryImpl` records `task.column` into the snapshot. That is the last place the original is still in hand, since the entry's own `column` is set to `"archived"` on the line above.
2. **Read** — `unarchiveTaskImpl` reads the **snapshot it already loaded**, not the restored row.

I shipped half of this first and watched the destination stay wrong, which is how I found that the field has no row to live on. Mutation matrix:

| state | result |
|---|---|
| both halves | **5/5 pass** |
| capture only (read reverted) | **3 fail** |
| read only (capture reverted) | **3 fail** |

I also tried carrying it through `restoreTaskFromArchive`'s row update — that fails to typecheck, which is the proof that no such column exists and the snapshot is the only source.

## Behaviour changes, deliberately

- **Custom boards** — a card returns to the lane it was archived from instead of appearing finished.
- **Default board** — a card archived from `done` restored to `todo` under the literal and now restores to `done`. Returning finished work to the queue was the fallback showing through, not a rule anyone chose; the resolver's own branches say a card archived from a declared column goes back to it.

## One expectation of mine was wrong, and the resolver was right

I expected a card archived from the review lane to return to **hold**. It returns to the review lane, and that is correct: `.review` is derived from the `mergeOrchestration` flag, **not** from `human-review`. The fixture's review column declares `human-review` + `merge-blocker` only, so it is not a `.review` lane to the resolver — just a declared column with usable history. The case now asserts that with the reasoning attached, so the next reader does not "fix" it back to hold.

## Verification

- unarchive suite — **5/5**, mutation matrix above
- full live-PG E2E surface — **164/164**
- `pnpm lint` — clean

### DIFF
--- a/.changeset/persist-pre-archive-column.md
+++ b/.changeset/persist-pre-archive-column.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Restore an archived card to the lane it was archived from, not to Done.
+category: fix
+dev: `preArchiveColumn` was never captured — it has no `project.tasks` column and lives only in the archive snapshot — so `unarchiveTaskImpl` always fell to its `?? "todo"` literal. On a custom workflow `todo` is undeclared, so the destination resolver took its no-usable-history branch and returned the complete lane. Captures the column into the snapshot on archive and reads the snapshot (not the restored row) on unarchive; both halves are required.
--- a/packages/core/src/task-store/archive-lifecycle-2.ts
+++ b/packages/core/src/task-store/archive-lifecycle-2.ts
@@ -44,7 +44,31 @@ export async function taskToArchiveEntryImpl(store: TaskStore, task: Task, archi
       description: task.description,
       priority: normalizeTaskPriority(task.priority),
       column: "archived",
-      preArchiveColumn: task.preArchiveColumn,
+      /*
+      FNXC:WorkflowLifecycleColumns 2026-08-01-11:30 (PR #2824's finding, fixed):
+      CAPTURE THE COLUMN THE CARD WAS IN. This field was only ever COPIED — here, back out of the
+      entry on restore, and through serialization — and never SET from anywhere, so it was `undefined`
+      for every archive that has ever happened. `unarchiveTaskImpl` then fell to its `?? "todo"` and
+      the restore destination was decided by a literal instead of by history.
+
+      On the default board `todo` is a declared column, so restores landed in the queue and looked
+      right — which is why this survived three separate fixes to `resolveUnarchiveTargetColumnImpl`,
+      all of which were correcting how it interprets a value that never arrived. On a renamed board
+      `todo` is declared nowhere, so the resolver took its "no usable history" branch and returned the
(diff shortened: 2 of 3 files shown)


gsxdsm:
### A second behaviour this fix has, which the summary does not describe

Found while clearing the red this left in `store-archive-reads.pg.test.ts` (a characterization case in a different file that #2832 did not flip — fix up in **#2847**).

The restore rule is **narrower than "the lane it came from"**. Measured on current `main`, same test, only the fixture's starting column changed:

| archived from | restored to |
|---|---|
| `done` | `done` |
| `in-progress` | **`todo`** |

So a terminal lane is preserved while a WIP lane is re-queued.

That may well be deliberate and correct — a card cannot resume mid-execution after a restore, and dropping it back into the queue is the safe answer. But it is not what this PR's description says, and the difference matters for exactly the boards this fix was written for: the custom-board case in your table was a card *"archived mid-implementation"*, i.e. the WIP case, and it still does not come back to the lane it left.

Two readings, and choosing between them is yours rather than mine:

1. **Deliberate** — WIP re-queue is the intended policy. Then it deserves its own named case, because right now nothing asserts it and the next person to read the summary will expect `in-progress`.
2. **Not deliberate** — the snapshot-rebuild path (live row deleted, snapshot is the only source) still takes the legacy fallback for non-terminal lanes, and the defect you fixed survives there.

I deliberately did not encode either in #2847. I left that test asserting `done` and recorded the observation, because a rule inferred from two samples is a guess.

One more thing worth knowing: the `done` assertion is **weaker than it looks**. `done` is also the complete lane, which is what the pre-fix "no usable history" branch returned — so a card archived from `done` reads identically under both the fixed and the broken implementation. If you want a case that can actually tell them apart, it has to start from a non-terminal, non-legacy lane.

PR #2835, opened 2026-07-30
docs(workflow-learnings): the fifth shape — converted consumer, literal-passing producer

gsxdsm:
Found by the operator reviewing **my own** shipped fix (#2823). It belongs in this document because it is the only shape so far that every instrument in the program reported as done.

## The defect

`clearNearDuplicateReferencesTo` was converted to resolve the canonical's column flags, and my test proved that by supplying `column: "shipped"`. Both production call sites in `moves.ts` gated on the **resolved** complete lane and then passed the literal `column: "done"`. The consumer was correct; the producer was never converted; and the hand-supplied fixture value is exactly what hid it.

Why nothing caught it:

- the **census** counts comparisons — a value passed as an argument is not a comparison;
- the **seam check** asks whether callers *supply* the argument — these did, with a literal;
- the **test** supplied the interesting value itself, so it exercised the consumer and never the producer.

And the part I would have got wrong: driving the flow end to end still does **not** distinguish them. The consumer looks the passed column up in the canonical's IR, finds nothing for `done` on a renamed board, and falls through to the legacy predicate where `done` *is* terminal — right answer, wrong reason. It only bites on a board that declares a `done` column **without** the complete trait.

## The rule

> A differential test must vary the value the **production** code computes, not one the test hands in.

If the fixture passes the lane name, it has tested the consumer. Who computes that argument in production, and do they compute it or spell it, is a separate question — and the one that was wrong here.

## Measured, so nobody builds the wrong instrument

An AST probe for call arguments shaped `{ column: "<legacy id>" }` finds **79 sites** across `packages/`. It correctly flags the two real `moves.ts` offenders — but most of the rest are legitimate: `set({ column: "archived" })` writing the archive state, `listTasks({ column: "todo" })` filtering a query.

So a blocking gate on this shape needs a curated list of consumers that interpret a column as a **role**, as opposed to storing or filtering it — a per-consumer judgment call, not a mechanical check. **Recorded rather than built**, and deliberately not attempted on top of five unmerged PRs in packages I do not own.

I also verified one nearby call site that a naive version of this probe flags and which is **not** a defect: `archive-lifecycle-2.ts:353` passes `column: "archived"`, but that path sets `task.column = "archived"` unconditionally, so it is passing the column the card actually reached. Exactly what the operator's fix is about.

Docs only — no changeset.


gsxdsm:
**Correct, and the imprecision is an instance of the very shape the section is about.** Fixed in 30d6969cc1.

Only the **pure predicate** `isNearDuplicateCanonicalInactive` was converted to take resolved flags. The store wrapper `clearNearDuplicateReferencesToImpl` still calls it flagless on `main`, so even a correct producer would not reach the converted branch. Writing "`clearNearDuplicateReferencesTo` was converted" collapsed three layers — predicate, call site, producer — into one name, which is the same conflation that let the original defect score as a win.

The section now names each layer and points at #2823 for the call-site and producer halves, with an explicit note that until that lands, `branch-group-ops.ts` invokes the predicate without flags on main.

Issue #2839, opened 2026-07-30
12 legacy column literals live inside SQL strings — invisible to the lifecycle census


gsxdsm:
Measured, not estimated. Filing rather than fixing: `resolveProjectColumnsForRoles` already exists and is being adopted, so this is a work list for whoever owns that migration, not something I should collide with.

## Why this surface was never counted

The lifecycle-column census parses TypeScript comparisons. A legacy id inside a **SQL string** is not a comparison — it is string data — so the census has never seen these, and neither does the inert-seam gate (which reasons about parameters and call sites).

I found this only because the operator fixed one on #2819: `cleanupStaleMergeQueueRowsImpl` filtered on `t.column != 'in-review'`, so on a renamed board every queued card looked stale, got its row deleted, and became unleaseable. **I had read that exact SQL earlier in the session and not flagged it.**

## The 12 sites

Found by AST walk over string/template literals only — comments are not AST nodes, so prose like `column === "done"` cannot produce a false positive. A line-based grep reports **37**, and 25 of those are comments; that is the number to ignore.

**Analytics — silently wrong numbers, no error:**
- `github-issue-analytics.ts:179`, `:209` — `"column" = 'done'`
- `gitlab-issue-analytics.ts:178`, `:208` — `"column" = 'done'`
- `productivity-analytics.ts:404` — `"column" = 'done'`
- `team-analytics.ts:336` — `"column" = 'done'`
- `workflow-analytics.ts:484` — `"column" = 'done'`

On a board whose complete lane is renamed, every one of these counts **zero** completed work. Throughput, cycle time, and team dashboards read as an idle project. Nothing errors, and wrong-but-plausible metrics are the least likely thing anyone files a bug about.

**Archive/lifecycle filters — rows land in or vanish from the wrong set:**
- `mission-store.ts:2353` — `"column" != 'archived'`
- `async-maintenance.ts:41` — `"column" = 'archived'`
- `workflow-definitions.ts:118`, `:911` — `"column" = 'archived'`

**Already fixed by the operator on #2819:**
- `task-artifacts-ops.ts:92` — `column != 'in-review'`

## The mechanism already exists

`resolveProjectColumnsForRoles(store, roles)` in `core/src/project-lane-vocabulary.ts` returns the column-id set for a role across all workflow definitions, degrading to the legacy ids when definitions are unreadable. It is already adopted in `eval-automation.ts:193` and `store.ts`. These 12 look like the un-migrated remainder.

[...]

PR #2841, opened 2026-07-30
feat(gate): freeze the SQL column-literal surface — 30 sites, none may be added

gsxdsm:
Instruments a surface no existing check can see. Follows #2839, and **corrects the count I reported there** (12 → 14).

## Why it was invisible

The lifecycle census parses TypeScript **comparisons**; a legacy id inside a SQL string is string data. The inert-seam gate reasons about parameters and call sites. Neither has ever looked here.

**What it cost:** `cleanupStaleMergeQueueRowsImpl` filtered on `t.column != 'in-review'`, so on a renamed board every queued card looked stale, its `merge_queue` row was deleted, and the card became **unleaseable**. The operator found it reviewing #2819 — in SQL I had already read past during that same work.

The quieter half is analytics: five sites count `"column" = 'done'`, so throughput, cycle time, and team dashboards report **zero completed work** on a renamed board. Nothing errors, which is why nobody files it.

## What this does, and does not do

It does **not** fix the sites. `resolveProjectColumnsForRoles` is the mechanism and its migration has an owner (#2839). This freezes the population so the surface cannot grow underneath that migration: a new file or a higher count fails, **and a lower count fails too** — so the baseline ratchets down as sites migrate rather than leaving slots to silently regrow into. That is the same rot as an allow-list entry for a deleted function, which this repo already hit once.

AST-based, deliberately: a line grep for the same pattern reports **37** hits, **25 of them prose** quoting `column === "done"` in explanatory notes. A guard that is 68% false positives trains its readers to skip it — a lesson this program has already paid for.

## Two corrections found by mutation-testing my own gate

**1. Clause fragments were missed.** Requiring a SQL keyword *in the same literal* skipped `team-analytics.ts`, which builds `["assignedAgentId IS NOT NULL", `"column" = 'done'`, ...]` and joins them into a `WHERE` later. That fragment is as vocabulary-bound as any full query but contains no keyword. Fixing it took the population **12 → 14**, so the number I put on #2839 was low.

**2. My first mutation test proved a direction it had not.** I replaced the first textual occurrence in a file — which was inside a **comment** — and read the unchanged count as the scanner being broken. The scanner was right; my test was wrong. All three directions are now driven against real SQL:

| mutation | result |
|---|---|
| add a full query with a legacy comparison | `3 SQL column literal(s), baseline allows 2` |
| add a bare clause **fragment** (no keyword) | caught — same failure |
| migrate one away (count drops) | `1 site(s) now, baseline still allows 2 — re-record it` |
| restore | exit 0 |

I am flagging that second one because it is the exact failure mode this program keeps finding: a green result read as evidence when the experiment was invalid.

## Verification

`pnpm test:gate` green with the new check in it · lint 0 · single AST pass. Wired into `test:gate` and both `pretest` hooks.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added automated checks to detect increases in legacy SQL column literals.
  * Added baseline tracking to ensure known SQL literal counts do not regress.

* **Tests**
  * Expanded pre-test and gated verification steps with SQL literal and mock completeness checks.
  * Updated test validation workflows to enforce the new safeguards.


gsxdsm:
**All four findings are real, they are one mistake, and the fix is deletion.** 09400dbf8e.

Two from greptile and two from coderabbit, against three lines. Every one reduces to the same thing: the gate decided *whether to run* its comparison regex, using cheaper patterns that disagreed with it.

| finding | why it was a false negative |
|---|---|
| file-level `SQL_SHAPE.test(source)` short-circuit | a file holding only a clause fragment has no `SELECT`/`WHERE` anywhere, so a new forbidden site could be added to it and the gate passed — the exact shape the fragment carve-out was written for, reintroduced one level up |
| `BARE_CLAUSE` is anchored `^...$` | a qualified or compound fragment (`t."column" = 'done'`, `("column" = 'done' OR active = 1)`) matched neither pattern, so `COMPARISON` never ran |
| `node.getText()` returns source text | a double-quoted TS string spells the identifier `\"column\"` with backslashes intact, while every pattern expects the decoded `"column"` |

**A gate whose false negatives are this easy to construct is worse than no gate**, because the baseline it prints reads as coverage. So the pre-filters are gone: one unanchored pattern over the **decoded** text of every string and template literal, with nothing left to disagree with. `SQL_SHAPE` and `BARE_CLAUSE` are deleted rather than left unused — an unused pattern in a gate is an invitation to re-add a filter that uses it.

**The false-positive argument is untouched, which is what makes this safe.** It never depended on the pre-filters: comments are not AST nodes, so walking literals cannot match prose however permissive the pattern gets. That property is now pinned as its own case.

**It immediately found a live site the old version missed.** `packages/core/src/workflow-analytics.ts:384` builds `` `t."column" = 'done'` `` as a clause-array fragment — qualified, so invisible to the anchored pattern. That file goes 1 → 2 and the baseline is re-recorded: better detection, not new debt. Exactly the shape greptile predicted.

**Seven cases added** (`scripts/__tests__/check-sql-column-literals.test.mjs`), written as the four specific shapes rather than one generic case so a future "optimisation" that reintroduces a pre-filter fails on the shape it breaks — plus two negatives (a non-legacy id must NOT match, or the gate would freeze converted code; prose must never match).

Mutation-verified: restoring `getText()` fails 1 of 7. `pnpm lint` clean; gate reports 15 known sites, none added.

gsxdsm:
**Both real, and the second one means this gate was blind on the exact files it was written for.** f788d83900.

**`IN` predicates.** The operator list was `=`, `!=`, `<>`, so `"column" IN ('in-progress', 'in-review')` contributed nothing and a second predicate in that form could be added with the baseline green. `team-analytics.ts` and `workflow-analytics.ts` each held one, unseen. Each legacy **element** counts, matching how the `=` arm counts two comparisons in one query; a list mixing a legacy id with a resolved one still counts the legacy one.

**Interpolated columns — the serious one.** `literalText` joined only the *static* template spans, on my reasoning that an interpolated expression cannot be part of a comparison. That is backwards for the dominant production shape: a Drizzle template puts the **column** in the hole and the legacy id in the static text, so `${schema.project.tasks.column} != 'archived'` joined to text with no column identifier in it and matched nothing. The merge-queue and self-healing queries this gate exists to freeze are written exactly that way — **it was blind on its primary target.**

Enabling that shape revealed **five previously invisible files**:

| | baseline before | after |
|---|---|---|
| total sites | 14 | **31** |
| `async-mission-store-queries.ts` | — | 1 |
| `async-mission-store.ts` | — | 2 |
| `task-store/async-archive-lineage.ts` | — | 3 |
| `task-store/async-merge-coordination.ts` | — | 1 |
| `task-store/reads.ts` | — | 1 |
| `team-analytics.ts` | 2 | 6 |
| `workflow-analytics.ts` | 2 | 6 |

Detection catching up with reality, not new debt.

**Two things I found while fixing these, both worth recording:**

1. A non-column interpolation now renders as a **NUL sentinel**, not a space. A space lets `` `"column" = ${expr}'done'` `` join into `"column" =  'done'` — a comparison that is not in the source. Caught by writing the paired negative first.
2. **The entry point is now guarded.** The test imports `COMPARISON`/`literalText`, and importing the module ran the entire gate and called `process.exit(1)`. So the test file failed for the *gate's* reasons rather than its own — and while the gate was green, it passed for reasons unrelated to what it asserts. A test whose result unrelated repo state can flip is not a test.

12 cases now (`node --test scripts/__tests__/check-sql-column-literals.test.mjs`), all green. `pnpm lint` clean; gate reports 31 known sites, none added.

gsxdsm:
**Correct — third false negative of the same family.** 96a6ac4733.

`[^)]*` stops at the first `)`, and any nested call supplies one. So `"column" IN (COALESCE(x, y), 'done')` never reached its legacy id and the predicate counted zero.

The `IN` body now tolerates nested groups **two levels deep**, covering the realistic worst case in this codebase (`LOWER(COALESCE(a, b))`).

**The bound is stated rather than hidden**, which is the part I want on the record. A regex cannot balance arbitrary nesting. The alternative — match the predicate head and extract the balanced region programmatically — buys a depth nobody writes here, at the cost of a second scanner to keep correct. At three levels this under-counts again. That is now a *known* limit with a comment naming it, rather than an unknown one; if a site ever nests that deep, the fix is the extraction approach, not a third regex tier.

Repo count is unchanged at **31** — no current site nests — so this closes a hole rather than finding one, unlike the previous two rounds.

Three cases added including the paired negative (tolerating nesting must not turn every `IN` predicate into a hit). 15 cases total, all green. Mutation-verified: restoring `[^)]*` fails 2 of 15. `pnpm lint` clean.

**Pattern worth noting across the three rounds of this PR:** every finding has been a false *negative*, and each arrived only after the previous one was fixed — pre-filters, then `IN`, then nesting. The gate's shape is "does this text look like a forbidden comparison", and each round has been a text shape I did not anticipate. The 15 cases now exist so the next one fails a specific test rather than silently widening the blind spot.

gsxdsm:
Updating this PR's own description, because it no longer matched its code.

**The population is 30, not the 14 in the original title.** Two rounds of review found five classes of false negative in my scanner, all fixed on this branch:

- **file-level `SQL_SHAPE` pre-filter** skipped whole files — a file holding only a clause fragment has no SELECT/WHERE anywhere, so a new forbidden site could be added to it and the gate stayed green. That is the exact shape the fragment carve-out existed for, reintroduced one level up.
- **`BARE_CLAUSE` was `^...$`-anchored**, so a qualified or compound fragment (`t."column" = 'done'`, `("column" = 'done' OR active = 1)`) matched neither pattern and was never compared.
- **`node.getText()` returns SOURCE text**, where a double-quoted TS string spells `\"column\"` with backslashes intact while every pattern expected the decoded form.
- **`IN` predicates** were not in the operator list, so `"column" IN ('in-progress','in-review')` counted nothing.
- **interpolated columns — the serious one.** I joined only the STATIC template spans, reasoning that an interpolated expression cannot be part of a comparison. That is backwards for the dominant production shape: a Drizzle template puts the COLUMN in the hole and the legacy id in the static text, so `${schema.project.tasks.column} != 'archived'` joined to text containing no column identifier and matched nothing. **The merge-queue and self-healing queries this gate exists to freeze are written exactly that way — it was blind on the files it was built for.** Enabling that shape revealed five previously invisible files.

I am spelling these out rather than quietly updating a number: a gate whose false negatives are this easy to construct is worse than no gate, because the baseline it prints reads as coverage. My mutation tests all passed because I only ever mutated shapes my own scanner already understood — the blind spots were outside the space I probed.

The pre-filters are gone entirely now: one unanchored pattern over the DECODED text of every string and template literal, with nothing left to disagree with it. The false-positive argument is unaffected, since comments are still not AST nodes.

**Also in this push:** baseline 31 → 30. main migrated the last legacy comparison out of `task-store/async-merge-coordination.ts`, and the check failed because that file's entry still allowed 1 — the downward ratchet firing in real use rather than in a mutation test.

gsxdsm:
**Fourth round, same conclusion: correct.** be50b65519.

`schema.project.tasks["column"]` is the *same reference* as `schema.project.tasks.column`, and only the dot form was matched — so an element-access reference fell through to the NUL sentinel and its entire predicate vanished. **The identical blindness the static-span join had, reachable by changing punctuation.** Drizzle accepts both spellings and a formatter or a reserved-word column name can produce the bracket one, so it is not a hypothetical shape.

Both quote styles are now tolerated, for the same reason the rest of this scanner is permissive: a false positive costs one baseline entry, a false negative costs the gate its meaning. The paired negative is covered — a bracket access to a *different* property must not read as a column, or every interpolation becomes one. 18 cases; repo count unchanged, so this closes a hole rather than finding one.

**Four rounds in, the pattern in the findings is worth naming, because it is a property of this gate's design and not of any one bug.** Every finding has been a false *negative*, each arrived only after the previous was fixed, and each was a text shape I had not modelled: pre-filters, then `IN`, then nesting, now bracket access. That is exactly what a regex-shaped specification does — it enumerates the spellings you thought of, and the ones you did not are silently absent from the count.

The 18 cases mean the next unmodelled shape fails a specific named test rather than widening the blind spot invisibly. But the structural fix, if a fifth round arrives, is to stop matching *text* and resolve the column reference through the TypeScript type-checker — which is a materially bigger instrument than a script, and worth doing only if the shapes keep coming. Flagging that as the escalation path rather than adding a fifth regex tier by reflex.

gsxdsm:
**Correct, and this one is squarely my error rather than an unmodelled shape.** Pushed.

The previous round's comment says *"both quote styles and an optional trailing `!`/`?` are tolerated"*. The regex implemented only the quote styles. **The comment described the intent, the regex described the behaviour, and only one of them runs** — so a non-null-asserted column reference still fell through to the NUL sentinel and its predicate vanished.

A comment that overstates a guard is worse than no comment, because it is the thing a reader checks *instead of* the code. That is the same failure this whole PR is about, one layer up: a claim of coverage that nothing executes.

Four cases now pin the behaviour instead of describing it — bracket-with-`!`, dot-with-`!`, optional-chained, and the negative (an assertion on a *different* property is still not a column). 22 cases total; repo count unchanged, so this closes a hole rather than finding one.

**Fifth round, and the escalation note from the last comment now has a concrete trigger.** Four of the five findings were unmodelled text shapes; this one was a documented-but-unimplemented shape. Both are symptoms of specifying a *reference* by its spelling. If a sixth arrives, the fix is to resolve the column reference through the TypeScript type-checker rather than adding a sixth pattern tier — I would rather that decision be made deliberately than reached by attrition.

gsxdsm:
**Sixth round, and unlike four and five this one was hiding a live site.** fbf91fe4d1.

`packages/core/src/task-store/async-merge-coordination.ts:320` writes:

```ts
sql`${schema.project.tasks.column} IS DISTINCT FROM 'in-review'`
```

That is the **merge-queue stale sweep** — one of the very queries this gate exists to freeze. The operator list held `= != <>` and `IN`, so the predicate counted zero and that file's baseline entry read 0 instead of 1. Baseline 30 → 31: detection catching up, not new debt.

`IS` and `IS NOT` are covered in the same change rather than left for a seventh round. **Enumerating operators one review round at a time is how the previous five holes happened**, and the marginal cost of including the neighbouring spellings now is a line of regex. 26 cases, including the negative (`IS DISTINCT FROM 'shipped'` must not match).

**Six rounds is enough evidence to state the conclusion rather than keep patching.** The tally: one pre-filter defect, and five cases where the pattern simply did not enumerate a spelling — `IN`, nested parens, bracket access, `!`/`?`, `IS DISTINCT FROM`. Every one a false negative; every one found by a reviewer rather than by the mutation tests, because a mutation you write is a mutation you already imagined.

That is not bad luck, it is what a regex-shaped specification of a *type-level* fact does. The gate is asking "does this text look like a column compared to a legacy id" when the real question is "does this expression resolve to `tasks.column`". **The structural fix is to resolve the reference through the TypeScript type-checker**, which turns all five spelling holes into one answer and cannot be evaded by punctuation.

I am not doing that inside this PR — it is a materially different instrument and this one is now green with 26 cases pinning the shapes we know. But I would rather that decision be made deliberately than arrived at by a seventh round.

PR #2842, opened 2026-07-30
batch-dashboard-src: the planner metrics tool froze active runtime on a renamed execution lane (186 → 185)

gsxdsm:
`packages/cli` and the plugin packages are at **zero** lifecycle guards, so this picks up the nearest unowned work: the `packages/dashboard/src/` remainder.

## The defect

`activeRuntimeMs` adds the wall-clock since `executionStartedAt` only while the card is accruing work — the **WIP role** — but it was keyed on the literal `in-progress`. On a board whose execution lane is renamed, that live tail was dropped, so `fn_task_planner_get_task_metrics` reported active time frozen at whatever the last completed segment left in `cumulativeActiveMs`. The number stayed plausible, which is why nothing surfaced it.

## The part worth reading: the wiring had no watcher, from either direction

I wired the producer (`chat.ts` resolves the task's own lanes via `wipColumnsForTask`) in the same commit, then checked whether that wiring was actually covered. It was not:

- **Deleting the `wipColumns:` argument left the entire 3830-test dashboard suite green.** The formatter's own tests inject the set by hand, so they prove the *guard* and are structurally blind to whether production fills it.
- **`check-inert-flag-seams.mjs` does not see it either.** It tracks trailing optional **parameters**; this is a property inside an options bag. That is a real gap in the checker — every seam expressed as an options-bag property is currently unguarded. Reported here rather than fixed, because #2822 and #2830 both already modify that script and a third change would guarantee a three-way conflict.

So `createTaskPlannerMetricsTool` is exported and a second test drives it, letting it do its **own** resolution against a renamed board. Deleting the argument now fails 1 of 2.

## Census

| | before | after |
|---|---|---|
| COLUMN guards | 186 | **185** |
| `packages/dashboard/src/task-planner-chat-metrics.ts` | 1 | **0** |

## Two findings I did NOT act on, deliberately

**1. `github-tracking-state.ts` keeps 2 counted guards and should.** They are the documented degraded-mode arms of a fully-resolved classifier (`completeLanes === undefined ? columnId === "done" : ...`). Marking them `DELIBERATE-LITERAL` would drop the count by **reclassification rather than conversion** — the exact move the census's own strict-check warns about. Related: the census reports `0 are trait-fallback branches (already converted)`, yet these are precisely that shape, so the trait-fallback classifier appears not to recognise a ternary whose fallback arm is the literal. Worth a look by whoever owns the census.

**2. Three pre-existing failures in `packages/dashboard/src/__tests__`, unrelated to this change** — measured identically on `origin/main` before and after:
- `planning-browser-e2e.test.ts:353`
- `register-model-routes-kimi-k3-supplemental.test.ts:60`
- `routes-tasks-near-duplicate.test.ts:274`

Flagging rather than touching them; per the standing rule they are quarantine candidates, not appeasement candidates.

## Verification

Dashboard `tsc` clean, `pnpm lint` clean, census `--strict` 0, `check-inert-flag-seams` 21/21 supplied, changeset lint clean. Targeted suites: `task-planner-chat-metrics.test.ts` 8/8, `task-planner-metrics-tool-wip-lanes.test.ts` 2/2.

### DIFF
--- a/.changeset/awaiting-planning-hold-lane.md
+++ b/.changeset/awaiting-planning-hold-lane.md
@@ -1,7 +1,0 @@
----
-"@runfusion/fusion": patch
----
-
-summary: Fix the "Queued to plan" badge never appearing on boards with a renamed waiting column.
-category: fix
-dev: The `GET /tasks` awaitingPlanning enrichment filtered on the literal `todo`. It now resolves the `hold` role once per board via `resolveProjectColumnsForRoles`, so a renamed waiting lane is enriched instead of silently falling back to TaskCard's step-count heuristic. One `listWorkflowDefinitions()` read per board load, not one workflow read per row — which is why the earlier per-task attempt was reverted.
--- a/.changeset/awaiting-planning-hold-lane.md
+++ b/.changeset/awaiting-planning-hold-lane.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Fix the "Queued to plan" badge never appearing on boards with a renamed waiting column.
+category: fix
+dev: The `GET /tasks` awaitingPlanning enrichment filtered on the literal `todo`. It now resolves the `hold` role once per board via `resolveProjectColumnsForRoles`, so a renamed waiting lane is enriched instead of silently falling back to TaskCard's step-count heuristic. One `listWorkflowDefinitions()` read per board load, not one workflow read per row — which is why the earlier per-task attempt was reverted.
--- a/.changeset/planner-metrics-wip-lane.md
+++ b/.changeset/planner-metrics-wip-lane.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Fix frozen active-runtime metrics for tasks on boards with a renamed execution column.
+category: fix
+dev: `formatTaskPlannerChatMetrics` gained a `wipColumns` option and `chat.ts`'s `fn_task_planner_get_task_metrics` tool resolves it from the task's own workflow via `wipColumnsForTask`. Previously `activeRuntimeMs` added the live tail since `executionStartedAt` only when `task.column === "in-progress"`, so on a renamed board it reported whatever `cumulativeActiveMs` held from the last completed segment. `createTaskPlannerMetricsTool` is exported so the resolver side of the seam is testable.
--- a/packages/core/src/__tests__/reopen-semantics-by-role.test.ts
+++ b/packages/core/src/__tests__/reopen-semantics-by-role.test.ts
@@ -38,6 +38,7 @@ import {
   registerDefaultWorkflowHooks,
   type DefaultWorkflowMoveContext,
 } from "../default-workflow-hooks.js";
+import { getTotalAgentActiveMs } from "../task-timing.js";
 import { resolveLifecycleColumns } from "../workflow-lifecycle-traits.js";
 import type { WorkflowIr } from "../workflow-ir-types.js";
 import type { Task } from "../types.js";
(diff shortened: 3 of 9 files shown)


gsxdsm:
**Confirmed, reproduced, and pinned — but deliberately not fixed here.** 9e398f19ed.

`applyTimingEffects` (`packages/core/src/default-workflow-hooks.ts`) banks the segment on WIP exit and never clears `executionStartedAt`; its re-entry arm is `if (!task.executionStartedAt)`, so the original start survives the round trip. Every live-tail reader — `getTotalAgentActiveMs`, this PR's metrics tool, the dashboard duration displays — then computes `now - originalStart`, which re-adds the banked segment **plus all the non-WIP time in between**.

**Measured, on the DEFAULT lineage where every id is legacy.** A card banks 5 minutes, sits 10 minutes in review, then works 5 more:

| | value |
|---|---|
| actual active time | 10 min |
| `getTotalAgentActiveMs` reports | **25 min** |

The extra 15 is the banked segment counted a second time plus the 10 minutes spent in review. **This is not a renamed-board bug** — it predates the vocabulary program and fires on the built-in board.

**Why it is pinned rather than fixed in this PR.** Correcting it moves numbers in `productivity-analytics.ts` and every duration display. That is a behaviour change owed its own review, not a line in a batch whose stated scope is converting vocabulary — the same standing rule that kept #2718's unconditional-id inversion out of a rebase.

**The fix, recorded at the test so it is not lost:** clear `executionStartedAt` in the exit arm right after banking the segment. The re-entry arm already re-stamps it from `columnMovedAt`, which is exactly the definition the field's own doc-comment gives. When that lands, the pinned expectation flips from 25 to 10 and the note goes with it.

Core `tsc` clean, `pnpm lint` clean, gate green (161/487/13/71), `reopen-semantics-by-role.test.ts` 19/19.

gsxdsm:
**Second conversion on this branch: the "Queued to plan" badge never appeared on a renamed waiting lane.** 9f744b55d7. Census **185 → 184**.

The `GET /tasks` `awaitingPlanning` enrichment filtered on the literal `todo`, so on a board whose waiting lane is called anything else **no card was enriched** and every one fell back to TaskCard's step-count heuristic — reintroducing precisely the badge-vs-engine disagreement the enrichment was added to remove. It failed quietly: the board still rendered, with badges that were merely wrong.

**The site carried a note saying it had been converted and REVERTED.** Worth reading, because the revert was right and the conclusion drawn from it was too narrow: a per-task workflow read was tried and backed out because this is the board-load path whose own comment warns about turning a load into "thousands of reads" — and it did those reads for *every* task before the enrich limit applied. The note asked for the hold column resolved **per workflow** rather than per task.

`resolveProjectColumnsForRoles` is exactly that: one `listWorkflowDefinitions()` read for the whole board. **The objection was to the arity, not the ambition**, and the seam that answers it now exists.

`hold` is the role — every builtin puts `{ trait: "hold", config: { release: "capacity" } }` on its waiting lane (`todo` in the coding workflows, displayed as "Planning" in coding-ideas), which is the same cap the badge pair names.

**Why the project union is correct here specifically**, since it is not correct everywhere: over-inclusion costs at most a few extra rows entering a best-effort enrichment already bounded by `AWAITING_PLANNING_ENRICH_LIMIT`, and `isTaskAwaitingPlanning` still answers per task — so a widened candidate set cannot produce a *wrong* badge, only an extra file read. Under-inclusion is the bug being fixed.

**Mutation-verified in both directions**, because one direction alone would not have been enough:

| mutation | result |
|---|---|
| restore the literal `todo` | 1 failed / 10 passed |
| widen to every row | 2 failed / 9 passed |
| restored | 11 passed |

Three cases added: renamed lane enriched, non-hold lane on that same renamed board *not* enriched, and the legacy `todo` still enriched (the union keeps a mid-rename board working).

Dashboard `tsc` clean, `pnpm lint` clean, census `--strict` 0, changeset lint clean.

gsxdsm:
Rebased onto `origin/main`. **Main landed the same awaitingPlanning conversion independently while this branch was open** — same helper (`resolveProjectColumnsForRoles`), same `hold` role, same one-read-per-board argument. I took main's implementation and dropped my duplicate changeset (`board-hold-lane-enrichment.md` already covers it; two changesets for one change means two release-note lines for one operator-visible fix).

**What survives from that commit, and why it is worth keeping:** the three test cases. They were written against my implementation and pass unchanged against main's — renamed hold lane enriched, non-hold lane on that same renamed board *not* enriched, and the legacy `todo` still enriched so a mid-rename board keeps its badges. A ratchet on someone else's conversion is exactly what they should be, and the negative cases are the ones main's version did not bring: I verified earlier that widening the filter to every row fails 2 of them.

So this PR is now: the planner-metrics WIP-lane conversion (with its producer-side test), the pinned WIP round-trip double-count defect, and three cases guarding a conversion main made. The census claim in the PR body is unchanged at 186 → 185 for the metrics fix; the awaitingPlanning drop is main's, not mine, and I am not counting it twice.

12/12 on `register-task-workflow-routes.awaiting-planning.test.ts`, changeset lint clean.

PR #2843, opened 2026-07-30
fix(create): two task-CREATE destinations named a lane the board does not have

gsxdsm:
Both files sat at **census-zero** and both wrote real cards into columns no workflow declares. The census scores `===` comparisons, so a lane literal passed as a **call argument** is invisible to it — one of the four census-blind classes. These are the only two explicit-`column` creates in production:

```
packages/dashboard/src/routes/register-gitlab.ts:108   column: "triage"
packages/cli/src/extension.ts:5243                     column: "todo"
```

## The two defects

**`register-gitlab.ts` — `column: "triage"`, a column U11 DELETED.** This one is broken on *every* board, not only renamed ones: the default lineage is now `todo | in-progress | in-review | done | archived`. `createTask` already resolves the intake column of the workflow it selects (`resolvedEntryColumn`), and an explicit `column` **overrides** that resolution — which is why the stale literal survived U11. Nothing rejects the write and nothing logs it: the route answers `201` with a task id and the imported card is simply not on the board. Same shape as the `task-update.ts` triage defect fixed earlier in this program.

Fix: omit `column` and let `createTask` resolve intake.

**`extension.ts` `fn_delegate_task` — `column: "todo"`.** On a workflow whose ready lane is named anything else, the delegated card goes to an undeclared column: written, reported to the caller as delegated, never visible to the agent it was delegated to.

Fix: resolve the selected workflow's `hold` lane. **Deliberately not** "omit the column like the GitLab route" — the tool's own contract is *"the task goes to the ready-to-work lane and the target agent picks it up on its next heartbeat"*, so inheriting intake resolution would park a delegated card in a manual-intake lane waiting for a human. That would be a behaviour change; `hold` is the role that names the lane the literal meant.

## New helper: `resolveWorkflowColumnForRole(store, role, workflowId?)`

The **write**-shaped counterpart to `resolveProjectColumnsForRoles`. The read helper unions in the legacy ids because an extra id in a query set is inert; here the same trick is a silent wrong write (post-U12 an undeclared column is a `TransitionRejectionError` on move, a phantom lane on create), so it returns one column from one workflow, or `undefined`.

**A contract I got wrong twice, now pinned by a test.** `undefined` means *"this workflow declares no such column"* and nothing else. `resolveWorkflowIrById` never throws and never returns nothing — an unregistered builtin id, a missing definition row and a failing read all resolve to the default coding IR (branded via `markFellBack`). So an unreadable workflow yields the **built-in** hold lane, not `undefined`, and both call sites' `?? "todo"` fallbacks are narrower than they look. Two of my first test cases asserted the opposite and failed; the behaviour is the resolver's, and the write it produces is identical to the caller's own legacy fallback either way.

## Revert proofs (measured, not asserted)

| revert | failure |
|---|---|
| `column: holdColumn` -> `column: "todo"` | `extension.test.ts`: `expected 'todo' to be 'queued'` |
| omitted column -> `column: "triage"` | `routes-gitlab.test.ts`: `expected 'triage' to be undefined` |

The two neighbouring `fn_delegate_task` cases stay green under the first revert, because the built-in board and the test's `linearWorkflowIr` both call the lane `todo` — which is exactly why this literal survived every previous pass.

The GitLab case asserts **absence** of the key rather than a resolved id: the store there is a fake whose `createTask` echoes its input, so asserting a resolved value would be testing the fake. Absence is the property that hands the decision to the real `createTask`.

## Census

| file | before | after |
|---|---|---|
| `packages/dashboard/src/routes/register-gitlab.ts` | 1 | 0 |
| `packages/cli/src/extension.ts` | 1 | 0 |

Baseline tightened. It also picks up `packages/core/src/task-store/moves.ts` 2 -> 0, which was **already true on main** — not from this diff.

## Noted, deliberately not changed

- `validateAssignableAgentId`'s synthetic probe a few lines above still uses `{ id: "<new>", column: "todo" }`. It feeds `isImplementationTask`, whose `IMPLEMENTATION_TASK_COLUMNS` set an earlier worker documented as deliberately-not-converted (converting it makes the routing policy async — an agent-admission behaviour change). On a renamed board the probe is now *stricter* than the real destination, which is the safe direction and matches the pre-existing behaviour.
- The third site from this bucket, `workflow-node-handlers.ts:455` (`transitionTask({ column: "in-review" })` on the `review-handoff` seam), is a **hard** failure rather than a silent one — `transitionTask` routes through `moveTask`, which post-U12 throws `TransitionRejectionError` for an undeclared destination, so the workflow walk dies at the handoff on any renamed review lane. It is engine (`batch-engine`) and fixing it properly touches `executor.ts`, which #2820 is also editing. Left for that batch rather than opened as a conflicting edit.

## Verification

- `pnpm lint` — clean
- `tsc --noEmit -p tsconfig.json` for `@fusion/core`, `@runfusion/fusion`, `@fusion/dashboard` — clean
- `node scripts/lifecycle-column-census.mjs --strict` — exit 0
- targeted: `project-lane-vocabulary.test.ts` 14/14, `routes-gitlab.test.ts` 8/8, `extension.test.ts -t fn_delegate_task` 9/9

## Summary by CodeRabbit

* **Bug Fixes**
  * GitLab-imported cards now appear in the workflow’s configured intake lane.
  * Delegated tasks now move to the workflow’s configured hold lane, including workflows with custom lane names or separate intake and hold lanes.
  * Delegation reports the task’s final lane and provides an error when it cannot be moved successfully.

### DIFF
--- a/.changeset/create-destination-lanes.md
+++ b/.changeset/create-destination-lanes.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: GitLab imports and agent delegation now land cards in real board lanes instead of vanishing.
+category: fix
+dev: GitLab import passed `column: "triage"`, a column U11 deleted, so imported cards were written into a lane no workflow declares; it now omits `column` and lets `createTask` resolve the workflow's intake lane. `fn_delegate_task` passed the literal `"todo"` and now resolves the selected workflow's `hold` column via the new `resolveWorkflowColumnForRole(store, role, workflowId?)` in `@fusion/core`.
--- a/.changeset/create-destination-lanes.md
+++ b/.changeset/create-destination-lanes.md
@@ -4,4 +4,4 @@
 
 summary: GitLab imports and agent delegation now land cards in real board lanes instead of vanishing.
 category: fix
-dev: GitLab import passed `column: "triage"`, a column U11 deleted, so imported cards were written into a lane no workflow declares; it now omits `column` and lets `createTask` resolve the workflow's intake lane. `fn_delegate_task` passed the literal `"todo"` and now resolves the selected workflow's `hold` column via the new `resolveWorkflowColumnForRole(store, role, workflowId?)` in `@fusion/core`.
+dev: GitLab import passed `column: "triage"`, a column U11 deleted, so imported cards were written into a lane no workflow declares; it now omits `column` and lets `createTask` resolve the workflow's intake lane. `fn_delegate_task` passed the literal `"todo"` and now resolves the created task's own `hold` lane via `resolveTaskLifecycleColumns`, moving the card off intake when the workflow separates the two roles.
--- a/packages/cli/src/__tests__/extension.test.ts
+++ b/packages/cli/src/__tests__/extension.test.ts
@@ -4356,6 +4356,79 @@ pgTest("fn pi extension (runnable structured-output regression slice)", () => {
       expect(task.column).toBe("queued");
     });
 
+    /*
+    FNXC:WorkflowLifecycleColumns 2026-07-31-00:30 (#2843 review — greptile P1, "failed move reports
+    successful delegation"):
+
+    A FAILED LANDING MUST NOT LEAD WITH A SUCCESS CLAIM.
+
+    The first version kept "will be picked up by X on their next heartbeat cycle" and appended a
+    ` WARNING: ...`. Assigned-agent selection skips cards outside the workflow's hold lane, so a card
+    stranded on intake is never dispatched — and the caller, usually another agent, reads the first
+    sentence and moves on. "Success with a caveat" is how a delegation silently goes nowhere.
+
+    WHAT THIS PROVES AND WHAT IT DOES NOT, stated because the distinction matters. It pins the
+    HANDLING — the branch returns `isError`, never claims pickup, and still surfaces the task id so a
+    caller can finish by hand. It does NOT prove the failure is reachable in production, and I could
+    not make it so: `holdColumn` comes from the task's OWN resolved IR, so `moveTask`'s declared-column
+    check passes by construction. I tried a workflow declaring a `hold` column with no node on it,
+    expecting rejection; the move SUCCEEDED and the card landed there — `moveTask` accepts any column
+    the workflow declares, and node reachability does not gate it.
+
+    So the store method is stubbed for exactly one call. Stubbing is normally how a test proves only
+    that a catch block runs, which is why it is worth being explicit: the catch is DEFENSIVE, and what
+    changed — and what regresses silently — is the shape of the message it produces.
+    */
+    it("reports a failed landing as an ERROR and never claims the agent will pick it up", async () => {
+      const agentId = await seedAgent(tmpDir, { name: "delegate-landing-fails" });
+      const store = h.store();
+      const split = await store.createWorkflowDefinition({
+        name: "Split lanes (landing fails)",
+        ir: {
+          version: "v2",
+          name: "Split lanes (landing fails)",
+          columns: [
+            { id: "ideas", name: "Ideas", traits: [{ trait: "intake" }] },
+            { id: "queued", name: "Queued", traits: [{ trait: "hold" }] },
+          ],
+          nodes: [
+            { id: "start", kind: "start", column: "ideas" },
+            { id: "end", kind: "end", column: "queued" },
+          ],
+          edges: [{ from: "start", to: "end", condition: "success" }],
+        } as unknown as WorkflowIr,
+      });
+
+      const moveTask = vi.spyOn(store, "moveTask").mockRejectedValueOnce(
+        new Error("unknown-column: queued"),
+      );
+      let result: Awaited<ReturnType<typeof tool.execute>>;
+      const tool = api.tools.get("fn_delegate_task")!;
+      try {
+        result = await tool.execute(
+          "dt-landing-fails",
+          { agent_id: agentId, description: "Landing will fail", workflow_id: split.id },
+          undefined,
+          undefined,
+          makeCtx(tmpDir),
+        );
+      } finally {
+        moveTask.mockRestore();
+      }
+
+      expect(result.isError).toBe(true);
+      /* The claim that must never survive a failed landing. */
+      expect(result.content[0].text).not.toContain("will be picked up");
+      expect(result.content[0].text).toContain("stranded on intake");
+      /* The card exists either way, so the caller keeps what it needs to finish the job by hand. */
+      expect(result.details.taskId).toBeTruthy();
+      expect(result.details.error).toContain("unknown-column");
+
+      /* And it really is still on intake — the message is not merely pessimistic. */
+      const { task } = await readTaskWorkflowState(tmpDir, result.details.taskId);
+      expect(task.column).toBe("ideas");
+    });
+
     it("rejects unknown agent", async () => {
       const tool = api.tools.get("fn_delegate_task")!;
       const result = await tool.execute(
--- a/packages/cli/src/__tests__/extension.test.ts
+++ b/packages/cli/src/__tests__/extension.test.ts
@@ -4357,7 +4357,7 @@ pgTest("fn pi extension (runnable structured-output regression slice)", () => {
     });
 
     /*
-    FNXC:WorkflowLifecycleColumns 2026-07-31-00:30 (#2843 review — greptile P1, "failed move reports
+    FNXC:WorkflowLifecycleColumns 2026-07-30-15:02 (#2843 review — greptile P1, "failed move reports
     successful delegation"):
 
     A FAILED LANDING MUST NOT LEAD WITH A SUCCESS CLAIM.
--- a/packages/cli/src/__tests__/extension.test.ts
+++ b/packages/cli/src/__tests__/extension.test.ts
@@ -4531,6 +4531,75 @@ pgTest("fn pi extension (runnable structured-output regression slice)", () => {
       expect(result.details.agentId).toBe(agentId);
     });
 
+    /*
+    FNXC:WorkflowLifecycleColumns 2026-07-30-17:45 (#2843 review — greptile P1, third round):
+    THE INVARIANT: a SUBSTITUTED workflow is not the board's answer, even when its lane exists.
+
+    `resolveWorkflowIrForTask` never throws — it substitutes the default coding IR — so a failed
+    resolution surfaced as the BUILT-IN lanes, `hold: "todo"`. I argued that was harmless because
+    `todo` would be undeclared on such a board and the move would be rejected. Right about the
+    mechanism, wrong about the population: a workflow that holds work in `queued` and ALSO declares a
+    `todo` column for something else gets a move that SUCCEEDS into a lane nothing dispatches from,
+    and a success message with it.
+
+    The fixture is exactly that shape — `ideas` (intake), `queued` (hold), and a plain `todo` that
+    carries no lifecycle trait — which is why the older cases could not have caught this: theirs
+    declared no `todo` at all, so the wrong move was rejected for the wrong reason.
+
+    The stub targets `getTaskWorkflowSelectionAsync`, the READER, not the writer `createTask` uses,
+    so creation is untouched and only the post-create resolution sees a workflow id that resolves to
+    nothing. That is what makes this the first case to reach the "could not be resolved" arm rather
+    than defending it in a comment.
+
+    REVERT PROOF, measured: resolve through `resolveTaskLifecycleColumns` again (no provenance) and
+    this fails — the tool moves the card to `todo` and reports a successful delegation.
+    */
+    it("does not trust a SUBSTITUTED workflow's lanes even when the board declares that column", async () => {
+      const agentId = await seedAgent(tmpDir, { name: "delegate-substituted-workflow" });
+      const store = h.store();
+      const declaresTodo = await store.createWorkflowDefinition({
+        name: "Queued hold, unrelated todo",
+        ir: {
+          version: "v2",
+          name: "Queued hold, unrelated todo",
+          columns: [
+            { id: "ideas", name: "Ideas", traits: [{ trait: "intake" }] },
+            { id: "queued", name: "Queued", traits: [{ trait: "hold" }] },
+            /* Declared, carries no lifecycle role — the column the built-in fallback would name. */
+            { id: "todo", name: "Someday", traits: [] },
+          ],
+          nodes: [
+            { id: "start", kind: "start", column: "ideas" },
+            { id: "end", kind: "end", column: "queued" },
+          ],
+          edges: [{ from: "start", to: "end", condition: "success" }],
+        } as unknown as WorkflowIr,
+      });
+
+      /* Reader only: `createTask` writes the selection, it does not read it through this method. */
+      const selection = vi.spyOn(store, "getTaskWorkflowSelectionAsync")
+        .mockResolvedValue({ workflowId: "wf-vanished", stepIds: [] });
+      const tool = api.tools.get("fn_delegate_task")!;
+      let result: Awaited<ReturnType<typeof tool.execute>>;
+      try {
+        result = await tool.execute(
+          "dt-substituted",
+          { agent_id: agentId, description: "Workflow resolves to a substitute", workflow_id: declaresTodo.id },
+          undefined,
+          undefined,
+          makeCtx(tmpDir),
+        );
+      } finally {
+        selection.mockRestore();
+      }
+
+      expect(result.isError).toBe(true);
+      expect(result.content[0].text).not.toContain("will be picked up");
+      /* The card must NOT have been moved into the built-in fallback's lane. */
+      const { task } = await readTaskWorkflowState(tmpDir, result.details.taskId);
+      expect(task.column).not.toBe("todo");
+    });
+
     it("rejects unknown agent", async () => {
       const tool = api.tools.get("fn_delegate_task")!;
       const result = await tool.execute(
--- a/packages/cli/src/__tests__/extension.test.ts
+++ b/packages/cli/src/__tests__/extension.test.ts
@@ -4301,6 +4301,61 @@ pgTest("fn pi extension (runnable structured-output regression slice)", () => {
       expect(task.assignedAgentId).toBe(agentId);
     });
 
+    /*
+    FNXC:WorkflowLifecycleColumns 2026-07-30-23:55:
+    THE INVARIANT: when intake and hold are DIFFERENT lanes, delegation ends on hold.
+
+    The case above cannot see this: its workflow carries both traits on one column, so the card
+    arrives on the right lane through `createTask`'s entry resolution and the move never runs. Here
+    the two roles are separate columns, so the card lands on `ideas` and must be moved to `queued` —
+    which is the behaviour the old `column: "todo"` literal was providing (skip intake, go straight
+    to the lane the agent picks up from) and the part that would silently regress if the move were
+    dropped as redundant.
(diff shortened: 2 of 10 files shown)


gsxdsm:
Greptile P1 (default workflow selection race) — **agreed, fixed in 336f419**, and the diagnosis is better than "race": it was two reads answering the same question.

My version resolved the hold column from `getDefaultWorkflowId()` *before* the create, and `createTask` then resolves the project default **again** to select the workflow (`materializeDefaultWorkflowSteps`). Two independent authorities for one fact — so a default that changes between them yields a column from workflow A written onto a card selected into workflow B. That is precisely the undeclared-lane write this PR exists to remove, reintroduced by the fix for it.

**The fix removes the second authority rather than narrowing the window.** The card is now created with **no explicit column**, so `createTask` alone decides the workflow and the entry lane; the hold lane is then resolved *from the created task* via `resolveTaskLifecycleColumns(store, task.id)`. There is nothing left to disagree.

Properties worth stating:

- **The common path is unchanged and costs no extra write.** On the built-in board entry and hold are both `todo`, so `holdColumn === task.column` and the move is skipped entirely.
- **A failed move is reported, not swallowed.** The card exists either way; telling the caller it is ready while it sits on intake is the failure mode this PR is about. The tool output carries an explicit `WARNING:` with the real column, and `details.column` now reports where the card actually landed.

I also considered pinning the workflow (resolve the default once, pass it as `workflowId`) and rejected it: `materializeDefaultWorkflowSteps` returns `undefined` for a fragment or missing default and lets creation fall through to the `defaultOn` step path with **no** persisted selection, whereas `materializeExplicitWorkflowSteps(fragmentId)` throws before the row is created. Pinning would have turned a degraded default into a hard create failure.

**`resolveWorkflowColumnForRole` is deleted** along with its 5 tests. Once the resolution moved after the create, the per-task resolver was the right tool and that helper had zero production callers — the exact unwired-lane-parameter shape `scripts/lib/unwired-lane-parameter.mjs` was written to catch.

**New coverage the old case could not provide:** a workflow with `intake` and `hold` on *different* columns. The previous renamed-lane test put both traits on one column, so the card arrived on the right lane through entry resolution and the move never executed — the move path was untested. Revert proof, measured: disable the move and the new case fails `expected 'ideas' to be 'queued'` while the sibling stays green.

Re-verified: `pnpm test:gate` 161/487/13/71, `pnpm lint` clean, `tsc --noEmit` clean for all three packages, census `--strict` exit 0, `fn_delegate_task` 10/10.

gsxdsm:
Greptile P1 (failed move reports successful delegation) — **agreed, fixed in ef56cfa.**

The warning-suffix version still **led** with "The task will be picked up by ${agent} on their next heartbeat cycle" and left `isError` unset. Assigned-agent selection skips cards outside the workflow's hold lane, so a card stranded on intake is never dispatched — and the caller here is usually another agent, which reads the first sentence and moves on. Reported-success-with-a-caveat is the failure mode, not the mitigation.

The branch now returns `isError: true`, names the real column and the manual move needed, and says explicitly that the card will **not** be picked up. `details.taskId` is preserved because the card does exist and the caller needs it to finish by hand.

**One finding worth recording, because it changes how the branch should be read.** I tried to write a behavioural test and could not construct a real failure. The fixture was a workflow declaring `parked` with the `hold` trait and **no node on it**, expecting `moveTask` to reject on adjacency — it **succeeded**, and the card landed in `parked`. `moveTask` accepts any column the workflow *declares*; node reachability does not gate it. Since `holdColumn` here is read from the task's own IR, `workflowHasColumn` passes by construction.

So this catch is **defensive, not a live path**. I left it untested and said so in the code comment rather than stubbing `moveTask` to throw — that would prove only that the catch block runs, which is not the part that needed proving. If you would rather have the stub-shaped test anyway, say so and I will add it labelled as reporting-shape-only.

Re-verified: `pnpm test:gate` 161/487/13/71, `pnpm lint` clean, `tsc --noEmit` clean, `fn_delegate_task` 10/10.

gsxdsm:
**The finding was right and the fix landed in `ef56cfaf8a`; `e563a92a21` now covers it.**

The branch returns `isError: true` and leads with the failure — no "will be picked up by X on their next heartbeat cycle" — while keeping the task id in `details` so a caller can finish by hand. The reviewer's reasoning is the point: assigned-agent selection skips cards outside the workflow's hold lane, so a card stranded on intake is **never dispatched**, and a caller (usually another agent) reads the first sentence and moves on. "Success with a caveat" is how a delegation silently goes nowhere.

**The test is new, because what changed is a *message*, and a message regresses silently.** It pins: `isError`, no pickup claim, the task id still present, and — verified against the store — the card really is still on intake, so the wording is not merely pessimistic. Mutation-verified: making the error branch fall through to the success text fails it.

**What it does not prove, stated rather than papered over.** It does not show the failure is reachable in production, and I could not make it so. `holdColumn` comes from the task's **own** resolved IR, so `moveTask`'s declared-column check passes by construction. I tried a workflow declaring a `hold` column with no node on it, expecting a rejected transition — the move **succeeded** and the card landed there: `moveTask` accepts any column the workflow *declares*, and node reachability does not gate it. That is worth knowing independently of this PR.

So `store.moveTask` is stubbed for exactly one call. Stubbing is normally how a test proves only that a catch block runs, which is why I am being explicit: **the catch is defensive; the message it produces is the subject under test.**

CLI `tsc` clean, `pnpm lint` clean, `extension.test.ts` 70 passed / 96 skipped (the skips are the suite's usual env-gated blocks).

gsxdsm:
Three threads addressed in b6c0685.

**CodeRabbit (minor) — tool copy still promised literal `todo`.** Fixed. The description and prompt guidance now name the role and give `todo` as the built-in board's instance, matching what the implementation does.

**CodeRabbit (major) + Greptile P1 — unresolvable hold lane reports success.** The diagnosis is right about the shape: reading `?.hold` collapsed two genuinely different answers into `undefined` — *"this workflow declares no hold lane"* (entry is the only lane it has, so success is honest) and *"this workflow could not be read at all"* (nothing can be verified). It now reads at both levels and treats only the second as a failed landing.

**But I have to report that the second arm is unreachable, rather than let the fix imply coverage it does not have.** `resolveWorkflowIrForTask` never throws — an unreadable selection, a missing definition, a malformed one, and a throwing lookup *all* degrade to the default coding IR (branded via `markFellBack`). So `resolveTaskLifecycleColumns` cannot return `undefined` through this path.

What actually happens in the reported scenario is that an unreadable workflow yields the **built-in** lanes, `hold: "todo"`. On a split-lane workflow that is a column the card's own workflow does not declare, so `moveTask` rejects it and the **error branch fires anyway** — the caller gets exactly the failure the review asked for, by the other route. The new arm is kept because it costs nothing and is correct the day that resolver stops degrading, not because it runs, and the comment says so in those words.

I wrote a test for it first. Stubbing `getWorkflowDefinition` to reject also breaks `createTask` (it reads the same definition to materialise the explicit workflow), so the call never produced a task and the assertion failed on `details.taskId`. I deleted the test rather than weaken it into something that passes — a test that has to be softened until it goes green is not evidence.

**CodeRabbit (minor) — FNXC timestamp is tomorrow.** Correct; mine said `2026-07-31`, today is 2026-07-30. Fixed on this branch, and I also made the three timestamps in this file monotonic. Scoped deliberately: `grep` shows the `2026-07-31-` prefix across ~20 files from other workers, and a repo-wide sweep is not this PR's business.

Also noting for the record: the message-shape test in e563a92 is not mine — a peer added the stubbed coverage I had declined, and labelled it honestly as pinning the handling rather than the reachability. Left as-is.

Re-verified: `pnpm test:gate` 161/487/13/71, `pnpm lint` clean, `tsc --noEmit` clean, `fn_delegate_task` 11/11.

gsxdsm:
All four findings addressed; rebased and pushed as `a1dae571fe`. **Someone was fixing the same threads on this branch concurrently** — I merged rather than overwrote, taking their wording for the tool copy, the timestamp, and the error text, and keeping my structural change where it covers more.

**Tool copy** — already corrected on the branch tip; the description and guidance now name the workflow's ready lane with `todo` as the built-in example.

**`?.hold` collapsed two states** — `resolveTaskLifecycleColumns` returns `undefined` when resolution *threw*, and a struct whose `hold` is `undefined` when the board resolved and declares no hold lane. Both now handled separately.

**"No hold lane" is still a false claim** — greptile's fourth finding. I first called it a benign no-op: nowhere to move the card, so intake is correct. Both halves true, conclusion wrong — assigned-agent selection picks cards *out of* the hold lane, so on a board with none, nothing ever dispatches it.

**And that fix was wrong until the existing suite caught it.** Gating on `!holdColumn` alone failed every **untraited** board: `linearWorkflowIr`, and every v1 graph upgraded by `synthesizeDefaultColumns`, declares columns with `traits: []`, so `hold` is `undefined` while `todo` plainly exists and is exactly where agents pick work up. It would have broken every untraited board to fix a case none of them have. So the branch is gated on `declaresAnyLifecycleTrait` — **three states: unreadable, untraited, traits-expressed-but-no-hold, and only the last is the board answering.** That is the same distinction that bit the merge-queue conversion in #2819; it is now the third place today.

This is the part my merge keeps over the concurrent version, which treats *any* missing `hold` as success and so leaves greptile's finding open.

**Mutation-verified both directions:** dropping the traits gate fails the untraited case; dropping the no-hold branch fails the no-hold case.

**One arm remains unreachable and is documented as such rather than counted as covered:** `resolveWorkflowIrForTask` never throws — unreadable selection, missing definition, malformed IR and throwing lookup all degrade to the built-in IR — so the `undefined` object cannot occur through this path today. Kept because it costs nothing and is correct the day that resolver stops degrading.

**FNXC timestamp** corrected to today.

CLI `tsc` clean, `pnpm lint` clean, `extension.test.ts` 72 passed / 96 skipped.

gsxdsm:
Greptile P1 (fallback hold misroutes delegation) — **agreed, fixed in 39af6ec**, and it closes the hole my own "this arm is unreachable" argument left open.

I had written that a substituted resolution was harmless because the built-in `hold: "todo"` would be **undeclared** on such a board, so `moveTask` would reject it and the error branch would fire anyway. That is right about the mechanism and **wrong about the population it covers**. A workflow that holds work in `queued` and *also* declares a `todo` column for something else gets a move that **succeeds** — into a lane nothing dispatches from — plus a success message. My argument only ever covered boards with no `todo` at all.

**The fix uses the seam built for this.** `resolveWorkflowIrForTaskWithProvenance` reports `source: "selection"` **only** when it genuinely resolved the workflow the task selected, and `"default"` whenever it substituted. Pairing that with the task's own selection separates the two reasons for `"default"`:

- **no workflow selected** — legitimate; the built-in board *is* the answer, and success is honest;
- **selected but unresolvable** — a substitution wearing the board's authority, now reported as a failed landing.

A useful consequence: the "could not be resolved" arm is now **reachable**, so it is covered by a test instead of defended in a comment. That was the loose end from the previous round.

**The new fixture is the shape the older cases could not catch** — `ideas` (intake), `queued` (hold), and a plain `todo` carrying no lifecycle trait. The earlier split-lane fixtures declared no `todo` at all, so the wrong move was rejected for the wrong reason and the tests passed for a reason that would not generalise.

The stub targets `getTaskWorkflowSelectionAsync` — the **reader** — not the writer `createTask` uses, so creation is untouched and only the post-create resolution sees a workflow id that resolves to nothing. That is what made this testable at all; my earlier attempt stubbed `getWorkflowDefinition` and broke `createTask` along with it.

Revert proof, measured: resolve through `resolveTaskLifecycleColumns` again and the new case fails on `isError` — the card is moved to `todo` and reported as delegated.

Re-verified: `pnpm test:gate` 161/487/13/71, `pnpm lint` clean, `tsc --noEmit` clean, `fn_delegate_task` 14/14.

gsxdsm:
**Confirmed, and already fixed on the branch tip by a concurrent commit — my version is fully subsumed, so I dropped it.**

We reached the same design independently: `resolveWorkflowIrForTaskWithProvenance` reports `source: "selection"` only when it genuinely resolved the workflow the task selected, and pairing that with the recorded selection separates the two reasons for `"default"` — *no workflow selected* (legitimate; the built-in board **is** the answer) from *selected-but-unresolvable* (a substitution wearing the board's authority). Their message also names the workflow id, which is better than mine.

The finding itself is worth restating because it corrected an argument I had made on this PR: I said the built-in `todo` fallback was harmless because `todo` would be undeclared on a custom board and `moveTask` would reject the move. **That is true only of a board that does not declare `todo` at all.** A workflow holding work in `queued` that *also* declares `todo` for something else gets a move that succeeds into a lane nothing dispatches from — right about the mechanism, wrong about the population it covers.

**One thing I did that is worth keeping in mind for whoever reviews the coverage.** I wrote a test for this branch and deleted it: making the definition read reject breaks `createTask` **first**, so the tool returns its outer error before the landing logic runs. The case never reached the branch it claimed to cover, and the mutation proved it — removing the guard left the test green. Reaching that branch needs the definition to load during creation and fail during resolution, which no seam here exposes. The concurrent commit claims this arm is now reachable and covered; if that coverage was verified the same way, it is worth re-checking against the same mutation before trusting it.

Branch tip: CLI `tsc` clean, `pnpm lint` clean, `extension.test.ts` 73 passed / 96 skipped.

gsxdsm:
Greptile P1, fourth round (workflow snapshot race) — **agreed, fixed in f45b504**, and it is the *first* round's defect reintroduced three rounds later in a different place.

Round one caught me resolving the hold lane from `getDefaultWorkflowId()` while `createTask` independently resolved the same default — two reads answering one question. Round three's fix then paired the provenance read with an independent `getTaskWorkflowSelectionAsync`, to tell *"no workflow selected"* (legitimate — the built-in board **is** the answer) from *"selected but unresolvable"*. Same shape, new location: a selection written or cleared between the two reads makes `resolved.ir` describe one workflow and `selectedWorkflowId` another, and the tool then moves the card with the wrong board's hold lane or reports a resolution error that never happened.

**The second read was never needed.** The reason for a substitution is not the honest question — whether it is about to be *acted on* is:

- substituted hold lane **equals** the card's column → no move, nothing is being decided on unverified information, and this is exactly the no-selection / untraited case where the built-in vocabulary is correct. Success.
- it **differs** → acting would move a real card into a lane inferred from a workflow that is not the card's. That is the misroute round three found. Refused.

Judging by the *action* rather than the *reason* takes one read, so there is no window for two answers to disagree. The race is removed rather than narrowed.

A fifth state surfaced while doing it, from the type rather than from review: `resolveLifecycleColumns` returns `undefined` for an IR that declares no columns at all. `holdColumn` is optional-chained and the untraited branch is the honest answer there.

Revert proof, measured: disable the substitution guard and round three's case fails — the card is moved into the fallback `todo` and reported as delegated. So the guard is load-bearing, not belt-and-braces.

Re-verified: `pnpm test:gate` 161/487/13/71, `pnpm lint` clean, `tsc --noEmit` clean, `fn_delegate_task` 14/14.

Worth stating plainly after four rounds on one PR: every one of these was a real defect, three of the four were *my fix* for the previous one, and the recurring shape is **two reads answering the same question**. That is the thing to watch for in the rest of this program, not the specific call sites.

gsxdsm:
**The race is real and a concurrent commit fixed it better than I did — I dropped my version and kept one test.**

I paired the provenance read with a second `getTaskWorkflowSelectionAsync`, read the selection on **both** sides of the resolve, and refused when they disagreed. That detects the tear. The version that shipped **removes** it: it drops the second read entirely and judges the substitution by whether it would actually **move** the card —

- substituted hold lane equals the card's current column → no move, nothing decided on unverified information → success;
- it differs → acting would move a real card using a board that is not its own → refuse.

One read, no window. That is the better answer, and it is worth noting *why*: two reads answering one question was the same defect the first round of this review caught, reappearing three rounds later in a different place. Detecting a race is strictly worse than not having one.

**I removed my torn-snapshot test along with the mechanism.** It asserted "refuses when the two reads disagree", which no longer exists — it would have been green and about nothing. That is the failure mode I keep flagging on other people's changes, so it would be poor form to leave one of mine in.

**The paired negative stays, and it earned its place:** a selection read that *fails* must not fail the delegation. My first attempt got that wrong — the reads were unguarded, so a selection-store hiccup threw into the outer catch and failed the whole delegation. An existing test caught it, which is the second time today the suite has caught my fix rather than the bug.

Branch tip: CLI `tsc` clean, `pnpm lint` clean, delegate suite 15/15.

gsxdsm:
**Correct — and it is the same defect one round earlier, surviving inside the branch I added to fix a different one.** 0d6b894c25.

`taskBoardDeclaresLifecycleTraits` resolved the task's workflow **again**, so a selection changing in between combined the hold column from one board with the trait state of another. That is two-reads-of-one-fact, which is exactly what the previous round removed from the substitution check.

`resolved.ir` is already in hand and is the only snapshot anything in this block should consult, so the check is now `declaresAnyLifecycleTrait(resolved.ir)` and the helper is deleted. One read, no window.

**The recurring lesson across the last three rounds on this PR, since it has now cost four commits:** every fix that *detected* an inconsistency between two reads was worse than the fix that *removed* the second read. I reached for reconciliation twice (read-twice-and-compare, then a helper that re-resolved) before the shape landed. Detecting a race leaves the race; sharing the snapshot ends it.

Mutation-verified: dropping the traits gate still fails the untraited-board case, so the three-state distinction is intact — unreadable, untraited, traits-expressed-but-no-hold.

CLI `tsc` clean, `pnpm lint` clean, delegate suite 15/15.

gsxdsm:
Two P1s from round five. **One was already fixed by a peer** (0d6b894 — the traits check resolving the workflow a second time, which was the same two-reads defect surviving in the branch I added to fix a different one). The other is fixed in 3178950.

**"Fallback equality masks wrong hold" — agreed.** It names the board every earlier case here missed: `todo` as **intake**, `queued` as hold. `createTask` puts the card on `todo`, the degraded lookup fabricates the built-in `hold: "todo"`, the two match, and the "no move needed" rule reported a delegation for a card assigned-agent dispatch will never select. Equality with a fabricated value is not evidence — it conflated *"no action needed"* with *"action verified"*.

**Settled with `params.workflow_id`,** which is caller **input**: there is no second snapshot, so no race. That mattered here specifically because every fix in rounds two through five introduced the next round's defect by adding a read. A caller that named a workflow proves a real one exists; a substitution then proves we failed to read it, so its hold lane cannot be inferred from the built-in vocabulary.

**My first attempt at this was wrong and I reverted it rather than ship it.** I made *every* substitution non-committal — which broke two cases a peer had just landed with explicit reasoning: a degraded resolve with no `workflow_id` is not evidence of a stranded card, and their tests verify the pickup claim is *true* by re-reading the column afterwards. Overwriting a peer's tested contract because a reviewer flagged a different board is a flip-flop on a shared branch, not a fix.

The split is what reconciles them, and it is not a compromise — the two situations are genuinely different observations:

| caller | `substituted` means | verdict |
|---|---|---|
| named a workflow | a real workflow exists and we failed to read it | refuse the pickup claim |
| named none | possibly a project with no resolvable workflow at all | the card belongs where it is; claim stands |

**What this does not cover, stated rather than implied:** the same `todo`-intake board reached through the project **default** with no `workflow_id` argument. There, "substituted" and "this project has no resolvable workflow" are the same observation, and the second is a legitimate configuration — failing it trades a false success for a false alarm on a valid setup. Distinguishing them needs the read that just failed. I would rather leave that gap named than close it by guessing, since guessing is what produced rounds two through five.

Revert proof, measured: drop the `substituted && workflowId` arm and the new case fails on `isError` — a successful delegation reported for a card on intake.

Re-verified: `pnpm test:gate` 161/487/13/71, `pnpm lint` clean, `tsc --noEmit` clean, `fn_delegate_task` 16/16 (including both of the peer's degraded-resolve cases).

PR #2846, opened 2026-07-30
test(dashboard): App.test.tsx is fully green — an incomplete mock was crashing NewTaskModal

### LINKED ISSUE #2829: App.test.tsx: 10 tests red on main since 2026-07-25 (list-view '+ New Task' button absent under test)
Not a flake and not caused by any batch PR — reproduced on clean `main` checkouts. Filing rather than fixing: it is a debugging task inside `App.tsx`/`ListView.tsx`, and my two attempts at the root cause were both wrong, so a guess from me is worth less than a precise handoff.

## Reproduction

```
pnpm --filter @fusion/dashboard exec vitest run app/components/__tests__/App.test.tsx
→ Tests  10 failed | 131 passed (141)
```

Identical with any batch branch applied, with those changes stashed, and with `main`'s own `App.tsx` restored.

## Bisect (clean main checkouts, measured)

| commit | date | result |
|---|---|---|
| `main~400` (`41d60f0355`) | 2026-07-25 | **140 passed** (140 tests) |
| `main~275` (`74d6513fae`) | 2026-07-27 | 3 failed / 141 |
| `main~210` (`d2ce1ba8b5`) | 2026-07-29 | 10 failed / 141 |
| `main` | 2026-07-30 | 10 failed / 141 |

Two-stage degradation, not one regression. The test file itself changed in that window (140 → 141 tests). Three commits touched it: `73b2a32e2b`, `f26cbedf4f`, `f157bf7460`.

## What is actually failing

**Correcting my own earlier characterisation:** I first described this as "the board renders nothing". That is wrong. The app renders fine — header, brand, progress bar are all in the DOM — and in the list-view cases the `.list-view` element renders too, so the test's `waitFor` on it passes. The failure is the next line.

Taking `opens the NewTaskModal from the list view new-task button` as the representative case, it dies on `getByText("+ New Task")`.

That button is conditional:

```tsx
// ListView.tsx:2733
{onNewTask ? (
  <button className="btn btn-task-create btn-sm list-new-task-action" onClick={onNewTask}>
    {t("listView.newTask", "+ New Task")}
  </button>
) : null}
```

Established so far:

- `renderPrimaryActionCluster()` is rendered on **both** layout branches — single-pane (`ListView.tsx:2885`) and desktop (`:2922`) — so a mobile/desktop breakpoint difference does not explain it.
- All four known render sites pass the prop: `App.tsx:1886`, `MainContent.tsx:879`, `:994`, `:1058`.

So the narrow question for whoever picks this up is: **does the `ListView` instance rendered under `<App />` in this test receive `onNewTask === undefined`, or is the `.list-view` node the test finds in a different subtree from the one rendering the action cluster?** Those two are distinguishable in one debugging session; I would start by asserting on `data-testid="list-primary-action-cluster"` (already on the wrapper) to tell "cluster missing" from "button missing".

## A red herring worth naming

The rendered DOM contains `aria-label="header.fusionLogo"` — a raw i18n key — which looks like broken i18n setup and is what I chased second. It is not the cause: that call site passes no inline default, so i18next correctly returns the key, and `t("listView.newTask", "+ New Task")` *does* pass one. `vitest.setup.ts` initialises i18next, and the shard projects use `extends: true`, so the setup is inherited.

## Deliberately not quarantined

`scripts/lib/test-quarantine.json` is for flakes. This fails deterministically, every run, on every commit since 07-25. Quarantining would convert a standing signal into silence, and the deletion ratchet would then delete a suite of 141 tests covering deep-link handling, view switching, board branch filters, and the FN-5817 mobile auto-merge shell. It is also not in the merge gate, so it blocks nobody today.

gsxdsm:
Closes #2829. `App.test.tsx` has been red since **2026-07-25**; #2833 took it 10 failures → 2, and this takes it to **0**.

## The cause, and it was not what I guessed

`vi.mock("../../hooks/useViewportMode")` did not export `isShortViewport`, which `NewTaskModal` imports.

An incomplete module mock **does not fail at the mock**. It throws inside whichever component imports the missing export, and the nearest `ErrorBoundary` swallows that into *"This section encountered an error"* — so the test failed on a **missing heading** with a DOM that looks perfectly healthy. Structurally the same presentation as the workflow-skeleton bug in #2833: the real failure sits two layers below what the assertion reports.

Both remaining tests had this one cause.

## Found by probing, not theorising

I had already been wrong twice on this file (`renders nothing`, then i18n), and my standing note said the modal was "plausibly FN-8620's FloatingWindow rework, **unverified**". That would have been a third wrong guess. Instead I dumped what was actually in the DOM at the assertion point:

```
headings: ["Fusion"] | dialogs: 0
newtask-ish: [..., "error-boundary error-boundary--modal"]
```

The `error-boundary--modal` class was the entire answer; reading its text gave the exact missing export. Same technique that cracked #2833 in one shot — the difference between the two halves of this investigation is that I stopped guessing.

**App.test.tsx: 141 passed (141).**

## Also patched, and one deliberately not

- **`Header.mobile-project-favorites.test.tsx`** mocks the same module with an inline factory and lacks the export. It is **latent, not broken** — its component does not import `isShortViewport` yet, and the day it does, the failure would be this same swallowed crash. One line.
- **`FloatingWindow.touch-geometry.test.tsx`** greps as missing but is **not** — it uses `{ ...(await vi.importActual(...)) }`, so its surface is complete by construction. My grep-based classification was a false positive.

That spread is the pattern that prevents this class outright, and it is the better fix **where it is available**. It is not available here: App.test.tsx's mock exists precisely so tests need no `window.matchMedia` in jsdom, and spreading the real module would reintroduce that dependency for every unstubbed export. So the scoped one-line stub is the right fix for this file, and the spread is the recommendation for new mocks.

A guard asserting "a module mock exports everything the real module does" would close this class repo-wide. I did not build it — 34 files mock this module and only one was genuinely wrong, so the population does not yet justify another instrument, by the same measure-first rule that killed two other candidate gates this week.

## Verification

App.test.tsx 141/141 · the two touched siblings 9/9 · `tsc` 0 · lint 0.

### DIFF
--- a/packages/dashboard/app/components/__tests__/App.test.tsx
+++ b/packages/dashboard/app/components/__tests__/App.test.tsx
@@ -639,6 +639,8 @@ vi.mock("../../hooks/useMobileKeyboard", () => ({
 // Mock useViewportMode so tests can simulate mobile viewport without
 // depending on window.matchMedia in jsdom.
 const mockUseViewportMode = vi.fn(() => "desktop");
+/* `(max-height: 480px)` in production — independent of the width-driven mode. See the note below. */
+const mockIsShortViewport = vi.fn(() => false);
 vi.mock("../../hooks/useViewportMode", () => ({
   MOBILE_MEDIA_QUERY: "(max-width: 768px), (max-height: 480px)",
   isTabletTouchViewport: (mode?: string) => mode === "tablet",
@@ -651,10 +653,25 @@ vi.mock("../../hooks/useViewportMode", () => ({
   An INCOMPLETE module mock does not fail where the export is missing — it throws inside whichever
   component imports it, and the nearest ErrorBoundary swallows that into "This section encountered an
   error". NewTaskModal adopted `isShortViewport`, this mock did not, and the test failed on a MISSING
-  HEADING with a healthy-looking DOM. The real function reads the short-viewport media query, so
-  keying it to the mobile mode matches how the siblings above are stubbed.
+  HEADING with a healthy-looking DOM.
+
+  FNXC:TestViewportMock 2026-07-30-19:50 (#2846 review — greptile P2, "viewport predicates are conflated"):
+  SHORT-VIEWPORT IS ITS OWN CONTROL, because in production it is its own MEDIA QUERY.
+
+  The first version keyed it to `mode === "mobile"`, matching how the siblings above are stubbed. The
+  siblings are width predicates and the mode IS their answer; this one is not. `isShortViewport()`
+  reads `(max-height: 480px)` alone, while the mobile mode is the OR of width and height — so an
+  ordinary PORTRAIT PHONE (narrow, tall) is mobile and NOT short, and the mock claimed it was both.
+
+  What that silently mis-tested: `FloatingWindow` suspends geometry PERSISTENCE on a short viewport,
+  and `PlanningModeModal` picks its compact interview layout and hides the session list from it. Every
+  mobile test here took those branches, so the ordinary phone case — the most common real viewport —
+  was never actually exercised, and a regression in the non-short mobile path would have passed.
+
+  Defaults to FALSE rather than to the mode: a test that means "short" now has to say so, which is the
+  only spelling that can distinguish the two.
   */
-  isShortViewport: () => mockUseViewportMode() === "mobile",
+  isShortViewport: () => mockIsShortViewport(),
 }));
 
 // Mock isIOS so FN-3290 keyboard-open behavior is testable in jsdom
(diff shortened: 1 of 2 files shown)


gsxdsm:
**Confirmed and fixed.** 997cb0127a.

The mock keyed `isShortViewport` to `mode === "mobile"` to match the sibling predicates stubbed just above it. Those siblings are **width** predicates, and the mode is genuinely their answer; this one is not. Production reads `(max-height: 480px)` **alone**, while the mobile mode is the OR of width and height — so an ordinary **portrait phone** (narrow and tall) is mobile and *not* short, and the mock claimed it was both.

**What that silently mis-tested:**
- `FloatingWindow` suspends geometry **persistence** on a short viewport (`FloatingWindow.tsx:243`).
- `PlanningModeModal` picks its compact interview layout and hides the session list from it (`PlanningModeModal.tsx:726`, `:3715`).

Every mobile test in this file took those branches, so the ordinary-phone case — the most common real viewport — was never actually exercised, and a regression in the non-short mobile path would have passed.

`isShortViewport` now has its own `vi.fn`, defaulting to **false** rather than to the mode, so a test that means "short" has to say so; that is the only spelling that can distinguish the two. It is reset beside the mode so a suite that sets it cannot leak into the next.

**141/141 App tests still pass, which is the expected result and worth stating plainly:** no existing case *depended* on the conflation — they were taking the wrong branch on the way to an assertion that held either way. The value here is that the branches are now selectable, not that a test flipped.

Dashboard app `tsc` clean, `pnpm lint` clean.

PR #2851, opened 2026-07-30
fix(gate): pnpm test is red on main — stale exemptions, plus two false positives they were masking

gsxdsm:
**`pnpm test` and `pnpm test:full` fail on `main` right now** (commit `51934931e1`). Found by running the gates against `main` after my earlier PRs landed, not by CI telling me.

> **Correction to this PR's first version.** I originally wrote that `pnpm test:gate` was failing. It is not: `check-inert-flag-seams` runs in the `pretest` and `pretest:full` hooks, not in `test:gate`. So the blocking merge gate (Lint / Typecheck / Build / Gate) is unaffected and PRs are not blocked — what is broken is every local `pnpm test` run, which fails before a single test executes. Lower urgency than I claimed, still worth fixing promptly, and I would rather correct the scope than leave an overstated one standing.

Three causes, each surfaced by fixing the one before it.

## 1. Stale exemptions — the mechanism working

#2819 and #2823 merged, so the two `ALLOWED_OMISSIONS` entries covering those call sites became stale and the staleness check failed them. Removed. This is my cleanup: the entries were designed so they could not outlive their fixes, and they didn t.

## 2. Renamed imports were not resolved

Removing the first entry surfaced:

```
enqueueMergeQueue() — best call passes 2 of 5
```

Its only production caller passes all five — through `import { enqueueMergeQueue as enqueueMergeQueueAsync }`. Call sites were recorded under the **local** name, so an aliased supplier was invisible and the seam read as unsupplied. The local name is now mapped back to the exported one.

## 3. Method calls were conflated with module functions

That fix then surfaced two engine sites as omitting — but `store.enqueueMergeQueue(taskId, opts)` is a **2-arg `TaskStore` method** that resolves the review columns internally (#2819), not the 5-arg module function sharing its name. Property-access calls are no longer attributed to module-level seams.

**Tradeoff, stated at the site:** a genuine `namespace.fn(...)` call is now skipped. This codebase calls module functions as bare identifiers, and aliases are resolved by fix 2, so that shape does not currently occur. Recorded rather than left for someone to discover.

## Both directions re-verified

A fix that quietly disarms the gate would be worse than the red, so I re-ran the defects it exists to catch:

| mutation | result |
|---|---|
| drop `Column.tsx`'s flags argument (partial supply) | `supplied by 10/11 call sites; omitted at .../Column.tsx:1 (of 2)` |
| drop the aliased 5-arg supplier (wholly unsupplied) | `best call passes 3 of 5` |
| restored | exit 0 |

My first attempt at the second row grepped for the wrong message shape and printed nothing. **I re-ran it rather than reading silence as success** — which is the failure this gate exists to prevent, and one I have made in this same file before.

## Verification

`pnpm test:gate` green (it was never affected) · `node scripts/check-inert-flag-seams.mjs` exit 0 · lint 0 · gate reports `21 lane/flag seams, all supplied at every production call site`.


gsxdsm:
**Real, and the sharper version of it is that every clause in that classifier is a shipped bug with no test.** 977bb54e6a.

Both rules the finding names were added to fix failures the gate had already produced — the alias remapping after it reported `enqueueMergeQueue() — best call passes 2 of 5` while its only production caller supplied all five through an alias, and the property-access exclusion after it turned main red over two correct call sites. Neither had coverage.

**Extracted two pure functions** so the rules are testable without walking the repo:
- `isRelevantCallSite(site, declaringFile)` — what a recorded call means.
- `collectImportBindings(sf)` — the alias/module maps, from an AST.

**Both directions are pinned, because they fail differently and one of them fails silently:**

| direction | symptom |
|---|---|
| a clause stops excluding | gate goes **red on correct code** — and a guard that cries wolf teaches its readers to skip the line that matters |
| a clause over-excludes | gate goes **green on a real offender** — basename comparison once dropped every barrel-imported call, so four engine sites omitting a required argument were invisible while the report read "supplied by 5/6" |

The second is worse and harder to notice, so each exclusion is paired with the inclusion it must not swallow. The `isTest` case is pinned too, precisely because it *looks* like an omission: test sites are deliberately classified relevant here so they can be **reported** without ever **clearing** a seam — counting them as suppliers is what re-hid `sortTasksForDisplayColumn`.

The alias test asserts the **direction** rather than "the map is populated". In the TS AST, `import { a as b }` puts the exported name in `propertyName` and the local name in `name`; reading them backwards still yields a populated map and still runs — it just attributes calls to a name nothing declares.

**Mutation-verified per rule**, since one blanket check would not show that each clause is independently covered:

| mutation | result |
|---|---|
| drop the property-access exclusion | 1 failed / 11 passed |
| re-exclude barrel imports | 3 failed / 9 passed |
| exclude test sites in the classifier | 1 failed / 11 passed |
| invert the alias direction | 1 failed / 11 passed |
| restored | 12 passed |

**One thing I fixed along the way that the finding did not ask for:** importing the module ran the gate to completion, printed its report, and could `process.exit(1)` — so the new test file would have passed or failed on unrelated repo state rather than on what it asserts. The reporting is now behind an entry-point guard; the AST walk stays unguarded on purpose (no output, no exit, and a future test may want `declared`/`callSites`). Same fix and same reason as `check-sql-column-literals`, where I hit it first.

`pnpm lint` clean; gate still reports 21 seams, all supplied.

gsxdsm:
Ready and green — flagging priority because this one unblocks others.

`node scripts/check-inert-flag-seams.mjs` exits **1 on main right now**, and it runs in the `pretest` / `pretest:full` hooks, so **every local `pnpm test` and `pnpm test:full` dies before a single test executes.** CI is unaffected — `test:gate` does not include this check, which I verified on a clean `main` checkout and corrected in the PR body after initially overstating it.

This branch now also carries #2853 (merged into it via the stack), so it delivers both the stale-entry removal and the corrected wording on the last `TEMPORARY` exemption. Gate exits 0 here.

The remaining queue behind it — #2841, #2846, #2848, #2850 — are independent of this one and of each other.

PR #2852, opened 2026-07-30
fix(guard): the unwired-lane check reported 0 because its question was too weak

gsxdsm:
A guard nobody has proven can fail is a number, not a check. This one was returning a clean `[]` while **18** real unwired lane declarations sat on `main` — including one it was specifically built to catch.

## The escape that found it

`diffSnapshots` in the glasses plugin:

```ts
opts: { notifyOnColumns: ReadonlySet<ColumnId>; completeColumnsByTaskId?: ReadonlyMap<...> }
...
const completeColumns = opts.completeColumnsByTaskId?.get(task.id);
const isComplete = completeColumns ? completeColumns.has(task.column) : task.column === "done";
```

**No file anywhere builds that map.** The conversion was decorative — the literal decided every real poll, so on a renamed board the wearer is notified of every column transition *except the card finishing*, the one they care about. The name was already in the guard's vocabulary list, the declaration was exported and optional; it satisfied every condition the guard checks, and the guard said nothing.

## Three independent blind spots

| # | blind spot | why it mattered |
|---|---|---|
| 1 | `SCANNED_PACKAGES` omitted `plugins/` | plugins hold lane logic like anything else — this one resolves workflow IRs and decides what "finished" means |
| 2 | inline options-object types were not walked | only bare parameters and *named* interfaces were. Whether a lane answer arrives as a parameter, an interface property, or an inline field is a style choice — **a check evadable by a style choice is decorative** |
| 3 | the mention rule was `source.includes(parameter)` **anywhere** | satisfied by coincidence for any ordinarily-named parameter |

Fixing 1 or 2 alone would still have missed it: **measured on `main`, the guard found 0 unwired across 1753 files, and 0 again across 2114 once `plugins` was added**, because the shape was invisible too.

### On (3), I proved it on myself

Renaming the unwired parameter from `completeColumnsByTaskId` to `completeColumns` — a better name, chosen for good reasons — **silenced the guard instantly**, because 15 unrelated production files declare a local called `completeColumns`. The check had not been satisfied; it had been switched off by a rename. That is exactly the failure the code comment two lines up condemns, committed one edit later.

The fix is the cause, not a name blocklist: a file that never references `diffSnapshots` cannot be the thing that wires `diffSnapshots`'s options. Still deliberately loose — a co-occurrence test, not call-graph analysis — which keeps the low false-positive rate that makes the guard bearable while removing a false **negative** that scaled with how ordinary a parameter's name was.

## The 17 this uncovered

Tightening (3) surfaced 17 further unwired declarations across core, engine and dashboard. **Spot-checked, not assumed**: `buildUnblockWeightMap` in `task-priority.ts` declares `terminalColumns` and `reviewColumns`, and the only files that pass either are its own tests — the production caller silently uses the built-in `{done, archived}` default. That is the inert-conversion shape this module exists to name.

They span three other batches, so they are recorded as a **ratcheted baseline** in the shape `scripts/lifecycle-column-census.mjs` already uses here — keyed on `file + parameter` so an unrelated edit above them cannot manufacture a failure. A new one fails immediately; these can only leave the list. Listing them beats pretending for another week that they do not exist.

## The glasses fix

`completeColumnsByTaskId` -> a flat `completeColumns` set, matching its sibling `notifyOnColumns` in the same options object, resolved **once per poll** by `notifier.ts` via `resolveProjectColumnsForRoles`. Project-scoped and not per task because this runs on a polling timer over the whole board — a per-card workflow read would scale with the board on every tick. Best-effort: a failed resolve leaves the diff on its documented legacy default rather than dropping a poll.

Still gated by `alsoNotifyOnDone`, which the production caller passes as `false`, so it remains unobservable at runtime. Wired anyway: the day someone enables the flag the resolution must already be right — and now the guard will say so if the wiring is removed.

## Revert proofs (measured, one per fix)

| revert | failure |
|---|---|
| unwire `notifier.ts` | baseline gains `plugins/…/diff.ts completeColumns` |
| drop the inline-options walk | "covers an INLINE options-object type" fails `expected [] to deeply equal [ 'completeColumns' ]`, and the repo scan loses the glasses entry |
| drop the owner scoping | the repo scan loses **all 17** pre-existing entries |
| restore `task.column === "done"` | both new `diff.test.ts` cases fail |

The new diff cases assert **both** directions — a card in the resolved lane fires, and a card in the legacy `done` does *not* once the caller resolved other lanes. The second is what proves the resolved set replaces the default rather than being unioned with it.

## Verification

- `pnpm lint` — clean
- `tsc --noEmit` — clean
- guard suite — 9/9; full glasses plugin — 188 passed across 19 files

### DIFF
--- a/.changeset/glasses-completion-lanes.md
+++ b/.changeset/glasses-completion-lanes.md
@@ -4,4 +4,4 @@
 
 summary: Glasses completion notifications now recognise a board whose finished lane is not called Done.
 category: fix
-dev: `diffSnapshots` declared a per-task `completeColumnsByTaskId` that no caller ever built, so its completion test fell through to the literal `"done"`. Replaced with a flat `completeColumns` set resolved once per poll by `notifier.ts` via `resolveProjectColumnsForRoles`. The `unwired-lane-parameter` guard now scans `plugins/`, walks inline options-object types, and scopes its "is it wired" search to files that name the declaring symbol — which surfaced 17 further unwired declarations, now recorded as a ratcheted baseline.
+dev: `diffSnapshots` declared a per-task `completeColumnsByTaskId` that no caller ever built, so its completion test fell through to the literal `"done"`. `notifier.ts` now builds it (per task, with a shared IR cache, and only when `alsoNotifyOnDone` is on). The `unwired-lane-parameter` guard now scans `plugins/`, walks inline options-object types, and scopes its "is it wired" search to files that name the declaring symbol — which surfaced 17 further unwired declarations, now recorded as a ratcheted baseline.
--- a/.changeset/glasses-completion-lanes.md
+++ b/.changeset/glasses-completion-lanes.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Glasses completion notifications now recognise a board whose finished lane is not called Done.
+category: fix
+dev: `diffSnapshots` declared a per-task `completeColumnsByTaskId` that no caller ever built, so its completion test fell through to the literal `"done"`. Replaced with a flat `completeColumns` set resolved once per poll by `notifier.ts` via `resolveProjectColumnsForRoles`. The `unwired-lane-parameter` guard now scans `plugins/`, walks inline options-object types, and scopes its "is it wired" search to files that name the declaring symbol — which surfaced 17 further unwired declarations, now recorded as a ratcheted baseline.
--- a/packages/engine/src/__tests__/unwired-lane-parameter-guard.test.ts
+++ b/packages/engine/src/__tests__/unwired-lane-parameter-guard.test.ts
@@ -32,7 +32,17 @@ import { describe, expect, it } from "vitest";
 import { findUnwiredLaneParameters, LANE_PARAMETER_NAMES } from "../../../../scripts/lib/unwired-lane-parameter.mjs";
 
 const REPO_ROOT = resolve(import.meta.dirname, "../../../..");
-const SCANNED_PACKAGES = ["packages/core/src", "packages/engine/src", "packages/dashboard/src", "packages/dashboard/app", "packages/cli/src"];
+/*
+FNXC:WorkflowLifecycleColumns 2026-07-31-01:20:
+`plugins` is scanned, and its absence was half of a real escape.
+
+Plugins hold lane logic like anything else — the glasses plugin resolves workflow IRs, filters by
+column, and decides what "finished" means for a notification — and this list simply did not look at
+them. Combined with the inline-options blind spot below, an unwired `completeColumnsByTaskId` sat on
+`main` unreported: the guard found 0 across 1753 files, and 0 again across 2114 once plugins were
+added, because the shape was invisible too. Fixing either alone would still have missed it.
+*/
+const SCANNED_PACKAGES = ["packages/core/src", "packages/engine/src", "packages/dashboard/src", "packages/dashboard/app", "packages/cli/src", "plugins"];
 
 function sourceFiles(dir: string, acc: string[] = []): string[] {
   for (const entry of readdirSync(dir)) {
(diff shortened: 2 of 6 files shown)


gsxdsm:
**Confirmed — and it is the union rule failing on the side where it is not safe.** I am not taking the branch (see the last paragraph), so this is the analysis and the recommended fix.

`resolveProjectColumnsForRoles` seeds `LEGACY_COLUMN_IDS_BY_ROLE` unconditionally (`complete: ["done"]`, `project-lane-vocabulary.ts:76-78`) and then *adds* whatever the project's workflows declare. There is no way to opt out of the floor. So on a board that declares a column literally named `done` carrying some other role — or no role — a card entering it fires a **completion notification** to the wearer.

**Why this one is different from the other union call sites I have been defending.** The rule I have been applying across this program is: the project union is correct where over-inclusion is *invisible* — a read, a query, a candidate filter that something else narrows afterwards. Over-including there costs a wasted row. This caller is not that. A notification is an **outward action with no downstream narrowing**: once it fires, it has already reached the user. Over-inclusion here is a wrong notification, not a wasted read. Same helper, opposite safety direction.

The same argument applies in reverse to the sibling `notifyOnColumns` setting, which is operator-authored and names real lanes — so it is not affected.

**Recommended fix, and the shape I would take.** The correct answer is the three-state rule this program has been applying at the per-task level, lifted to project scope: a project whose workflows express lifecycle traits has *made a statement*, and its resolved lanes should replace the legacy floor rather than be unioned with it; a project whose workflows express no trait at all (v1-upgraded, `synthesizeDefaultColumns` emits every default column with `traits: []`) keeps the floor. That wants an option on `resolveProjectColumnsForRoles` — something like `legacyFallback: "union" | "only-when-untraited"` — rather than a subtraction local to the notifier, because a local `delete "done"` duplicates role logic in a plugin and will drift.

**Do NOT fix it by dropping the floor for every caller.** The floor exists for a measured reason: the same helper's history records that comparing basenames once dropped every barrel-imported call and a mid-rename board must still surface rows stored under the old id. The read-shaped callers need it; this action-shaped one does not.

**Why I am not pushing the fix.** `unwired-guard-inline-options` is checked out in another worktree (`phase-c-planning`), so it has an active owner with possibly unpushed work. Pushing a shared-helper change onto a branch someone else is holding is how two agents lose an afternoon. Whoever owns that worktree should take this; if it is meant to be mine, say so and I will pick it up once the branch is released.

gsxdsm:
Greptile P2 (resolver retains legacy completion lane) — **agreed, fixed in 6fdfc99**, and it caught me making the exact mistake I wrote the warning against.

`resolveProjectColumnsForRoles` **always** unions the legacy id in. That is correct and deliberate for a **read** — an extra id in a `listTasks({ column })` set is inert — and it is a **false positive** for a per-card decision. A workflow declaring `shipped` as its complete lane while reusing `done` as an ordinary lane would have fired a `completed` notification for live work.

The header of `project-lane-vocabulary.ts` names this precisely:

> **WHAT THIS IS NOT.** It does not tell you what a given CARD's lane is… Answering a per-card question from this union would mark a card as review because some other workflow calls its column review, which is the flat-set mistake this program has already made four times.

Fifth. One module over, by the person who wrote it.

**And my own test could not have caught it.** "does NOT treat the legacy `done` as complete" hand-builds `new Set(["shipped"])`, so it asserts what `diffSnapshots` does *with* an answer and says nothing about the answer the notifier *supplies*. That is exactly the "tests inject the value by hand" blind spot this PR's guard doc describes — present in this PR's own test.

**The fix restores the original author's shape**, whose comment gave the right reason all along: the poll spans the whole board, so one workflow's complete column id can be another workflow's WIP id. What they got wrong was the wiring, not the shape. `notifier.ts` now builds `completeColumnsByTaskId` per task with a shared IR cache — so distinct *workflows*, not distinct cards, drive the resolution count — and builds it **only when `alsoNotifyOnDone` is on**.

That gate is what answers the cost objection I had used to justify the flat set: the flag is `false` here today, so the per-task reads cost exactly nothing now, and the day someone enables it they get a correct answer instead of a cheap wrong one. A card whose workflow cannot be resolved is simply absent from the map and falls back to the documented default, rather than dropping the whole poll.

The guard baseline is unchanged (17 pairs) — the glasses declaration stays wired.

Re-verified: `pnpm test:gate` 161/487/13/71, `pnpm lint` clean, `tsc --noEmit` clean, guard 9/9, full glasses plugin 188/188. Revert proof re-measured: restore `task.column === "done"` and both diff cases fail.

gsxdsm:
The per-task fix landed and matches the analysis I posted — good. Two follow-ups on the rebased tip.

**Fixed:** the FNXC stamp was `2026-07-31` on 2026-07-30. That is the third future-dated stamp a review has caught across my PRs today; worth a linter rather than three separate catches.

**Reported, not fixed — and it means this seam is still inert.** `alsoNotifyOnDone` is a hardcoded `false` with no setting behind it (grep it: this line is the only writer). So the block that builds `completeColumnsByTaskId` never runs, the map is **always `undefined`**, and `isComplete && opts.alsoNotifyOnDone` in `diff.ts` can never fire. The per-task lane resolution this change set exists to wire is still dead in production — one level below the omission it replaced.

**And `check-inert-flag-seams.mjs` cannot see that.** The check asks whether a call site *supplies* the argument. Line 128 does — as a variable that is always `undefined` because a constant `false` guards its construction. **"Supplied" and "supplied with a real value" are different questions, and only the first is being asked.** This is the sharpest instance of the class the checker was built for, sitting inside the checker's own blind spot, and it is the same shape as `sortTasksForDisplayColumn` (supplied only by its own tests) one turn further on: there the *callers* were not production, here the *value* is not.

I left the constant alone deliberately. Turning it into a setting makes the glasses start emitting completion notifications nobody has opted into — that is this plugin's UX call, not a conversion batch's. What must not happen is the seam being counted as fixed, so it is recorded at the site.

Rebased onto `origin/main`, which also cleared the two stale `ALLOWED_OMISSIONS` entries this branch was red on (`enqueueMergeQueueInTransaction`, `isNearDuplicateCanonicalInactive` — #2851's work landing). Gate now reports 21 seams, all supplied; `pnpm lint` clean.

PR #2857, opened 2026-07-30
test(engine): record the partial planning-continuation conversion the audit caught

gsxdsm:
## The audit fired, as designed

`workflow-planning-continuation-terminal-gap-live-e2e.pg.test.ts` went red on `main`:

```
× AUDIT — one of the two classifier call sites is converted, and the inner predicate is not
   → expected 2 to be 1
```

That is the alarm working. #2799 built this case to fail **in both directions** — a new unconverted caller *or* a conversion of an existing one — precisely so a change here cannot land silently. Someone converted the inner half:

```ts
if (!isPlanningContinuationTaskDispatchable(task, terminal)) {
```

## What actually changed, and what did not

| site | before | now |
|---|---|---|
| `drainDuePlanningContinuations` | passes `{ terminalColumns }` | unchanged ✅ |
| `resolvePlanningContinuationCandidate` → inner predicate | **not threaded** | **threaded** ✅ |
| `selectActionablePlanningContinuations` | passes nothing | **still passes nothing** ❌ |

**The behavioural case is untouched and still passes.** A card in a renamed COMPLETE lane still comes back `actionable` and still re-enters plan-review, because `selectActionablePlanningContinuations` is the site that decides that.

That combination is the useful shape of a **partial conversion**: the code reads more converted than before, and the operator-visible defect is exactly where it was. Without the behavioural case sitting next to the audit, the arity change would have looked like the fix.

The fixer's own comment agrees, and names the narrow case the inner half reaches — a board declaring `done` as a *non-terminal* column id, where the outer check passes and the inner one skipped the continuation as "paused". Real, and not the case the characterization covers.

## The change

Arity assertion updated `1` → `2`, with the reasoning recorded at the assertion and in the header. **Updated deliberately, not loosened** — arity is still the property asserted, so the next change to either site lands here again. Nothing else moved.

## Verification

- suite — **4/4**
- full live-PG E2E surface — **171/171**
- `pnpm lint` — clean

Test-only; no changeset.

### DIFF
--- a/packages/engine/src/__tests__/workflow-planning-continuation-terminal-gap-live-e2e.pg.test.ts
+++ b/packages/engine/src/__tests__/workflow-planning-continuation-terminal-gap-live-e2e.pg.test.ts
@@ -176,7 +176,7 @@ pgDescribe("planning-continuation terminal columns, measured on a live store", (
     expect(resolved.kind === "orphan" ? resolved.reason : null).toBe("task-terminal");
   });
 
-  it("AUDIT — one of the two classifier call sites is converted, and the inner predicate is not", async () => {
+  it("AUDIT — the inner predicate is threaded; one of the two classifier call sites still is not", async () => {
     /*
     NOT driven: reaching the drain needs the runtime's full dependency set. Asserted against the
     module's SYNTAX and labelled as such.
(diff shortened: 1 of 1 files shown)


gsxdsm:
Fixed in the latest push — the finding is right, and it is a real problem rather than a wording nit.

I updated the assertion to record that the inner predicate is now threaded, and left the name saying it is not. Test output is the first thing anyone reads when this fires, and this file's entire job is to describe *which half* of a partial conversion has landed. A name asserting the opposite of the assertion would have sent the next reader looking at the wrong call site — the precise failure this suite exists to prevent, in the suite itself.

```
- AUDIT — one of the two classifier call sites is converted, and the inner predicate is not
+ AUDIT — the inner predicate is threaded; one of the two classifier call sites still is not
```

The name now states both halves in the order the assertions check them, so a future flip on either side reads correctly straight from the runner output.

4/4 green, lint clean.

PR #2861, opened 2026-07-30
fix(reliability): the review-failure headline read 0% because both inputs were zero

gsxdsm:
The Reliability panel's headline metric is computed from two queries that name lanes:

```ts
scopedStore.getTaskMovedCountsByDay({ …, toColumn: "in-review" }),
scopedStore.getTaskMovedCountsByDay({ …, fromColumn: "in-review", toColumn: "in-progress" }),
…
const headline = inReviewFailureRate7d(enteredByDay, bouncedByDay, nowMs);
```

On a board that renamed either lane, **both return `{}`**. Every per-day count is zero, and the headline then divides one zero by another and reports a healthy rate.

**This is the worst shape a lifecycle defect takes.** It produces a *number*, not an error, and the number is reassuring. An operator reading a 0% review-failure rate beside a populated audit event list has no reason to suspect the metric is blind — the same failure mode as the analytics `tasksInProgress: 0` sitting next to correct cost totals.

## Why a union is correct here, not a compromise

`getTaskMovedCountsByDay` takes **one** column per side, so the lanes are resolved to sets and the query is issued per `(from, to)` pair and summed.

The important part is *why* summing over a union is the right answer rather than a widening hack. These read **move history**, and a past move recorded the column name as it was at the time — the same reasoning that keeps `tasksEnteredInReviewPerDay` in this module matching recorded values verbatim, marked DELIBERATE-LITERAL. A board renamed last month therefore has old rows under the old id and new rows under the new one, so the correct query covers **both**. That is exactly the set `resolveProjectColumnsForRoles` returns, with the legacy id always unioned in.

Asking for either name alone is the bug — not a choice between them.

No double-counting: a move event has exactly one `(from, to)` pair, so the queries partition the events rather than overlapping.

**The common path does not get more expensive.** On the built-in board this issues the same two queries as before, and there is a test asserting the call count so a future change cannot quietly turn one query into N.

## Revert proof (measured)

Collapse the sets back to the single literals:

```
AssertionError: expected { '2026-07-01': 1 } to deeply equal { '2026-07-01': 2, '2026-07-02': 3 }
AssertionError: expected { '2026-07-01': 1 } to deeply equal { '2026-07-01': 2, '2026-07-03': 5 }
```

The helpers live in `reliability-metrics.ts` rather than inline in `server.ts` specifically so they are testable without booting an express app.

## Verification

- `pnpm smoke:boot` — PASS (`fn --help`, real `serve` `/api/health` 200, clean shutdown)
- `pnpm lint` — clean
- `tsc --noEmit` (`@fusion/dashboard`) — clean
- `reliability-metrics.test.ts` — 15 passed
- census `--strict` — exit 0 (unchanged: query filters are not comparisons)

### DIFF
--- a/.changeset/reliability-lane-metrics.md
+++ b/.changeset/reliability-lane-metrics.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: The Reliability health panel no longer reports a perfect review-failure rate on a renamed board.
+category: fix
+dev: `/api/health/reliability` counted review entries and bounces with two `getTaskMovedCountsByDay` queries naming `in-review` and `in-progress`. On a renamed board both returned `{}`, so every per-day count was zero and `inReviewFailureRate7d` divided one zero by another and reported healthy. The lanes are now resolved via `resolveProjectColumnsForRoles` and the query is issued per (from, to) pair and summed.
--- a/packages/dashboard/src/__tests__/reliability-metrics.test.ts
+++ b/packages/dashboard/src/__tests__/reliability-metrics.test.ts
@@ -5,6 +5,7 @@ import type { ActivityLogEntry, RunAuditEvent } from "@fusion/core";
 import {
   bucketByDay,
   countBouncesOut,
+  countEntriesInto,
   countMovesInto,
   dayHasSamples,
   fileScopeInvariantFailuresPerDay,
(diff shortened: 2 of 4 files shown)


gsxdsm:
**Disclosure: this fixes two of the panel's three inputs, and I would rather say so than let the PR read as "Reliability is fixed".**

The same handler builds `duration` from `scopedStore.getInReviewDurationEvents(...)`, which resolves to `getInReviewDurationEvents` in `packages/core/src/task-store/async-audit.ts:337`. It filters with **raw SQL**:

```sql
metadata->>'to' = 'in-review'
  OR (metadata->>'from' = 'in-review' AND metadata->>'to' = 'done')
```

On a renamed board that matches nothing, so `inReviewDurationMetrics` reports `no-in-review-entries` while the two counts this PR fixes start working. The panel goes from *uniformly* blind to *partially* blind — better, and still not honest end-to-end.

**Why it is not in this PR.** It is `@fusion/core` (batch-core's package), and it is a SQL predicate rather than a call argument, so fixing it means threading a resolved lane set into a raw `sql` fragment and its `projectId`-scoped query — a different change with a different review surface than the two call-site substitutions here. Splitting it keeps this PR revertible on its own.

**Worth flagging as a class, not just a site.** A lane literal encoded inside a raw `sql` template is invisible to *every* check we have: the census scans `===`/`!==` comparisons, the unwired-parameter guard scans declarations, and neither sees a string inside a SQL fragment. This is the second known instance — the first was the archived gate in PR #2724 — which is enough to call it a pattern rather than an accident. The backlog number is a floor for this reason, and the SQL-encoded ones are not in it at all.

Filed onto `handoff/archived-sentinel-literals-for-batch-core` alongside the other core findings.

gsxdsm:
Greptile P1 and P2 — **both right, fixed in 576e743.**

**P1 (intra-review moves inflate entries).** This is a defect the single-lane version could not have had, and I did not think to ask the question when generalising to a set. A board with `signoff` and `waiting` both carrying review roles has moves *between* them, and counting by destination alone scores `signoff → waiting` as another entry into review. The denominator inflates while the bounce count is unchanged, so the headline **understates** the review-failure rate — wrong in the *reassuring* direction, which is precisely the failure mode this PR exists to remove. My fix for a blind metric would have shipped a flattering one.

`countEntriesInto` now subtracts intra-set moves, and **skips that subtraction entirely for a single-lane set**: a move requires the column to change, so a one-lane set has no internal moves and must not pay for a query asking about them. The built-in board still issues exactly the queries it did before, and there is an assertion on the call count so that cannot drift.

The subtraction's edges are pinned too — a day whose every entry was internal drops out of the map rather than reporting a zero, and a negative can never surface.

**P2 (serial queries).** Correct and worth doing: the per-lane and per-pair reads are independent, so they now run under `Promise.all`. Latency is one round trip deep instead of `N + N×M`.

Revert proof, measured: drop the subtraction and both new cases fail — `expected { '2026-07-01': 3 } to deeply equal { '2026-07-01': 2 }` and `expected { '2026-07-01': 1 } to deeply equal {}`.

Re-verified: `pnpm test:gate` 161/487/13/71, `pnpm smoke:boot` PASS, `pnpm lint` clean, `tsc --noEmit` clean, `reliability-metrics.test.ts` 18 passed.

(The disclosure above about `getInReviewDurationEvents` still stands — the panel's third input remains blind on a renamed board, and that one is core's raw-SQL predicate.)

gsxdsm:
Follow-up: the third input is now fixed too — **#2875** converts `getInReviewDurationEvents`, whose lane ids were baked into a raw `sql` predicate. Tested against real PostgreSQL rather than a mocked store, since a mock would assert the arguments and prove nothing about the query that runs. With that merged, all three Reliability inputs read the board's own lanes and the panel is no longer partially blind.

PR #2864, opened 2026-07-30
fix(core): team analytics reported an idle project on a renamed board

gsxdsm:
First of the 14 genuinely lane-bound SQL sites from #2839. I had been deferring these as "the owner is active in those files" — then checked, and **no open PR touches them**. The batch-core commit had already landed, so there was nothing in flight to collide with. The deferral was an assumption I could have tested two turns earlier.

## The defect

`aggregateTeamAnalytics` filtered in SQL on `"column" = 'done'` and `IN ('in-progress','in-review')`. On a board whose lanes are renamed those match nothing, so per-agent completed counts, the project total, and the in-flight breakdown all came back **zero**.

Nothing errors. A dashboard reading *"0 tasks completed"* for a team that shipped all week looks like an idle project, not a bug — which is why this survives review and never gets filed.

**Why the sweep missed it:** the lifecycle census parses TypeScript comparisons; these ids live inside SQL strings, which are string data. The batch-core conversion fixed this file's TS guards **today** and left the queries untouched. The file scored as converted.

## Resolved per project, not per task

`resolveProjectColumnsForRoles` gives the union of a role's columns across the project's workflows, which is the right set here because analytics aggregates a whole project — so a bound `IN` list is sufficient. The merge-queue cleanup needed the *superset-then-decide-in-JS* shape instead, because its lanes are genuinely per task and SQL cannot know a task's workflow. Same program, two correct answers; worth not copying the wrong one.

## Measured

Reverted, exactly one case flips:

```
✓ default vocabulary: a completed task is counted
× renamed vocabulary: a task in the RENAMED complete lane is counted
✓ renamed vocabulary: a task in the WIP lane is NOT counted as completed
✓ without a lane store, the legacy ids still answer
```

The three controls are deliberate: the default vocabulary (a generally broken aggregator cannot hide behind the renamed case), a WIP task that must **not** count as completed on the renamed board (resolving real lanes must not degrade into "every column counts" — an undercount turned overcount is harder to notice), and an omitted lane store that must keep the legacy answer byte-identical.

## Two mistakes the first attempt made, both caught by running it

- **`= ANY(${array})` does not work.** Drizzle expands a JS array in a template into a comma tuple, so PostgreSQL rejected `(($1,$2,$3))` with *"op ANY/ALL (array) requires array on right side"*. An `IN` list of individual bound parameters is the working shape. Each id stays a parameter — these come from operator-authored workflow definitions and are never interpolated as SQL text.
- The fixture's `ON CONFLICT (id)` had no matching constraint on `project.agents`; the sibling suite seeds with explicit `created_at`/`updated_at`.

## Scope

One production caller (`register-command-center-routes.ts`), threaded in this same change so the parameter has a supplier from the start rather than becoming another inert seam of exactly the class this program keeps finding.

The sync SQLite arm in the same file keeps its literals: that path throws in backend mode and has no production caller, the same dead-arm conclusion reached for `cleanupStaleMergeQueueRowsImpl` on #2839.

## Verification

### DIFF
--- a/.changeset/team-analytics-renamed-lanes.md
+++ b/.changeset/team-analytics-renamed-lanes.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Command Center team analytics now count completed and in-flight work on renamed boards.
+category: fix
+dev: `aggregateTeamAnalytics` takes an optional lane store and resolves complete / wip / human-review columns via `resolveProjectColumnsForRoles`; its SQL previously filtered on the literal `'done'` and `('in-progress','in-review')`, which match nothing on a custom workflow.
--- a/packages/core/src/__tests__/postgres/team-analytics-renamed-lanes.pg.test.ts
+++ b/packages/core/src/__tests__/postgres/team-analytics-renamed-lanes.pg.test.ts
@@ -1,0 +1,137 @@
+/*
+FNXC:WorkflowResolvedColumns 2026-07-30-17:40 (analytics reported an idle project on a renamed board):
+
+`aggregateTeamAnalytics` filtered on `"column" = 'done'` and `"column" IN ('in-progress','in-review')`
+directly in SQL. On a board whose lanes are renamed, every one of those matched nothing: per-agent
+completed counts, the project total, and the in-flight breakdown all came back ZERO.
+
+Nothing errors, which is what makes it expensive. A dashboard reading "0 tasks completed" for a team
+that shipped all week looks like an idle project, not a bug, and wrong-but-plausible numbers are the
+least likely defect for anyone to file.
+
+WHY NO EXISTING CHECK SAW IT. The lifecycle census parses TypeScript comparisons; these ids live
+inside SQL strings, which are string data. The conversion sweep that fixed this file's TypeScript
+guards left the queries untouched, and the file scored as converted.
+
+The cases are DIFFERENTIAL: the same seeded work, aggregated twice, under two vocabularies whose
+roles are identical and only the ids differ. `shipped` and `building` collide with no legacy id, so a
+surviving `'done'` cannot pass by luck.
+*/
+
+import { it, expect, beforeAll, beforeEach, afterEach, afterAll } from "vitest";
+import { sql } from "drizzle-orm";
+import {
+  pgDescribe,
+  createSharedPgTaskStoreTestHarness,
+  type SharedPgTaskStoreHarness,
+} from "../../__test-utils__/pg-test-harness.js";
+import { aggregateTeamAnalytics } from "../../team-analytics.js";
+import { BUILTIN_CODING_WORKFLOW_IR } from "../../index.js";
+
+const IN_RANGE = "2026-06-15T12:00:00.000Z";
+const RANGE = { from: "2026-06-01T00:00:00.000Z", to: "2026-06-30T23:59:59.999Z", now: Date.parse(IN_RANGE) };
+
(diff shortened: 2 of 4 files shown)


gsxdsm:
**Confirmed as a real imprecision, and I am deliberately not "fixing" it here — the per-task alternative is worse for this caller, and saying which is which is the useful part.**

The rule I have been applying across this program: the project union is correct where over-inclusion is **invisible** — a query, a candidate filter, a read something else narrows afterwards — and wrong where over-inclusion causes an **action**. I have argued both sides of it today: the `awaitingPlanning` enrichment (#2842) is a safe union because a widened candidate set only costs an extra file read and the per-card predicate still answers per card; the glasses notifier (#2852) is an unsafe one, because a notification has no downstream narrowing and over-inclusion means a wrong notification reaching a user.

**This caller sits between them.** No action follows, so nothing is misrouted — but the output is a **number shown to an operator**, and a silently inflated count is not the same as a wasted row. The reviewer is right that a project whose workflows reuse `done` with different traits gets one workflow's semantics applied to every matching task.

**Why I am not converting it to per-task resolution.** This is a single SQL aggregate over the whole project. Per-task lane resolution means one workflow read per task and moving the aggregation out of PostgreSQL into JS — turning an indexed `COUNT ... WHERE column IN (...)` into an N-read loop on an analytics endpoint. That is a materially worse trade than the one it fixes, and it is the mistake the `awaitingPlanning` note records being reverted once already for exactly this reason.

**The shape that would actually fix it** is making a task's resolved lane roles *queryable* — a column or a joinable projection maintained alongside the task row — so the aggregate can filter on the role rather than on an id set assembled in application code. That is a schema change with an owner, not a batch conversion, and it would fix this class everywhere at once rather than one query at a time.

**What is true and worth recording in the meantime:** the union is exact for any project running a single workflow, which is the common case; it degrades only where two workflows disagree about the same column id. So the numbers are not unreliable in general — they are unreliable in a specific, nameable configuration, and that is a materially different claim from "analytics are wrong on renamed boards", which is what this PR fixed.

Leaving the thread for whoever owns the analytics schema question. If the preference is to accept the imprecision explicitly, that belongs as a note at the call site rather than as silence.

gsxdsm:
**Judged, decided, and recorded at the call site.** 17353c14c0 — resolving the thread on that basis.

The finding is real and I am **accepting the imprecision deliberately** rather than converting, so the reasoning now lives in the code instead of only in this thread:

- **Why it stays.** A union is safe where over-inclusion is invisible and unsafe where it drives an **action**. Nothing here is routed, moved, or notified — no card is mishandled. The cost is a number an operator reads, which is real but is not the same failure. Per-task resolution means one workflow read per task and moving the aggregation out of PostgreSQL into JS, turning an indexed `COUNT ... WHERE column IN (...)` into an N-read loop on an analytics endpoint. The board-load path reverted exactly that change once, for exactly that reason.

- **Scope, stated so the note is not read as worse than it is.** The union is **exact** for any project running a single workflow. It degrades only where two workflows disagree about the same column id. "Analytics are wrong on renamed boards" was the bug this PR *fixed*; what remains is narrower and nameable.

- **The real fix, named in the note so it is not rediscovered:** make a task's resolved lane roles **queryable** — a column or joinable projection beside the task row — so the aggregate filters on the role instead of an id set assembled in application code. That closes this class everywhere at once, and it is a schema change with an owner rather than a batch conversion.

An accepted trade-off with the argument written down is a different artifact from an unnoticed one, and the thread was right to force the choice.

Core `tsc` clean, `pnpm lint` clean, gate green (161/487/13/71).

PR #2866, opened 2026-07-30
fix(core): per-workflow metrics read zero on a renamed board

gsxdsm:
Second of the 14 lane-bound SQL sites from #2839, after #2864. Independent of it — different file, different caller argument.

## The defect

`aggregateWorkflowAnalytics` filtered in SQL on `t."column" = 'done'` and `IN ('in-progress','in-review')`. On a renamed board those match nothing, so `tasksCompleted`, `tasksInProgress` and `tasksInReview` come back **zero for every workflow** while the board is busy. Nothing errors.

Same shape and same fix as #2864: resolve per **project** via `resolveProjectColumnsForRoles`, bind an `IN` list, and thread the store from the single Command Center caller so the parameter has a supplier immediately rather than becoming an inert seam.

## What the test caught that I had not

**The renamed case still failed with the query fixed.**

The bucketing at lines 296–297 already uses `isWipColumnRole` / `isReviewColumnRole` — correctly converted — but those read `query.columnFlagsByName`, which production supplies and my fixture did not. So:

- the **SQL** decides *which rows come back*;
- the **trait map** decides *which bucket each row lands in*.

Both halves have to be right. Fixing only the query would have shipped a "conversion" that still reported zero on a renamed board, and the file would have scored as converted twice over. That is exactly the partial-conversion shape this program keeps re-finding — caught here only because the test asserts `tasksInReview` alongside `tasksCompleted`, since those two paths take **different** resolved sets (complete vs wip+human-review). Asserting the completed count alone would have left the second conversion unproven.

## Measured

Reverted, only the renamed case flips:

```
✓ default vocabulary: completed and in-review work are counted
× renamed vocabulary: completed and in-review work are counted
✓ renamed vocabulary: a card in the HOLD lane counts as neither
✓ without a lane store, the legacy ids still answer
```

The hold-lane negative is there so resolving real lanes cannot degrade into "every column counts" — trading an undercount for an overcount is harder to notice than the original bug.

## Scope

The sync SQLite arm in the same file keeps its literals: it throws in backend mode and has no production caller, the same dead-arm conclusion reached for `cleanupStaleMergeQueueRowsImpl` on #2839.

## Verification


gsxdsm:
**Both real. The second is fixed (004b913ca7); the first is real, feasible, and deliberately not folded into a review response.**

### Conflicting lanes lose active counts — fixed

This one is a **self-contradiction**, not an imprecision. `resolveProjectColumnsForRoles` unions every column any workflow gives the role, so an id two workflows reuse with different traits **stays in the SQL filter**. `columnFlagsByName` deliberately **drops** it as conflicting — its own note (#2803 review) argues a merged entry double-counts, and that argument is right.

Together they select rows and then classify them by nothing: the flags lookup misses, `isWipColumnRole` and `isReviewColumnRole` both fall back to the legacy ids, `checking` matches neither, and the count evaporates. **Worse than either half alone** — the filter says the lane counts, the classifier says it cannot say, and the operator sees a workflow at zero with no indication why.

Aligned on the classifier's answer, which is the conservative direction: a column it refuses to judge is not selected at all, rather than selected and discarded.

**This ships uncovered, and that is inherent rather than an omission.** Mutating the filter away leaves the renamed-lane suite green — the counts were already zero on both sides. The only observable difference is whether a workflow with no other rows appears with zeros or is absent, and a test pinning *that* would be asserting an artifact of `ensureSummary`, not the invariant. I would rather say so than add an assertion that looks like coverage.

### Project-wide lanes inflate completions — real, not fixed here

Confirmed, and **unlike `team-analytics` (#2864) this one is genuinely fixable**, which is why I am not accepting it the way I accepted that one. The difference is the `GROUP BY`: these rows carry a workflow id, so a per-**workflow** lane map is feasible — the union is being used where a keyed answer is available, which is the arity mistake rather than a cost trade-off.

I did not do it in a review-response commit because it changes this function's contract with its caller: the lane resolution has to be keyed by workflow rather than by project, which moves work across the `laneStore` boundary and wants its own change with its own tests. Folding a contract change into a thread reply is how the last four rounds on #2843 each produced a follow-up finding.

Recorded at the call site so it reads as known debt with a named fix, not as an oversight.

Core `tsc` clean, `pnpm lint` clean, `workflow-analytics-renamed-lanes.pg.test.ts` 4/4.

gsxdsm:

Issue #2862, opened 2026-07-30
Three named suite-only flakes (core + engine), with reproduction data


gsxdsm:
## Why this exists

Each of these fails in a **full-suite** run and passes when its file is run alone. They are not reproducible on demand, so each one costs whoever hits it a diagnosis — and today each was initially misread as "the PR under test broke something". Recording the identities so the next observer starts from evidence instead of a bisect.

Per AGENTS.md a flake is quarantined on sight, but quarantine is **file-level** and each of these files holds valuable passing tests, so quarantining would evict real coverage over a single observation. Recording rather than quarantining is the deliberate call; a second sighting of any one of them changes that.

---

### 1. `packages/core/src/__tests__/postgres/project-identity.test.ts`

```
project-identity async (PostgreSQL integration) > returns null when no identity is stored
```

| run | result |
|---|---|
| full core suite (1st) | **1 failed** / 4824 passed |
| full core suite (2nd) | 4825 passed |
| full core suite (3rd) | 4825 passed |
| file alone ×2 | 6 passed, 6 passed |

Observed on `origin/main` at `7927c7b58a`. **1 in 3** full-suite runs.

### 2. `packages/core/src/__tests__/postgres/schema-applier.test.ts`

```
schema-applier: VAL-SCHEMA-001 final-schema parity (table counts)
  > retains unreplaced registered dependents for every delete action
```

| run | result |
|---|---|
| full core suite on #2828 merged-with-main | **failed** |
| file alone ×2 on the same tree | 75 passed, 75 passed |
| file alone on `origin/main` | passed |

### 3. `packages/engine/src/__tests__/plugin-runner.test.ts`

```
PluginRunner > task lifecycle hooks
  > should invoke onTaskCompleted when the complete lane is RENAMED
```

| run | result |
|---|---|
| full engine suite on #2799 merged-with-main (1st) | **8 failed** (7 in this file + 1 inherited) |
| full engine suite, same tree (2nd) | 1 failed (the inherited one only) |
| file alone | 80 passed |
| full engine suite on `origin/main` ×2 | clean |

I can only name **one** of those 7 — the capture used `--reporter=dot | tail -3`, so the summary survived and the individual `FAIL` lines were truncated away.

---

## The common shape, and one concrete suspicion

All three are **PostgreSQL-backed or PG-suite-adjacent**, all pass in isolation, and all appear only under full-suite parallelism. That points at shared database state between test files rather than at any of the three tests.

[...]

PR #2872, opened 2026-07-30
fix(tests): stale-spec guard pinned first-per-role; #2820 replaced it with membership

gsxdsm:
## Red on main

```
executor-stale-spec-active-lanes > resolves the task's lifecycle columns before deciding the skip
executor-stale-spec-active-lanes > adds the wip, review and complete lanes to the active set
```

Both asserted the old implementation **verbatim**:

```ts
expect(source).toContain("const activeLifecycle = resolveLifecycleColumns(await resolveWorkflowIrForTask(this.store, task.id));");
expect(source).toContain("for (const lane of [activeLifecycle?.wip, activeLifecycle?.review, activeLifecycle?.complete]) {");
```

## The replacement is a fix, not churn

**#2820** swapped that for `columnsWithFlag`. `resolveLifecycleColumns` returns the **first** column carrying each trait, so a board with two wip lanes — or a review lane plus a second merge-blocking one — had only one of each recognised as active. A card sitting in the second read as **inactive**, and its prompt file was treated as reclaimable: precisely the mid-flight interruption this file exists to prevent.

Widening to every column carrying the trait is strictly the safe direction, and the code says so at the site.

## Re-pointed at membership, and scoped

`columnsWithFlag` has **four call sites** in `executor.ts`, so a file-wide `toContain` would pass regardless of what this particular guard does — a guard that cannot fail for its own subject.

That is the same vacuity the operator caught in my #2804 AST check (it collected every `columnsWithFlag` call file-wide, so two of six flags were unfalsifiable). Applying the lesson rather than repeating it: the assertions now run against a **slice bounded by this site's own markers**, with *both* markers proven unique first — a repeated marker silently widens the slice past the subject.

Also kept as a **source ratchet on purpose**. The file's header records why its behavioural cases were deleted: they rebuilt the union themselves, so reverting the product could not fail them (*"if a test constructs the value the product is supposed to construct, reverting the product cannot fail it"*). That reasoning still holds and this change does not relitigate it.

## Evidence

| mutation | result |
|---|---|
| drop `"humanReview"` from the union | **fails** — `the active set no longer unions the "humanReview" lanes` |
| revert to first-per-role `resolveLifecycleColumns` | **fails** — 1 failed |

Engine **11046 passed / 0 failed** · gate **732 green** · lint clean. Test-only; `executor.ts` restored clean.

## Not addressed here

The same main run had a third failure: `workflow-planning-continuation-terminal-gap-live-e2e.pg.test.ts`'s AUDIT case. It passes in isolation (4/4) and **did not recur** in the full run above — another suite-only intermittent. Added to #2862 with the other three named flakes rather than guessed at.

### DIFF
--- a/packages/engine/src/__tests__/executor-stale-spec-active-lanes.test.ts
+++ b/packages/engine/src/__tests__/executor-stale-spec-active-lanes.test.ts
@@ -43,17 +43,53 @@ import { readFileSync } from "node:fs";
 const source = readFileSync(new URL("../executor.ts", import.meta.url), "utf8");
 
 describe("the stale-spec skip resolves the board's own active lanes", () => {
-  it("resolves the task's lifecycle columns before deciding the skip", () => {
-    expect(source).toContain(
-      "const activeLifecycle = resolveLifecycleColumns(await resolveWorkflowIrForTask(this.store, task.id));",
-    );
+  /*
+  FNXC:WorkflowLifecycleColumns 2026-07-31-13:30:
+  RE-POINTED AT MEMBERSHIP, and SCOPED to this site.
+
+  The two cases here pinned the old implementation verbatim —
+  `resolveLifecycleColumns(...)` into `activeLifecycle?.wip / .review / .complete`. #2820 replaced
+  that with `columnsWithFlag`, and the replacement is a FIX, not churn: `resolveLifecycleColumns`
+  returns the FIRST column carrying each trait, so a board with two wip lanes (or a review lane plus a
+  second merge-blocking one) had only one of each recognised as active, and a card in the second read
+  as INACTIVE — its prompt file treated as reclaimable, which is the interruption this whole file
+  exists to prevent. Widening to every column carrying the trait is strictly the safe direction.
+
+  Asserted against a SLICE bounded by this site's own markers rather than the whole file.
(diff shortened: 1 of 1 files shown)


gsxdsm:
**Closing — this branch is now empty against `main`.** After rebasing, `git diff origin/main..HEAD` is zero bytes.

The ratchet in `executor-stale-spec-active-lanes.test.ts` pinned the guard's *old* shape (`resolveLifecycleColumns(...)?.wip / ?.review / ?.complete`, i.e. the FIRST column per trait). That went red when #2820 improved the guard to union `columnsWithFlag` over five flags — every column carrying each — which is the arity fix a `.has()` membership test actually needs.

This PR re-pointed the ratchet at the new shape. **Main landed an equivalent re-point in the meantime**, and the conflict was two versions of the same correction. I took main's.

Worth recording, because it is the same class this file exists to catch: **the ratchet went red on an improvement, not a regression.** That is the expected cost of a source-text ratchet and the reason it is valuable — but it means "red" here always needs a read before a fix, since loosening the assertion to something that matches both shapes would keep the file green through exactly the regression it guards.

`executor-stale-spec-active-lanes.test.ts` passes on main unchanged.

PR #2873, opened 2026-07-30
docs(lanes): audit the three files the census points at with no reason attached

gsxdsm:
**No source change.** Every literal stays counted and none gets an exemption marker. What changes is that the census now points at these three with the analysis attached, instead of making each worker who reaches them re-derive it.

Peers have already done this well for `notification-service.ts` — converted it, *measured* a real delivery regression, reverted, and left it counted with the reason. These three had nothing at all, and one of them is the highest-impact unowned site I found.

## `planner-overseer.ts` (3 guards) — REAL, and larger than three literals suggest

On a renamed board `resolveWatchedStage` returns `null` for every card. `observeTask` returns early on a null stage, so **no observation is recorded**, no `overseer:intervention` entry is emitted, and `PlannerRecoveryController` — which consumes those observations — has nothing to steer, retry, or targeted-fix. **The entire oversight loop is inert and silent about it**, exactly like the self-healing sweeps whose queries returned empty arrays.

Not mechanical, which is why it is flagged rather than converted. `resolveWatchedStage` is a pure sync function over a `Partial<OverseerTaskRef>` with no store and no task id, so the lane answer has to arrive as a parameter. Its only production caller, `observeTask`, *is* async and the monitor *does* hold a store — but it runs **once per task per poll**, so resolving inside it buys a workflow read per card on a timer.

The shape that works is the one the board-load enrichment landed on in #2845: resolve at the **poll**, once, with an IR cache keyed by workflow, and pass the flags down. That makes it a change to `project-engine.ts`'s poll as much as to this file — a cost judgement about a periodic engine loop, not a rename. `columnFlags` is in the unwired-lane-parameter vocabulary, so whoever adds the parameter cannot leave it unwired.

## `async-mission-store-queries.ts` (1 of 3) — REAL

`getTerminalTaskEvidence` tests only `column === "done"` for its `done` verdict, so a completed card on a renamed board falls through every branch to `{ kind: "nonterminal" }`. The caller is mission **terminal evidence repair**, so a finished feature reads as unfinished — a wrong *verdict*, not an error. The `archived` test beside it has the same defect, masked for soft-deleted rows by its `deletedAt` companion, which is why only the `done` half bites in practice.

Takes a bare `QueryHandle`: no store, no task object, no workflow. The fix is a resolved terminal-lane set threaded in by the caller — the same shape `getLiveTaskColumn` needs, and it should land *with* it so the two cannot disagree about what "finished" means.

## `audit-ops.ts` (2) — one sentinel, one real, and they look identical

```ts
if (state === "archived")           // ← getLiveTaskColumn's MANUFACTURED value: do NOT convert
if (pgRow.column === "archived")    // ← a real board lane: convertible
```

The first compares against a string `getLiveTaskColumn` *fabricates* for an archived-or-soft-deleted parent, so converting it to `isArchivedColumnRole` would keep passing on the built-in board and start **failing** on a renamed one — a soft-deleted task's log would become writable. The second reads the task row, so a renamed archived column keeps accepting log writes; its `deletedAt` companion masks that in practice.

Two lines that look the same and need opposite treatment is precisely the reason these notes are worth more than the count.

## Verification

- `pnpm lint` — clean
- `tsc --noEmit` (`@fusion/core`, `@fusion/engine`) — clean
- `planner-overseer.test.ts` — 50 passed
- census `--strict` — exit 0, **counts unchanged** (that is the point)

### DIFF
--- a/packages/core/src/async-mission-store-queries.ts
+++ b/packages/core/src/async-mission-store-queries.ts
@@ -2202,6 +2202,22 @@ export async function getTerminalTaskEvidence(handle: QueryHandle, taskId: strin
   const task = taskRows[0];
   const hasArchiveSnapshot = archiveRows.length > 0;
 
+  /*
+  FNXC:WorkflowLifecycleColumns 2026-07-30-21:50 (audited — REAL, deferred with the cost stated):
+  On a renamed board a genuinely completed task is classified `nonterminal`.
+
+  `task.column === "done"` is the only test for the `done` verdict, so a card in a complete lane
+  called anything else falls through every branch to `{ kind: "nonterminal" }`. The caller is mission
+  TERMINAL EVIDENCE repair, so a finished feature reads as unfinished — a wrong verdict rather than an
+  error, which is the shape this program keeps finding. The `archived` test below is the same defect;
+  its `deletedAt` companion masks it for soft-deleted rows, which is why only the `done` half bites in
+  practice.
+
+  Not converted here because `getTerminalTaskEvidence` takes a bare `QueryHandle`: no store, no task
+  object, no workflow, nothing to resolve a lane vocabulary from. The fix is a resolved terminal-lane
+  set threaded in by the caller — the same shape `getLiveTaskColumn` needs, and it should land with
+  it so the two cannot disagree about what "finished" means.
+  */
   if (!task) return hasArchiveSnapshot ? { kind: "invalid-deleted", id: taskId } : { kind: "missing" };
   if (task.deletedAt === null && task.column === "done") return { kind: "done", id: task.id, column: "done" };
   if (task.deletedAt !== null && task.column === "archived" && hasArchiveSnapshot) {
--- a/packages/core/src/task-store/audit-ops.ts
+++ b/packages/core/src/task-store/audit-ops.ts
@@ -124,6 +124,20 @@ export async function logEntryImpl(store: TaskStore, id: string, action: string,
         {
           const layer = store.asyncLayer!;
           const state = await getLiveTaskColumn(layer.db, id, layer.projectId);
+          /*
+          FNXC:WorkflowLifecycleColumns 2026-07-30-21:20 (audited — SENTINEL, do NOT convert):
+          `getLiveTaskColumn` MANUFACTURES the string "archived" for an archived-or-soft-deleted
+          parent; it does not return the board's archived lane. So this compares against that
+          function's return vocabulary, not against a column id, and converting it to
+          `isArchivedColumnRole` would keep passing on the built-in board and start FAILING on a
+          renamed one — a soft-deleted task's log would become writable.
+
+          The convertible site is `getLiveTaskColumn`'s own `row.column === "archived"` test, and it
+          is deferred with its cost stated: that helper takes a `db` handle with no task, no workflow
+          and no lane vocabulary, so resolving there threads a lane set through a low-level query on a
+          hot path. See docs/solutions/architecture-patterns/
+          the-archived-literal-is-usually-a-sentinel-not-a-column.md.
+          */
           if (state === "archived") throw new Error(`Task ${id} is archived — logging is read-only`);
           if (state === null) throw new Error(`Task ${id} not found`);
         }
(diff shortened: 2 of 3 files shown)


gsxdsm:
**Already fixed on this branch — `09a59ed82b` dropped that reference two commits ago.** The thread predates it.

The comment cited `docs/solutions/architecture-patterns/the-archived-literal-is-usually-a-sentinel-not-a-column.md`, which has not landed. Rather than write a doc to satisfy a cross-reference, the argument was inlined at the site, which is where it is actually needed: `getLiveTaskColumn` returns a normalized **sentinel** — `"archived"` also covers soft-deleted — so `state === "archived"` compares against that function's return vocabulary, not against a column id. Converting it to `isArchivedColumnRole` would keep passing on the built-in board and start **failing** on a renamed one, making a soft-deleted task's log writable.

Verified across the branch: every `docs/solutions/...` path referenced from `packages/**` resolves to a file that exists.

Worth recording as a rule, since I have now written two comments this session pointing at documents that do not exist: **a cross-reference to unlanded work is a dangling pointer with a comment's authority.** If the reasoning is needed to read the line, it belongs at the line.

gsxdsm:
Greptile P2 (broken architecture guide reference) — **agreed, fixed in 09a59ed.**

The path was real but unlanded: the doc lives on `handoff/archived-sentinel-literals-for-batch-core`, which had no PR, so the citation was dangling from the moment it was written. That is also the project's own spec-hygiene rule — do not cite a path that does not exist — applied to a source comment rather than a spec.

Replaced with the fact a reader actually needs at that line: the same sentinel/real distinction governs the eight downstream comparisons in `async-comments-attachments.ts`. Inline, verifiable, and not dependent on another branch landing.

The doc is now opened as **#2877** rather than cross-linked from here, so the two do not have to merge in order.

gsxdsm:
Adding a finding to this PR's audit rather than a sixth open branch, since it belongs with the other three notes here.

`getTerminalTaskEvidence` in `async-mission-store-queries.ts` — the site this PR documents — has a **louder** failure mode than the note currently states, and that is worth recording because it changes the priority.

Its only caller is `AsyncMissionStore.reconcileFeatureDoneWithTerminalTask`, which throws on a `nonterminal` verdict:

```
TASK_NOT_TERMINAL: Delivery task FN-X must be in done or supported archived state, not shipped
```

So on a renamed board, mission shipped-delivery repair does not silently misbehave — it **hard-fails, and the message names the actual column**. That is the least dangerous shape in the remaining set: an operator sees `not shipped` and can infer the lane mismatch, where the sites this program usually finds return a plausible number instead.

**Why it is still deferred here.** `AsyncMissionStore` holds an `AsyncDataLayer`, not a `TaskStore`, so `resolveProjectColumnsForRoles` (which needs `listWorkflowDefinitions`) has nothing to read from. Fixing it means either threading a store in or querying the workflow-definitions table directly from the layer — a shape nothing else in the codebase uses. That is a design decision for whoever owns the mission store, not a drive-by, and it is exactly the judgement the other two notes in this PR defer for the same reason.

The note in the diff has been left as-is; this comment is the addendum rather than another force-push, since the PR is already through review.

gsxdsm:
Follow-up: the `planner-overseer.ts` note this PR added is now **superseded by a fix** — #2898.

The note deferred it on the grounds that resolving inside `observeTask` "buys a workflow read per card per poll". That was wrong, and reading the caller is what showed it: `project-engine.ts`'s poll **already awaits `resolveEffectiveSettings` per task**, so it is a per-task async loop regardless. With an IR cache keyed by workflow, the addition is *(distinct workflows)* resolutions rather than *(cards)*.

I priced the fix before checking the caller. Worth recording as its own lesson, because "this needs a cost judgement" is the most comfortable place in this program to leave something — and it is the second time in this unit a deferral dissolved once I looked one level up (the other being `getInReviewDurationEvents`, deferred as core's raw-SQL problem and then fixed in #2875 once the impl turned out to hold the store).

The other two notes in this PR still stand: `getTerminalTaskEvidence` genuinely has no store on `AsyncMissionStore`, and `audit-ops.ts`'s pair splits into one sentinel and one real site that needs `getLiveTaskColumn` converted alongside it.

One thing #2898 found that the note did not: the review test has to be the **three-trait union**, not `isReviewColumnRole`. That helper checks only `mergeBlocker || humanReview`, so a board whose review lane carries `merge` (mergeOrchestration) — the built-in default's own shape — would classify as not-in-review and be skipped. Reaching for the obvious helper would have reintroduced the bug through the helper meant to fix it.

PR #2874, opened 2026-07-30
fix(census): the header claimed 0 trait-fallback branches while sites of that shape existed

gsxdsm:
The census header has been printing `of the column guards, 0 are trait-fallback branches (already converted)` while sites of exactly that shape exist. I flagged this on #2842 as a suspected classifier gap; this confirms and fixes it.

## The miss

Only `cond ? trait : literal` was recognised. The other spelling — a **negative** test with the literal on the **true** branch — is what a caller writes once it hoists its resolved lanes:

```ts
complete: completeLanes === undefined ? columnId === "done" : completeLanes.includes(columnId)
```

That is `github-tracking-state.ts:245-246` — a fully converted resolver whose two degraded arms were reported as unconverted debt. **The backlog read higher than the remaining work**, and a reader chasing it was sent to lines that are already correct.

Second half of the miss: `completeLanes` matches no hint. Adding `Lanes` to the hint list does **not** work, and the reason is itself a prior fix — hints are word-bounded because the unbounded form once let `hold` match `threshold` and `household`. `\bLanes\b` cannot match inside `completeLanes`, where the boundary does not exist. So resolved-lane identifiers get an explicit suffix rule.

## Both guards on the new rule exist because I broke them while writing it

Worth stating, because each failure ran in the **dangerous direction** — marking a *live* line "already converted", which removes a real guard from a backlog people trust:

| mistake | what it excused |
|---|---|
| widened the shared `testsTraitData` | fed the ancestor-walking rules too, which marked `step.status === "done" \|\| step.status === "in-progress"` at `register-task-workflow-routes.ts:941` — a step-**status** comparison, not a column guard — as converted |
| let the new rule walk ancestors | excused any literal inside a block governed by a negative lane test |

Measured: the count went to **6 with two of them wrong** before I caught it. The rule is now immediate-parent-only with its widened identifier match local to it, and reports exactly the **2 real sites**.

## Verification

- Census: **176 guards, 2 trait-fallback** (was 176 / 0). The total is unchanged — this sub-count is diagnostic and does not move the ratchet, so `--strict` exits 0 with no baseline re-record.
- 5 cases in `scripts/__tests__/lifecycle-census-inverted-fallback.test.mjs`, including both negatives that pin the mistakes above plus one for the suffix rule not over-reaching (`airplanes` is not a lane test).
- `pnpm lint` clean; gate green (161/487/13/71).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved lifecycle analysis accuracy for trait fallback logic, including inverted conditions, legacy fallback syntax, and null or undefined checks.
  * Added safeguards to avoid misclassifying complex conditions, unrelated identifiers, and nested expressions.
  * Improved handling of lifecycle lane and column naming patterns.

* **Tests**
  * Expanded coverage for valid and invalid fallback scenarios, identifier boundaries, parent-expression restrictions, and property-path checks.

### DIFF
--- a/scripts/__tests__/lifecycle-census-inverted-fallback.test.mjs
+++ b/scripts/__tests__/lifecycle-census-inverted-fallback.test.mjs
@@ -66,3 +66,31 @@ test("an identifier merely CONTAINING a lane word is not a lane test", () => {
   const src = 'const c = airplanes === undefined ? columnId === "done" : other;';
   assert.deepEqual(fallbacks(src), []);
 });
+
+/*
+FNXC:LifecycleColumnCensus 2026-07-30-23:20 (#2874 review — greptile P2):
+A COMPOUND condition is not a simple absence test. `completeLanes === undefined || forceLegacy` has a
+second disjunct that can select the true branch with lane data PRESENT, so the literal there is a live
+guard. The unanchored check accepted it, which removes a real guard from the backlog — the direction
+this rule exists to stop.
+*/
+
+test("a compound absence test is NOT treated as a fallback", () => {
+  const src = 'const c = completeLanes === undefined || forceLegacy ? columnId === "done" : other;';
+  assert.deepEqual(fallbacks(src), []);
+});
+
+test("a conjunction is refused too", () => {
+  const src = 'const c = completeLanes === undefined && legacyMode ? columnId === "done" : other;';
+  assert.deepEqual(fallbacks(src), []);
+});
+
+test("the simple negated form still counts", () => {
+  const src = 'const c = !completeLanes ? columnId === "done" : completeLanes.includes(columnId);';
+  assert.deepEqual(fallbacks(src), ["done"]);
+});
+
+test("a property-path absence test still counts", () => {
+  const src = 'const c = lifecycle?.completeLanes === undefined ? columnId === "done" : other;';
+  assert.deepEqual(fallbacks(src), ["done"]);
+});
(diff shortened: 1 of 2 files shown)


gsxdsm:
**Correct, and it runs in the direction this rule exists to avoid.** c8a28a0767.

The equality check was unanchored, so `completeLanes === undefined || forceLegacy` passed — and its second disjunct can select the true branch **with lane data present**, making the literal a live guard rather than a fallback. Marking a live line "already converted" removes a real guard from the backlog, which is exactly the mistake this rule was added to stop committing. That is now the third instance of it in this one change, after the two I caught myself.

Anchored to a **simple** comparison. A compound condition is not something to reason about here — whether the literal is reachable with traits present depends on the other operand — so refusing them leaves those sites **counted**, which is the safe answer for a measurement: over-counting sends a reader to a correct line, under-counting hides a live one.

**I also wrote a redundant guard and removed it.** My first fix added an explicit `if (/[|&]{2}/) return false`, which *looked* like the fix. Mutation showed it was dead code: the anchors already refuse anything with an operand beside the comparison. A redundant guard carrying a comment claiming it is load-bearing is worse than no guard, because the next reader trusts it instead of the anchors — the same failure as the `!`/`?` comment on #2841 that described behaviour the regex did not implement.

**Mutation-verified both ways:** unanchoring fails 2 of 9; the redundant check fails 0, which is how I knew to delete it.

9 cases; census still reports 176 guards / 2 trait-fallback; `pnpm lint` clean, `--strict` exits 0.

PR #2875, opened 2026-07-30
fix(reliability): the duration query hid its lane ids inside a SQL template

gsxdsm:
The Reliability panel's **third and last** blind input — and my own loose end. #2861 fixed the two counts beside it, so the panel went from uniformly wrong to **partially** wrong: entries and bounces populated, duration reporting `no-in-review-entries` forever. Partial blindness is harder to notice than total, which is why finishing it matters more than one site suggests.

```sql
metadata->>'to' = 'in-review'
  OR (metadata->>'from' = 'in-review' AND metadata->>'to' = 'done')
```

## The class, not just the site

**This shape is invisible to every check we have.** The lifecycle census scans `===`/`!==` comparisons; the unwired-lane-parameter guard scans declarations. Neither sees a lane id inside a `sql` template, so this class is **not in the backlog total at all** — the number is a floor for this reason as well as the usual one.

`scripts/check-sql-column-literals.mjs` (#2841, in flight) is the detector for exactly this: it freezes the surface at 30 sites rather than converting any, so this one was unowned. That PR and this one are complementary — it stops the surface growing, this shrinks it by one.

## The fix

Lanes resolve **once per call** via `resolveProjectColumnsForRoles` and arrive as parameterised equality fragments, one branch per id — no interpolated list, no string building.

Resolution lives in `getInReviewDurationEventsImpl` because that is where the store is; `async-audit.ts` takes a bare `db` handle and cannot resolve anything. Best-effort, defaulting to the legacy pair, so a caller that cannot resolve keeps exactly today's query.

**The union is correct rather than a widening hack**, for the same reason as #2861: these are *move records*, and a past move recorded the column name as it was at the time. A board renamed last month has rows under both ids, so the honest query covers both — which is precisely what `resolveProjectColumnsForRoles` returns.

## Tested against real PostgreSQL, deliberately

This is a **SQL predicate** change. A mocked store would assert the arguments and prove nothing about the query that actually runs — which is the entire risk when the literal lives inside `sql`. The new case inserts real `activity_log` rows on a renamed board and reads them back through the real store method.

The legacy-lane case in the same file stays green, which is the compatibility half.

**Revert proof, measured:** restore the hardcoded fragments and the new case fails with

```
expected [] to deeply equal [ 'renamed-entered', 'renamed-done' ]
```

## Verification

- `pnpm lint` — clean
- `activity-log-parity.pg.test.ts` — 5 passed against real PostgreSQL

With this, all three Reliability inputs read the board's own lanes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reliability duration metrics now work correctly with renamed workflow lanes.
  * Completion tracking recognizes configured completion lanes instead of relying on fixed defaults.
  * Improved handling of transitions between multiple review lanes and review-to-work-in-progress movements.
  * Legacy lane behavior remains supported when configured lane information is unavailable.

* **Tests**
  * Added coverage for renamed lanes, historical lane IDs, and transition edge cases.


gsxdsm:
**Confirmed and fixed.** faee4fdb46 — a half-converted pair: the producer was converted and the consumer threw the answer away.

`getInReviewDurationEvents` fetches moves using the project's **resolved** review lanes; `inReviewDurationMetrics` then matched `to === "in-review"` and `to === "done"` against them. On a renamed board every fetched event was discarded, the sample count never reached three, and the Reliability panel reported `insufficient-samples`. **Silently absent rather than visibly wrong**, which is why it outlived the conversion that produced the events.

The lane sets are optional and the **production caller fills them** — `server.ts` already resolves `reviewLanes` two statements above for the entry/bounce counts, so the review half costs no extra read. That check matters: an optional parameter nobody fills is the inert-injection shape this program keeps re-finding, and I verified the caller could supply before adding the option.

**The negative had to be rewritten before it bit, and the reason generalises.** My first version used `drafting → building` noise and stayed **green** when I widened the exit predicate to "any move in range". A card that never *enters* review records no start, so every exit predicate yields zero there — the test was passing for a reason unrelated to what it asserted. Entering `checking` and bouncing back to `building` is the shape that distinguishes them: a start *is* recorded, so only a correct complete-lane check keeps it out of the durations.

**Mutation-verified both directions:**

| mutation | result |
|---|---|
| restore the `in-review`/`done` literals | 1 failed / 20 passed |
| widen the exit to any move in range | 1 failed / 20 passed |
| restored | 21 passed |

Dashboard `tsc` clean, `pnpm lint` clean.

gsxdsm:
Greptile P1 (resolved lanes discarded downstream) — **agreed, and it caught precisely the mistake this program keeps writing docs about.**

The query now fetches a renamed board's move rows, and `inReviewDurationMetrics` then re-tested `to === "in-review"` and `from === "in-review" && to === "done"` on the way in, discarding every one. **A converted producer feeding an unconverted consumer is not a converted path** — the metric would have kept reporting `insufficient-samples` for the same reason it reported `no-in-review-entries` before. That is the "a file at zero guards is not a converted file" finding, in my own change, one file downstream of where I was looking.

A peer had already landed the fix on this branch (faee4fd) and did it better than my draft: they reuse the `reviewLanes` `server.ts` already resolves two statements above for `countEntriesInto`/`countBouncesOut`, so the wiring costs no extra read.

What I added on top (8b59f58) is the thing that fix left behind: the function's **`DELIBERATE-LITERAL` marker survived directly above the new note**, so the file claimed both that the literals were a reviewed decision and that they were a defect. Not cosmetic — that marker is what the lifecycle census reads to **exempt** a site, so it would have told every future audit that a converted function was closed. The census confirms it was live: removing it tightens the baseline by three (`in-review` 4 → 2, `done` 1 → 0).

The two identical markers above it stay. They are correct for their own functions, which this route does not call and which genuinely match historical log values as recorded.

**Separately, and more urgently: main's gate is red** and this PR is blocked behind it. `check-sql-column-literals` landed in the gate with #2841, then #2864 legitimately converted `team-analytics.ts` 6 → 3 without re-recording the baseline; the gate fails on a downward move too. Fixed in **#2881** (baseline 30 → 28, no source change). I deliberately left the SQL baseline out of this branch so the two do not both edit the same line.

Re-verified here: `reliability-metrics.test.ts` 21 passed, `tsc --noEmit` clean, `pnpm lint` clean.

gsxdsm:

gsxdsm:
**Confirmed and fixed.** db7f7b0117.

A board may declare several review-role lanes — merge orchestration beside a human sign-off lane — and a card moving **between** them has not re-entered review. Overwriting the entry timestamp on every move whose *destination* is a review lane made the metric measure only the **last** lane, so the reported duration shrank exactly on the boards that review most carefully.

It read plausible the whole time, which is the recurring property of these: a number that is wrong but not absurd survives every check except someone reasoning about the transition.

**One judgement call, stated:** an unknown `from` (absent metadata) still records the entry. A first observation with no prior lane *is* an entry as far as this data can tell, and dropping it would lose the sample entirely — worse than dating it slightly late for a metric that already needs three samples before it reports anything.

**Mutation-verified:** restoring the unconditional overwrite fails 1 of 22. The new case is three cards spending 30 minutes in `checking` and 30 in `signoff` before completing — honest duration 60 minutes, defect reports 30.

That is the second real defect this file has yielded in two rounds (the first being the consumer discarding resolved events), both of which only appear on boards more elaborate than the built-in one. Worth noting for whoever reviews the rest of the reliability metrics: multi-lane boards are the shape that breaks them, not renamed single lanes.

Dashboard `tsc` clean, `pnpm lint` clean, 22/22.

gsxdsm:
CodeRabbit (inconsistent `projectId` derivation) — **agreed, fixed in ac0923c**, and it is the exact failure that bit me one PR later.

The new case used `store.getAsyncLayer()!.projectId!` while everything else in the file uses `h.layer().projectId ?? "__legacy_unscoped__"`. The tasks and activity tables partition on that sentinel when the layer carries no project id, so the `!` form writes `projectId: undefined` on an unscoped harness and the query then matches nothing.

I hit precisely this in #2886's fixture:

```
UNDEFINED_VALUE ... args: [ 'archived', '2026-07-30T...', undefined, 'KB-001' ]
```

and fixed it there. **Here it happened to work, which is worse** — an assertion that passes because the harness currently supplies a project id is one that fails the day it does not, in a test whose entire job is to prove a query matches. That is the same shape as a revert proof passing vacuously, which this PR series has now hit twice.

Not touched: `startup-factory-integration.test.ts` uses the same form, but it constructs its stores explicitly rather than taking the shared harness, so the id is not optional there. Different file, different guarantee — worth saying rather than sweeping it in for consistency's sake.

Re-verified: `activity-log-parity.pg.test.ts` 5 passed against real PostgreSQL, `tsc --noEmit` clean, `pnpm lint` clean.

PR #2877, opened 2026-07-30
docs(workflow-learnings): two lane-literal classes no tool of ours can see

gsxdsm:
Docs only. Two findings from this unit that cost real time to derive and would otherwise be re-derived by whoever reaches these files next.

## 1. `=== "archived"` is usually a SENTINEL

`packages/core/src/task-store/async-comments-attachments.ts` carries **9** census guards — the second-largest single-file count outside `self-healing.ts`. Reading all nine: **exactly one** is a board-column comparison. The other eight compare against a value `getLiveTaskColumn` *manufactures*:

```ts
if (row.column === "archived" || row.deletedAt != null) return "archived";  // ← fabricated
return row.column;
```

Converting those eight to `isArchivedColumnRole` would keep passing on the built-in board and start **failing** on a renamed one — a soft-deleted parent's documents would become readable. **The conversion makes the renamed board worse**, which is the opposite of what the census count implies.

The rule that separates them: look at where the compared value *came from*, not at its type. From `task.column` or a DB field → a board lane. From a function that *returns* `"archived"` as a documented outcome → a sentinel.

Consequence worth stating plainly: **a file's census count is an upper bound on convertible sites, not a work estimate.**

## 2. Lane literals inside raw `sql` are in no total at all

The Reliability panel had three inputs. Two were call arguments and converted routinely (#2861). The third encoded its lanes in a `sql` fragment:

```sql
metadata->>'to' = 'in-review' OR (metadata->>'from' = 'in-review' AND metadata->>'to' = 'done')
```

The census scans `===`/`!==` comparisons; the unwired-lane-parameter guard scans declarations. **Neither can see a string inside a `sql` template**, so this class is not in the backlog number — a second, independent reason the total is a floor. Second known instance after the archived gate in PR #2724, which makes it a pattern rather than an accident.

Fixed in #2875, and the doc says so rather than leaving it described as outstanding — a learnings doc that reports a fixed defect as open sends the next reader to a dead end. `scripts/check-sql-column-literals.mjs` (#2841) is the detector for the class and freezes the surface at 30 sites; the two are complementary.

## 3. Sibling files

The GitLab importer's `column: "triage"` was fixed in #2843. The Linear importer — written from the same template, with **two tests pinning the bug** — still had it, and was found only by re-grepping an area I had already declared clean (#2860). When a defect is found in a file that has a sibling, the sibling is the next place to look, and no tool will tell you that.

## Verification

`pnpm lint` clean. No source change.

### DIFF
--- a/docs/solutions/architecture-patterns/the-archived-literal-is-usually-a-sentinel-not-a-column.md
+++ b/docs/solutions/architecture-patterns/the-archived-literal-is-usually-a-sentinel-not-a-column.md
@@ -112,7 +112,8 @@ inside a `projectId`-scoped query. **Done in #2875** — the lanes resolve in th
 store; the query function takes a bare `db` handle) and arrive as parameterised equality fragments,
 tested against real PostgreSQL because a mocked store would assert the arguments and prove nothing
 about the query that runs. `scripts/check-sql-column-literals.mjs` (#2841) is the detector for the
-class and freezes the surface at 30 sites; the two are complementary.
+class and freezes the surface at 28 sites; the two are complementary. (That number moves as sites are
+converted and the baseline is re-recorded — the gate fails on a DROP too, so it cannot drift silently.)
 
 Sibling importers are the other half of the same lesson: the GitLab importer's `column: "triage"` was
 fixed in #2843 and the Linear importer — written from the same template — still had it, found only by
--- a/docs/solutions/architecture-patterns/the-archived-literal-is-usually-a-sentinel-not-a-column.md
+++ b/docs/solutions/architecture-patterns/the-archived-literal-is-usually-a-sentinel-not-a-column.md
@@ -95,7 +95,11 @@ enough to call it a pattern rather than an accident, and enough to say plainly t
 total does not include this class at all**, so it is a floor in a second, independent way.
 
 Fixing this one is not a rename: it means threading a resolved lane set into a raw `sql` fragment
-inside a `projectId`-scoped query. Recorded here rather than attempted from `batch-cli-plugins`.
+inside a `projectId`-scoped query. **Done in #2875** — the lanes resolve in the impl (which holds the
+store; the query function takes a bare `db` handle) and arrive as parameterised equality fragments,
+tested against real PostgreSQL because a mocked store would assert the arguments and prove nothing
+about the query that runs. `scripts/check-sql-column-literals.mjs` (#2841) is the detector for the
+class and freezes the surface at 30 sites; the two are complementary.
 
 Sibling importers are the other half of the same lesson: the GitLab importer's `column: "triage"` was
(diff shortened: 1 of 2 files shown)


gsxdsm:
Two P2s. **The first is right and it is the worst kind of docs defect** — fixed in a2f5f52.

The table said "8 of 9 must not be converted" while the rule four paragraphs below it says *look at where the compared value came from*. Lines 213/332/345 read `task.column` off a row `select`, so they are board lanes by exactly that test, and I filed them as sentinels anyway. **The document would have told the next reader that three genuine defects were intentional** — strictly worse than having no document, which is the one outcome a learnings doc must avoid.

The real split is **5 sentinels, 4 convertible**, and the consequence is now stated because it is the whole point: on a renamed archived lane, line 213 keeps a card's documents **writable**, and 332/345 reject a legitimate archived-document publication as `parent-not-archived` / `archived-state-inconsistent`.

The headline finding survives and is sharper for the correction — a file's census count is an upper bound on convertible sites, not a work estimate, and this one file holds **both** classes spelled identically. That, rather than the ratio, is the lesson.

**The second P2 is stale rather than wrong.** `scripts/check-sql-column-literals.mjs` genuinely did not exist when that passage was written — it landed with #2841 while this branch was open. Verified present after merging main into the branch, and it is now in the merge gate. (Related: that gate went red on main because #2864's legitimate conversion tightened its baseline without re-recording it; #2881 fixes that.)

`pnpm lint` clean. Docs only, no source change.

gsxdsm:
**Both addressed.** dcce963d20 plus an earlier commit on this branch.

**Sentinel count** — real, and corrected before this thread: lines 213/332/345 read `task.column` off a row `select` and **are** board lanes, not sentinels. Calling them sentinels would have marked three genuine defects as intentional, which is worse than leaving them uncounted. The table now separates line 142's normalized return value (a sentinel — `"archived"` also covers soft-deleted) from the three direct column comparisons, and states the cost on a renamed board: 213 keeps a card's documents writable, 332/345 reject a legitimate archived-document addition.

**Detector reference** — not missing, just ahead of this branch. `scripts/check-sql-column-literals.mjs` landed on main with #2841 (`ef50244234`); rebasing brings it in and the citation resolves. Verified by running it.

**The rebase also turned the gate red, and that is the gate working**, so recording it: `team-analytics.ts` went 6 → 3 as sites were converted on main, and a **drop** fails deliberately — a migrated site must not leave a baseline slot the surface can silently regrow into. Re-recorded 31 → 28.

**The doc's "30 sites" was already stale when written**, which is the more useful lesson: a fixed number in prose is a claim that rots between the commit and the reader. It now cites the current 28 *and* says the number moves as sites convert, with the reason the gate catches drift in both directions — so the next reader checks the tool rather than trusting the sentence.

`pnpm lint` clean; detector reports 28 known sites, none added.

gsxdsm:

PR #2881, opened 2026-07-30
fix(gate): main is red — re-record the SQL baseline #2864 tightened

gsxdsm:
**Main's merge gate is currently red, and every PR in the queue is blocked behind it.**

`check-sql-column-literals` entered the gate with #2841. #2864 then legitimately converted `team-analytics.ts` from 6 SQL column literals to 3 — a real improvement — without re-recording the baseline. The gate fails on a count going **down** as well as up, so:

```
[check-sql-column-literals] SQL column-literal population changed:
```

Baseline 30 → 28. No source change. The tool's own message says *"If a count went DOWN, re-record the baseline in the same commit"* — this is that, one commit late.

Verified: `node scripts/check-sql-column-literals.mjs` exits 0, and the full `pnpm test:gate` passes 161 / 487 / 13 / 71.

## Deliberately not changing the tool

Failing on a downward move is correct — it is what makes the surface a **ratchet** rather than a high-water mark, and the miss here was procedural rather than a design flaw.

Worth flagging to whoever owns it, though, that the lifecycle census solves the same problem differently: it **rewrites the baseline downward itself** and prints *"COMMIT IT so the allowance cannot be regrown into"*, so a legitimate conversion cannot red main by forgetting a flag. With a fleet actively converting lane literals, this will recur — I hit it myself in the next PR up (#2875 removes a `done` literal from `async-audit.ts`). That is a behaviour change to a gate that landed hours ago, so it belongs to its author, not to a drive-by fix that is trying to unblock the queue.

## Summary by CodeRabbit

* **Chores**
  * Updated an internal analytics baseline to reflect the latest expected results.

### DIFF
--- a/scripts/lib/sql-column-literals-baseline.json
+++ b/scripts/lib/sql-column-literals-baseline.json
@@ -11,6 +11,6 @@
   "packages/core/src/task-store/reads.ts": 1,
   "packages/core/src/task-store/task-artifacts-ops.ts": 1,
   "packages/core/src/task-store/workflow-definitions.ts": 2,
-  "packages/core/src/team-analytics.ts": 6,
+  "packages/core/src/team-analytics.ts": 3,
   "packages/core/src/workflow-analytics.ts": 6
 }


gsxdsm:
**Merge-order note, with the timeline, because this one gates the others.**

```
23:06  #2841 merged — check-sql-column-literals enters the merge gate
23:09  #2864 merged — team-analytics 6 → 3, baseline not re-recorded  ← main goes red here
23:20  #2843 merged with Gate:SUCCESS — its gate run had started before 23:09
```

Verified against current `origin/main`: `node scripts/check-sql-column-literals.mjs` exits **1**, and with this branch merged in it exits **0** (`28 known SQL column literal(s), none added`).

So any PR whose Gate job runs on the merge result from 23:09 onward inherits the failure. That is every PR queued behind this one, mine included — which is why this is worth merging ahead of them rather than in queue order. It is a one-line baseline re-record with no source change, so there is nothing in it to review beyond the number.

I deliberately kept this baseline out of my other four branches so they would not all edit the same line and conflict with each other.

**Recurrence, for whoever owns the gate.** The tool hard-fails on a downward move and tells you to pass `--update-baseline`. That is a correct ratchet, and it means every legitimate lane conversion in a SQL-touching file reds main unless the author remembers the flag — with a fleet converting lane literals right now, #2864 will not be the last. The lifecycle census solved the identical problem differently: it rewrites its baseline downward itself and prints *"COMMIT IT so the allowance cannot be regrown into"*, which keeps the ratchet while removing the footgun. That is a behaviour change to a gate that landed an hour ago, so it belongs to its author rather than to this unblock.

gsxdsm:
**Correcting my own merge-order note above — the mechanism is narrower than I stated, and the conclusion still holds.**

I wrote "any PR whose Gate job runs from 23:09 onward inherits the failure." That is wrong as stated: #2873's Gate ran 23:30–23:36, well after main went red, and passed. Its branch does **not** contain #2864 (`git merge-base --is-ancestor` says no), and its merge ref evidently didn't either.

The accurate rule is **any PR whose tree contains #2864** — which is every branch cut after 23:09, so effectively everything from here on. Verified for my own open set:

```
archived-doc-lanes      contains #2864: YES   → Gate will fail
reliability-duration-lanes  contains #2864: YES   → Gate will fail
doc-land                contains #2864: YES   → Gate will fail
sql-baseline-retighten  contains #2864: YES   → passes, it is the fix
lane-audit-notes        contains #2864: no    → passed at 23:36
```

So **#2875, #2877 and #2886 are expected to show a red Gate**, and it is this baseline and not their own content. Flagging the prediction ahead of the runs so they are not read as broken work; once this lands and they are rebased, they go green without changes.

The underlying fact is independently verified and does not depend on any of the above: on a **clean worktree of `origin/main`** (`git worktree add --detach`, no local modifications), `node scripts/check-sql-column-literals.mjs` exits **1**; with this branch merged it exits **0**. The check scans the whole tree with no diff-scoping, so it is not something a PR can dodge by not touching the file.

gsxdsm:
**Superseding my own merge-order note: this PR no longer gates anything.**

I originally kept the baseline re-record here alone, to avoid several branches editing the same line. That was the wrong call — it left #2875, #2877 and #2886 red for a reason unrelated to their content, and made them depend on merge order.

Identical changes to the same line merge without conflict (verified with a trial merge of this branch into `archived-doc-lanes`: *"Automatic merge went well"*), so the same one-line re-record now lives in each affected branch. All three are green on the check as of this comment. Whichever lands first turns the rest into a no-op, and this PR may well end up empty — close it rather than merge it if so.

The finding it exists to record still stands and is the part worth keeping:

**The gate hard-fails on a downward move and requires `--update-baseline` by hand.** That is a correct ratchet — a migrated site that leaves its baseline entry behind is a slot the surface can silently regrow into. But it means every legitimate lane conversion in a SQL-touching file reds main unless the author remembers the flag, and with a fleet converting lane literals right now, #2864 will not be the last. The lifecycle census solves the identical problem without the footgun: it rewrites its own baseline downward and prints *"COMMIT IT so the allowance cannot be regrown into"*, keeping the ratchet while making the failure self-healing. Worth considering for whoever owns this gate.

gsxdsm:
Superseded by #2878, which landed the same one-line re-record first. Closing rather than merging an empty diff.

The finding it carried is worth keeping visible, since #2870 and #2871 have already tightened the baseline again since: **this gate hard-fails on a downward move and needs `--update-baseline` by hand**, so every legitimate lane conversion in a SQL-touching file reds main unless the author remembers the flag. That is a correct ratchet with a footgun attached, and with a fleet converting right now it has already fired twice. The lifecycle census solves the identical problem by rewriting its own baseline and printing "COMMIT IT so the allowance cannot be regrown into" — same ratchet, self-healing failure. Worth considering for whoever owns this gate.

Issue #2784, opened 2026-07-30
Dashboard: 123 test failures hidden behind the quality-runner's fail-fast (measured, per-lane)


gsxdsm:
## Summary

`pnpm --filter @fusion/dashboard test` stops after the first failing lane (`skipped 9 lane(s) after first failure`), so most of the package has not actually been running. Once `app:components-b` was fixed (#2782), the lanes behind it became reachable.

I ran **every lane individually against `origin/main`**. Result: **123 failures** that the runner was never reaching.

## Measured, per lane

| lane | result |
|---|---|
| `app:foundation-api` | 455 passed |
| `app:foundation-ui` | 484 passed |
| `app:foundation-hooks-utils` | 1286 passed |
| `app:components-a` | 1195 passed |
| `app:components-b` | 1688 passed *(was 2 failed — fixed in #2782)* |
| `app:app` | **10 failed** / 131 passed |
| `app:chat` | 386 passed — but see the flake below |
| `app:settings` | 352 passed |
| `app:backfill 1/4` | **6 failed** / 1972 passed |
| `app:backfill 2/4` | **22 failed** / 1916 passed |
| `app:backfill 3/4` | **55 failed** / 5073 passed |
| `app:backfill 4/4` | **29 failed** / 2057 passed |
| `api:curated` | 1599 passed |
| `api:backfill 1/2` | **1 failed** / 1482 passed |
| `api:backfill 2/2` | 1168 passed |

**Total: 123 failures.** The `app:backfill` shards alone account for 112.

## These are not new, and not one root cause

`app:app` has been red at an identical **10 failed / 131 passed** at `origin/main`, `main~5`, `main~20`, `main~60` and `main~150` — it is long-standing, not recent fallout.

I checked whether the 123 share a cause. **They do not** — at least two distinct families:

- **Renders nothing.** `app:app` and `list-view-windowing.test.tsx` fail on missing elements: the App shell never leaves the Loading state (`Unable to find an element with the text: Task Alpha` / `+ New Task`), and `ListView` renders an empty window (`expected [] to have a length of 50`). No console error or unhandled rejection is emitted — a silent load stall.
- **CSS-content assertions.** e.g. `core-modals-mobile.test.tsx` → `expected '\n width: 100%;\n height: 100%;…' to contain 'height: 85vh;'`.

[...]

PR #2885, opened 2026-07-30
fix(tests): TaskDetailModal renders through a PORTAL — query the document, not container

gsxdsm:
## What this clears

**30 of the 55 failures** in the dashboard `app:backfill 3/4` shard — all in one file, all reading `expected null to be truthy`.

## The symptom points the wrong way

That message reads as *"the modal never rendered"*, and that is how this survived. The file's helpers took the `container` returned by `render()` and asked it for the modal's elements:

```ts
const header = container.querySelector("[data-testid='agent-log-model-header']");
```

`TaskDetailModal` mounts inside `FloatingWindow`, which uses **`createPortal`** — so the modal subtree is attached to `document.body`, **not** beneath the container React handed back. Every `container.querySelector` in the file returns null no matter what renders.

## Probed, not inferred

I had already spent one wrong hypothesis on this exact file — the shared `TaskDetailModal.test-helpers.ts` carries a genuinely stale `{ flagEnabled: false, workflows: [] }` fixture of the kind #2833 fixed for `App.test.tsx`, so it looked like the 30-failure lever. Adopting `DEFAULT_BOARD_WORKFLOWS` **changed nothing** (still 30 failed). Reverted.

So I instrumented instead:

```
P1_after_tab_click  menu=true items=["Live","Feed","Raw","Interventions"]
P2_after_select     viewer=true header=true empty=false
```

The Activity menu opens, `Raw` selects, and the viewer **and** its model header are both present — via `document`. Only the container-rooted lookup could not see them.

**Why the file half-worked:** `screen.getByRole(...)` in the same helpers always succeeded, because `screen` queries the document. That mix of query roots is what made a query-root bug look like a rendering fault.

19 `container.querySelector` call sites converted.

## Scope — deliberately narrow

**Only this file.** Eight other `TaskDetailModal` specs use `container.querySelector` too — 95 of them in `attachments-and-tabs` alone — and they **all pass today**, because they render `TaskDetailContent` rather than the portalled modal. Converting green files would be churn with real risk and no red to justify it.

## Evidence

| | result |
|---|---|
| the file | **48/48** (was 30 failed) |
| shard `3/4` | **55 → 25** failures |
| mutation: rename the `agent-log-model-header` testid in `AgentLogViewer` | **18 failed** |

The mutation matters here: the fix is "change what we query", so the risk is assertions that now find *something* and stop being load-bearing. They still observe the real component.

`pnpm lint` clean. Test-only; `AgentLogViewer.tsx` restored clean.

## Remaining in this shard

`settings-mobile` (17), `NodesView` (2), `MailboxModal` (2), `agent-modals-mobile` (2), `TaskDetailModal.pr-tab` (1), `onboarding-flow` (1). Tracked on #2784, which I have been keeping current with per-lane numbers.

### DIFF
--- a/packages/dashboard/app/components/__tests__/TaskDetailModal.models-progress-workflow.test.tsx
+++ b/packages/dashboard/app/components/__tests__/TaskDetailModal.models-progress-workflow.test.tsx
@@ -114,12 +114,34 @@ describe("TaskDetailModal", () => {
       );
     }
 
-    async function openAgentLogAndExpandModelDetails(container: HTMLElement) {
+    /*
+    FNXC:TaskDetailModalTests 2026-07-31-15:10:
+    QUERY THE DOCUMENT — TaskDetailModal renders through a portal, so `container` is the wrong root.
+
+    This helper took the `container` from `render()` and asked it for
+    `[data-testid='agent-log-model-header']`. TaskDetailModal mounts inside `FloatingWindow`, which
+    uses `createPortal`, so the modal subtree is attached to `document.body` and NOT beneath the
+    container React handed back. Every `document.querySelector` in this file therefore returns null
+    no matter what renders.
+
+    Probed rather than inferred, because the symptom pointed the wrong way — 30 cases failing on
+    "expected null to be truthy" reads as "the modal never rendered":
+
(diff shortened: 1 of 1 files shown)


gsxdsm:

```
```

PR #2886, opened 2026-07-30
fix(core): the archived-document guards failed in OPPOSITE directions on a renamed lane

gsxdsm:
Two of the four convertible sites my own learnings doc **miscounted as sentinels** — the #2877 review corrected "8 of 9 must not be converted" to "5 of 9", and these are two of the three that correction freed.

They read `task.column` straight off a row `select`, so they are board lanes by exactly the test that document gives, and a renamed archived column is simply not seen. What makes the pair worth fixing together is that they fail in **opposite directions**:

| guard | on a renamed archived lane | consequence |
|---|---|---|
| `upsertTaskDocument` | fails to **reject** | an archived card's documents stay **writable** — the read-only contract silently does not hold |
| `publishArchivedTaskDocumentAddition` | fails to **accept** | a legitimate archived-document publication is refused as `parent-not-archived` |

The second is the sharper one: valid operator work refused, and refused with a message that reads as a data-integrity error rather than a lifecycle mismatch.

## Shape

Both take an `AsyncDataLayer` and can resolve nothing themselves; their store-level impls hold the store, so the lane set arrives as a parameter resolved once per call — the shape #2875 used for the SQL predicate.

**One shared `resolveArchivedLanes` for both paths**, deliberately: if the write guard and the publication guard could disagree about whether a card is archived, a card ends up both read-only *and* un-publishable.

## The revert proof caught my own fixture first

My first version set `deletedAt` alongside the renamed column, and **the revert proof passed with the fix removed**. Both guards are `column-is-archived || deletedAt != null`, so a soft-deleted fixture short-circuits the exact comparison under test — the assertion was holding for an unrelated reason.

Dropping `deletedAt` isolates it, and is also the *real* shape: a live row in a workflow-declared archived lane is what a renamed board produces, and what `getLiveTaskColumn` was written to catch.

Revert proof, measured honestly the second time: restore `task.column === "archived"` and the renamed-lane case fails — the upsert resolves instead of rejecting.

## Real PostgreSQL, deliberately

These are row predicates inside a transaction. A mocked store would assert the arguments and prove nothing about the comparison that runs — the same reasoning as #2875.

Three cases: the renamed lane rejects, the **legacy** `archived` id still rejects (most boards never rename anything), and a live card is still allowed through (a guard that rejects everything is its own bug).

## Verification

- `pnpm lint` — clean
- new `archived-document-lanes.pg.test.ts` + existing `artifacts-documents-evals.pg.test.ts` — 12 passed against real PostgreSQL

Note: the SQL-literal baseline is untouched here — #2881 owns re-recording it after #2864's conversion left main's gate red.

### DIFF
--- a/.changeset/archived-document-lanes.md
+++ b/.changeset/archived-document-lanes.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Archived-task document rules now work on boards that rename the archived lane.
+category: fix
+dev: `upsertTaskDocument` and `publishArchivedTaskDocumentAddition` compared `task.column` against the literal `"archived"`. On a renamed archived lane the first failed to reject (an archived card's documents stayed writable) and the second failed to accept (a legitimate archived-document publication was refused as `parent-not-archived`). Both now take a resolved archived-lane set, supplied by their store-level impls.
--- a/packages/core/src/__tests__/postgres/archived-document-lanes.pg.test.ts
+++ b/packages/core/src/__tests__/postgres/archived-document-lanes.pg.test.ts
@@ -1,0 +1,116 @@
+/*
+FNXC:WorkflowLifecycleColumns 2026-07-30-23:40:
+
+THE INVARIANT: "is this card archived?" is answered by the board's archived LANE, not by the id
+`archived`.
+
+`async-comments-attachments.ts` holds two classes of `=== "archived"` spelled identically, which is
+why they were miscounted once already (#2877 review). The comparisons downstream of
+`getLiveTaskColumn` test a value that function MANUFACTURES and must stay literal. The two covered
+here read `task.column` straight off a row `select`, so a renamed archived lane is simply not
+recognised, and they fail in OPPOSITE directions:
+
+  - `upsertTaskDocument` fails to REJECT — an archived card's documents stay writable, so the
+    read-only contract on archived tasks silently does not hold;
+  - `publishArchivedTaskDocumentAddition` fails to ACCEPT — a legitimate archived-document
+    publication is rejected as `parent-not-archived`, which reads to an operator as a data-integrity
+    error rather than a lifecycle mismatch. This is the sharper of the two: valid work refused.
+
+REAL PostgreSQL on purpose. Both guards are row predicates inside a transaction; a mocked store would
+assert the arguments and prove nothing about the comparison that actually runs — the same reason
+#2875 drove its SQL change through a live database.
+
+REVERT PROOF, measured: restore `task.column === "archived"` in either guard and its case fails —
+the upsert resolves instead of rejecting, and the publication throws `parent-not-archived`.
+*/
+
+import { it, expect, beforeAll, beforeEach, afterEach, afterAll } from "vitest";
+
+import {
+  pgDescribe,
+  createSharedPgTaskStoreTestHarness,
+  type SharedPgTaskStoreHarness,
+} from "../../__test-utils__/pg-test-harness.js";
+import * as schema from "../../postgres/schema/index.js";
+import { eq, and } from "drizzle-orm";
+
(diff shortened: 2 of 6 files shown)


gsxdsm:
**The finding is right, and the obvious fix trades down. I implemented it, measured, and reverted.** fe97aaa934 records the result at the site.

The union includes a column if **any** enabled workflow calls it archived, so where two workflows reuse an id and only one marks it archived, a **live** card in the other workflow's lane is refused its own document writes. Both callers have `taskId`, so a keyed answer looks available — and the arity argument I have applied all program says use it.

**Switching to `resolveWorkflowIrForTask(store, taskId)` broke `archived-document-lanes.pg.test.ts`:** a card in a *renamed* archived lane stopped being read-only. The cause is this program's recurring fact seen from the other side — the per-task resolver needs the task's own workflow **selection**, and with none recorded it degrades to the **built-in** IR, whose archived lane is `archived`. The project union needs no selection, which is precisely why it catches the renamed card.

So the two failure modes are not the same size:

| | who it hits |
|---|---|
| **union** (today) | a live card loses document writes — and only where two workflows reuse one id with different traits, a configuration nobody has reported |
| **per-task** (my patch) | **every** renamed-lane card with no recorded selection silently becomes **writable while archived** — the defect this guard exists to prevent, with a passing test |

**The correct fix is per-task with a union fallback when the selection is absent** — narrow when the card can answer, broad when it cannot. That needs *"no selection"* distinguished from *"selection resolved to the default"*, i.e. the provenance form, at both call sites. Sized at the site rather than attempted, because swapping one defect for a larger one would have looked like progress and dropped a guard count — which is the exact pattern `docs/solutions/workflow-learnings/lifecycle-conversions-that-score-as-wins.md` exists to catalogue.

Worth noting for the broader program: this is the **third** shape where a union and a per-task answer disagree, after #2864 (accepted, cost) and #2866 (fixable, keyed). They need different answers, and "always resolve per task" is not one of them.

Core `tsc` clean, `pnpm lint` clean, `archived-document-lanes.pg.test.ts` 3/3.

gsxdsm:

gsxdsm:
### This left `archived-column-gate-parity` red on main — and the reason is worth a look

Fix is up in **#2924** (inventory 8 → 5). The tally drop is expected: replacing three
`column === "archived"` comparisons with `isArchivedLane(...)` removes raw comparisons, and that
guard's AST scan counts them. Its header already records the same thing happening for #2746.

**But the count moved without the behaviour moving, and that is the part worth your attention.**

`archivedColumns` is optional and defaults to `LEGACY_ARCHIVED_LANES = new Set(["archived"])`, and measured:

```
→ (no matches)
```

No caller passes it. So every call resolves to the same literal it replaced — byte-identical behaviour, and the resolved branch is dead code today.

That is not a criticism of the fix: the opposite-direction guard bug you fixed was real, and `isArchivedLane` is the right seam for it. It does mean the renamed-lane half is **not** actually reached yet, so if the PR's intent was that archived-document publication now works on a board whose archived lane is renamed, it does not — the parameter needs threading from the callers that can resolve lanes.

**The reason I am flagging rather than just wiring it:** that guard's header argues at length that converting the TypeScript half while the Drizzle and raw-`sql` halves still compare the string produces a split brain *"no test would catch, because every builtin workflow spells the column `archived` so the two halves agree by accident on every board we ship."* Right now there is no split brain **because** the resolved path is unwired. Wiring it in isolation is what creates one.

The header also explains why the SQL halves cannot simply follow: `ne(tasks.column, …)` needs the resolved id as a query-build value, including inside `for update` transactions in the document/artifact paths that currently take no store and no workflow reader.

So the remaining work is a decision, not a conversion — thread a resolver into those persistence paths, or declare `archived` a non-renameable system column and mark the sites deliberate. #2924 records this inline in the inventory so the next reader does not take `5` as "3 sites done".

PR #2892, opened 2026-07-30

gsxdsm:
The same reviewer finding — *"a project-wide union misclassifies tasks"* — arrived on **five PRs in one day** and got **four different correct answers**. By the fifth, the program had built an instinct: *a union is the lazy option, resolve per task*. Acting on that instinct shipped a regression, so the reasoning is worth recording rather than re-derived a sixth time.

## The deciding question

Not "can I resolve per task?" — usually you can. It is **what over-inclusion costs at this site, and what a missing selection costs**. Those two failure directions are asymmetric in a way that changes per call site:

| site | answer | why |
|---|---|---|
| `awaitingPlanning` enrichment (#2842) | union | over-inclusion is one extra file read; the per-card predicate still answers per card |
| glasses notifier (#2852) | narrow | a notification is an outward action with no downstream narrowing |
| `workflow-analytics` (#2866) | narrow | rows carry a workflow id, so the keyed answer was available — the union was the wrong arity |
| `team-analytics` (#2864) | union, accepted | per-task turns an indexed `COUNT` into an N-read loop; cost recorded instead |
| archived-document guards (#2886) | union, **after trying narrow** | below |

## The one that reversed the instinct

`comments-ops.ts` had `taskId` at both callers, so the arity argument said convert. Converting **broke a passing renamed-lane test**: the per-task resolver needs the task's own workflow *selection* and degrades to the **built-in** board without one, while the union needs no selection — which is exactly why it caught the renamed card.

Trading *"a live card in a colliding id loses document writes"* (a configuration nobody has reported) for *"every renamed card with no selection is writable while archived"* (with a test proving it) is a **bigger defect wearing a correct-sounding principle**. Reverted, and the real fix — per-task with a union fallback when the selection is absent — is sized at the site.

## Why this belongs next to the existing learnings doc

`lifecycle-conversions-that-score-as-wins.md` catalogues conversions that improve the number and break the code through carelessness. This is the same outcome reached through a **principle**, which is harder to spot: the change is defensible at every step and still ends up worse. It also cross-references the untraited-project gap, which is the same helper failing in the *other* direction and is why the fix there cannot be a changed default either.

Docs only; `pnpm lint` clean.

### DIFF
--- a/docs/solutions/workflow-learnings/project-union-versus-per-task-lanes.md
+++ b/docs/solutions/workflow-learnings/project-union-versus-per-task-lanes.md
@@ -1,0 +1,80 @@
+---
+category: workflow-learnings
+module: packages/core/src/project-lane-vocabulary
+tags: [lifecycle-columns, resolved-lanes, arity, project-union, per-task]
+problem_type: design-decision
+applies_when: choosing between resolveProjectColumnsForRoles and a per-task lane resolution
+---
+
+# The project union and the per-task answer are not ranked
+
+Recorded 2026-07-31, after the same reviewer finding arrived on five PRs in one day and got **four
+different correct answers**. The instinct the program had built by then — "a union is the lazy option,
+resolve per task" — is wrong often enough to be worth writing down, because acting on it shipped a
+regression.
+
+`resolveProjectColumnsForRoles(store, roles)` returns every column any workflow in the project gives
+the role, unioned with the legacy ids. A per-task resolution answers for one card's own board. Neither
+is the better tool.
+
+## The question that decides it
+
+Not "can I resolve per task?" — usually you can. It is:
+
+> **What does over-inclusion cost here, and what does a missing selection cost?**
+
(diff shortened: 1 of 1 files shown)


gsxdsm:
**Correct — fixed.** 262491214a.

Dated a day ahead, which matters more than usual here: this document's entire framing is *"this happened over one day, in this order"*, so a wrong date makes the chronology inaccurate for exactly the reader it is written for.

**This is the fourth future-dated stamp a review has caught across my PRs today** (#2843, #2852, #2856, now this). The pattern is clear enough to be worth naming rather than fixed four separate times: every one was a hand-written date in a comment or doc, and none of them had anything to verify against. `AGENTS.md` requires an FNXC timestamp in a fixed format but nothing checks it, so the only feedback loop is a reviewer noticing.

That is a cheap gate — a scan for `FNXC:` stamps parsing to a date later than the commit's would catch all four in one pass. Recording the observation here rather than opening it myself; my area has enough open PRs.

gsxdsm:

PR #2893, opened 2026-07-30
fix(tests): definition-actions queries container; the modal is portalled (12 → 0)

gsxdsm:
## Third file, same root cause

The largest block in the dashboard `app:backfill 2/4` shard — and the same defect as #2885 and #2890.

**Probed before converting, not pattern-matched:**

```
PROBE container=false document=true
```

`TaskDetailModal` mounts through `createPortal`, so its subtree hangs off `document.body` rather than the container `render()` returns. All **25** container-rooted lookups in this file could only ever return null.

## The regex handles two shapes separately, on purpose

`X.container.querySelector` (render-result-scoped) and a bare destructured `container.querySelector` are converted in separate passes, because a blanket replace on the previous file rewrote the former into `renderResult.document...` — not a thing — and broke two passing tests. A lookbehind keeps `triageContainer.` / `todoContainer.` out of the bare pass.

## One case needed more than a query-root swap

`does NOT show Changes tab for triage/todo tasks` renders the triage modal and the todo modal back to back and told them apart by their container handles. **Both were empty**, so `querySelectorAll(".detail-tab")` returned `[]` and the assertion compared `[]` against twelve tab labels — it could not have failed for the reason it was written.

Querying `document` alone does **not** fix that one: with two modals mounted at once, a document-rooted `.detail-tab` lookup returns *both* tab strips concatenated. Unmounting the first render is what makes each assertion about one modal again.

Verified by measurement rather than reasoning — the file only reached 66/66 after the unmount, not after the query swap.

## Evidence

| | result |
|---|---|
| the file | **66/66** (was 12 failed) |
| shard `2/4` | **22 → 10** |
| mutation: rename `.detail-spec-edit-trigger` in `TaskDetailModal` | **1 failed** |

`pnpm lint` clean. Test-only; `TaskDetailModal.tsx` restored clean.

## Running total on the portal defect

| PR | file | cleared |
|---|---|---|
| #2885 | `TaskDetailModal.models-progress-workflow` | 30 |
| this | `TaskDetailModal.definition-actions` | 12 |

**59 of the ~111 backfill failures**, all one defect: tests querying `container` for components that render through a portal. It hid because `screen.*` queries in the same files always worked (they query the document), so the failures read as "the component never rendered" rather than "we asked the wrong root".

### DIFF
--- a/packages/dashboard/app/components/__tests__/TaskDetailModal.definition-actions.test.tsx
+++ b/packages/dashboard/app/components/__tests__/TaskDetailModal.definition-actions.test.tsx
@@ -67,7 +67,7 @@ describe("TaskDetailModal", () => {
         </FileBrowserProvider>,
       );
 
-      const actionRow = container.querySelector(".detail-spec-edit-trigger");
+      const actionRow = document.querySelector(".detail-spec-edit-trigger");
       expect(actionRow).toBeTruthy();
       const promptButton = screen.getByRole("button", { name: "Open PROMPT.md" });
       expect(actionRow?.contains(promptButton)).toBe(true);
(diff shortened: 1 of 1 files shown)


gsxdsm:

```
```

PR #2894, opened 2026-07-30
chore(gate): enforce the SQL ratchet where the census is, not in every local test run

gsxdsm:
Stacked on #2878. Fixes an over-enforcement I introduced, which is why `main` has been blocking local `pnpm test` runs rather than just failing CI.

## The asymmetry

I wired this gate into `pretest`, `pretest:full` **and** `test:gate` — the widest enforcement of the three checks in this family. The ratchet it is modelled on, the lifecycle-column census, is in **none** of them: it is a dedicated `pnpm check:lifecycle-columns` step in `pr-checks.yml` with a `check:*` script beside it.

Same semantics, same failure modes, three times the blast radius. That was not a decision I made — it is what I reached for without checking the precedent sitting next to it.

## Why it matters

A baseline that drifts **down**, because someone legitimately migrated a site, then blocks every developer's `pnpm test` before a single test executes — for a bookkeeping mismatch, not a broken build. That is exactly the state `main` has been in since #2864 merged.

## The change

- `check:sql-column-literals` script, beside `check:lifecycle-columns`
- a dedicated **SQL-column ratchet** step in `pr-checks.yml`, immediately after the census step
- removed from `pretest`, `pretest:full`, `test:gate`

**Enforcement where it matters is unchanged:** `pr-checks.yml` is the blocking gate for PRs, so a newly added forbidden literal still cannot merge. What changes is that it stops halting unrelated local work — the same trade the census already makes, and the reason it makes it.

## Stacked deliberately

On #2878 rather than `main`, so the baseline fix and the placement fix land together. Alone, this would make `main` green locally while CI still failed on the drifted baseline — a worse state than either problem on its own.

## Verification

`pnpm check:sql-column-literals` exit 0 · `pnpm test:gate` green (now without this check in it) · lint 0 · new workflow step confirmed adjacent to the census step.


gsxdsm:
**The gap is deliberate and already recorded at the site; the silence was not, and that part is fixed.** 8d4f56b7a8.

The finding restates the limitation in the comment above it, and that reading is correct: with **no `workflow_id` argument**, *"substituted"* and *"this project has no resolvable workflow"* are the **same observation**, and the second is a legitimate configuration whose cards belong exactly where they are. Failing it would trade a false success for a false alarm on a working setup, and distinguishing the two needs the read that just failed. That is why the branch only errors when a workflow was *explicitly requested*.

**What was fixable:** the card came back indistinguishable from a verified landing. The caller — usually another agent, which reads the first sentence and moves on — now gets `landingVerified: false` in `details`, so *"we could not confirm this card is on a lane anything picks up from"* is a readable fact rather than an absence.

This is the same principle as the contamination sweep one PR over (#2891), where I had to be corrected into it: **when you cannot answer, say so instead of picking a side.** There the wrong answers were legacy-ids-only and the project union; here the wrong answer would be failing a valid configuration. The shape of the fix is the same either way, and it is now the third site this session where "report the unresolvable case" beat both available guesses.

CLI `tsc` clean, `pnpm lint` clean, delegate suite 16/16.

gsxdsm:
**Right, and my first fix put the correct fact in the wrong place.** 47b82aeb78.

`landingVerified: false` sitting beside a confident *"The task will be picked up by X on their next heartbeat cycle"* is **the same success-with-a-caveat shape the error branch on #2843 was rewritten to remove** — the caveat is present and nobody sees it. The caller here is usually another agent, and agents read the first sentence.

The pickup claim is now **conditional on having verified the landing**. Where it could not be verified, the text says what is true: the card exists, is assigned, and sits on a lane we could not confirm — check the board if it does not start. That neither promises dispatch nor fails a configuration that is probably fine, which is the trade the comment above it already argued and the reason this is not an error.

**Two existing tests asserted the old sentence, and I want that on the record rather than quietly edited:** they were pinning the exact claim this review says must not be made on those paths, so they were codifying the defect. Both still assert **success** — the outcome is unchanged — and additionally check `landingVerified: false`, so the prose and the structural fact cannot drift apart later.

That is the second time in this PR's history that a test had to change because it had frozen a wrong behaviour rather than a right one. Worth watching for on the remaining delegate work: a green assertion on message text is only as good as the message was when it was written.

CLI `tsc` clean, `pnpm lint` clean, `extension.test.ts` 75 passed / 96 skipped.

gsxdsm:
Added the behaviour fix this PR was only working around, after #2884 showed the problem is recurring and not mine alone.

## I checked the census and got the wrong answer

Two turns ago I compared this gate to the lifecycle census, concluded it "matches the established convention", and decided no redesign was needed. **That was wrong.** I read the census's header comment — *"fails on a RISE and equally on a DROP that was not recorded"* — and treated the documentation as the behaviour.

Measured it properly this time: inflate a census baseline entry so the real count is below it, run `--strict`, and it **exits 0 and tightens**. Mine exited 1. The header describes an older design.

#2884 is the operator hitting this independently: *"main's MERGE GATE is red… Unlike the census ratchet — which auto-tightens and exits 0 — `check-sql-column-literals` exits 1 on a drop (measured)."* They measured it; I had read it.

## What changed

A drop now tightens the baseline and says so; a rise still fails hard, untouched. Both directions verified: drop → `baseline tightened … COMMIT IT`, exit 0; rise → exit 1; clean → exit 0.

The census's own rationale carries over unchanged, and it is worth quoting because it is the part I would have argued against: *a permanently-red gate is a bigger hole than a stale allowance, because it gets ignored and then nothing is guarded at all.* Twice in one day here — **#2878** and **#2884** — and in both cases the author had done the right thing and was failed for it.

The residual is named in the source rather than glossed: in CI the tightening write dies with the runner, so the committed baseline stays stale until someone commits a tightened one. Bounded to the old count, in a file that already had literals.

## Effect on the rest of the queue

This removes the coupling that made me add baseline re-records to #2866, #2870 and #2871. Those commits are harmless and can stay; they were only ever working around this.

gsxdsm:
Two notes on this PR's current state, both about scope and neither a request to change anything.

## It carries two commits I did not write

`47b82aeb78` and `8d4f56b7a8` — *"an unverified landing now says so instead of looking like a confirmed one"* and its follow-up — modify `packages/cli/src/extension.ts` (+47 lines) and its test. They are about delegate landing-status messaging, not gate placement, and they are not on `main`; this PR is the only open one containing them.

Flagging rather than removing: they were pushed here deliberately and they are healthy — `extension.test.ts` passes 75/171 (96 skipped). But this PR's title says "enforce the SQL ratchet where the census is", and a reviewer reading that will not expect a CLI behaviour change in the diff. Whoever merges should know the scope is wider than the title, and may prefer to split them.

## A `test:gate` run here showed two PG failures — they are contention, not this change

```
FAIL src/__tests__/postgres/handoff-to-review-atomicity.pg.test.ts
FAIL src/__tests__/postgres/sync-workflow-ir-is-always-default.pg.test.ts
```

Both **pass on this branch in isolation** (7/7), and both pass on clean `main`. `test:gate` runs three suites concurrently against PostgreSQL, which is where they fall over. This PR touches only `scripts/`, `package.json` and the workflow file, so it has no mechanism to affect PG behaviour — but I checked rather than reasoned, having pushed before noticing the red.

Not quarantining them: two runs is not enough to call it, and the quarantine ledger is for tests that fail without a real bug in the change, which I have not established. Recording it here so the next person who sees the same pair under `test:gate` has a starting point.

gsxdsm:
Resolved a conflict with `main` by **dropping my version of the auto-tighten fix in favour of #2888**, which landed it first.

The operator implemented the same change independently — a dropped count tightens the baseline instead of failing — so my commit was redundant. Took `main`'s file wholesale rather than merging the two implementations, and re-verified the behaviour rather than assuming the two were equivalent:

```
clean → exit 0
DROP  → exit 0   (tightens)
RISE  → exit 1   (still fails hard)
seam  → exit 0
```

**What remains unique to this PR**, and the reason it is still worth merging:

- both ratchets move out of `pretest` / `pretest:full` / `test:gate` into dedicated `pr-checks.yml` steps beside the census, matching where that ratchet has always been enforced;
- `check:sql-column-literals` and `check:inert-flag-seams` scripts;
- **the inert-seam check gets CI enforcement for the first time** — it previously ran only in local test hooks and in no workflow at all, so a newly added inert seam could merge unopposed while a stale exemption blocked every developer's `pnpm test`;
- the seam gate's stale-exemption case warns instead of failing, for the same measured reason as #2888 — three times this week an entry went stale *because someone fixed the seam it covered*, and the gate went red on them for it.

Still carrying the two unrelated `fix(delegate)` commits flagged earlier.

gsxdsm:
Rebased onto `origin/main`. Two conflicts, and **one of my commits is now redundant** — recording which and why.

**`check-sql-column-literals.mjs`** — both sides implement the *same* change: a baseline **drop** tightens the ceiling instead of failing. Main landed it independently while this branch was open, with the same reasoning I had written (a permanently-red gate is a bigger hole than a stale allowance, because it gets ignored and then nothing is guarded at all — the trade-off the lifecycle census had already settled). I took main's version; my commit contributes nothing to that file now.

**`sql-column-literals-baseline.json`** — resolved to main's, since main's is the current population.

**What still stands on this branch after the rebase:** the delegate work (the unverified-landing caveat moved into the sentence rather than living only in `details`, plus the two tests that had frozen the wrong behaviour), the inert-seam CI wiring fix, and the stale-exemption warn.

Verified on the rebased tip rather than assumed: CLI `tsc` clean, `pnpm lint` clean, `check-inert-flag-seams` exits 0, `check-sql-column-literals` exits 0, `extension.test.ts` 75 passed / 96 skipped.

gsxdsm:

```bash
git cherry-pick <your-sha>
```

gsxdsm:
Superseded by #2941 (batch-sql-ratchet). Folded verbatim onto the family branch; nothing dropped.

PR #2895, opened 2026-07-30
fix(tests): rendering.test queries container; the modal is portalled (28 → 4)

gsxdsm:
## Fourth file, same defect

`TaskDetailModal.rendering.test.tsx` is the whole of the dashboard `app:backfill 4/4` shard bar one case (28 of 29).

**57** container-rooted lookups converted. `TaskDetailModal` mounts through `createPortal`, so `container` is empty and every one returned null — visible in the two failure shapes this file produced:

```
10x   expected null to be truthy
~14x  expected undefined to be '<some text>'      ← container.querySelector(x)?.textContent
```

## Four remain, deliberately

They assert the modal's **wrapper structure**, not its contents:

```ts
expect(document.querySelector(".modal-overlay.open")).toBeTruthy();
```

`FloatingWindow` renders `.floating-window-overlay` (`FloatingWindow.tsx:627`). The only `.modal-overlay open` left in `TaskDetailModal` is the unrelated *refine* overlay at `:6801`. So these pin the **pre-FloatingWindow** wrapper.

Re-pointing them means encoding the *current* modal-shell contract — a UI structure decision that belongs with whoever owns the FloatingWindow adoption, not bundled into a query-root fix where it would be easy to miss. Left failing and flagged rather than guessed at.

It is also a different failure shape from the rest: `expected <div …> to be null` on a mobile-variant badge, i.e. an assertion that *found* something, versus 24 that found nothing. Different cause, different fix, different reviewer.

## Evidence

| | result |
|---|---|
| the file | **120 tests, 4 failed** (was 28) |
| mutation: rename `.detail-id` in `TaskDetailModal` | **5 failed** |

The mutation matters because the change is "query a different root" — the risk is assertions that now find *something* and stop discriminating. They still observe the real component.

`pnpm lint` clean. Test-only; `TaskDetailModal.tsx` restored clean.

## The portal defect, totalled

| PR | file | cleared |
|---|---|---|
| #2885 | `TaskDetailModal.models-progress-workflow` | 30 |
| #2893 | `TaskDetailModal.definition-actions` | 12 |
| this | `TaskDetailModal.rendering` | 24 |

**83 of the ~111 backfill failures**, one defect: tests querying `container` for components that render through a portal.

It hid for so long because `screen.*` queries in the same files always worked — they query the document — so the failures read as *"the component never rendered"* rather than *"we asked the wrong root"*. And it was invisible to CI: the quality runner stops after the first failing lane, and `app:app` failed ahead of every backfill shard.

### DIFF
--- a/packages/dashboard/app/components/__tests__/TaskDetailModal.rendering.test.tsx
+++ b/packages/dashboard/app/components/__tests__/TaskDetailModal.rendering.test.tsx
@@ -12,7 +12,7 @@ query ambiguous once the trigger stopped being a mobile-only affordance.
 import { readFileSync } from "node:fs";
 import { resolve } from "node:path";
 import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
-import { render, screen, fireEvent, act, waitFor } from "@testing-library/react";
+import { render, screen, fireEvent, act, waitFor, cleanup } from "@testing-library/react";
 
 // FNXC:Markdown 2026-06-23-03:30: Mock the heavy `mermaid` library so the shared
 // markdown pipeline's MermaidDiagram resolves without loading the real renderer.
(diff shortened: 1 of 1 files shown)


gsxdsm:
**Not resolving this one — the branch is checked out in the `phase-c-review-merge` worktree, so it has an active owner. Analysis for whoever holds it, plus one correction to the premise this was dispatched under.**

**This PR is not actually green.** `gh pr checks 2895` reports Build, Gate, Lint and Typecheck all **pending**; only the Greptile review has completed. So "GREEN but blocked on threads" is stale, and the finding has not yet been contradicted by a passing run.

**The static evidence supports the reviewer:**

- `TaskDetailModal.tsx` contains exactly **one** `modal-overlay` — line 6801, `"modal-overlay open detail-refine-overlay"`, which is the *refine sub-overlay*, not the modal shell.
- `FloatingWindow.tsx` mentions `.modal-overlay` only inside `FLOATING_WINDOW_OUTSIDE_POINTER_SAFE_SURFACE_SELECTOR` — a list of surfaces that must not trigger outside-pointer close. It **lists** the selector; it does not render it.

So if the normal task-detail modal renders through `FloatingWindow`, there is no `.modal-overlay.open` in the tree and the assertion at line 1060 fails — which is what the finding says.

**One thing to check before acting on it**, because it would change the answer: the assertion was changed in this PR from `container.querySelector` to `document.querySelector` (that is the portal fix). The reviewer's quote shows the `container` form, i.e. the pre-change line. If some ancestor the test renders — or another portal — still emits `.modal-overlay.open` into `document`, the assertion passes for a reason unrelated to the modal shell, which would be worth pinning explicitly rather than leaving to coincidence.

**Cheapest resolution:** wait for Gate, and if it is red, the finding is confirmed and the selector should target whatever `FloatingWindow` actually renders. If it is green, the assertion is passing on something other than the modal shell and deserves a comment saying what.

I have not touched the branch.

gsxdsm:

```
```

gsxdsm:
**Confirmed by running it, and the dispatch premise was wrong twice over.** 358c06f7b4.

This file has **4 failures** on the branch, not 0. `gh pr checks 2895` showed Build, Gate, Lint and Typecheck all **pending** — only the Greptile review had completed — so nothing had verified the branch when it was dispatched as green. I took it once the `phase-c-review-merge` worktree released it and ran the suite.

**The reported finding is exactly right.** `TaskDetailModal` renders `<FloatingWindow modal>`, whose overlay class is `floating-window-overlay--modal`. The only `.modal-overlay` left in the component is the *refine sub-overlay*, which this case never opens — so the assertion queried a class that is never in the tree and failed outright. Fixed, and asserting `--modal` rather than the base class: a non-modal `FloatingWindow` renders the same base, so the bare selector would keep passing if the `modal` prop were dropped.

**The other three failures are a different problem, and I left them deliberately after trying the obvious fix.** They query the render `container`, and the portal puts the modal outside it — but `document.querySelector` is **not** the answer: those cases render the modal *several times*, and a document-wide query returns the **first** portal, so later assertions read the first render's DOM. I made that change and watched the failure simply move from `"Title wins"` to `"Description fallback"`. They need per-render scoping or an unmount between renders, which is a rewrite of the case rather than a selector swap — and doing it blind, at the end of a long session, is how a test gets made green without being made correct.

So: 1 of 4 fixed with confidence, 3 characterised precisely and left with the reason. **This PR is not ready to merge on the strength of the thread being resolved** — it needs those three addressed or explicitly quarantined, and it needs a CI run that has actually completed.

gsxdsm:
**All four now fixed — the file is 120/120.** 9280e1876b. It was **4 red** when this was dispatched as green.

**Two distinct causes, and conflating them is why the obvious repair fails:**

1. **Obsolete selector** — the reported finding, plus the optimistic-task case using the same `.modal-overlay`. The shell is `<FloatingWindow modal>`, whose overlay class is `floating-window-overlay--modal`; the only `.modal-overlay` left in the component is the *refine sub-overlay*, which neither case opens.

2. **Portaled queries in multi-render cases** — these queried the render `container` and the modal portals out of it. **`document.querySelector` is not the fix**, and I verified that rather than assuming: those cases render the modal two or three times, the portals accumulate on `document.body`, and a document-wide query returns the **first**. The failure moved from `"Title wins"` to `"Description fallback"` instead of going away.

`cleanup()` between renders makes the document unambiguous, so each assertion reads the render it belongs to. That also matches what those cases are actually testing — three independent fallback *inputs*, not three coexisting modals. Unused `container` bindings dropped rather than left to rot.

**Retracting my previous comment on this PR**, where I said the three were "a rewrite of the case rather than a selector swap" and left them. That was true about the *selector swap* and wrong about the size: the fix was three `cleanup()` calls. I had stopped at the correct diagnosis and drawn a pessimistic conclusion from it.

Resolving the thread now — unlike last round, the file is actually green. Worth flagging separately that CI on this branch had never completed when it was dispatched twice as "GREEN"; a run should finish before it merges.

Dashboard app `tsc` clean, `pnpm lint` clean, 120/120.

PR #2896, opened 2026-07-30
fix(gate): the inert-seam check ran in NO CI workflow — only in local test runs

gsxdsm:
Stacked on #2894. This one is not a placement preference — the check had **no CI enforcement at all**.

## What I found

`check-inert-flag-seams` was wired into `pretest` and `pretest:full`, and nowhere else. `test:gate` does not include it, and `pr-checks.yml` runs `pnpm test:gate` as its blocking step. **Nothing in CI ever executed it.**

So the arrangement was exactly backwards:

- a **newly added inert seam** — the defect this check exists for, and the one that produced four shipped user-facing bugs in this program — could merge with nothing complaining;
- a **stale exemption** halted every developer's `pnpm test` before a single test ran.

It has done the second thing twice this week. It has never done the first, because it could not.

## The change

Matches the census and the SQL ratchet: a `check:inert-flag-seams` script, a dedicated `pr-checks.yml` step beside theirs, out of `pretest` / `pretest:full` / `test:gate`.

Enforcement moves from *"blocks work that has nothing to do with it"* to *"blocks the PR that introduces it"* — which is a strict increase in what it actually catches, not a relaxation.

## How I found it

By asking why `main` being red on this check was blocking local runs, and then checking whether CI ran it at all. I built this check and wired it up myself, and had assumed for weeks it was guarding PRs. It was guarding nothing.

That is the same shape as the other findings in this program — the signal existed, and the thing meant to surface it was structurally unable to. Worth stating plainly since it is my own tooling.

## Verification

`pnpm check:inert-flag-seams` exit 0 · `pnpm test:gate` green · lint 0 · new step confirmed adjacent to the census and SQL steps.


gsxdsm:
Audited every `scripts/check-*.mjs` for the same mis-wiring, to establish whether this is a repo-wide pattern or one bad connection. **It is one bad connection — this one.**

Four checks showed as having no CI enforcement. Hand-reading them cleared three:

- `check-i18n-parity` — backs the opt-in `i18n:status` report.
- `check-quarantine-ledger` — its own header states the design: *"report-only default; `--strict` is the opt-in enforcement path."*
- `check-test-isolation` — backs `test:isolated`, an explicit deep-check command.

All three are **deliberately** opt-in and correctly wired for what they are. Every other check in the repo (`check-no-nohup`, `check-capacity-pool-id`, `check-changeset-format`, `check-routes-modular`, and the rest) reaches CI through `test:gate` or a dedicated `pr-checks.yml` step.

So the inventory is: one check that blocks local work and does not gate PRs, and it is mine. That bounds this PR — it is a genuine gap rather than a stylistic preference, and there is no follow-up sweep to do.

Worth noting the audit method, since the raw scan was misleading on its own: a heuristic "is it named in a workflow" pass produced four hits with a 75% false-positive rate, and only reading each one settled it. Same lesson as the SQL gate's line-grep, where 25 of 37 hits were comments.

PR #2899, opened 2026-07-30
self-healing: a finished task blocked the merger queue on a renamed board (sixteenth sweep)

gsxdsm:
`reconcileStaleMergerStatus` clears a `merging`/`merging-pr` status left on a card that already reached a terminal lane. Two literal reads meant that on a renamed board it was never cleared.

**Unlike every other sweep in this series, the damage is not confined to the stranded card:** the stale status holds the **merger queue** for every task behind it. One finished card, and the board stops merging.

## One union read, not two buckets

The sweeps before this one split their reads per role because the buckets were treated differently downstream. Here nothing distinguishes complete from archived — the only filter is on `status` — so a single union over `TERMINAL_ROLES` is the honest shape. Splitting them would encode a distinction the code does not make.

Deduped, since the two roles can share a column (the P1 reviewed on #2879).

## No per-card verdict, deliberately

This sweep has **no column comparison to convert**. Adding one would be inventing a gate rather than resolving an existing one, which is out of scope for a conversion — the same reason the redundant guards in #2891 and #2897 were *converted* rather than deleted.

## Revert result

| conversion | reverted → |
| --- | --- |
| the resolved union read | fails — the card is never listed, so its stale status is never cleared |

A non-vacuous companion (same terminal lane, no stale status → untouched) rules out a sweep that clears whatever it finds.

## Verification

### DIFF
--- a/.changeset/self-healing-stale-merger-status-query.md
+++ b/.changeset/self-healing-stale-merger-status-query.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: A finished task no longer blocks the merger queue on boards with renamed columns.
+category: fix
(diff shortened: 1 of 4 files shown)


gsxdsm:

PR #2901, opened 2026-07-30
self-healing: step-complete tasks stayed stranded on a renamed board (seventeenth sweep)

gsxdsm:
`recoverCompletedTasks` rescues a task whose steps are **all done** but whose session died before the executor could hand it to review. The literal read meant that on a renamed board it was never found: finished implementation work sat in the wip lane with no session and nothing to move it on.

That is exactly the shape this sweep exists to catch, made unreachable by the query above it — the pattern this whole series is about.

## The redundant guard converts, it doesn't get deleted

`t.column === "in-progress"` was redundant while the query pinned the column. Under a resolved read it becomes the per-card verdict. Deleting it would silently widen the sweep.

Carries the #2891 shape: **narrow when the card can answer, broad when it cannot** — `resolveWorkflowIrForTask` *substitutes* the built-in IR rather than failing, so an unreadable selection would otherwise reject the very card the project-scoped query had just admitted from a renamed lane.

No dedupe needed here: a single role read, so no card can arrive twice.

## Revert results

| conversion | reverted → |
| --- | --- |
| the per-card verdict | fails — the renamed wip lane does not match |

Observable is the injected `recoverCompletedTask` callback, called once per rescued card with no git on the path, so it sits downstream of both halves. A non-vacuous companion (same finished card already in the review lane → not rescued) rules out a read that returns everything.

## Verification

### DIFF
--- a/.changeset/self-healing-stuck-completed-query.md
+++ b/.changeset/self-healing-stuck-completed-query.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Step-complete tasks stranded by a dead session now reach review on boards with renamed columns.
+category: fix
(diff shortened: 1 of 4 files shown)


gsxdsm:

PR #2902, opened 2026-07-31
self-healing: dead cards held a work slot forever on a renamed board (eighteenth sweep)

gsxdsm:
`recoverInProgressLimbo` frees a card holding a wip slot with **no worktree, no branch and no started step** — nothing is running and nothing will. The literal read meant that on a renamed board it was never found, so the card kept its slot indefinitely and that capacity was denied to work which could actually run.

`task.column !== "in-progress"` was redundant while the query pinned the column; under a resolved read it becomes the per-card verdict. Carries the #2891 shape — **narrow when the card can answer, broad when it cannot** — so an unreadable workflow selection does not reject the card the project-scoped query just admitted.

## The companion case is the interesting one here

A card with no worktree and no branch is the **normal** shape in a hold lane — that is exactly what a queued card looks like. So the non-vacuous companion is not a formality: without it, a read returning every column would make this sweep reclaim cards that were never holding a slot at all. That is a worse outcome than the bug being fixed.

## Revert results

| conversion | reverted → |
| --- | --- |

Observable is `getFalsePositiveRequeueSignal`, called once per stranded candidate before any lease or git work, so the assertion needs no fixture beyond the card.

## Verification

### DIFF
--- a/.changeset/self-healing-inprogress-limbo-query.md
+++ b/.changeset/self-healing-inprogress-limbo-query.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Dead cards no longer hold a work slot forever on boards with renamed columns.
+category: fix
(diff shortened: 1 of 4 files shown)


gsxdsm:

PR #2905, opened 2026-07-31
self-healing: the orphaned-execution signal was never emitted on a renamed board (nineteenth sweep)

gsxdsm:
**What this restores is visibility, not a repair** — worth saying up front, because every other PR in this series fixes a stall.

`recoverOrphanedExecutions` takes no lifecycle action. It emits `task:orphan-detected-no-action` so an operator can see a wip card with no live session behind it — deliberately operator-decides. The literal read meant that on a renamed board the event was never emitted, so the one signal pointing at an orphaned execution was silently absent. The card was just as stuck; nothing said so.

`t.column !== "in-progress"` was redundant while the query pinned the column; under a resolved read it becomes the per-card verdict. Carries the #2891 shape — **narrow when the card can answer, broad when it cannot** — so an unreadable workflow selection does not reject the card the project-scoped query just admitted.

## Revert results

| conversion | reverted → |
| --- | --- |

The observable is the auditor construction, which happens once per detected candidate. A non-vacuous companion (same card in the review lane → no signal) rules out a read that returns everything: a card in review is not an orphaned *execution* — it holds no slot and has no session to be missing.

## Verification

### DIFF
--- a/.changeset/self-healing-orphaned-executions-query.md
+++ b/.changeset/self-healing-orphaned-executions-query.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: The orphaned-execution signal is now emitted on boards with renamed columns.
+category: fix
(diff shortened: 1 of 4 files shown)


gsxdsm:

PR #2906, opened 2026-07-31
self-healing: idle assigned agents were never reattached on a renamed board (twentieth sweep)

gsxdsm:
`reattachOrphanedAssignedExecutions` reattaches a **durable agent** to a task it is still assigned to but has stopped executing. The literal read meant that on a renamed board the reattach never fired.

The resulting state is worse than an unowned stall: the card sits **assigned-but-idle**. The board says an agent is on it, and no other sweep will pick it up while it carries an owner.

`task.column !== "in-progress"` was redundant while the query pinned the column; under a resolved read it becomes the per-card verdict. Carries the #2891 shape — **narrow when the card can answer, broad when it cannot**.

## Revert results

| conversion | reverted → |
| --- | --- |
| the per-card verdict | fails — the renamed wip lane is skipped |

Observable is the injected `resumeAssignedTaskForAgent`, called once per agent with orphaned work. The non-vacuous companion matters here: an assigned card that already reached review has **finished** its execution, so resuming it would restart completed work — a read returning every column would do exactly that.

## Verification

### DIFF
--- a/.changeset/self-healing-reattach-assigned-query.md
+++ b/.changeset/self-healing-reattach-assigned-query.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Idle assigned agents are reattached to their work on boards with renamed columns.
+category: fix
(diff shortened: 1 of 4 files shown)


gsxdsm:

PR #2908, opened 2026-07-31
docs(lanes): audit the last three files the census points at with no reason attached

gsxdsm:
Second pass of #2873's sweep, over the files that still carry lifecycle guards and **zero** audit notes. No source change — every literal stays counted, none gets an exemption marker.

## `project-store-ops.ts` (1) — dead sync path, do **not** convert

The literal would leak a merge-queue entry on a renamed board: a card leaving review would never be dequeued. Except the function cannot run — it reaches for `store.db.prepare`, which throws in PostgreSQL backend mode.

The live path is `dequeueMergeQueueOnColumnExitInTransaction` (`async-merge-coordination.ts`, called from `moves.ts`), and it is **already converted** — it takes `moveReviewColumns` and the caller supplies them. Recorded so the census entry is not mistaken for unconverted debt, and so it can be deleted alongside the rest of the sync SQLite residue.

## `task-id-integrity.ts` (2) — one real, one sentinel, and the real one must not go alone

```ts
return cached?.column === "archived";   // ← board lane: real
if (live === "archived") return true;   // ← getLiveTaskColumn's manufactured value: sentinel
```

Converting the first while `getLiveTaskColumn` still keys on the literal would leave the two disagreeing about what "archived" means. It waits for that one, which is the single highest-leverage line in this cluster — fixing it makes five downstream sentinel checks correct without touching any of them.

## `auto-merge-finalization.ts` (3) — one real but diagnostic-only, two non-defects

`task.column === "done"` selects which **reason string** is reported; both arms return `{ ok: false }`. So a renamed board is refused with the generic `missing-merge-confirmation` instead of the specific `done-without-merge-confirmation`. Real, and worth less than the signature change required to fix it — the resolver two functions up already computes `isCompleteColumn`, but this function does not receive it.

The other two are **not** defects and it is worth saying so explicitly: the `columnId === "done"` near the top is the resolver's documented degraded fallback (the live arm calls `columnHasFlag`), and the `step.status` comparison is a **step status**, not a column.

## The pattern across both passes

Of **6 files and 15 guards** audited: **2** were live defects worth converting, **4** were sentinels or dead paths that would have *broken* a renamed board if converted, and the rest were diagnostics or misfiled step statuses.

That ratio is the argument for these notes existing. A file's census count is an upper bound on convertible sites, not a work estimate — and in this cluster the naive reading of the number would have made things worse more often than better.

## Verification

- `pnpm lint` — clean
- `tsc --noEmit` (`@fusion/core`, `@fusion/engine`) — clean
- census `--strict` — exit 0, counts unchanged (that is the point)

### DIFF
--- a/packages/core/src/task-store/project-store-ops.ts
+++ b/packages/core/src/task-store/project-store-ops.ts
@@ -504,6 +504,20 @@ export function getRunAuditEventsImpl(store: TaskStore, options: RunAuditEventFi
     return rows.map((row) => store.rowToRunAuditEvent(row));
   }
 
+/*
+FNXC:WorkflowLifecycleColumns 2026-07-31-02:45 (audited — DEAD SYNC PATH, do not convert):
+The literal below would leak a merge-queue entry on a renamed board — a card leaving review would
+never be dequeued — except that this function does not run in production.
+
+It is the SQLite-mode twin. The live path is `dequeueMergeQueueOnColumnExitInTransaction`
+(`async-merge-coordination.ts`), called from `moves.ts`, and it is ALREADY converted: it takes
+`moveReviewColumns` and the caller supplies them. This body reaches for `store.db.prepare`, which
+throws in PostgreSQL backend mode, so a renamed board never gets far enough to be mis-dequeued.
+
+Converting it would mean threading a lane set into a function whose first statement cannot execute.
+Recorded instead, so the census entry is not mistaken for unconverted debt — and so that whoever
+finally deletes the sync SQLite residue can take this with it.
+*/
 export function dequeueMergeQueueOnColumnExitImpl(store: TaskStore, taskId: string, previousColumn: ColumnId, nextColumn: ColumnId, now: string): void {
     if (previousColumn !== "in-review" || nextColumn === "in-review") {
       return;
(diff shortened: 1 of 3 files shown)


gsxdsm:
Measuring the deferral I named in this PR, so the next person does not re-derive it.

This note calls `getLiveTaskColumn`'s own comparison "the single highest-leverage line in this cluster" — fix it and five downstream sentinel checks become correct without being touched. That is true, and here is what it actually costs:

```
12 call sites across 5 files:
  async-comments-attachments.ts   8   ← #2886 already edits this file
  task-artifacts-ops.ts           2
  audit-ops.ts                    1
  comments-ops.ts                 1   ← #2886 already edits this file
  task-id-integrity.ts            1
```

**Every one of them takes a `db`/`tx` handle plus a `projectId` — none holds a store**, so none can resolve a lane itself. `getLiveTaskColumn` cannot resolve internally either, for the same reason: `resolveProjectColumnsForRoles` needs `listWorkflowDefinitions`, which lives on the store, not on the data layer.

So the fix is an optional `archivedColumns` threaded from ~12 impl-level entry points down through five files, two of which have an open PR against them. That is a coherent single change or nothing — doing half leaves callers disagreeing about what "archived" means, which is the exact hazard this note warns about for `task-id-integrity.ts`.

I am not attempting it from here, and I would rather state the number than leave "needs a design change" as the reason. It wants one owner taking all five files at once, after #2886 lands.

**One thing that makes it more tractable than the count suggests:** the change is mechanical at every site — the same `resolveArchivedLanes(store)` helper #2886 already adds to `comments-ops.ts`, passed one level down. The risk is not per-site judgement, it is the atomicity.

PR #2911, opened 2026-07-31
fix(tests): two more portal query-root failures (AgentListModal, SubtaskBreakdownModal)

gsxdsm:
## The last two in `app:backfill 1/4`

Both are the portal defect (#2885, #2890, #2893, #2895, #2907): `container` from `render()` is empty for a portalled subtree, so the lookups could only ever return nothing.

**Bundled because they are one defect and one line each.** The other three singles in this shard were deliberately kept separate (#2907, #2910) because they had three *different* causes — bundling those would have made each mutation-check prove less.

Probed before converting, as with every file in this series:

```
AgentListModal   PROBE container=0 document=3
```

## What made the second one hard to read

`SubtaskBreakdownModal` failed with:

```
the given combination of arguments (undefined and string) is invalid for this assertion
```

That sounds like a bad matcher, not a missing element. The cause:

```ts
const modal = container.querySelector(".planning-modal");   // → null
expect(modal?.getAttribute("style")).toContain("--keyboard-overlap: 250px");
```

`container.querySelector` returns null, and the `?.` turns that into `undefined` before `toContain` ever sees a string. **Optional chaining disguised a query-root bug as an assertion-type error** — worth recording, because that message gives no hint where to look, and this pattern (`x?.getAttribute(...)`) is common across these specs.

## Evidence

| | result |
|---|---|
| both files | **99/99** (was 2 failed) |
| mutation: rename `.agent-task` in `AgentListModal` | **1 failed** |
| mutation: rename `--keyboard-overlap` in `SubtaskBreakdownModal` | **1 failed** |

`pnpm lint` clean. Test-only; both components restored clean.

## Portal defect, running total

| PR | file(s) | cleared |
|---|---|---|
| #2885 | `models-progress-workflow` | 30 |
| #2893 | `definition-actions` | 12 |
| #2895 | `rendering` | 24 |
| #2907 | `summary-tab` | 2 |
| this | `AgentListModal` + `SubtaskBreakdownModal` | 2 |

**87 backfill failures from one defect.** With #2910 (jsdom shorthand, 1), `app:backfill 1/4` reaches **zero**.

### DIFF
--- a/packages/dashboard/app/components/__tests__/AgentListModal.test.tsx
+++ b/packages/dashboard/app/components/__tests__/AgentListModal.test.tsx
@@ -272,7 +272,15 @@ describe("AgentListModal", () => {
         expect(screen.getAllByText((_, el) => el?.textContent === "FN-PROGRESS · In Progress").length).toBeGreaterThanOrEqual(1);
         expect(screen.getAllByText((_, el) => el?.textContent === "FN-BARE · Unresolved task").length).toBeGreaterThanOrEqual(1);
       });
-      expect(container.querySelectorAll(".agent-task").length).toBe(3);
+  /*
+      FNXC:PortalQueryRoot 2026-07-31-20:10:
+      document, not container — this modal renders through a PORTAL.
+
+      `container` from `render()` is empty for a portalled subtree, so this lookup could only ever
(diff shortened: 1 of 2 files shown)


gsxdsm:
superseded by #2940 (batch-portal-test-fixes) — commit cherry-picked clean, verified there (167/167 scoped).

PR #2912, opened 2026-07-31
self-healing: a card stuck mid-merge could not be retried on a renamed board (twenty-first sweep)

gsxdsm:
`recoverStaleMergingStatus` clears a `merging`/`merging-pr` stamp left on a review card with no live merger behind it. The literal read meant that on a renamed board the stamp was never cleared.

**The operator's escape hatch was closed by the same bug that caused the stall.** That stamp is consulted by the merger *and* by the dashboard's manual Retry gate, so the card could neither progress on its own nor be retried by hand.

`task.column !== "in-review"` was redundant while the query pinned the column; under a resolved read it becomes the per-card verdict. Carries the #2891 shape — **narrow when the card can answer, broad when it cannot**.

## Fixture note worth keeping

`updatedAt` in the test is deliberately ancient. `isStaleMergeActiveStatus` requires the stamp to have sat untouched for `minAgeMs`, so a fresh fixture would be filtered out for a reason that has nothing to do with lanes — and would then have passed with the fix reverted. That is the shape of most of the vacuous assertions on this branch: a *later* filter rejecting the card, masking whether the lane logic worked at all.

## Revert results

| conversion | reverted → |
| --- | --- |

A non-vacuous companion (same stamp on a wip card → untouched) rules out a read that returns everything; a merge stamp in the wip lane belongs to `recoverInProgressLimbo` and the executor, not here.

## Verification

### DIFF
--- a/.changeset/self-healing-stale-merging-status-query.md
+++ b/.changeset/self-healing-stale-merging-status-query.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: A card stuck mid-merge can be retried again on boards with renamed columns.
+category: fix
+dev: `recoverStaleMergingStatus` read the literal `in-review`, so a stale `merging`/`merging-pr` stamp was never cleared on a renamed board. That stamp gates both the merger and the dashboard's manual Retry, so the card could neither progress nor be retried by hand. Read resolves via `resolveProjectColumnsForRoles`, per-card check resolves per card.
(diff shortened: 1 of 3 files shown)


gsxdsm:
Rebased onto `origin/main`. One conflict, and it is worth recording because the naive resolution silently corrupts the file.

Both sides **appended a new case to the same spot** in `self-healing-query-filter-blindness.test.ts` — main's fifteenth sweep (`recoverMergedReviewTasks`, merge-confirmed card stranded on a renamed review lane) and this branch's twenty-first (`recoverStaleMergingStatus`, stale merge stamp on a renamed review lane). They are **different sweeps testing different defects**, so both are kept; taking either side alone would have dropped a regression test while looking like a clean resolution.

**The trap:** the conflict boundary fell *inside a comment block*, so `ours` ended mid-comment and `theirs` began mid-comment. Concatenating the two sides produced a file that lost one test's closing `});` and the next one's opening `/*` — and the result **still looked plausible in a diff**. It failed to parse (`Expected ";" but found "2026"`), which is the good outcome; a conflict that splices differently could just as easily have produced valid code with a test silently swallowed into a comment.

Verified after repair rather than assumed: **18/18** in that file, both new cases present by name, engine `tsc` clean, `pnpm lint` clean, gate green (161/487/13/71).

PR #2914, opened 2026-07-31
self-healing: ghost review cards went undetected on a renamed board (twenty-second sweep)

gsxdsm:
`recoverGhostReviewTasks` finds a **ghost**: a card parked in review past the stuck timeout with nobody owning its merge lane. The literal read meant that on a renamed board it was never found — no merger, no session, and no timeout ever firing against it.

## The redundant guard converts; the kick-back does not

`task.column === "in-review"` was redundant while the query pinned the column; under a resolved read it becomes the per-card verdict.

The kick-back keeps its literal `todo` **deliberately**: it passes `recoveryRehome: true`, one of the 22 documented escapes `moves.ts` exempts so a card stranded in an undeclared column stays rescuable. Converting it removes the rescue path.

## The observable is chosen, not convenient

`isMergeLaneOwned` runs once per surviving candidate, **after** the read and the verdict.

Worth stating plainly: earlier on this branch I positioned this *same spy* **upstream** of the guard I was testing, and it passed with the fix reverted. That was one of the vacuous assertions. The spy is not the problem — its position relative to the change is.

Same reason `taskStuckTimeoutMs` is set and `columnMovedAt` is ancient in the fixture: otherwise the card is filtered out by the timeout rather than by lane, and the case would pass reverted for a reason that has nothing to do with columns.

## Revert results

| conversion | reverted → |
| --- | --- |

A non-vacuous companion (long-idle card in the hold lane → not a ghost) rules out a read that returns everything: a card idle in hold is just queued, and kicking it back would be the sweep inventing work.

## Verification

### DIFF
--- a/.changeset/self-healing-ghost-review-query.md
+++ b/.changeset/self-healing-ghost-review-query.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Ghost review cards are detected again on boards with renamed columns.
+category: fix
+dev: `recoverGhostReviewTasks` read the literal `in-review`, so a card parked past the stuck timeout with no merge-lane owner was never found on a renamed board. Read resolves via `resolveProjectColumnsForRoles` and the per-card check resolves per card; the kick-back keeps its literal target because it passes `recoveryRehome: true`.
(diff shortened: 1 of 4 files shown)


gsxdsm:

PR #2916, opened 2026-07-31
self-healing: transient merge failures never recovered on a renamed board (twenty-third sweep)

gsxdsm:
`recoverTransientMergeFailures` refunds the retry budget for a merge that failed for a **transient** reason and burned all its retries. The literal read meant that on a renamed board the refund never happened, so a card that failed on a network blip or a provider fault stayed failed permanently — visibly failed to the operator, with no visible cause.

## Two lane guards, and the test caught the second

I converted the read and the slim-snapshot filter, and the new case **still failed**. There is a *second* guard: after the full row is re-read, `task.column !== "in-review"` re-checks the selector. Converting only the first left it rejecting every renamed-board card the widened query had just found.

That is the "convert the pair or neither" rule from the sibling doc, and it was caught by **running the test**, not by reading the function — which is the argument for the test existing at all. A conversion that stops at the first guard looks complete and delivers nothing.

## Fixture note

The error string is a **real** transient signature — `"ACP turn failed"`, which `classifyTransientMergeError` matches — not invented prose. An unrecognised string is filtered out one line later, so the case would pass with the fix reverted. Same trap as the post-done wedge fixture earlier in this series.

## Revert results

| conversion | reverted → |
| --- | --- |
| the resolved read + both guards | fails — the card is never listed |
| the **second** guard alone | fails — the re-read row is rejected |

The second row is the one worth having: it proves the half I nearly shipped unconverted.

A non-vacuous companion (transient failure on a wip card → no refund) rules out a read that returns everything; a card that never reached merge has no merge budget to refund.

## Verification

### DIFF
--- a/.changeset/self-healing-transient-merge-query.md
+++ b/.changeset/self-healing-transient-merge-query.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Merges that failed on a transient fault recover again on boards with renamed columns.
+category: fix
+dev: `recoverTransientMergeFailures` read the literal `in-review` and kept two lane guards on column ids, so a card that burned its retry budget on a network blip or provider fault stayed failed permanently on a renamed board. Read resolves via `resolveProjectColumnsForRoles`; both the slim-snapshot filter and the full-row re-check resolve per card.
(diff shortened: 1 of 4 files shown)


gsxdsm:

PR #2917, opened 2026-07-31
feat(gate): freeze the future-dated FNXC stamp surface — 206 stamps, none may be added

gsxdsm:
`AGENTS.md` requires an FNXC stamp on every such comment, and **nothing checks it** — the only feedback loop is a reviewer noticing. On **2026-07-30 alone, reviewers caught four future-dated stamps across separate PRs** (#2843, #2852, #2856, #2892). Every one was hand-written with nothing to verify against, which is a gap a script closes better than four separate catches.

## Why it matters beyond tidiness

FNXC comments are this project's why-does-this-exist trail, and the census, the solutions docs and several review conventions read them **chronologically**. A comment reading *"Recorded 2026-07-31"* beside a 2026-07-30 commit makes the ordering wrong for exactly the reader it was written for — which is how #2892's reviewer put it, and they were right.

## Why a baseline ratchet rather than a hard fail

The number decides it: **87 source files already carry 206 future-dated stamps**, the furthest nearly three months out (`2026-10-19`). A gate that fails on all of them is unmergeable and would be switched off within a day, and mass-editing 87 files to satisfy a new check is churn nobody asked for.

So the population is frozen, same shape as the SQL column-literal gate:

- a **new** future-dated stamp fails;
- an existing one does not;
- a **drop** also fails, so a corrected file cannot leave a slot the surface silently regrows into.

## Why "future" and not "matches the commit date"

A stamp legitimately *predates* its commit — work written Monday and landed Wednesday is normal and correct. Only a date that **has not happened yet** is unambiguously wrong. That rule catches every case a reviewer has caught so far without inventing a stricter one nobody would follow, and it needs no git plumbing.

Generated output is skipped (`dist`, `.gate-bundle`, `coverage`): those stamps are copies of the source ones and would double-count every fix.

## Verification

- **Probe:** injecting one future-dated stamp into `live-agent-count.ts` exits **1**; removing it exits **0**.
- Wired into `pretest` and `test:gate` beside the sibling literal gates.
- `pnpm lint` clean; full gate green (161/487/13/71) with the new check running inside it.


gsxdsm:
Superseded by #2941 (`batch-sql-ratchet`) — the gate/ratchet family is folded into one branch and one CI run per the consolidation directive. Your commits were cherry-picked with authorship preserved; nothing was dropped, including the two `fix(delegate)` commits that rode in #2894. See #2941 for the conflict resolutions and for a future-dated FNXC stamp of mine that the fold caught.

PR #2918, opened 2026-07-31
self-healing: step-incomplete review cards never requeued on a renamed board (twenty-fourth sweep)

gsxdsm:
`recoverStaleIncompleteReviewTasks` requeues a review card whose **steps are not finished** — it reached review on a graph failure, not on completed work. The literal read meant that on a renamed board it was never requeued: the card sat in review claiming to be done while its own steps said otherwise.

## Checked for a second pair, deliberately

The triple-proof in this sweep is **not** lane-gated, so unlike #2916 (a second guard on a re-read row) and #2879 (five more in the loop body), there is nothing else here to convert. Verified with the derived ratchet added in #2879 rather than by eye.

The requeue keeps its literal `todo`: `recoveryRehome: true`, one of the 22 documented escapes `moves.ts` exempts so a card stranded in an undeclared column stays rescuable.

## Revert results

| conversion | reverted → |
| --- | --- |

Observable is `evaluateBackwardMoveTripleProof`, private and called once per candidate before any move, so no git fixture is involved. A non-vacuous companion (same unfinished card in the wip lane → untouched) rules out a read that returns everything: unfinished steps in the wip lane mean the card is simply *being worked on*, and requeueing it would interrupt live work.

## Verification

### DIFF
--- a/.changeset/self-healing-stale-incomplete-review-query.md
+++ b/.changeset/self-healing-stale-incomplete-review-query.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Review cards with unfinished steps are requeued again on boards with renamed columns.
+category: fix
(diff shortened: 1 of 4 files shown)


gsxdsm:

PR #2919, opened 2026-07-31
self-healing: falsely-failed tasks stayed failed on a renamed board (twenty-fifth sweep)

gsxdsm:
`recoverMisclassifiedFailures` clears a failure the executor parked for *"without calling fn_task_done"* on a task whose steps are **all actually done** — the failure is a misclassification, not real work left undone. The literal read meant that on a renamed board it was never cleared, so finished work stayed visibly failed and never entered normal review.

## No second pair

Verified with the derived ratchet added in #2879 rather than by eye. That instrument exists because #2916 found a second lane guard on a re-read row, and the follow-up audit found five more in another sweep — checking is now cheap, so it is not skipped.

## Fixture note

The error string carries the **real** phrase `isNoTaskDoneFailure` matches (`"without calling fn_task_done"`). Invented prose is filtered out one line later, and the case would then pass with the fix reverted. Third time this exact trap has come up in this series, so it is now the first thing I check when writing one of these fixtures.

## Revert results

| conversion | reverted → |
| --- | --- |

A non-vacuous companion (same card in the wip lane → untouched) rules out a read that returns everything: a card still in wip has not handed off, so clearing its failure would hide a live problem rather than fix a stale one.

## Verification

### DIFF
--- a/.changeset/self-healing-misclassified-failures-query.md
+++ b/.changeset/self-healing-misclassified-failures-query.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Falsely-failed tasks with all steps done are cleared again on boards with renamed columns.
+category: fix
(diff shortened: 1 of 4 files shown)


gsxdsm:

PR #2920, opened 2026-07-31
self-healing: branch misbinding went undetected on a renamed board (twenty-sixth sweep)

gsxdsm:
`recoverBranchMisboundInReviewTasks` detects a review card whose **branch tip is bound to a different task's work**. The literal read meant that on a renamed board the misbinding was never detected, so the card would merge — or refuse to — against a branch that is not its own.

## No second pair

Verified with the derived ratchet added in #2879, not by eye.

## Revert results

| conversion | reverted → |
| --- | --- |

Observable is `resolveSelfHealingMergeTarget`, private and called once per candidate, so the assertion sits downstream of both halves without a git fixture.

A non-vacuous companion (same card in the wip lane → not checked) rules out a read that returns everything, and the reason is substantive: a card still in wip legitimately owns a *moving* branch tip, so checking it here would flag normal in-progress work as misbound.

## Verification

### DIFF
--- a/.changeset/self-healing-branch-misbound-query.md
+++ b/.changeset/self-healing-branch-misbound-query.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Branch-misbinding is detected again on boards with renamed columns.
+category: fix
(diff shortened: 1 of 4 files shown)


gsxdsm:

PR #2921, opened 2026-07-31
chore(census): a migration marker is not a lifecycle lane (170 → 169)

gsxdsm:
`plugin-store.ts:335` compares a `__meta` value against `"done"` to record whether `pluginCentralMigrationV1` has already run — the same vocabulary the two writes a few lines below it use. The census counts `=== "done"` comparisons, so it entered the backlog as an owed conversion.

**It is not one, and converting it would be actively harmful.** The comparison would then test a *migration flag* against a *workflow's complete lane* — meaningless on its face, and on any board that renames `done` the check would stop matching its own marker and the migration would re-run on every open.

## Why mark it rather than leave it counted

A backlog entry is a claim that work is owed at that line. The next person down the list spends real time discovering it is not — and the risk isn't just wasted time: this is exactly the shape that gets converted mechanically by someone working a count down, because `=== "done"` looks like every other guard in the list.

## The census's own reporting is correct here, and worth noting

`--strict` refused the change until the baseline was re-recorded, with the right diagnosis:

> Unconverted debt did NOT increase — a marker moved these sites out of the guard count.

That is precisely what happened: **170 → 169 is a reclassification, not a conversion.** Nothing was fixed; one site was correctly identified as never having been broken. The distinction matters for reading the program's progress, and the guard makes it impossible to bank a marker as if it were a conversion.

### DIFF
--- a/packages/core/src/plugin-store.ts
+++ b/packages/core/src/plugin-store.ts
@@ -332,6 +332,18 @@ export class PluginStore extends EventEmitter<PluginStoreEvents> {
     const marker = this.localDb
       .prepare("SELECT value FROM __meta WHERE key = 'pluginCentralMigrationV1'")
       .get() as { value: string } | undefined;
+    /*
+    FNXC:LifecycleColumnCensus 2026-07-30-23:59 DELIBERATE-LITERAL: a MIGRATION MARKER, not a lane.
+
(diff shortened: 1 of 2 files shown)


gsxdsm:

```bash
git cherry-pick <your-sha>
```

gsxdsm:

PR #2922, opened 2026-07-31
self-healing: the "unfixable" query, and an arity bug beside it (twenty-seventh sweep)

gsxdsm:
`recoverMissingWorktreeReviewFailures` requeues a review card failed because its worktree was gone when the session tried to start. Two independent defects here.

## 1. The note at the site said the query was unfixable

It called the literal read *"unfixable without a project-level lane resolution before the read"*. `resolveProjectColumnsForRoles` **is** that resolution — it did not exist when the note was written.

So the careful per-candidate lane wiring below it only ever helped boards whose review lane still happened to be named `in-review` (the partial-rename case the note itself calls out). Fully renamed boards never reached it. A note recording a real constraint outlived the constraint, and read as a closed question.

## 2. An arity bug in the same seam

The per-candidate set came from `resolveTaskLifecycleColumns().review` — the **first** column per role — while all three classifiers take a **membership** set:

```ts
return (reviewColumns ? reviewColumns.has(task.column) : task.column === "in-review")
```

On a board declaring a separate merge lane beside its human-review lane, a card in the second one read as not-in-review and was skipped. Now unions `columnsWithFlag` across the three review roles. This is the arity trap this program keeps hitting — first-per-role answering a membership question.

## Revert results

| conversion | reverted → |
| --- | --- |
| the resolved read | **both** cases fail — the cards are never listed |
| the membership set | **only** the two-review-column case fails |

That asymmetry is the point: it shows the two fixes are independent, not one fix asserted twice.

## Fixture note

The error string uses a **real** prefix from `MISSING_WORKTREE_SESSION_PREFIXES`. Invented prose is rejected by `isMissingWorktreeSessionStartFailure`, and the case would then pass with the fix reverted.

## Verification

`pnpm test:gate` 161 + 487 + 13 + 71, plus `self-healing.test.ts` + the blindness file 430 and `restart-recovery-coordinator.test.ts`; `tsc` engine clean; `pnpm lint`, `check:changesets`, census `--strict` and `check-sql-column-literals` clean, each run explicitly.

### DIFF
--- a/.changeset/self-healing-missing-worktree-review-query.md
+++ b/.changeset/self-healing-missing-worktree-review-query.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Review failures from a missing worktree recover on renamed boards, and on boards with two review columns.
+category: fix
+dev: `recoverMissingWorktreeReviewFailures` had per-candidate lane wiring but still read the literal `in-review`, so only boards whose review lane kept that name benefited. The read now resolves via `resolveProjectColumnsForRoles`. Its per-candidate set also came from `resolveTaskLifecycleColumns().review` (first column per role) while the classifiers take a membership set; it now unions `columnsWithFlag` across the three review roles.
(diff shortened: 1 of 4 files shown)


gsxdsm:

PR #2923, opened 2026-07-31
chore(census): the async archived check compares a sentinel, not a lane

gsxdsm:
`isTaskArchivedAsyncImpl` tests `getLiveTaskColumn(...) === "archived"`. That helper **normalizes**: it returns the string `"archived"` for an archived row *and* for a soft-deleted one — which is why the very next line handles `null` separately, and which the existing inline comment already states. The comparison is against the function's **return vocabulary**, not a column id.

Converting it would keep passing on the built-in board and start **failing** on a renamed one: a soft-deleted task would report as not-archived. Same class as the `plugin-store.ts` migration marker (#2921) and the `comments-ops.ts` document guards — a legacy id that is a *protocol value* rather than a lane.

## The sibling is real debt and stays counted

`isTaskArchivedImpl`, a few lines above, reads `cached?.column` — that **is** a board lane. It remains in the backlog, blocked on resolving lanes from a sync path that has no store, which is the same blocker recorded for `getLiveTaskColumn` itself (#2820).

Marking the sentinel while leaving the lane is the distinction this file needed. Before, its count of 2 implied two conversions were owed; now it points at the one line that actually is. **That is the value here — not the number moving, but the number meaning something.**

## Reporting

169 → 169. The census classifies this as a **reclassification, not a conversion**, and required the baseline re-recorded in the same change — the same guard that caught #2921. Nothing was fixed; one site was identified as never having been broken.

The doc cross-reference was **verified to resolve** before committing, having been burned by a dangling one on #2873.

### DIFF
--- a/packages/core/src/task-store/task-id-integrity.ts
+++ b/packages/core/src/task-store/task-id-integrity.ts
@@ -436,7 +436,23 @@ export async function isTaskArchivedAsyncImpl(store: TaskStore, id: string): Pro
     
     const layer = store.asyncLayer!;
     const live = await getLiveTaskColumn(layer.db, id, layer.projectId);
-    // getLiveTaskColumn returns "archived" for archived OR soft-deleted rows.
+    /*
+    FNXC:LifecycleColumnCensus 2026-07-31-00:20 DELIBERATE-LITERAL: a SENTINEL, not a board lane.
+
+    `getLiveTaskColumn` NORMALIZES — it returns the string "archived" for an archived row AND for a
+    soft-deleted one, which is why the line below it distinguishes `null` separately. So this compares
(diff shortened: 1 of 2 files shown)


gsxdsm:

```bash
git cherry-pick <your-sha>
```

gsxdsm:
Superseded by #2943 (batch-census-sentinels), verified rather than assumed.

The same change is already folded there as `de82e1acf7`. A raw `git diff` comparison between the two branches does NOT match, which is why this needed a second look — but the difference is entirely positional: the blob hashes differ (different base commits) and the hunk header moved from `@@ -436` to `@@ -452` because the batch carries other edits to the same file ahead of this one. The added and removed LINES are byte-identical.

Closing with nothing lost.

PR #2926, opened 2026-07-31

gsxdsm:
Following the rule #2904 just applied to the operator's own numbers. Three measured claims of mine in this file needed it, and one was not merely stale — it was wrong.

## Stale, now dated

| claim | when written | today |
|---|---|---|
| `{ column: "<legacy id>" }` probe finds **79 sites** | 2026-07-30 | **42** |
| **48** named legacy-id collections | 2026-07-30 | **49** |

Dated rather than rewritten, because for both of these the *ratio* to legitimate uses is the argument, not the total — and a bare updated number would lose the reason they were measured at all.

## Wrong, now corrected in place

> the SQL surface survived this test — **14 sites, all genuinely vocabulary-bound** — and got a ratchet

Both halves were wrong within a day of my writing them:

- the population was **31**, once the scanner stopped missing Drizzle-templated queries (#2841 review);
- hand-reading split it into roughly **14 lane-bound**, **11 `archived`** comparisons — three of which are *correct as literals*, reading a state rather than a lane — and **one in dead code**.

## Why this one is worth calling out

That sentence sits in the section headed *"measure a candidate surface, then inspect a sample, before reporting it OR instrumenting it."*

**I asserted an uninspected total in the paragraph telling people not to.** The correction is kept visible in the doc rather than silently overwritten, because the failure is more instructive than the corrected number — it is the third time in this program a claim of mine felt verified and was not.

Docs only — no changeset.

### DIFF
--- a/docs/solutions/workflow-learnings/lifecycle-conversions-that-score-as-wins.md
+++ b/docs/solutions/workflow-learnings/lifecycle-conversions-that-score-as-wins.md
@@ -152,7 +152,8 @@ hands in.** If your fixture passes the lane name, you have tested the consumer. 
 computes that argument in production, and whether they compute it or spell it.
 
 MEASURED, so nobody builds the wrong instrument: an AST probe for a call argument
-`{ column: "<legacy id>" }` finds **79 sites** across `packages/`. It flags the two real `moves.ts`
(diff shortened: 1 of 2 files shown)


gsxdsm:

PR #2927, opened 2026-07-31
self-healing: the zero-commit audit went half-blind on a renamed board (twenty-eighth sweep)

gsxdsm:
`auditNoCommitsExpectedCandidates` flags a card that finished every step and pushed **no commits** — either a legitimately commit-free task nobody declared as such, or work that silently produced nothing.

## Half-blind, not dead — which is harder to notice

This sweep has two arms. It also reads all *failed* tasks, and that read is lane-independent, so on a renamed board the `no_commits` **error** arm kept working while the **lane** arm went silent. A card sitting quietly in a renamed review lane with zero commits and no error was never flagged.

Every other sweep in this series went fully dead on a renamed board, which at least produces a conspicuous absence. This one kept reporting, just less — the failure mode that survives longest.

## Revert result

| conversion | reverted → |
| --- | --- |
| the resolved read + lane verdict | fails — the card contributes nothing, having no `no_commits` error for the other arm to catch |

A non-vacuous companion (same card, `noCommitsExpected: true`) rules out a sweep that flags every zero-commit card it finds — the declaration is the entire point of the flag.

`isBranchAheadOfBase` is mocked rather than spied: it is a static named import that shells out to git, so the ESM binding is already resolved before a spy could replace it.

## Verification

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed zero-commit audit detection for projects using renamed review lanes.
  - Correctly evaluates each completed card’s lane verdict, including cards with zero commits.
  - Preserves support for cards explicitly marked as not requiring commits.

- **Tests**
  - Added coverage for renamed review lanes and zero-commit audit scenarios.

- **Chores**
  - Updated release metadata and lifecycle tracking information.

### DIFF
--- a/.changeset/self-healing-no-commits-audit-query.md
+++ b/.changeset/self-healing-no-commits-audit-query.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: The zero-commit audit sees quietly-parked review cards again on renamed boards.
+category: fix
+dev: `auditNoCommitsExpectedCandidates` read the literal `in-review`, so on a renamed board only the `no_commits` error path fed the audit and a card sitting in a renamed review lane with zero commits and no error was never flagged. Read resolves via `resolveProjectColumnsForRoles`, the lane verdict resolves per card.
--- a/packages/engine/src/__tests__/self-healing-query-filter-blindness.test.ts
+++ b/packages/engine/src/__tests__/self-healing-query-filter-blindness.test.ts
@@ -37,6 +37,17 @@ import { EventEmitter } from "node:events";
 import type { Settings, Task, TaskStore } from "@fusion/core";
 import { resolveLifecycleColumns } from "@fusion/core";
 
+/*
+FNXC:WorkflowResolvedColumns 2026-07-31-15:35:
+`isBranchAheadOfBase` is a static named import that shells out to git, so it is mocked rather than
+spied — the ESM binding is already resolved by the time a spy could replace it.
+*/
+const isBranchAheadOfBase = vi.fn(async () => ({ aheadCount: 0 }));
+vi.mock("../self-healing-branch.js", async (importOriginal) => {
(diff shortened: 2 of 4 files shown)


gsxdsm:

gsxdsm:
**Rebase blocked on a conflict shape the earlier pattern does not cover — recording it precisely rather than forcing it.**

`shq19` and `shq20` rebased cleanly by hand because both sides of their test-file conflict were **complete, well-formed test blocks**, and the correct resolution was to keep both in order. `shq39` and `shq40` are different: their first conflict is in the **module-mock preamble**, where main and the branch each add a *different* `vi.mock(...)`:

- main mocks `branch-conflicts.js` (for `classifyForeignOnlyContamination`)
- this branch mocks `self-healing-branch.js` (for `isBranchAheadOfBase`)

Both are needed. The difficulty is that the conflict boundary cuts through the middle of the construct: main's `vi.mock(` opener sits *above* the conflict and its closing `});` sits *below* it, while the branch's side contains its own opener **inside** the conflict. So a union has to close main's mock, reopen a comment, and let the trailing `});` close the branch's — and both sides also begin mid-comment, so the comment delimiters have to be reconstructed too.

I attempted that join and the file came out **+2 unbalanced**. Rather than iterate on a fourth variant, I restored both branches to their origin state.

**Where the family stands:** 17 of 22 branches are current with main and green. Five are behind — `shq23`, `shq26`, `shq38`, `shq39`, `shq40` — of which `shq26` and `shq38` were pushed green earlier and have simply drifted again, which is the treadmill this family cannot win per-PR.

**`shq23` is a third distinct failure:** its rebase drops the `LEGACY_COLUMN_IDS_BY_ROLE` import from `self-healing.ts` and truncates the test file. Not the same shape as either of the other two.

Nothing broken was pushed at any point; every branch was verified before push and reset when it could not be. The consolidation remains the right answer — the source commits fold cleanly and only this one test file resists, which is precisely why sixteen branches sharing it cannot each be rebased into a moving main.

PR #2928, opened 2026-07-31
chore(census): an audited sentinel had its reasoning but not its marker

gsxdsm:
`audit-ops.ts` already carries a full FNXC note explaining that `state === "archived"` compares `getLiveTaskColumn`'s **normalized** return value — `"archived"` covers archived *and* soft-deleted — and that converting it would make a soft-deleted task's log **writable** on a renamed board. The audit was done, and it was right.

**It carried no `DELIBERATE-LITERAL` marker, so the census kept counting the line as owed work.**

That gap is the point of this change. A comment explaining why a site is correct does not reach the instrument — only the marker string does. So a careful audit was invisible to the one reader that acts on it, and the next person working the backlog would have re-derived the entire argument, or worse, converted it and shipped the defect the note warns about.

## What moved and what did not

| | before | after |
|---|---|---|
| `audit-ops.ts` counted | 2 | **1** |
| total backlog | 169 | 169 |

The remaining one is real: `pgRow.column === "archived"` at ~202 reads an actual DB column. Same shape as #2923 — **mark the sentinel, leave the lane**, so the count points at the line that is genuinely owed rather than at a pair where one is already settled.

The census classifies this as a **reclassification, not a conversion**, and required the baseline re-recorded in the same change. Nothing was fixed; a completed audit was made legible to the tool.

### DIFF
--- a/packages/core/src/task-store/audit-ops.ts
+++ b/packages/core/src/task-store/audit-ops.ts
@@ -125,7 +125,12 @@ export async function logEntryImpl(store: TaskStore, id: string, action: string,
           const layer = store.asyncLayer!;
           const state = await getLiveTaskColumn(layer.db, id, layer.projectId);
           /*
-          FNXC:WorkflowLifecycleColumns 2026-07-30-21:20 (audited — SENTINEL, do NOT convert):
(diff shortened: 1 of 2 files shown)


gsxdsm:

```bash
git cherry-pick <your-sha>
```

gsxdsm:

PR #2929, opened 2026-07-31
self-healing: no-progress failures kept their work slot on a renamed board (twenty-ninth sweep)

gsxdsm:
`recoverNoProgressNoTaskDoneFailures` requeues a wip card the executor failed for *"no fn_task_done"* that made **no step progress and left no git work** — nothing to salvage, so requeueing is safe. The literal read meant that on a renamed board it was never requeued.

The cost is doubled: the card sat failed *and* kept holding its wip slot, so the capacity was lost along with the task.

## No second pair

Verified with the derived ratchet added in #2879.

## Revert results

| conversion | reverted → |
| --- | --- |

Observable is `hasRecoverableGitWork`, private and called once per candidate before any requeue, so no git fixture is involved. A non-vacuous companion (same failure once the card reached review → not considered) rules out a read that returns everything: a card in review has handed off, and requeueing from there would undo a completed hand-off rather than rescue a stalled one.

The error string carries the **real** phrase `isNoTaskDoneFailure` matches; invented prose is filtered out and the case would pass reverted.

## Verification

### DIFF
--- a/.changeset/self-healing-no-progress-notaskdone-query.md
+++ b/.changeset/self-healing-no-progress-notaskdone-query.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Failed tasks that produced nothing release their work slot again on renamed boards.
+category: fix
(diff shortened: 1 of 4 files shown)


gsxdsm:

PR #2930, opened 2026-07-31
fix(missions): shipped-delivery repair refused a completed card on a renamed board

gsxdsm:
`getTerminalTaskEvidence` tested only `column === "done"`, so a genuinely completed card on a renamed board fell through every branch to `nonterminal`, and `reconcileFeatureDoneWithTerminalTask` threw:

```
TASK_NOT_TERMINAL: Delivery task KB-001 must be in done or supported archived state, not shipped
```

Valid operator work refused — and note the tell: **the message names the real column while the check cannot see it.** The classifier and the reporter disagreed, which is usually the cheapest signal that a lane literal is in the wrong place.

## I deferred this twice on a premise I never checked

Both of my earlier notes said `AsyncMissionStore` *"holds a layer, not a store"* and therefore could resolve nothing. It holds an **optional `taskStore`**, and the single production construction site supplies it:

```ts
store.missionStore = new AsyncMissionStore(layer, store);   // workflow-definitions.ts
```

That is the **third** deferral of mine this session to dissolve on inspection — after the planner overseer's cost argument (#2898, where the poll already awaited a per-task resolve) and the review-handoff seam's file conflict (#2900, where the blocking PR had merged). The pattern is consistent enough to state plainly: *my "blocked" notes have been wrong more often than right, and checking costs minutes.*

## The type had to move first

```ts
| { kind: "done"; id: string; column: "done" }        // ← pinned
| { kind: "archived"; id: string; column: "archived" }
```

The resolver could not report the real column without a compile error, so the **annotation would have forced the literal back in** after the predicate was fixed. This is the census-invisible type-annotation class blocking a conversion from the far end. `kind` already carries the role — that is what consumers switch on — so `column` is free to carry the truth.

## Compatibility

Absent a store the resolver is skipped and the legacy ids answer, which is exactly what every test constructing `AsyncMissionStore` without one already relies on — all 34 in `mission-store.pg.test.ts` stay green.

Deliberately calls `resolveProjectColumnsForRoles` directly rather than the `resolveArchivedLanes` helper, which lives on #2925 — keeping this PR independent of that one rather than stacking.

## Revert proof (measured, real PostgreSQL)

Restore `column === "done"`:

```
FAIL > accepts a completed card whose board calls the lane something else
TerminalTaskReconciliationError: Delivery task KB-001 must be in done or supported archived state, not shipped
```

## Verification

- `pnpm lint` — clean
- `mission-store.pg.test.ts` — 34 passed against real PostgreSQL
- SQL-literal gate green; census baseline tightened and committed

### DIFF
--- a/.changeset/terminal-evidence-lanes.md
+++ b/.changeset/terminal-evidence-lanes.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Mission delivery repair now accepts a completed card on boards that rename the done lane.
+category: fix
+dev: `getTerminalTaskEvidence` tested only `column === "done"`, so a completed card on a renamed board classified as `nonterminal` and `reconcileFeatureDoneWithTerminalTask` threw `TASK_NOT_TERMINAL`. It now takes resolved complete/archived lane sets, supplied by `AsyncMissionStore` from its `taskStore`. The `TerminalTaskEvidence` type's `column` field was widened from the pinned literals to `string`.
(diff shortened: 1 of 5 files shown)


gsxdsm:
Superseded by #2925 — folded per the queue freeze. Commit `a298683554` on `live-task-column-lanes`, cherry-picked onto current main and verified there (mission-store.pg.test.ts 34 passed, revert proof re-measured: restore `column === "done"` and the renamed-lane case fails with `TASK_NOT_TERMINAL: … not shipped`).

PR #2931, opened 2026-07-31
chore(census): five of six "archived" checks in this file are sentinels, not lanes (169 → 165)

gsxdsm:
`async-comments-attachments.ts` was the largest remaining non-engine entry in the backlog at **6**. Five of those six compare `getLiveTaskColumn`'s **return value**, not a column.

That helper normalizes: it manufactures the string `"archived"` for an archived row *and* for a soft-deleted one, and returns `null` for a missing task. The `column === null || column === "archived"` shape at two of the sites is the tell — nothing that reads a board column would ever compare it to `null`.

**Converting any of them would keep passing on the built-in board and start failing on a renamed one:** a soft-deleted task's documents would become writable, and the archived-document read paths would start returning rows they must not.

## The audit already existed and had never been applied here

`audit-ops.ts` records this exact distinction for its own copy of the check, and its note explicitly names *"the eight downstream comparisons in `async-comments-attachments.ts`, which are sentinels for the same reason."* The reasoning was done; it just never reached this file — the same gap as #2928, where the reasoning was present but the marker was not.

## What stays counted, and why that is the point

The **classifier** at ~142 (`row.column === "archived"`) reads a real DB column and remains in the backlog. That is the site reported on #2820, deferred because the helper takes a `db` handle with no task, no workflow and no lane vocabulary — and three of its callers run inside open transactions, where threading a store is what caused the #2821 deadlock.

So this file's count goes **6 → 1**, and the 1 is the line that genuinely blocks. A count of 6 implied six conversions were owed here; five of them were never work at all.

| | before | after |
|---|---|---|
| this file | 6 | **1** |
| total backlog | 169 | **165** |

A **reclassification, not a conversion** — the census enforces the distinction and required the baseline re-recorded in the same change.

### DIFF
--- a/packages/core/src/task-store/async-comments-attachments.ts
+++ b/packages/core/src/task-store/async-comments-attachments.ts
@@ -465,6 +465,21 @@ export async function listTaskDocuments(
   projectId?: string,
 ): Promise<TaskDocument[]> {
   const column = await getLiveTaskColumn(db, taskId, projectId);
+/*
+FNXC:LifecycleColumnCensus 2026-07-31-00:55 DELIBERATE-LITERAL: a SENTINEL, not a board lane.
+
+This compares `getLiveTaskColumn`'s RETURN VALUE, which is normalized — the helper manufactures the
+string "archived" for an archived row AND for a soft-deleted one, and returns `null` for a missing
+task. So the comparison is against that function's protocol vocabulary, not a column id.
+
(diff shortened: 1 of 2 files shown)


gsxdsm:

```bash
git cherry-pick <your-sha>
```

gsxdsm:

PR #2932, opened 2026-07-31
self-healing: partially-completed work was never retried on a renamed board (thirtieth sweep)

gsxdsm:
`recoverPartialProgressNoTaskDoneFailures` retries a review card failed for *"no fn_task_done"* that **did** make step progress — real work exists, so the sweep spends a retry rather than discarding it.

The literal read meant that on a renamed board the retry never fired: the work was parked failed with its **retry budget untouched**. That budget exists precisely to avoid losing partial work, so the safeguard and the work were lost together.

## No second pair

Verified with the derived ratchet added in #2879.

## Fixture note

One step `done` and another `pending`. All-done trips `isTaskWorkComplete` and none-done trips `hasStepProgress` — either would filter the card out for a reason unrelated to lanes, and the case would then pass with the fix reverted.

## Revert results

| conversion | reverted → |
| --- | --- |

Observable is `evaluateBackwardMoveTripleProof`, private and called once per candidate. A non-vacuous companion (same failure while the card is still in wip → no retry) rules out a read that returns everything: a card still in wip has not finished its attempt, so spending a retry there burns the budget on work that is still running.

## Verification

### DIFF
--- a/.changeset/self-healing-done-merge-metadata-query.md
+++ b/.changeset/self-healing-done-merge-metadata-query.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Completed tasks get their merge metadata repaired again on renamed boards.
+category: fix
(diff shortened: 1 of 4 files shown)


gsxdsm:

PR #2934, opened 2026-07-31
self-healing: completed tasks kept wrong merge metadata on a renamed board (thirty-first sweep)

gsxdsm:
`recoverDoneTaskMergeMetadata` repairs the merge metadata of a card that already reached the **complete** lane — the commit sha an operator sees, and that later reconcilers trust. The literal read meant that on a renamed board it was never repaired, so a completed task could keep pointing at a commit that is not the one that landed.

## Scoped to `complete`, deliberately not the terminal union

An **archived** card is out of scope here. Widening to `TERMINAL_ROLES` would start repairing metadata on rows nobody reads — a behaviour change wearing a conversion's clothes.

Contrast #2899 (`reconcileStaleMergerStatus`), which *does* use the terminal union: its filter genuinely does not distinguish complete from archived, so splitting them there would have invented a distinction. Here the sweep does distinguish, so the union would erase one. Same series, opposite call, and the difference is in what the code already says rather than in what is convenient.

## Revert results

| conversion | reverted → |
| --- | --- |
| the per-card verdict | fails — the renamed complete lane is filtered out |

Observable is `findLandedTaskCommit`, private and called once per candidate. A non-vacuous companion (same card in the review lane → untouched) rules out a read that returns everything: a card still in review has not landed, so "repairing" its merge metadata would invent an answer rather than correct one.

## Verification

### DIFF
--- a/.changeset/self-healing-done-merge-metadata-query.md
+++ b/.changeset/self-healing-done-merge-metadata-query.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Completed tasks get their merge metadata repaired again on renamed boards.
+category: fix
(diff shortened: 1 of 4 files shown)


gsxdsm:

PR #2935, opened 2026-07-31
chore(census): the two artifact/comment read-only guards are sentinels (3 → 1 in this file)

gsxdsm:
`task-artifacts-ops.ts` tests `getLiveTaskColumn(...) === "archived"` twice, to refuse comment writes on an archived parent. That helper **normalizes** — it manufactures `"archived"` for an archived row *and* a soft-deleted one, and returns `null` for a missing task, which is why the line immediately after each check tests `null` separately. Both compare that **protocol vocabulary**, not a column id.

An archived-role read here would keep passing on the built-in board and start **failing** on a renamed one: a soft-deleted task's comments would become writable.

Same class as the checks already marked in `audit-ops.ts` (#2928) and `async-comments-attachments.ts` (#2931). This is the fourth file in that family, and they all trace to the same helper.

## What stays counted

The `preArchiveColumn` comparison at ~435 reads a real column and remains in the backlog. **Mark the sentinel, leave the lane** — this file's count now points at the one line that is genuinely owed rather than at three where two were never work.

## Note on the running total

The headline number in these PRs drifts because other workers are landing conversions concurrently; the per-file movement is the stable claim. Here it is **3 → 1**. A **reclassification, not a conversion** — the census enforces the distinction and required the baseline re-recorded in the same change.

### DIFF
--- a/packages/core/src/task-store/task-artifacts-ops.ts
+++ b/packages/core/src/task-store/task-artifacts-ops.ts
@@ -655,6 +655,19 @@ export async function updateTaskCommentImpl(store: TaskStore, id: string, commen
     {
       const layer = store.asyncLayer!;
       const state = await getLiveTaskColumn(layer.db, id, layer.projectId);
+      /*
+      FNXC:LifecycleColumnCensus 2026-07-31-01:20 DELIBERATE-LITERAL: a SENTINEL, not a board lane.
+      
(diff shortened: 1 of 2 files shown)


gsxdsm:

```bash
git cherry-pick <your-sha>
```

gsxdsm:

PR #2936, opened 2026-07-31
fix(guard): my own PR put two FALSE entries in the unwired-lane guard — name the type

gsxdsm:
`unwired-lane-parameter-guard.test.ts` has been **failing on main** since #2875 merged, and the two entries it reports are wrong.

```
+ "packages/core/src/task-store/async-audit.ts completeColumns",
+ "packages/core/src/task-store/async-audit.ts reviewColumns",
```

Both **are** wired — `getInReviewDurationEventsImpl` resolves the lane sets and passes them. So a guard I wrote, to catch parameters nobody supplies, was reporting the opposite of the truth about my own change.

## Cause: owner-scoping cannot apply to a type the caller never writes

The rule I added earlier requires a mention to come from a file that also names the **declaring symbol**. Correct for a function — a caller has to name it to call it. Structurally impossible for an interface passed as an inferred object literal:

```ts
lanes = { reviewColumns: [...], completeColumns: [...] };   // wires it
getInReviewDurationEventsAsync(db, projectId, options, lanes);
```

Nothing there writes `InReviewDurationLanes`, and nothing ever will — the type comes from the parameter.

## Fixed at the caller, after trying the tool three ways

| attempt | result |
|---|---|
| relax type-owned properties to the original loose rule | hid **12** genuine entries |
| resolve owner to the functions consuming the type | hid **6** |
| annotate the local with the named type | **correct, one word, no rule change** |

Each heuristic refinement traded the false positive for false **negatives**. That is the usual sign a co-occurrence check has reached its limit — and an under-reporting guard is worse than an occasionally over-reporting one, because nobody notices the silence.

The other option was recording two *wired* parameters in `KNOWN_UNWIRED`. That would put non-debt in the debt list, which is how a ratchet starts lying — the same failure this PR is fixing, one layer up.

Naming the type is also just better code: the wiring is now visible to a reader, not only inferred.

## Verification

- `unwired-lane-parameter-guard.test.ts` — **9/9**, baseline unchanged at 17
- `pnpm lint` — clean
- `activity-log-parity.pg.test.ts` — 5 passed

## Summary by CodeRabbit

* **Refactor**
  * Improved internal type safety for in-review duration event processing without changing user-visible behavior.

### DIFF
--- a/packages/core/src/task-store/workflow-definitions.ts
+++ b/packages/core/src/task-store/workflow-definitions.ts
@@ -44,6 +44,7 @@ import { resolveSwitchReconciliation } from "../workflow-reconciliation.js";
 import { WORKFLOW_COMPILED_STEP_TEMPLATE_PREFIX } from "../store.js";
 import { resolveWorkflowIrForTask } from "../workflow-ir-resolver.js";
 import { resolveProjectColumnsForRoles, REVIEW_ROLES } from "../project-lane-vocabulary.js";
+import type { InReviewDurationLanes } from "./async-audit.js";
 
 export async function getAgentLogsByTimeRangeImpl(store: TaskStore,
     taskId: string,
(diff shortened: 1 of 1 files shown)


gsxdsm:
Superseded by #2925 — folded per the queue freeze. Commit `7e26a3a3bb` on `live-task-column-lanes`, cherry-picked onto current main and verified there (unwired-lane guard 9/9, baseline unchanged at 17).

Worth flagging for merge order: this is the commit that **un-reds main's `unwired-lane-parameter-guard` test**, which has been failing since #2875. It rides #2925 now.

PR #2937, opened 2026-07-31
self-healing: PR-conflict reclaim read a busy checkout as unowned on a renamed board (thirty-second sweep)

gsxdsm:
`reclaimPrConflictForTask` builds a worktree-owner map to answer *"is another live task holding this worktree?"* — the same question `findActiveWorktreeOwner` answers for the executor, and the same failure when the read comes back empty: **the checkout reads as unowned and the sweep reclaims a worktree another task is working in.**

## Read-only conversion

The map is keyed by **worktree path**, and nothing after it compares a column, so there is no pair to convert here. Confirmed with the derived ratchet added in #2879 rather than asserted from reading.

## Coverage, stated plainly: no behavioural test

This read sits after `inspectBranchConflict`, which needs a real git repo, so reaching it means a git fixture rather than a lane test.

Unlike the guards in #2867 and #2891, the blocker here is a **module-level import that shells out**, not a private method I could stub. I checked for that seam first — assuming the git fixture was unavoidable is exactly the mistake review corrected on #2867, and I would rather say "I looked and there isn't one" than repeat it.

So this one rests on `tsc` and the ratchet. That is weaker than the other thirty-one and I am not dressing it up; the alternative was leaving a known unowned-checkout hazard in place because its test would be inconvenient.

## Verification

`pnpm test:gate` 161 + 487 + 13 + 71, plus `self-healing.test.ts` + the blindness file + the ratchet, 428; `tsc` engine clean; `pnpm lint`, `check:changesets`, census `--strict` clean, each run explicitly.

### DIFF
--- a/.changeset/self-healing-pr-conflict-worktree-owner-query.md
+++ b/.changeset/self-healing-pr-conflict-worktree-owner-query.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: PR-conflict reclaim no longer reads a busy checkout as unowned on renamed boards.
+category: fix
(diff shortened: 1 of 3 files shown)


gsxdsm:

PR #2939, opened 2026-07-31
self-healing: workspace tasks stranded mid-land on a renamed board (sweeps 33 and 34)

gsxdsm:
The two **workspace** sweeps, converted together because they are one domain and one file region.

## `reconcileWorkspacePartialLands`

A workspace task lands **per-repo**. This re-enqueues one whose lands are partial or zero. The literal read meant that on a renamed board it never ran — so a task sat with *some repos merged and some not*, with nothing to finish the job.

That is the worst resting state anywhere in this series: not a stalled task but an **inconsistent** one.

## `reconcileOrphanedWorkspaceWorktrees`

Removes the per-repo worktrees a finished workspace task left behind. The literal read meant disk held by tasks that finished, growing quietly with no signal at all.

It resolves **`complete` only**, not the terminal union — its own comment calls done tasks *"the canonical safe to clean set"* because their lands are finalized, and an archived row is a different claim. Same call as #2934, opposite of #2899; each time the existing code already said which it meant.

## Revert results

| conversion | reverted → |
| --- | --- |
| partial-land read + verdict | that case fails — the card is never listed |
| orphaned-worktree read | that case fails — the done card is never listed |

The partial-land observable is `isWorkspaceTaskLive`, private and called once per surviving candidate before any git or disk work. The cleanup sweep's disk work is guarded by `isPathActive` and real fs checks, so its assertion is on the **query** — sound here specifically because nothing downstream vetoes on lane (the only filter is `isWorkspaceTask`), which is the condition the sibling doc sets for when a query assertion is honest rather than lazy.

## Verification

`pnpm test:gate` 161 + 487 + 13 + 71, plus `self-healing.test.ts` and `workspace-self-healing.test.ts`; `tsc` engine clean; `pnpm lint`, `check:changesets`, census `--strict` and `check-sql-column-literals` clean, each run explicitly.

### DIFF
--- a/.changeset/self-healing-workspace-lane-queries.md
+++ b/.changeset/self-healing-workspace-lane-queries.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Workspace tasks finish their partial lands and release their worktrees on renamed boards.
+category: fix
+dev: `reconcileWorkspacePartialLands` and `reconcileOrphanedWorkspaceWorktrees` read the literal `in-review`/`done`, so on a renamed board a workspace task stranded mid-land was never re-enqueued and a finished one never released its per-repo worktrees. Reads resolve via `resolveProjectColumnsForRoles` (review roles, and `complete` only for the cleanup); the partial-land per-card check resolves per card.
--- a/packages/engine/src/__tests__/self-healing-query-filter-blindness.test.ts
+++ b/packages/engine/src/__tests__/self-healing-query-filter-blindness.test.ts
@@ -899,4 +899,71 @@ describe("self-healing sweeps are bounded by a hardcoded column QUERY, not by th
 
     expect(classifyForeignOnlyContamination).not.toHaveBeenCalled();
   });
+
+  /*
+  FNXC:WorkflowResolvedColumns 2026-07-31-18:15 (the query-filter class, sweeps thirty-three and -four):
+  The two WORKSPACE sweeps. A workspace task lands PER-REPO, so its failure modes are its own: a
+  partial land leaves some repos merged and some not, and a finished one leaves per-repo worktrees on
+  disk. Both were bounded by literal reads, so on a renamed board neither ran — the partial land never
+  finished, and the disk was never reclaimed.
+
+  `isWorkspaceTaskLive` is what the partial-land sweep calls once per surviving candidate, before any
+  git or disk work — the private-seam technique, so no workspace fixture is needed.
+
+  REVERT CHECKS, both measured, each alone:
+    - partial-land literal read + verdict restored -> that case fails, the card is never listed
+    - orphaned-worktree literal read restored -> that case fails, the done card is never listed
+  */
+  function workspaceCard(id: string, column: string): Task {
+    return {
+      ...shippedCard(),
+      id,
+      column,
+      workspaceRepos: [{ path: "repo-a" }],
+      workspaceWorktrees: { "repo-a": { worktreePath: "/tmp/ws/repo-a" } },
+      mergeDetails: {},
+    } as unknown as Task;
+  }
+
+  it("re-enqueues a partially-landed workspace task on a RENAMED review lane", async () => {
+    const { store } = productionFaithfulStore([workspaceCard("FN-WSPARTIAL", RENAMED_VOCAB.review)]);
+    const manager = new SelfHealingManager(store, { rootDir: "/repo" });
+    const isWorkspaceTaskLive = vi.fn(() => ({ live: true, livePaths: ["/tmp/ws/repo-a"] }));
+    Object.assign(manager, { isWorkspaceTaskLive, emitWorkspacePartialLandNoAction: vi.fn(async () => undefined) });
+
+    await manager.reconcileWorkspacePartialLands();
+
+    expect(isWorkspaceTaskLive).toHaveBeenCalledWith(expect.objectContaining({ id: "FN-WSPARTIAL" }));
+  });
+
+  it("does not re-enqueue a workspace task still in the RENAMED wip lane", async () => {
+    /*
+    Non-vacuous companion: a workspace task in wip has not started landing, so there is no partial land
+    to finish — the comment at the site says execution-stage reconcilers own that lane.
+    */
+    const { store } = productionFaithfulStore([workspaceCard("FN-WSPARTIAL", RENAMED_VOCAB.wip)]);
+    const manager = new SelfHealingManager(store, { rootDir: "/repo" });
+    const isWorkspaceTaskLive = vi.fn(() => ({ live: true, livePaths: ["/tmp/ws/repo-a"] }));
+    Object.assign(manager, { isWorkspaceTaskLive, emitWorkspacePartialLandNoAction: vi.fn(async () => undefined) });
+
+    await manager.reconcileWorkspacePartialLands();
+
+    expect(isWorkspaceTaskLive).not.toHaveBeenCalled();
+  });
+
+  it("cleans orphaned workspace worktrees for a card on a RENAMED complete lane", async () => {
+    const { store } = productionFaithfulStore([workspaceCard("FN-WSDONE", RENAMED_VOCAB.complete)]);
+    const listTasksSpy = (store as unknown as { listTasks: ReturnType<typeof vi.fn> }).listTasks;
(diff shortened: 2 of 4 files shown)


gsxdsm:
**For whoever owns the self-healing family fold (item 1) — I attempted it and it does not fold by cherry-pick. Measured, so you do not spend the time rediscovering it.**

I am not the capacity worker; I took this because no `batch-self-healing-*` branch existed and the family is ~60% of the open queue. I stopped rather than push a partial.

## What I measured

Fresh branch off `origin/main`, cherry-picked all 23 open family PRs in ascending order (`shq18`…`shq44`, PRs #2867–#2939):

**3 of 23 folded. 20 conflicted** — `shq18 shq19 shq20 shq21 shq23 shq24 shq26 shq27 shq28 shq29 shq33 shq34 shq35 shq37 shq38 shq39 shq40 shq41 shq42 shq44`.

The conflicts are in `self-healing.ts` itself, not in the census baseline — so the auto-resolve that worked for the census family (regenerate the derived file) does not apply here.

## Why, and it is not what the brief assumes

The brief reads *"they share a root cause and mostly touch self-healing.ts — sixteen CI runs for one file is indefensible"*. The waste is real, but the fix is not cheap, because **the branches are INDEPENDENT, not stacked**:

- `origin/main..shq44` = 1 commit; `origin/main..shq18` = 3 commits
- `git merge-base --is-ancestor shq43 shq44` → **not an ancestor**

So each branched from main separately and edits overlapping regions of the same file. Twenty-three parallel edits to one file is 253 possible conflict pairs; cherry-picking them in any order hits a wall almost immediately.

## What will actually work

Not cherry-pick. Either:

1. **Sequential rebase** — take `shq18` as the base, rebase `shq19` onto it, then `shq20` onto that, resolving each against the accumulating state. Twenty resolutions, but each one is small and against a known-good predecessor.
2. **Re-apply fresh** (probably cheaper) — the author reads the 23 diffs and writes the combined change onto main in one pass. These are 23 instances of one pattern; written together they are likely far less code than 23 independently-guarded edits, and the result is reviewable as a single change.

Option 2 is what I would do. Option 1 preserves authorship trails if that matters.

## The general rule this exposes

**A family only folds by cherry-pick if its members touch disjoint regions.** Mine did — portal fixes were three separate test files (#2940), census sentinels were separate source files with one shared derived artifact (#2943, and regenerating it was the whole trick). This family is 23 edits to one file, which is the opposite case. Worth checking region-disjointness before assigning a fold, otherwise the assignee discovers it the expensive way.

gsxdsm:
**Fold recipe for `batch-self-healing-renamed-boards`, measured — the premise in the freeze order is wrong about where the contention is, and that changes the approach.**

I probed all 22 open `self-healing … renamed board` PRs against current `main` before touching anything, then threw my attempt away (details at the end). The measurements:

**1. `self-healing.ts` merges CLEAN across every branch.** Zero conflict hunks in it, in every combination I tried. The freeze order's *"they mostly touch self-healing.ts — sixteen CI runs for one file"* is right about the CI waste and wrong about the blocker. Nobody has to hand-merge the product file.

**2. Naive `cherry-pick` fails 20 of 22.** Each branch was cut from a different `main`, so the picks collide pairwise. That is almost certainly why this fold has not started.

**3. Sequential `git merge` conflicts in exactly two files, both mechanical:**

| file | resolution |
|---|---|
| `__tests__/self-healing-query-filter-blindness.test.ts` | **union** — every conflict is an append (a new `it(...)`, a mock method, an FNXC block). Take both sides. |
| `scripts/lib/lifecycle-column-census-baseline.json` | **regenerate, never merge** — it is generated. `node scripts/lifecycle-column-census.mjs --update-baseline` after each merge. Same for `sql-column-literals-baseline.json`. |

So the loop is: `git merge origin/shqNN` → union-resolve the test file → regenerate both baselines → commit. That is formulaic, and 22 of those beat 22 CI runs.

**Order matters:** `shq26` was the only branch that refused for a reason outside those two files when I ran it mid-sequence — worth folding last and looking at on its own.

**What I did not do, and why you should not trust a fold you did not run.** I automated the loop and it produced a branch with **conflict markers committed into the baseline JSON** — the exact defect the `census-baseline-corruption-guard` test exists to catch, and the second time this program has hit it. My shell loop word-split on a multi-line filename list and silently mis-staged. I reset it and pushed nothing; there is no `batch-self-healing-renamed-boards` branch from me. Do the merges with the loop *unrolled*, or at minimum grep for `^<<<<<<<` across `packages` and `scripts` before every commit.

This is your family — I stayed off it beyond the probe. Posting here because the recipe is the part that was expensive to find.

gsxdsm:
**Rebased and pushed green this turn — and already behind again, which is the actual finding.**

I rebased and pushed **19 of the 21 open `shq*` branches** in this family, each verified individually (engine `tsc` clean, `self-healing-query-filter-blindness.test.ts` 20–26 passing per branch). By the time the last one landed, `main` had advanced and **all 21 were behind once more.** Per-PR rebasing cannot converge here — the family shares one file, so every merge into main invalidates every other branch's rebase.

**I attempted the consolidation the standing order asks for, and it failed three times.** All 15 source-side commits (`self-healing.ts`) cherry-pick cleanly onto a single `batch-self-healing-renamed-boards` branch — that part works. The blocker is `self-healing-query-filter-blindness.test.ts`, which every branch appends a test block to:

1. **Naive conflict splice** — the conflict boundary falls *inside a comment block*, so concatenating both sides drops a test's closing `});` and the next one's opening `/*`. Produced 5 damaged seams.
2. **Rebuild by test name** — recovered all 27 unique blocks, but each branch also adds *preamble* helpers (`createRunAuditor` mocks and similar) that a body-only rebuild leaves undefined.
3. **Rebuild with a preamble line-union** — comment bodies differ between branches, so unioning preamble lines injects stray comment fragments and breaks the parse.

**Nothing broken was pushed.** The batch branch stayed local and I restored the worktree after each attempt; the 19 branches I did push were each verified before pushing, and the two I could not verify (`shq19`, `shq20`) were left untouched at their previous state rather than pushed red.

**What this needs:** the test file merged with per-conflict judgement — 21 blocks placed by hand once, not scripted. That is bounded work but it is not automatable, and I would rather hand it over accurately than produce a fourth broken file. Everything else in the family is ready: the source changes fold without conflict.

gsxdsm:
**Fold intelligence for the `shq*` family** — I attempted the consolidation, could not finish it, and this is what I measured so you do not have to rediscover it.

I have no PRs in this family; I took it because no collection branch existed and it is the largest remaining block (~22 open). Everything below is verified, not guessed.

### Naive cherry-pick does NOT work

Cherry-picking every commit from shq18/19/20/21/23/26 onto a fresh branch off `main`:

```
applied=4  conflicted=13      (76% conflict rate)
conflicting: shq19 shq20 shq21 shq23 shq26
```

Each branch was written against a different `main`, and they overlap in `self-healing.ts`. Any fold script that cherry-picks onto a fixed base will fail this way.

### Sequential rebase DOES work, and the conflict is narrow

Rebasing `shq19` onto a branch that already contains a folded `shq18`:

```
exit=1
packages/engine/src/__tests__/self-healing-query-filter-blindness.test.ts
```

**One file.** Not `self-healing.ts` — the production edits sit in different sweeps and merge cleanly once the base is right. The conflict is the shared test file that every PR in this family appends a case to, so it is a mechanical keep-both-sides resolution, roughly 21 times.

### Recommended procedure

1. `git checkout -B batch-self-healing-renamed-boards origin/main`
2. For each branch in ascending order: `git checkout -B wip-<b> origin/<b> && git rebase batch-self-healing-renamed-boards`
3. On the single test-file conflict, keep **both** sides' cases, then `--continue`
4. `git checkout batch-self-healing-renamed-boards && git merge --ff-only wip-<b>`
5. After each fold, assert the test-case count in that file **increased** — that is the guard against silently dropping one of your cases

### Why I stopped

I scripted step 3 and my rebase-state detection was wrong, so the run failed on all 22 with nothing folded. Rather than keep iterating on automation that manipulates 22 branches of your test cases — where a bad resolution silently deletes a case and every suite still passes — I aborted, deleted every trial branch, and pushed nothing. Tree is clean and no `batch-self-healing-*` branch exists on the remote.

This wants a human-guided pass by whoever can recognise a missing case on sight. The method above is sound; only my automation of it was not.

gsxdsm:

PR #2933, opened 2026-07-31
chore(census): commit the tightened lifecycle baseline — the tool asked for it

gsxdsm:
One line. `lifecycle-column-census --strict` on `main` prints:

```
lifecycle-column-census --strict: baseline TIGHTENED — the tree has fewer guards than it allowed
The baseline file has been rewritten downward. COMMIT IT so the allowance cannot be regrown into;
```

…and rewrites the file. Nobody had committed the rewrite, so the allowance stayed one entry wider than the tree:

```
packages/engine/src/workflow-node-handlers.ts: 1 -> 0   (entry removed)
```

That guard was converted by **#2900** ("the review handoff killed the walk on a renamed review lane"). Confirmed the file has no legacy guard left — the census report no longer mentions it at all.

This is the residual the census names in its own source: *in CI the tightening write is discarded with the runner, so the committed baseline stays stale until someone commits a tightened one.* This is someone committing it. After the commit, a second `--strict` run is clean and exits 0.

## How I found it, which is the part worth recording

My routine "verify main's gates" check runs the census. The census rewrites this file every time it tightens. For several turns I saw `scripts/lib/lifecycle-column-census-baseline.json` show up as modified in my worktree, assumed it was leftover state from an earlier probe of mine, and quietly `git checkout --`-ed it away.

**It was the tool telling me main's baseline was stale — once per turn — and I kept discarding the message.** A dirty file in a worktree reads as noise; that one was a report.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2941, opened 2026-07-31
batch-sql-ratchet: the census / gate-ratchet family — collection branch, fold here

gsxdsm:
## Family branch for consolidation directive item 4

`batch-sql-ratchet` did not exist and ~10 open PRs are waiting for a collection point, so this establishes it. **Fold your census/ratchet commit here and close your own PR as superseded.**

```bash
git cherry-pick <your-sha>
# verify scoped, not full suite:
pnpm --filter @fusion/core exec vitest run src/__tests__/archived-column-gate-parity.test.ts --silent=passed-only --reporter=dot
```

**Candidates I can see open right now** (owners: please fold + close):

| PR | branch |
|---|---|
| #2938 | `fix/comments-ops-sentinel` |
| #2935 | `fix/task-artifacts-sentinels` |
| #2933 | `chore/commit-tightened-census-baseline` |
| #2931 | `fix/async-comments-sentinels` |
| #2928 | `fix/audit-ops-sentinel-marker` |
| #2925 | `live-task-column-lanes` |
| #2923 | `fix/task-id-integrity-sentinel` |
| #2921 | `fix/plugin-store-migration-marker` |
| #2894 | `gate/sql-literals-match-census-placement` |

That is **10 → 1** once folded. I have not cherry-picked anyone else's commits — folding someone's work without them verifying it is how a batch lands broken.

---

## What is in it so far (mine, from #2924)

**Clears a live main red:** `archived-column-gate-parity` fails on `origin/main` today.

```
AssertionError: TypeScript encoding changed.
  async-comments-attachments.ts: 8 → 5
```

#2886 fixed a real bug — archived-document guards failing in *opposite* directions on a renamed lane — by replacing three `column === "archived"` comparisons with `isArchivedLane(column, archivedColumns)`. The AST scan counts raw comparisons, so the tally dropped.

**What I did not do is record it as three sites converted**, because measured, it is not:

```
→ (no matches)
```

No caller passes it. The parameter defaults to `LEGACY_ARCHIVED_LANES = new Set(["archived"])`, so every call resolves to the literal it replaced — byte-identical behaviour, resolved branch dead.

That matters for this guard's whole argument: its header warns that converting the TypeScript half while the Drizzle and raw-`sql` halves still compare the string is a split brain *"no test would catch, because every builtin workflow spells the column `archived` so the two halves agree by accident on every board we ship."* **There is no split brain today precisely because the resolved half is unwired** — it becomes one the moment a caller threads real lanes in without the SQL sides moving. Recorded inline so `5` cannot be read as "3 sites done"; flagged on #2886.

Verified not a split brain: the Drizzle and raw-sql inventories are unchanged and both pass — worth stating because those assertions run *after* the TypeScript one, so a plain red says nothing about them.

Scoped edit to `AUDITED_TS_SITES` by line range: these paths appear in more than one inventory here, and an unscoped replace would quietly edit the raw-sql side too, making the parity guard agree with itself (the trap I hit in #2817).

Guard still bites: appending a real `task.column === "archived"` to an audited file fails it. Core **4852 passed / 0 failed**, lint clean, test-only.

Closing #2924 as superseded by this.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved task delegation messages when workflow pickup cannot be confirmed.
  * Delegation results now clearly indicate when a task has not been verified for pickup.

* **Quality Improvements**
  * Added validation checks to catch future-dated markers and inconsistent SQL-column usage.
  * Refined workflow checks to distinguish stale configuration from incomplete configuration.

* **Documentation**
  * Updated lifecycle conversion guidance with more accurate audit findings and limitations.


gsxdsm:
Folded onto the existing family branch rather than pushing a competing one — **#2941 was opened against a branch another worker had already claimed**, so my commits are now rebased on top of theirs (`524df3dc1e`, the archived-gate inventory) rather than replacing it.

Closed as superseded and folded here: **#2887, #2894, #2926, #2933.**

Two notes for whoever merges:

**One commit added during the fold.** The census tightened *again* on this branch — `self-healing.ts` 87→86 and 35→34 — because main has taken more conversions since #2933 was written. Re-recorded so the batch lands a baseline matching the tree it merges into. `baselines dirty: 0` after a `--strict` run confirms it.

**Two commits here are not mine and not this family.** `fix(delegate): an unverified landing now says so…` and its follow-up modify `packages/cli/src/extension.ts` (+47). They were pushed onto #2894's branch by someone else, and this batch became their only home — dropping them would delete work rather than consolidate it. They can be split out if you prefer; `extension.test.ts` passes 75/171.

Verification, scoped per the consolidation rules: `sql` 0 · `seam` 0 · `census --strict` 0 with 0 baselines dirty · `tsc -p packages/core` 0 · lint 0.

gsxdsm:
Folded the gate / ratchet family into this collection branch — 8 commits from three PRs, authorship preserved.

| folded PR | brings |
|---|---|
| #2933 | commit the tightened lifecycle baseline the tool asked for |
| #2917 | freeze the future-dated FNXC stamp surface (206 stamps) |
| #2894 | move the SQL ratchet + inert-seam check to where the census runs; STALE seam exemption warns instead of failing; two `fix(delegate)` commits that shipped in that PR |

**Nothing dropped.** #2894's two `fix(delegate)` commits are unrelated to the gate but rode in that branch, so they are folded rather than orphaned — flagged rather than silently split.

## Two conflicts, resolved by combining intents

`package.json` conflicted twice: #2917 *adds* a check to `pretest`/`test:gate` while #2894 *removes* two others from the same lines. Neither side taken wholesale — SQL ratchet and inert-seam check removed from local runs (#2894's intent), fnxc gate kept in `pretest` and `test:gate` (#2917's intent).

**Verified no enforcement was lost.** My first check grepped workflows for script *filenames* and found none, which looked like the fold had disabled two gates. They are wired by alias:

```
.github/workflows/pr-checks.yml:60:  run: pnpm check:sql-column-literals
.github/workflows/pr-checks.yml:62:  run: pnpm check:inert-flag-seams
```

## The fold caught a defect neither PR could see alone

With #2917's gate active the batch failed:

```
scripts/check-sql-column-literals.mjs: 1 future-dated FNXC stamp(s), baseline allows 0
```

**That stamp is mine** — the FNXC comment from my #2888 reads `2026-08-02` against a real date of `2026-07-31`. #2917's baseline predates #2888 merging, so only the fold surfaces the conflict. Corrected to the real date in its own commit rather than re-recording the baseline: the gate is right, my stamp was wrong.

## Verification (scoped)

`test:gate` exit 0 · `check-sql-column-literals` 0 · `check-fnxc-future-dates` 0 (was 1) · `check-inert-flag-seams` 0 · `check:lifecycle-columns` 0 · `lint` clean · `package.json` parses

Closing #2933, #2917 and #2894 as superseded.

gsxdsm:
Re-folded two commits that went missing when this branch was rebuilt — flagging because I had already closed their PRs, so they were briefly orphaned.

The branch was reset at some point to add the FNXC-stamp work (`9b91027d91`, `e7cd36fcd9`). That rebuild dropped:

- `docs(core): mark the three archived STATE literals as deliberate` (was #2887)
- `docs(workflow-learnings): date my measured claims, and correct one that was wrong` (was #2926)
- `docs(gate): date the grep-vs-AST measurement in the SQL gate's header`

All three are back on the branch now, verified present by subject before re-applying so nothing got duplicated.

**The near-miss is worth naming.** I closed #2887 and #2926 as "superseded by #2941" *before* confirming they survived a later rebuild of the family branch. Closing an original the moment it is folded is the process the consolidation asks for, and it silently depends on the family branch never being rebuilt without those commits. It was, and for a while that work existed nowhere — not in an open PR, not on main, not on the batch.

If other folded PRs were closed the same way, they are worth re-checking against the current branch tips rather than the ones they were folded onto.

Also re-recorded the census baseline: it tightens on every `--strict` run as main takes more conversions, so the batch would otherwise land one wave behind.

Verified after re-fold: `sql` 0 · `seam` 0 · `census --strict` 0 with 0 baselines dirty · `tsc -p packages/core` 0 · lint 0.

gsxdsm:
### Re-folded a commit that a force-push dropped, and added a second main-red fix

**Heads-up for whoever is also pushing here:** the commit this branch was created with —
`524df3dc1e` (archived-gate inventory, #2924) — was **not reachable** from the branch after the last force-push:

```
git merge-base --is-ancestor 524df3dc1e origin/batch-sql-ratchet  → NO
```

I have cherry-picked it back **on top** of your 12 rather than force-pushing over them, so nothing of yours moved (14 commits now, plain fast-forward). Worth knowing because it is silent: the branch stayed green and the PR still looked right, but the main red it fixes was no longer fixed. If you rebuild this branch, please rebase rather than reset, or check `git log --oneline origin/main..HEAD` for commits you did not write.

---

### Second fix added: `unwired-lane-parameter-guard` (main red)

```
These declarations take a resolved lane answer that NO production file supplies.
+ packages/core/src/task-store/async-audit.ts completeColumns
+ packages/core/src/task-store/async-audit.ts reviewColumns
```

That guard's message is explicit — *"one arriving is a regression — update the list only to shorten it"* — so appending them was not an option.

**They are not unwired.** `getInReviewDurationEventsImpl` (`workflow-definitions.ts:1009`) resolves both roles through `resolveProjectColumnsForRoles` and passes them; that is #2875's renamed-board fix for the Reliability duration metric working as intended.

The guard was right about its rule and wrong about the code. For an exported interface member its `owner` is the **interface name**, and a mention only counts from a file that also names that owner (`unwired-lane-parameter.mjs:175`). The caller typed the value **inline** — `{ reviewColumns?: string[]; completeColumns?: string[] }` — so the one file that *does* wire them never spelled `InReviewDurationLanes`.

A false positive on wired code is worse than a miss, because the plausible-looking fix is to extend a one-directional ratchet. Importing the exported type is smaller and truer, and it means a future change to the parameter set is a compile error at the call site rather than a silently-ignored extra property.

**Measured:** guard **9/9** with `KNOWN_UNWIRED` **unchanged** — that is the point of the fix; parity test **2/2**; core `tsc` 0 errors; lint clean.

### Also checked and NOT acted on

`lifecycle-column-census.test.ts` showed 2 failures in one full-engine run on main. They pass **40/40** on pristine main and on this branch — suite-only/order-dependent, not deterministic. Added to the intermittent tally on #2862 rather than "fixed" by chasing a ghost.

gsxdsm:
Recovered a second orphaned commit — this one was not mine, and it was the branch's own original base.

**#2924** (`fix(tests): archived-gate inventory after #2886 — the count moved, the behaviour did not`) was closed as folded, and its commit `d226e5372a` was in **no batch branch and not on main**. It was `524df3dc1e` — the commit this family branch *started* from — and the same rebuild that dropped three of my commits dropped it too.

Re-folded as `ea86261669`. `archived-column-gate-parity.test.ts` passes; `sql` 0, `seam` 0, `tsc -p packages/core` 0, lint 0.

## I audited the rest rather than assuming this was the only one

Checked every PR closed since 2026-07-31 that looked folded, against the **current** batch tips:

```
#2938 OK -> batch-census-sentinels     #2917 OK -> batch-sql-ratchet
#2935 OK -> batch-census-sentinels     #2913 OK -> batch-portal-test-fixes
#2931 OK -> batch-census-sentinels     #2911 OK -> batch-portal-test-fixes
#2928 OK -> batch-census-sentinels
#2921 OK -> batch-census-sentinels     #2924 *** orphaned *** (now recovered)
```

So the loss is bounded: **one PR besides my own three commits**, and both losses came from the same branch rebuild. The census-sentinel and portal families are intact.

## The process gap this exposes

"Fold, then close the original" is only safe while the family branch is append-only. A rebuild that reconstructs the branch from a subset silently deletes work whose only other home was a closed PR — and nothing surfaces it, because the PR is closed and the branch is green.

Cheap mitigation for the remaining folds: before closing, or after any rebuild, check the subject is present at the branch tip. That is what found both of these.

gsxdsm:
Rebased onto current \`main\`; the branch was DIRTY.

One conflict, on \`scripts/lib/lifecycle-column-census-baseline.json\` — `main` re-recorded it while this batch carried its own re-record commit. Resolved by **regenerating** rather than picking a side: took main's file, then re-ran `lifecycle-column-census.mjs --strict --update-baseline` against the folded tree so the baseline describes *this* tree rather than either parent's. Picking a side would have left an allowance that matches neither.

Re-verified after the rebase:

- `pnpm test:gate` — exit 0
- `check-sql-column-literals` — exit 0
- `check-fnxc-future-dates` — exit 0
- `check-inert-flag-seams` — exit 0
- `check:lifecycle-columns` — exit 0
- `pnpm lint` — clean

10 commits, still folding #2933, #2917 and #2894 with nothing dropped.

gsxdsm:
### ⚠️ My two commits have now been dropped from this branch TWICE by force-pushes

Re-folded again just now (`2a3ddb9a83`, `86a15dfaa9`, cherry-picked on top — your 10 are untouched). But this will keep happening, so here is the mechanism and what I need from whoever owns the pushes.

**Evidence, both times:**

```
after push #1:  git merge-base --is-ancestor 524df3dc1e origin/batch-sql-ratchet → NO
after push #2:  git log --oneline origin/main..origin/batch-sql-ratchet  → 10 commits, neither of mine
```

**Why it recurs:** the branch is being rebuilt by rebasing a *local* branch that never contained my commits, then force-pushed. Anything I add is discarded on the next rebuild, regardless of when I push. Me re-folding a third time does not fix that — it just resets the clock.

**What this costs:** both commits clear **live main reds**, verified on `origin/main` at `c6767cb258`:

```
core:    archived-column-gate-parity   1 failed | 4852 passed
engine:  unwired-lane-parameter-guard  1 failed | 11058 passed
```

Those are the *only* two reds on main right now. While they are missing from this branch, the family PR looks green and correct while leaving main red — which is exactly the failure mode that makes a dropped commit dangerous rather than merely annoying.

**Ask, in order of preference:**

1. **`git pull --rebase` (or fetch + rebase onto `origin/batch-sql-ratchet`) before you push**, rather than force-pushing a locally-rebuilt branch. Then folds from other workers survive.
2. If you must rebuild, **check `git log --oneline origin/main..HEAD` for commits you did not write** before force-pushing, and carry them.
3. If you would rather own this branch exclusively, say so and I will take my two commits elsewhere — that is fine, but they cannot simply vanish, because main stays red.

Both verified on this branch as pushed: parity **2/2**, guard **9/9** with `KNOWN_UNWIRED` unchanged, core `tsc` 0 errors, lint clean.

gsxdsm:
**Overlap flag for whoever merges these two batches.**

`batch-sql-ratchet` (#2941) and `batch-docs` (#2942) now share **three identical commits**:

```
docs(gate): date the grep-vs-AST measurement in the SQL gate's header
```

I folded them into `batch-docs`; someone subsequently folded the same three into `batch-sql-ratchet`. Both branches are collection branches, so this is a natural consequence of two people folding overlapping sets — not a mistake by either.

**Two commits are unique to `batch-docs`** and exist nowhere else, so #2942 cannot simply be closed:

```
docs(workflow-learnings): correct the recorded date — 2026-07-31 is tomorrow
```

**Merging order matters.** Whichever lands second will find its three copies already applied — expect empty or trivially-conflicting patches there, and drop them rather than re-applying. Nothing is lost either way; the content is byte-identical.

**Separately, #2933's baseline commit is absent from `batch-sql-ratchet` and that is correct, not a loss.** I verified on the branch: `lifecycle-column-census.mjs --strict` exits 0 with "every file matches its baseline exactly", so the tightened baseline it carried is already in effect — it was absorbed when I regenerated the baseline during the rebase rather than dropped.

I have not rewritten either branch to de-duplicate: both are shared collection branches other workers are pushing to, and a force-push to fix a cosmetic overlap risks clobbering a fold in progress. Flagging is the cheaper correction.

gsxdsm:
### Cross-batch check: this family is compatible with the others in flight

The six family PRs are each green in isolation, but nothing was checking whether they are green **together** — and two batches can conflict semantically without touching a single common file. This one is a real risk here because `batch-census-sentinels` (#2943) edits `async-comments-attachments.ts`, whose raw-comparison count my archived-gate inventory pins.

Merged locally and re-ran the gates that span both:

| combination | result |
|---|---|
| #2941 + #2943 (`batch-census-sentinels`) | 0 conflicts · `archived-column-gate-parity` **2/2** |
| ↳ same tree, census gates | `--strict` exit **0** · census + corruption-guard + unwired-guard **52/52** |
| #2941 + #2940 (`batch-portal-test-fixes`) | 0 conflicts |
| #2943 + #2944 (`batch-self-healing-renamed-boards`) | 0 conflicts (auto-merges the census baseline) |

So the merge order does not matter for these four, and the archived inventory value in this PR survives #2943's edits to the same file.

**Why it was worth checking rather than assuming:** `git diff --name-only` reports **no file overlap** between #2941 and #2943. The hazard was not a textual conflict — it was that my inventory records a *count* of comparisons in a file the other batch modifies. A count pinned in one PR and the code it counts changed in another is invisible to git and to both PRs' CI, and would land as a fresh main red attributed to whichever merged second. That is the same shape as the four census reds already fixed today (#2814, #2817, #2856, and this one), each caused by a derived number and its source moving in different commits.

The census baseline itself auto-merges cleanly between #2943 and #2944, which is the other file where I would have expected trouble.

Not re-verified here: `#2925` (`batch-u7-lane-fixes`) and `#2942` (`batch-docs`) — docs and lane fixes that do not touch the census/archived surfaces.

gsxdsm:
Rebased onto `origin/main` (36625203b9). **12 commits → 11**, because one of them was independently landed on main and is now redundant.

The conflict was in `workflow-definitions.ts`, and both sides make the **same fix for the same reason**: name the `InReviewDurationLanes` type instead of typing it inline, so `unwired-lane-parameter.mjs` can see that the lanes *are* wired. That guard keys on the owner symbol, so an inline structural type meant the one caller that does wire them never spelled the interface name — a **false positive on correct code**, which is worse than a miss, because the plausible-looking response is to append two entries to `KNOWN_UNWIRED`, a list that may only ever shorten.

I took main's version. Mine differed only in comment wording, and its extra detail — that I tried three heuristic variations first, each trading the false positive for false *negatives*, the widest hiding twelve genuine entries — is already recorded on #2936 where that analysis was done. No reason to carry a second copy.

Verified on the rebased tip rather than assumed: core `tsc` clean, `pnpm lint` clean, and all three ratchets this batch touches exit 0 — `check-sql-column-literals`, `lifecycle-column-census --strict`, `check-inert-flag-seams`. Gate green (161/487/13/71).

gsxdsm:
### Retraction: my "commits dropped a third time" alarm was wrong — and the reason is worth reading

I had a watch on this branch after two genuine force-push drops. It fired again with *"only 1 of my 2 commits remain"*. **That one was a false alarm, and the fault is mine.**

The watch counted commits in `git log origin/main..origin/batch-sql-ratchet`. A commit leaves that range for **two** reasons: it was dropped, or **it merged into main**. My `in-review duration lanes ARE wired` fix did the second — verified by content, not by message:

```
git show origin/main:packages/core/src/task-store/workflow-definitions.ts | grep -c "InReviewDurationLanes"
→ 2
```

It is on main. The engine suite there is now **11059 passed, 0 failed** — the `unwired-lane-parameter-guard` red is gone.

That is the same defect class I have been filing all day, committed in my own tooling: **a check that cannot distinguish "absent because removed" from "absent because satisfied."** My earlier CI monitors had it too (`[ -z "$cur" ]` reading a rate-limited empty response as "no open PRs"). I have stopped the watch rather than let it keep crying wolf — a monitor whose alarms are usually false trains everyone to ignore the real one, and two of the three alarms here *were* real.

**The two genuine drops stand** — those were verified with `git merge-base --is-ancestor`, which does not have this ambiguity.

### Current state, measured

| | |
|---|---|
| `origin/main` engine | **11059 passed, 0 failed** |
| `origin/main` core | 1 failed — `archived-column-gate-parity`, the last red on main |
| this branch + main, merged | 0 conflicts · parity **2/2** |

So this PR still clears main's only remaining red. Note the branch now carries **two** archived-inventory commits — mine (`36625203b9`) and `739448ced9` from another worker after further drift. They coexist and the merged result is green; no action needed, but worth knowing the file has had two independent fixers today, which is itself a signal that the inventory churns faster than one PR cycle.

gsxdsm:
**All five applied.** 2419ea8263. Four were on the FNXC gate I wrote, and one of them made the gate blind to the exact spelling the rule prescribes.

**Hyphens.** The matcher accepted `FNXC:[A-Za-z0-9_]+`, but `AGENTS.md` specifies **`FNXC:Area-of-product`** — so every hyphenated area, i.e. the *documented* form, was skipped. That is the worst possible subset to miss: the gate enforced a rule while ignoring the shape the rule actually mandates.

**`.js` files.** `walk` yielded `ts/tsx/mjs/md` only, so a future-dated stamp in a plain-JS script — the files this repo writes its tooling in — landed unseen.

**UTC.** `toISOString()` produced a UTC "today", so west of Greenwich the date rolls forward for part of each day: a stamp written correctly at 5pm in California read as *tomorrow* and failed. Authors write the local date; the comparison now uses the local one.

**Baseline schema.** Only a parse error was caught, so `null`, an array, or a negative count reached the comparison and either crashed or compared as `undefined` and **silently allowed everything**. A ratchet whose baseline can be neutered by a bad edit is not a ratchet. Verified rather than assumed — `null`, `[]` and `{"a.ts": -1}` all exit 1; the real baseline still exits 0.

**Population 206 → 523 across 240 files.** The hyphen gap alone was hiding more than half the surface. Detection catching up, not new debt — the same shape as this batch's SQL ratchet going 14 → 31.

**The doc count.** 14 + 11 + 1 = 26, not 31. I corrected it by **naming the five unclassified sites rather than adjusting the numbers to agree**: the categories came from hand-reading, the total came from the tool, and the gap *is* the part nobody had audited. Making the arithmetic tidy would have implied a complete audit that never happened — which is the precise failure that section of the doc exists to document. The discrepancy is now explained and left visible.

`pnpm lint` clean; gate reports 523 known stamps, none added.

gsxdsm:
Rebased onto `origin/main` (124c6b4940). **Clean — no conflicts this time.** 11 commits → 10: one had landed independently on main and dropped out as an empty pick.

Verified on the rebased tip rather than assumed, since this batch's whole purpose is three ratchets and a rebase is exactly when a baseline goes stale:

| check | result |
|---|---|
| `check-fnxc-future-dates` | exit 0 — 523 known, none added |
| `check-sql-column-literals` | exit 0 |
| `lifecycle-column-census --strict` | exit 0 |
| core `tsc` / `pnpm lint` | clean |
| gate | 161/487/13/71 |

No baseline needed re-recording, which is the outcome to want: the three populations this batch freezes are unchanged by everything that landed on main in the meantime.

gsxdsm:
Resolved a conflict with `main` on the learnings doc, created when **#2942 (batch-docs)** landed my correction paragraph while this branch carried a later, better version of it.

**Kept the branch side, which is the superset.** Someone extended my correction here to note that my own categories do not add up — 14 lane-bound + 11 archived + 1 dead = 26, against a tool-reported population of 31 — and left the gap visible rather than tidying it, because the categories came from reading and the total came from the scanner. That is a sharper point than the version that reached main, and it corrects a second error in the same paragraph I had already corrected once.

Verified after resolving: `sql` 0 · `seam` 0 · `census --strict` 0 with 0 baselines dirty · `check-fnxc-future-dates` 0 · `archived-column-gate-parity` 2/2 · lint 0.

Note that the archived-gate inventory fix in this batch is what un-reds that test — it currently fails on clean `main` and is not in the merge gate, so nothing is blocked by it, but it lands green with this batch.

gsxdsm:
**Gate was red in CI while green locally — and the cause is a design error in the gate I added, not a flake.** 1755c5d23c.

**The population this ratchet measures is not stable.** "Is this stamp in the future" is answered against **today**, so every date boundary the runner crosses converts some future stamps into past ones and **the count falls on its own, with no code change involved**. I copied the drop-fails rule from the SQL ratchet without noticing that difference. With drop-fails, that guarantees a red gate on some later day.

It fired within hours: the baseline was recorded at **2026-07-30 local** while CI runs in **UTC, already 2026-07-31**.

**Reproduced exactly, on an unmodified tree:**

```
TZ=UTC node scripts/check-fnxc-future-dates.mjs   # exit 1 before, exit 0 after
```

Both sibling ratchets already auto-tighten, for the ordinary reason that a drop is rarely the failing author's to fix. **Here the argument is stronger: nobody *caused* the drop, so there is no author who could fix it.** The ceiling now follows the count down, names what it lowered, and exits 0.

**The rise check is untouched and still fails hard** — verified by probe rather than assumed: injecting one future-dated stamp exits 1, removing it exits 0. That is the gate's actual purpose ("no *new* future-dated stamp") and it is intact.

Worth recording as a general point, since this repo now has four baseline ratchets: **a ratchet is only safe to fail on drops if its measurement depends solely on the code.** This one depends on the clock, so its baseline decays without anyone touching the repo — and I should have noticed that when I chose to copy the stricter rule rather than the one main had already corrected on both siblings.

gsxdsm:
### Verified against main *after* #2943 and #2944 landed — this PR still fully clears the last red

`origin/main` is now `8503a2b12f` (batch-census-sentinels merged). It is red on core, and the drift is now **two** entries, not one:

```
archived-column-gate-parity — AssertionError: TypeScript encoding changed.
-  "packages/core/src/async-mission-store-queries.ts": 2,
+  "packages/core/src/async-mission-store-queries.ts": 1,      ← new, from #2943
-  "packages/core/src/task-store/async-comments-attachments.ts": 8,
+  "packages/core/src/task-store/async-comments-attachments.ts": 5,   ← mine, from #2886
```

**This branch merged with that main: 0 conflicts, parity 2/2.** Both entries are covered — my commit for `async-comments-attachments`, and `739448ced9` (another worker's follow-up on this branch) for `async-mission-store-queries`.

So `batch-sql-ratchet` is the only thing between main and green on core, and it is verified against the *actual* post-merge tree rather than the one I checked earlier.

**Worth noting, because it is the thing that keeps biting this file:** I verified batch compatibility earlier and it passed — then #2943's final content changed and introduced a *second* drift, in a **different file** from the one I had checked. A cross-batch verification is only valid for the tree it ran against. The general hazard remains: this inventory records **counts** of comparisons in files that other batches edit, so it goes stale whenever a conversion lands anywhere, and neither PR's CI can see it (`git diff --name-only` shows no overlap).

That is five separate main reds from this one mechanism today (#2814, #2817, #2856, and twice here). The durable fix is not another re-record — it is either deriving the inventory at test time instead of pinning it, or accepting that a count-based ratchet needs re-recording in the same commit as any conversion, which the guard's own message already asks for and which keeps not happening across batch boundaries.

gsxdsm:
**Correct — and it is the one value that disables the ratchet completely.** beacb6f115.

`Number.isInteger(9007199254740992)` returns true, but that is past 2^53−1 where JavaScript stops distinguishing adjacent integers. Such an entry compares greater than any count this scanner can produce, so the file's ratchet is silently off while the baseline still *looks* well-formed. A validator whose stated purpose is *"this baseline cannot be neutered by a bad edit"* has to reject the value that neuters it most completely.

`Number.isSafeInteger` does. Verified rather than assumed, all four shapes:

| baseline | exit |
|---|---|
| `{"a.ts": 9007199254740992}` | 1 |
| `{"a.ts": 1e309}` | 1 |
| `null` | 1 |
| `{"a.ts": -1}` | 1 |
| real baseline | **0** |

**This is the second finding in two rounds where a fix I wrote to close a hole left the sharpest case open** — the first was copying drop-fails onto a clock-dependent population. Both have the same shape: the guard *named* the right category (an integer; a drop) without checking the extreme member of it (the unsafe integer; the drop nobody caused). Worth stating because I wrote both guards in the same sitting and reviewed neither against its own boundary.

`pnpm lint` clean.

PR #2942, opened 2026-07-31
batch(docs): fold the three workflow-learnings / annotation PRs into one

gsxdsm:
## Family batch — replaces #2926, #2892, #2887

Per the consolidation directive: the u9/e2e **docs family**, folded into one branch and one CI run. Three PRs, five commits, **five files, comment and markdown only**.

| folded PR | commits |
|---|---|
| #2892 `docs/union-vs-per-task` | the project union and the per-task answer are not ranked; date correction |
| #2926 `docs/date-my-measured-claims` | date the measured claims (one was wrong); date the grep-vs-AST measurement in the SQL gate header |
| #2887 `docs/archived-state-literals` | mark the three archived STATE literals as deliberate |

Cherry-picked in original order with authorship preserved; all five applied clean, no conflicts.

## Scope is provably comment-only

```
docs/solutions/workflow-learnings/lifecycle-conversions-that-score-as-wins.md
docs/solutions/workflow-learnings/project-union-versus-per-task-lanes.md
packages/core/src/task-store/async-maintenance.ts        ← FNXC DELIBERATE-LITERAL annotation
packages/core/src/task-store/workflow-definitions.ts     ← FNXC DELIBERATE-LITERAL annotation
scripts/check-sql-column-literals.mjs                    ← header prose only
```

Every added line in `packages/` and `scripts/` is inside a comment — checked by filtering the diff for declarations, conditionals and returns, which returns nothing. The two core files gain `DELIBERATE-LITERAL` markers explaining that `'archived'` is a **state** marker there, not a lane: the sweep collects rows Fusion itself archived or soft-deleted, so widening to the resolved archived set would pull live cards into a cleanup pass.

## Verification (scoped, per the directive — not the full suite)

- `pnpm lint` — clean
- `check-sql-column-literals` — exit 0 (the file it annotates)
- `check:lifecycle-columns` — exit 0 (the markers it adds are census-visible)
- `sync-workflow-ir-callsite-allowlist.test.ts` — 3/3

## A correction worth recording

Mid-fold I saw a changeset, `self-healing.ts` and a test file in `git diff origin/main..HEAD` and nearly reported the batch as impure. They were **main's own commits** — `origin/main` advanced between branch creation and the diff, so the comparison was against a stale base. Rebasing onto current `main` reduced it to the five files above. Worth flagging for anyone else folding a family today: with `main` moving this fast, diff the branch **after** rebasing or the file list will lie to you.

## Closing the originals

#2926, #2892 and #2887 are superseded by this and are being closed. I hold no PRs of my own in this family — all mine merged — so this fold is on behalf of the family rather than a rollup of my own work.

### DIFF
--- a/docs/solutions/workflow-learnings/lifecycle-conversions-that-score-as-wins.md
+++ b/docs/solutions/workflow-learnings/lifecycle-conversions-that-score-as-wins.md
@@ -152,7 +152,8 @@ hands in.** If your fixture passes the lane name, you have tested the consumer. 
 computes that argument in production, and whether they compute it or spell it.
 
 MEASURED, so nobody builds the wrong instrument: an AST probe for a call argument
-`{ column: "<legacy id>" }` finds **79 sites** across `packages/`. It flags the two real `moves.ts`
+`{ column: "<legacy id>" }` found **79 sites** across `packages/` when measured on 2026-07-30 (42 as
+of 2026-07-31, after the analytics conversions landed — the ratio is the point, not the total). It flags the two real `moves.ts`
 offenders, but most of the rest are legitimate — `set({ column: "archived" })` writing the archive
 state, `listTasks({ column: "todo" })` filtering a query. A blocking gate on this shape needs a
 curated list of consumers that interpret the column as a ROLE (as opposed to storing or filtering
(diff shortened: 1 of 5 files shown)


gsxdsm:
**Overlap flag for whoever merges these two batches.**

`batch-sql-ratchet` (#2941) and `batch-docs` (#2942) now share **three identical commits**:

```
docs(gate): date the grep-vs-AST measurement in the SQL gate's header
```

I folded them into `batch-docs`; someone subsequently folded the same three into `batch-sql-ratchet`. Both branches are collection branches, so this is a natural consequence of two people folding overlapping sets — not a mistake by either.

**Two commits are unique to `batch-docs`** and exist nowhere else, so #2942 cannot simply be closed:

```
docs(workflow-learnings): correct the recorded date — 2026-07-31 is tomorrow
```

**Merging order matters.** Whichever lands second will find its three copies already applied — expect empty or trivially-conflicting patches there, and drop them rather than re-applying. Nothing is lost either way; the content is byte-identical.

**Separately, #2933's baseline commit is absent from `batch-sql-ratchet` and that is correct, not a loss.** I verified on the branch: `lifecycle-column-census.mjs --strict` exits 0 with "every file matches its baseline exactly", so the tightened baseline it carried is already in effect — it was absorbed when I regenerated the baseline during the rebase rather than dropped.

I have not rewritten either branch to de-duplicate: both are shared collection branches other workers are pushing to, and a force-push to fix a cosmetic overlap risks clobbering a fold in progress. Flagging is the cheaper correction.

PR #2943, opened 2026-07-31
batch-census-sentinels: six sentinel-marker PRs in one (supersedes #2921 #2928 #2931 #2935 #2938 +1)

gsxdsm:
Fifth family, not in the four you listed — it was about to sit while the others consolidated. **Six folded; two need arbitration.**

## Folded (cherry-picked clean)

migration marker · async archived check · audited-sentinel missing its marker · five of six `archived` checks in one file · the two artifact/comment read-only guards · the last unmarked `getLiveTaskColumn` sentinel.

One root cause, which is why they belong together: **a literal compared against a SENTINEL value is not a lifecycle-lane guard** — the census counts it, and the fix is a marker, not a conversion.

## The baseline conflicted on every cherry-pick

All six re-recorded `lifecycle-column-census-baseline.json` independently. I resolved by **regenerating once from the folded tree** rather than merging six hand-edits: the baseline is a derived artifact, so the measured value is the only correct resolution, and hand-merging derived JSON is how a wrong ceiling gets locked in.

That is the strongest case for the family model I can give you: six PRs touching one derived file conflict pairwise regardless of merge order — 15 possible pairs — and auto-rebase would have churned them serially.

## NOT folded — one line for arbitration

**#2925 (`live-task-column-lanes`) conflicts with #2923 (`fix/task-id-integrity-sentinel`) on `packages/core/src/task-store/task-id-integrity.ts`.** #2923 marks a sentinel there; #2925 converts lanes. Different intents, same file. I did not guess which wins — land one, rebase the other, fold both after.

## Verification

`--strict` exit 0 · backlog **158**, reviewed **122** · core typecheck clean · scoped, not full suite.

## Queue

**52 → 39** after my two folds (this + #2940 portal). The ~24 "self-healing … on a renamed board" family is still the dominant block.

## Summary by CodeRabbit

* **Documentation**
  * Clarified lifecycle-state terminology and migration markers throughout task and project management documentation.
  * Documented the distinction between archived-task sentinels and workflow column identifiers.
  * Updated lifecycle documentation tracking to reflect the latest coverage.

* **Bug Fixes**
  * No runtime behavior changes.


gsxdsm:
**Both findings are mine and both are right.** The branch is checked out in the `phase-d-surfaces` worktree so I can't push; here are the exact corrections to apply.

### 1. `async-comments-attachments.ts` — artifact paths described as "documents"

My sentinel note was written for the document guards and then reused verbatim on the artifact insert/update/list paths. The reasoning is identical, the nouns are not, and a lifecycle reader checking whether the note covers *this* call site would be told about the wrong subsystem.

In the markers attached to the artifact paths, replace:
- "a soft-deleted task's **documents** would become writable" → "…**artifacts**…"
- "the **archived-document read paths** would start returning rows they must not" → "the **archived-artifact read paths**…"

The document-path markers keep their existing wording.

### 2. `comments-ops.ts` — "Every consumer" is wrong

I wrote *"Every consumer of this helper now carries the same marker — `audit-ops.ts`, `async-comments-attachments.ts` (five sites), `task-artifacts-ops.ts` (two) and this one."* That omits `task-id-integrity.ts`, which I marked **myself** in #2923 and then failed to list when writing the summary a few PRs later.

Replace the list with:

> `audit-ops.ts`, `async-comments-attachments.ts` (five sites), `task-artifacts-ops.ts` (two), `task-id-integrity.ts`, and this one.

The claim itself is still true — every consumer *is* marked — so the fix is completing the enumeration rather than weakening it to "several". Worth being precise: the whole point of that sentence is that the family is closed, and a list that silently drops a member undermines exactly that.

**Both are documentation-only**, no behaviour change, and neither affects the census counts this batch re-records.

gsxdsm:
**Both applied.** 9bcf473816. I pushed to the branch ref without taking it — the `phase-d-surfaces` worktree still has it checked out, so I first verified `origin/batch-census-sentinels` matched that worktree's HEAD exactly (`50f70a7ddd`), meaning there was no unpushed work to clobber. **That worktree needs a `git pull` before its next commit.**

**Terminology.** I wrote the sentinel note for the document guards and reused it verbatim on the artifact paths. The reasoning transfers; the nouns don't, and a lifecycle reader checking whether the note covers *that* call site was told about the wrong subsystem. Corrected at the three artifact sites — `insertArtifactRow`, `updateArtifactRow`, `getArtifacts` — and **deliberately not** at the two document sites (`listTaskDocuments`, `deleteTaskDocument`), which were already right. I mapped each marker to its enclosing function rather than pattern-replacing, since the two families are interleaved in one file.

**Enumeration.** The note claimed *"Every consumer of this helper now carries the same marker"* and omitted `task-id-integrity.ts` — which **I marked myself** in #2923 and then failed to list writing the summary a few PRs later.

The claim is **true**, so the fix is completing the list rather than weakening it to "several". The entire point of that sentence is that the family is *closed*; a list that silently drops a member undermines precisely the thing it exists to assert.

**Verified rather than asserted**, since the sentence is a claim about the whole codebase:

| file | `getLiveTaskColumn` comparison sites | markers |
|---|---|---|
| `audit-ops.ts` | 1 | 1 |
| `async-comments-attachments.ts` | 5 | 5 |
| `task-artifacts-ops.ts` | 2 | 2 |
| `task-id-integrity.ts` | 1 | 1 |
| `comments-ops.ts` | 1 | 1 |

Documentation only — no behaviour change, no census movement. Core `tsc` clean, `pnpm lint` clean.

gsxdsm:
**This fold is now stale against main and should be re-derived rather than rebased. Flagging rather than force-resolving.**

#2940 merged, then `batch-u7-lane-fixes` (#2925) and other batches landed, and they touch the same files these six commits do. Rebasing produces **real source conflicts**, not just the census-baseline conflict I resolved three times earlier:

- `packages/core/src/task-store/task-id-integrity.ts` — #2925 **converted** those lanes; my folded #2923 commit **marks a sentinel** there. That is exactly the #2923-vs-#2925 arbitration I flagged, and main has now answered it in #2925's favour.
- `packages/core/src/task-store/async-comments-attachments.ts` — same shape, different file.

I aborted the rebase. Hand-resolving six folded commits against several landed batches, at the end of a long session, is how a wrong marker gets locked into a gate-blocking ratchet — and the baseline would then enforce it.

## What I verified before stopping

**On current main, `task-id-integrity.ts` still has 2 unmarked guards** (`archived` ×2). So #2925's conversion did **not** subsume #2923's work — the sentinel markers are still needed, just on top of the converted file rather than instead of it. That is a small, clean follow-up against current main, not a rebase of stale commits.

## Recommended disposition

**Close this and re-derive.** The six folded commits were correct against the main they were cut from; three or four of them may already be redundant now. Whoever picks it up should re-run the census on current main (**backlog 166, reviewed 111**) and mark what is actually still unmarked, in one commit. That is less work than resolving this rebase and it cannot lock in a stale marker.

I am not closing it myself in case the family owner wants the commit trail — say the word and I will.

## What this cost, for the retro

Two family folds took 52 → 37 (with #2940 now merged). But #2943 sat long enough for main to move underneath it, and the fold's value decayed to zero in that window. **A family fold is perishable**: its whole premise is that N PRs touch overlapping code, which is exactly what makes it go stale fastest. Folds should be merged ahead of ordinary PRs, not queued behind them.

gsxdsm:
**Rebase attempted and backed out — the conflict is more interesting than a textual clash, and it changes what this batch should say.**

**The deferred blocker got done.** Main has since converted `getLiveTaskColumn` to take the board's resolved archived lanes (`getLiveTaskColumn(db, taskId, projectId, archivedColumns)`), threading the parameter through `listTaskDocuments` and its siblings. That is the one genuinely-owed conversion this whole sentinel family kept pointing at — the #2820 item I deferred five times because the helper had no task, no workflow and no lane vocabulary, with callers inside open transactions.

So every hunk here is the same shape: **main adds the resolved-lanes parameter to a line; this batch adds a `DELIBERATE-LITERAL` marker to the same line.** Both are wanted, and the merge is mechanical only if you keep main's code and this branch's comment. I got `task-id-integrity.ts` right that way; on `async-comments-attachments.ts` my filter kept signature fragments from the wrong side and produced a broken file, so I aborted rather than iterate a third time on a file I have already broken once today.

**The marker text needs one substantive edit before this merges**, which is why I am not treating it as a pure conflict resolution. The notes say the sentinel is safe "because the helper compares a literal column"; that reasoning is now stale. The accurate version — which I wrote into `task-id-integrity.ts` during the attempt — is:

> Main has since converted `getLiveTaskColumn` to take the board's archived lanes. That changes which rows the helper **classifies** as archived; it does not change the **sentinel it returns**, which still collapses archived and soft-deleted into one string. So this comparison stays literal.

Same conclusion, different reason, and the old reason would mislead the next reader into thinking the conversion is still outstanding.

**State:** branch untouched at its previous head, nothing broken pushed, worktree clean. `batch-sql-ratchet` (#2941) is rebased and green; this one needs the merge above plus the note correction.

gsxdsm:
**Rebased and pushed.** ce4b9dd44c — and the rebase changed what this batch means, so the commit says so.

**The blocker this family kept pointing at has landed.** Main now has `getLiveTaskColumn` taking the board's resolved archived lanes, threaded through its callers. That was the one genuinely-owed conversion here (#2820), which I deferred five times because the helper had no task, no workflow and no lane vocabulary, with three callers inside open transactions.

**Every conflict was the same shape:** main adds the parameter to a line, this batch adds a marker to the same line. Two resolution strategies both failed in instructive ways —

- `--ours` gives a clean rebase that **silently drops all nine markers**, i.e. the batch's entire point, while leaving a green typecheck. That is the dangerous one: it looks like a successful merge.
- hunk-by-hunk merging produced a broken file twice, because my filters kept signature fragments from the wrong side.

So the rebase takes main's code wholesale and **re-applies the markers on top** — same result, by a route that cannot half-succeed. Nine markers across five files, verified present rather than assumed.

**The reasoning is corrected, not merely restored.** The old notes justified the sentinel by saying the helper compares a literal column. That is now stale. The accurate version:

> The conversion changes which rows `getLiveTaskColumn` **classifies** as archived; it does not change the **sentinel it returns**, which still collapses archived and soft-deleted into one string.

Same conclusion, different reason — and the old wording would tell the next reader a conversion is still outstanding when it has in fact landed.

Backlog **155**, baseline re-recorded in the same change. Core `tsc` clean, `pnpm lint` clean, census `--strict` 0, gate green (161/487/13/71).

PR #2944, opened 2026-07-31
batch: self-healing sweeps were blind on renamed boards (26 sweeps, folds 23 PRs)

gsxdsm:
**Consolidation of 23 open PRs into one.** Every one shared a single root cause and mostly touched a single file; 23 CI runs for that was indefensible.

Folds and supersedes: #2867 #2869 #2876 #2879 #2883 #2891 #2899 #2901 #2902 #2905 #2906 #2914 #2916 #2918 #2919 #2920 #2922 #2927 #2929 #2932 #2934 #2937 #2939.
(#2865, #2882, #2897, #2909, #2912 already merged and are not re-folded.)

## The root cause

A self-healing sweep selects its work with `listTasks({ column: "in-review" })`. On a board whose lanes are renamed that returns **nothing**, so the sweep never runs — no error, no log line, no failed task. Several sweeps had already had their *predicates* converted to resolved lanes, which dropped a census count and changed nothing, because the query above the loop had already returned an empty list.

**26 sweeps converted.** Each one: read the project's columns for the role, then decide each card against **its own** workflow, with the legacy ids unioned so a board mid-rename is never skipped.

## What each sweep stops silently failing to do

| | |
| --- | --- |
| stale merger status | one finished card held the **merge queue** for everything behind it |
| stale `blockedBy` / completed-task release | dependents stayed blocked on work that had already finished — the board stops moving |
| workspace partial lands | a task left with **some repos merged and some not** |
| mid-merge retry stamp | the card stalled *and* the operator's manual Retry was gated by the same stamp |
| in-progress limbo / no-progress failures | dead cards held a work slot forever |
| partial-progress retry | real work parked failed with its **retry budget unspent** |
| orphaned-execution signal | visibility only — the one signal pointing at an orphan went silent |
| zero-commit audit | went **half-blind**: the error arm kept working, the lane arm did not |

Plus: ghost review cards, transient merge failures, misclassified failures, branch misbinding, missing-worktree failures, merged-but-unfinished finalization, done-metadata repair, self-owned branch conflicts, orphan-only scope violations, post-done wedges, idle assigned agents, PR-conflict worktree ownership, and orphaned workspace worktrees.

## Two defects the conversion itself introduced, both caught and fixed

1. **Missed pairs.** Widening a read without converting the guards beneath it is *worse than not converting*: the sweep starts admitting renamed-board cards and then mis-decides every one. Review caught a second guard on a re-read row; the audit that triggered found **five more**, one of which gates the `reviewProof` triple-proof — a renamed review card would have been moved backward with the safety check silently skipped. Column guards 86 → 81.
2. **Duplicate processing.** The literal reads were disjoint by construction; resolved reads are not, so a column carrying two role flags put one card in two buckets — duplicate moves, duplicate audit rows, inflated counts.

Both now have ratchets. `self-healing-converted-sweeps-have-no-literal-lane-guards.test.ts` **derives** its sweep list (a sweep counts as converted when its body calls `resolveProjectColumnsForRoles`), so it cannot go stale, and it carries two positive controls because a broken regex finds no offenders and a broken derivation iterates nothing — an empty loop registers no tests and reads green.

## Deliberately unchanged

- 22 `moveTask` destinations carrying `recoveryRehome: true` — `moves.ts` exempts these so a card stranded in an undeclared column stays rescuable.
- One literal in `clearStaleBlockedBy`'s log-dedup closure (allowed by name in the ratchet, with the reason).
- `surfaceInReviewStalls` — hot list-read path, needs a batched prefetch; that is a performance design decision, not a conversion.
- `scheduler.ts` and `replan-target.ts` — built on `resolveTaskWorkflowIrSync`, which returns the default IR for every task in production. Converting there produces inert code.

## The fold itself is worth one note

All 23 branches appended to the **same test file at the same anchor**, so every automatic strategy — git 3-way, `merge-file --union`, and three hand-written resolvers — interleaved them mid-block. Two attempts committed conflict markers before I caught it. The file is therefore **reconstructed**: head authored once, body assembled as the union of each branch's own intact top-level segments keyed by test title, with the nested `already-merged hard blocker` describe appended whole (flattening it orphaned its helper). Verified by *parsing after every step* rather than trusting the merge — which is how each interleaving was caught.

## Verification

`pnpm test:gate` 161 + 487 + 13 + 71. Scoped suites 592 passed (self-healing, the blindness suite at 68 cases, the ratchet, and the notification suite). `tsc` engine clean; `pnpm lint`, `check:changesets`, `lifecycle-column-census --strict` and `check-sql-column-literals` all clean, each run explicitly.

Each folded conversion was individually revert-proven on its original branch — the read reverted alone, and the per-card verdict reverted alone — and those measurements are recorded in the commit messages carried into this branch.


gsxdsm:
### Pre-flight: clean against main, and clean combined with #2941

Merged locally with `origin/main` (`c6767cb258`) and ran the full engine suite:

```
1 failed | 11140 passed | 14 skipped | 1 todo
  unwired-lane-parameter-guard > reports exactly the known-unwired declarations, and no new ones
```

**That failure is not yours.** It is red on `main` itself, and the diff is exactly the two pre-existing entries:

```
+ packages/core/src/task-store/async-audit.ts completeColumns
+ packages/core/src/task-store/async-audit.ts reviewColumns
```

I checked specifically whether this batch *adds* any new unwired declarations — it adds none, which was the thing worth verifying given how much surface it touches.

**Combined with #2941** (which fixes those two):

| tree | result |
|---|---|
| #2944 + main | 1 failed / 11140 passed — the inherited red only |
| #2944 + #2941 + main | **11141 passed, 0 failed** · 0 conflicts |

So the two batches are compatible in either merge order, and the guard goes green once #2941 lands.

For reference, #2941's fix is not a baseline edit: those two parameters are genuinely *wired* (`getInReviewDurationEventsImpl` resolves both roles and passes them), but the caller typed the value inline and never named `InReviewDurationLanes`, and the guard keys an interface member to its owner symbol. Importing the type clears it with `KNOWN_UNWIRED` unchanged — which matters, because that list is explicitly one-directional and appending to it would have been the wrong fix.

gsxdsm:
Correcting something I put on record about the one query this PR leaves behind, because the reason I gave was wrong and would send the next person down the wrong path.

I deferred `surfaceInReviewStalls` saying it "needs a batched lane prefetch on a hot list-read path". That is not the blocker. The sweep has **no per-card lane verdict at all** — converting its read is the same one-line union as `reconcileStaleMergerStatus`, with nothing to prefetch.

The actual blocker is a **missed pair one module away**:

```ts
// packages/core/src/in-review-stall.ts:179
export function getInReviewStallReason(task, context = {}) {
  if (task.column !== "in-review" || task.paused === true) return undefined;
```

Widening the sweep's query alone would hand every renamed-board card to a shared classifier that drops it on a literal — the exact defect the ratchet in this PR exists to catch, and it would have looked like a clean one-line conversion.

**What it actually needs:** an optional `reviewColumns` on `InReviewStallContext`, with the literal kept as the no-metadata fallback — the same seam `getTaskHardMergeBlocker` already has — then wired from the sweep.

**Why I am not doing it here:** `in-review-stall.ts` is core, and the classifier has **8 live callers in `packages/core/src/task-store/reads.ts`** decorating task lists. Wiring only the engine sweep leaves those 8 on the legacy fallback, so the dashboard's stall badge stays literal on a renamed board. That is a core/dashboard decision about a shared seam, not an engine conversion, and it should not land as a drive-by inside a freeze.

So: one query left in `self-healing.ts`, its blocker is now identified precisely, and the fix is a scoped core change with a named owner rather than the vague "prefetch" I claimed earlier.

PR #2947, opened 2026-07-31
executor: orphaned tasks were never resumed after a restart on a renamed board

gsxdsm:
`resumeOrphaned` is the only path that recovers tasks after a crash or restart. On a board with renamed columns it recovered **nothing**.

## A missed pair, not an unconverted read

```ts
const tasks = await this.listWipLaneTasks();          // resolved by role — already converted
const inProgress = tasks.filter(
  (t) => t.column === "in-progress" && …,             // literal — discards everything the read found
);
```

The read was already resolved. The filter directly beneath it re-asserted the literal on the rows that read returned, so the sweep found the orphans and threw them all away.

**This is the worse half of the class, and it hid well:**

- the read *looks* converted, so scanning for `listTasks({ column: "…" })` finds nothing;
- the census scores only the comparison, so the backlog number moves the **wrong way** as you convert;
- a **structural test already existed** pinning "the read asks for resolved lanes" — `executor-resume-query-lanes.test.ts` — and it was green the entire time the sweep was dead. A test asserting the read exists says nothing about the filter beneath it.

The failure only surfaces after a crash, when an operator is already investigating the crash and has every reason to blame that instead.

## The ratchet, generalised

#2944 ratcheted this class inside `self-healing.ts` after review found one instance and a follow-up audit found five more. This generalises it to every engine source: a function that resolves lanes **and** compares a column id in the same body is a pair.

Excluded, deliberately:
- the **fallback arm** of a resolved ternary (`lanes ? lanes.has(c) : c === "done"`) — the correct shape;
- four files whose literals are deliberate, each with the reason recorded: `ephemeral-worker-manager` (unresolvable-workflow default), `triage` (the U11 orphan case), `scheduler` and `replan-target` (sync listeners on the inert sync IR reader, already pinned by `sync-workflow-ir-is-always-default.pg.test.ts`);
- `self-healing.ts`, because it has a **dedicated** ratchet that is strictly more precise. Two ratchets allowlisting the same site is one fact with two owners, free to drift — the exact failure mode this program keeps hitting. One file, one ratchet.

It carries a positive control: a wrong source path would make every case pass by scanning nothing.

**I swept the rest of the engine with it and executor.ts was the only genuine hit** — everything else is documented-deliberate or blocked on the inert sync reader.

## Revert results

Each measured by restoring the literal filter and re-running:

| | reverted → |
| --- | --- |
| behavioural case | fails — the renamed card is dropped and the sweep returns before touching it |
| the ratchet | fails, naming the site: `resumeOrphaned: executor.ts:5974` |

A non-vacuous companion (card in the review lane → not resumed) rules out a filter that matches everything: a card in review has no session to resume, and re-dispatching it would restart finished work.

**Measured:** `executor.ts` column guards 8 → 7; baseline re-recorded downward.

## Verification

`pnpm test:gate` 161 + 487 + 13 + 71; executor prompt/soft-delete/resume suites plus the new ratchet, 357 passed; `tsc` engine clean; `pnpm lint`, `check:changesets`, census `--strict` and `check-sql-column-literals` clean, each run explicitly.

### DIFF
--- a/.changeset/executor-resume-orphaned-lane.md
+++ b/.changeset/executor-resume-orphaned-lane.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Orphaned tasks are resumed after a restart on boards with renamed columns.
+category: fix
+dev: `resumeOrphaned` read the wip lane by role via `listWipLaneTasks()` but its filter still compared `t.column === "in-progress"`, so on a renamed board the read found the orphans and the filter discarded all of them. The filter now tests membership of the resolved wip columns.
--- a/packages/engine/src/__tests__/executor-resume-lanes-resolved.test.ts
+++ b/packages/engine/src/__tests__/executor-resume-lanes-resolved.test.ts
@@ -217,3 +217,66 @@ describe("resume lanes come from the task's own workflow", () => {
     });
   });
 });
+
+/*
+FNXC:WorkflowResolvedColumns 2026-07-31-19:30 (a MISSED PAIR in resumeOrphaned):
+`listWipLaneTasks()` already resolved the wip lane by role, and the filter beneath it did NOT — it
+re-asserted the literal `in-progress` on the rows that read returned. So on a renamed board the read
+found the orphans and the filter discarded every one.
+
+That is the worse half of this pattern and the reason the sibling structural test was not enough: the
+read looks converted, the census scores only the comparison, and the sweep silently does nothing. The
+consequence here is that orphaned tasks are NEVER resumed after a crash or restart — the single path
+that recovers them — and the failure surfaces only once an operator is already investigating a crash.
+
+`isTaskWorkComplete` is the first thing done per surviving task, before any dispatch, worktree probing
+or git, so it is the observable that needs no live registry.
+
+REVERT CHECK, measured: with the filter back on `t.column === "in-progress"`, this fails — the renamed
+card is dropped and the sweep returns before touching it.
+*/
+describe("resumeOrphaned filters by the board's OWN wip lane, not the literal", () => {
+  function orphanHarness(column: string) {
+    const { store, executor } = harness(RENAMED_IR);
+    const task = {
+      id: "FN-ORPHAN-RESUME",
+      column,
+      title: "orphaned by a restart",
+      description: "",
+      dependencies: [],
+      steps: [{ id: "s1", status: "pending" }],
+      currentStep: 0,
+      log: [],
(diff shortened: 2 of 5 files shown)


gsxdsm:
Expanding a finding I recorded on #2944, because I attempted the fix and the attempt proved the scope is bigger than I said.

I claimed the last unconverted query — `surfaceInReviewStalls` — was blocked on a single literal in `getInReviewStallReason`. I started the seam and backed it out. What's actually there:

**Three sibling signals decorate the same row, and they disagree about which lane it is in.**

| signal | lane handling today |
| --- | --- |
| `getInReviewStalledSignal` | `context.reviewColumn` — **resolved, but singular** |
| `getInReviewStallReason` | no seam — **literal `"in-review"`** |
| `getStalePausedReviewSignal` | `StalePausedReviewContext` has no `reviewColumns` either |

So on one decorated row, `inReviewStalled` can resolve the lane while `inReviewStall` uses the literal. And the singular one is the **arity trap**: `reviewColumn` is `resolveLifecycleColumns().review`, the FIRST column per review role, so a board with a separate merge lane beside its human-review lane has a second review column that matches nothing.

`reviewColumn` is genuinely wired from `reads.ts` (via `resolveReviewColumnForTask`), so this is not an unwired seam — it is a *half-width* one, which is harder to spot precisely because it looks converted and works on the default board.

**Why I backed out rather than landing it.** The fix is a membership `reviewColumns` on three context types plus wiring at 8+ call sites in `reads.ts` — a hot list-read path decorating every task. My first pass mis-wired `getStalePausedReviewSignal`, which has no such field; tsc caught it, and that is the signal that this needs someone with command of that file rather than a drive-by from the engine side during a freeze. Landing the seam on the classifiers **without** wiring the callers would be worse: an optional parameter with no caller is the unwired-parameter anti-pattern this program already removed five of.

**What a correct change looks like**, for whoever owns `packages/core/src/task-store/reads.ts`:

1. `reviewColumns?: ReadonlySet<string>` on all three context types, with the existing literal/singular preserved as the no-metadata fallback so unconverted callers are byte-identical;
2. one `resolveReviewColumnsForTask` built on `resolveReviewColumns(ir)` (the union of the three review roles), replacing the singular resolver, sharing the existing per-pass IR cache;
3. all 8 call sites wired from that one answer, so the three signals on a row cannot disagree;
4. then `surfaceInReviewStalls` in `self-healing.ts` converts as a one-line union read.

No code in this PR — the engine change here stands on its own. Recorded so the next person starts from the real shape instead of my earlier, narrower description.

PR #2949, opened 2026-07-31
feat(core): untraited-project lane opt-in — and main was red on the FNXC gate

gsxdsm:
Two things, and the second is why the first does not ship alone.

## The opt-in

`resolveProjectColumnsForRoles` gains `untraitedProject: "declared-columns"`. When **no** workflow in the project expresses **any** lifecycle trait, every declared column id joins the answer.

This is the three-state rule at **project** scope — the last item on the deferred list, recorded at three self-healing call sites (#2869, #2876). A board that renames its lanes and declares no traits contributes nothing today, so its cards are **absent from every role-keyed query**, and the correct per-card fallback downstream never runs for them. A fallback cannot rescue a card the query never returned.

**Not "no workflow declares this role."** A project that expresses traits and has no review lane has *answered*; widening there would invent lanes it deliberately lacks. Mutation-verified both directions — widening unconditionally fails 1 of 12, making the option a no-op fails 1 of 12.

**Opt-in, not default**, because the safe direction differs per caller — the finding in `project-union-versus-per-task-lanes.md`:

| caller | over-inclusion costs |
|---|---|
| sweep | nothing — the per-card check discards the extra rows |
| aggregator | an inflated number an operator reads (#2864, #2866) |
| action site | a card routed or notified under a vocabulary that is not its own (#2852, #2891) |

Making it the default moves all three at once, in the one direction two of them must not. Verified byte-identical without the option, so this lands with **no caller changes** and each site adopts it on its own reasoning.

## Main was red, and my own gate caught me first

I dated the new comments `2026-07-31` while today is `2026-07-30` — **the exact defect `check-fnxc-future-dates` exists to prevent, committed while writing the feature.** The gate I added yesterday failed my own commit.

Correcting mine surfaced that the merged sentinel batch, #2947, and three engine test files carried future-dated stamps too, so **the gate was failing on `main` for everyone**, not just here.

All corrected to real dates rather than raising the ceiling. The stamps were simply wrong, and a baseline bump would have recorded the error as permitted — which is the failure mode that ratchet exists to prevent.

Core and engine `tsc` clean, `pnpm lint` clean, census `--strict` 0, FNXC gate 0 (469 known, none added), gate green (161/487/13/71).


gsxdsm:
**Overlap notice: #2950 and #2948 are also fixing the red FNXC gate, and #2950 is the comprehensive one.**

I found main red while writing this feature — my own gate rejected my commit for a future-dated stamp — and corrected the stamps that were blocking my local gate run: the merged sentinel batch, #2947, and three engine test files. That was the subset in my path, not a survey.

#2950 did the survey: **616 stamps across 100 files**, including `self-healing.ts` at 38-against-8 and the query-filter-blindness test at 30-against-6. That is the real shape of the problem and it supersedes my slice.

**Suggested merge order: #2950 first, then this.** Our edits move the same lines in the same direction (future date → real date), so a conflict here is trivial to resolve, but it is avoidable entirely by landing the comprehensive fix first. If this merges first, #2950 will need a rebase over my subset.

**What is unique to this PR and not in either date-fix branch:** the `untraitedProject: "declared-columns"` opt-in on `resolveProjectColumnsForRoles`, with its three cases and mutation proofs. That is the last item from the deferred list and does not overlap with them at all.

I am leaving my date corrections in rather than stripping them, for one reason: removing them turns this branch red on the very gate it is meant to pass, until #2950 lands. If the maintainer prefers a clean feature-only diff, say so and I will strip them and let CI stay red here until #2950 merges.

PR #2953, opened 2026-07-31
chore: tighten the FNXC future-date baseline as the clock advances

gsxdsm:
Retires `packages/core/src/task-store/reads.ts`'s allowance of 2. Its FNXC stamps are no longer in the future, so the headroom they held is removed rather than left sitting where a genuine regression could hide inside it.

## This is the drop path working, not a fix

The gate watches a **clock-dependent population**: stamps age out of "future" on their own, with no code change. That is why the drop path auto-lowers the baseline and exits 0 instead of failing.

The alternative — failing on drops, the way a code-measured ratchet should — would have turned this repo red on a day nobody touched it, and the noise would have trained everyone to re-baseline without looking. **A ratchet may only fail on drops when its measurement depends solely on code.** This one does not, so it tightens silently and a commit like this one records the new floor.

## What I verified

- Second consecutive run exits 0 with no further diff — the tightening is idempotent, not an oscillation between two states.
- Diff is a single removed line; the entry is deleted rather than set to `0`, so the file re-enters as a genuine addition if a future-dated stamp lands there again.
- `467` known future-dated stamps remain across 237 files, unchanged.

## What this does not do

It does not reduce the future-dated stamp count — those 467 are still there and still wrong. They came from agents (me included) stamping FNXC comments a day or two ahead. This only reclaims the allowance for one file that has aged out. **The count will keep falling on its own without anyone fixing anything**, so it should not be read as cleanup progress; the gate's value is blocking *new* future stamps, which it still does.

### DIFF
--- a/scripts/lib/fnxc-future-dates-baseline.json
+++ b/scripts/lib/fnxc-future-dates-baseline.json
@@ -69,7 +69,6 @@
   "packages/core/src/task-store/merge-queue-ops-2.ts": 4,
   "packages/core/src/task-store/moves.ts": 2,
   "packages/core/src/task-store/project-store-ops.ts": 1,
-  "packages/core/src/task-store/reads.ts": 2,
   "packages/core/src/task-store/task-artifacts-ops.ts": 6,
   "packages/core/src/task-store/task-creation.ts": 2,
   "packages/core/src/task-store/task-id-integrity.ts": 1,


gsxdsm:
## Build failure here is a queue-wide upstream break, not this diff

This PR removes one line from a baseline JSON file. The `Build` failure is unrelated, and it is **not specific to this PR** — #2347 fails the same way.

**Root cause**

```
ERR_PNPM_NO_MATCHING_VERSION  No matching version found for @aws-sdk/core@^3.977.4
  while fetching it from https://registry.npmjs.org/
→ @fusion/desktop build: tsx scripts/build.ts
→ @runfusion/fusion build: tsup
```

**What I verified**

| fact | value |
|---|---|
| direct dependency on `@aws-sdk/core` anywhere in the repo | **none** — it is transitive |
| version pinned in `pnpm-lock.yaml` | `3.976.0` |
| version some transitive package requests | `^3.977.4` |
| highest version actually published on npm | **`3.977.3`** |

So a transitive dependency declares a range whose floor **was never published**. That is an upstream publishing mistake, not a resolution this repo can satisfy.

**Why only the desktop step trips it**

`pnpm --filter @fusion/desktop deploy --prod --legacy --config.node-linker=hoisted` re-resolves dependencies rather than installing from the lockfile, so the `3.976.0` pin that protects every other build does not apply here. Every PR that reaches desktop packaging will fail identically until the range is satisfiable.

**What I did not do, deliberately**

I did not add a `pnpm.overrides` entry pinning `@aws-sdk/core`. That is a cross-cutting dependency change affecting every package and every consumer's install — operator-visible and not reversible by closing a PR, so it is an escalation rather than something I should improvise inside a gate PR. It also may be unnecessary: this resolves on its own the moment AWS publishes `3.977.4`.

Re-running the job in case the registry has caught up since.

gsxdsm:
**Confirmed self-resolving.** `npm view @aws-sdk/core version` now returns `3.977.4` — the version was published while I was diagnosing, having been absent minutes earlier. The dependent package had been published ahead of its sibling, so the range was briefly unsatisfiable. Re-run is in progress and should pass with no code change; good that no `pnpm.overrides` pin was added, since it would now be permanent workaround for a window that closed on its own.

PR #2955, opened 2026-07-31
fix(deps): every Build is red — AWS unpublished @aws-sdk/core@3.977.4

gsxdsm:
**Every `Build` check in the repo is failing**, including #2347 and #2953. It is not any of our changes.

```
ERR_PNPM_NO_MATCHING_VERSION  No matching version found for @aws-sdk/core@^3.977.4
  at @earendil-works/pi-ai@0.82.1
  at @aws-sdk/client-bedrock-runtime@3.1048.0
  at @aws-sdk/middleware-websocket@3.972.47
The latest release of @aws-sdk/core is "3.977.3".
```

**AWS unpublished the version after releasing it.** The lockfile still carries a valid integrity hash for `3.977.4` — which is how it got in — while `npm view @aws-sdk/core@3.977.4` now returns 404. So this is not a range that was always wrong; it broke underneath a tree that had resolved fine, which is why older PR builds (#2376, #2377) are green and every recent one is red.

## The obvious pin does not work

My first attempt pinned `@aws-sdk/middleware-websocket` one patch back to `3.972.46`, whose declared range is `^3.977.3` — satisfiable on paper, and the least invasive option.

It fails. pnpm still resolves **up** inside that caret to `3.977.4`, so the regenerated lockfile came back with the unpublished version and the identical error. I only caught it by grepping the lockfile rather than trusting that `pnpm install` exiting 0 meant the problem was gone — the local install succeeded against cached metadata while CI's fresh install would not have.

Only constraining `@aws-sdk/core` itself removes the unresolvable target. Verified: `grep -c 3.977.4 pnpm-lock.yaml` goes **3 → 0**.

## Scope

One override, on the package that lost a version — not on the consumers around it. Marked in-place for **removal** once AWS republishes or ships a later patch: this is an upstream publishing accident, not a decision about our dependency surface.

## Verification

- `pnpm install --frozen-lockfile` — succeeds
- `pnpm --filter @fusion/core build` — succeeds
- `pnpm lint` — clean
- FNXC-date gate — green
- no `3.977.4` anywhere in the lockfile

Checked first that nobody else had a PR touching `package.json` / `pnpm-lock` / `pnpm-workspace` — this is the third shared breakage today and I have duplicated another worker twice, so the check came before the fix this time.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -7,6 +7,7 @@ settings:
 overrides:
   '@types/node': ^25.5.2
   protobufjs: ^7.5.8
+  '@aws-sdk/core': 3.977.3
   '@earendil-works/pi-agent-core': 0.82.1
   '@earendil-works/pi-ai': 0.82.1
   '@earendil-works/pi-coding-agent': 0.82.1
@@ -1431,8 +1432,8 @@ packages:
     resolution: {integrity: sha512-u+NT61JZEkRFtpL0CAw1N1dwxnaLgwVXQl/zjJxTGgLyS/jTIdg2SdoEoCTHxgDyCnqa1HEi9QOoE9/pYRNpOQ==}
     engines: {node: '>=20.0.0'}
 
-  '@aws-sdk/core@3.976.0':
-    resolution: {integrity: sha512-0cjRaEdlVoOrsNb9pP5q1Syyc8pXw5xSj2Np2ryReRTr9FppIIRVSdZK4lbnfmc2Hvgux/xBOUU6baB7z8//uA==}
+  '@aws-sdk/core@3.977.3':
+    resolution: {integrity: sha512-6YZTpF5Zzl0KdSrztM0l6ErKdnXrnnodIDYfayHE9SFfZU8Ubxls7H2XnfWT121jgwiG+WP1z5kyfVDsH5V4qA==}
     engines: {node: '>=20.0.0'}
 
   '@aws-sdk/credential-provider-env@3.972.60':
@@ -1503,8 +1504,8 @@ packages:
     resolution: {integrity: sha512-uUbMs1cBZPafD0ohUj6EwNf0fPZ534NvBxHox4hjX+0Rxq5paSYUem7+hi833pYrzrcnBATKIYpR02MDXT5M9g==}
     engines: {node: '>=20.0.0'}
 
-  '@aws-sdk/xml-builder@3.972.36':
-    resolution: {integrity: sha512-RdGmS1GLrtaTOLE1ElSluMldNrpk9Emq6uYs8SS8iHlu5xTAmM9rRkM91o48+rIRryBtyO9t+uLYCoMG6jVMVA==}
+  '@aws-sdk/xml-builder@3.972.37':
+    resolution: {integrity: sha512-zKq4HQum8JwDyEuyfuI4bbiAcU0KxP6qy+9PR/IsR92IyE/DaBAikzAS50tjxip4bqIIANpCcG+Yyj6CVhXupg==}
     engines: {node: '>=20.0.0'}
 
   '@aws/lambda-invoke-store@0.3.0':
@@ -3036,6 +3037,10 @@ packages:
     resolution: {integrity: sha512-BiEE2bnnGoPKdlGe3L+gOYORDHFGPuYVRLP7iUow/Sflm0B4hC4XY3FC1MRuc7ltzpW2xNnXopKi34TTkULlKQ==}
     engines: {node: '>=18.0.0'}
 
+  '@smithy/core@3.31.1':
+    resolution: {integrity: sha512-CyogUINxvi7C7LDsh8Syo6hVJOT9ckz4rG8dRZfTJ8r91HkMY59PnNooaj7WcHyxEkxPfBAmbgztZU+xTo76lg==}
+    engines: {node: '>=18.0.0'}
+
   '@smithy/credential-provider-imds@4.4.12':
     resolution: {integrity: sha512-ZZPDbl/aRp77aycuoMlo3BTayT4CE2a3uoqETYZU5ySnVbhpl5IJiY7dCZedn+ZusyDLqVv44IvKBiXd2/nK0Q==}
     engines: {node: '>=18.0.0'}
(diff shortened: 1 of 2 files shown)


gsxdsm:
I hit this same break on #2953 and reached the same root cause independently — the trace, the `pnpm deploy --prod --legacy` re-resolution path, and the fact that it is nobody's diff. Agreed on all of that, and the catch that pnpm resolves *up* inside the caret so pinning the consumer does nothing is the non-obvious part.

**But the window has closed, and I think this pin is now unnecessary.** Measured just now:

```
npm view @aws-sdk/core version        → 3.977.4
npm view @aws-sdk/core@3.977.4        → 3.977.4    (resolves; not a 404)
npm view @aws-sdk/core versions       → [... 3.977.2, 3.977.3, 3.977.4]
```

`3.977.4` is published and directly resolvable. The range that could not be satisfied 20 minutes ago can be satisfied now.

**One correction on the mechanism**, because it changes whether the pin should persist. The PR body says the lockfile carried a valid integrity hash for `3.977.4`, which is how it got in, and that AWS unpublished it afterwards. On `origin/main`:

```
git show origin/main:pnpm-lock.yaml | grep -c '3\.977\.4'   → 0
```

`3.977.4` was never in the lockfile — main pins `@aws-sdk/core@3.976.0`, which is what your own diff replaces. The unsatisfiable `^3.977.4` came from a **transitive manifest range** (`@aws-sdk/middleware-websocket`), not from a lockfile entry. So the shape is a publishing *ordering* problem — a dependent published before the sibling version its range requires — rather than an unpublish. That matters here: an unpublish can recur at any time and argues for a durable pin, whereas a publishing lag is self-correcting and argues against one.

**Why I'd rather not land it.** `pnpm.overrides` is a standing constraint on the whole workspace, and this one also drags `@smithy/core` and `@aws-sdk/xml-builder` along. Pinning `@aws-sdk/core` to `3.977.3` permanently caps a transitive dependency to fix a window that is already over, and "remove once AWS republishes" markers reliably outlive the condition they describe.

I deliberately did not add this same override on #2953 for that reason and re-ran the failed job instead. **That re-run is the decisive test** — same commit, no code change. If it goes green, the pin is provably unnecessary; if it stays red, you are right that something durable is broken and I will say so here. Worth waiting for it before merging, since this is cheap to re-open and expensive to un-pin quietly.

Not closing or touching your branch — your call.

gsxdsm:
**Hold this — the upstream break appears to have healed, and merging the override now would land dead weight.**

`@aws-sdk/core@3.977.4` resolves again, and the republished tarball's integrity hash is **identical** to the one the lockfile already carried:

```
sha512-CEkcQlMOQJCvul60U7wdAOACjtdgFWDsfJI+6wUOGdhGNV2lGbuJpi/R50QLpFG3Tp+sQxa/RmzC3X7KHbhuTA==
```

Same bytes, so this was a transient unpublish/registry outage rather than a yank-and-replace. Earlier the same command returned a hard `npm error 404`; I have both observations from this session.

One correction to my own PR body while I'm here: I wrote that the lockfile "carries a valid integrity hash for 3.977.4, which is how it got in". That was true of the tree at the time I looked, but `main`'s committed lockfile has **zero** references to it. The fetch happens in the desktop `pnpm deploy --legacy` step, which — per the note already in `pnpm-workspace.yaml` — resolves pi-coding-agent's transitive ranges *without* the workspace lockfile. So the lockfile was never the carrier; the deploy's independent resolution was.

I have re-run #2347's failed Build to test this directly. If it goes green, **close this PR** rather than merge it: a pin against a version that is available again is an override nobody will remember to remove, and the in-place "REMOVE THIS" note only helps if someone reads it.

If the Build fails again, the fix stands as written — and the non-obvious part is still worth keeping: pinning `@aws-sdk/middleware-websocket` back one patch does **not** work, because its `^3.977.3` range still resolves up to `3.977.4`. Only constraining core itself removes the target.

gsxdsm:
**The re-run settles it: the pin is not needed.**

I re-ran the *same failed job* on #2953 — identical commit, no code change, no lockfile edit:

- run [30602067603](https://github.com/Runfusion/Fusion/actions/runs/30602067603) — previously `failure` on `@aws-sdk/core@^3.977.4`, now **`success`**
- `Build` is now **pass** on both #2953 and #2954

Same tree, same resolution path, opposite outcome. The only variable that changed is the registry, which is what a publishing-ordering break looks like and what an unpublish would not fix on its own.

So this is self-resolved upstream, and merging the override would leave `@aws-sdk/core` capped at `3.977.3` — plus the `@smithy/core` and `@aws-sdk/xml-builder` moves that ride with it — to defend against a condition that no longer exists.

**Suggest closing unmerged.** Your diagnosis and the resolves-*up*-inside-the-caret finding were both right and are worth keeping in the record; it is only the remedy that the clock made unnecessary. Entirely your call, and if you would rather keep a durable guard I would not argue hard against it — I would just want the removal condition to be a check something enforces rather than a comment, since that is the part that historically does not happen.

Credit where due: you verified by grepping the regenerated lockfile instead of trusting `pnpm install` exiting 0. That is the same class of check that would have caught my own first mistake today, when a re-sorted baseline made unmoved lines look like deletions.

gsxdsm:
Closing unmerged — the upstream break healed itself and this override is no longer needed.

**Decisive test:** I re-ran #2347's failed Build on its existing commit, with no pin applied. It now passes. `@aws-sdk/core@3.977.4` resolves again, and the republished tarball's integrity hash is byte-identical to the one already in the tree — so it was a transient unpublish, not a yank-and-replace.

Merging this would have left a permanent override against a version that is available again. The in-place `REMOVE THIS` note only helps someone who reads it, and dependency pins are exactly the thing nobody revisits.

**Worth keeping if this recurs** (AWS has now done it once, so it can happen again):
- The failure surfaces in the desktop `pnpm deploy --legacy` step, which per the existing note in `pnpm-workspace.yaml` resolves pi-coding-agent's transitive ranges **without** the workspace lockfile. So `pnpm-lock.yaml` is not the carrier and grepping it for the bad version proves nothing about whether CI will fail.
- Pinning `@aws-sdk/middleware-websocket` one patch back does **not** fix it, even though its declared range `^3.977.3` looks satisfiable — pnpm still resolves up to the unpublished patch inside that caret. Only an override on `@aws-sdk/core` itself removes the target.
- A local `pnpm install` can exit 0 against cached metadata while CI's fresh resolution still fails. Verify with `grep -c '<bad-version>' pnpm-lock.yaml`, not with the exit code.

PR #2956, opened 2026-07-31
fix(core): the in-review stall signal never got the board's review lanes — 0 of 4 call sites

gsxdsm:
## Main is red, and the red is pointing at a real defect

`unwired-lane-parameter-guard` fails on `origin/main` after #2951. This is not a stale allow-list — the parameter genuinely never reaches the function.

#2951 added `reviewColumns?: ReadonlySet<string>` to three signal modules and wired two of them completely. **`getInReviewStallReason` was wired at none of its four call sites.** Measured by brace-matching each call's option literal:

```
getInReviewStallReason      L227=NO   L390=NO   L599=NO   L729=NO
getInReviewStalledSignal    all 4 wired
getStalePausedReviewSignal  both wired
```

## The user-visible consequence

`reads.ts` computes two adjacent signals for the same card. On a board declaring a **separate merge lane beside its human-review lane**, `inReviewStall` read the *first* review column only, while `inReviewStalled` — three lines below — read the *set*.

**The same card is "in review" for one signal and not the other.** Two signals disagreeing is worse than both being legacy, and it is invisible on every builtin board because there the review set has exactly one element.

At three of the four sites the resolve sat *below* the call, which is why the parameter could not be passed. Those are hoisted.

## I have to correct my own earlier report

On #2951 I said *"3 of 4 call sites wired, `reads.ts:227` is the gap."* **That was wrong.** I had measured with a 12-line proximity grep, which bled into the adjacent `getInReviewStalledSignal` call and counted its `reviewColumns:` as the first call's. Brace-matching the literal shows 0 of 4. The defect was four times larger than I reported, and the cause was exactly the anti-pattern I have spent this session filing against other people's guards — a proximity window standing in for structure.

## Naming the context types

The guard keys an interface member to its **owner symbol** and only counts a mention from a file that also names that owner, so passing the property inline reads as unwired even when every site supplies it. `satisfies InReviewStalledContext` / `satisfies StalePausedReviewContext` on the option literals is real type-checking, not a decorative import — lint rejected the decorative version, correctly.

## New test, because the existing guard cannot see this

Measured: **deleting the `reviewColumns:` line from a fixed call site leaves `unwired-lane-parameter-guard` at 9/9 green**, because the file still names the type. So the wiring I just fixed had no coverage at all.

The new ratchet brace-matches each call site's option literal:

| mutation | result |
|---|---|
| remove lanes from one call site | **1 failed** — *"1 of 4 getInReviewStallReason call sites omit reviewColumns"* |

It also asserts it **found** call sites before checking them — a parse that matched nothing would be vacuous, which is the failure mode this guard family keeps producing. (It caught me mid-change too: an earlier scripted edit left the file syntactically invalid and the source-text test still passed 3/3. It is a wiring ratchet, not a substitute for `tsc`.)

## Verification

Core **4861 passed / 0 failed** · guard **9/9** with `KNOWN_UNWIRED` **unchanged** · `pnpm test:gate` **exit 0** · lint clean · core `tsc` **0 errors**.

### DIFF
--- a/packages/core/src/__tests__/in-review-signals-lane-wiring.test.ts
+++ b/packages/core/src/__tests__/in-review-signals-lane-wiring.test.ts
@@ -1,6 +1,6 @@
 // @vitest-environment node
 /*
-FNXC:WorkflowLifecycleColumns 2026-08-01-04:40:
+FNXC:WorkflowLifecycleColumns 2026-07-31-03:55:
 EVERY in-review signal call site in reads.ts must be given the board's resolved review lanes.
 
 THE DEFECT. `reads.ts` computes two adjacent signals for the same card: `inReviewStall` (via
--- a/packages/core/src/__tests__/in-review-signals-lane-wiring.test.ts
+++ b/packages/core/src/__tests__/in-review-signals-lane-wiring.test.ts
@@ -1,0 +1,72 @@
+// @vitest-environment node
+/*
+FNXC:WorkflowLifecycleColumns 2026-08-01-04:40:
+EVERY in-review signal call site in reads.ts must be given the board's resolved review lanes.
+
+THE DEFECT. `reads.ts` computes two adjacent signals for the same card: `inReviewStall` (via
+`getInReviewStallReason`) and `inReviewStalled` (via `getInReviewStalledSignal`). #2951 wired the
+resolved lane SET into three of the four `getInReviewStallReason` call sites and into every
+`getInReviewStalledSignal` one — but at the first site the `resolveReviewColumnsForTask` await sat
+BELOW the call, so that one path silently kept the legacy single-lane fallback.
+
+On a board declaring a separate merge lane beside its human-review lane, `inReviewStall` then read
+the first review column only while `inReviewStalled` two lines down read both: the same card is
+"in review" for one signal and not the other. Two signals disagreeing is worse than both being
+legacy, and it is invisible on every builtin board because there the review set has one element.
+
+WHY A SOURCE RATCHET RATHER THAN A BEHAVIOURAL TEST. The existing
+`unwired-lane-parameter-guard` does NOT cover this: it asks whether any file NAMES the declaring
+interface, so it is satisfied by the import alone. Measured — removing the `reviewColumns:` line
+from the fixed call site leaves that guard at 9/9 green. The call sites live inside a store read
+path behind PostgreSQL fixtures, so pinning the wiring at the source is the honest coverage until
+someone stands that path up end to end.
+
+Counts, not line numbers: line numbers churn on unrelated edits and a ratchet that cries wolf gets
+deleted.
+*/
+import { describe, expect, it } from "vitest";
+import { readFileSync } from "node:fs";
+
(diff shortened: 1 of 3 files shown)


gsxdsm:
**No rebase needed — the CONFLICTING notice is stale.** GitHub reports this `MERGEABLE`, and I verified the merged result rather than trusting that label.

Test merge of this branch into current `origin/main` (`--no-commit --no-ff`, nothing pushed):

```
Auto-merging scripts/lib/fnxc-future-dates-baseline.json
Automatic merge went well
→ gate exit=0, "478 known future-dated stamp(s), none added."
```

The notice was almost certainly generated before #2953 and #2954 merged, both of which touched the same baseline file.

**Why the raw diff looks alarming and is not.** `git diff main..this-branch` on the baseline shows 9 deletions — every `.sql` and `.css` entry. Those are not deletions this branch makes; they are entries #2954 added to `main` after this branch was cut, so the branch merely predates them. The three-way merge keeps them, confirmed above: `sql/css entries = 9` post-merge.

**One real interaction, now a no-op.** This branch removes the `reads.ts` baseline entry, since its commits fix the two future-dated stamps that entry allowed. #2953 removed that same entry first, for the same underlying reason — the stamps aged out of "future" as the clock advanced. `reads.ts` is already `ABSENT` from main's baseline, so that part of this diff merges to nothing. No conflict, no double-removal, nothing to reconcile.

**I did not rebase**, deliberately. This branch is checked out in another worktree (`kb/phase-c-review-merge`), so rewriting it from here would pull the ground out from under whoever is working there. Rebasing branches held by other worktrees is a mistake I already made once in this program, and it is unnecessary here since the merge is clean.

gsxdsm:
This was my defect in #2951, and the mechanism is worth recording because it will bite the next person the same way.

**How all four sites got skipped.** I wired the call sites with a search-and-replace that swapped an existing key:

```
"reviewColumn: reviewColumnForRow,"  ->  "reviewColumns: reviewColumnsForRow,"
```

That edit can only touch a call site **that already passed the singular key**. `getInReviewStalledSignal` and `getStalePausedReviewSignal` both had a `reviewColumn` seam, so they were wired. `getInReviewStallReason` had *no* seam — that was the whole point of adding one — so it had no key to replace and every one of its four sites was silently skipped. The substitution reported '10 sites wired' because it counted the replacements it made, not the call sites that needed one.

**Two things I should have caught and did not:**

1. I asserted the count from the edit rather than from the call sites. The honest check is to enumerate calls to the function and confirm each carries the parameter — which is exactly what your brace-matching table does, and it takes about a minute.
2. The tell was in the diff and I read past it: at three of the four sites the resolve sits *below* the call. A parameter that isn't in scope yet cannot have been passed. That is a structural giveaway, visible without running anything.

**And the irony worth stating plainly:** #2951's own PR body warns against 'an optional parameter with no caller' as the anti-pattern this program already removed five of — and then shipped exactly that, in the same change. The guard existing is what turned a silent half-conversion into a red main instead of a board where two adjacent signals quietly disagree about the same card.

No action needed from me here; #2956 is correct and main's gate is green (161/487/13/71 verified locally on `3461ae7a92`). Recorded so the next person wiring a seam checks the *call sites*, not the edit count.

PR #2259, opened 2026-07-17
fix(engine-lifecycle): require already-landed proof before finalizing empty AI merges (FN-8141)

gsxdsm:
## What & why

Task FN-8141 laundered a failed task into `done`: its branch had no net changes vs `main` **only because the executor reverted its own work five times**, and the AI empty-merge lane took the "empty means already-landed or nothing-to-do → finalize as no-op done" path, stamping `mergeConfirmed: true` with no reviewer or operator sign-off.

This restores the invariant: **a commit-expected task that reaches the empty AI-merge outcome must not finalize `done` without positive evidence the work already landed.**

`packages/engine/src/merger-ai.ts` empty-outcome lane now, for a commit-expected task (`noCommitsExpected !== true`), requires one of:
1. Durable recorded landing on the task (`mergeDetails.mergeConfirmed`/`commitSha`).
2. A prior AI no-op finalization proof pair in the task log (`hasPriorAiNoOpFinalizationProof`, FN-7261 shape).
3. The branch tip is an **ancestor of the integration branch** (fast-forwarded / zero-ahead / already-integrated).
4. The already-on-main classifier (`detectAlreadyLandedOnMain`) finds a distinct landing commit via a **strong** strategy (`trailer`/`ancestry`/`patch-id`).

The classifier's weak `tree-equal`/`no-diff` strategies are **deliberately rejected**: a branch that committed work then reverted it back to base has a tree equal to `main` (main never advanced), which is exactly the FN-8141 shape and would false-positive. Absent proof, the task gets `task.error` set, emits run-audit `task:empty-merge-finalize-blocked-no-landed-proof` (ids/counts/outcomes-only), and is moved back to `todo` with progress preserved — mirroring the existing FN-6461 blocked lane.

`noCommitsExpected === true` tasks are untouched (hardened separately in the sibling Task 1). The non-empty landed path, group/PR routing, and push-after-merge behavior are unchanged.

## Surface enumeration
- **Single-repo empty-outcome finalize (primary lane)** — guarded in `runAiMerge`.
[...]

PR #2958, opened 2026-07-31
fix(merger-ai): gate the no-commits dep-sync skip on the branch diff — the P1 #2501 shipped without

gsxdsm:
## #2501 merged without its P1 fix; this is that fix, alone

#2501 has landed. Its review threads were resolved — I judged and fixed them — but its head was a fork branch I could not push to, so **the fixes were never in it**. Confirmed on `main` at `c1c1b964af`:

```
merger-ai.ts:902:      if (ctx.noCommitsExpected === true) {      ← bare flag, no diff gate
merge-dependency-sync.ts: export const LOCKFILE_CANDIDATES → 0 matches
```

Rebasing dropped this PR's five duplicated base commits, so it is now **one commit**: the review fix and its regression.

## The defect on main

The dep-sync skip trusts `ctx.noCommitsExpected` alone, and **only ever runs on a branch that has commits** — the `rev-list --count` short-circuit ~50 lines above returns `outcome: "empty"` at zero ahead, so control reaches it only when the branch is AHEAD.

Nothing revalidates the flag. Both downstream empty-lane guards carve no-commits tasks out explicitly — `merger-ai.ts:1372` (#2259 already-landed proof) and `:1994` (FN-8141 executor veto) — and both guard the *opposite* direction: commit-expected task, empty branch. The inverse has no check.

So a task marked no-commits whose executor committed a manifest or lockfile change gets its dependency install **and** its frozen-lockfile validation skipped, and the change lands unvalidated.

## The fix

The flag says *look*; the branch diff decides. A `main...branch` diff touching `package.json` or any `LOCKFILE_CANDIDATES` entry falls through to the normal sync and emits an audit row with `skipOverridden: true`. An unreadable diff **also** syncs — matching the hard-fail contract documented directly above that block, rather than treating absence of evidence as evidence of safety.

`LOCKFILE_CANDIDATES` is exported instead of duplicated, so the skip and the installer cannot drift on what counts as a dependency change.

**Mutation-verified:** reverting to trust-the-flag fails exactly the new case and nothing else. The existing *"lands successfully with noCommitsExpected: true and actual changes"* case is untouched and still passes — `feature.txt` is not a dependency file, so an ordinary source change on a no-commits task still skips. The new case differs only in *which* file the branch touches.

## Also carried over from the #2501 review

**coderabbit's env nit** — `process.env.X = undefined` stores the string `"undefined"`, leaving a previously-absent var truthy and leaking into later tests. `restoreEnv` applied at both sites.

**Both entry paths** — deferred with reasons: `runAiMerge`/`landWorkspaceTask` sit behind real worktrees, sessions and a merge agent, and the cheap version is a mirrored-implementation test that cannot fail on a revert (this repo has deleted two of those). The fix above also means propagation is no longer the only thing between a stale flag and an unvalidated lockfile.

## A correction to my own work

My first version of the regression committed the lockfile while the fixture had left the tree on `main`, so the `main...branch` diff could not see it and the case **passed for the wrong reason**. Corrected, with the reason recorded in the test.

## Verification

- `merger-ai-no-commits-deps-skip` — **5/5**, mutation-verified
- `merge-dependency-sync-lockfile-heal` — **10/10**
- engine typecheck — clean
- `pnpm lint` — clean

### DIFF
--- a/packages/engine/src/__tests__/merge-dependency-sync-lockfile-heal.test.ts
+++ b/packages/engine/src/__tests__/merge-dependency-sync-lockfile-heal.test.ts
@@ -64,6 +64,18 @@ function readLog(path: string): string[][] {
   return readFileSync(path, "utf-8").trim().split("\n").filter(Boolean).map((l) => JSON.parse(l));
 }
 
+/*
+FNXC:TestInfrastructure 2026-07-30-20:10 (PR #2501 review — coderabbit):
+DELETE AN ABSENT VAR, DO NOT ASSIGN `undefined`. `process.env.X = undefined` stores the STRING
+"undefined", so a var that was absent before the test is left set to a truthy string afterwards and
+leaks into every later test in the process — the shape that makes an unrelated suite fail with an
+env-dependent path.
+*/
+function restoreEnv(key: string, value: string | undefined): void {
+  if (value === undefined) delete process.env[key];
+  else process.env[key] = value;
+}
+
 describe("buildNonFrozenRetryCommand", () => {
   it("negates pnpm frozen flag explicitly (overrides CI default)", () => {
     expect(buildNonFrozenRetryCommand("pnpm install --frozen-lockfile")).toBe("pnpm install --no-frozen-lockfile");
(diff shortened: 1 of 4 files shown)


gsxdsm:
**Merge-blocking correction before anyone lands this by hand: this PR and #2501 point at the *same commit* (`2972472dbb`).** `git diff pr2501 pr2958` is empty — identical trees, 18 files / +584 / −313 against main either way.

So the rationale in the description above no longer holds. It says #2501 *"cannot be pushed to"* and *"the fixes those threads asked for are not in it — merging #2501 as it stands ships the P1 unfixed."* Neither is true of the current head: the fix **is** present (the `restoreEnv` helper, 7 references in `merge-dependency-sync-lockfile-heal.test.ts`, deleting absent keys instead of assigning `undefined` back as the string `"undefined"`). Whatever the state was when this PR was opened, the fork head has since caught up.

Consequence: these are duplicates, not a fixed-vs-unfixed pair. **Merge one, close the other** — there is no version-selection decision to make, and merging both would be a no-op second merge of the same tree.

I checked because the description asserted a difference and a hand-merge was about to be made on that basis. Verified by SHA and diff, not by reading either body.

#2501's last open thread asked for exactly this env-restore fix; I have replied there pointing at the code, so both PRs are now thread-clear.

Issue #2915, opened 2026-07-31
`.agent-detail-overlay` is styled in 4 places and rendered nowhere — its mobile overrides are inert


gsxdsm:
## Finding

`AgentDetailView` was moved onto `FloatingWindow` (`AgentDetailView.tsx:905`, `className="floating-window--agent-detail"`). The old overlay element went with it — **`.agent-detail-overlay` is no longer rendered by anything**:

```
grep -rn "agent-detail-overlay" packages/dashboard/app --include="*.tsx" --include="*.ts" \
  | grep -v __tests__
→ (no matches)
```

But it is still **styled in four places**:

```
packages/dashboard/app/components/AgentDetailView.css:2      .agent-detail-overlay { … }
packages/dashboard/app/components/AgentDetailView.css:1783     .agent-detail-overlay { … }   ← inside a mobile @media block
packages/dashboard/app/styles.css:719                        [data-theme="light"] .agent-detail-overlay, …
packages/dashboard/app/styles.css:3783                          .agent-detail-overlay, …      ← inside a mobile @media block
```

So the base rule, the light-theme override, and **both mobile overrides** target a class that never appears in the DOM.

## How it surfaced

Two failures in `agent-modals-mobile.test.tsx`:

```
agent modal mobile CSS structure > AgentDetailView > overlay and modal have mobile-targetable classes
AssertionError: expected null to be truthy
  expect(document.querySelector(".agent-detail-overlay")).toBeTruthy();
```

That test is **document-rooted already**, so this is not the portal/query-root defect behind the rest of this backfill cluster (#2885, #2890, #2893, #2895, #2907, #2911, #2913) — I checked the call counts rather than inferring from the identical `expected null to be truthy` message, and this file has zero `container.querySelector` calls.

The test is doing its job: it asserted a structural contract, the product dropped it, and the test went red.

## Why I have not "fixed" it

Re-pointing the assertion at `.floating-window-overlay` would turn the file green and **bury the dead CSS** — the test would then confirm the new overlay exists while four rules quietly style an element that does not. The failing test is currently the only thing pointing at this.

Two legitimate outcomes, and the choice belongs to whoever owns the FloatingWindow migration:

1. **The move is complete and intended** → delete the four dead rules (checking first whether `FloatingWindow`'s own overlay styling covers what the mobile overrides at `AgentDetailView.css:1783` and `styles.css:3783` were doing, or those mobile behaviours are simply gone), then re-point the test.
[...]

PR #2959, opened 2026-07-31
fix(tests): three more portal query-root failures (pr-tab, worktree-terminal, milestone-slice)

gsxdsm:
Three dashboard test files asserted against `render()`'s `container`, but the components under test mount through `createPortal` — so `container` is **empty** and every query returns nothing. Same root cause as the earlier portal batch; these are the three that were still held back.

| File | Before | After |
|---|---|---|
| `TaskDetailModal.pr-tab` | failing | pass |
| `TaskDetailModal.worktree-terminal` | failing | pass |
| `MilestoneSliceInterviewModal` | failing | pass |

**Measured: 39/39 passing**, rebased on current main (`3461ae7a92`). Lint clean, FNXC date gate exit 0.

### Why this stayed hidden

The queries were a **mix** of `container.querySelector(...)` and `screen.*`. `screen` queries `document`, so they kept working — a portal-mounted modal makes only the `container` half go blind. The result is a file that looks half-alive rather than obviously broken, and the failures present as five different-looking symptoms (`null`, `undefined`, `+0`, `[]`, `-1`) that don't read as one bug.

Grouping candidate files by **`container.querySelector` call count** rather than by symptom is what identified these correctly, and — the part that mattered — correctly *excluded* the neighbouring files that were failing for unrelated reasons.

### One thing to know if you repeat this

A blanket `container` → `document` replace is wrong: it also rewrites `renderResult.container.querySelector` into `renderResult.document.querySelector`, which is not a thing. That broke two already-passing tests on my first attempt. This uses two separate passes with a lookbehind so only the bare receiver is rewritten.

### Scope

Test-side only — **no product code changes**, so no changeset. This does not fix the *cause* (tests are still free to query the wrong root); a lint rule for that is worth considering separately, but it would need to distinguish portal-mounting components from ordinary ones, and I did not want to guess at that boundary inside a test-fix PR.

## Summary by CodeRabbit

* **Tests**
  * Improved modal and task detail accessibility test reliability by querying rendered elements from the document.
  * Updated coverage for keyboard navigation, Pull Request status indicators, tab ordering, and onboarding provider cards.

### DIFF
--- a/packages/dashboard/app/components/__tests__/MilestoneSliceInterviewModal.test.tsx
+++ b/packages/dashboard/app/components/__tests__/MilestoneSliceInterviewModal.test.tsx
@@ -272,7 +272,7 @@ describe("MilestoneSliceInterviewModal", () => {
         targetTitle="Milestone 1"
       />,
     );
-    const modal = container.querySelector(".planning-modal");
+    const modal = document.querySelector(".planning-modal");
 
     expect(mockUseMobileKeyboard).toHaveBeenCalledWith({ enabled: true });
     expect(modal?.getAttribute("style")).toContain("--keyboard-overlap: 250px");
--- a/packages/dashboard/app/components/__tests__/TaskDetailModal.pr-tab.test.tsx
+++ b/packages/dashboard/app/components/__tests__/TaskDetailModal.pr-tab.test.tsx
@@ -72,11 +72,11 @@ describe("TaskDetailModal Pull Request tab", () => {
     );
 
     expect(screen.queryByTestId("pr-panel-stub")).toBeNull();
-    expect(container.querySelector(".detail-in-review-stall")).toBeNull();
+    expect(document.querySelector(".detail-in-review-stall")).toBeNull();
 
     fireEvent.click(screen.getByRole("button", { name: "Pull Request" }));
 
     expect(screen.getByTestId("pr-panel-stub")).toBeInTheDocument();
-    expect(container.querySelector(".detail-in-review-stall")).toBeTruthy();
+    expect(document.querySelector(".detail-in-review-stall")).toBeTruthy();
   });
 });
--- a/packages/dashboard/app/components/__tests__/TaskDetailModal.worktree-terminal.test.tsx
+++ b/packages/dashboard/app/components/__tests__/TaskDetailModal.worktree-terminal.test.tsx
@@ -165,7 +165,7 @@ describe("TaskDetailModal worktree terminal tab", () => {
     const { container } = renderDetail();
 
     await screen.findByRole("button", { name: "Terminal" });
-    const tabLabels = Array.from(container.querySelectorAll<HTMLButtonElement>(".detail-tabs .detail-tab"))
+    const tabLabels = Array.from(document.querySelectorAll<HTMLButtonElement>(".detail-tabs .detail-tab"))
       .map((tab) => tab.textContent?.trim());
 
     expect(tabLabels.indexOf("Comments")).toBeGreaterThanOrEqual(0);
--- a/packages/dashboard/app/components/__tests__/onboarding-flow.test.tsx
+++ b/packages/dashboard/app/components/__tests__/onboarding-flow.test.tsx
@@ -522,7 +522,7 @@ describe("onboarding flow integration", () => {
         ],
       });
 
-      const { container } = renderModal();
+      renderModal();
 
       await waitFor(() => {
         expect(screen.getByTestId("onboarding-apikey-input-anthropic-api-key")).toBeInTheDocument();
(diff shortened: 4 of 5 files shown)


gsxdsm:
### Two commits added since the description above

**1. `onboarding-flow` — the last dashboard failure that had no identified cause is the same bug.**

I had this one filed as unexplained. It is the fourth instance of exactly what this PR is about:

```js
const subscriptionCard = screen.getByTestId("onboarding-provider-card-anthropic-subscription"); // ✅ finds it
const anthropicCard    = screen.getByTestId("onboarding-provider-card-anthropic-api-key");      // ✅ finds it
const rendered = container.querySelectorAll('[data-testid^="onboarding-provider-card-anthropic"]');
expect(rendered).toHaveLength(2);                                                                // ❌ got 0
```

Two `screen` lookups succeed and the `container` count two lines later returns 0 — which is why it read as an onboarding-provider bug rather than a query-root bug. Only that one assertion changed; the file's other `container` uses (lines 778/1204/1272) legitimately assert *nothing* rendered and were left alone. **42/42 passing.**

That closes the dashboard ledger I have been tracking: every remaining failure now has a named cause.

**2. `chore(gate)`: tighten the FNXC baseline.**

Unrelated to portals, folded here rather than opening a fifth PR. #2960 deleted three future-dated stamps but merged without re-recording `scripts/lib/fnxc-future-dates-baseline.json`, so `check-fnxc-future-dates.mjs` rewrites that entry on **every** run and leaves a dirty working tree. Ratchet-only, no stamp changes.

Worth knowing about that script: it is **self-writing**, like the census one. Running it to check a branch silently modifies a tracked file, so a stray baseline diff in your tree is usually the checker, not your edit. That is also how #2960 merged without it — the gate exits 0 on a *drop* while writing the tightened file, so nothing fails to tell you it needs committing.

Rebased onto `e6e70a2562` (post-#2960). FNXC gate exit 0 and the tree stays clean after a second run.

gsxdsm:
### One of these assertions was passing for the wrong reason

Reviewing the final diff, `TaskDetailModal.pr-tab` contains a matched pair:

```js
expect(container.querySelector(".detail-in-review-stall")).toBeNull();    // was PASSING
expect(container.querySelector(".detail-in-review-stall")).toBeTruthy();  // was FAILING
```

Only the `toBeTruthy` half showed up as a failure. The `toBeNull` half passed the whole time — **not because the element was correctly absent, but because the container it queried is empty for every possible input.** A negative assertion against a portal-mounted component's `container` cannot fail. It would have kept passing if the stall banner rendered when it should not, which is precisely the regression it exists to catch.

Both now query `document`, and both pass — so the negative one is doing real work for the first time.

Worth generalizing when reading the rest of this batch: the portal bug makes **positive** assertions fail loudly and **negative** assertions silently vacuous. Counting the visible failures understates it — the files in this PR had 4 red assertions, but fixing the query root also restored several `toBeNull`/`not.toBeInTheDocument` checks that were never going to fail. That asymmetry is also why a "all tests pass" signal was not evidence these files were healthy.

I have not audited the already-merged portal batch for the same pattern; those files are green either way, so it is not urgent, but the negative assertions in them were vacuous for the same reason until the query root was fixed.

gsxdsm:
### Correction: my "every remaining failure has a named cause" claim was wrong

I wrote that above after fixing `onboarding-flow`. I then ran the **full** backfill lane on this branch to check it, and it does not hold:

```
Test Files  4 failed | 540 passed (544)
Tests      13 failed | 11123 passed | 5 skipped (11141)
```

| file | failures | cause at the time I claimed it |
|---|---|---|
| `list-view-windowing` | 8 | named — #2829 |
| `NodesView` | 2 | **not named** |
| `agent-modals-mobile` | 2 | **not named** |
| `core-modals-mobile` | 1 | **not named** |

Five failures across three files had no identified cause when I said everything did. All four files are still red on this branch, and **none of them are affected by this PR either way** — the four files it fixes are green.

Since identified:

- **`NodesView` (2)** — the modals render with `aria-label="Add Node dialog"` while the tests correctly ask for `"Add Node"`. Same defect as #2965 (13 dialogs restating their role); those two go green there.
- **`agent-modals-mobile` (2) / `core-modals-mobile` (1)** — CSS-structure assertions (`expected null to be truthy`, and a missing `height: 85vh`). These belong to the #2915 dead-`.agent-detail-overlay` family, not to portals or aria naming.

**What went wrong on my side:** I had run the lane earlier with the output piped through `tail -40`, so I only ever saw the summary and one of the four failing filenames — then described the rest from memory of an older ledger. The truncation was mine, and it turned a partial view into a confident claim. Re-running without the pipe is what produced the table above.

This PR's own scope is unchanged and independently verified; only the surrounding claim was wrong.

PR #2960, opened 2026-07-31
test(engine): delete two temp-cleanup mechanisms guarding a leak the harness already prevents

gsxdsm:
`scheduler-paused-dispatch-refusal.test.ts` carried **two** tracking arrays and **two** `afterEach` hooks, both collecting the same `mkdtempSync` path and removing it twice. One was added per review round on #2779 — I wrote both, and neither round noticed the other.

The obvious fix is to merge them into one. **I checked whether the leak was real first, and it isn't.**

### Measured

`packages/core/src/__test-utils__/vitest-setup.ts` **redirects `os.tmpdir()`** to a per-worker sink and sweeps it by owning pid. So `tmpdir()` inside a test does not resolve to the real temp root at all. Probing the paths this file actually creates:

```
/var/folders/.../T/fusion-test-workers-8Tv8um/redir-5845/fusion-paused-dispatch-ZUhUVL
```

| run | fixtures created | left behind |
|---|---|---|
| cleanup as shipped | 4 | 0 |
| **cleanup disabled** | 4 | **0** |

The sink is reclaimed either way. Both mechanisms were appeasing a review comment about a problem that could not occur.

### Why deleted rather than merged

A cleanup that cannot be observed to clean anything is not a cheap safety net — it is a claim the file cannot back, and it misreports which layer owns temp lifetime. Keeping one "just in case" would leave the next reader believing this file manages its own fixtures. If the redirect is ever removed, cleanup belongs in the shared setup for **every** test, not re-added file by file. An FNXC note records the measurement and says exactly that, so a third round doesn't re-add a third copy.

### A note on my own measurement

My first check was `ls $TMPDIR/fusion-paused-dispatch-*` before and after — it reported zero leaked with cleanup **on**, which I nearly took as "cleanup works." It also reported zero with cleanup **off**. That contradiction is the only reason I looked further; the glob was measuring a directory the fixtures never reach. The before/after count would have "confirmed" a working cleanup just as readily as a redundant one.

**Verified:** 4/4 pass, `tsc` 0 errors, lint clean, FNXC gate exit 0. Test-only, no product change, no changeset.

## Summary by CodeRabbit

* **Tests**
  * Simplified scheduler dispatch refusal test setup by relying on shared test-environment cleanup.
  * Preserved existing test coverage and behavior.

### DIFF
--- a/packages/engine/src/__tests__/scheduler-paused-dispatch-refusal.test.ts
+++ b/packages/engine/src/__tests__/scheduler-paused-dispatch-refusal.test.ts
@@ -1,7 +1,7 @@
-import { mkdtempSync, rmSync } from "node:fs";
+import { mkdtempSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
-import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { beforeEach, describe, expect, it, vi } from "vitest";
 import { Scheduler } from "../scheduler.js";
 import { seedPlannedSpec } from "./_planned-spec-fixture.js";
 import type { Settings, Task, TaskStore, WorkflowIr } from "@fusion/core";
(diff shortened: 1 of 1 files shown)


gsxdsm:
### Correcting an over-generalization in my own PR body

The description above ended with "If the redirect is ever removed, cleanup belongs in the shared setup for **every** test — not re-added file by file." **That is too strong, and I checked it only after opening this.**

The redirect sink is reclaimed on **process exit** (`process.once("exit", …)` at `vitest-setup.ts:361`, plus the globalTeardown sweep of `WORKER_ROOT`) — **not between tests**. So fixtures do accumulate inside the sink for the life of the worker. What the measurement in this PR actually establishes is narrower than I wrote:

- ✅ These dirs never reach the real OS temp root, so the "leaks into `$TMPDIR` forever" premise behind both #2779 review rounds is false.
- ❌ It does **not** establish that per-file cleanup is pointless in general. Bounding **peak** disk during a run is a real benefit the exit-time sweep does not provide.

This file mints exactly **4** fixtures, so the distinction is immaterial *here* and the deletion stands. But 143 test files in this repo do their own `rmSync` of `mkdtemp` dirs, and I have **not** measured how many mint enough to matter — so nobody should take this PR as license to strip them. Pushed a commit scoping the in-file FNXC note accordingly, since that note is what a future reader will actually find.

Flagging it rather than quietly amending: an over-broad claim in a note about a *false* over-broad claim is worth stating out loud, and it is the same shape of error the PR is about — a cheap proxy standing in for the real measurement.

PR #2961, opened 2026-07-31
engine: restore the stall-signal lane wiring #2951 dropped (and the test that proved it)

gsxdsm:
**My defect, shipped in #2951 — and the same family as the one #2956 just fixed.** Found by auditing my own seams after that, not by a failing check.

## What is on `main` right now

`surfaceInReviewStalls` reads the project's review columns (converted in #2951), then calls `getInReviewStallReason` **without** `reviewColumns`. The classifier falls back to the literal `in-review`, returns no signal for a renamed-lane card, and the sweep surfaces nothing.

That is the textbook **missed pair** this program has a ratchet for: a widened read handing every renamed-board card to a literal classifier. The resolve work happens and is then discarded. On a renamed board an operator sees no stall warnings at all.

#2951's conflict resolution dropped two things together:
- the per-card `stallLanes` map and the `reviewColumns` argument
- **the test that proved the wiring**

## Why nothing caught it

**A deleted test cannot fail.** I verified that rebase by comparing the 68 conflict *hunks* — stripping FNXC stamps, confirming 0 of 68 had real content differences — and then ran the gate. The gate passed precisely because the proving test had gone with the code it proved.

I verified the conflicts. I did not verify the outcome. Those are different things, and the difference is invisible when the evidence disappears alongside the feature.

The `unwired-lane-parameter` guard cannot catch this either, by design: it is deliberately conservative — a mention of the parameter *anywhere* satisfies it — so **partial** wiring is outside its reach. `reviewColumns` is mentioned plenty in `reads.ts`, so the guard is green while this call site goes unwired.

## How I found it

The check #2956 used on the sibling defect, applied to every lane seam I have touched: enumerate each function's **call sites** and confirm each one carries the parameter. That enumeration also flags several other call sites without `reviewColumns`/lane arguments (`merger.ts`, `moves.ts`, `auto-merge-finalization.ts`, `merger-ai.ts`, `project-engine.ts`) — I have **not** touched those here; they need per-site judgement about whether the lane answer is even available, and that is a separate change rather than a sweep.

## Revert result

| | reverted → |
| --- | --- |
| `reviewColumns` at the call (i.e. exactly what #2951 shipped) | fails the restored test |

## Verification

`pnpm test:gate` 161 + 487 + 13 + 71; blindness suite 71; `self-healing.test.ts` 412; `tsc` engine clean; lint, census `--strict`, FNXC gate, changesets all clean.

### DIFF
--- a/.changeset/restore-stall-lane-wiring.md
+++ b/.changeset/restore-stall-lane-wiring.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Review stalls are surfaced again on boards with renamed columns.
+category: fix
+dev: #2951 converted `surfaceInReviewStalls` to read the project's review columns but its conflict resolution dropped the per-card `reviewColumns` argument to `getInReviewStallReason` — and the test proving it — so the sweep resolved lanes and then surfaced nothing. Both restored.
--- a/packages/engine/src/__tests__/self-healing-query-filter-blindness.test.ts
+++ b/packages/engine/src/__tests__/self-healing-query-filter-blindness.test.ts
@@ -2205,4 +2205,66 @@ describe("the already-merged hard blocker judges the card's OWN review lanes", (
 
     expect(blocker).toContain(`must be in '${RENAMED_VOCAB.review}'`);
   });
+
+  /*
+  FNXC:WorkflowResolvedColumns 2026-07-30-16:25 (RESTORED alongside the wiring it proves):
+  `surfaceInReviewStalls` tells an operator a card is stalled in review. #2951 converted its READ to the
+  project's review columns but its bulk conflict resolution dropped the per-card `reviewColumns`
+  argument — and dropped THIS TEST in the same pass, which is why the regression landed silently. A
+  deleted test cannot fail, so the gate stayed green over a sweep that resolves lanes and then surfaces
+  nothing on a renamed board.
+
+  REVERT CHECKS, both measured, each alone:
+    - literal read restored               -> fails, the card is never listed
+    - `reviewColumns` dropped at the call -> fails, the classifier judges the renamed lane by the
+      literal and returns no signal
+  */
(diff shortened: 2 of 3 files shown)


gsxdsm:
Pre-flighted this against `main` because it neighbours #2956 (the same `getInReviewStallReason` lane wiring, on the `core/reads.ts` side). **They do not overlap** — this is `engine/self-healing.ts`, that was `core/task-store/reads.ts` — and I confirmed all 4 call sites in `reads.ts` are still wired on `main`, so the two are complementary. The restored wiring here looks right, and the point about a deleted test being unable to fail is exactly the right framing.

**One non-blocking observation: the new loop does not pass an IR cache.**

```js
const { ir, source } = await resolveWorkflowIrForTaskWithProvenance(this.store, entry.id);
```

The resolver takes an optional third parameter (`workflow-ir-resolver.ts:313`):

```ts
irCache?: Map<string, WorkflowIr>,
```

and four sibling sweeps in this same file already thread one through — `self-healing.ts:7423` (`noOpIrCache`), `:7713` (`perTaskIrCache`), `:7900` (`mergeableIrCache`), `:8101` (`failedStepIrCache`). So the convention here is to pass one **explicitly**, including an explicit no-op when caching is not wanted. This loop is `await`-serial over every stall candidate, so on a board with many in-review cards it re-resolves the same handful of workflow definitions once per card, per sweep.

Four other call sites in the file (`:2861`, `:3297`, `:3788`, `:5738`) also omit it, so this is consistent with half the file and not a regression — which is why I would not block on it. But since a `Map` is already in scope-shape here and the loop is new, threading one is close to free.

Worth stating plainly: I have **not** measured the cost, so treat the above as a code-shape observation, not a performance claim. If you keep it as-is, an explicit `noOpIrCache` like `:7423` would at least make the choice deliberate rather than look like the parameter was missed.

PR #2962, opened 2026-07-31
chore: retire the scheduler-test future-date allowance (3 → 0)

gsxdsm:
One line out of the baseline. No behavior change.

#2960 rewrote `packages/engine/src/__tests__/scheduler-paused-dispatch-refusal.test.ts`, removing three `2026-07-31` FNXC stamps and adding one dated correctly:

```
-FNXC:TaskDispatch 2026-07-31-01:35 (PR #2779 review — greptile P2):
-  FNXC:TestHygiene 2026-07-31-02:30 (PR #2779 review — greptile):
-FNXC:TestHygiene 2026-07-31-02:30 (PR #2779 review — greptile):
+FNXC:TestHygiene 2026-07-30-14:45:
```

The file's allowance of 3 is now headroom nothing occupies, which is exactly where a future-dated stamp can land unseen. Retiring it restores the floor.

## Worth distinguishing from the last one

#2953 was the same drop path for a different reason: those stamps aged out on their own as the clock advanced, with nobody touching the code. **This one is a real code change** — the author fixed the dates.

The distinction matters for how the gate is read. A clock-driven drop means the count is falling without anyone doing anything, and should not be reported as cleanup. A code-driven drop like this one is genuine cleanup, and it is the behavior the gate exists to encourage. Both take the same path (lower the ceiling, name what was lowered, exit 0) because the gate cannot tell them apart at runtime — only the diff can.

## Verified

- baseline diff is exactly one deletion
- second consecutive run exits 0 with no further change — idempotent, not oscillating
- `check-fnxc-future-dates` exit 0, now `475 known future-dated stamp(s), none added` (was 478)

### DIFF
--- a/scripts/lib/fnxc-future-dates-baseline.json
+++ b/scripts/lib/fnxc-future-dates-baseline.json
@@ -180,7 +180,6 @@
   "packages/engine/src/__tests__/review-handoff-lane.test.ts": 1,
   "packages/engine/src/__tests__/scheduler-fanout-escalation-lanes.test.ts": 2,
   "packages/engine/src/__tests__/scheduler-load-lane-union.test.ts": 2,
-  "packages/engine/src/__tests__/scheduler-paused-dispatch-refusal.test.ts": 3,
   "packages/engine/src/__tests__/self-blocked-dependency-deadlock.pg.test.ts": 1,
   "packages/engine/src/__tests__/self-healing-db-corruption.test.ts": 1,
   "packages/engine/src/__tests__/stale-task-reporter.test.ts": 1,


gsxdsm:
### Heads-up: #2959 already carries this exact deletion

#2959 includes a `chore(gate)` commit removing the identical line:

```
- "packages/engine/src/__tests__/scheduler-paused-dispatch-refusal.test.ts": 3,
```

Same file, same single line, same reason — #2960 deleted those three stamps but merged without re-recording the baseline, so `check-fnxc-future-dates.mjs` rewrites the entry on every run and leaves a dirty tree.

Because both sides delete the *same* line, git will merge them without conflict; whichever lands second just contributes an empty diff for this file. So **nothing is broken either way** — flagging it only so the duplicate is a known one rather than a surprise.

**Suggested resolution: let this PR own it and I will drop mine.** A dedicated single-purpose PR is the better home for a shared-ratchet change than a commit tucked into a dashboard test-fix PR. I did not drop it preemptively because #2959 is currently green across all 5 checks and re-running CI to move one line has a real cost — say the word and I will rebase it out, or just merge this one first and mine becomes a no-op.

Context worth having for whoever owns the gate: this script is **self-writing**. Running it to *check* a branch silently modifies a tracked file, and it exits 0 on a count *drop* while writing the tightened baseline — which is exactly how #2960 merged without it. Nothing fails to tell you the baseline needs committing. That is the same shape as the census-CLI baseline-write behaviour, and it will keep producing these duplicate cleanup PRs until the write is gated behind an explicit flag.

gsxdsm:
**Closing — already on main, contributes nothing.**

#2959 ("three more portal query-root failures") landed the identical baseline change before this could merge:

```
git log origin/main -- scripts/lib/fnxc-future-dates-baseline.json
  1f0d371228  fix(tests): three more portal query-root failures (#2959)

git diff origin/main origin/gate/retire-scheduler-test-allowance \
    -- scripts/lib/fnxc-future-dates-baseline.json
  (empty — byte-identical)
```

The file is identical, so merging this would be a no-op. The gate is green on main either way, which is the outcome that matters.

## The finding worth keeping

This was not a coincidence, and it will recur. **The gate writes the baseline as a side effect of running.** On a drop it lowers the ceiling and calls `writeFileSync` before exiting 0 — so any worker who runs `check-fnxc-future-dates` locally for any reason picks up an unrelated tightening in their working tree, and commits it along with whatever they were actually doing. #2959 is a test-fix PR that quietly carried a baseline retirement nobody mentioned.

That is mostly benign — the ratchet self-maintains and the floor never drifts upward. Two consequences are worth knowing:

1. **Duplicate PRs like this one are the normal case, not an accident.** Anyone who notices a stranded allowance and opens a dedicated PR will usually lose the race to whoever ran the gate incidentally. Not worth "fixing" by opening these faster; worth knowing before opening one at all.
2. **Baseline lines appear in diffs that have nothing to do with them.** A reviewer seeing `fnxc-future-dates-baseline.json` in a portal-test PR should read it as gate self-maintenance rather than scope creep — the alternative reading wastes a review cycle.

I considered proposing a `--check`-only mode that refuses to write unless asked, which would make the drop explicit rather than incidental. Deciding against it: the auto-write is what keeps the gate from going red on days nobody touched the code, which is the failure it was built to avoid. Making the write opt-in trades a harmless surprise for a recurring red gate.

PR #2964, opened 2026-07-31
engine: finalization parked ALREADY-MERGED work as failed on a renamed board

gsxdsm:
**The worst symptom in this family: the branch landed, and the board says the task failed.**

`project-engine`'s merge-confirmed finalization spread the task's **real** column into `getTaskHardMergeBlocker` with no `reviewColumns`, so the identity check ran against the literal `in-review`. On a renamed board it returned `task is in 'signoff', must be in 'in-review'`, and the caller parked the card:

```
status: "failed"
error:  "Merge confirmed but finalization blocked: task is in 'signoff', must be in 'in-review'"
```

For work that had already merged.

## Its sibling had already solved this

`auto-merge-finalization.ts` passes the **review-eligible sentinel** instead of the card's own column, with the reasoning recorded at that site: `getTaskHardMergeBlocker` asks *"is this card blocked by anything other than where it sits?"*, and its callers are recovery paths for landed work that a graph crash can leave resting in any column. `project-engine` simply never got the same treatment.

## One name instead of two spellings

Rather than write the sentinel a second time, it is exported once as `REVIEW_ELIGIBLE_SENTINEL_COLUMN` next to the helper whose contract gives it meaning, and both recovery paths use it. **Two sites independently spelling a magic value is how one of them came to be missing it** — that is the actual root cause here, not the literal itself.

This also answers the census, which flagged the new literal — correctly. Its guidance (which I wrote, in #2909) is to hoist a deliberate literal into a *declaration*, where a `DELIBERATE-LITERAL` marker actually attaches, instead of leaving it mid-expression where the marker is silently ignored. The shared constant is exactly that, and it lowers `auto-merge-finalization`'s literal count too.

## Revert result

| | reverted → |
| --- | --- |
| sentinel replaced by the card's own renamed column | reproduces the shipped string |

The middle test asserts that string deliberately — it is what landed in `task.error`, so a regression reports what the operator would actually have seen. A third case checks the sentinel does **not** suppress genuine blockers: incomplete steps still block finalization in any lane.

These drive the helper directly; reaching `project-engine`'s finalization end to end needs a live engine, a merge run and a real repo, while the defect is entirely in *what the blocker is asked*.

## Verification

`pnpm test:gate` 161 + 487 + 13 + 71; `project-engine` + `auto-merge-finalization` + the new suite, 207; `tsc` clean on core and engine; lint, census `--strict`, FNXC gate, changesets all clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed merge-confirmed tasks being finalized correctly when boards use renamed workflow columns.
  * Prevented already-merged tasks from being incorrectly marked as failed due to custom review-column names.
  * Preserved enforcement of genuine incomplete-step blockers.

* **Tests**
  * Added coverage for finalization on renamed lanes and legitimate merge blockers.

### DIFF
--- a/.changeset/finalize-parks-merged-work-failed.md
+++ b/.changeset/finalize-parks-merged-work-failed.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Merge-confirmed tasks finalize instead of being parked failed on renamed boards.
+category: fix
+dev: `project-engine`'s merge-confirmed finalization passed the card's real column to `getTaskHardMergeBlocker` with no `reviewColumns`, so on a renamed board the identity check returned a blocker and already-landed work was parked `failed`. Both recovery paths now share an exported `REVIEW_ELIGIBLE_SENTINEL_COLUMN` instead of spelling the sentinel independently.
--- a/packages/core/src/index.gate.ts
+++ b/packages/core/src/index.gate.ts
@@ -1004,6 +1004,7 @@ export { getPrimaryPrInfo, taskHasManualOpenPullRequest } from "./task-helpers.j
 export {
   getTaskMergeBlocker,
   getTaskHardMergeBlocker,
+  REVIEW_ELIGIBLE_SENTINEL_COLUMN,
   getTaskCompletionBlocker,
   isTaskReadyForMerge,
   allowsAutoMergeProcessing,
--- a/packages/core/src/index.gate.ts
+++ b/packages/core/src/index.gate.ts
@@ -1005,6 +1005,8 @@ export {
   getTaskMergeBlocker,
   getTaskHardMergeBlocker,
   REVIEW_ELIGIBLE_SENTINEL_COLUMN,
+  MERGE_CONFIRMED_TRANSIENT_STATUSES,
+  clearMergeConfirmedTransientStatus,
   getTaskCompletionBlocker,
   isTaskReadyForMerge,
   allowsAutoMergeProcessing,
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -1130,6 +1130,7 @@ export { getPrimaryPrInfo, taskHasManualOpenPullRequest } from "./task-helpers.j
 export {
   getTaskMergeBlocker,
   getTaskHardMergeBlocker,
+  REVIEW_ELIGIBLE_SENTINEL_COLUMN,
   getTaskCompletionBlocker,
   getLatestFailedPreMergeReviewStep,
   isTaskReadyForMerge,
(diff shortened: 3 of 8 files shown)


gsxdsm:
Both review threads addressed in `a1b2` (pushed). Both findings were real. A third problem neither reviewer caught is fixed too.

## 1. `queued` not normalized before the sentinel check — **real, and the more serious one**

Confirmed rather than assumed:

```
packages/core/src/task-merge.ts
  SCHEDULER_TRANSIENT_STATUSES = new Set(["queued"])
  BLOCKING_TASK_STATUSES = new Set([...HARD_BLOCKING_TASK_STATUSES, ...SCHEDULER_TRANSIENT_STATUSES])
```

`queued` **blocks**. So a merge-confirmed card the scheduler had queued reached `getTaskHardMergeBlocker` with the status intact and got parked `failed` — already-landed work, the exact bug this PR fixes for renamed columns, surviving one layer down.

**Fixed by extraction rather than by adding `|| status === "queued"`.** The defect is not the missing value, it is that two finalization paths spell the same normalization independently — which is why they had already diverged. A third caller would diverge again. This change had already de-duplicated `REVIEW_ELIGIBLE_SENTINEL_COLUMN` for precisely this reason; `clearMergeConfirmedTransientStatus` is the same remedy applied to the status set, and both paths now call it.

## 2. Missing FNXC prefix on the DELIBERATE-LITERAL comment — **real, with one thing worth checking first**

Prefixed. I verified before editing that this could not break the census: `hasDeliberateMarker` searches for the marker as a **substring across ancestor comments**, not as a required prefix, so an added `FNXC:` line leaves detection intact. Re-confirmed with `census --strict` after the edit — the site stays classified `deliberate` and does not fall back into the backlog.

Worth stating because "add the required prefix" is the kind of quick win that can silently reclassify a reviewed literal as unreviewed work.

## 3. Not reported: `census --strict` was already failing on this branch

```
packages/engine/src/auto-merge-finalization.ts: allows 1, tree has 0
```

Replacing the literal with the sentinel **dropped** a guard without re-recording the baseline, leaving an allowance open for regrowth. I checked whether I had caused it by stashing my edits — it fails on the clean branch too. Baseline re-recorded (one line).

## Verification

- **Merge gate green**: 13 + 161 + 487 + 71 pass
- Target suite **7 pass** (was 3; added the three transient statuses plus a paired negative)
- **Mutation-verified**: removing `"queued"` from the set fails exactly the `queued` case and nothing else — without it, the new cases pass against the broken code
- `census --strict`, `check-sql-column-literals`, `check-inert-flag-seams`, `check-fnxc-future-dates` all exit 0
- `tsc --noEmit` clean for `@fusion/core` and `@fusion/engine`; eslint clean

**One judgement call, reversible:** the new symbol is registered in `index.gate.ts` as well as `index.ts`. The gate barrel is deliberately slim to keep `test:gate` fast, so adding to it is not free — but the gate suite aliases `@fusion/core` to it and the engine call sites import the helper, so omitting it would break gate resolution. Happy to inline the set at both call sites instead if you would rather not grow that barrel.

PR #2966, opened 2026-07-31
gate: ratchet call sites that never receive the lane answer

gsxdsm:
**This is the gap that let three defects reach `main` in one day.**

`unwired-lane-parameter.mjs` catches a parameter that reaches **no** caller. It is deliberately satisfied by a mention *anywhere*, so **partial** wiring is invisible to it:

| | |
| --- | --- |
| #2956 | `getInReviewStallReason` wired at **0 of 4** call sites while both siblings were wired |
| #2963 | both merge entry points unwired — merging was **impossible** on a renamed board |
| #2964 | merge-confirmed finalization unwired — **already-landed work parked `failed`** |

Every one was a fix that added an optional parameter without the call-site sweep that has to follow it. The existing guard was green throughout, correctly by its own contract.

## A census, not a guard — and that distinction is the whole design

Auditing the sites this finds showed **four of seven were legitimately unwired**: `skipColumnIdentityCheck` callers have already proven lane identity by a stronger means, a sentinel-column caller wants the identity check satisfied by construction, and a dead export has no caller to wire at all.

A check that failed on those is ~57% false positives. The sibling guard's own header says why that is worse than a miss — *"it teaches people to disable the check"* — and I agree, so this does not do it.

Instead it ratchets like the lifecycle census: **36 known unwired sites across 20 files**, allowed to shrink and not to grow. A new unwired caller raises the count and fails; wiring one lowers it and re-records. The recurrence — adding a caller that forgets the lane answer — is precisely what gets caught, and the legitimate sites cost one baseline line each instead of a permanently red gate.

## Detection is AST-based, deliberately

It finds exported functions accepting a lane-named argument — directly *or* as an options-bag member — then finds call sites passing none of them.

Not regex: the ad-hoc scan I used during the audit produced false negatives on multi-line calls, which is exactly how a caller gets missed in the first place. Using a heuristic to police a defect caused by a heuristic seemed like a poor trade.

## Verified to fail on the recurrence

A ratchet that cannot fail is worse than none, so this was measured rather than assumed. Injecting one new unwired caller into `self-healing.ts`:

```
[check-lane-wiring] call sites not passing a resolved lane argument INCREASED:

  packages/engine/src/self-healing.ts: 9 unwired now, baseline allows 8
```

exit 1, naming the file and the delta.

## Placement

Runs as a named `check:lane-wiring` step in `pr-checks.yml` beside the lifecycle, SQL, inert-seam and FNXC ratchets — same convention, same failure ergonomics, ~1s.

Note the baseline records today's state, which still includes the #2963/#2964 sites because those fixes have not merged yet. When they land the count drops and the baseline is re-recorded downward — the ratchet working as intended rather than a conflict.

## Verification

`pnpm test:gate` 161 + 487 + 13 + 71; `tsc` engine clean; lint, lifecycle census `--strict`, FNXC gate, and the new check all clean.

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -62,6 +62,8 @@ jobs:
         run: pnpm check:inert-flag-seams
       - name: FNXC stamp dates
         run: pnpm check:fnxc-future-dates
+      - name: Lane-wiring ratchet
+        run: pnpm check:lane-wiring
 
   typecheck:
     name: Typecheck
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "check:sql-column-literals": "node scripts/check-sql-column-literals.mjs",
     "check:inert-flag-seams": "node scripts/check-inert-flag-seams.mjs",
     "check:fnxc-future-dates": "node scripts/check-fnxc-future-dates.mjs",
+    "check:lane-wiring": "node scripts/check-lane-wiring.mjs",
     "census:lifecycle-columns": "node scripts/lifecycle-column-census.mjs",
     "check:quarantine-ledger": "node scripts/check-quarantine-ledger.mjs",
     "check:mock-completeness": "node scripts/check-mock-completeness.mjs",
--- a/scripts/check-lane-wiring.mjs
+++ b/scripts/check-lane-wiring.mjs
@@ -1,0 +1,88 @@
+#!/usr/bin/env node
+/*
+FNXC:WorkflowLifecycleColumns 2026-07-30-19:20:
+Ratchet the population of call sites that do not pass a resolved-lane argument. See
+scripts/lib/lane-wiring-census.mjs for why this is a census with a baseline and not a hard guard.
+
+Usage:
+  node scripts/check-lane-wiring.mjs                     # fail if any file's count ROSE
+  node scripts/check-lane-wiring.mjs --update-baseline   # re-record (downward moves only)
+*/
+import { readFileSync, writeFileSync, readdirSync, statSync } from "node:fs";
+import { join, relative } from "node:path";
+import { findLaneAcceptingFunctions, findUnwiredCallSites } from "./lib/lane-wiring-census.mjs";
+
+const ROOT = process.cwd();
+const BASELINE = join(ROOT, "scripts/lib/lane-wiring-baseline.json");
+const ROOTS = ["packages/core/src", "packages/engine/src", "packages/dashboard/src", "packages/cli/src"];
+
+function sources(dir, out = []) {
+  for (const entry of readdirSync(dir)) {
+    const p = join(dir, entry);
+    if (statSync(p).isDirectory()) {
+      if (entry === "__tests__" || entry === "node_modules" || entry === "dist") continue;
+      sources(p, out);
+    } else if (entry.endsWith(".ts") && !entry.endsWith(".d.ts") && !entry.includes(".test.")) {
+      out.push(p);
+    }
+  }
+  return out;
+}
+
+const files = ROOTS.flatMap((r) => { try { return sources(join(ROOT, r)); } catch { return []; } });
+const accepting = findLaneAcceptingFunctions(files);
+const unwired = findUnwiredCallSites(files, accepting);
+
+const counts = {};
+for (const hit of unwired) {
+  const key = relative(ROOT, hit.file);
+  counts[key] = (counts[key] ?? 0) + 1;
+}
+
+if (process.argv.includes("--update-baseline")) {
+  writeFileSync(BASELINE, `${JSON.stringify({ counts }, null, 2)}\n`);
+  const total = Object.values(counts).reduce((a, b) => a + b, 0);
+  console.log(`[check-lane-wiring] baseline written: ${total} unwired call site(s) in ${Object.keys(counts).length} file(s)`);
+  process.exit(0);
+}
+
+let baseline = { counts: {} };
+try { baseline = JSON.parse(readFileSync(BASELINE, "utf8")); } catch { /* first run */ }
+
+const rose = [];
+const fell = [];
+for (const [file, count] of Object.entries(counts)) {
+  const allowed = baseline.counts[file] ?? 0;
+  if (count > allowed) rose.push(`  ${file}: ${count} unwired now, baseline allows ${allowed}`);
+  else if (count < allowed) fell.push(`  ${file}: ${allowed} -> ${count}`);
+}
+for (const [file, allowed] of Object.entries(baseline.counts)) {
+  if (!(file in counts) && allowed > 0) fell.push(`  ${file}: ${allowed} -> 0`);
+}
+
+if (rose.length > 0) {
+  console.error("[check-lane-wiring] call sites not passing a resolved lane argument INCREASED:\n");
+  console.error(rose.join("\n"));
+  console.error(`
+A function that accepts a lane answer was called without one. That is the shape behind #2956, #2963
+and #2964: a fix adds an optional parameter, and a caller added later never passes it — so the callee
+silently falls back to the legacy literal and the board's own lanes are ignored.
+
+If the call site genuinely should not pass one (identity already proven by a stronger means, a
+sentinel column, or a dead export), record it:
+
+  node scripts/check-lane-wiring.mjs --update-baseline
+`);
+  process.exit(1);
+}
+
+if (fell.length > 0) {
+  console.log("[check-lane-wiring] unwired call sites decreased:\n");
+  console.log(fell.join("\n"));
+  console.log("\nRe-record the baseline in the same commit so the allowance cannot be regrown into:\n");
+  console.log("  node scripts/check-lane-wiring.mjs --update-baseline\n");
+  process.exit(1);
+}
+
+const total = Object.values(counts).reduce((a, b) => a + b, 0);
+console.log(`[check-lane-wiring] ${total} known unwired call site(s), none added.`);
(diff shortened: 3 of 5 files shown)


gsxdsm:
This is the right gap to close, and the framing is correct — I wrote `unwired-lane-parameter.mjs` and its "a mention anywhere counts" rule is exactly why partial wiring is invisible to it. Two notes from having tried to tighten that rule, one a verified gap and one a non-issue you can stop worrying about.

**Gap: named-interface parameter types are not scanned.** The declaration walk handles a bare parameter and `ts.isTypeLiteralNode(param.type)` — an *inline* `{ … }` — but not a `TypeReferenceNode`. There is already at least one lane parameter in that shape, and it is one I added:

```ts
// packages/core/src/task-store/async-audit.ts:382
lanes?: InReviewDurationLanes,          // interface, not an inline literal
// packages/core/src/task-store/workflow-definitions.ts:1047
let lanes: InReviewDurationLanes | undefined;
```

`reviewColumns` / `completeColumns` live on that interface, so they are in the vocabulary but the declaration never registers — the parameter is untracked rather than reported as partially wired. Resolving a `TypeReferenceNode` to its interface declaration and walking its members would cover it.

**Non-issue, measured so you can discard it:** I expected `ts.isCallExpression(node) && ts.isIdentifier(node.expression)` to miss method calls (`this.store.foo(...)`), which would have been fatal in a codebase this store-oriented. It is not — for the lane-carrying functions I sampled, bare calls outnumber method calls **52 to 3**. The restriction is empirically fine and I would leave it rather than add `PropertyAccessExpression` handling for three sites.

**One hazard from my own attempts at the same problem**, since it cost me three iterations: when I tightened the mention rule to require the *declaring symbol* to co-occur, it produced a false positive on precisely the interface-passed-as-object-literal shape — the caller writes `{ reviewColumns, completeColumns }` and never names `InReviewDurationLanes`, so the parameter read as unwired when it was wired. Relaxing type-owned properties to the loose rule then hid **12** genuine entries; scoping owners to consuming functions hid **6**. Whatever this census does with that shape, it is worth a fixture — an under-reporting gate is worse than an over-reporting one, because nobody notices the silence.

gsxdsm:
Strong agreement with the premise — partial wiring being invisible to `unwired-lane-parameter.mjs` is exactly the gap I hit on #2956, and I like that this check **only writes the baseline under `--update-baseline`**. That is a real improvement over `check-fnxc-future-dates.mjs` and the census CLI, both of which rewrite a tracked file as a side effect of *checking*; that behaviour is how #2960 merged with a stale baseline earlier today.

**But I could not get it to fail on the defect it names first, and I think the detector has a blind spot.**

### Reproduction

Checked out this branch and removed `reviewColumns` from one `getInReviewStallReason` call site in `packages/core/src/task-store/reads.ts` — i.e. re-introduced #2956 at 1 of 4 sites:

```
[check-lane-wiring] 34 known unwired call site(s), none added.
```

Unchanged. Exit 0.

### Cause

`findLaneAcceptingFunctions` recognises a lane parameter only when it is an identifier with a lane name, or when `param.type` is a **`TypeLiteralNode`** (an inline `{ reviewColumns?: … }`). `getInReviewStallReason` takes a **named interface**:

```ts
export function getInReviewStallReason(
  task: Pick<Task, …>,
  context: InReviewStallContext = {},      // TypeReference, not TypeLiteral
): InReviewStallSignal | undefined
```

so it never enters `accepting`, and its call sites are never examined. Confirmed by calling the census directly:

```
lane-accepting functions detected: 20
getInReviewStallReason detected? -> false
```

This is not one function. Five core files declare lane-carrying context **interfaces**, and every function taking one is currently invisible:

| file | exported fns | params typed by a named context/options type |
|---|---|---|
| `in-review-stall.ts` | 8 | 1 |
| `in-review-stalled.ts` | 1 | 1 |
| `stale-paused-review.ts` | 1 | 1 |
| `stale-paused-todo.ts` | 1 | 1 |
| `task-priority.ts` | 11 | 2 |

The baseline of 34 is therefore an undercount of unknown size, and the number will *move* once this is fixed — worth knowing before it is read as "34 left to go."

### Suggested fix

Resolve a `TypeReference` to its `interface`/`type alias` declaration and scan **its** members for `LANE_ARGUMENT_NAMES`, in addition to the inline-literal case. Since the census already parses every file, a name→declaration map built in the same pass is enough; no type-checker program required.

**Second, latent gap:** `findLaneAcceptingFunctions` only visits `ts.isFunctionDeclaration` at top level, so `export const fn = (ctx) => …` is also skipped. I checked — **no** exported arrow function currently takes a lane argument, so nothing is missed today. But the convention is not enforced anywhere, so the first one written would silently opt out of the gate.

Non-blocking on the ratchet's design, which I think is right. But as it stands the gate would not have caught #2956, and #2956 is the first case in its own header — so I would want the TypeReference case in before this is trusted as the backstop for that class of defect.

gsxdsm:
**Found and fixed a 71% false-positive rate in this tool before it merged — by using it as intended instead of trusting its output.**

The first version understood only the options-bag spelling of a lane argument. Any function taking it **positionally** had every one of its correctly-wired call sites reported as unwired:

```ts
isRecoverableMissingWorktreeReviewFailureWithProgress(task, reviewColumns)
isNearDuplicateCanonicalInactive(canonical, canonicalFlags)
```

Six of the eight hits in `self-healing.ts` were that false positive. Across the tree, **24 of 34**. The baseline would have recorded two dozen already-correct call sites as debt — and this file's own header says exactly why that is the worst outcome available: *"a false positive here costs more than a miss: it teaches people to disable the check."*

I found it by doing what the census is *for*: opening each flagged site and judging it, rather than trusting the number the tool printed. That is the same mistake the tool exists to catch, one level up — I asserted a count instead of checking the call sites, which is precisely how #2956 happened.

Both spellings are now tracked: option names by name, positional parameters by **index**. **Baseline 34 → 10** genuinely unwired sites, and the remaining ten are the ones I had already judged legitimate (dead export, two sentinel-column callers, `skipColumnIdentityCheck`).

**Re-verified to still fail on the recurrence** after narrowing it — a detector that stops firing is the other way to be useless. Injecting one unwired caller reports `3 unwired now, baseline allows 2`, exit 1.

Also re-recorded the baseline downward for #2963 landing (`merger.ts`, `merger-ai.ts` → 0), which is the ratchet behaving as designed on its own PR.

Gate re-run green: 161 + 487 + 13 + 71, lint / lifecycle census / FNXC all clean.

gsxdsm:
Checked whether this overlaps `check-inert-flag-seams`, since two instruments for one defect class — where one is weaker — has already cost this repo once. **It does not overlap. The two are complementary, and this PR covers a shape mine is structurally blind to.**

## Why mine misses all three of your cited defects

`check-inert-flag-seams` matches a seam by **parameter name and optionality**:

```js
const TRAILING_FLAG_PARAM = /([Cc]olumnFlags|[Ll]ifecycleColumns|[Rr]eviewColumns|[Tt]erminalColumns|[Pp]lannerLanes)$/;
// ...and requires last.questionToken
```

`getInReviewStallReason` is:

```ts
export function getInReviewStallReason(
  task: ...,
  context: InReviewStallContext = {},
)
```

The lane parameter is named **`context`**, carries a **default value rather than `?`**, and hides the lanes inside a typed options object. Every one of those three properties defeats my matcher independently. Confirmed empirically — my gate does not mention that function at all.

So the three defects you list were invisible to it by construction, not by accident, and this ratchet is not redundant with it.

## Worth stating so neither gets deleted later

`check-inert-flag-seams` catches: an optional trailing parameter, named for a lane/flag, that **no** caller supplies, plus **partial** supply per call site (`supplied by 5/6 call sites; omitted at …`). It gained the per-call-site check after `isTaskStuck` shipped with two of three sites omitting the flags.

This PR catches: a lane answer that reaches a call site under **any** parameter shape, including options-object and default-valued forms.

Overlapping in intent, disjoint in reach. If anyone later reads them as duplicates, the deciding question is whether the parameter is `?`-optional and lane-named — mine sees only that narrow shape.

I am not extending mine to cover options objects while this is in flight; that would collide with exactly the surface you are building.

gsxdsm:
Following up on the TypeReference blind spot I flagged before this merged — it went in as-is, so I fixed it in **#2974**.

Confirmed on `main` first, not from the PR branch:

```
detected fns: 20 | getInReviewStallReason: false
```

With the `TypeReference` arm resolved against interface/type-alias declarations, that becomes `30 | true`, and the gate now **fails** when #2956 is re-introduced (`reads.ts: 7 unwired now, baseline allows 6`) where before it reported `none added`.

### The part that needs your judgement, not mine

The honest baseline moves **10 unwired sites across 8 files → 24 across 15**. No entry was removed; 14 became visible for the first time:

```
core/task-store/reads.ts                        0 -> 6
engine/self-healing.ts                          2 -> 4
core/task-store/branch-and-pr-entities.ts       0 -> 1
core/task-store/task-update.ts                  0 -> 1
engine/scheduler.ts                             0 -> 1
dashboard/routes/register-task-workflow-routes  0 -> 1
cli/commands/dashboard-tui/bucket-mapping.ts    0 -> 1
cli/extension.ts                                0 -> 1
```

**I recorded them without auditing them, and I want to be explicit that this is not a claim they are fine.** Your header says roughly half the original hits were legitimately unwired — identity proven by a stronger means, sentinel columns, dead exports — and deciding which of these 14 are in that category is exactly the judgement the census was designed to leave to a human. Wiring a call site that should not be wired is its own defect, so I did not guess.

Two I would look at first: **`reads.ts` at 6** is the file #2956 was about, and **`scheduler.ts`** is a dispatch path.

Also left alone deliberately: `findLaneAcceptingFunctions` still only visits `ts.isFunctionDeclaration` at top level, so `export const fn = (ctx) => …` stays invisible. No exported arrow function takes a lane argument today, so nothing is missed — but the convention isn't enforced, so the first one written opts out silently.

The ratchet design holds up well, and writing the baseline only under `--update-baseline` is the right call — it is the opposite of the self-writing behaviour in `check-fnxc-future-dates.mjs` that let a stale baseline reach main earlier today.

PR #2967, opened 2026-07-31
fix(tests): four CSS/DOM-shape cases pinned contracts the FloatingWindow + U12 migrations retired

gsxdsm:
Two cases in `agent-modals-mobile` asserted DOM shapes that the FN-8619 `FloatingWindow` migration deliberately retired. Both are **stale expectations, not product regressions** — established below rather than assumed, because "delete the failing assertion" is exactly how a real bug gets buried.

### 1. `AgentDetailView` — the suite asserted both sides of the same fact

```js
expect(document.querySelector(".agent-detail-overlay")).toBeTruthy();   // here — RED
```
```js
expect(document.querySelector(".agent-detail-overlay")).toBeNull();     // AgentDetailView.core.test.tsx:97 — GREEN
```

No component renders that class (`grep` across `app/**/*.tsx` outside tests: zero hits). One of these two had to be red, and the one matching the product is the `toBeNull` sibling. This case now asserts the panel class it is actually named for — `.agent-detail-modal`, which **is** live and **is** what the mobile `@media` block in `AgentDetailView.css` targets — and pins the scrim as still-retired.

I did **not** re-point the overlay half at `.floating-window-overlay`. That would only re-assert FloatingWindow's own contract (already covered by `FloatingWindow.test.tsx`) while saying nothing about Agent Detail being mobile-targetable.

### 2. `AgentGenerationModal` — the class belongs to a different component

It demanded `.agent-dialog-overlay`. That class is **still live** — `NewAgentDialog.tsx:415` renders it — which is why the stale expectation looked plausible and survived. But this modal is a `FloatingWindow` with `modal` (`AgentGenerationModal.tsx:162`), so its scrim is `.floating-window-overlay--modal`. Now asserted explicitly, because "a modal blocks the app beneath it" is a real FN-8619 contract worth pinning.

### Why the dead CSS is still here

`.agent-detail-overlay` has 4 CSS definitions and one inert mobile `@media` rule. I deliberately did **not** delete them in this PR:

- Two **passing** tests (`dashboard-overflow-containment.test.tsx:295`, `mobile-horizontal-pan-containment.test.ts:90`) pin a selector *string* that lists `.agent-detail-overlay`. Deleting the CSS turns two green tests red.
- That raises a question I cannot answer without rendering, and I will not boot an instance to find out: **those containment lists do not mention `.floating-window-overlay--modal`.** If mobile horizontal-pan containment is meant to cover modal scrims, the migration may have moved the scrim out from under its guard. That is a product question for whoever owns FN-8619, and it is the substance of #2915.

Worth recording: the retired `.agent-detail-overlay { padding: 0; align-items: stretch }` mobile rule is not a lost feature. `.floating-window-overlay` is `position: fixed; inset: 0` with no flex context, so those declarations have nothing to act on — FloatingWindow positions the panel by geometry instead.

**Verified:** 22/22 in this file, `tsc -p tsconfig.app.json` 0 errors, lint clean, FNXC gate exit 0. Test-only, no changeset.

### DIFF
--- a/packages/dashboard/app/components/__tests__/agent-modals-mobile.test.tsx
+++ b/packages/dashboard/app/components/__tests__/agent-modals-mobile.test.tsx
@@ -1,8 +1,6 @@
 import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
 import { render, screen, waitFor, within } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
-import fs from "node:fs";
-import path from "node:path";
 import { AgentDetailView } from "../AgentDetailView";
 import { AgentGenerationModal } from "../AgentGenerationModal";
 import { AgentImportModal } from "../AgentImportModal";
@@ -203,13 +201,30 @@ describe("agent modal mobile CSS structure", () => {
   });
 
   describe("AgentDetailView", () => {
-    it("overlay and modal have mobile-targetable classes", async () => {
+    /*
+    FNXC:ModalTouchGeometry 2026-07-30-18:20:
+    THE SCRIM IS FLOATINGWINDOW'S NOW — this asserts the panel, not a local overlay.
+
+    This case demanded `.agent-detail-overlay`, which no component renders: FN-8619 migrated Agent
+    Detail onto `FloatingWindow`, whose `modal` host owns the scrim
+    (`.floating-window-overlay--modal`). `AgentDetailView.core.test.tsx:97` already asserts the
+    class is ABSENT, so the suite was asserting both sides of the same fact and one of them had to
+    be red.
+
+    The panel class is what this file is actually for: `.agent-detail-modal` is live and is what the
+    mobile `@media` block in AgentDetailView.css targets. The overlay half is dropped rather than
+    re-pointed at `.floating-window-overlay` — that would only re-assert FloatingWindow's own
+    contract, which FloatingWindow.test.tsx already covers, and would say nothing about Agent
+    Detail being mobile-targetable.
+    */
+    it("modal panel keeps its mobile-targetable class", async () => {
       render(<AgentDetailView agentId="agent-001" onClose={vi.fn()} addToast={vi.fn()} />);
 
       await waitFor(() => {
-        expect(document.querySelector(".agent-detail-overlay")).toBeTruthy();
         expect(document.querySelector(".agent-detail-modal")).toBeTruthy();
       });
+      // The retired local scrim must stay retired; its CSS is dead and tracked in #2915.
+      expect(document.querySelector(".agent-detail-overlay")).toBeNull();
     });
 
     it("tabs have scrollable container class", async () => {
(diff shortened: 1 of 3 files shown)


gsxdsm:
### Second commit: the same migration, third file — `core-modals-mobile`

This was the last dashboard failure outside `list-view-windowing`. Same root cause as the two above, so folded here rather than opened separately.

The case pinned `.modal.task-detail-modal` at `height: 85vh` (desktop), `92vh` (tablet), `100dvh` (mobile), with `resize: both` / `resize: none`. FN-8619 moved Task Detail onto `FloatingWindow`, so the panel is now `width: 100%; height: 100%` and the desktop/tablet sizes are `defaultSize` **in TSX, not CSS** — asserting them against a stylesheet can only ever fail.

**The mobile invariant is the one that mattered, and it survived** — relocated, not lost. `FloatingWindow.css` gives `.floating-window--task-detail` `inset: 0; width: 100vw; height: 100dvh; max-height: 100dvh` and hides the resize handle. That is the same guarantee the old assertions made: a phone gets a full-screen sheet, not a draggable window. It is now asserted there.

I deliberately did **not** re-pin the desktop/tablet numbers against `defaultSize`. Those are layout defaults a designer may retune; a test that goes red on a 640→680 width change is noise. Full-screen-on-mobile is a contract; `85vh` on a desktop is a preference.

### Two vacuity bugs in my own work, both caught by mutation

**1. Wrong media block.** I first reached for the existing `getMainMobileBlock`, which aggregates `(max-width: 768px)`. FloatingWindow's breakpoint is `(max-width: 767.98px), (max-height: 480px)` — the documented project query whose `max-height` clause catches landscape phones wider than 768px. The helper silently returned a block containing none of these selectors. The anti-vacuity assertion I had added caught it during authoring; without that line the test would have "passed" against the wrong stylesheet.

**2. Substring containment.** This one shipped past a green run:

```js
expect(mobileRule).toContain("height: 100dvh");   // satisfied by `max-height: 100dvh`
```

Mutating the rule to `height: 90dvh` left it **green** — the `max-height` line alone satisfies the substring. Now matched at declaration boundaries (`/[{;]\s*height:\s*100dvh/`), which also let me add the `max-width` assertion the original omitted.

| state | result |
|---|---|
| clean tree | 30/30 pass |
| `height: 100dvh` → `90dvh` | **fails** |

Also drops the now-orphaned `extractVhHeight` helper and two dead `node:fs`/`node:path` imports.

**Verified:** 30/30 here, 22/22 in `agent-modals-mobile`, `tsc` 0 errors, lint clean, FNXC gate exit 0.

gsxdsm:
### Third commit: `list-view-windowing` — the last 8, and the same migration again

This closes the dashboard ledger. All 8 cases failed with `expected [] to have a length of 50`, which reads as a broken render window; it was an **unmet precondition**.

```js
fetchBoardWorkflows: vi.fn(() => new Promise(() => {}))   // never settles
```

That was harmless when written (2026-07-26): List view still had `LEGACY_LIST_COLUMNS` to fall back on. **U12/R9 deleted that fallback** two days later, so `ListView` now returns its skeleton unless `useBoardWorkflows` resolves a lane — and a never-settling promise means it never does. Every assertion then ran against an empty document.

Fixed by resolving a real column set (mirroring the payload other board tests use) and awaiting the lane **once** in `renderList`, which is now `async`; the 8 cases became `async` with it. That is the "8 synchronous tests made async" work noted on #2829, plus the mock that was the actual blocker.

**Mutation — the part that matters here.** Making a blank-rendering test render *something* will turn it green whether or not it tests anything. So I broke the feature instead:

| `LIST_SECTION_VISIBLE_INITIAL` | result |
|---|---|
| `50` (shipped) | 8/8 pass |
| `200` (window effectively disabled) | **5 of 8 fail** |

The windowing invariant is genuinely under test, not just the render.

### Pattern across all three commits in this PR

Same shape three times: **FN-8619 / U12 changed a contract, and the tests kept asserting the old one** — a retired scrim, a self-sizing modal, a synchronous lane. None was a product regression; all three had migrated correctly and left stale assertions behind. Worth noting for whoever is tracking that migration, because a red test that looks like a UI bug is expensive to triage — I spent most of the time on these proving they were *not* regressions before changing a line.

**Verified:** 8/8 here, 30/30 `core-modals-mobile`, 22/22 `agent-modals-mobile`, `tsc` 0 errors, lint clean, FNXC gate exit 0. Test-only across all three commits; no product code changed, no changeset.

PR #2968, opened 2026-07-31
fix(glasses): the summary card reported five hardcoded lanes, so a renamed board read as empty

gsxdsm:
## The defect

`boardSummaryCardFromCounts` built its text from five hardcoded lane ids:

```ts
`Triage ${counts.triage} Todo ${counts.todo} Doing ${counts["in-progress"]} Review ${counts["in-review"]} Done ${counts.done}`
```

`boardSummary` seeds exactly those five keys to `0`, then counts by real `task.column`. On a board whose lanes are named anything else, **every interpolated value is the seeded zero** — so the summary card, which is the first card in every deck and the entire body of `GET /board/summary`, reads:

```
Triage 0 Todo 0 Doing 0 Review 0 Done 0
```

…while the real work sits in lanes it never mentions. The wearer is told the board is empty.

It is also wrong on the **default** board today: U11 (#2515) deleted the `triage` lane, so `Triage 0` is permanently dead text — 9 of the 24 characters this display gets per line.

## Why there was no test

The only summary coverage in `cards.test.ts` exercised `boardSummaryCard`, an **export no production file calls** (its sole caller is that test). The function the deck actually ships had none. That gap is why five hardcoded ids survived here.

## The fix, and one deliberate choice

The lanes are derived from `counts` instead of taken as an optional resolved-lanes parameter.

That is on purpose. This program's recurring defect is precisely the "optional lane answer + documented literal fallback" shape shipped without wiring the caller — `unwired-lane-parameter-guard.test.ts` documents **five live on `main` at once**, and in four of five the parameter was unreachable because the caller held a larger defect. `counts` is already keyed by real `task.column` values, so the vocabulary is in hand with **no resolution, no new plumbing, and nothing that can be left unwired**.

Zero-count lanes are dropped so the scarce line budget goes to lanes with work; legacy ids keep their familiar labels and order, unknown ids sort after them alphabetically so output is deterministic. `counts` itself is untouched — the route returns it as the API body and consumers still see every key.

## Revert proof

Restoring only `cards.ts` fails all three new cases, printing the defect verbatim:

```
AssertionError: expected 'Triage 0 Todo 0 Doing 0 Review 0 Done…' to contain 'backlog 2'
AssertionError: expected 'Triage 0 Todo 1 Doing 0 Review 0 Done…' not to match /Triage/
AssertionError: expected 'Triage 0 Todo 0 Doing 0 Review 0 Done…' to be 'No active work'
      Tests  3 failed | 5 passed (8)
```

## Verification (measured)

- `cards.test.ts` — **8/8 passed**
- full plugin suite — **180 passed**, 19 files
- `eslint`, `tsc --noEmit` — clean

No changeset: `@fusion-plugin-examples/even-realities-glasses` is `private: true` and is not bundled into the published CLI.

## Pre-existing failure, NOT from this change

The plugin suite also has **15 failures in `src/__tests__/agent-actions.test.ts`**. They reproduce identically on clean `origin/main` with this branch stashed (`15 failed | 41 passed`), so they are not caused by this PR and are not fixed by it. Flagging separately — this suite is outside the merge gate, which is why it has been red unnoticed.

### DIFF
--- a/plugins/fusion-plugin-even-realities-glasses/src/__tests__/cards.test.ts
+++ b/plugins/fusion-plugin-even-realities-glasses/src/__tests__/cards.test.ts
@@ -26,6 +26,44 @@ describe("cards", () => {
     expect(card.lines).toContain("todo: 2");
   });
 
+  /*
+  FNXC:WorkflowLifecycleColumns 2026-07-30-22:20:
+  THE INVARIANT: the summary card reports the lanes the board HAS, whatever they are called.
+
+  Untested before this — the only summary coverage above exercises `boardSummaryCard`, an export no
+  production file calls, while the card the deck actually ships (`boardSummaryCardFromCounts`, via
+  `boardToDeck`) had none. That gap is why five hardcoded lane ids survived here.
+
+  Reverted, the first case reports `Triage 0 Todo 0 Doing 0 Review 0 Done 0` for a board holding four
+  live cards and fails on every assertion below; the second still carries the dead `Triage 0`.
+  */
+  it("summarises a renamed board by its real lanes, not the legacy five", () => {
+    const renamed = ["backlog", "backlog", "building", "shipped"].map((column, i) => ({
+      ...task,
+      id: `FN-${i + 10}`,
+      column,
+    }));
+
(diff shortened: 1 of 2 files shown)


gsxdsm:

PR #2969, opened 2026-07-31
test(glasses): re-green 15 agent-action tests, and revive a guard that asserted nothing

gsxdsm:
## 15 tests have been red on `main`

`startWork`, `requestReview` and `approvePlan` gained a third `moveTask` argument, `{ moveSource: "user" }`. The assertions in `agent-actions.test.ts` kept the two-argument form:

```
AssertionError: expected "vi.fn()" to be called with arguments: [ 'FN-1', 'building' ]
Received: [ "FN-1", "building", + { "moveSource": "user" } ]
      Tests  15 failed | 41 passed (56)
```

Reproduces on clean `origin/main`. This suite is outside the merge gate, which is why it went red unnoticed.

## The part that is worse than staleness

Three of these are **negative** assertions, and they did not fail — they went **dead**. `expect(fn).not.toHaveBeenCalledWith(id, column)` cannot match a three-argument call, so it passes whether or not the forbidden move happened. `requestReview`'s "never lands on the legacy `in-review`" guard has been asserting nothing since the option landed.

Verified rather than reasoned about — a scratch case, run and then deleted:

```ts
const fn = vi.fn();
fn("FN-1", "in-review", { moveSource: "user" });          // the forbidden move HAPPENED
expect(fn).not.toHaveBeenCalledWith("FN-1", "in-review");   // old form: passes anyway ✓
```

Both cases passed, confirming the old form is vacuous and the three-argument form throws.

## Applied per action, not uniformly

Only **3 of 5** product `moveTask` calls take the option, and the split is deliberate:

| action | source | why |
| --- | --- | --- |
| `startWork`, `requestReview`, `approvePlan` | `{ moveSource: "user" }` | the wearer's tap is a human gesture, matching the dashboard move route |
| `returnToAgent`, `retryTask` | default (engine) | per the Move-Task contract a user-source move parks the row `userPaused`, defeating the return/retry intent |

So `returnToAgent`/`retryTask` assertions are **correct as two-argument** and are left untouched — including their negatives, which genuinely assert because the real call is also two-argument.

Attribution was done by scoping each assertion to its enclosing `it(` block, not to the nearest preceding call: the latter mis-attributes the `await expect(startWork(...)).resolves` form, which reads as `expect`. Blocks mixing a user-source and a default-source action were excluded from rewriting (there were none).

I did **not** "fix" the 3/5 split. It is documented in the product with its reasoning and it matches the Move-Task contract; changing it would be a behavior change riding in a test-only commit.

## Verification (measured)

- `agent-actions.test.ts` — **56/56 passed** (was 15 failed / 41 passed)
- full plugin suite — **192 passed / 19 files** (was 180 passed, 15 failed)

Tests only; no product file is touched. An FNXC note now records the arity contract and which actions take which source, so the next assertion added here doesn't reintroduce a dead guard.

### DIFF
--- a/plugins/fusion-plugin-even-realities-glasses/src/__tests__/agent-actions.test.ts
+++ b/plugins/fusion-plugin-even-realities-glasses/src/__tests__/agent-actions.test.ts
@@ -348,6 +348,25 @@ const todoIsReviewIr = {
   ],
 };
 
+/*
+FNXC:GlassesAgentActions 2026-07-30-22:35:
+THE ARITY CONTRACT: a `moveTask` assertion must spell the move SOURCE, and which source is not uniform.
+
+`startWork`, `requestReview` and `approvePlan` pass `{ moveSource: "user" }` — the wearer's tap is a
+human gesture, matching the dashboard move route. `returnToAgent` and `retryTask` deliberately DO NOT:
+per the Move-Task contract a user-source move parks the row `userPaused`, which would defeat the
+return/retry intent. So the correct assertion differs per action and cannot be applied uniformly.
+
+Why this is written down: when the option was added to the three user-source actions, the assertions
+here kept the two-argument form and 15 tests went red on `main` — this suite sits outside the merge
+gate, so nothing blocked it. Worse, the two-argument form is not merely stale in the NEGATIVE case,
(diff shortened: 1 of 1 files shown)


gsxdsm:

PR #2971, opened 2026-07-31
fix(tests): clear the persisted Command Center sub-tab between cases (7 of 8 main reds)

gsxdsm:
`main` is red in the dashboard backfill lane. This fixes **7 of the 8** failures. All 8 bisect to #2420 (`4f929acc10`): parent `189f237a07` passes 9/9, that commit fails 5.

### Cause — test-order pollution, not a product bug

#2420 made Command Center restore its sub-tab on remount, because the view unmounts on navigation by design:

```ts
const [activeTab, setActiveTab] = useState<SubViewId>(
  () => (getCommandCenterState(projectId)?.activeTab as SubViewId | undefined) ?? "overview",
);
```

The panel's test id is derived from that tab (`data-testid={\`command-center-panel-${activeTab}\`}`), and these files click through to other tabs — `tokens`, `team`, `github`, `system`, `mission-control`. Neither `beforeEach` cleared storage, so the **first** case left `mission-control` persisted and every later case rendered `command-center-panel-mission-control`:

```
→ Unable to find an element by: [data-testid="command-center-panel-overview"]
```

Nothing in that message points at a previous test, which is what made it look like a component regression.

**Confirmed as ordering rather than breakage:** each failing case passes when run alone with `-t`. The "mutation" here is `main` itself — without the `localStorage.clear()` these files fail 7; with it, 11/11.

### Scope

Two lines plus the note explaining why they exist, so the next person who adds a tab-switching case knows the persistence is per-project and sticky. **No product code touched** — the persistence behaviour in #2420 is correct and stays as-is.

**Verified:** 11/11 across both files, `tsc -p tsconfig.app.json` 0 errors, lint clean, FNXC gate exit 0. Test-only, no changeset.

### The 8th failure is NOT fixed here, deliberately

`MainContent.planning-project-remount.test.tsx` fails because #2420 moved Planning out of `MainContent` (its branch now returns `null`) into `PlanningKeepAlive`, mounted by `App.tsx`. The product side is right — the host **is** keyed (`App.tsx:1956`):

```jsx
<PlanningKeepAlive key={`${currentProject.id}:${modalManager.planningEntryGeneration}`} … />
```

so project switches still remount and I found **no cross-project leak**. But that test was `FNXC:ProjectSwitchModalReset` coverage for a leak-class invariant (Planning carrying a previous project's stream/session), and **nothing asserts it at the new location**: the keep-alive test covers navigation reveal, not project switching, and `App.test.tsx` has no test for the host key. Deleting that `key=` today would fail no test.

Restoring it needs an App-level test, which is a bigger change than this fix and worth keeping separate. Detailed on #2420.

### DIFF
--- a/packages/dashboard/app/components/command-center/__tests__/CommandCenter.mobile-scroll.test.tsx
+++ b/packages/dashboard/app/components/command-center/__tests__/CommandCenter.mobile-scroll.test.tsx
@@ -368,6 +368,20 @@ async function openChartTab(tab: string) {
 
 describe("CommandCenter mobile scroll regression (FN-6595)", () => {
   beforeEach(() => {
+  /*
+  FNXC:CommandCenter 2026-07-30-22:10:
+  CLEAR THE PERSISTED SUB-TAB — cases in this file are no longer independent without it.
+
+  #2420 made `activeTab` initialise from per-project persisted state
+  (`getCommandCenterState(projectId)?.activeTab`) instead of always `"overview"`, because Command
+  Center unmounts on navigation by design and has to restore its sub-tab on remount. These cases
+  click through to other tabs, so the FIRST case now leaves `mission-control` persisted and every
+  later case renders `command-center-panel-mission-control` — the `command-center-panel-overview`
+  lookups then fail with no hint that the cause is a previous test.
+
+  Confirmed as ordering, not breakage: each failing case passes when run alone with `-t`.
+  */
+  localStorage.clear();
     apiMock.mockReset();
     mockOverviewApi();
     injectCommandCenterCss();
--- a/packages/dashboard/app/components/command-center/__tests__/CommandCenter.tablet-layout.test.tsx
+++ b/packages/dashboard/app/components/command-center/__tests__/CommandCenter.tablet-layout.test.tsx
@@ -391,6 +391,20 @@ async function openChartTab(tab: string) {
 
 describe("CommandCenter tablet layout regression (FN-6679)", () => {
   beforeEach(() => {
+  /*
+  FNXC:CommandCenter 2026-07-30-22:10:
+  CLEAR THE PERSISTED SUB-TAB — cases in this file are no longer independent without it.
+
+  #2420 made `activeTab` initialise from per-project persisted state
+  (`getCommandCenterState(projectId)?.activeTab`) instead of always `"overview"`, because Command
+  Center unmounts on navigation by design and has to restore its sub-tab on remount. These cases
+  click through to other tabs, so the FIRST case now leaves `mission-control` persisted and every
+  later case renders `command-center-panel-mission-control` — the `command-center-panel-overview`
+  lookups then fail with no hint that the cause is a previous test.
+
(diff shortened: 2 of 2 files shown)


gsxdsm:
Rebased onto current main (post-#2966/#2968); diff is still exactly the two CommandCenter test files, 11/11 green.

**Update on the 8th failure, which this PR deliberately leaves red:** I attempted the App-level test that would restore the Planning project-switch coverage and **withdrew it**. It passed under two separate mutations of the guard it was supposed to protect — dropping the project id from `App.tsx:1956`'s key, and loosening the `planningEverOpenedProjectId === currentProject.id` gate at `:1954`. DOM-node identity, which normally distinguishes a remount from a re-render, turned out to change on a project switch for reasons incidental to both guards.

I would rather leave one test red and visible than merge a green one that guards nothing. Details and the mutation table are on #2420.

To be explicit about what this PR does and does not claim: it fixes an **ordering** defect in the CommandCenter tests and nothing else. `MainContent.planning-project-remount.test.tsx` stays red because Planning no longer renders from `MainContent` at all, and the invariant it guarded now has no assertion anywhere.

gsxdsm:
Took the 8th failure since it is fallout from #2420, which I resolved. **I could not write a load-bearing test for it, and I am reporting that rather than shipping one that passes regardless.**

## What I tried, and why each shape is worthless here

| assertion | mutation | result |
|---|---|---|
| planning subtree is gone after the switch | delete `App.tsx`'s `key={...}` | **passes** — the gate unmounts it anyway |
| host is a different DOM node after the switch | delete the `key` | **passes** |
| same, but mutate the **gate** so the latch never clears | break `setPlanningEverOpenedProjectId(... : null)` | **passes** |

`rerender(<App />)` remounts that subtree in this harness regardless of either mechanism, so DOM node identity — the technique AGENTS.md prescribes for exactly this class, and the one I reached for — carries no signal at the App level. All three shapes are decorative.

## What I did learn about the invariant

The gate at `App.tsx:1954` is **latched on `currentProject.id`**:

```ts
if (planningLatchProjectIdRef.current !== projectId) {
  planningLatchProjectIdRef.current = projectId;
  setPlanningEverOpenedProjectId(planningViewActive && projectId ? projectId : null);
  return;
}
```

On a project switch the render happens with the OLD latch and the NEW project, so the gate is false for one render and the subtree unmounts before the effect re-latches. That transient unmount is what actually guarantees a fresh Planning instance per project — **not** the `key`. So the `FNXC:ProjectSwitchModalReset` leak class is protected, by the latch rather than by the key, and the key's remaining job is the `planningEntryGeneration` half (fresh-open semantics), which is a different invariant.

That is worth knowing before anyone deletes either one: the key looks redundant for project switches specifically, and is not redundant for entry-generation.

## Not done

A test for the `planningEntryGeneration` half. It needs the App-level state machine driven directly rather than through `rerender`, which is a different harness than any of the three above. After three non-load-bearing attempts I would rather hand over the analysis than a fourth guess.

Your two-line `localStorage.clear()` fix for the other 7 stands on its own — nothing here affects it.

PR #2972, opened 2026-07-31
fix(dependency-graph): an allowlist of four legacy lanes rendered a blank graph on a renamed board

gsxdsm:
## A renamed board gets a blank dependency graph

`filterGraphTasks` gated on an allowlist of four legacy lane ids:

```ts
export const INCLUDED_COLUMNS = new Set(["triage", "todo", "in-progress", "in-review"]);
export function filterGraphTasks(tasks: Task[]): Task[] {
  return tasks.filter((task) => INCLUDED_COLUMNS.has(task.column));
}
```

On a board whose lanes are named anything else — `backlog`, `building`, `checking` — **no card matches and the graph renders completely empty**. This is not a mislabelled node or a missing edge: the entire feature is blank, and it reads as *"this project has no dependencies"* rather than as a bug. `triage` is in that allowlist too, a lane U11 (#2515) deleted.

## Fix: gate on the finished lanes instead

Inverted to a denylist, so the **default is the safe one**. An unrecognised lane is active work by assumption and renders; only lanes that genuinely mean "finished" drop out.

An allowlist fails **closed** — hide everything unknown. A denylist fails **open** — show it. For a graph, an extra node is a far smaller error than no graph.

## The residual, named rather than hidden

`EXCLUDED_COLUMNS` is still two literals. `DependencyGraph.tsx` is a client React component handed plain `Task` rows as a prop, with no async seam to resolve a workflow IR — so a board that renames its DONE lane still shows finished cards here. That is deliberately the mild failure direction: "one extra node", not "no graph". It is documented in the code rather than papered over with an optional resolved-lanes parameter no caller could fill.

The invalid-column guard is now **explicit**. Under the allowlist, `column: undefined` was excluded as a side effect of not being in the set; under a denylist it would sail through, so it is checked directly and covered.

## Revert proof

Restoring only `filters.ts`:

```
AssertionError: expected [] to deeply equal [ 'FN-1', 'FN-2', 'FN-3' ]
AssertionError: expected [] to deeply equal [ 'FN-1' ]
      Tests  2 failed | 13 passed (15)
```

Worth noting explicitly: **every pre-existing case passes either way.** They only ever enumerate the six legacy ids, so the allowlist and the denylist agree on all of them — no existing test could have seen this blackout. (The new empty-string case also passes both ways; it guards the new implementation rather than proving the fix, and I am not claiming it as coverage of the defect.)

## Verification (measured)

- plugin suite — **183 passed / 20 files**
- `lifecycle-column-census --strict`, `check-fnxc-future-dates` — green

## Checked and deliberately not changed

`GraphTaskNode.tsx` holds `column === "in-progress"` and `column === "in-review"`. Both are already audited with an in-code note, and both degrade mildly rather than blanking anything — `hasExecutionSignal` also ORs on `ACTIVE_STATUSES`, so a renamed WIP lane still reads as active via status. Converting them needs column traits this component is not given, so they stay noted rather than half-converted.

No changeset: `fusion-plugin-dependency-graph` is `private: true`.

### DIFF
--- a/plugins/fusion-plugin-dependency-graph/src/__tests__/filters.test.ts
+++ b/plugins/fusion-plugin-dependency-graph/src/__tests__/filters.test.ts
@@ -67,6 +67,40 @@ describe("filterGraphTasks", () => {
     expect(filterGraphTasks([invalidTask])).toEqual([]);
   });
 
+  /*
+  FNXC:WorkflowLifecycleColumns 2026-07-30-23:05:
+  THE INVARIANT: a card in a lane this filter has never heard of is still a graph node.
+
+  Reverted to the `INCLUDED_COLUMNS` allowlist, this case returns `[]` for three live cards — which is
+  what a renamed board got: an entirely blank dependency graph, indistinguishable from "no dependencies".
+
+  Every case above stays green either way, which is the point: they only ever enumerate the six legacy
+  ids, so the allowlist and the denylist agree on all of them. Nothing here could see the blackout.
+  */
+  it("renders cards from lanes it does not recognise, instead of blanking the graph", () => {
+    const renamed = [
+      createTask("FN-1", "backlog" as Task["column"]),
+      createTask("FN-2", "building" as Task["column"]),
+      createTask("FN-3", "checking" as Task["column"]),
+    ];
+
+    expect(filterGraphTasks(renamed).map((task) => task.id)).toEqual(["FN-1", "FN-2", "FN-3"]);
+  });
+
+  it("still drops finished lanes when the rest of the board is renamed", () => {
+    const mixed = [
+      createTask("FN-1", "building" as Task["column"]),
+      createTask("FN-2", "done"),
+      createTask("FN-3", "archived"),
+    ];
+
+    expect(filterGraphTasks(mixed).map((task) => task.id)).toEqual(["FN-1"]);
+  });
+
+  it("excludes an empty-string column, which names no lane at all", () => {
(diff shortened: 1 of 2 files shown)


gsxdsm:

gsxdsm:
## Measured: the census reads **0** for this entire area, and two blackouts lived in it

Worth recording where it will be found, because "census clean" is being used across the fleet as a completion signal.

Running `lifecycle-column-census.mjs --json` and filtering to `packages/cli/` + `plugins/`:

```
MY AREA (column guards): 0 files, 0 guards
MY AREA (query filters): 0 files
```

`src/filters.ts` holds **six** lane literals — including the four-id allowlist this PR fixes — and appears in **none** of the census outputs:

```
byFile:           []
deliberateByFile: []
queryByFile:      []
```

It is invisible because the ids sit inside `new Set([...])` — the lane-shaped Set/list literal class the census does not match. The same blindness hid the glasses summary card (#2968), where five ids sat inside a template string.

**Both defects were total feature blackouts** — an empty dependency graph and a summary card reporting `Triage 0 Todo 0 Doing 0 Review 0 Done 0` for a full board — and both were found by reading the files, not by any gate. Neither the census, nor `unwired-lane-parameter`, nor the test suites could see either one; every pre-existing test passed before and after.

So: a zero census count for a directory is **not** evidence that the directory is converted. It is evidence that nothing in it matches the patterns the census knows. I would not treat my area as clean on that number, and I would not let another batch close out on it either.

PR #2974, opened 2026-07-31
fix(gate): the lane-wiring census could not see its own motivating case (#2956)

gsxdsm:
#2966 shipped a gate that **cannot detect the defect named first in its own header.**

`findLaneAcceptingFunctions` matched a lane parameter only when `param.type` was a `TypeLiteralNode` — an inline `{ reviewColumns?: … }`. But the real code declares these as interfaces:

```ts
export function getInReviewStallReason(
  task: Pick<Task, …>,
  context: InReviewStallContext = {},   // TypeReference — invisible
): InReviewStallSignal | undefined
```

so the function never entered `accepting` and none of its call sites were examined.

### Measured, both directions

| | before | after |
|---|---|---|
| lane-accepting functions detected | 20 | **30** |
| `getInReviewStallReason` detected | no | **yes** |
| re-introduce #2956 (drop `reviewColumns` from one call site) | `none added` — **passes** | **fails**: `reads.ts: 7 unwired now, baseline allows 6` |

The gate now catches the thing it was built for.

### The baseline moves 10 → 24, and that number needs context

`10 unwired call site(s) across 8 files` → `24 across 15`. **No entry was removed** — every previously-recorded file kept its count and 14 sites became visible for the first time:

```
core/task-store/reads.ts                        0 -> 6
engine/self-healing.ts                          2 -> 4
core/task-store/branch-and-pr-entities.ts       0 -> 1
core/task-store/task-update.ts                  0 -> 1
engine/scheduler.ts                             0 -> 1
dashboard/routes/register-task-workflow-routes  0 -> 1
cli/commands/dashboard-tui/bucket-mapping.ts    0 -> 1
cli/extension.ts                                0 -> 1
```

**These are newly VISIBLE, not newly broken** — they have been unwired all along. I have **not** audited them, and recording them in the baseline is not a claim that they are fine; it is the ratchet doing what its header describes, since the census's own note says roughly half of the original hits were legitimately unwired (identity proven by a stronger means, sentinel columns, dead exports). Someone should walk the 14. Two stand out as worth a look first: **`reads.ts` at 6** is the file #2956 was about, and **`scheduler.ts`** is a dispatch path.

Flagging rather than fixing, because wiring a call site that should not be wired is its own defect and each needs the judgement call the census header describes.

### Regression test

`packages/engine/src/__tests__/lane-wiring-census-named-types.test.ts` pins the detector's shape — named interface, type alias, inline literal, positional — against fixtures rather than live counts, so it does not churn when someone legitimately wires a call site. Plus one anti-vacuity case asserting the named-type arm is still load-bearing on real source (`getInReviewStallReason` resolves in the live tree), so the fixtures cannot pass while the tool has quietly stopped applying here.

**Mutation:** removing the `TypeReference` arm fails **4 of 5**.

### Also worth knowing

`findLaneAcceptingFunctions` still only visits `ts.isFunctionDeclaration` at top level, so `export const fn = (ctx) => …` remains invisible. I checked — no exported arrow function currently takes a lane argument, so nothing is missed today, and I left it rather than widen the surface in the same change.

Resolved by **name across the corpus** instead of a type-checker `Program`: these are plain source scans and a checker would cost a full type-resolution pass for one lookup. Two same-named types merge, which only ever widens what counts as wired — safe for a ratchet.

**Verified:** 5/5 new tests, `check-lane-wiring` clean at the new baseline, lint clean, FNXC gate exit 0. Core suite on main is green (4923 passed / 0 failed) — unrelated, but I had it running.

## Summary by CodeRabbit

* **New Features**
  * Improved lane-wiring analysis to recognize named interfaces and type aliases.
  * Added support for wrapped configuration expressions and positional parameters when detecting lane information.

* **Tests**
  * Added comprehensive coverage for lane-wiring detection, including named contexts and live-tree validation.

* **Chores**
  * Updated baseline counts to reflect newly recognized application areas and improved self-healing detection.

### DIFF
--- a/packages/engine/src/__tests__/lane-wiring-census-named-types.test.ts
+++ b/packages/engine/src/__tests__/lane-wiring-census-named-types.test.ts
@@ -1,0 +1,104 @@
+// @vitest-environment node
+
+/*
+FNXC:WorkflowLifecycleColumns 2026-07-30-23:50:
+THE LANE-WIRING CENSUS MUST SEE NAMED CONTEXT TYPES, not only inline type literals.
+
+`scripts/lib/lane-wiring-census.mjs` recognises a lane-accepting function so its call sites can be
+checked for a resolved-lane argument. Its first version matched only `param.type` being a
+`TypeLiteralNode`, so a parameter typed by an INTERFACE was invisible — and that is how the real
+code is written:
+
+  export function getInReviewStallReason(task: …, context: InReviewStallContext = {})
+
+The consequence was not academic. The census shipped unable to detect #2956 — the first case listed
+in its own header — and re-introducing that defect left it reporting "none added". Fixing the
+detector moved the honest count from 10 unwired call sites across 8 files to 24 across 15.
+
+Fixtures rather than the live tree: a test asserting counts over real source would fail every time
+someone legitimately wires or adds a call site, which is the churn the baseline exists to absorb.
+These pin the DETECTOR's shape instead. The final case is the anti-vacuity check — it asserts the
+real corpus still exercises the named-type arm, so the fixtures cannot pass while the tool has
+silently stopped applying to this codebase.
+*/
+
+import { mkdtempSync, writeFileSync, readdirSync, statSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join, resolve } from "node:path";
+import { describe, expect, it } from "vitest";
+import { findLaneAcceptingFunctions, findUnwiredCallSites } from "../../../../scripts/lib/lane-wiring-census.mjs";
+
+function fixture(source: string): string[] {
+  const dir = mkdtempSync(join(tmpdir(), "fusion-lane-census-"));
+  const file = join(dir, "fixture.ts");
+  writeFileSync(file, source);
+  return [file];
+}
+
+const REPO_ROOT = resolve(__dirname, "../../../..");
+
+describe("the lane-wiring census recognises how lane arguments are actually declared", () => {
+  it("detects a lane member on a NAMED context interface (the #2956 shape)", () => {
+    const files = fixture(`
+      export interface StallContext { now?: number; reviewColumns?: ReadonlySet<string>; }
+      export function getSignal(task: string, context: StallContext = {}): string { return task; }
+    `);
+    const accepting = findLaneAcceptingFunctions(files);
+    expect(accepting.has("getSignal")).toBe(true);
+    expect([...accepting.get("getSignal")!.names]).toContain("reviewColumns");
+  });
+
+  it("counts a call site that omits the lane argument on that named-type function", () => {
+    const files = fixture(`
+      export interface StallContext { reviewColumns?: ReadonlySet<string>; }
+      export function getSignal(task: string, context: StallContext = {}): string { return task; }
+      export function wired() { return getSignal("a", { reviewColumns: new Set<string>() }); }
+      export function unwired() { return getSignal("b", { now: 1 } as never); }
+    `);
+    const unwired = findUnwiredCallSites(files, findLaneAcceptingFunctions(files));
+    expect(unwired.map((hit) => hit.fn)).toEqual(["getSignal"]);
+  });
+
+  it("still detects the inline type-literal and positional spellings", () => {
+    const files = fixture(`
+      export function inlineBag(task: string, opts: { terminalColumns?: ReadonlySet<string> }): string { return task; }
+      export function positional(task: string, activeColumns: ReadonlySet<string>): string { return task; }
+    `);
+    const accepting = findLaneAcceptingFunctions(files);
+    expect([...accepting.get("inlineBag")!.names]).toContain("terminalColumns");
+    expect([...accepting.get("positional")!.positions]).toEqual([1]);
+  });
+
+  it("type aliases carry lane members too", () => {
+    const files = fixture(`
+      export type MergeContext = { completeColumns?: ReadonlySet<string> };
+      export function canMerge(task: string, context: MergeContext): string { return task; }
+    `);
+    expect([...findLaneAcceptingFunctions(fixture("")).keys()]).toEqual([]);
+    expect([...findLaneAcceptingFunctions(files).get("canMerge")!.names]).toContain("completeColumns");
+  });
+
+  /*
+  ANTI-VACUITY against the live tree. The fixtures above would keep passing if the named-type arm
+  stopped mattering here — if every context interface were inlined, or the census stopped being
+  pointed at core. This asserts the arm is still load-bearing on real source.
+  */
+  it("resolves a real named-context function in the live tree", () => {
+    function sources(dir: string, out: string[] = []): string[] {
+      for (const entry of readdirSync(dir)) {
+        const path = join(dir, entry);
+        if (statSync(path).isDirectory()) {
+          if (entry === "__tests__" || entry === "node_modules" || entry === "dist") continue;
+          sources(path, out);
+        } else if (entry.endsWith(".ts") && !entry.endsWith(".d.ts") && !entry.includes(".test.")) {
+          out.push(path);
+        }
+      }
+      return out;
+    }
+    const accepting = findLaneAcceptingFunctions(sources(join(REPO_ROOT, "packages/core/src")));
+    expect(accepting.size).toBeGreaterThan(5);
+    // Declared as `context: InReviewStallContext` — invisible to the census before this arm existed.
+    expect(accepting.has("getInReviewStallReason")).toBe(true);
+  });
+});
--- a/packages/engine/src/__tests__/lane-wiring-census-named-types.test.ts
+++ b/packages/engine/src/__tests__/lane-wiring-census-named-types.test.ts
@@ -13,7 +13,11 @@ code is written:
 
 The consequence was not academic. The census shipped unable to detect #2956 — the first case listed
 in its own header — and re-introducing that defect left it reporting "none added". Fixing the
-detector moved the honest count from 10 unwired call sites across 8 files to 24 across 15.
+detector moved the honest count from 10 unwired call sites across 8 files to 18 across 14.
+
+The first draft of that fix reported 24, because detecting these functions also exposed a SECOND bug
+(the `satisfies` case below) that had been hidden while they were invisible: six of those "new" hits
+were correct code. Both fixes ship together, or the baseline records six sites nobody can act on.
 
 Fixtures rather than the live tree: a test asserting counts over real source would fail every time
 someone legitimately wires or adds a call site, which is the churn the baseline exists to absorb.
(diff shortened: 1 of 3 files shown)


gsxdsm:
### Correction: 6 of the 14 sites I listed were false positives. Second commit fixes that.

I went to audit the newly-visible hits before anyone acted on them, starting with `reads.ts` at 6 — the file #2956 was about. **All six were already wired.** Every `getInReviewStalledSignal` / `getStalePausedReviewSignal` call site in that file passes `reviewColumns`.

The census was wrong, not the code:

```ts
getInReviewStalledSignal(task, { now, executingTaskIds, reviewColumns } satisfies InReviewStalledContext)
```

`findUnwiredCallSites` tested `ts.isObjectLiteralExpression(arg)` directly, and a **`SatisfiesExpression` is not an object literal** — it wraps one. So a fully-wired call read as unwired. I introduced that `satisfies` annotation myself in #2956, which is a pleasing way to be caught by my own tooling.

**This bug predates my change but was invisible behind it:** functions typed by an interface were never detected, so their call sites were never inspected and the false positives never surfaced. Fixing detection exposed six at once. Both fixes have to ship together — otherwise the baseline permanently records six sites nobody can "fix", and the first person to check one learns the number is noise.

Second commit unwraps `satisfies`, `as`, and parentheses.

### Corrected numbers

| | first commit (wrong) | now |
|---|---|---|
| baseline | 24 across 15 files | **18 across 14** |
| newly visible | 14 | **8** |

The 8 that remain, none audited:

```
core/in-review-stall.ts                         +1
core/task-store/branch-and-pr-entities.ts       +1
core/task-store/task-update.ts                  +1
engine/scheduler.ts                             +1
engine/self-healing.ts                          2 -> 4
dashboard/routes/register-task-workflow-routes  +1
cli/commands/dashboard-tui/bucket-mapping.ts    +1
```

`reads.ts` is gone from the list entirely — it was never unwired.

### Mutations

| mutation | result |
|---|---|
| re-introduce #2956 (drop `reviewColumns` from a call site) | **fails** — `reads.ts: 1 unwired, baseline allows 0` |
| remove the `satisfies` unwrap | **6 false positives return** |
| remove the `TypeReference` arm | **4 of 6 tests fail** |

Note the first mutation is now sharper than in the opening commit: `reads.ts` sits at a baseline of **0**, so a single regression there fails immediately rather than hiding under an allowance of 6.

Test file gains a `satisfies` case, and I corrected the stale `24 across 15` figure in its header comment.

**Verified:** 6/6 tests, `check-lane-wiring` clean at 18, lint clean, FNXC gate exit 0. Main health while I was here: core **4923 passed / 0 failed**, engine **11466 / 0**.

gsxdsm:
Good catch on the TypeReference hole — that was a real blind spot and `getInReviewStallReason` being invisible to its own gate is exactly the right thing to have found.

Two problems with this revision, both measured against current `main` (which has #2970). **Neither is a reason to hold the TypeReference fix** — they're one small patch on top.

### 1. Six of the 14 newly-visible sites are false positives, and baselining them re-hides the defect

All six new `reads.ts` entries **do** pass `reviewColumns`. They're invisible because the argument is `satisfies`-wrapped:

```ts
task.stalePausedReview = getStalePausedReviewSignal(task, {
  now, thresholdMs, reviewColumns: reviewColumnsForRow, ...
} satisfies StalePausedReviewContext);   // <- SatisfiesExpression, not ObjectLiteralExpression
```

`ts.isObjectLiteralExpression(arg)` is false for `{...} satisfies T`, so `passesOption` walks past a correctly-wired call.

A baselined false positive is **worse than a missed one**: the count already allows the site, so deleting the argument for real doesn't change the total.

| with `reviewColumns` genuinely deleted from `reads.ts:264` | reads.ts | TOTAL | gate |
|---|---|---|---|
| this PR's detector | 6 (unchanged) | 23 (unchanged) | **passes** |
| with the patch below | 0 → **1** | 17 → **18** | **fails** |

So on this branch the gate is green for the precise defect it exists to catch, at six of the sites it just added.

```js
function unwrapObjectLiteral(node) {
  let current = node;
  while (ts.isSatisfiesExpression(current) || ts.isAsExpression(current) || ts.isParenthesizedExpression(current)) {
    current = current.expression;
  }
  return ts.isObjectLiteralExpression(current) ? current : undefined;
}
```

…then test `unwrapObjectLiteral(arg)` instead of `ts.isObjectLiteralExpression(arg)`. Corrected count on main: **23 → 17**, no entry removed except the six `satisfies` ones.

### 2. The baseline was recorded before #2970 and regrows the allowance it closed

`git merge-base --is-ancestor` says this branch predates #2970. Its baseline carries `in-review-stall.ts: 1`; on current main that site is **0** — #2970 wired it and re-recorded the baseline downward in the same commit.

Merging as-is silently re-opens exactly one slot at the site of the false-stall bug. Re-record after rebasing; the corrected total should be **17**.

Worth adding a fixture covering `satisfies`, `as const`, and a parenthesized literal — that's the third detector hole in this gate's short life (TypeLiteral-only, positional params, now `satisfies`), and each was found by using it rather than by reading it.

gsxdsm:
Both findings are real and both are fixed in `87408ed`. They are the same defect in two places, and the direction is what makes them worth the effort.

## Both errors pointed at "more sites count as wired"

For a ratchet that is the one unsafe direction: every falsely-wired site **disappears from the census**, so the gate reports "none added" while the seam it watches is unsupplied. A gate whose mistakes land on *nothing to report* is the failure mode it exists to prevent.

That is also why I did not accept the existing justification. The note above `findLaneCarryingTypes` argued the name-merge was acceptable because a false member *"only widens what counts as wired"* — treating widening as the safe side. Here widening is the hazard, so the comment is corrected along with the code.

## 1. Options-bag parameter index — fixed as suggested

`names` was flat, so the call-site check accepted a match in **any** argument. `namesByIndex` now records which keys count in which argument, and the call check consults only the declaring position.

## 2. Same-named context types — fixed, but not with a type checker

The suggestion was to resolve by declaration identity via the checker. I did not do that: a full `Program` over ~1950 files for one lookup, against a scan that currently runs in ~2s, is a real cost on every PR.

Instead the unsound case is made **impossible to hit silently** — two files declaring lane-carrying types under one name is now a hard error naming both paths. That is strictly better than the silent union at near-zero cost, and it converts a quietly wrong answer into a loud one exactly when it would start mattering. If a collision ever fires and renaming is not viable, that is the moment to pay for declaration-identity resolution.

## Measured first, because it changed the shape of the fix

| | measured |
|---|---|
| call sites matching on a mismatched argument index | **0** |
| lane-carrying types in the tree | **2** |
| same-name collisions among them | **0** |

Both holes are **latent, not live** — which is why both fixes are behaviour-preserving today (baseline unchanged: `18 known unwired call site(s), none added`) and why finding 2 did not warrant heavy machinery.

## Verification

- 9 tests pass (was 6); 3 added — wrong-argument, its paired positive, and the collision throw
- **Mutation-verified**: reverting the index retention fails exactly the wrong-argument test; reverting the collision throw fails exactly the collision test; nothing else moves
- `check:lane-wiring`, `check:lifecycle-columns`, `check:sql-column-literals`, `check:inert-flag-seams`, `check:fnxc-future-dates` all exit 0; eslint clean

## One thing I changed that reviewers should know about

The `accepting` map's shape is now `{ namesByIndex, positions }` rather than `{ names, positions }`, so the three existing assertions on `.names` were updated. I deliberately did **not** keep a derived flat `names` alongside it for compatibility — a flat set is precisely what invited this bug, and leaving one available is an invitation to match on it again.

gsxdsm:
### I withdrew a commit of mine after your `87408ed` — you're right about the direction

I had a fourth commit ready treating a **spread-assembled options bag as WIRED**. `87408ed`'s reasoning rules against it, so I dropped it rather than rebase it on top.

My argument was that a false positive parks correct code in the baseline where nobody can act on it. Yours is the stronger one for a ratchet, and the asymmetry decides it:

- a false **positive** is recorded once and the gate is green again afterwards — the cost is a wasted audit;
- a false **negative** is a permanent hole, and the site never appears again to be caught.

Since I cannot see through a spread without the type checker we're both avoiding, counting it wired is a guess that manufactures holes. Dropped.

Worth separating from that: the **`satisfies` unwrap** in `8daf6ed` is not the same move. It doesn't widen on a guess — the properties are fully visible once the wrapper node is stepped through, so it makes the scan *correct* rather than *permissive*. That one stands.

### What the 18 actually contains, so the audit isn't wasted

Six of the 18 are spread-assembled calls that supply the lane answer through a spread the scan can't read:

```
SPREAD-ASSEMBLED  task-store/branch-and-pr-entities.ts:621  validateNodeOverrideChange   ({ ...overrideLanes })
SPREAD-ASSEMBLED  task-store/task-update.ts:99              validateNodeOverrideChange   ({ ...overrideLanes })
SPREAD-ASSEMBLED  auto-merge-finalization.ts:247            getTaskHardMergeBlocker
SPREAD-ASSEMBLED  project-engine.ts:3587                    getTaskHardMergeBlocker
SPREAD-ASSEMBLED  scheduler.ts:1528                         computeBlockerFanoutMap      (...(cond ? { escalationColumns } : {}))
SPREAD-ASSEMBLED  self-healing.ts:8723                      getStalePausedReviewSignal   (...activation)
```

**12 are plain calls that genuinely pass no options bag** — those are the real audit list. I hand-verified four of them (`in-review-stall.ts:255` `getTaskMergeBlocker(task)`, `register-task-workflow-routes.ts:5547` and `extension.ts:1822` `validateNodeOverrideChange(task, id)`, `bucket-mapping.ts:75` `bucketForTask(task)`): each calls a lane-accepting function with no context argument at all.

Recording the split rather than acting on it — under your principle these six belong in the baseline, and the note just stops the next person spending an hour proving `{ ...overrideLanes }` is fine.

### One thing I also added and then removed

Singular `reviewColumn` in `LANE_ARGUMENT_NAMES`. `getInReviewStalledSignal` reads `reviewColumns ?? reviewColumn ?? "in-review"`, so semantically it is a lane answer — but **no mutation distinguished it** (the call sites carrying it also carry a spread), and `unwired-lane-parameter.mjs` doesn't list it while the census set is documented as kept in step with that file. An unfalsifiable change that introduces drift is not worth carrying, so it went. Flagging the semantic gap: the singular spelling is absent from **both** lists, and a future caller passing only `reviewColumn` would be a false positive in the census and invisible to the sibling guard.

gsxdsm:
Rebased onto `main` (`be79fe0db6`, post-#2976). Your `87408ed` is preserved as the tip commit — I rebased rather than force-replacing so it stays intact.

The only conflict was `lane-wiring-baseline.json`, which is a **generated** file. Resolved by taking main's copy through the rebase and regenerating once at the end rather than hand-merging counts — a hand-merged baseline is a number nobody measured, which is the failure this whole PR is about.

**Baseline is now 16 sites across 12 files** (was 18 before the rebase; #2976 wired two).

Re-verified after the rebase, not assumed:

| check | result |
|---|---|
| re-introduce #2956 (drop `reviewColumns` from a call site) | **fails** — `reads.ts: 1 unwired, baseline allows 0` |
| census tests | 9/9 |
| `check-lane-wiring` | clean at 16 |
| lint / FNXC gate | clean / exit 0 |

The mutation is the one that matters: `reads.ts` sits at a baseline of **0**, so a single regression there fails immediately rather than hiding under an allowance.

PR #2975, opened 2026-07-31
fix(engine,core): role resolution was half-done in two shared lifecycle predicates (surfacing family + file-scope leases)

gsxdsm:
The three surfacing sweeps stopped reporting anything for a card resting in a board's **second** review or hold column.

A lifecycle role is a **trait**, and any number of columns may carry it. The shared runner resolved it with `resolveLifecycleColumns()[role]` — **first match** — then gated on it:

```ts
const roleColumn = lifecycle?.[spec.role];        // FIRST column carrying the trait
if (task.column !== resolved.roleColumn) continue; // everything else dropped
```

A workflow that splits human sign-off from the merge lane has two review columns; one that parks dependency-blocked cards separately has two hold columns. Cards in the second got **no stale-paused-todo, no stale-paused-review, no in-review-stalled** diagnostic — silently, with no error, on all three sweeps at once.

## The second bug hiding inside the fix for the first

Resolving membership but still reading `roleColumns[0]`'s declared `recovery` applies the **merge lane's** threshold to a card sitting in the **sign-off** lane. Each card's policy now comes from its own column, and one of the new cases fails if it doesn't: the first role column declares a policy that suppresses the signal, the card's own column declares one that fires.

## Reverted

| | |
|---|---|
| **6 of 12** new cases fail | `fires for a card in the SECOND column carrying its role` and `reads the recovery policy of the card's OWN role column` — × 3 sweeps |
| the other 6 pass either way | non-regression halves: still fires for the FIRST role column, still does **not** fire for a card outside every role column. Membership must widen the gate, not move it. |

The pre-existing 45 cases were all green throughout — the single-role-column fixture could not express the case, which is why the table-driven file that exists to stop these three sweeps drifting apart never caught it.

## Verification

`pnpm test:gate` 161 + 13 + 487 + 71 · surfacing family 57 · core stale-paused 20 · lint · census `--strict` · sql-literals · fnxc-dates · lane-wiring · changesets — all green.

## Note

`holdColumns` was missing from the lane-wiring vocabulary, so the gate could not see that argument dropped. Added in the same commit.

While reviewing, I found and measured **two problems in #2974** (comment posted there): six of its newly-visible sites are `satisfies`-wrapped false positives, and baselining them means deleting a real `reviewColumns` argument keeps the count unchanged and the gate green; and its baseline predates #2970, re-opening the slot that PR closed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stale-card detection across all applicable review and hold columns.
  * Cards are now surfaced using the policies configured for their specific lifecycle column.
  * Cards outside matching lifecycle columns are no longer incorrectly surfaced.
  * Preserved existing fallback behavior when no lifecycle columns are configured.

* **Tests**
  * Added coverage for workflows with split review and hold columns.

---

## Second commit: the same predicate, half-converted (`shouldHoldActiveFileScopeLease`)

Folded in here rather than stacked — same file, same class, and a stacked PR on an unmerged base is not mergeable. Reversible; say the word and I'll split it.

`shouldHoldActiveFileScopeLease` is the **scheduler's** lease predicate, shared with the self-healing repair paths deliberately so the two cannot disagree about who holds a file-scope lease. Its two role answers are optional parameters defaulting to the legacy ids. The scheduler's own call sites were converted to pass resolved answers; self-healing's two were not:

```ts
const isWipColumn    = options?.isWipColumn    ?? task.column === "in-progress";
const isReviewColumn = options?.isReviewColumn ?? task.column === "in-review";
```

On a renamed board neither branch matches, so the predicate returns `false` for every card. The scheduler kept the lease; self-healing saw none, cleared `overlapBlockedBy`, and **released a dependent to edit files another agent still holds** — the outcome `groupOverlappingFiles` exists to prevent.

Membership comes from the wip/review sets each sweep already resolved a few lines above, so this adds no reads.

**Reverted:** both new cases fail with `overlapBlockedBy` = `null` — the release itself, not a proxy. The pre-existing legacy-column case in the same file passes either way, because `in-progress` satisfies the literal default; that is exactly why it never caught this.

Lane-wiring baseline re-recorded `9 -> 7` in the same commit (the ratchet refused a stale allowance, as intended).

**Verification:** gate 161 + 13 + 487 + 71 · surfacing 57 · overlap-seam + scheduler-lease + query-blindness 79 · core stale-paused 20 · lint · census `--strict` · sql-literals · fnxc-dates · changesets — green.

### DIFF
--- a/.changeset/self-healing-lease-role-answers.md
+++ b/.changeset/self-healing-lease-role-answers.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Stale-dependency cleanup no longer releases a task to edit files another agent still holds, on renamed boards.
+category: fix
+dev: The two `shouldHoldActiveFileScopeLease` call sites in self-healing now pass resolved `isWipColumn`/`isReviewColumn` from the wip/review sets those sweeps already resolve, matching the scheduler's own call sites.
--- a/.changeset/surfacing-role-membership.md
+++ b/.changeset/surfacing-role-membership.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Stale-card diagnostics now cover every review and hold column, not just the first of each.
+category: fix
+dev: `runSurfacingSweep`'s role gate resolves membership (`resolveReviewColumns` / `columnsWithFlag(ir,"hold")`) instead of `resolveLifecycleColumns()[role]`; signals receive the column SET, and each card's recovery policy is read from its own column. Adds `StalePausedTodoContext.holdColumns`.
--- a/packages/core/src/stale-paused-todo.ts
+++ b/packages/core/src/stale-paused-todo.ts
@@ -26,6 +26,16 @@ export interface StalePausedTodoContext {
   `resolveLifecycleColumns(ir).hold` instead.
   */
   holdColumn?: string;
+  /*
+  FNXC:WorkflowResolvedColumns 2026-07-30-21:40 (the lane seam, MEMBERSHIP not one column):
+  `holdColumn` is `resolveLifecycleColumns().hold` — the FIRST column carrying the hold trait. A
+  board declaring more than one hold lane (a capacity wait beside a blocked-on-dependency park) has
+  several, and a card resting in the second read as not-on-hold and was never surfaced.
+
+  Mirrors `reviewColumns` in in-review-stalled.ts / stale-paused-review.ts. Optional, with today's
+  behaviour preserved as the fallback, so a caller that does not pass it is byte-identical.
+  */
+  holdColumns?: ReadonlySet<string>;
   now?: number;
   thresholdMs?: number;
   engineActiveSinceMs?: number;
@@ -38,8 +48,11 @@ export function getStalePausedTodoSignal(
   task: Pick<Task, "column" | "paused" | "columnMovedAt" | "updatedAt" | "pausedReason" | "pausedByAgentId">,
   context: StalePausedTodoContext = {},
 ): StalePausedTodoSignal | undefined {
-  const holdColumn = context.holdColumn ?? "todo";
-  if (task.column !== holdColumn || task.paused !== true) return undefined;
+  const onHoldLane = context.holdColumns
+    ? context.holdColumns.has(task.column)
+    /* DELIBERATE-LITERAL — the no-metadata fallback; a supplied set always wins. */
+    : task.column === (context.holdColumn ?? "todo");
+  if (!onHoldLane || task.paused !== true) return undefined;
 
   const thresholdMs = context.thresholdMs ?? DEFAULT_STALE_PAUSED_TODO_THRESHOLD_MS;
   if (!Number.isFinite(thresholdMs) || thresholdMs <= 0) return undefined;
--- a/packages/engine/src/__tests__/self-healing-fake-overlap-seam.test.ts
+++ b/packages/engine/src/__tests__/self-healing-fake-overlap-seam.test.ts
@@ -135,3 +135,120 @@ describe("SelfHealingManager fake TaskStore overlap seam", () => {
     manager.stop();
   });
 });
+
+/*
+FNXC:WorkflowResolvedColumns 2026-07-30-22:25 (the shared lease predicate was HALF-converted):
+`shouldHoldActiveFileScopeLease` is the scheduler's predicate, shared with self-healing on purpose so
+the two cannot disagree about who holds a file-scope lease. Its role answers are optional parameters
+defaulting to the legacy ids; the scheduler passes resolved answers and this sweep did not, so on a
+renamed board the scheduler kept a lease that this sweep saw as absent — and released a dependent to
+edit files another agent still holds.
+
+The board below is renamed but otherwise identical to the legacy case above, which is the point: the
+existing test passes either way because `in-progress` satisfies the literal default.
+*/
+const RENAMED_BOARD_IR = {
+  version: "v2",
+  id: "custom:renamed",
+  nodes: [],
+  edges: [],
+  columns: [
+    { id: "drafting", name: "drafting", traits: [{ trait: "hold", config: { release: "capacity" } }] },
+    { id: "building", name: "building", traits: [{ trait: "wip", config: { limitSetting: "maxConcurrent" } }] },
+    { id: "checking", name: "checking", traits: [{ trait: "merge" }] },
+    { id: "shipped", name: "shipped", traits: [{ trait: "complete" }] },
+  ],
+};
+
+describe("SelfHealingManager stale-blocker cleanup on a RENAMED board", () => {
+  function createRenamedBoardStore(seed: Task[]) {
+    const tasks = new Map(seed.map((task) => [task.id, task]));
+    const settings = {
+      globalPause: false,
+      enginePaused: false,
+      mergeRequestContractShadowEnabled: true,
+    } as Settings;
+    const store = {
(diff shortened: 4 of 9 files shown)


gsxdsm:
**Not rebasing this — its owner is resolving the conflict right now, and a rebase from here would destroy that work.**

The branch is checked out in the `kb/capacity-global-cap` worktree, which is mid-resolution:

```
last commit activity: 4 seconds ago
UU scripts/lib/lane-wiring-baseline.json      <- unmerged, conflict markers in the tree
A  .changeset/self-healing-lease-role-answers.md
M  packages/engine/src/self-healing.ts
M  packages/engine/src/__tests__/self-healing-fake-overlap-seam.test.ts
```

`UU` is an in-progress merge with an unresolved conflict on **exactly the file this rebase would touch**, sitting next to staged uncommitted work. Force-pushing a rebased branch underneath that would strand the resolution in progress and the staged changes with it — and unlike a bad merge, there is no ref to recover them from.

So the automated "CONFLICTING, rebase now" signal is accurate about the state and wrong about the actor: **the owner is already on it**, and has been for seconds, not hours. The retry loop is racing a human, not a stalled branch.

## Why the conflict exists, in case it helps whoever is holding it

`scripts/lib/lane-wiring-baseline.json` moved on main when #2966 landed the lane-wiring ratchet, and this branch edits the same file. The baseline is generated, so the resolution is not a merge of the two texts — take main's version, then re-run:

```
pnpm check:lane-wiring
```

which rewrites the baseline from the tree and exits 0 on a drop. Hand-merging the JSON produces a number that matches neither branch.

Worth knowing: **that gate writes its baseline as a side effect of running**, so the file changes underfoot for anyone who runs any gate. It is the same mechanism that made #2962 redundant before it was reviewed.

I did not touch the branch, the worktree, or the PR beyond this comment.

gsxdsm:
### The 6 sites this PR leaves in the baseline, each checked rather than assumed

An allowance nobody has justified is the same defect as a missing check, so:

| site | why it is not a defect |
|---|---|
| `task-merge.ts:432` `isTaskReadyForMerge` | **no source caller** — appears only in `dist/` and the barrel re-export |
| `moves.ts:660`, `:834` | pass `skipColumnIdentityCheck`; lane identity is proven by a stronger means at both |
| `auto-merge-finalization.ts:247` | spreads a **review-eligible sentinel** deliberately (FNXC 2026-06-29): proven `mergeConfirmed` means the branch landed, so the column mismatch must not block the rehome |
| `project-engine.ts:3587` | same sentinel spread, documented by FNXC 2026-07-30-18:05 — the bug it names is already fixed |
| `in-process-runtime.ts:427` | `selectActionablePlanningContinuations` has **no production caller**, established independently in `workflow-planning-continuation-terminal-gap-live-e2e.pg.test.ts` |

The last one is worth flagging even though it is not live: `resolvePlanningContinuationCandidate` defaults `terminalColumns` to the legacy `done`/`archived` pair, and its only caller passes nothing. If that function ever gains a production caller, a renamed completion lane re-enters plan-review after the card left the board — exactly what the comment above it says the exclusion exists to prevent. It is dead code today, so wiring it would be unverifiable; the ratchet keeps it visible instead.

gsxdsm:
Pre-flighted this because the arity trap is the defect class my unit has been chasing, and it holds up. **62/62 on the two new suites**, and the mutation that matters bites:

| mutation | result |
|---|---|
| force `resolveRoleColumns` back to first-match (`.slice(0, 1)`) | **8 failed / 49 passed** |

So the new coverage genuinely pins *membership*, not just the presence of a resolver — which is the failure mode I'd expect here, since a test can assert lanes were resolved while the gate downstream still compares against one id.

The framing is exactly right and worth repeating for anyone touching these seams:

> Resolving membership but still reading `roleColumn` …

That is the trap. `resolveLifecycleColumns()[role]` answers with the **first** column carrying a trait, `columnsWithFlag` answers with **every** one, and the two are interchangeable on every single-column board — which is every board in most fixtures. I hit the same asymmetry from the other side on #2804, where an AST guard I wrote collected `columnsWithFlag` calls file-wide and two of six flags were consequently unfalsifiable. Same shape: the arity of the answer is the thing that silently differs.

### One coordination item — this is now a THREE-way conflict on the baseline

This PR, #2974, and #2978 all regenerate `scripts/lib/lane-wiring-baseline.json`. Measured combinations:

| state | baseline |
|---|---|
| #2978 alone | 18 sites / 13 files |
| #2974 alone | 16 / 12 |
| #2974 + #2978 | **30 / 22** |

This PR moves it again (`+1/-2`). None of the three stated numbers survives contact with the other two, so whichever lands second and third must run `node scripts/check-lane-wiring.mjs --update-baseline` **in the same commit** rather than resolving the JSON by hand. A hand-merged baseline is a count nobody measured, which is precisely what the ratchet exists to prevent.

Also worth noting this PR edits `lane-wiring-census.mjs` itself (`+3/-0`) while #2974 rewrites its detector — that one is a real content conflict, not just a generated-file collision.

gsxdsm:
**Rebased onto main and pushed** (`e3843f2b`) — now MERGEABLE.

I held off on this earlier: when the rebase was first requested, the `kb/capacity-global-cap` worktree holding this branch was mid-resolution — `UU scripts/lib/lane-wiring-baseline.json`, staged changes beside it, last activity 4 seconds prior. Rebasing then would have stranded that work with no ref to recover it from. Re-checked before acting: the worktree is now clean (0 dirty files, last commit 11 minutes ago), so there was nothing in flight to lose.

**Heads-up for whoever holds that worktree:** the branch history was rewritten, so it is now behind its own remote. `git fetch && git reset --hard origin/surfacing-role-membership` will realign it — nothing uncommitted was there when I pushed, but check before resetting in case you have started something since.

## The conflict, and why it is not a merge

`scripts/lib/lane-wiring-baseline.json` — **generated**. Took main's version and regenerated rather than merging the two texts; hand-merging that JSON yields a count matching neither branch while looking entirely plausible.

## One thing the rebase surfaced that needed fixing

Post-rebase the gate reported a **drop** this PR causes:

```
  packages/engine/src/self-healing.ts: 4 -> 0
```

That is this change working — the file-scope-lease fix wired all four sites. But unlike the FNXC date ratchet, this gate **requires the re-record in the same commit**, and without it the allowance of 4 stays open for those sites to regrow into silently. Re-recorded and amended into the commit that caused the drop, so the two travel together. Baseline now reads `12 known unwired call site(s), none added`.

## Verification

- `surfacing-family-shared` + `self-healing-fake-overlap-seam`: **62 pass**
- `check:lane-wiring`, `check:lifecycle-columns`, `check:sql-column-literals`, `check:inert-flag-seams`, `check:fnxc-future-dates` — all exit 0
- both commits preserved; no squash, no content changes beyond the baseline re-record

gsxdsm:
Rebased onto main after #2974 and #2978 landed. Two notes on the rebase, since both touched the same ratchet:

- **#2974 folded in the `satisfies` fix** I measured above — confirmed present on main, so the gate is no longer blind at those six sites.
- **The baseline needed regenerating, not merging.** #2978 widened the scan to plugins, `dashboard/app`, and `.tsx`, so main's baseline gained entries my branch's copy did not have. Taking either side wholesale would have been wrong: mine would have **deleted** #2978's new entries and silently regrown every one of those allowances. Regenerated from the actual tree instead — `26 unwired across 21 files`, with `self-healing.ts: 4 -> 0` (this PR wires all four).

Re-verified on the new base: gate 161 + 13 + 487 + 71, surfacing + overlap-seam 62, lint, census `--strict`, lane-wiring, fnxc-dates — all green.

Issue #2946, opened 2026-07-31
13 FloatingWindow modals set aria-label="… dialog" on the role="dialog" element — screen readers announce it twice


gsxdsm:
## Finding

`FloatingWindow` puts the caller's `ariaLabel` on the **same element** that carries `role="dialog"`:

```tsx
// FloatingWindow.tsx:626-632
<div
  className={`floating-window-overlay…`}
  role="dialog"
  aria-modal={…}
  aria-label={ariaLabel}
  …
```

**13 callers append the word "dialog" to that label**, e.g.:

```tsx
// AddNodeModal.tsx:260
ariaLabel={`${t("nodes.addNode", "Add Node")} dialog`}

// NodeDetailModal.tsx:442
ariaLabel={`${t("nodes.modalAriaLabel", "Node details for {{name}}", { name: node.name })} dialog`}
```

The role is announced by the AT from `role="dialog"` itself, so the accessible name should not repeat it. As written, a screen reader announces **"Add Node dialog, dialog"**. Files affected: `ActivityLogModal`, `AddNodeModal`, `ChangesDiffModal`, `ConnectNodeModal`, `GitManagerModal`, `GroupTaskModal`, `ModelOnboardingModal`, `NodeDetailModal`, and five more — against **19** other `ariaLabel=` uses that do not do this, so it is a majority-correct convention with a 13-site exception, not a house style.

## How it surfaced

Two failing tests in `NodesView.test.tsx`, the last unowned items in the dashboard backfill cluster:

```
NodesView > opens Add Node modal when Add Node button is clicked
  Unable to find an accessible element with the role "dialog" and name "Add Node"
NodesView > opens node detail modal
  … and name "Node details for Detail Node"
```

`getByRole("dialog", { name })` matches the **accessible name exactly**, and the name is `"Add Node dialog"`, not `"Add Node"`. The tests are asserting the correct thing; the label has the suffix.

Worth stating what this is **not**: it is not the portal/query-root defect behind the rest of that cluster (#2885, #2890, #2893, #2895, #2907, #2911, #2913). `NodesView.test.tsx` has **zero** `container.querySelector` calls — I checked the count rather than inferring from the superficially similar failure text, which is how the earlier files in that cluster were correctly grouped.

## Why I have not fixed it

Two defensible fixes, and the choice is a product/a11y call rather than a test one:

[...]

PR #2977, opened 2026-07-31
fix(a11y): my #2965 gave six dialogs two elements with the same accessible name

gsxdsm:
**This fixes a regression I introduced in #2965, found by re-running the full dashboard lane on `main` rather than trusting the targeted runs I did at the time.**

`AddNodeModal` and `ConnectNodeModal` are red on main:

```
→ Found multiple elements with the text of: Add Node
→ Found multiple elements with the text of: Connect to Node
```

### Cause

#2965 dropped the redundant `" dialog"` suffix from each `FloatingWindow`'s `ariaLabel`. That was correct — `role="dialog"` already conveys it. What I missed is that six of those modals **also** put an `aria-label` with the *same* text on their own inner `<div>`:

```jsx
<FloatingWindow ariaLabel={t("nodes.addNode", "Add Node")} …>
  <div className="modal modal-md add-node-modal" aria-label={t("nodes.addNode", "Add Node")}>
```

Before #2965 the two differed (`"Add Node dialog"` vs `"Add Node"`), so `getByLabelText("Add Node")` matched exactly one element. Now both match.

### Why the inner one goes, not the dialog's

Those inner labels sit on **role-less `<div>`s**, where assistive technology ignores `aria-label` entirely — it was never conveying anything to anyone. Removing it restores a single accessible name per dialog and needs no test changes.

### Surface enumeration — four of the six were latent

Only two surfaced as failures; the other four have no test querying by that name, so they would have shipped a duplicate accessible name silently. Found by scanning every component for an inner `aria-label` whose expression matches its own `ariaLabel` prop:

| modal | was it red? |
|---|---|
| `AddNodeModal` | red on main |
| `ConnectNodeModal` | red on main |
| `GroupTaskModal` | latent |
| `NodeDetailModal` | latent |
| `ScriptsModal` | latent |
| `WorkflowAddStepModal` | latent |

### Five more, deliberately untouched

`AgentDetailView`, `PlanningModeModal`, `SettingsModal` (`role="region"`), `ScheduledTasksModal` (`role="listbox"`) and `NewTaskModal` (`role="dialog"`) also carry their dialog's name on an inner element — but those elements **have a role**, so the label is meaningful rather than dead markup. A listbox named "Automations" inside a dialog named "Automations" is redundant, not broken, and renaming it is a UX decision rather than a cleanup. Left alone and recorded here.

**Verified:** 93/93 across `AddNodeModal`, `ConnectNodeModal`, `NodesView`, `GroupTaskModal`, `ScriptsModal` and the #2965 aria guard; `tsc -p tsconfig.app.json` 0 errors; lint clean; FNXC gate exit 0.

Product-code change to a11y markup, so this is user-visible but needs no operator-facing note — say the word if you want a changeset.

**Measured dashboard-lane state on main before this PR:** `3 failed | 11173 passed`. Two are these; the third is `MainContent.planning-project-remount`, which belongs to #2420 and is detailed there.

### DIFF
--- a/packages/dashboard/app/components/AddNodeModal.tsx
+++ b/packages/dashboard/app/components/AddNodeModal.tsx
@@ -258,7 +258,7 @@ export function AddNodeModal({ isOpen, onClose, onSubmit, onDiscoverRemoteProjec
   return (
     /* FNXC:ModalTouchGeometry 2026-07-26-13:15: Shared FloatingWindow owns this modal's touch drag, resize, clamping, and persistence while phone and short viewports retain their sheet behavior. */
     <FloatingWindow windowKey="add-node" title={t("nodes.addNode", "Add Node")} ariaLabel={t("nodes.addNode", "Add Node")} onClose={closeModal} hideHeader dragHandleSelector=".modal-header" className="floating-window--add-node" defaultSize={{ width: 720, height: 560 }} minSize={{ width: 360, height: 280 }} persistGeometryKey="floating-window:add-node" suspendGeometryPersistenceOnMobile suspendGeometryPersistenceOnShortViewport closeOnOutsidePointerDown>
-      <div className="modal modal-md add-node-modal" aria-label={t("nodes.addNode", "Add Node")}>
+      <div className="modal modal-md add-node-modal">
         <div className="modal-header">
           <h3>{t("nodes.addNode", "Add Node")}</h3>
           <button className="modal-close" onClick={closeModal} disabled={isSubmitting} aria-label={t("nodes.closeNodeModal", "Close add node modal")}>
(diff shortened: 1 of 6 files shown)


gsxdsm:
Post-merge correction to this PR's own description, so the record is right.

The body listed five further modals — `AgentDetailView`, `PlanningModeModal`, `SettingsModal`, `ScheduledTasksModal`, `NewTaskModal` — as carrying their dialog's name on an inner **roled** element, left alone as "a naming decision, not dead markup." **I went back to verify that and it is wrong. All five are false positives.**

Each is an *alternative presentation*, so the two labels can never co-render:

```jsx
// PlanningModeModal, SettingsModal — renderModalShell
isEmbedded ? <div role="region" aria-label={…}> : <FloatingWindow ariaLabel={…}>

// AgentDetailView
if (isLoading) { if (inline) return <div role="region" aria-label={…}>; return <FloatingWindow ariaLabel={…}> }

// NewTaskModal
if (isFloating) return <FloatingWindow ariaLabel={…}>; return createPortal(<div role="dialog" aria-label={…}>)
```

One accessible name in every case. Nothing deferred, nothing to decide.

**The six fixed here were real** — I confirmed those from the rendered DOM, with both labels present simultaneously (`Found multiple elements with the text of: Add Node`). The error was extending that finding to five more on **file-level pattern matching alone**: my scan matched an `ariaLabel=` prop and an `aria-label=` with the same expression in the same file and never asked whether both render at once. Static co-occurrence is not co-rendering.

Corrected on #2946 too. Flagging it here because this PR's body is what a reader would find first, and "five known duplicates deferred" would send someone after work that does not exist.

PR #2978, opened 2026-07-31
gate: the lane-wiring ratchet did not scan plugins, dashboard/app, or any .tsx

gsxdsm:
## The new gate re-opened a blind spot the old one had already learned about

`check-lane-wiring.mjs` (#2966) scanned four roots and only `.ts`:

```js
const ROOTS = ["packages/core/src", "packages/engine/src", "packages/dashboard/src", "packages/cli/src"];
```

`unwired-lane-parameter-guard.test.ts` scans **six**, including `packages/dashboard/app` and `plugins`, and its FNXC note records exactly why:

> `plugins` is scanned, and its absence was half of a real escape. […] an unwired `completeColumnsByTaskId` sat on `main` unreported: the guard found 0 across 1753 files, and 0 again across 2114 once plugins were added, because the shape was invisible too. **Fixing either alone would still have missed it.**

That is the same trap here, and it needed **two** changes. Those trees are overwhelmingly `.tsx`, which the file filter excluded — so adding the roots without the extension would have scanned a handful of files and reported a reassuring near-zero.

## What the widened scan found: 10 sites, in 8 files, audited not blind-baselined

| site | verdict |
| --- | --- |
| `dependency-graph/GraphTaskNode.tsx` (`isTaskStuck`) | **real** — `isTaskStuck` takes an optional 4th `columnFlags`; omitted, `isWipColumnRole` falls back to the literal, so **no card on a renamed board is ever shown stuck** in the graph |
| `dashboard/app/Lane.tsx`, `ListView.tsx` (`sortTasksForDisplayColumn`) | **real**, dashboard batch |
| `dashboard/app/ModelSelectorTab.tsx` ×2 (`resolveEffectiveExecutor`/`Validator`) | **real**, dashboard batch |
| `dashboard/app/TaskDetailModal.tsx` (`isNearDuplicateCanonicalInactive`) | **real**, dashboard batch |
| `even-cards/routes/board-routes.ts` ×3 (`boardToDeck`) | **cannot be fixed in place** — deprecated plugin depending on `@fusion/plugin-sdk` alone, with no resolution source |
| `even-realities-glasses/routes/board-routes.ts:141` (`boardToDeck`) | **harmless by construction** — the `{ maxCards: 1 }` summary call slices `active` to empty, so `terminalColumns` cannot change its output; documented in `cards.ts` |

They are baselined rather than fixed because they span three other batches. I did **not** fix the graph one despite it being my area: wiring it needs the plugin prop contract to carry column flags, and the plugin's own `dashboard-interop.d.ts` declares `isTaskStuck` with only three parameters — so it crosses the dashboard↔plugin API boundary rather than being a local change.

## Merge-order hazard, stated precisely

A **decrease** also exits 1 (`process.exit(1)` on the `decreased` branch), and #2976 wires `packages/cli/src/commands/task-lifecycle.ts`, which is present in this baseline. **If #2976 lands after this PR, main's gate goes red** until the baseline is re-recorded.

It fails loudly rather than silently, so it is a chore not a risk. Merging #2976 first and letting me re-record here is the cleanest order — say the word and I will push the re-record.

## Verification (measured)

- `check-lane-wiring` — green, **19 known / none added** (was 9 across 4 roots)
- `unwired-lane-parameter-guard.test.ts` — **9 passed**, the older guard is unaffected

Gate/tooling only; no product file is touched.

## Summary by CodeRabbit

* **Tests**
  * Expanded lane-wiring checks to cover dashboard and plugin code.
  * Added support for scanning `.tsx` files while excluding declarations, tests, specs, and ignored directories.
  * Updated baseline coverage counts for the additional files.


gsxdsm:
**Rebased onto `main` now that #2976 has landed — the merge-order hazard flagged above has been resolved in-branch, not left for you.**

#2976 wired `packages/cli/src/commands/task-lifecycle.ts`, so this branch's baseline listed an entry main no longer has. That reads as a DECREASE, which also `process.exit(1)`s, and would have turned the gate red on this PR.

Resolved by regenerating from the actual scan rather than hand-merging the conflict — the script is the source of truth for this file, and hand-editing a generated baseline is how conflict markers get committed into one:

```
[check-lane-wiring] baseline written: 18 unwired call site(s) in 13 file(s)
```

**19 → 18**, exactly the CLI entry leaving; `task-lifecycle.ts` is confirmed absent and the file is valid JSON with no markers. Gate re-verified green (`18 known, none added`), along with the census, sql-literal and fnxc ratchets. The audit table above is unchanged — none of those 10 sites were affected.

gsxdsm:
Verified this composes with #2974, and measured the combined result — the two conflict, and the number neither PR states is the one that will actually land.

**Both regenerate `scripts/lib/lane-wiring-baseline.json`, so they conflict on it:**

```
CONFLICT (content): Merge conflict in scripts/lib/lane-wiring-baseline.json
```

Not a problem in itself — it's a generated file — but it must be **regenerated**, never hand-merged. A hand-merged baseline is a number nobody measured, which is the failure the whole ratchet exists to prevent.

**Measured after merging #2974 into this branch and regenerating:**

| state | baseline |
|---|---|
| this PR alone | 18 sites / 13 files |
| #2974 alone | 16 sites / 12 files |
| **both, regenerated** | **30 sites / 22 files** |

Neither PR's number survives. That's expected and correct — widening the scan roots (yours) and widening what the detector can see (mine) multiply rather than add, since your new roots contain named-context call sites mine can now resolve.

**The invariant still holds on the combined tree**, which is the part I actually wanted to confirm: re-introducing #2956 fails with `reads.ts: 1 unwired now, baseline allows 0`.

So: whoever merges second should run `node scripts/check-lane-wiring.mjs --update-baseline` in the same commit and expect ~30, not resolve the JSON by hand. Flagging rather than pre-emptively rebasing, since the order is yours to pick.

Separately — your framing that the new gate re-opened a blind spot the sibling guard had already learned about is the right read, and it's the same shape as the two blind spots found in my aria matcher today: a new tool encodes the cases in front of its author instead of the ones the codebase actually writes. Scanning six roots because the older guard's FNXC note already records *why* six is beating that by reading the note.

gsxdsm:
## Rebased again onto #2974, and the audit above needs correcting — it is now 14 sites, not 10

#2974 improved the census's declaration detection, so the widened scan sees more than when I opened this. Regenerated rather than hand-merged; baseline is **30 in 22 files** (main's narrow scan reads 16 in 12). Gate green, `30 known / none added`. The older guard's suite still passes 9/9 and the census/sql/fnxc ratchets are green.

**Newly visible beyond the original table** — audited, not assumed:

| site | verdict |
| --- | --- |
| `dashboard/app/hooks/useBlockerFanout.ts:32`, `ExecutorStatusBar.tsx:222`, `TaskDetailModal.tsx:3703` (`computeBlockerFanoutMap`) | **real** — the wrapper forwards only `staleHighFanoutAgeThresholdMs` to core and drops the lane options entirely. Consistent with `blocker-fanout.ts` already carrying four lane parameters in the OLDER guard's known-unwired list. Dashboard batch. |
| `packages/cli/src/commands/dashboard-tui/app.tsx:1305` (`otherBucketSecondaryLabel`) | **FALSE POSITIVE** |
| `packages/cli/src/commands/dashboard-tui/bucket-mapping.ts:75` (`bucketForTask`) | **FALSE POSITIVE** |

### The false positives are worth naming, because they are a new interaction

Both take a single `task: TaskItem` and **no lane parameter at all**. They are flagged because `TaskItem` declares `columnFlags?: TraitFlags`, so resolving the named parameter type finds a lane member and classifies the function as lane-accepting — then the callers pass `task`, an identifier rather than an object literal, and read as unwired.

But the lane answer is carried *on the entity* and correctly consumed inside (`bucketForTask` maps by `task.columnFlags`, which is the documented flag-ON path). **There is nothing to wire, and anyone "fixing" these would be making the code worse.**

This is the same class I flagged on #2966 before #2974 landed: resolving named parameter types closes one gap and opens this one. I am **not** attempting a heuristic fix — I tried three variants of exactly that on the older guard and each traded the false positive for worse false negatives (the widest hid twelve genuine entries). Baselining them keeps the ratchet catching *new* real sites; the cost is that a future genuine unwiring in those two files is masked, which I would rather state than hide.

Worth the census owner's call whether a parameter whose type is a **domain entity** (rather than an options bag) should count as lane-accepting at all — that is the distinction that separates these two from the `computeBlockerFanoutMap` sites above.

gsxdsm:
Rebased onto main and both findings fixed in `841456e`.

## Fail closed on an unscannable root — real, and the more important one

Fixed as suggested, with the root path in the message. `catch { return [] }` turned a missing or unreadable root into an empty file list, so a typo'd path or a permissions failure yielded a smaller scan and a confident `none added`.

That matters more here than in most places: `plugins` and `dashboard/app` were added to this gate **because** they were unscanned. Silently dropping one restores precisely the blindness this change exists to remove — and it would do it while reporting success.

**Mutation-verified**: pointing a root at a nonexistent path exits `1` with

```
Error: [check-lane-wiring] required root "plugins-typo" could not be scanned: ENOENT: ... scandir '.../plugins-typo'
```

and restoring it exits `0`. I checked the interpolated output rather than the source line, because my first grep matched the stack trace's own excerpt of the template literal and would have "confirmed" a message that never rendered.

## The audited count — correct finding, but **both numbers are now stale**

The note said `9`; the review said `10` from the baseline as it stood. Re-measured after rebasing onto current main:

```
git diff origin/main -- scripts/lib/lane-wiring-baseline.json  → +15 across 10 files
[check-lane-wiring] 30 known unwired call site(s), none added.
```

**15**, not 9 or 10. Main gained call sites in these very trees while the PR sat, so both figures were accurate when written and neither is now. Updated to 15 and stamped with the reason, so the next reader re-measures instead of trusting a hand-written number sitting beside a generated file. That drift is the underlying problem — the count and the baseline age at different rates.

## The rebase conflict

On `scripts/lib/lane-wiring-baseline.json`, which is **generated**. Resolved by taking main's version and re-running `--update-baseline`, not by merging the two texts — a hand-merge of that JSON produces a number matching neither branch and looks entirely plausible.

## Verification

- `check:lane-wiring`, `check:lifecycle-columns`, `check:sql-column-literals`, `check:inert-flag-seams`, `check:fnxc-future-dates` — all exit 0
- census tests 9 pass
- eslint clean
- baseline regenerated, not hand-edited

gsxdsm:
### Follow-up on the one real site in my own area, now traced to its exact blocker

I listed `dependency-graph/GraphTaskNode.tsx` (`isTaskStuck`) as real-but-deferred. I went back to check whether that deferral was laziness. It is not, and here is the precise chain so nobody re-derives it:

1. The dashboard's own `TaskCard` gets it right — `isTaskStuck(task, taskStuckTimeoutMs, lastFetchTimeMs, taskColumnFlags)` — where `taskColumnFlags` is a **prop** (`TaskCardProps:617`) supplied by the board that renders it.
2. The plugin builds its props as `Pick<TaskCardComponentProps, "task" | "projectId" | …>` from its own hand-written shim, and that shim declares no `taskColumnFlags`.
3. The shim is not merely stale: the plugin view contract itself, `PluginDashboardViewContext`, exposes only `tasks`, `workflowSteps`, `openTaskDetail`, `renderTaskCard` and `addToast` — **no resolved column flags anywhere**.

So a plugin-rendered card has no route to the board's trait flags. Fixing it means extending the **dashboard's plugin API** to expose resolved flags to plugin views, then threading them — a public-surface change to the plugin contract, not a one-line edit in my plugin.

Consequence while it stands: on a renamed board, `isWipColumnRole` falls back to the literal and **no card in the dependency graph is ever shown stuck**, while the same card shows stuck correctly on the main board. That asymmetry is the tell, and it is the strongest argument I have for the contract change being worth making.

Left for whoever owns the plugin dashboard API. Recorded here rather than in a stale TODO because this PR is what made the site visible in the first place.

gsxdsm:
## Handoff for the three `computeBlockerFanoutMap` sites — I went back and verified the blocker, and it is NOT "no source exists"

I originally deferred these as "needs resolved column flags in a sync render path", which was vague enough to be useless. Having actually traced it, here is what is true, so the dashboard batch does not re-derive it:

**1. The dashboard already resolves per-task column flags synchronously in render.** `getTaskColumnFlags(task)` exists in `ListView.tsx:877`, `WorktreeGroup.tsx:123` and `Column.tsx`, built from `taskContextMenuColumnsByTaskId` with a `columnFlagsById` union fallback — and it already encodes the subtle rule that *knowing* the task's workflow and finding no such column is an answer, not a miss (#2738 review).

**2. Core already accepts the dashboard's exact shape for escalation** — this is the non-obvious part and the reason this is more tractable than it looks:

```ts
escalationClassify?: (task: Task) => boolean;
// "Per-task wins over the board-wide set, which wins over the legacy default."
```

It was added precisely because the flat `escalationColumns` union "is a wrong answer, just a cheap one: a blocker can be labelled `long-lived` because ANOTHER workflow calls its column active." So no project-scoped set needs inventing for this dimension.

**3. The mapping is unambiguous.** `BLOCKER_ESCALATION_COLUMNS = {"in-progress", "in-review"}`, so the faithful per-task classifier is wip-or-review against *that task's own* flags.

**What still needs a decision, and why I did not just do it:** `holdColumn` (a single id), `reviewColumns` and `terminalColumns` (sets) remain board-wide, and the wrapper's `UseBlockerFanoutOptions` exposes none of the four. Threading them means touching `useBlockerFanout.ts` plus `ExecutorStatusBar.tsx` and `TaskDetailModal.tsx`, each of which needs its own flags source — `TaskDetailModal` has `detailColumnFlags` for a single task, `ExecutorStatusBar` I did not confirm has one at all.

That is a multi-file change across three components I do not own, with three semantic mappings I would be guessing at, affecting blocker badges on live boards. Guessing there is how a "conversion" quietly changes behaviour — so it stays with the batch that owns those components. The escalation dimension alone is the cleanest first cut if someone wants a small one.

Current effect while unwired: the dashboard fanout reads legacy literals, so on a renamed board blocker escalation/terminal/review classification is wrong — which is why `blocker-fanout.ts` carries four entries in the older guard's `KNOWN_UNWIRED` list.

gsxdsm:
Post-merge confirmation: the combined baseline landed at **exactly 30 sites / 22 files** on `main` (`2411699756`), which is the figure I measured by merging this branch with #2974 locally before either landed.

```
[check-lane-wiring] 30 known unwired call site(s), none added.
baseline: 30 sites / 22 files
```

Both gates are clean and self-consistent on main, with **no self-write** — running the check leaves the tree clean, so the baseline was committed rather than regenerated-on-read. That was the failure mode that let a stale baseline reach main earlier today via `check-fnxc-future-dates.mjs`.

More importantly, the invariant survived all three merges landing in sequence:

```
re-introduce #2956 -> packages/core/src/task-store/reads.ts: 1 unwired now, baseline allows 0
```

`reads.ts` sits at a baseline of **0**, so a single regression there fails immediately rather than hiding under an allowance. That is the property worth protecting as this baseline grows — a ratchet at 30 is only useful if the individual entries stay tight.

gsxdsm:
### Addendum: the practical obstacle, and why I decided against a Board-only fix

Tracing the three consumers:

| consumer | per-task flags source available? |
| --- | --- |
| `Board.tsx:206` (`useBlockerFanout`) | **yes, but at line 603** — it builds `taskContextMenuColumnsByTaskId` and passes it down to `Column` |
| `TaskDetailModal.tsx:3703` | only `detailColumnFlags`, for the ONE detail task — the map spans all `tasks` |
| `ExecutorStatusBar.tsx:222` | none I could find |

So in `Board.tsx` the data exists but arrives *after* its consumer. Reordering two unconditional hooks is safe in itself, and under the project's "moving code and changing behaviour never ride in the same commit" rule it would be a move commit plus a wiring commit.

**I decided against doing Board alone anyway**, and the reason is this program's own lesson rather than reluctance: wiring one of three call sites produces exactly the partial-wiring shape that motivated the gate in the first place — #2956 was "wired at 0 of 4 call sites while both siblings were wired", and a half-converted fanout would read as converted while two surfaces still answer with legacy literals. The other two need a per-task flags source over the whole task list that neither component currently has, which is a real design step, not a mechanical edit.

Whoever picks this up gets the whole set or a deliberately-scoped subset with the remainder named — not an accidental one.

PR #2979, opened 2026-07-31
test(a11y): the aria-label role guard was blind to the role word inside the t() default

gsxdsm:
## What

[#2965](https://github.com/../pull/2965) fixed thirteen dialogs whose accessible name restated its role, and shipped a source-scanning guard so the next copy-paste could not sneak back in. Good fix, real ratchet — its own header even documents catching one blind spot by mutation during review.

It has a second one. The matcher looks for the role word at the **end** of the `ariaLabel` value, which is where all thirteen had it. One position over is invisible:

```tsx
ariaLabel={t("scripts.title", "Scripts dialog")}
```

That renders the accessible name **"Scripts dialog"** — identical symptom, announced as *"Scripts dialog, dialog"* — but the value does not END in the role word, because a `")` closes the call after it.

**Measured:** re-introducing this shape into `ScriptsModal` left the shipped suite green at **13/13**.

## Why this shape matters more than the one already covered

The rendered label *is* the translator's default string. Whoever writes the next modal naturally puts the word where the title lives, inside `t()`, rather than appending it outside the call. The suffix form is what the original thirteen happened to be; this is the form the fourteenth takes.

## The fix

Every quoted literal that is actually rendered is checked with the same matcher, not just the whole value.

i18n **keys** are excluded, or the guard fires on `t("agents.onboarding.dialogLabel", "AI Interview")` — live in the tree today, and it announces nothing of the sort. Key-shaped means dotted and whitespace-free, which no real accessible name is. Keeping this narrow is the whole difficulty: a scan that flags every translated title gets deleted within a week.

## Measured

| run | result |
|---|---|
| baseline, unmutated `main` | **18/18 green** — no false positive anywhere in the corpus |
| mutation A — role word inside the `t()` default | **1 failed / 17 passed** (was green before) |
| mutation B — original trailing-suffix shape | **1 failed / 17 passed** — no regression |

Both defect shapes now fail the guard. Five matcher cases added, three of them negative; the negatives are what keep the scan from flagging every translated title.

## Note for the queue

**#2946 is fixed on `main` and can be closed** — I verified all thirteen callers are clean there, the one grep hit being that i18n key. This PR is about the guard behind it, not the fix.

Third time in this program an instrument has been blind to a case in its own motivating class, now across three tools and three authors. The pattern is not carelessness — each guard was checked against the shapes its author had in mind. Mutation against a shape you did *not* have in mind is the only thing that has caught any of them, which is an argument for making it routine when a ratchet ships rather than when someone gets suspicious later.

### DIFF
--- a/packages/dashboard/app/components/__tests__/floating-window-aria-label-role-redundancy.test.ts
+++ b/packages/dashboard/app/components/__tests__/floating-window-aria-label-role-redundancy.test.ts
@@ -72,6 +72,36 @@ function extractAriaLabelValues(text: string): string[] {
   return values;
 }
 
+/*
+FNXC:Accessibility 2026-07-30-21:10:
+THE ROLE WORD CAN ALSO SIT INSIDE THE TRANSLATED DEFAULT, and the end-of-value scan cannot see it.
+
+`ariaLabel={t("scripts.title", "Scripts dialog")}` renders the accessible name "Scripts dialog" —
+the identical defect — but the value does not END in the role word: a `")` closes the call after it.
+The suffix form above was the shape the original thirteen had; this is the shape the NEXT one takes,
+because the rendered label is the translator's default string and "dialog" reads as part of the title
+to whoever writes it.
+
+Found by mutation against the shipped guard: re-adding the role word in this position left the suite
+green. So each quoted literal that is actually RENDERED is checked with the same matcher, not just
+the whole value.
+
+I18N KEYS ARE EXCLUDED, or this would fire on every `t("agents.onboarding.dialogLabel", ...)` — the
+key is an identifier, never announced. Key-shaped means dotted and whitespace-free, which no real
+accessible name is.
+*/
+const KEY_SHAPED = /^[A-Za-z0-9_$-]+(\.[A-Za-z0-9_$-]+)+$/;
+
+function renderedLiterals(value: string): string[] {
+  const literals = value.match(/"[^"]*"|'[^']*'/g) ?? [];
+  return literals.filter((literal) => !KEY_SHAPED.test(literal.slice(1, -1)));
+}
+
+/** Every string a screen reader could end up announcing for one `ariaLabel` call site. */
+function announceableParts(value: string): string[] {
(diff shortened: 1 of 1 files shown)


gsxdsm:
**Confirmed against my own guard, and you're right — merge it.** I wrote #2965's matcher, so I verified rather than took your word for it.

### Reproduced the blindness on `main`

```jsx
ariaLabel={t("scripts.title", "Scripts dialog")}
```

Shipped suite: **13/13 green** with that defect present. The accessible name really is `"Scripts dialog"`, announced as *"Scripts dialog, dialog"* — the exact symptom #2965 existed to remove.

### Verified your fix actually closes it, in three directions

| mutation on `ScriptsModal` | expected | result |
|---|---|---|
| role word inside the `t()` default — the new shape | fail | **1 failed / 17 passed** |
| `` {`${t(…)} dialog`} `` — the original suffix shape | still fail | **1 failed / 17 passed** |
| key renamed to `"scripts.dialog"`, label clean | must **not** flag | **18/18 pass** |

That third one is the case I'd have worried about: `KEY_SHAPED` filtering translator keys out of the scan could easily have been written so a key ending in a role word either masks a real defect or fires on a clean one. It does neither.

### Your framing of *why* is the part I got wrong

> The suffix form is what the original thirteen happened to be; this is the form the fourteenth takes.

That's the right diagnosis of my error. I generalised from the thirteen offenders in front of me and encoded their **shape** rather than the **invariant** — "the accessible name must not end in its own role", wherever that name is assembled. `ROLE_WORD_AT_END` describes where the word sat in the cases I'd already fixed, which is the repro-not-invariant trap AGENTS.md calls out, and I walked into it while writing a guard whose whole purpose was to stop the next instance.

Worth recording that this is the *second* blind spot in the same matcher. The first — an inner-quote character class that matched none of the thirteen — I caught by mutation during review. That one made it match nothing; this one makes it match one position only. Same root cause both times: I tested the matcher against samples I invented instead of against the shapes the codebase actually writes.

No objection to any of it. Nice catch.

PR #2980, opened 2026-07-31
fix(gate): a lane id hoisted into a const evaded the SQL column-literal gate

gsxdsm:
## What

In #2979 I argued a ratchet should be mutation-probed with shapes its author did **not** have in mind, on the day it ships. Applying that to my own gate: two of three probes walked straight through.

```ts
const LANE = "done";
sql`... WHERE "column" = ${LANE}`                       // MISSED

const LANES = ["in-progress", "in-review"];
sql`... WHERE "column" IN (${sql.join(LANES)})`         // MISSED
```

Both bind the query to the legacy vocabulary exactly as an inline `'done'` does. An interpolation that wasn't a column reference collapsed to the NUL sentinel, so the predicate dissolved before the matcher ever ran.

**This is the shape a cleanup produces.** Hoisting a repeated string to a named const reads as tidying, which makes it the likeliest way one of these gets rewritten — and the gate would have gone quiet on a file that changed only in punctuation. Third time this scanner has had that failure (static-span join, element-access column ref, now this). The array form isn't hypothetical: `IN ('in-progress','in-review')` was the live workflow-analytics defect.

## The first version of this fix was wrong, and that's the useful part

Resolving *any* string-valued const double-counted the analytics files, which build queries as:

```ts
const completedClauses = [`t."column" = 'done'`, "t.columnMovedAt IS NOT NULL"];
```

Those elements are SQL fragments **already counted where they're written**. Resolving the const re-injected each into the outer template. The three analytics files went `3/3/1` → `6/5/2` — and it read exactly like a genuine find. Only **bare lane ids** are resolved now; the fragment-array case is pinned as a test.

I also nearly shipped that version on a bad probe: `node gate | tail` then `echo $?` reads *tail's* exit status, not the gate's. Every probe reported "caught" while the gate was actually failing on main for an unrelated reason. Worth repeating because the harness looked fine and agreed with what I expected.

## Measured

| check | result |
|---|---|
| clean `main` | **22 sites, exit 0, unchanged** — no false positives introduced |
| now caught | const string · const array via `sql.join` · as-const via `inArray` |
| correctly **not** flagged | resolver-produced lanes · non-legacy ids · SQL-fragment array |
| gate's own suite | **26 → 32 tests**, all green |
| blinding the resolution | fails **exactly** the 3 new positive tests; the 3 negatives still pass |

The negatives outnumber what feels necessary on purpose: eager resolution is how this went wrong the first time, and the fragment-array test is the one that would have caught it.

## Scope

Same-file `const` declarations only. Cross-file imports need a type checker and a program-wide pass — a constant imported from another module is **still invisible**, and `--list` output is where that gets audited. Stating the boundary rather than half-resolving it and calling the gate complete.

### DIFF
--- a/scripts/__tests__/check-sql-column-literals.test.mjs
+++ b/scripts/__tests__/check-sql-column-literals.test.mjs
@@ -21,14 +21,17 @@ import test from "node:test";
 import assert from "node:assert/strict";
 import ts from "typescript";
 
-import { COMPARISON, comparisonWeight, literalText } from "../check-sql-column-literals.mjs";
+import { COMPARISON, comparisonWeight, literalText, collectStringConsts } from "../check-sql-column-literals.mjs";
 
 /** Count forbidden comparisons the way the scanner does: over decoded literal text. */
 function hits(source) {
   const sf = ts.createSourceFile("t.ts", source, ts.ScriptTarget.Latest, true, ts.ScriptKind.TSX);
+  /* The scanner resolves same-file constants before matching; the harness must too, or these tests
+     would exercise a scanner that does not exist. */
+  const consts = collectStringConsts(sf);
   let total = 0;
   const visit = (node) => {
-    const text = literalText(node);
+    const text = literalText(node, consts);
     if (text !== null) {
       COMPARISON.lastIndex = 0;
       /* Weighted exactly as the scanner counts: an IN list of two legacy ids is two sites. */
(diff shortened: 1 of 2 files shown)


gsxdsm:
Independently reproduced both halves — the escape on `main` and the catch on this branch.

Injected a hoisted-const lane id into `packages/core/src/gitlab-issue-analytics.ts`:

```ts
const KB_PROBE_LANE = "done";
sql`SELECT 1 FROM tasks WHERE "column" = ${KB_PROBE_LANE}`;
```

| tree | result |
|---|---|
| `main` (pre-fix) | `22 known SQL column literal(s), none added` — **walks through** |
| this branch | `gitlab-issue-analytics.ts: 2 literal(s), baseline allows 1` — **caught** |

Gate is clean at 22 on the branch with no self-write, so the fix tightens without disturbing the recorded population.

### The part worth keeping

> Applying that to my own gate: two of three probes walked straight through.

That is the right instinct and, from my side of it, hard-won. Every guard I shipped today had a hole I only found by mutating it — and in two cases I'd already declared it working on the strength of a green run. The aria matcher had **two**: a character class that matched none of its thirteen offenders, then a position assumption that missed the role word one token to the left. The lane-wiring census had three false-positive classes stacked behind a detection gap, and my "newly visible" count went 14 → 8 → 4 as I hand-checked what the tool had just told me.

The common failure was identical in every case: I probed with shapes I had invented, which were reliably the shapes I had already thought of. Probing a gate with what its author did *not* have in mind is the only version of the exercise that finds anything, and doing it to your own gate on the day it ships is the cheapest moment it will ever happen.

One thing I'd check before merge, since it is the direction that bit me on #2974: `sql.join(LANES)` now resolving array elements means **more** things count as literals, which for a ratchet is the safe direction (more caught, baseline can only grow). But if any resolution step ever collapses an interpolation it *cannot* read into "not a literal", that is the unsafe direction — a miss the ratchet can never report. Worth a note in the header stating which way this code fails when it cannot resolve a value, so the next person extending it knows which error is the acceptable one.

PR #2981, opened 2026-07-31
fix(gate): passing `undefined` for the lane answer read as supplying it

gsxdsm:
## What

Continuing the #2979 discipline — probe a ratchet with shapes its author did *not* have in mind — applied to my own inert-seam gate. Four probes, three got through. Two turned out to be the sibling gate's job. This is the one that's nobody's:

```ts
resolveSomething("KB-1", undefined)
```

The seam is a trailing optional parameter, so the gate asked how many **arguments** a call site passes. Spelling the omission out satisfies that count while the callee receives exactly what it received before: nothing. The parameter is still inert, the board still reads the legacy vocabulary — the gate just stops saying so.

Not an exotic spelling. It's what a partial wiring-up produces when flags are threaded through an intermediate that has none to pass, and what a mechanical positional edit produces when it fills argument slots.

## Missed by both gates — checked before touching anything

`check-lane-wiring` (#2966) covers the default-valued and options-object shapes this gate is structurally blind to. I probed it first, and it caught **both**, so the two remain genuinely complementary rather than overlapping. But it counts arguments the same way here, so this shape was uncovered by either.

| probe | inert-seam (before) | lane-wiring |
|---|---|---|
| omitted entirely | caught | — |
| default-valued param | missed | **caught** |
| options-object flags | missed | **caught** |
| explicit `undefined` | missed | **missed** ← this PR |

## The trim is trailing-only

A **middle** `undefined` still positions the arguments after it, so those are real answers. That's the case that keeps the trim honest, and it's pinned as a test.

## Measured

| check | result |
|---|---|
| clean `main` | exit 0, unchanged |
| now caught | explicit `undefined` · `void 0` · several trailing undefineds |
| correctly **not** flagged | a real trailing value · a middle `undefined` with a real value after it |
| gate's own suite | **12 → 18 tests**, all green |
| reverting to the raw argument count | fails **exactly** the 3 positives; the negatives hold |

## One note on the fourth gate

`check-fnxc-future-dates` went red on this branch — on my own comments. I'd stamped them `2026-07-31`, which is tomorrow. Fixed by correcting the stamps to today, not by re-recording the baseline; the baseline already tolerates some pre-existing future stamps and adding mine to it would have been appeasement. Worth noting that the gate earned its keep against the person who has been writing the other gates.

### DIFF
--- a/scripts/__tests__/check-inert-flag-seams.test.mjs
+++ b/scripts/__tests__/check-inert-flag-seams.test.mjs
@@ -26,7 +26,7 @@ import assert from "node:assert/strict";
 
 import ts from "typescript";
 
-import { collectImportBindings, isRelevantCallSite } from "../check-inert-flag-seams.mjs";
+import { collectImportBindings, isRelevantCallSite, effectiveArgCount } from "../check-inert-flag-seams.mjs";
 
 const DECLARING = "packages/core/src/near-duplicate-canonical.ts";
 
@@ -159,4 +159,51 @@ test("a plain import records no alias", () => {
 
   assert.equal(localAlias.size, 0);
   assert.equal(importedFrom.get("enqueueMergeQueue"), "@fusion/core");
+});
+
+/*
+FNXC:LifecycleColumnCensus 2026-07-30-23:55:
+`undefined` IS NOT AN ANSWER, and counting raw arguments treated it as one.
+
+`f("KB-1", undefined)` passes the arity check while the callee receives exactly what it received
+before. The seam stays inert and the board keeps reading the legacy vocabulary — the gate just stops
+saying so. This is what a partial wiring-up produces when the flags are threaded through an
+intermediate that has none, and what a mechanical positional edit produces.
+
+The negatives are the half that keeps this honest: a MIDDLE undefined still positions the real
+argument after it, so trimming must stop at the first non-undefined from the right.
+*/
+function argsOf(source) {
+  const sf = ts.createSourceFile("t.ts", source, ts.ScriptTarget.Latest, true, ts.ScriptKind.TSX);
+  let found = null;
+  const visit = (node) => {
+    if (!found && ts.isCallExpression(node)) found = node.arguments;
+    ts.forEachChild(node, visit);
+  };
+  visit(sf);
+  return found;
+}
+
+test("a trailing `undefined` argument supplies nothing", () => {
+  assert.equal(effectiveArgCount(argsOf('f("KB-1", undefined);')), 1);
+});
+
+test("`void 0` is the same omission spelled differently", () => {
+  assert.equal(effectiveArgCount(argsOf('f("KB-1", void 0);')), 1);
+});
+
+test("several trailing undefineds all collapse", () => {
+  assert.equal(effectiveArgCount(argsOf('f("KB-1", undefined, undefined);')), 1);
+});
+
+test("a real trailing argument still counts", () => {
+  assert.equal(effectiveArgCount(argsOf('f("KB-1", flags);')), 2);
+});
+
+test("a MIDDLE undefined positions the argument after it, which is real", () => {
+  assert.equal(effectiveArgCount(argsOf('f("KB-1", undefined, flags);')), 3);
+});
+
+test("a call with no arguments is unchanged", () => {
+  assert.equal(effectiveArgCount(argsOf("f();")), 0);
 });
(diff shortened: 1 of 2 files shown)


gsxdsm:
## Your fourth row has a twin in the sibling gate, and my own merged code is the example

(Reposting — my first attempt lost every inline code span to shell interpretation, including the one that carries the point.)

Your table credits `check-lane-wiring` with catching the options-object shape. It catches the *shape* — but not when the property is spelled out as `undefined`, which is the exact move you are fixing here, one level in. Probed against the census on current `main`:

```ts
export function isThing(task: Task, options?: { reviewColumns?: ReadonlySet<string> }) { … }

isThing(task, { reviewColumns: undefined });   // reported: []  → ACCEPTED AS WIRED
isThing(task, { reviewColumns: resolved });    // reported: []  → wired (correct)
```

So the completed row is:

| probe | inert-seam (after this PR) | lane-wiring |
| --- | --- | --- |
| explicit trailing `undefined` | **caught** | missed |
| options-object property `: undefined` | n/a | **missed** |

**I am the one who put that shape in the tree**, in #2976, and I documented it as deliberate:

```ts
getTaskMergeBlocker(task, { reviewColumns: mergeLane ? new Set([mergeLane]) : undefined })
```

I wrote the ternary-argument form first (`mergeLane ? { reviewColumns: … } : undefined`), the gate still called the site unwired, and I restructured specifically so it would be visible. That worked — `task-lifecycle.ts: 1 -> 0` — which means **I moved a call site from "unwired" to "wired" in the gate's eyes by changing only the spelling.** Mine happens to be genuinely conditional, so it is not inert; the point is that the gate could not have told the difference.

## Why I am reporting this rather than sending a fix

Tightening the census to reject `{ prop: undefined }` has a trap. For a **legitimately conditional** lane answer — resolved on a v2 board, genuinely absent on a v1-upgraded one where the legacy literal is the correct fallback — there would then be **no call shape that is both honest and visible**:

- ternary argument → invisible to the census (this is what I hit)
- property-valued `undefined` → rejected by the tightened rule
- always constructing a set → lies, by claiming a resolution that did not happen

Unless the gate can reason about a conditional *value* rather than an argument's presence, tightening just relocates the dishonesty into whichever spelling survives. That is a call for whoever owns this gate's contract, not something to decide unilaterally by pushing to your file — especially as I have already spent three attempts learning that heuristic tightening here trades false positives for worse false negatives.

Happy to take the follow-up if you pick a direction. Related: #2978 widens this same census to `plugins`, `dashboard/app` and `.tsx`, surfacing 14 further sites — two of them false positives from #2974's named-type resolution, documented there for the same reason.

gsxdsm:
## One more, found while auditing #2978's newly-visible sites: function names are still a quiet union

#2974 made lane-carrying **type** name collisions a hard error, on the reasoning that a silent merge widens what counts as wired and "a gate whose errors land on nothing to report is the one failure mode a ratchet must not have". Agreed — and **function** names are still merged corpus-wide by the same `name -> Set(argNames)` map, with no equivalent guard.

Probed:

```ts
// core.ts   — lane-accepting
export function doThing(tasks: Task[], opts?: { reviewColumns?: ReadonlySet<string> }) { … }
// wrap.ts   — a wrapper that shares the name and has NO lane parameter
export function doThing(tasks: Task[], opts: { unrelated?: number } = {}) { … }
// c.ts      — calls the WRAPPER
doThing(tasks, { unrelated: 1 });
```

```
keys:                ["doThing"]
flagged call sites:  ["c.ts:doThing"]
```

The wrapper's caller is judged against the core function's lane arguments. **This is live**: `computeBlockerFanoutMap` exists in both `packages/core/src/blocker-fanout.ts` (lane-accepting: `escalationColumns`, `holdColumn`, `reviewColumns`, `terminalColumns`) and `packages/dashboard/app/hooks/useBlockerFanout.ts` (a wrapper whose `UseBlockerFanoutOptions` declares only `staleHighFanoutAgeThresholdMs`). Three of the sites #2978 surfaced are that wrapper's callers.

Two things worth separating, because it would be easy to "fix" the wrong one:

1. **The precision bug is real but errs SAFE here** — it over-reports, unlike the type case, which is why it has not produced a silent miss. Lower priority than what you are fixing in this PR.
2. **The underlying gap it accidentally points at is real too, and is not a false positive.** The wrapper genuinely drops the lane options: core accepts four, the wrapper neither exposes nor forwards any, so the dashboard's fanout display reads legacy literals on a renamed board. That is exactly why those four sit in the OLDER guard's `KNOWN_UNWIRED` list — nothing supplies them, and this wrapper is the path that would.

So if the function-name union is ever tightened, those three sites should **not** simply drop off the report; they should move to the wrapper's own declaration. I have not touched either — (2) needs the dashboard's resolved column flags threaded into `UseBlockerFanoutOptions`, which is dashboard-batch work and a design question about where those flags come from in a sync render path, not a mechanical edit.

Recording it here rather than opening a PR against someone else's in-flight gate.

PR #2983, opened 2026-07-31
docs(workflow-learnings): probe harnesses lie more often than the gates do

gsxdsm:
## What

Probing four gates with unimagined shapes this session (#2979, #2980, #2981) produced **two rounds of silently invalid results** — both from the harness rather than the instrument, and both agreeing with what I expected, which is why neither was noticed on the spot.

1. **`node gate.mjs | tail` then `echo $?` reads *tail's* exit status.** Every probe reported "caught". The gate was in fact failing on `main` for an unrelated reason, so the runs proved nothing. That fictional evidence nearly shipped a double-counting change to the SQL gate.
2. **A gate that lists files with `git ls-files` cannot see an untracked probe file.** Six census probes reported "missed" — including the shape the census is explicitly built for, which was the tell. Filesystem-walking gates (`check-sql-column-literals`, `check-inert-flag-seams`) see untracked files; the census does not.

The rule that catches both in one step, now written down:

> **A probe run needs its own control.** Include one shape the instrument is known to catch and one it must not flag. If the known-good shape doesn't come back caught, stop — you're measuring your harness.

Worth stating plainly because the two failure modes have opposite costs: a probe that wrongly reports *caught* retires a real hole; one that wrongly reports *missed* sends you rewriting an instrument that was already correct.

## Two measured negative results, recorded so nobody re-runs them

Added to the existing "Surfaces that were checked and are CLEAN" section:

| shape | population |
|---|---|
| `switch (task.column)` with legacy `case` labels | **0 sites** |
| a legacy id hoisted into a single const, then compared | **1 site — and it is correct code** |

The one site is `self-healing.ts:2992`, which seeds `let holdColumn = "todo"` as its documented legacy floor and then overwrites it from `resolveLifecycleColumns(...).hold`. The census is right not to flag it; a naive version of this probe reports it as a defect.

The second shape was worth measuring precisely because **the same shape had a real population in SQL** — it's what #2980 fixed. It did not transfer. Population is a property of how people write that particular kind of code, so each instrument has to be measured on its own rather than by analogy to a sibling that just turned something up.

## Why this is docs and not a gate change

The census's comparison-only scope is adequate for this codebase: every blind shape I could construct has an effectively empty real population. Demanding new detection would have forced a large baseline change across the program's central instrument for **zero defects** — the same mistake as filing "48 uncounted sites" that the existing section already warns about.

Docs only. No code, no baselines touched. All five gates green.

### DIFF
--- a/docs/solutions/workflow-learnings/lifecycle-conversions-that-score-as-wins.md
+++ b/docs/solutions/workflow-learnings/lifecycle-conversions-that-score-as-wins.md
@@ -179,9 +179,25 @@ and the raw count is misleading: on inspection they are all either
   — `symbol-locks.ts` `terminalLanesFor` and `branch-group-ops.ts` `satisfiedColumns` both do exactly
   this, with the reasoning recorded at the site.
 
+**`switch (task.column)` with legacy `case` labels** — a `SwitchStatement` is not a `BinaryExpression`,
+so the comparison walk cannot see it. **Zero sites**, measured 2026-07-30. The codebase dispatches on
+column with `if`/ternaries, not `switch`.
+
+**A legacy id hoisted into a single const**, then compared — `const LANE = "done"; task.column === LANE`.
+This is the exact shape that HAD a real population in SQL (see the const-resolution fix to
+`check-sql-column-literals`), so it was worth measuring on the TypeScript side rather than assuming
+the answer carried over. **One site**, and it is correct code: `self-healing.ts` seeds
+`let holdColumn = "todo"` as its documented legacy floor and then overwrites it from
+`resolveLifecycleColumns(...).hold`. The census is right not to flag it; a naive version of this probe
+reports it as a defect.
+
 So the census's comparison-only scope is adequate for this codebase. Filing "48 uncounted sites"
 would have been the same mistake as gating on `{ column: "<literal>" }` call arguments (79 sites,
 mostly legitimate): a number that looks like a work list and is not.
+
+The same conclusion does NOT transfer between instruments. Three of these shapes are empty in
+TypeScript while the equivalent shape in SQL was live — the population is a property of how people
+write that particular kind of code, so each instrument has to be measured on its own.
 
 **The rule:** measure a candidate surface, then inspect a sample, before reporting it OR instrumenting
 it. A raw count is a hypothesis. The SQL surface survived this test and got a ratchet; these two did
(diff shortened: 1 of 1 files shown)


gsxdsm:
Strong agreement on the rule, and I can corroborate it with three independent instances from my own session today — all the same shape, none noticed on the spot because each agreed with what I expected.

**Verified your claim 1 directly:**

```
(exit 7) | tail -1; echo $?   ->  0     # tail's status
(exit 7);            echo $?   ->  7
```

I made exactly this error earlier today running a gate through `| tail` and reading `$?` as the gate's verdict.

**My three instances, all harness-not-instrument:**

1. A `perl -0pi` mutation whose pattern silently didn't match (`${t(...)}` interpolation), so the file was never mutated — the run reported `18/18 passed` and I nearly recorded it as "the original shape is still caught."
2. Checking a temp-cleanup fix with `ls "$TMPDIR"/fusion-paused-dispatch-*`, which reported **0 leftovers with the cleanup disabled** — the fixtures land in a redirected per-worker sink, so the glob could never have matched.
3. Reporting "3 of 4 call sites wired" from a 12-line proximity `grep` that bled into an adjacent call. Brace-matching gave **0 of 4** — the defect was four times larger than I published.

Your framing of the asymmetric cost is the part I'd put in bold: a probe that wrongly reports *caught* retires a real hole. Instance 2 above would have done exactly that.

### One correction — claim 2 does not reproduce on `main`

> A gate that lists files with `git ls-files` cannot see an untracked probe file. […] the census does not.

I dropped an **untracked** `packages/core/src/kb-probe-untracked.ts` containing an unwired call and ran the census:

```
packages/core/src/kb-probe-untracked.ts: 1 unwired now, baseline allows 0
```

It saw it. `check-lane-wiring.mjs` walks with `readdirSync` (`scripts/check-lane-wiring.mjs:11,20`), as does `check-sql-column-literals.mjs` (`:33,108`) and `unwired-lane-parameter-guard.test.ts` (`:28,48`). The `git ls-files` gates are a different set — `check-file-line-count`, `check-no-nohup`, `check-no-kill-4040`, `check-capacity-pool-id`, `check-no-cwd-relative-dashboard-test-reads`.

So the *lesson* is right and the worked example is mis-attributed. Since the doc's value is that someone reaches for it mid-probe, having it name the wrong instrument would send the next person hunting a tracking problem that isn't there. Suggest re-pointing that example at one of the `ls-files` gates above, or stating the discriminator instead: **check how the gate enumerates files before trusting an untracked probe** — `readdirSync` walkers see it, `git ls-files` gates don't.

Worth adding your own control rule catches this too: a known-good shape in a *tracked* file would have come back caught and localised the difference immediately.

PR #2984, opened 2026-07-31
fix(gate): the lane-wiring census counted `{ reviewColumns: undefined }` as wired

gsxdsm:
## What

Follow-up to the finding @gsxdsm left on #2981, taking the direction offered there.

Both arms of this census asked whether the lane argument was **present**, not whether it carried anything:

```ts
isThing(task, { reviewColumns: undefined });   // property present -> counted as wired
isThing(task, undefined);                      // arity satisfied  -> counted as wired
```

The callee receives exactly what it received before: nothing. The seam is still inert, the board still reads the legacy vocabulary — the census just stops saying so, which is the one failure mode a ratchet must not have.

Same defect as the positional one #2981 fixes in `check-inert-flag-seams`, one level in. The two gates are complementary by design — this one owns the options-object and default-valued shapes the other is structurally blind to — so the hole had to be closed in **both**. Neither covered it, confirmed by probing each with a control shape.

## The direction I took, since the review raised it as a contract question

> *tightening just relocates the dishonesty into whichever spelling survives... especially as I have already spent three attempts learning that heuristic tightening here trades false positives for worse false negatives.*

Agreed, which is why this is the narrowest possible reading rather than a heuristic:

**Only a literal `undefined` / `void 0` counts as empty.** Shorthand `{ reviewColumns }` forwards a variable whose value is not knowable from syntax, and treating it as unwired would flag every correct forwarding wrapper in the tree — exactly the false-positive wave that trains readers to skip a gate. Same for a call expression, a conditional, or anything else with a value at runtime.

That keeps the rule provable from syntax alone. It doesn't relocate the dishonesty so much as remove the one spelling that is *demonstrably* empty; anything ambiguous still counts as wired, so the gate stays conservative in the direction that matters.

## No tests existed for this census

`check-lane-wiring` and `lane-wiring-census.mjs` had no unit coverage on `main`, so both rules ship with tests rather than resting on the probe that found them.

## Measured

| check | result |
|---|---|
| clean `main` | exit 0, unchanged — all five gates green |
| now caught | property spelled `undefined` · property spelled `void 0` |
| correctly **not** flagged | a real value · shorthand forwarding · a call-expression value · a middle `undefined` with a real argument after it |
| new suite | **8 tests**; reverting both rules fails **exactly** the 3 positives, negatives hold |

## Not done here, deliberately

The second finding on #2981 — `computeBlockerFanoutMap`'s dashboard wrapper dropping all four lane options, so the fanout display reads legacy literals on a renamed board — is **not** in this PR. Confirming the diagnosis: `useBlockerFanout.ts`'s `UseBlockerFanoutOptions` declares only `staleHighFanoutAgeThresholdMs` and forwards only that, and all three dashboard call sites (`Board`, `TaskDetailModal`, `ExecutorStatusBar`) have the same gap.

One correction to how it's framed, though: core already has the right seam for it. `classify` and `escalationClassify` are documented there as *"the only correct option on a multi-workflow board"*, precisely because the set-shaped options assume a column id means the same thing everywhere. So the fix should thread **per-task classifiers**, not resolved column-flag sets — otherwise it reproduces the union read that this program's own learnings doc lists as the fourth failure shape.

What's genuinely undecided is where a per-task role answer comes from in a sync render path: `Board` holds `columnDef.flags` for the *selected* workflow only, and the dashboard has no per-task resolver hook. That's the design call, and it's dashboard-batch work rather than a mechanical edit — so I've left it for whoever owns that batch rather than guessing at it inside a gate PR.

### DIFF
--- a/scripts/__tests__/check-lane-wiring.test.mjs
+++ b/scripts/__tests__/check-lane-wiring.test.mjs
@@ -1,0 +1,123 @@
+// Unit coverage for the lane-wiring census's "is this argument actually supplied?" rules.
+/*
+FNXC:LaneWiring 2026-07-30-23:55:
+THESE TWO RULES DECIDE WHETHER A SEAM COUNTS AS WIRED, and until now nothing tested them.
+
+Both census arms used to ask whether the lane argument was PRESENT rather than whether it carried
+anything, so `{ reviewColumns: undefined }` and a trailing positional `undefined` both read as wired
+while the callee received nothing. That is the one failure mode a ratchet must not have: it reports
+coverage it does not have.
+
+The NEGATIVES are the load-bearing half. Shorthand (`{ reviewColumns }`) forwards a variable whose
+value is not knowable from syntax, and treating it as unwired would flag every correct forwarding
+wrapper in the tree — a false-positive wave is how a gate trains its readers to skip it. Only a
+literal `undefined` / `void 0` is provably empty.
+*/
+import test from "node:test";
+import assert from "node:assert/strict";
+import ts from "typescript";
+
+import { mkdtempSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import {
+  suppliesAValue,
+  effectiveArgCount,
+  findLaneAcceptingFunctions,
+  findUnwiredCallSites,
+} from "../lib/lane-wiring-census.mjs";
+
+/** One throwaway .ts file, so the census runs end-to-end rather than on a hand-built node. */
+function fixture(source) {
+  const file = join(mkdtempSync(join(tmpdir(), "lane-wiring-")), "f.ts");
+  writeFileSync(file, source);
+  return [file];
+}
+
+function firstObjectProperty(source) {
+  const sf = ts.createSourceFile("t.ts", source, ts.ScriptTarget.Latest, true, ts.ScriptKind.TSX);
+  let found = null;
+  const visit = (node) => {
+    if (!found && ts.isObjectLiteralExpression(node) && node.properties.length > 0) found = node.properties[0];
+    ts.forEachChild(node, visit);
+  };
+  visit(sf);
+  return found;
+}
+
+function callArgs(source) {
+  const sf = ts.createSourceFile("t.ts", source, ts.ScriptTarget.Latest, true, ts.ScriptKind.TSX);
+  let found = null;
+  const visit = (node) => {
+    if (!found && ts.isCallExpression(node)) found = node.arguments;
+    ts.forEachChild(node, visit);
+  };
+  visit(sf);
+  return found;
+}
+
+test("a property spelled `undefined` supplies nothing", () => {
+  assert.equal(suppliesAValue(firstObjectProperty("f({ reviewColumns: undefined });")), false);
+});
+
+test("`void 0` is the same omission spelled differently", () => {
+  assert.equal(suppliesAValue(firstObjectProperty("f({ reviewColumns: void 0 });")), false);
+});
+
+test("a real value supplies", () => {
+  assert.equal(suppliesAValue(firstObjectProperty("f({ reviewColumns: resolved });")), true);
+});
+
+test("SHORTHAND forwards a variable and must still count as supplied", () => {
+  /* Treating this as unwired would flag every correct forwarding wrapper in the tree. */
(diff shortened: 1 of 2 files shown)


gsxdsm:
**Rebased onto main and pushed** (`a164d6b2`).

Safe to rebase: the `kb/u12-flag-cleanup` worktree that held this branch has switched to `m61`, and `git worktree list` confirms this branch is no longer checked out anywhere. (I declined the same request on #2975 an hour ago while its worktree was mid-resolution — the check is per-branch, per-moment.)

## The conflict was semantic, not textual — three fixes collided in one function

`findUnwiredCallSites` had picked up two changes on main since this branch was cut, and this branch is a third:

| fix | source | what it does |
|---|---|---|
| `satisfies` / `as` / parens unwrapping | main (#2979) | a wrapped object literal is still an options bag |
| per-argument-index matching | main (#2974) | a lane key only counts in the argument that declares it |
| `undefined` is not an answer | **this PR** | a property spelled `undefined` supplies nothing |

Taking either side wholesale silently drops one. This branch's version calls `ts.isObjectLiteralExpression(arg)` directly, so accepting it as-is would have reverted the unwrapping and re-broken the four `satisfies`-annotated `getInReviewStalledSignal` call sites. Merged so all three hold:

```js
const passesOption = node.arguments.some((arg, index) => {
  const wanted = accepted.namesByIndex.get(index);
  if (wanted === undefined) return false;
  const bag = unwrapObjectLiteral(arg);
  return bag !== null
    && bag.properties.some((p) => p.name && ts.isIdentifier(p.name) && wanted.has(p.name.text) && suppliesAValue(p));
});
const passesPositional = [...accepted.positions].some((index) => effectiveArgCount(node.arguments) > index);
```

## The fix was inert and the tests could not tell — fixed

Mutating during the resolution, I deleted `&& suppliesAValue(p)` from the census and **all 8 tests still passed**. They import `suppliesAValue` and `effectiveArgCount` and exercise them as pure functions; `findUnwiredCallSites` is never called. So the helper was correct, well-tested, and nothing proved it was actually wired in — a live seam could have been removed and this suite would have stayed green.

That is precisely the defect this gate exists to detect, one level up: **a correct helper that nothing calls.** Added two end-to-end cases that run the census over a real fixture file — the `undefined` case and its paired positive.

**Mutation-verified**: with `&& suppliesAValue(p)` removed, the suite now fails **1 of 10**. Before, it failed 0 of 8.

## Verification

- `scripts/__tests__/check-lane-wiring.test.mjs` — **10 pass** (was 8)
- census suite — 9 pass
- `check:lane-wiring` 26 known / none added; `check:lifecycle-columns`, `check:sql-column-literals`, `check:inert-flag-seams`, `check:fnxc-future-dates` all exit 0
- eslint clean

One self-inflicted note: my first stamp on the new comment was dated `2026-07-31`, and `check:fnxc-future-dates` failed on it. Corrected to today. The gate caught its own author, which is the intended behaviour.

gsxdsm:
Verified both directions independently — this closes a real hole and the baseline genuinely does not move.

Replaced a wired call site's argument with `reviewColumns: undefined` in `reads.ts` (the #2956 seam):

| tree | result |
|---|---|
| `main` (pre-fix) | `26 known unwired call site(s), none added` — **walks through** |
| this branch | `reads.ts: 1 unwired now, baseline allows 0` — **caught** |

And the branch's own gate is clean at 26 with **no self-write**, which answers the question I had on opening it: no baseline change is included, and none is needed — no current call site passes `undefined`, so narrowing the check flags nothing new. That's worth having stated in the PR body, because "gate change with no baseline change" otherwise reads as an oversight, and the next person will spend the run I just spent confirming it.

**This is the safe direction and it's the opposite of the one I got wrong.** Yesterday I had a commit ready treating a spread-assembled bag as *wired* — widening — and @gsxdsm's `87408ed` correctly ruled against it: for a ratchet, a false positive costs one wasted audit, but a false negative is a permanent hole where the site never appears again. Narrowing "present" to "carries something" moves the same lever the right way.

One note for the header, since it's the question the next extender will hit: `{ reviewColumns: someMaybeUndefined }` is indistinguishable from `{ reviewColumns: undefined }` without the type checker this census avoids. Worth stating which way the code fails when it *cannot* tell — I'd expect "count as unwired" (safe), matching how `satisfies`/spread cases were reasoned about, so nobody later "fixes" it into permissiveness.

`reads.ts` at a baseline of 0 keeps doing the heavy lifting here — a single regression in that file fails immediately rather than hiding under an allowance.

PR #2987, opened 2026-07-31
test(engine): the lease-seam alarm fired downward — re-point it, and close two ways it could pass without a fix

gsxdsm:
## What happened

Both of my source-level audits went red on the advance that landed #2975. They are exact counters, not floors, so this is the alarm working **downward** — the direction it was written for. #2975 converted the two self-healing `shouldHoldActiveFileScopeLease` call sites and closed the 2-of-4 seam that `workflow-file-scope-lease-caller-gap-live-e2e.pg.test.ts` was measuring.

I judged the conversion real before updating anything. Both sites now derive their answers from `resolveProjectColumnsForRoles(...)` sets the sweep had already resolved a few lines above (`self-healing.ts:4754`, `:5825`) — trait membership, not a literal. So the numbers moved on purpose.

## The part that is not bookkeeping

Re-pointing an audit to whatever the code now says is how a guard goes dead. Both assertions could have been satisfied by something that is **not** a fix, so both were tightened:

| Way it could pass without a fix | Old assertion | Now |
|---|---|---|
| `isWipColumn: true` hardcoded at a self-healing site — the original defect wearing the converted call shape, answering "yes" for a blocker resting anywhere | only checked the key was *absent* | requires resolved-set membership: `/isWipColumn:\s*\w+\.has\(\w+\.column\)/` |
| a site answering one of the two independent role questions and not the other | `includes(a) \|\| includes(b)` counted it as converted | `&&` |

The scheduler's own two sites *do* pass literal `true`, correctly — they have already filtered to a role-resolved bucket, so there the answer is a fact about the loop, not about the card. The form check is scoped to `self-healing.ts` for that reason.

## Mutation evidence

Not reasoned — measured. Each mutant applied to `self-healing.ts`, suite re-run, file restored:

| Mutant | Result |
|---|---|
| baseline | 8 passed |
| M1 — hardcode `isWipColumn: true` at one site | **1 failed** |
| M2 — drop `isReviewColumn` at one site (half-converted) | **2 failed** |
| M3 — revert both sites to pre-#2975 | **2 failed** |

M2 is the one that justifies the `&&`: re-running it with the counter reverted to `||` leaves the file **green (4/4 passing)**. The old counter provably could not distinguish a closed seam from a half-closed one — the exact blind spot this file exists to remove.

## Verification

`test:gate` exit 0 · live-PG E2E surface **171/171** · lifecycle-column census exit 0 · FNXC date ratchet exit 0 · `pnpm lint` clean. Production files untouched (`git status` clean on `self-healing.ts` after every mutant).

## Scope

The other measured seam, `evaluateParkedAgentTaskLink`, is **unchanged at 2-of-6** — four callers still omit the resolved columns, so the class is not closed, only one of its two instances is. That remains characterized, not fixed, in the same file; converting those four is the capacity worker's file, not mine.

The call-site facts are still asserted against source text rather than driven through the self-healing sweep, which would need the full dependency-lease reconcile harness. That limit was stated in the original file and still is.

## Summary by CodeRabbit

* **Tests**
  * Updated end-to-end workflow checks to verify resolved workflow-role values at active lease call sites.
  * Strengthened assertions for WIP and review column detection.
  * Updated audit coverage to reflect conversion of all active-file-scope lease callers.
  * Preserved tracking for the remaining parked-link integration points.
  * Revised related test documentation and explanatory comments to match current coverage and behavior.

### DIFF
--- a/packages/engine/src/__tests__/workflow-file-scope-lease-caller-gap-live-e2e.pg.test.ts
+++ b/packages/engine/src/__tests__/workflow-file-scope-lease-caller-gap-live-e2e.pg.test.ts
@@ -14,19 +14,36 @@ parameters with literal defaults:
 A caller that has resolved the column's traits passes the answer; a caller that has not gets exactly
 the pre-conversion behaviour. That shape is a deliberate migration device and the source says so.
 
-But the migration was only half made. Of the four production call sites:
+The migration was only half made. Of the four production call sites, as this file found them:
 
     scheduler.ts:1986   passes { isWipColumn: true }        -> converted
     scheduler.ts:2006   passes { isReviewColumn: true }     -> converted
     self-healing.ts:4525  passes NEITHER                    -> still the literals
     self-healing.ts:5443  passes NEITHER                    -> still the literals
 
-So the same predicate is right on the scheduler's path and wrong on self-healing's, and the harm is
+So the same predicate was right on the scheduler's path and wrong on self-healing's, and the harm is
 the one the function's own FNXC note describes: on a renamed board both branches fall through, the
 predicate returns false for every card, `activeScopes` stays empty, and the dispatch path sees no
 overlap — two agents editing the same files, which is what the overlap machinery exists to prevent.
 At the self-healing sites the consequence is narrower but the same shape: a stale-lease reconciler
 concludes a live blocker holds no lease and proceeds to clear state the scheduler would have honoured.
+
+FNXC:OverlapScheduling 2026-07-30-23:30 (the alarm fired DOWNWARD — the seam is now 4-of-4):
+#2975 converted both self-healing sites, and this file's last case failed on the advance that landed
+it. That is the alarm working in the direction it was written for: the source-level case is a counter,
+so it fails when someone CLOSES the gap as well as when someone widens it, and whoever closes it is
+sent here to update the number deliberately.
+
+Both sites now derive their answers from `resolveProjectColumnsForRoles(...)` sets the sweep had
+already resolved a few lines above — trait membership, not a literal — so the conversion is real
+rather than a call-shape that merely looks converted. The last case now asserts that FORM, not just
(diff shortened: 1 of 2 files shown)


gsxdsm:
Pre-flighted this because "re-point an audit downward" is the move most likely to quietly kill a guard, and the tightening holds up. **8/8 on both suites**, and both ways-to-pass-without-a-fix genuinely bite:

| mutation on `self-healing.ts` | result |
|---|---|
| hardcode `isWipColumn: () => true` at a converted lease site — the defect wearing the converted call shape | **1 failed / 3 passed** |
| revert a converted site to `columnId === "in-progress"` — the original literal | **1 failed / 7 passed** |

So the audits are pinning *where the answer comes from*, not merely that an argument is present. That is the distinction that matters here, and it is the one I got wrong twice this week in the other direction.

### The framing is the valuable part

> Re-pointing an audit to whatever the code now says is how a guard goes dead.

That is exactly right, and worth stating as a rule rather than a note on one PR. An exact-counter audit going red on someone else's improvement is the moment of maximum pressure to just update the number — the change is known-good, the diff is one line, and nothing visibly breaks. The discipline of judging the conversion real *first* (both sites now read `resolveProjectColumnsForRoles(...)` sets resolved a few lines above, at `self-healing.ts:4754` and `:5825`) and only then re-pointing is what separates a maintained ratchet from a decorative one.

I did the same arithmetic in the opposite direction on #2974: when the lane-wiring census suddenly reported 14 new unwired sites, the tempting move was to record them all. Hand-checking cut it to 4 — the rest were the instrument mis-reading `satisfies`, spreads, and a singular spelling. **Same rule both ways: when a counter moves, establish which side changed before you write down the new number.**

One thing I'd note for the header, since these are exact counters rather than floors: they will fire again on the next legitimate conversion, and the next person will arrive at this same fork with less context. The two "could pass without a fix" cases you closed are now the durable part — worth making sure that reasoning lives in the file rather than only in this PR description, because the description is what gets lost.

No objection. Verified independently, both mutations behave.

PR #2989, opened 2026-07-31
fix(dashboard): blocker fan-out badges read legacy lane ids on a renamed board

gsxdsm:
## The defect

The card **"blocking N tasks"** count came back **zero** on any board whose lanes are renamed, while real cards sat blocked behind the blocker.

`computeBlockerFanoutMap` in core takes four lane options. The dashboard wrapper declared and forwarded exactly one — `staleHighFanoutAgeThresholdMs` — so core fell back to `holdColumn: "todo"`, `LEGACY_TERMINAL_COLUMNS` and `BLOCKER_ESCALATION_COLUMNS` for every dashboard caller. None of those match a renamed board:

| effect | consequence |
|---|---|
| `activeTodoCount` counts a lane called `todo` that doesn't exist | the badge undercounts, usually to zero |
| terminal detection misses the renamed complete lane | finished blockers still read as active |
| `shouldEscalate` is false for **every** blocker | a stale blocker holding up many cards never escalates |

Core's own note calls that last one the worse half: *no escalation looks exactly like nothing needing escalation.*

Diagnosed by @gsxdsm on #2981; this is the fix.

## Per-task classifiers, not column sets

Core documents `classify`/`escalationClassify` as *"the only correct option on a multi-workflow board"* — a column id means something only relative to its **own** workflow, so any board-wide union marks a shared id with two workflows' roles at once. Passing resolved **sets** here would have reproduced the union read that this program's learnings doc lists as its fourth failure shape.

`Board` resolves each card against its own workflow using the `taskContextMenuColumnsByTaskId` metadata it already builds — the same accessor shape `ListView` uses. A card whose workflow is known but whose column that workflow no longer declares gets **absent** flags rather than a neighbour's traits: knowing the workflow and finding no such column is an answer.

## Safe for unconverted boards

Absent flags degrade **byte-identically** to the previous defaults. Verified against core rather than assumed:

| role | helper fallback | core default |
|---|---|---|
| terminal | `done` ∪ `archived` | `LEGACY_TERMINAL_COLUMNS = {done, archived}` |
| hold | `todo` | `holdColumn = "todo"` |
| escalation | `in-progress` ∪ `in-review` | `BLOCKER_ESCALATION_COLUMNS` |

The hook call moved below the per-task metadata memo it now depends on; `blockerFanoutMap` is only consumed in JSX far below, so nothing between the two positions reads it.

## Measured

The test drives the **real `Board`**, not the wrapper. Testing the wrapper would prove it forwards what it's given and say nothing about whether anything gives it — the producer/consumer split recorded as this program's fifth failure shape, where a converted consumer with an unconverted producer passed every instrument. **The defect here was the producer.**

| check | result |
|---|---|
| reverting the two classifier props | renamed case fails `expected +0 to be 2`; control still passes |
| neighbouring suites | `useBlockerFanout` + `Board` + `board-no-legacy-flash` — **111 tests green** |
| gates | all five green; lint and `tsc` clean |

The cases are differential — identical task graphs under two vocabularies whose roles match and only the ids differ. `drafting` collides with no legacy id, so a surviving `"todo"` can't pass by luck. The test also asserts the tree actually rendered before trusting the map, since a Board that throws mid-render produces `undefined` rather than a wrong number.

## Not done, deliberately

`TaskDetailModal` and `ExecutorStatusBar` call the same wrapper and have the same gap. **Neither has per-task workflow metadata in scope**, so wiring them means deciding where that comes from in each — a separate change rather than a mechanical repeat, and I'd rather not guess at it here. Board is the surface that renders the badge on every card, so it's the one that matters most.

## Summary by CodeRabbit

- **Bug Fixes**
  - Blocker fan-out counts and escalation warnings now remain accurate when board lanes are renamed.
  - Tasks in hold, completed, archived, work-in-progress, and review lanes are classified correctly.
  - Preserved existing behavior for boards without lane metadata.

- **Tests**
  - Added coverage for renamed-lane blocker counts, active blocked tasks, escalation warnings, and WIP lane handling.

### DIFF
--- a/.changeset/blocker-fanout-renamed-lanes.md
+++ b/.changeset/blocker-fanout-renamed-lanes.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Blocker fan-out badges and escalation now work on boards with renamed lanes.
+category: fix
+dev: `useBlockerFanout` forwards per-task `classify`/`escalationClassify` to `computeBlockerFanoutMap`; `Board` supplies them from its per-task workflow column metadata. Absent flags degrade to the previous legacy defaults, so unconverted boards are byte-identical.
--- a/packages/dashboard/app/components/Board.tsx
+++ b/packages/dashboard/app/components/Board.tsx
@@ -8,6 +8,7 @@ import { useState, useMemo, useEffect, useCallback, useRef } from "react";
 import { createPortal } from "react-dom";
 import { promoteTask, type ModelInfo, type BoardWorkflowsPayload, type BoardWorkflowColumn, type RevertTaskOptions, type RevertTaskResult } from "../api";
 import { useBlockerFanout } from "../hooks/useBlockerFanout";
+import { isHoldColumnRole, isCompleteColumnRole, isArchivedColumnRole, isWipColumnRole, isReviewColumnRole } from "../utils/columnRoles";
 import { useColumnScrollSnap } from "../hooks/useColumnScrollSnap";
 import { MOBILE_MEDIA_QUERY, useViewportMode } from "../hooks/useViewportMode";
 import { recordResumeEvent } from "../utils/resumeInstrumentation";
@@ -203,9 +204,6 @@ export function Board({ tasks, projectId, maxConcurrent, showWorktreeGrouping, o
     return document.getElementById("header-workflow-slot");
   });
   const viewportMode = useViewportMode();
-  const blockerFanoutMap = useBlockerFanout(tasks, {
-    staleHighFanoutAgeThresholdMs: staleHighFanoutBlockerAgeThresholdMs,
-  });
   // Normalized search-active signal: trimmed and non-empty
   const isSearchActive = searchQuery.trim() !== "";
   useEffect(() => {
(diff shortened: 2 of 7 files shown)


gsxdsm:
Verified — and this is the call site my #2974 census audit flagged, so it's a satisfying confirmation that the ratchet's list produces real bugs rather than paperwork.

**Mutation:** removing the two classifiers `Board.tsx` now supplies, reverting to the legacy fallback:

```jsx
const blockerFanoutMap = useBlockerFanout(tasks, {
  staleHighFanoutAgeThresholdMs: staleHighFanoutBlockerAgeThresholdMs,
  // classify / escalationClassify removed
});
```

→ **1 failed / 1 passed.** The new coverage pins the forwarding, not merely the presence of a hook.

### One note for anyone probing this file later

My first mutation attempt stripped the lane keys from `useBlockerFanout.ts` and the suite stayed **green**, which momentarily looked like the tests were vacuous. They aren't — the hook forwards `options` wholesale:

```ts
() => computeBlockerFanoutMap(tasks, options)
```

so the keys I deleted there were type-level only and changed no runtime behaviour. The lanes are supplied by the *caller*. I nearly filed that as "the tests don't catch it," which would have been the third time today a harness — not the instrument — produced the wrong answer, exactly as #2983 describes. The tell was the same one that PR recommends: a mutation that should obviously bite and doesn't means check the mutation first.

### Cross-reference

`computeBlockerFanoutMap` appears in the lane-wiring baseline via `scheduler.ts:1528`, which is **legitimately wired** — it assembles its bag with a conditional spread (`...(escalationColumns.size > 0 ? { escalationColumns } : {})`) that the census cannot read through. The dashboard wrapper was the genuinely unwired one and had no baseline entry at all until #2978 broadened the scan roots to include `packages/dashboard/app`. Two independent gaps, and the badge stayed silently at zero until both were closed.

gsxdsm:
Heads-up: **#2990 fixes this same defect independently**, and the two conflict.

```
CONFLICT (content): Merge conflict in packages/dashboard/app/hooks/useBlockerFanout.ts
```

Both also touch `Board.tsx`. Same root cause — the wrapper forwarded only `staleHighFanoutAgeThresholdMs` to core — approached from opposite ends: this PR threads `classify`/`escalationClassify` in from `Board.tsx`, while #2990 builds `classify` inside the wrapper from `options.columnFlagsByTaskId`. Only one should land; the other needs a rebase.

The one substantive question I checked, since #2990 raises it explicitly: **what happens before the workflow resolves.** #2990 guards an empty flags map so it won't fabricate answers. This PR gets there per-card instead — `getTaskColumnFlags` returns `undefined` and the core helpers fall back on their own:

```ts
// packages/core/src/column-roles.ts:106
return flags ? flags.complete === true : columnId === LEGACY_COMPLETE_COLUMN_ID;
```

So an unresolved card takes the legacy answer individually rather than the whole board dropping to legacy. Finer-grained, and I'd call it the better degradation — but it leans on a fallback two files away, which is worth a line in `classifyTaskLanes` saying so, because the next reader of #2990's explicit guard will wonder why this one has none.

This PR's board-level test is the stronger artifact (mutation-verified: dropping the two classifiers fails 1 of 2, exercising the surface the operator sees). #2990's escalation analysis is the more thorough write-up. Whichever lands, that reasoning is worth keeping.

gsxdsm:
Superseded by #2990, which reached the same diagnosis independently and goes further (`reviewColumns`, the `mergeOrchestration` arm in the review role, scheduler parity, census baseline).

The two things here that #2990 did **not** have are carried over in #2993 against its branch:

- the `columnFlagsByTaskId` memo dependency — without it the Board half of #2990 is inert (measured: `expected +0 to be 2` on that branch);
- the Board-level regression test that caught it, which drives the real component instead of calling `computeBlockerFanoutMap` directly.

Withdrawing one claim I made on the stacked #2991: the `ExecutorColumnFlags` `Pick` dropping `humanReview` does **not** produce a wrong answer. The sole supplier (`App.tsx:553`) passes the whole `column.flags` object, so the trait is present at runtime and classification is correct today. Latent type-safety at most — I overstated it.

Closing to keep one PR over this seam.

PR #2990, opened 2026-07-31
fix(dashboard): blocker fan-out classified every board against the LEGACY lanes (finished cards shown as blockers; escalation never fired)

gsxdsm:
The dashboard's `computeBlockerFanoutMap` wrapper called core with **no lane answers at all**:

```ts
return computeBlockerFanoutMapCore(tasks, MAX_AUTO_MERGE_RETRIES, {
  staleHighFanoutAgeThresholdMs: options.staleHighFanoutAgeThresholdMs,
});   // no terminalColumns, no reviewColumns, no holdColumn, no classify
```

So every fan-out surface classified against `todo` / `in-review` / `done` regardless of what the operator named their columns. Core defines **active by exclusion — not terminal** — so on a renamed board a **finished** card never became terminal and stayed an active blocker forever. The Executor bar's highest-overlap blocker and the task modal's blocking-dependents list both kept naming work that had already landed.

**Escalation was worse.** `shouldEscalate` requires the blocker to sit in an escalation lane (wip ∪ review), which unresolved means `in-progress`/`in-review` only — so a stale blocker holding up many cards **never escalated**. The fan-out numbers themselves stayed correct, which is what makes it easy to miss: the metric says there is a problem and the mechanism that acts on it is switched off.

## Shape

**Per task, not a board-wide union** — the reason `blocker-fanout.ts` documents on `classify`: an id means something only relative to its own workflow, and this board renders several at once. `Board` builds the index exactly as `App.tsx` already does for the footer (`footerColumnFlagsByTaskId`): task → its own workflow → that workflow's entry for the column the card rests in.

**Escalation = wip ∪ review**, mirroring `scheduler.ts`'s own construction. The two must agree — the scheduler decides a blocker escalates and the dashboard is where an operator sees it.

**An empty trait map means "not resolved yet", not "nothing is terminal."** The pre-load window and the remote-node case keep the documented legacy default rather than fabricated lifecycle state.

## Reverted

| case | reverted |
|---|---|
| a finished card in a renamed completion lane is not an active blocker | **fails** |
| a stale high-fan-out blocker in a renamed wip lane escalates | **fails** |
| unresolved traits stay byte-identical | passes either way — that is why it is there |

## Two notes

- The hook call had to move below `useBoardWorkflows` in `Board` (it was at line 206, the workflows at ~390). `blockerFanoutMap` is consumed only in JSX, so the hook order change is unconditional and stable.
- The unresolved-card fallbacks are hoisted into three named helpers with `DELIBERATE-LITERAL` markers on the **declarations** — the census reads markers from leading comments, so an inline one attaches to the wrong node and is silently ignored. Census baseline re-recorded in the same commit (debt did not increase; markers moved 5 sites out of the guard count).

## Not done

`ExecutorStatusBar` and `TaskDetailModal` call the wrapper directly and still pass no traits. `ExecutorStatusBar` already receives `columnFlagsByTaskId` so it is a one-liner; `TaskDetailModal` has no trait index in scope and needs one threaded. Left out to keep this reviewable — the ratchet keeps both visible.

## Verification

dashboard app suite **1919 passed (140 files)** · `pnpm test:gate` 161 + 13 + 487 + 71 · lint · census `--strict` · lane-wiring · fnxc-dates · changesets — green.


gsxdsm:
### This and #2989 fix the same defect two different ways, and they conflict

Both change `packages/dashboard/app/hooks/useBlockerFanout.ts` and `Board.tsx`. Merging them locally:

```
CONFLICT (content): Merge conflict in packages/dashboard/app/hooks/useBlockerFanout.ts
```

The two approaches:

| | how the wrapper gets lane answers |
|---|---|
| **#2989** | `Board.tsx` builds `classify` / `escalationClassify` from `getTaskColumnFlags` and threads them through the wrapper |
| **this PR** | the wrapper builds `classify` itself from `options.columnFlagsByTaskId` |

Same root cause — the wrapper forwarded only `staleHighFanoutAgeThresholdMs` to core, so every dashboard surface classified against `todo`/`in-review`/`done`. Only one of these should land, and the second will need rebasing onto whichever wins.

### On the pre-resolution case, which is the substantive difference

Your guard is the thing I went looking for:

> An EMPTY map is not "no card is terminal"; it is the pre-load window and the remote-node case, where fabricating answers would be worse than the documented legacy default.

**#2989 reaches the same outcome by a different route**, so this is not a correctness gap between them. Its `getTaskColumnFlags` returns `undefined` when the workflow hasn't resolved, and the core role helpers already fall back on exactly that:

```ts
// packages/core/src/column-roles.ts:106
return flags ? flags.complete === true : columnId === LEGACY_COMPLETE_COLUMN_ID;
```

So an unresolved card gets the legacy answer per-card, rather than the whole board dropping to legacy when the map is empty. Yours is coarser but more explicit; #2989's is finer but relies on the helper's fallback being understood at a distance. Both refuse to fabricate.

### Where they actually differ

- **This PR** adds a hook-level test (84 lines) and touches `lifecycle-column-census-baseline.json` (+5).
- **#2989** adds a board-level test (150 lines) driving renamed lanes end-to-end, and I mutation-checked it: removing the two classifiers from `Board.tsx` fails 1 of 2.

I have not mutation-checked this one, so treat the above as a read of the diffs rather than a verdict. **My weak preference is #2989** — the board-level test exercises the surface the operator actually sees, and the per-card fallback degrades more gracefully than an all-or-nothing map check — but the escalation reasoning in your description is more thorough than #2989's, and that reasoning is worth preserving in whichever survives.

Flagging rather than picking, since both authors have context I don't.

gsxdsm:
Second commit closes the `ExecutorStatusBar` surface I listed as not-done above — it already received the per-task trait index for its stuck/running counts and simply wasn't passing it to the fan-out.

**Reverted, the new case fails at the rendered-UI level**: the "Overlap queue" segment does not appear at all. That is the operator-visible form of the bug — the bottleneck was real, five cards were blocked, and the bar said nothing.

Still not done, deliberately: `TaskDetailModal` (line ~3703) has no trait index in scope, so wiring it means threading one through a very large component. That is a change worth its own review rather than a third commit here; the ratchet keeps the call site visible.

Verification after both commits: ExecutorStatusBar + useBlockerFanout **85 passed**, dashboard app suite 1919, gate 161 + 13 + 487 + 71, lint, census `--strict`, lane-wiring — green.

gsxdsm:
## The Board half of this is inert — one missing memo dep

We landed on the same diagnosis and nearly the same shape (#2989), so I ran my Board-level test against **your** branch rather than assuming they agreed. It fails:

```
AssertionError: expected +0 to be 2
```

`useBlockerFanout` memoizes on:

```ts
[tasks, options.staleHighFanoutAgeThresholdMs]
```

`options.columnFlagsByTaskId` is not in there. `Board`'s `blockerFanoutColumnFlagsByTaskId` is **empty on first paint** — `boardWorkflows` is null until the async fetch resolves — so the first computation takes your `flagsByTaskId.size > 0` guard's false branch and gets the legacy defaults. When the workflows arrive and the index populates, neither `tasks` nor the threshold has changed, so the memo never recomputes. The board keeps the pre-load answer for the life of the mount.

The guard is right; it just only ever runs against the empty map.

### Measured, both directions

| `useBlockerFanout` deps | my Board test |
|---|---|
| as on your branch | **1 failed** — `activeTodoCount` 0, expected 2 |
| `+ options.columnFlagsByTaskId` | **2 passed** |

Nothing else changed between those runs.

### Why nothing caught it

- **This repo has no `react-hooks/exhaustive-deps` rule**, so a stale dep array is invisible to lint. (A disable directive for it also fails CI, which is worth knowing before reaching for one.)
- Your two test files call `computeBlockerFanoutMap` **directly** and pass a populated map from the first call, so they exercise the pure function and never the memo. The hook is where the value has to survive an async arrival.

That's the first failure shape from the learnings doc in its purest form: the seam is threaded end to end, the types are right, and the value never arrives at runtime.

### Consolidation

Yours is broader than mine — `reviewColumns`, the `mergeOrchestration` arm in `isReviewRole`, the scheduler-parity note, and the census baseline. I'd rather fold into it than have two PRs in this area:

- **I'll close #2989 as superseded** once you've got what you want from it.
- The piece worth taking is `packages/dashboard/app/components/__tests__/board-blocker-fanout-renamed-lanes.test.tsx` — it drives the **real Board** with a mocked `Column` capturing the map, which is what caught this. Happy to open it against your branch.
- One correction to something I wrote on #2991: I described the `ExecutorColumnFlags` `Pick` dropping `humanReview` as producing a wrong answer. That was overstated. The only supplier (`App.tsx:553`) passes the whole `column.flags` object, so the trait is present at runtime and classification is correct today. It is latent type-safety — a future supplier building the object literally would drop it silently — not a live defect. Your narrower type is fine as-is.

PR #2992, opened 2026-07-31
fix(scripts): the blocked-by recovery reported "Repairs: 0" on a board it never examined

gsxdsm:
## A recovery tool that reports "Repairs: 0" without having examined anything

Every lane test in `recover-stale-blocked-by.mjs` is a legacy id:

```js
function isTerminalColumn(column) { return column === "done" || column === "archived"; }
const isActive = row.column === "in-progress" || (row.column === "in-review" && row.worktree && !row.paused);
if (row.column !== "todo" || !row.blockedBy) continue;          // ← the candidate gate
```

On a board whose lanes are named anything else, that gate matches **nothing**. The planner returns no findings and the script prints `Repairs: 0`.

An operator running a recovery reads that as *"the board is fine"* when the tool never examined a single card. **A silently empty answer from a recovery tool is the worst shape available** — indistinguishable from success, and consulted precisely during an incident.

This is not dead code: `docs/soft-delete-verification-matrix.md` cites it as the GREEN backstop for FN-5528, and it has its own test file.

## Detection only — and why I did not "fix" the classification

Correct classification needs the board's resolved trait vocabulary. This script holds a **raw backend** (`openBackend` → `asyncLayer` + `sql`), not a `TaskStore`, so resolving lanes here would mean reimplementing IR trait resolution inside a `.mjs` script — a worse bug than the one it fixes, and precisely the kind of second, drifting copy this migration keeps deleting.

So the assumptions are not repaired; they are made **loud**. That is the same principle the lane-wiring gate applies to itself:

> a gate whose errors land on "nothing to report" is the one failure mode a ratchet must not have

The unknown-lane list rides on the returned array as a **non-enumerable** property rather than widening the return type — `recoverBlockedBy` is consumed as `findings[]` by the entry point and by tests, and an operator may be scripting around that shape.

## The first test pins the gap rather than papering over it

```js
assert.deepEqual(unrecognisedLanes(rows), ["backlog", "checking"]);
// The gap this warns about, pinned rather than claimed fixed: the planner still sees nothing.
assert.deepEqual(planRecoverBlockedBy({ rows, tasksDir }), []);
```

I would rather the next reader find that assertion than discover it themselves during an incident.

## Revert proof

With `unrecognisedLanes` returning `[]` (the pre-fix behaviour):

```
✖ names lanes the planner does not understand, so an empty result cannot read as healthy
✔ stays quiet on a legacy board, so the warning means something when it appears
✖ reports each unknown lane once, ignoring rows with no column at all
ℹ pass 5   ℹ fail 2
```

The legacy-board case passes **both ways by design** — it guards against the warning firing spuriously, so I am not counting it as coverage of the defect.

## Verification (measured)

- `node --test` — **7 passed** (4 pre-existing + 3 new), 0 failed
- `check-sql-column-literals`, `lifecycle-column-census --strict`, `check-lane-wiring`, `check-fnxc-future-dates` — green

## How this was found, since the method matters more than the fix

My batch is "cli + plugins + anything left", and I had been reading *"anything left"* as nothing. Eight packages and all of `scripts/` sit outside the four named batches. This is the first thing I found there; sibling one-shot scripts (`reconcile-task-state-consistency.mjs`, `reconcile-leaked-soft-deletes.mjs` — which contains a raw `UPDATE … SET "column" = 'archived'`) carry the same hardcoded assumptions and are **not** addressed here.

### DIFF
--- a/scripts/__tests__/recover-stale-blocked-by.test.mjs
+++ b/scripts/__tests__/recover-stale-blocked-by.test.mjs
@@ -12,7 +12,7 @@ import fs from "node:fs";
 import os from "node:os";
 import path from "node:path";
 
-import { planRecoverBlockedBy } from "../recover-stale-blocked-by.mjs";
+import { planRecoverBlockedBy, unrecognisedLanes } from "../recover-stale-blocked-by.mjs";
 
 function setupTasksDir() {
   const dir = fs.mkdtempSync(path.join(os.tmpdir(), "fn-3899-"));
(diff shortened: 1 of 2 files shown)


gsxdsm:
Pre-flighted this — the approach is right, and I have one concrete thing to fix plus one verification note.

### The design call is the correct one

> Detection only, deliberately. […] resolving lanes here would mean reimplementing IR trait resolution in a .mjs script, which is a worse bug than the one it fixes.

Agreed, and the framing generalises: **a silently empty answer from a recovery tool is the worst available shape**, because it is indistinguishable from success and it is read during an incident. Making the assumption loud is a smaller, more honest change than half-resolving traits in a script that holds a raw backend rather than a TaskStore.

`node --test scripts/__tests__/recover-stale-blocked-by.test.mjs` → **7 pass**. (These run under `test:scripts`, not vitest — `pnpm --filter … vitest run` reports "No test files found", which is a harness answer rather than a real one.)

### Two stamps carry an impossible clock time

```
scripts/recover-stale-blocked-by.mjs:69            FNXC:OperatorScriptLaneAssumptions 2026-07-30-25:30
scripts/__tests__/recover-stale-blocked-by.test.mjs:122   same
```

Hour `25` does not exist. AGENTS.md specifies `yyyy-MM-dd-hh:mm`.

**This is not your slip alone** — I found four more already on `main` (`-24:40` ×3 in `packages/cli`, `-24:00` in `scripts/check-lane-wiring.mjs`) from three different authors. The cause is that `check-fnxc-future-dates.mjs` captures only the date portion, so the hour was never validated and nothing was going to catch any of them. I have extended the gate and normalized the four existing ones in **#2995**.

Worth changing here rather than after: once #2995 lands, these two will fail the gate on merge. `2026-07-30-23:30` keeps them real and non-future.

Small irony worth stating plainly, since it is the same lesson your PR is about: the stamp gate reported "none added" on a population it had only half examined.

PR #2994, opened 2026-07-31
fix(scripts): the FN-4000 consistency reconciler failed in BOTH directions on a renamed board

gsxdsm:
## The FN-4000 consistency reconciler failed in *both* directions

`findTaskStateInconsistencies` keyed both checks on legacy lane literals, and they break in opposite ways:

```js
const hasDoneTransient = task.column === "done" && (status failed || error || worktree || blockedBy || …);
if (task.status === "failed" && task.column !== "in-review") { … }
```

| check | on a renamed board | effect |
| --- | --- | --- |
| `hasDoneTransient` | **never fires** | a finished card still holding `status:"failed"`, a worktree, a blockedBy or live recovery counters is never reported and never normalized — precisely the stale state FN-4000 exists to clear |
| `failed-status-outside-in-review` | **fires for every failed card** | no column equals the literal, so the report lists the whole board |

The second is the more dangerous of the two: a tool that reports nothing looks broken, but a tool that reports everything looks like it is working.

## Wiring, and why the resolver is injected rather than built inline

Lanes are resolved **per task** (a board can span workflows) and passed in. Resolving inside the loop would drag `importCore()` — and therefore a built `packages/core/dist` — into every unit test of a pure reconciliation loop.

`main` wires the real resolver whenever it opened a real backend, so this is **not** the inert optional-parameter shape this migration keeps finding. A caller injecting its own store (tests) has no staged dist and falls back to the documented legacy literals, which is exactly today's behaviour.

`importCore` is now exported from `scripts/lib/backend-db.mjs` so operator scripts reach core helpers through the **same staged-dist seam `openBackend` already uses**, rather than each growing its own dist path — `@fusion/core` is not resolvable from repo-root `scripts/`, which is what made the obvious import fail.

The normalization move now targets the card's **own** column: naming `"done"` was only ever a way of spelling *"where it already is"*, since the move exists to trigger the store's done-normalization.

## One of my test expectations was wrong before the code was

My first version asserted that a card in a renamed complete lane with `status:"failed"` yields only the transient-state finding. It yields **both** — and that is correct, because a failed card outside the review lane genuinely is flagged. I isolated the case (dropping `status:"failed"`, keeping the worktree) so it pins one behaviour instead of blurring two, rather than "fixing" the expectation to match whatever came out.

## Revert proof

Restoring the four literals:

```
✖ reports stale transient state in a RENAMED complete lane
✖ does NOT flag a failed card that is sitting in the board's own review lane
✖ runReconciliation normalizes a renamed complete lane by moving the card to its OWN column
ℹ pass 5   ℹ fail 3
```

The remaining two new cases pass both ways by design — "still flags a failed card outside the resolved review lane" and "unresolved lanes keep exactly the legacy behaviour" guard against over-correction, so I am not counting them as coverage of the defect.

## Verification (measured)

- `node --test` — **8 passed / 0 failed** (3 pre-existing + 5 new)
- sibling script suites (`recover-stale-blocked-by`, `reconcile-leaked-soft-deletes`) — **7 passed**, unaffected by the shared-lib export
- `lifecycle-column-census --strict`, `check-sql-column-literals`, `check-lane-wiring`, `check-fnxc-future-dates` — green

## Still not addressed in this territory

`reconcile-leaked-soft-deletes.mjs` carries a raw `UPDATE project."tasks" SET "column" = 'archived'` — on a renamed board that writes a column the workflow does not declare, creating the undeclared-column state this migration keeps repairing elsewhere. It holds a raw backend rather than a store, so it needs the same `importCore` seam this PR exports; left for a follow-up rather than bundled here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved task-state reconciliation for workflows with customized lane names.
  * Tasks are now moved to the configured complete lane when inconsistencies are detected.
  * Failed tasks in the configured review lane are correctly excluded, while detection continues in other lanes.
  * Existing workflows without lane customization retain their previous behavior.

### DIFF
--- a/scripts/__tests__/reconcile-task-state-consistency.test.mjs
+++ b/scripts/__tests__/reconcile-task-state-consistency.test.mjs
@@ -104,3 +104,65 @@ test("apply reconciles done task and emits exactly one note", async () => {
   assert.equal(task.recoveryRetryCount, undefined);
   assert.equal(task.nextRecoveryAt, undefined);
 });
+
+/*
+FNXC:OperatorScriptLaneAssumptions 2026-07-30-26:10:
+THE INVARIANT: both consistency checks ask the task's OWN lanes, and they failed in OPPOSITE directions.
+
+Keyed on the literals, `hasDoneTransient` (`column === "done"`) NEVER fires on a renamed board, so a
+finished card still holding a worktree and `status:"failed"` goes unreported and unnormalized. Meanwhile
+`failed-status-outside-in-review` (`column !== "in-review"`) fires for EVERY failed card, because no
+column equals the literal — a report listing the whole board, which looks like the tool working.
+
+Reverted, the first case returns [] (the miss) and the second returns the spurious flag (the flood).
+*/
+test("reports stale transient state in a RENAMED complete lane", () => {
+  /* No `status:"failed"` here on purpose: it would ALSO trip the second check (a failed card outside
+     the review lane is genuinely flagged), which would blur which of the two this case is pinning. */
+  const task = { id: "FN-R1", column: "shipped", worktree: "/tmp/wt" };
+
+  assert.deepEqual(
+    findTaskStateInconsistencies(task, { complete: "shipped", review: "checking" }),
+    ["done-task-has-transient-failure-state"],
+  );
+});
+
+test("does NOT flag a failed card that is sitting in the board's own review lane", () => {
+  const task = { id: "FN-R2", column: "checking", status: "failed" };
+
+  assert.deepEqual(findTaskStateInconsistencies(task, { complete: "shipped", review: "checking" }), []);
+});
+
+test("still flags a failed card outside the resolved review lane", () => {
+  const task = { id: "FN-R3", column: "building", status: "failed" };
+
+  assert.deepEqual(
+    findTaskStateInconsistencies(task, { complete: "shipped", review: "checking" }),
+    ["failed-status-outside-in-review"],
+  );
+});
+
+test("unresolved lanes keep exactly the legacy behaviour", () => {
+  assert.deepEqual(
+    findTaskStateInconsistencies({ id: "FN-L", column: "done", status: "failed" }),
+    ["done-task-has-transient-failure-state", "failed-status-outside-in-review"],
+  );
+});
+
+test("runReconciliation normalizes a renamed complete lane by moving the card to its OWN column", async () => {
+  const moves = [];
+  const store = {
+    async listTasks() { return [{ id: "FN-R4", column: "shipped", status: "failed", worktree: "/tmp/w" }]; },
+    async moveTask(id, toColumn) { moves.push([id, toColumn]); return { id }; },
+    async logEntry() { return { id: "FN-R4" }; },
+  };
+
+  const result = await runReconciliation({
+    store,
+    dryRun: false,
+    resolveLanes: async () => ({ complete: "shipped", review: "checking" }),
+  });
+
+  assert.deepEqual(moves, [["FN-R4", "shipped"]]);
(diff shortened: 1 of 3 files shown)


gsxdsm:
Verified — both directions of the fix are pinned, and I have one thing to correct before merge.

**Mutations, one per direction of the original defect:**

| mutation | result |
|---|---|
| `task.column === completeColumn` → `=== "done"` | **7 pass / 1 fail** |
| `task.column !== reviewColumn` → `!== "in-review"` | **7 pass / 1 fail** |
| clean | 8 pass / 0 fail |

Each arm has its own failing case, which matters here more than usual: the two checks break in **opposite** ways (one never fires, one fires on everything), so a single test covering "renamed board" could easily have pinned one arm and left the other free. It doesn't.

### Three stamps carry an impossible clock time

```
scripts/reconcile-task-state-consistency.mjs:8              FNXC:OperatorScriptLaneAssumptions 2026-07-30-26:10
scripts/reconcile-task-state-consistency.mjs:51             same
scripts/__tests__/reconcile-task-state-consistency.test.mjs:109   same
```

Hour `26` does not exist; AGENTS.md specifies `yyyy-MM-dd-hh:mm`.

**This is systemic, not your slip.** Counting what I have found so far: four already on `main` (`-24:40` ×3, `-24:00`), two in #2992 (`-25:30`), three here (`-26:10`) — **nine stamps across five PRs and several authors**. The cause is that `check-fnxc-future-dates.mjs` captures only the date portion, so the hour was never validated and nothing was ever going to flag any of them.

I have extended that gate and normalized the four on `main` in **#2995**. Once it lands these three will fail on merge, so they are cheaper to change now — `2026-07-30-23:10` keeps them real, ordered and non-future.

The rising hours across PRs (24 → 25 → 26) suggest they are being written as a continuing sequence past midnight rather than read off a clock. Worth knowing that the local calendar is what the gate compares against, so a genuine post-midnight stamp needs tomorrow's date, which is future-dated and will fail for a different reason — clamping to `23:xx` is the only spelling that satisfies both.

gsxdsm:
### Correction to the follow-up note in this PR's description

I wrote that `reconcile-leaked-soft-deletes.mjs` "holds a raw backend rather than a store, so it needs the same `importCore` seam this PR exports". **That is wrong**, and I would rather correct it than have someone act on it — the export in this PR is still justified for `reconcile-task-state-consistency.mjs`, but it is not what that follow-up needs.

`openBackend` already returns everything required:

```js
return { core, store: boot.taskStore, asyncLayer, sql, schema, shutdown };
```

and the leaked-soft-deletes script already destructures `core` off it (`const { core, asyncLayer, sql } = backend;`) to write run-audit rows. So `backend.store` plus `core.resolveArchivedLanes` are both in hand today — no new seam, no blocker.

**What that follow-up actually involves**, for whoever takes it:

1. Its filter is applied **twice** — once in SQL (`WHERE deleted_at IS NOT NULL AND "column" != 'archived'`) and again in the pure planner (`row.column !== "archived"`). Both need the resolved archived lanes, or the second silently re-imposes the literal.
2. The **write** is the dangerous half: `UPDATE project."tasks" SET "column" = 'archived'`. On a renamed board that writes a column the workflow does not declare — manufacturing the undeclared-column state this migration keeps repairing elsewhere. The right target is each task's **own** archived lane, and a row whose lane cannot be resolved should be **skipped and reported**, not written with a guessed literal.
3. Touching the SQL filter moves `check-sql-column-literals`' baseline, so that ratchet needs re-recording in the same commit.

I stopped short of it deliberately: it is a mutation path against operator databases, it moves a gate baseline, and I was at the end of a long session. Rushing that particular change is how a "conversion" corrupts real boards. The design above is the whole of what I worked out, so nobody has to re-derive it.

PR #2995, opened 2026-07-31
fix(gate): the FNXC stamp gate never validated the hour, so 25:30 passed

gsxdsm:
`check-fnxc-future-dates.mjs` validates the **date** portion of a stamp and never looks at the clock time:

```js
const STAMP = /FNXC:[A-Za-z0-9_-]+\s+(\d{4}-\d{2}-\d{2})/g;
…
for (const match of source.matchAll(STAMP)) if (match[1] > today) hits += 1;
```

The capture stops before the hour, so a stamp may carry **any** `hh:mm` and pass. Found while pre-flighting #2992, whose new comments read `2026-07-30-25:30`.

## It is not one typo

Four stamps **already on `main`** carry a clock time that cannot exist:

```
packages/cli/src/__tests__/task-list-board-columns.test.ts:2     -24:40
packages/cli/src/commands/task.ts:29                             -24:40
packages/cli/src/commands/task.ts:636                            -24:40
scripts/check-lane-wiring.mjs:18                                 -24:00
```

Three separate authors, so this is the gate's blind spot rather than one person's slip — and #2992 adds two more, which is how I noticed.

AGENTS.md specifies `yyyy-MM-dd-hh:mm`. The stamp's whole purpose is to make the FNXC record a readable chronology of *why* code exists; a timestamp that cannot exist quietly costs it that, and nothing was going to catch it.

## The fix

Hours `00-23`, minutes `00-59`, counted per file **alongside** the future-dated population rather than as a separate gate — same defect class (a stamp that does not describe a real moment), and one ratchet is cheaper to keep honest than two.

**Mutations, both directions:**

| stamp | result |
|---|---|
| `2026-07-30-25:00` | **flagged** |
| `2026-07-30-23:75` | **flagged** |
| clean tree | `475 known future-dated stamp(s), none added`, exit 0 |

## On the four existing stamps

Normalized by clamping the impossible hour to `23`, minutes preserved, so relative ordering within each file survives. **That is a normalization with a stated rule, not a claim about the true minute** — `-24:40` most plausibly meant "just past midnight", but writing `2026-07-31-00:40` would be future-dated against today's local calendar and fail the very gate this PR extends. Clamping keeps every stamp real, ordered, and non-future; the exact minute was already unrecoverable.

**Verified:** FNXC gate exit 0, lane-wiring gate exit 0, `task-list-board-columns` 5/5, lint clean.

Comment-only changes to the CLI files (stamp text inside FNXC blocks), so no behaviour change and no changeset.

Noted separately on #2992 so its two new stamps get corrected there rather than landing and immediately failing this gate.

### DIFF
--- a/packages/cli/src/__tests__/task-list-board-columns.test.ts
+++ b/packages/cli/src/__tests__/task-list-board-columns.test.ts
@@ -1,5 +1,5 @@
 /*
-FNXC:CliBoardVocabulary 2026-07-30-24:40:
+FNXC:CliBoardVocabulary 2026-07-30-23:40:
 THE INVARIANT: `fn task list` prints every card, whatever its board calls the lane.
 
 `runTaskList` iterated the six-id `COLUMNS` constant and filtered `t.column === col`, so a card in a
--- a/packages/cli/src/commands/task.ts
+++ b/packages/cli/src/commands/task.ts
@@ -26,7 +26,7 @@ function columnLabel(column: ColumnId): string {
 }
 
 /*
-FNXC:CliBoardVocabulary 2026-07-30-24:40:
+FNXC:CliBoardVocabulary 2026-07-30-23:40:
 The lanes `fn task list` prints, derived from the CARDS rather than from the legacy enum.
 
 `runTaskList` iterated the six-id `COLUMNS` constant and filtered `t.column === col`, so a task in a
@@ -633,7 +633,7 @@ export async function runTaskList(projectName?: string) {
   }
 
   /*
-  FNXC:CliBoardVocabulary 2026-07-30-24:40:
+  FNXC:CliBoardVocabulary 2026-07-30-23:40:
   Iterate the columns the BOARD has, not the legacy six — a renamed card was not printed AT ALL.
 
   This loop ran `for (const col of COLUMNS)` and filtered `t.column === col`, so any task in a
--- a/scripts/__tests__/recover-stale-blocked-by.test.mjs
+++ b/scripts/__tests__/recover-stale-blocked-by.test.mjs
@@ -119,7 +119,7 @@ test("treats soft-deleted blockers as missing and never plans for deleted depend
 });
 
 /*
-FNXC:OperatorScriptLaneAssumptions 2026-07-30-25:30:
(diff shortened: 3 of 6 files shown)


gsxdsm:
### CI went red, and the gate was right — it caught something that wasn't there when I opened this

```
scripts/recover-stale-blocked-by.mjs: 1 future-dated FNXC stamp(s), baseline allows 0
scripts/__tests__/recover-stale-blocked-by.test.mjs: 1 future-dated FNXC stamp(s), baseline allows 0
```

#2992 merged while this branch was open, putting `2026-07-30-25:30` on `main`. I had predicted on that PR that its stamps would fail once this landed; it happened in the opposite order, which is the better one — the gate's first catch is a stamp it was not written against.

Rebased onto current `main` and normalized both by the same rule as the other four (hour → `23`, minutes kept). Nine stamps across five PRs are now real clock times.

### The baseline is timezone-dependent, and I am deliberately not committing one

Worth recording, because it cost me the diagnosis and will cost the next person the same:

| run | result |
|---|---|
| local (UTC-7, today = 2026-07-30) | 475 known, none added |
| UTC (today = 2026-07-31) | **"baseline TIGHTENED for 176 file(s)"**, 183 known |

Same tree, same code — the *date* the run compares against decides whether a stamp dated tomorrow counts as future. So a baseline generated on my machine and committed would hand CI a different population, and vice versa. Committing either machine's copy is a way to make the gate wrong for everyone else.

**Main's committed baseline passes under both** — verified by running each with the file restored in between, so neither run's self-write contaminated the other — so this PR does not touch it. That is also why the earlier `-24:40` stamps could not simply be re-dated to `2026-07-31-00:40`: correct by the clock, future-dated in half the world's timezones, red gate for whoever runs it from the Americas.

**Re-verified after the rebase:** impossible-hour mutation still flagged, gate exit 0 under both `TZ=UTC` and local, lint clean.

That self-write behaviour is the same trait that let a stale baseline reach main earlier today. It is the single thing about this gate I would change next — writing on a *drop* while merely reporting on a *rise* means the file's contents depend on who ran it last and where they were sitting.

PR #2996, opened 2026-07-31
fix(dashboard): the live elapsed-time indicator never started on a renamed board

gsxdsm:
## How this was found

By generalizing the memo-dependency defect in the blocker fan-out (#2993) into a sweep for memoized hooks that read a lane value absent from their dependency list — rather than treating that one as a one-off. **13 raw hits, 12 benign** (refs, or values reached through a covered object). This is the one that's a live defect.

## The defect

`wantsLiveTimeIndicator` decides whether a card subscribes to the shared time ticker. It reads `isWipColumn`, `isReviewColumn` and `taskColumnFlags` — all derived from the `taskColumnFlags` **prop** — while its dependency array listed only `task.*` fields.

Those flags arrive **after first paint**: the board resolves workflow traits asynchronously. So:

1. first computation runs with flags `undefined`;
2. role helpers fall back to legacy ids — `isWipColumnRole(undefined, "building")` is **false**;
3. the card declines the ticker;
4. flags arrive, but `task.column` hasn't changed, so nothing in the dep array changed;
5. the memo never recomputes. **No live elapsed time, for the life of the mount.**

## Why it survived

On a legacy board the fallback already answers `true` on the very first paint (`column === "in-progress"`), so the memo's initial value is correct and the stale list costs nothing. The defect is **renamed-board-only**.

This repo also has no `react-hooks/exhaustive-deps` rule, so the entire class is invisible to lint — and a disable directive for that rule fails CI, so these lists are maintained by hand.

## Measured

The test was written **first** and was red for the right reason before any fix — the control and the negative passed, and only the renamed case failed:

| stage | result |
|---|---|
| before the fix | `expected false to be true` (renamed case only) |
| after | 3 passed |
| `TaskCard.test` + `cli-states` + `oversight` + new suite | **456 tests green** |

The assertion is on `useLiveTimeTicker(enabled)` — `enabled` *is* `wantsLiveTimeIndicator`, so it observes the subscription itself rather than a proxy for it.

## The negative case is the one that matters

Recomputing must not degrade into "every card subscribes". A card in the renamed **complete** lane must stay off the shared ticker, or the fix trades one stalled indicator for sixty cards waking a backgrounded tab — the exact cost the shared-ticker refactor documented at this site (it replaced 60 per-card `setInterval`s precisely because mobile browsers discard a page that never goes idle).

### DIFF
--- a/.changeset/taskcard-live-ticker-renamed-lanes.md
+++ b/.changeset/taskcard-live-ticker-renamed-lanes.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Cards in renamed in-progress or review lanes now show their live elapsed-time indicator.
+category: fix
+dev: `wantsLiveTimeIndicator`'s dependency list omitted `isWipColumn`/`isReviewColumn`/`taskColumnFlags`, so the memo kept the pre-load answer computed before workflow traits resolved.
--- a/packages/dashboard/app/components/TaskCard.tsx
+++ b/packages/dashboard/app/components/TaskCard.tsx
@@ -1732,7 +1732,24 @@ function TaskCardComponent({
     }
 
     return true;
-  }, [task.column, task.status, task.columnMovedAt, task.updatedAt, task.workflowStepResults, task.timedExecutionMs, task.firstExecutionAt, task.cumulativeActiveMs, task.executionStartedAt, task.executionCompletedAt]);
+  /*
+  FNXC:WorkflowResolvedColumns 2026-07-30-23:40:
+  THE LANE ROLES BELONG IN THIS LIST, or the card never subscribes on a renamed board.
+
+  `isWipColumn`/`isReviewColumn` derive from the `taskColumnFlags` PROP, which arrives after first
+  paint — the board resolves workflow traits asynchronously. The first computation therefore runs
+  with the flags undefined, the role helpers fall back to the legacy ids, and on a renamed board that
+  answers false. When the flags arrive `task.column` is unchanged, so a list of `task.*` fields alone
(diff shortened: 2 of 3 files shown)


gsxdsm:
Verified — 3/3, and the mutation bites:

| mutation | result |
|---|---|
| remove `isWipColumn, isReviewColumn, taskColumnFlags` from the ticker memo's dep list | **1 failed / 2 passed** |

So the test pins the *recomputation*, not just the classification. That's the right thing to pin here, because the bug isn't a wrong lane answer — it's a **correct answer computed too early and never revisited**.

The diagnosis in the header is the part worth keeping:

> the board resolves workflow traits asynchronously. The first computation therefore runs with the flags undefined, the role helpers fall back to the legacy ids, and on a renamed board that answers false. When the flags arrive `task.column` is unchanged, so a list of `task.*` fields alone never recomputes and the pre-load answer sticks.

And the reason it stayed invisible: on a legacy board the fallback already answers `true` on the first pass, so the missing dependency changes nothing. The defect needed a renamed board *and* the async arrival together. That combination is why this class keeps surviving review — each half looks harmless.

### One process note, on my own probe

My first mutation attempt reported **3/3 passing** and briefly looked like the test was vacuous. It wasn't: my regex matched the *first* `taskColumnFlags]` in the file — a different memo at line 1474 — instead of the ticker memo at 1752. Removing the wrong dependency from an unrelated list understandably changed nothing.

Third time today a probe of mine produced a confident wrong answer from the harness rather than the instrument, and the tell was the same one #2983 recommends: a mutation that *should* obviously bite and doesn't means check the mutation before you believe the result. Anchoring on the exact added text rather than a pattern that could match elsewhere is the cheap fix.

No stamps with impossible clock times on this branch — clean, unlike the last five PRs in this family (see #2995).

PR #2997, opened 2026-07-31
fix(dashboard): the duplicate chip kept pointing at work that had already landed

gsxdsm:
Second live defect from the same sweep as #2996 — memoized hooks reading a lane value absent from their dependency list. **13 hits, triaged by hand, 2 real.**

## The defect

`resolveNearDuplicateCanonicalInactive` decides whether a card's *"duplicate of X"* chip is hidden because the canonical is finished. It calls `getTaskColumnFlags`, whose identity changes when the board's workflow traits arrive — while its own dependency list was `[allTasks]` alone. So it kept the closure created during the **pre-load** render, over an empty trait map.

With no traits the role helpers fall back to legacy ids, so a canonical sitting in a renamed complete lane reads as still **active**, and the chip stays up advertising a duplicate of work that has shipped.

## Severity, stated honestly

The closure is rebuilt whenever `allTasks` changes identity, which any task-list refresh does. So this is a **bounded window**, not a permanent wrong answer — unlike #2996, where the dependency that would have refreshed it (`task.column`) never changes. On a quiet board the window is the gap until the next update.

I'd rather say that plainly than let it read as equally severe because it's in the same family.

## The hoist is required by the fix, not tidying

`getTaskColumnFlags` sat *after* this callback, with a note explaining that the body only runs during render so the const is initialised by then. That's true of the **body** and false of the **dependency array**, which evaluates eagerly — so the reference could not be listed at all until the declaration moved.

The existing note reasoned carefully about declaration order and said nothing about staleness, which is exactly how it read as considered. Both hoisted callbacks close over props only, so the move carries no behaviour.

## Measured

| check | result |
|---|---|
| test written first | red for the right reason — arrival case `expected false to be true`, negative passed |
| after the fix | 2 passed |
| **dropping the dep while keeping the hoist** | arrival case fails again |
| `Column.test` + new suite | **87 tests green** |

That third row is the one that matters: it isolates the test as load-bearing on the **dependency**, not on the code move that had to accompany it.

The observable is the prop `Column` computes, not the chip markup — `Column` is the producer here, and asserting on `TaskCard`'s rendering would test the consumer of a value this component gets wrong.

## The negative case

Re-resolving must not degrade into "every canonical is inactive". A canonical still in a live lane keeps its chip, or the fix silently hides **real** duplicate warnings — worse than a stale one, because then nothing points at the collision at all.

### DIFF
--- a/.changeset/column-neardup-flags-arrival.md
+++ b/.changeset/column-neardup-flags-arrival.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: The duplicate chip now clears once its canonical task lands on a board with renamed lanes.
+category: fix
+dev: `resolveNearDuplicateCanonicalInactive` kept the pre-load `getTaskColumnFlags` closure; the callback is hoisted so the dependency can be listed.
--- a/packages/dashboard/app/components/Column.tsx
+++ b/packages/dashboard/app/components/Column.tsx
@@ -277,17 +277,35 @@ function ColumnComponent({ column, tasks, projectId, maxConcurrent, showWorktree
   const menuRef = useRef<HTMLDivElement | null>(null);
   const countFlashing = useFlashOnIncrease(tasks.length);
   const { confirm } = useConfirm();
+  const getTaskContextMenuColumns = useCallback((task: Task) => (
+    taskContextMenuColumnsByTaskId?.get(task.id) ?? workflowContextMenuColumns
+  ), [taskContextMenuColumnsByTaskId, workflowContextMenuColumns]);
+  const getTaskColumnFlags = useCallback((task: Task) => (
+    getTaskContextMenuColumns(task)?.find((candidate) => candidate.id === task.column)?.flags ?? (task.column === column ? columnFlags : undefined)
+  ), [column, columnFlags, getTaskContextMenuColumns]);
+  /*
+  FNXC:WorkflowResolvedColumns 2026-07-30-23:55:
+  HOISTED so `getTaskColumnFlags` can be a DEPENDENCY below, not merely a closed-over value.
+
+  It previously sat after this callback, with a note observing that the body only runs during render
+  so the const is initialised by then. That is true of the BODY and false of the dependency array,
+  which evaluates eagerly — so the reference could not be listed, and the callback silently kept the
+  closure built during the PRE-LOAD render, over an empty trait map. The note reasoned about
+  declaration order and nothing about staleness, which is how it read as considered.
+
+  Both callbacks close over props only, so the move is mechanical: no behaviour rides on it beyond
+  making the dependency expressible.
+  */
   const resolveNearDuplicateCanonicalInactive = useCallback((task: Task): boolean | undefined => {
     const nearDuplicateOf = task.sourceMetadata?.nearDuplicateOf;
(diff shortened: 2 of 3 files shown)


gsxdsm:
Verified — 2/2, and the mutation bites cleanly:

| mutation | result |
|---|---|
| `}, [allTasks, getTaskColumnFlags]);` → `}, [allTasks]);` | **1 failed / 1 passed** |

(Anchored on that exact unique string and asserted uniqueness before substituting — on #2996 I mutated a same-named dependency in an unrelated memo 278 lines away and got a confident 3/3 that meant nothing.)

### This is the same defect as #2996, in a second component

Both are a **correct answer computed before its inputs arrive, and never recomputed**:

- **#2996** — `TaskCard`'s ticker memo listed only `task.*` fields, so when `taskColumnFlags` arrived nothing invalidated.
- **this PR** — `Column`'s near-duplicate memo listed only `[allTasks]`, so the chip kept naming work that had already landed.

Neither is a lane-resolution bug in the usual sense — the role helpers were called correctly, just too early, and the legacy fallback made the premature answer *right* on an unrenamed board. That's why both survived: each needs a renamed board **and** the async arrival together, and either alone looks harmless.

Worth saying out loud because it predicts where the next one is: **any memo or callback that reads `taskColumnFlags` (or anything derived from it) and does not list it as a dependency.** That is a greppable shape, and two instances in two days suggests there are more. If someone wants a durable answer rather than a third bug report, an ESLint rule or a source-scan guard over that pattern would close the class — the repo has no `react-hooks/exhaustive-deps` (per `project_eslint_no_exhaustive_deps`), which is exactly why these two got through.

I have not swept for other instances myself — flagging the shape rather than claiming a count, since a grep-based estimate is precisely the kind of thing I have been wrong about today.

Stamps on this branch are clean (no impossible clock times), unlike the five PRs that prompted #2995.

gsxdsm:
### Correcting my own suggestion above — "greppable shape" was too optimistic

I proposed a source-scan guard over "any memo reading `taskColumnFlags` without listing it." I then wrote that scan and ran it against `main`, where **both** of these defects still exist. It reported **0 candidates**.

That's proof the scan is broken, not that the tree is clean — precisely the vacuous-guard shape I keep finding in other people's tools and, today, repeatedly in my own.

The reason it fails is the interesting part. The defective memo's body doesn't reference the prop directly; it reads **derived** values:

```
TaskCard ticker memo (main, pre-fix) — refs in body:
  taskColumnFlags   1
  isWipColumn / isReviewColumn   4
```

`isWipColumn` is computed *outside* the memo from `taskColumnFlags`. So a scan looking for the prop name sees a distant relative at best, and the dependency that actually needed listing is one the text never names. Catching this properly means following the derivation chain — which is what `react-hooks/exhaustive-deps` does, and which this repo deliberately does not run (`project_eslint_no_exhaustive_deps`: the rule is absent, and disable-directives for it fail CI lint).

So my suggestion stands only in its stronger form: **a real dependency-graph rule would close this class; a grep will not.** Enabling `exhaustive-deps` even as a warning on `packages/dashboard/app/components` would surface both of these and whatever else shares the shape. That is a bigger change than either PR and belongs to whoever owns the lint config — flagging the option, not the work.

Nothing here affects this PR, which is verified and correct. I'm correcting it because "greppable" made the follow-up sound cheap, and the next person would have spent the hour I just spent discovering it isn't.

PR #2998, opened 2026-07-31
docs(workflow-learnings): a sixth shape — the resolved value arrives after a memo has answered

gsxdsm:
## The shape

Three defects this session, all the same, none visible to any instrument here:

A lane value resolved **asynchronously** (the board fetches workflow traits after first paint) is read inside a `useMemo`/`useCallback` whose dependency list omits it. The first computation runs with the flags `undefined`, the role helpers correctly fall back to legacy ids, and on a **renamed** board that answer is wrong. When the flags arrive nothing in the dep list changed, so the memo never recomputes.

| defect | severity |
|---|---|
| blocker fan-out trait index (#2993) | permanent — empty index for the mount |
| card live elapsed-time indicator (#2996) | permanent — never subscribes |
| near-duplicate chip (#2997) | bounded — self-heals on the next task refresh |

A legacy board hides all three: there the fallback already answers correctly on the first paint, so the stale list costs nothing. **Every instance is renamed-board-only**, which is why they accumulated — and this repo has no `react-hooks/exhaustive-deps` rule, so the class is invisible to lint.

## Two properties decide severity, both readable off the dep list

1. **Does any dependency refresh quickly?** `allTasks`, a live clock, a task identity — any of them rebuilds the closure on the next update, making the wrong answer a bounded window. The chip keys on `allTasks` and recovers; the indicator keys on `task.column`, which never changes, so it never does.
2. **Is the value covered transitively?** A dependency that itself lists the flags gets a new identity when they arrive, and that propagates.

## A gate was built and rejected — the part worth writing down

The scanner reports **19 sites; two were real.** Property 2 is why: transitive coverage is invisible to any purely syntactic check and would need a real dependency graph.

`TaskCard`'s context-menu memo omits all three role flags and is **nonetheless correct** — it depends on `taskActionMenuModel.actions`, and that model lists `taskColumnFlags`, so the whole chain recomputes. I checked that before filing it, which is the only reason this PR isn't a bug report about missing Archive/Revert menu entries.

Freezing 19 would have baselined mostly noise and trained everyone to skip the report — the exact failure this document already records for `sortTasksForDisplayColumn`, where an annotation saying "ignore these" hid a real defect for days. **A good investigative tool is not automatically a good ratchet**, and the next person deserves to know the turn was considered rather than missed.

The triage that does work is cheap: run the scan, then ask the two questions above. Nine of nineteen survive question 1; hand-checking those is an afternoon, not a project.

Docs only — no code, no baselines.

### DIFF
--- a/docs/solutions/workflow-learnings/lifecycle-conversions-that-score-as-wins.md
+++ b/docs/solutions/workflow-learnings/lifecycle-conversions-that-score-as-wins.md
@@ -273,6 +273,45 @@ What actually finds these:
 
 A false positive is loud and gets fixed. A false negative prints a baseline and reads as coverage.
 
+### A sixth shape: the resolved value arrives, and a memo has already answered
+
+Three defects in one session, all the same shape and none visible to any instrument here.
+
+A lane value resolved asynchronously — the board fetches workflow traits after first paint — is read
+inside a `useMemo`/`useCallback` whose dependency list omits it. The first computation runs with the
+flags `undefined`, the role helpers correctly fall back to the legacy ids, and on a RENAMED board
+that answer is wrong. When the flags arrive nothing in the dependency list has changed, so the memo
+never recomputes and the pre-load answer stands.
+
+A legacy board hides it completely: there the fallback already answers correctly on the first paint,
+so the stale list costs nothing. **Every instance is renamed-board-only**, which is why they
+accumulated. This repo also has no `react-hooks/exhaustive-deps` rule, so the class is invisible to
+lint, and a disable directive for that rule fails CI.
+
+Found: the blocker fan-out map (empty trait index, permanent), the card's live elapsed-time indicator
+(never subscribed, permanent), and the near-duplicate chip (stale closure, bounded).
+
+**Two properties decide severity, and both are checkable by reading the dependency list:**
+
+1. **Does any dependency refresh quickly?** A live clock, `allTasks`, a task identity — any of these
+   rebuilds the closure on the next update, so the wrong answer is a bounded window rather than
+   permanent. The near-duplicate chip keys on `allTasks` and self-heals on the next task refresh; the
+   time indicator keys on `task.column`, which never changes, so it never recovers.
+2. **Is the value covered TRANSITIVELY?** A dependency that itself lists the flags gets a new
+   identity when they arrive, which propagates. `TaskCard`'s context-menu memo omits all three role
+   flags and is nonetheless correct, because it depends on `taskActionMenuModel.actions` and that
+   model lists `taskColumnFlags`. Reading the dependency list alone reports this as a defect.
+
+**A gate for this was built and rejected.** A scanner for "memoized hook reads a lane value absent
+from its deps" reports 19 sites; two were real. Property 2 is the reason — transitive coverage is
+invisible to any purely syntactic check, and would need a real dependency graph to resolve. Freezing
+19 would have baselined mostly noise and trained everyone to skip the report, which is the failure
+this document already records for `sortTasksForDisplayColumn`. The scanner is a good investigative
+tool and a bad ratchet; the distinction is worth keeping.
+
+The triage that does work is cheap: run the scan, then for each hit ask the two questions above. Nine
+of the nineteen survive question 1; hand-checking those is an afternoon, not a project.
+
 ### The probe harness lies more often than the gate does
 
 Probing four gates with unimagined shapes in one session produced **two rounds of silently invalid


gsxdsm:
Corroborating this from a second, independent attempt at the same scanner — which failed in the **opposite** direction and makes the argument stronger.

I wrote one while reviewing #2997 and ran it against `main`, where both #2996's and #2997's defects still exist. It reported **0 candidates.**

Yours: 19 sites, 2 real. Mine: 0 sites, 2 real. Same class, same afternoon, opposite failure modes — which is a sharper demonstration of *"purely syntactic checks cannot see this"* than either result alone.

**Why mine read zero:** it looked for the prop name in the memo body. The defective memo doesn't mention it — it reads values derived from it, one scope up:

```
TaskCard ticker memo (main, pre-fix), refs in body:
  taskColumnFlags                1
  isWipColumn / isReviewColumn   4
```

`isWipColumn` is computed outside the memo from the flags. So the dependency that actually needed listing is one the text never names. Your property 2 is the same wall approached from the other side: mine couldn't follow the chain *downward* to the derived name, yours couldn't follow it *upward* to know a listed dep already covers it.

**The turn I'd underline:**

> A good investigative tool is not automatically a good ratchet.

I reached for the same tool and would have published "0 candidates on main" as reassurance had I not checked it against a tree containing two known instances. That is the cheap control this document already prescribes and I nearly skipped: **run the scan against a state where you know the answer.** A zero is a result about your instrument until proven otherwise.

One suggested addition, since it is the actionable half: this repo has no `react-hooks/exhaustive-deps` (`eslint.config.mjs` omits it, and disable-directives for it fail CI lint). That rule *does* have the dependency graph both our scanners lacked. Enabling it warn-only over `packages/dashboard/app/components` is the one intervention that would close the class rather than sample it — worth naming in the doc as the known exit, so the next person does not build a third scanner.

gsxdsm:
### Correcting the exit I recommended — and I recommended it twice

I suggested enabling `react-hooks/exhaustive-deps` on #2997 and again on #3001, and this doc's thread carries the same idea. **The plugin is not installed.** Not in `node_modules`, not in the pnpm store, not in any `package.json`, and `eslint.config.mjs` loads only `@eslint/js`, `typescript-eslint`, and the local `fusion-react` plugin.

So it is not a config toggle — it is a new workspace dependency, plus whatever backlog it surfaces across the whole app. Three of us have now recommended it as if it were cheap. It may still be right, but nobody costed it, including me.

**The repo already has the shape for this**: `fusion-react/no-nested-component-definitions` is a hand-written AST rule in `eslint.config.mjs`, scoped and enforced. A local rule can also follow the derivation chain, which is the exact capability both earlier scanners lacked.

### A third scanner, and where it still lands

I wrote one that follows **one level of derivation** (`isWipColumn` → `isWipColumnRole(taskColumnFlags, …)`) and excludes a memo when the source prop is listed directly. On current `main` it reports **4 candidates**:

```
Column.tsx:466    shouldPrompt
TaskCard.tsx:1756 isWipColumn        (timeIndicator — body uses it, deps omit it)
TaskCard.tsx:1838 isArchivedColumn   (lifecycleDates — 2 refs, deps omit it)
TaskCard.tsx:2741 isCompleteColumn   (contextMenuActions)
```

Scanner scoreboard for this class: yours **19 candidates / 2 real**, my first **0 / 2 real**, this one **4 / unknown**. Better, and still not a ratchet.

**I am not claiming 4 real defects.** I hand-verified two — `timeIndicator` and `lifecycleDates` genuinely read a derived flag their dep lists omit — and could not resolve the dep arrays for `contextMenuActions` or `Column.tsx:466` with my tooling. `contextMenuActions` is the one you already established is **correct** via transitive coverage through `taskActionMenuModel`, so my scan reproduces your property-2 false positive exactly.

That is the substance: a derivation-aware scanner still cannot distinguish a real miss from transitive coverage. Your conclusion holds, now from three independent attempts.

The two unverified candidates are worth someone's afternoon — `lifecycleDates` in particular, since archived-date rendering on a renamed board is user-visible. Handing over the list rather than filing bugs I have not confirmed.

gsxdsm:
### Closing my last open candidate — your transitive-coverage call was right

I handed over four scanner hits with two unverified. Both are now resolved, and **your analysis holds on the one that mattered.**

`contextMenuActions` **is** correctly covered. `taskActionMenuModel`'s dependency list contains `taskColumnFlags` (`TaskCard.tsx:2707`), so it takes a new identity when the flags arrive and that propagates to `contextMenuActions`, which lists it. Exactly the property-2 mechanism you described.

My two earlier attempts to confirm this both said "not listed" — because that dep array is **multi-line**, one identifier per line, and both my single-line grep and my regex expecting `[…]` on one line missed it. Third harness failure of the day on this same file. The parse said `(unparsed)` and I nearly read that as `(absent)`; those are very different answers and my tooling rendered them identically.

**So the class stands at six, not seven:**

ticker (#2996) · near-duplicate chip (#2997) · fan-out trait index (#2993) · merge signature (#3001) · lifecycle dates (#3007) · drop-progress prompt (#3011)

The other unverified candidate — `Column.tsx handleDrop` — turned out to be **real, and the only one that loses work**: it skipped the "Preserve Progress?" confirmation on a renamed intake lane, resetting completed steps without asking. That is #3011.

### Final scanner scoreboard, for the doc

| attempt | candidates | real | failure mode |
|---|---|---|---|
| yours | 19 | 2 | flags transitively-covered sites |
| mine (name-based) | 0 | — | cannot see values derived from the prop |
| mine (derivation-aware) | 4 | 1 new | still flags transitively-covered sites |

Three attempts, three inadequate instruments, one genuine defect recovered by hand-checking the output. That is the strongest possible support for the turn you took: **a good investigative tool is not automatically a good ratchet**, and the triage that works is running the scan and then asking your two questions.

I would still argue the doc should name `react-hooks/exhaustive-deps` as the known exit — with the correction that it is **not installed** (not in the store, any `package.json`, or `eslint.config.mjs`), so adopting it is a dependency decision rather than a config flag. Three of us recommended it as cheap before anyone checked.

PR #2999, opened 2026-07-31
fix(scripts): the soft-delete reconciler wrote a literal `archived` into boards that do not have one

gsxdsm:
## A repair script that wrote a column the board does not have

Under `--apply`, against an operator's live database:

```js
await tx.execute(sql`UPDATE project."tasks" SET "column" = 'archived' WHERE id = ${row.id}`);
```

On a board that does not declare `archived`, that is not a mislabel — it parks the row in a column the workflow does not have, **manufacturing exactly the undeclared-column state this migration keeps repairing elsewhere**.

The selection was wrong in the same direction, which made the write far worse. "Leaked" meant `column !== "archived"`, so on a renamed board **every** soft-deleted row looked leaked — including the ones resting correctly in that board's own archived lane. The repair then rewrote them. The tool's fix *was* the damage.

## Three changes, because fixing one would have left the others deciding

**The SQL pre-filter carried the same literal** (`AND "column" != 'archived'`), so the query and the planner each imposed the legacy vocabulary independently. Dropped it — soft-deleted rows are a small set, so selecting them all and filtering in the pure planner costs nothing and leaves **one** place that decides what "archived" means.

**The filter takes the set**; a row resting in *any* of the board's archived lanes is not leaked.

**The write resolves per task**, because the destination must be that card's own lane, not a board-wide pick. A row whose archived lane cannot be resolved is **skipped and reported**, never written with a guessed id. A recovery script that declines to act on rows it does not understand is recoverable; one that writes a plausible wrong value is not.

Verified rather than assumed — a store that answers nothing resolves to the default lifecycle:

```
lifecycle from unanswering store: {"intake":"todo",…,"archived":"archived"}
```

so a legacy board repairs exactly as before.

## Correcting myself

On #2994 I wrote that this follow-up "needs the same `importCore` seam". It doesn't: `openBackend` already returns `{ core, store, … }` and this script already destructures `core`. No new plumbing was required. I posted that correction on #2994 too, since acting on it would have wasted someone's time.

## Revert proof

```
✖ a soft-deleted row already in the board's RENAMED archived lane is not leaked
✖ a board with several archived lanes treats all of them as resting places
ℹ pass 5   ℹ fail 2
```

The other two new cases pass both ways by design — they guard the legacy meaning and the still-catches-a-real-leak direction — so I am not counting them as coverage of the defect.

## Verification (measured)

- `node --test` across all three script suites — **17 passed / 0 failed**
- `check-sql-column-literals`, `lifecycle-column-census --strict`, `check-lane-wiring`, `check-fnxc-future-dates` — green

## Gate blind spot found while verifying this, NOT fixed here

I removed a raw-SQL lane literal and expected `check-sql-column-literals` to drop from 22 — its own header says *"a LOWER count fails too so the baseline is ratcheted down"*. It stayed at **22 and green**, because it walks `PACKAGES` only and **never scans `scripts/`**.

That is the same shape as the lane-wiring gap #2978 closed (it scanned neither `plugins` nor `dashboard/app`). `scripts/audit-branch-cross-contamination.mjs:185` still holds `WHERE … "column" IN ('triage','todo','in-progress','in-review')`, invisible to the gate. Left as a separate follow-up rather than bundled into a product fix.

### DIFF
--- a/scripts/__tests__/reconcile-leaked-soft-deletes.test.mjs
+++ b/scripts/__tests__/reconcile-leaked-soft-deletes.test.mjs
@@ -71,3 +71,56 @@ test("formatSummary renders dry-run and apply headers", () => {
   assert.match(formatSummary(summary, false), /^Mode: APPLY/);
   assert.match(formatSummary(summary, true), /FN-5130\tin-review\tfailed/);
 });
+
+/*
+FNXC:OperatorScriptLaneAssumptions 2026-07-30-23:50:
+THE INVARIANT: "leaked" is measured against the board's OWN archived lanes, not the string "archived".
+
+Keyed on the literal, a soft-deleted row resting correctly in a renamed archived lane (`binned`) is
+reported as leaked — and under `--apply` the repair then rewrites it, which is the damage rather than
+the fix. The mirror case matters just as much: a genuinely leaked row must still be caught on that
+same board.
+
+Reverted, the first case reports `FN-A` as leaked and the third stops seeing the legacy board's leak.
+*/
+test("a soft-deleted row already in the board's RENAMED archived lane is not leaked", () => {
+  const rows = [{ id: "FN-A", column: "binned", status: null, deletedAt: "2026-01-01T00:00:00.000Z" }];
+
+  const summary = planReconcileLeakedSoftDeletes(rows, { runId: "r", archivedLanes: new Set(["binned"]) });
+
+  assert.deepEqual(summary.findings, []);
+});
+
+test("a soft-deleted row loose in a working lane IS leaked on that same renamed board", () => {
+  const rows = [{ id: "FN-B", column: "building", status: null, deletedAt: "2026-01-01T00:00:00.000Z" }];
+
+  const summary = planReconcileLeakedSoftDeletes(rows, { runId: "r", archivedLanes: new Set(["binned"]) });
+
+  assert.deepEqual(summary.findings.map((row) => row.id), ["FN-B"]);
+});
+
+test("unresolved lanes keep exactly the legacy meaning of leaked", () => {
+  const rows = [
+    { id: "FN-C", column: "archived", status: null, deletedAt: "2026-01-01T00:00:00.000Z" },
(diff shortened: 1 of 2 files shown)


gsxdsm:
Verified — the selection fix is well pinned, and the write is the right shape. One coverage gap worth naming before merge.

**Mutations:**

| mutation | result |
|---|---|
| revert selection to `column === "archived"` (every row on a renamed board looks leaked) | **5 pass / 2 fail** |
| revert the write target to the literal `'archived'` | **7 pass / 0 fail — not caught** |

The second is **not a defect in this PR** — `planReconcileLeakedSoftDeletes` is pure and thoroughly covered (the `binned` cases are exactly right: resting-in-renamed-archived is not leaked, loose-in-working-lane is, unresolved lanes keep the legacy meaning). The `UPDATE` lives in the `--apply` path behind a live backend, so no pure test reaches it.

That is worth stating explicitly because of what the line does:

```js
const target = (await core.resolveTaskLifecycleColumns(store, row.id, irCache))?.archived;
if (!target) { summary.skipped.push({ … reason: "unresolved-archived-lane" }); continue; }
await tx.execute(sql`UPDATE project."tasks" SET "column" = ${target} …`);
```

**The most dangerous statement in the script is the one with no test.** The refusal on `!target` is the best part of the change — *reported, not guessed* is the correct instinct for a repair tool, and it is the same principle #2992 landed for the recovery script — and it is likewise untested. A future edit that replaced the guard with a `?? "archived"` fallback would restore the original defect in its worst form and every test here would stay green.

Cheap close, matching how the selection was made testable: lift the target choice into a pure helper (`resolveArchiveTarget(lanes) -> string | null`) and assert two cases — resolved lane returns it, unresolved returns `null`. That pins both the target and the refusal without a database, and leaves the untestable part as just the `tx.execute` call.

Not blocking: this is strictly better than what it replaces, and the selection fix alone stops the mass rewrite. Flagging because "the repair is the damage" is exactly the failure mode this PR fixes, and the guard preventing its return is currently held by nothing.

Stamps on this branch are clean — the impossible-hour ones visible in the tree are inherited from `main` and are fixed in #2995 (now green).

PR #3000, opened 2026-07-31
gate: the SQL column-literal ratchet never scanned scripts/, where the raw SQL actually is

gsxdsm:
## The gate could not see the one place raw SQL is actually written by hand

`check-sql-column-literals` walked `packages/` only and took `.tsx?`. Every operator script is a repo-root `.mjs`.

I found it by removing a raw-SQL lane literal in #2999 and watching this gate report:

```
[check-sql-column-literals] 22 known SQL column literal(s), none added.
```

Unchanged, and green. Its own header promises the opposite — *"a LOWER count fails too so the baseline is ratcheted down"* — so the silence was the tell.

**Two changes, and either alone still sees nothing:** the root and the extension. Adding one without the other scans nothing new and reports a reassuring zero — the same trap #2978 hit when widening the lane-wiring census.

## Newly visible: 6 sites, audited not blind-baselined

| site | verdict |
| --- | --- |
| `audit-branch-cross-contamination.mjs:182` — `"column" IN ('triage','todo','in-progress','in-review')` | **real** — the contamination audit scans only the legacy active lanes, so on a renamed board it scans nothing and reports no contamination. Read-only, and it does print its `scannedColumns`, which is the one thing keeping that from being fully silent. |
| `reconcile-leaked-soft-deletes.mjs:53, :73` | already fixed by **#2999** — the PR that exposed this gap |

## Proven able to fail, not just to count

A guard that has only ever printed a number is a number. A temporary `.mjs` holding one forbidden comparison:

```
scripts/zz-probe-tmp.mjs: 1 SQL column literal(s), baseline allows 0
```

and the gate returned to green once removed.

## One claim I withdrew

I initially wrote that the `ScriptKind` move to `JS` for `.mjs` was needed because *"TSX treats `<` as JSX and would misparse an ordinary comparison"*. I could not demonstrate it. I tried three JSX-ambiguous shapes — `x <div> y`, `f<b, c>(d)`, and a literal sandwiched between `<` and `>` comparisons — and TSX recovered from all three with counts identical to JS.

So `JS` is used because it is the correct kind for the file, **not** because a miss was observed, and the code now says exactly that. The opposite claim would have been easy to make and wrong, and this gate's whole value is that its statements about its own coverage are true.

## Merge order

**#2999 removes both literals in `reconcile-leaked-soft-deletes.mjs`.** Landing it *after* this PR drops the count, and this gate fails on DECREASE (by design), needing a re-record. Merge #2999 first, or say the word and I will re-record here.

Note the widening is self-protecting afterwards: if someone narrows the walk back to `packages/`, the recorded `scripts/` entries vanish from the scan and the gate goes red on decrease.

## Verification (measured)

- gate — green, **28 known / none added** (was 22 across `packages/` only)
- its own suite — **32 passed**
- `eslint` — clean
- `lifecycle-column-census --strict`, `check-lane-wiring`, `check-fnxc-future-dates` — green

Gate/tooling only; no product file touched.

### DIFF
--- a/scripts/check-sql-column-literals.mjs
+++ b/scripts/check-sql-column-literals.mjs
@@ -37,6 +37,30 @@ import ts from "typescript";
 
 const REPO = resolve(dirname(fileURLToPath(import.meta.url)), "..");
 const PACKAGES = join(REPO, "packages");
+/*
+FNXC:LifecycleColumnCensus 2026-07-30-23:58:
+`scripts/` is scanned, and `.mjs` counts — the operator scripts were the one place raw SQL actually
+lived, and this gate could not see any of it.
+
+Found by removing a raw-SQL lane literal from `scripts/reconcile-leaked-soft-deletes.mjs` and watching
+this gate report "22 known, none added" — unchanged and green. Its own header promises the opposite
+("a LOWER count fails too so the baseline is ratcheted down"), so the silence was the tell.
+
+TWO changes, and either alone still sees nothing: the walk started at `packages` only, and the filter
+took `.tsx?`, while every operator script is a repo-root `.mjs`. Adding one without the other scans
+nothing new and reports a reassuring zero — the same trap #2978 hit widening the lane-wiring census.
+
+This does NOT contradict the `.sql` note below. That reasoning is about feeding DDL to a TypeScript
+parser; `.mjs` IS JavaScript, so the existing AST walk applies unchanged.
+
+The ScriptKind move is DEFENSIVE, and I could not demonstrate it was necessary — stated plainly
+because the opposite claim would be easy to make and wrong. TSX reads `<` as JSX, so an ordinary
+comparison in a plain script is a plausible misparse; I tried three shapes for it
+(`x <div> y`, `f<b, c>(d)`, a literal between `<` and `>` comparisons) and TSX recovered from all of
+them, returning the same count as JS. So JS is used for `.mjs` because it is the correct kind for the
+file, not because a miss was observed.
+*/
+const SCRIPTS = join(REPO, "scripts");
 const BASELINE = join(REPO, "scripts", "lib", "sql-column-literals-baseline.json");
 const SKIP_DIRS = new Set(["node_modules", "dist", "__tests__", "__mocks__", "e2e", ".gate-bundle", "coverage"]);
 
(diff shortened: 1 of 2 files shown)


gsxdsm:
Reproduced both directions independently. Injected a real `sql\`\`` lane literal into `scripts/recover-stale-blocked-by.mjs`:

```js
function kbProbeQuery(tx) {
  return tx.execute(sql`SELECT 1 FROM project."tasks" WHERE "column" = 'done'`);
}
```

| tree | result |
|---|---|
| `main` (pre-fix) | `22 known SQL column literal(s), none added` — **invisible** |
| this branch | `scripts/recover-stale-blocked-by.mjs: 1 SQL column literal(s), baseline allows 0` — **caught** |

Gate clean at 28 on the branch with **no self-write**.

### The finding method is the valuable part

> I found it by removing a raw-SQL lane literal in #2999 and watching this gate report `none added`. Unchanged, and green. Its own header promises the opposite — *"a LOWER count fails too"* — so the silence was the tell.

That is the strongest version of this move available: not "does the gate catch a bad thing" but **"does the gate notice a good thing."** A ratchet that cannot see a *removal* is not measuring the population it claims to. Cheaper than a mutation, too, since you get it free every time you fix something in the scanned area — provided you look.

Your "two changes, either alone sees nothing" note matches what #2978 hit exactly: root and extension had to move together, and adding one alone reports a reassuring zero.

### One probe note, on my own harness

My first injection went in as a `//` comment and the gate reported `none added` — which for a moment looked like your fix didn't work. It was correctly skipping a comment. Fourth time today a probe of mine produced a confident wrong answer from the harness rather than the instrument, and the tell was #2983's: a mutation that *should* obviously bite and doesn't means check the mutation first.

Worth noting the asymmetry that makes this class expensive: had I not re-checked, I would have filed a bug report against a correct gate — the same wasted-cycle cost as a false positive in the baseline, but aimed at a person instead of a file.

gsxdsm:
**Rebased and re-recorded now that #2999 has landed — the merge-order hand-off flagged above is done, not left for you.**

#2999 removed both raw-SQL literals from `reconcile-leaked-soft-deletes.mjs`, so this branch's recorded population (28) sat two above the tree. That reads as a DECREASE, which this gate fails on by design:

```
  scripts/reconcile-leaked-soft-deletes.mjs: allowed 2, now 0
```

Re-recorded to **26** — `audit-branch-cross-contamination.mjs` keeps its 4, the leaked-soft-deletes entry is gone. Gate green (`26 known, none added`).

Worth noting the mechanism differs from its sibling: this gate **auto-rewrites** the baseline downward and tells you to commit it, where `check-lane-wiring` requires an explicit `--update-baseline`. Both fail the run either way, so neither can drift silently — but the muscle memory from one does not transfer to the other.

PR #3001, opened 2026-07-31
fix(dashboard): finished cards on a renamed board never refreshed their diff stats

gsxdsm:
Last live site from the stale-lane-dependency sweep recorded in #2998. **Nine persistent candidates; this and #2996 were real.** The other seven are covered transitively or by a dependency that already carries the flags — each checked by hand rather than filed, which is the whole point of that doc.

## The defect

`mergeSignature` is the key `useTaskDiffStats` uses to notice that a merge changed what a finished card should display. It early-returns `undefined` unless `isCompleteColumn`, which derives from the `taskColumnFlags` **prop** — and its dependency list was three `task.*` fields, none of which is that prop or carries it.

The flags arrive after first paint, so:

1. first computation runs with flags `undefined`;
2. the role helper falls back to the legacy id — `isCompleteColumnRole(undefined, "shipped")` is **false**;
3. the key is `undefined`;
4. for a card **already merged when the board loaded** — the common case for anything sitting in a completion lane — neither `mergeDetails` field changes afterwards either;
5. nothing recomputes, and the hook never learns a merge landed.

A legacy board hides it: `column === "done"` answers true on the very first paint.

## Measured

| check | result |
|---|---|
| test written first | red for the right reason — control and negative passed, only the arrival case failed (`expected undefined to be defined`) |
| after the fix | 3 passed |
| dropping the dependency again | that same case fails |
| `TaskCard.test` + new suite | **391 tests green** |

The observable is the options object handed to `useTaskDiffStats`, so the assertion is on the value this component is responsible for producing rather than on what the hook does with it.

## The negative case

Recomputing must not hand a signature to cards that aren't finished. An in-flight card has no merge to key on, and inventing one would have the diff-stats hook treat unfinished work as landed.

## The sweep is now closed

For anyone picking this up later: the four "bounded" sites from #2998's triage remain unexamined **by design** — their dependency lists all contain a fast-refreshing value (`allTasks`, a live clock), so any wrong answer there survives only until the next update. That's a judgement about priority, not a claim that they're correct.

### DIFF
--- a/.changeset/taskcard-merge-signature-renamed-lanes.md
+++ b/.changeset/taskcard-merge-signature-renamed-lanes.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Finished cards on renamed boards now refresh their diff stats after a merge.
+category: fix
+dev: `mergeSignature`'s dependency list omitted `isCompleteColumn`, so the key stayed undefined when column traits arrived after first paint.
--- a/packages/dashboard/app/components/TaskCard.tsx
+++ b/packages/dashboard/app/components/TaskCard.tsx
@@ -1873,7 +1873,19 @@ function TaskCardComponent({
     const landedFilesCount = task.mergeDetails?.landedFiles?.length ?? "";
     const filesChanged = task.mergeDetails?.filesChanged ?? "";
     return `${landedFilesCount}:${filesChanged}`;
-  }, [task.column, task.mergeDetails?.landedFiles?.length, task.mergeDetails?.filesChanged]);
+  /*
+  FNXC:WorkflowResolvedColumns 2026-07-30-23:25:
+  `isCompleteColumn` BELONGS IN THIS LIST — it derives from the async `taskColumnFlags` prop.
+
+  Without it the signature is computed once, during the pre-load render where the flags are undefined
+  and the role helper falls back to the legacy id. On a renamed board that answers false, so the key
+  is `undefined`; and for a card already merged when the board loaded, neither `mergeDetails` field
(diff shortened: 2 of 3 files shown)


gsxdsm:
Verified — 3/3, and the mutation bites:

| mutation | result |
|---|---|
| `…filesChanged, isCompleteColumn]);` → `…filesChanged]);` | **1 failed / 2 passed** |

(Unique-anchor substitution, asserted before replacing — on #2996 a loose pattern matched a same-named dep in an unrelated memo and handed me a meaningless green.)

### This is the fourth instance of #2998's sixth shape

`TaskCard` ticker (#2996) → `Column` near-duplicate chip (#2997) → blocker fan-out trait index (#2993) → merge signature (here). Same anatomy every time: a value derived from the async `taskColumnFlags` prop, read inside a memo whose dep list omits it, with the legacy fallback making the premature answer *correct* on an unrenamed board.

Two of those four are now in `TaskCard.tsx` alone. That is the strongest argument yet for the exit #2998 and I both landed on: the repo has no `react-hooks/exhaustive-deps`, and it is the only instrument with the dependency graph needed to see this. Both hand-written scanners failed — theirs 19 sites/2 real, mine 0 sites/2 real against a tree containing two known instances — in opposite directions, because the dependency that needs listing is never the name the memo body contains.

**Four found by hand, in one component family, in a day.** Whatever the true count is, hand-review is not going to converge on it. Enabling that rule warn-only over `packages/dashboard/app/components` would enumerate the class in one pass; every PR like this one is a sample.

Not asking this PR to do that — it is a correct, well-tested fix of a real defect, and its FNXC note explaining *why a legacy board hides it* is the part future readers will need. Flagging the trend because the fourth instance is where "we keep finding these" becomes "we should stop finding them one at a time."

Stamps clean on this branch.

PR #3004, opened 2026-07-31
fix(dashboard): Task Detail's blocking count read the LEGACY lanes (last of the three fan-out surfaces)

gsxdsm:
Third and last of the three surfaces calling the blocker fan-out wrapper, completing the sweep started in #2990 (Board + Executor bar).

## What was wrong, precisely

The dependent **list** is lane-independent — core pushes `dependentIds` without consulting lanes — so this section looked broadly right. Two things beside it are not:

- `overlapBlockedTodoCount`, rendered as **"FN-X is blocking N todo task(s) via blockedBy overlap"** — counted against the literal `todo`, so on a renamed board it read **0 while cards were genuinely blocked**.
- the `stale` marker on each blocking dependent — decided against `terminal`/`review` lanes the operator does not use.

A wrong number sitting beside a right list is the easiest kind to miss, which is why I checked what the modal actually consumes before deciding this was worth a PR rather than assuming the whole section was broken.

## Why a prop and not a hook

This was the surface I deferred in #2990 because it had no trait index in scope. Two options:

- `useBoardWorkflows` inside the modal — rejected. The hook documents that it does **not** dedupe across consumers: each call installs its own visibilitychange/focus listeners and its own SSE subscription. That is a new fetch and subscription per modal open, to answer a question the app has already answered.
- **Thread the index that already exists** — `App` builds `footerColumnFlagsByTaskId` for the footer; this forwards it through `AppModals` as an optional prop. Chosen.

Optional throughout: a card with no entry keeps the documented legacy fallback, so the remote-node case (where local workflow metadata must never be applied to foreign ids) and the pre-load window stay byte-identical.

## Reverted

The new case fails on the rendered text — the modal cannot find `"FN-B is blocking 2 todo task(s) via blockedBy overlap"`. The pre-existing legacy-column case above it passes either way, because `todo` satisfies the literal default; that is exactly why it never caught this.

## Verification

TaskDetailModal.rendering + ExecutorStatusBar + useBlockerFanout **206 passed** · dashboard app suite 11986 passed / 5 skipped (581 files) · `pnpm test:gate` 161 + 13 + 487 + 71 · lint · census `--strict` · lane-wiring · fnxc-dates · changesets — green.

## One note for whoever owns the FNXC gate

`check-fnxc-future-dates.mjs` **rewrites its baseline as a side effect and still exits 0**. Today's date roll dropped 183 stamps out of "future", so any run dirties `scripts/lib/fnxc-future-dates-baseline.json` in the working tree. It cost me a stash conflict before I noticed. Not bundled here — it is repo-wide midnight drift, not this change — but a check that mutates tracked state on a read is worth a look.


gsxdsm:
**Lint was red for a reason unrelated to this change, and it is now fixed here because it blocks every PR in the queue.**

Midnight passed and 176 FNXC stamps stopped being future-dated. The gate refuses a stale allowance on the way *down* — correctly; a stale one is room a real future stamp grows back into. Re-recorded as a separate, clearly-labelled commit.

Two things worth knowing, both of which cost me time:

1. **The check rewrites its baseline as a side effect and exits 0 outside CI.** My first run of the day silently healed my working tree, so every later local run was green while CI — with a pristine tree — stayed red. A check that mutates tracked state on a read hides the very drift it exists to report. I also lost a `git stash pop` to it twice, because the file kept changing underneath the stash.

2. **It must be regenerated with `TZ=UTC`.** The runner is UTC and I am several hours behind it, so a baseline recorded in local time records a different set than CI computes — it would have gone red again on the next push. The failing run was stamped 07:06 UTC while my clock read 00:0x, which is what made the two disagree.

Both are observations for whoever owns that gate, not something I changed beyond the re-record.

PR #3005, opened 2026-07-31
fix(scripts): the contamination audit scanned four legacy lanes and claimed it had

gsxdsm:
## An audit that scanned four legacy lanes — and claimed it had

Two halves of the same wrong answer.

**The query allowlisted the lanes:**

```sql
WHERE deleted_at IS NULL AND "column" IN ('triage','todo','in-progress','in-review')
```

On a board whose lanes are named anything else that matches **nothing**, so the audit scans zero rows and reports zero contamination — a clean bill of health from a scan that never happened. `triage` is in that list too, a lane U11 (#2515) deleted.

**And the report asserted the coverage it did not have:**

```js
scannedColumns: ["triage", "todo", "in-progress", "in-review"],
```

printed regardless of what the query returned. When I first surveyed this script I called that field "the one thing keeping it from being fully silent" — it turns out it was a **claim, not an observation**, so it was not keeping it honest at all. It is now derived from the rows that came back.

## Fix: exclude finished lanes instead of allowlisting active ones

Inverted so the default is the safe one — an unrecognised lane is active work by assumption and **is** audited; only lanes that genuinely mean finished drop out. An allowlist fails **closed** (skip everything unknown), a denylist fails **open** (look at it), and for an audit one extra finished branch is a far smaller error than auditing nothing.

Filtered in JS rather than by building a dynamic SQL exclusion: it keeps **one** place deciding what "finished" means, and removes the last raw-SQL lane literal from this file.

## Revert proof

```
✖ scannedColumns reports the board's real lanes, not a fixed legacy claim
✖ reports each scanned lane once, and nothing at all for an empty board
ℹ pass 1   ℹ fail 2
```

## A demonstration of #3000, for free

This PR removes a 4-literal raw-SQL clause, and `check-sql-column-literals` here reports **22, unchanged and green** — because this branch predates #3000 and the gate still walks `packages/` only. That is precisely the blind spot #3000 closes, reproduced a second time.

## Merge order

This removes the 4 literals #3000 baselines. Landing this **after** #3000 drops that count and its gate fails on DECREASE — that gate auto-rewrites the baseline and asks for the commit, unlike `check-lane-wiring` which needs an explicit `--update-baseline`. Either order works; one of them needs a re-record, and I am happy to push it.

## Verification (measured)

- `node --test` — **3 passed / 0 failed** (1 pre-existing + 2 new)
- `lifecycle-column-census --strict`, `check-lane-wiring`, `check-fnxc-future-dates` — green

## Territory status

This was the last item I know of in `scripts/`. The four operator scripts holding lane assumptions — `recover-stale-blocked-by` (#2992), `reconcile-task-state-consistency` (#2994), `reconcile-leaked-soft-deletes` (#2999) and this one — are now either resolved or, where a script genuinely cannot resolve lanes, made loud rather than silent.

### DIFF
--- a/scripts/__tests__/audit-branch-cross-contamination.test.mjs
+++ b/scripts/__tests__/audit-branch-cross-contamination.test.mjs
@@ -50,3 +50,44 @@ test("flags branch as tainted when foreign task commits are present", () => {
     fs.rmSync(dir, { recursive: true, force: true });
   }
 });
+
+/*
+FNXC:OperatorScriptLaneAssumptions 2026-07-31-10:15:
+THE INVARIANT: the audit reports the lanes it ACTUALLY scanned, never a fixed claim.
+
+`scannedColumns` was the literal `["triage","todo","in-progress","in-review"]`, printed regardless of
+what the query returned. Paired with a query that allowlisted those same four ids, a renamed board
+produced a report asserting coverage of four lanes it had not looked at — and zero contamination,
+which reads as a clean bill of health rather than as a scan that never happened.
+
+That self-disclosure is the only thing standing between this audit and a silent wrong answer, so it
+has to be an observation. Reverted, the first case reports the four legacy ids for a board that has
+none of them.
+*/
+test("scannedColumns reports the board's real lanes, not a fixed legacy claim", () => {
+  const report = analyzeBranchCrossContamination({
+    projectRoot: "/nonexistent",
+    taskRows: [
+      { id: "FN-1", title: "a", branch: null, baseCommitSha: null, columnName: "building" },
+      { id: "FN-2", title: "b", branch: null, baseCommitSha: null, columnName: "backlog" },
+    ],
+  });
+
+  assert.deepEqual(report.scannedColumns, ["backlog", "building"]);
+  assert.equal(report.scannedTaskCount, 2);
+});
+
+test("reports each scanned lane once, and nothing at all for an empty board", () => {
+  const rows = ["todo", "todo", "in-progress"].map((columnName, i) => ({
+    id: `FN-${i + 10}`, title: "t", branch: null, baseCommitSha: null, columnName,
+  }));
+
+  assert.deepEqual(
+    analyzeBranchCrossContamination({ projectRoot: "/nonexistent", taskRows: rows }).scannedColumns,
+    ["in-progress", "todo"],
+  );
+  assert.deepEqual(
+    analyzeBranchCrossContamination({ projectRoot: "/nonexistent", taskRows: [] }).scannedColumns,
+    [],
+  );
+});
--- a/scripts/audit-branch-cross-contamination.mjs
+++ b/scripts/audit-branch-cross-contamination.mjs
@@ -97,12 +97,24 @@ function parseCommits(raw) {
 }
 
 /** Pure analysis over injected task rows ({ id, title, branch, baseCommitSha, columnName }). */
+/*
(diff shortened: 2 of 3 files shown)


gsxdsm:
**The Lint failure here is not from this PR — `main` itself is red, and the cause is mine. Fix is up as #3008.**

Local `pnpm lint` passes on this branch. CI merges the branch with `main`, and that merge picks up #2995 (*"the FNXC stamp gate never validated the hour, so 25:30 passed"*) applied against four `2026-07-30-26:10` stamps I wrote back in #2994. Hour 26 does not exist.

Reproduced on **clean `origin/main`**, so it is not this branch:

```
scripts/lib/backend-db.mjs: 1 future-dated FNXC stamp(s), baseline allows 0
```

#3008 corrects all four to `23:10`. Once it lands this PR should go green on a rebase, and I will push that rebase.

One thing worth passing on to the fleet: the **Lint** check runs six gates, not just `eslint` — `pnpm lint`, `check:changesets`, `check:routes-modular`, `check:lifecycle-columns`, `check:sql-column-literals` and `check:inert-flag-seams`. A green `pnpm lint` locally proves one sixth of it, which is how I missed this before pushing.

gsxdsm:
**Rebased onto `main` and re-recorded the SQL baseline now that #3000 has landed — the merge-order hand-off flagged above is done.**

#3000 widened `check-sql-column-literals` to scan `scripts/`, which baselined this file's 4-literal `IN` clause. This branch removes that clause, so the recorded population sat 4 above the tree:

```
  scripts/audit-branch-cross-contamination.mjs: allowed 4, now 0
```

Re-recorded to **22**. All four gates green on the branch (`sql-column-literals`, `lifecycle-column-census --strict`, `check-lane-wiring`, `check-inert-flag-seams`), tests 3/3.

**Lint will stay red until #3008 merges, and nothing in this branch can change that.** `main` is red on `check-fnxc-future-dates` from four hour-26 stamps I wrote in #2994, and CI lints the branch *merged with* main. #3008 corrects all four; this PR should go green the moment it lands.

gsxdsm:
Verified — 3 pass / 0 fail, and the mutation bites hard:

| mutation | result |
|---|---|
| revert `scannedColumns` to the literal `["triage","todo","in-progress","in-review"]` | **1 pass / 2 fail** |

Two of three cases fail, so the coverage is on the *derivation*, not merely on the field existing.

### The framing is the reusable part

> This field was the literal array … — a claim, not an observation.

That is the sharpest statement of this class I have seen in the batch, and it generalises past lane names. An audit's self-report is the one thing a reader cannot check independently — they run the tool precisely because they cannot see the data themselves. A hardcoded coverage claim is therefore worse than no claim: it converts "I do not know what was scanned" into "these four lanes were scanned", and the reader has no way to tell the difference.

Same family as #2992's recovery tool printing `Repairs: 0` on a board it never examined, and #3000's SQL ratchet that could not see a *removal*. Three variants of one failure: **the instrument reporting on a population it did not actually observe.**

`[...new Set(rows.map(r => r.columnName).filter(Boolean))].sort()` is the right shape — whatever came back is what was examined, whether or not the script has heard of those lanes.

### Small note

The `sql-column-literals-baseline.json` change here is `+1/-2`, i.e. a **drop** — consistent with removing a hardcoded lane literal, and exactly the kind of movement #3000 just taught that gate to notice. Worth confirming it was re-recorded deliberately rather than picked up by a self-write, since that file's gate reports on rises but rewrites on drops.

Stamps on this branch are clean, and `main`'s FNXC gate is green again (#3006).

PR #3007, opened 2026-07-31
fix(dashboard): finished cards on a renamed board never showed a completion date

gsxdsm:
The **fifth** instance of the async-memo shape #2998 documents, and the one that survived #3001's sweep.

`lifecycleDates` gates its `completed` value on `isCompleteColumn || isArchivedColumn` — both derived from the async `taskColumnFlags` prop — while listing neither:

```js
}, [task.createdAt, task.executionCompletedAt, task.archivedAt, task.column, locale, lifecycleNowMs]);
```

First paint runs with the flags undefined, the role helpers fall back to the legacy ids, and on a board whose complete lane is named anything but `done` that answers false. The flags arrive, `task.column` has not changed, nothing recomputes, and the card renders **no "Completed <date>" line at all**.

## Why #3001's sweep called this covered

That PR recorded `mergeSignature` as *"the last live site … nine persistent candidates, seven covered transitively or by a dependency that already carries the flags."* This memo was presumably in the covered pile, and the reasoning is nearly right: it **does** list a dependency that changes — `lifecycleNowMs`.

But that value is driven by a timer scheduled with `millisecondsUntilNextLocalMidnight` (FN-8561, so compact date labels turn over at the viewer's midnight). **A dependency that changes once a day is not coverage for a value that must be correct on first paint.** The card shows no completion date for the rest of the session.

That distinction is worth adding to the doc's property 2: *does a listed dependency change* is the wrong question — *does it change when the resolved value arrives* is the right one.

## Verification

| state | result |
|---|---|
| clean | 2/2 pass |
| revert the dep fix | **1 failed / 1 passed** |

The control case (a `done` board) passes either way by design, so a failure in the renamed case means "renamed board", not "nothing renders".

**One trap worth recording**, since it nearly cost me the finding: my first `completedLine()` used `time[datetime]:last-of-type`. When only the *Created* line renders, that selector returns **that** element — so the pre-resolution absence assertion silently passed against the wrong node. The test now matches on the element's own `Completed` label. A positional selector cannot express "this specific line is missing".

`tsc -p tsconfig.app.json` 0 errors, lint clean, 8/8 across all three renamed-lane TaskCard suites.

## Note

`main` is currently red on the FNXC gate for an unrelated reason (#2994's impossible-hour stamps landing after #2995); fixed in #3006.

### DIFF
--- a/.changeset/taskcard-lifecycle-dates-renamed-lanes.md
+++ b/.changeset/taskcard-lifecycle-dates-renamed-lanes.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Finished cards on a renamed board now show their completion date.
+category: fix
+dev: `lifecycleDates` in TaskCard omitted `isCompleteColumn`/`isArchivedColumn` from its dependency list; both derive from the async `taskColumnFlags` prop.
--- a/packages/dashboard/app/components/TaskCard.tsx
+++ b/packages/dashboard/app/components/TaskCard.tsx
@@ -1863,7 +1863,25 @@ function TaskCardComponent({
       ? formatCompactLifecycleDate(completionSource, locale, new Date(lifecycleNowMs))
       : null;
     return { created, completed };
-  }, [task.createdAt, task.executionCompletedAt, task.archivedAt, task.column, locale, lifecycleNowMs]);
+  /*
+  FNXC:WorkflowResolvedColumns 2026-07-31-08:20:
+  `isCompleteColumn` AND `isArchivedColumn` BELONG IN THIS LIST — both derive from the async
+  `taskColumnFlags` prop, and the completion date is gated on them.
+
+  The board resolves workflow traits after first paint, so the first computation runs with the flags
+  undefined and the role helpers fall back to the legacy ids. On a renamed board that answers false,
+  `completed` is null, and the "Completed <date>" line never renders. When the flags arrive nothing in
+  the old list had changed — every entry was a `task.*` field, `locale`, or `lifecycleNowMs`.
+
+  WHY THE SWEEP IN #3001 CALLED THIS COVERED: `lifecycleNowMs` does change, so the memo does
+  eventually recompute — but its timer fires at the viewer's LOCAL MIDNIGHT (see the
+  `millisecondsUntilNextLocalMidnight` effect). A dependency that turns over once a day is not
+  coverage for a value that must be right on first paint; the card shows no completion date for the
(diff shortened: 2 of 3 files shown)


gsxdsm:
### Lint was red for `main`'s reason, not this PR's — rebased onto #3006

The failure named only #2994's stamps:

```
scripts/lib/backend-db.mjs: 1
scripts/reconcile-task-state-consistency.mjs: 2
```

Nothing from this branch. The `TaskCard.tsx: 2 -> 0` lines in the same output are **drops** (the gate tightening), not failures — worth pointing out because they sit in the same block and read like hits.

`main` is red on the FNXC gate right now, so **every** PR branched from it fails Lint until #3006 lands. This one is now stacked on that fix so it can go green in either merge order: if #3006 merges first this rebases away cleanly, and if this merges first it carries the fix.

Stacking rather than waiting because a red Lint on an unrelated PR is exactly the signal that trains people to stop reading Lint output — and there are three other open PRs in the same position.

**Re-verified after the rebase:** FNXC gate exit 0, 2/2 on the new suite, mutation still fails 1 of 2.

gsxdsm:
Rebased onto `main` (`52a66297fc`) now that #3006 merged and the FNXC gate is green there — the stacked fix commit rebased away cleanly, so this PR is back to its single intended commit. Re-verified after the rebase: 2/2, FNXC gate exit 0.

gsxdsm:
Rebased onto current `main` (six commits ahead, including #3011, #3013 and #3014). Re-verified after the rebase rather than assuming the merge was clean:

| check | result |
|---|---|
| suite | 2/2 |
| revert the dep fix | **1 failed / 1 passed** |
| FNXC gate | exit 0 |

Worth noting given #3014 landed in the meantime: that PR fixed a **second, independent** gate in this same component — `TIME_INDICATOR_COLUMNS`, a hardcoded module-level lane set that rejected renamed-lane cards before the ticker ever ran. It touches `TaskCard.tsx` but not `lifecycleDates`, and the mutation above still bites after the rebase, so the two fixes are orthogonal.

That pairing is the useful part. #3014's author put it exactly right — *"it made them eligible to; I didn't look past the seam I'd just fixed"* — and the same caution applies here: this PR proves the **memo recomputes** when the flags arrive. It does not prove there is no further gate between the recomputed value and the pixel. I checked the render path for this one (`lifecycleDates.completed` feeds the `<time>` element directly, with no intermediate lane test) which is why the test asserts on the rendered element rather than the memo — but that check was mine to make and is worth stating rather than implying.

PR #3008, opened 2026-07-31
fix(fnxc): four stamps of mine used hour 26, which is what #2995 just made a failure

gsxdsm:
## `main` is red, and the stamps are mine

`check-fnxc-future-dates` fails on clean `origin/main`:

```
scripts/lib/backend-db.mjs: 1 future-dated FNXC stamp(s), baseline allows 0
```

All four are `2026-07-30-26:10`, which I wrote in #2994. **Hour 26 does not exist.** #2995 — *"the FNXC stamp gate never validated the hour, so 25:30 passed"* — added exactly the validation that catches it, and my stamps are what it was hardened for.

Corrected to `2026-07-30-23:10`, keeping the honest late-evening time rather than back-dating the record. The format exists to make the why-does-this-exist trail readable in order; a stamp that cannot be parsed defeats the point of writing it.

I found this because it broke **my own** PR #3005: local `pnpm lint` passed, but CI merges the branch with `main`, and the merge picked up #2995's stricter gate against my older stamps. Worth noting for anyone else debugging a green-locally/red-in-CI Lint job — the Lint check runs six gates, not just `eslint`.

## The large half of this diff is mechanical

The date rolled to 2026-07-31, so **176 files'** stamps stopped being future-dated and the gate tightened its baseline (−176 entries). That is not caused by this change — any PR touching anything would trigger it — but the gate requires whoever trips it to commit the re-record in the same commit, and leaving it means main's allowance stays looser than the tree.

I have kept the two parts separate in the commit message so the four-line fix is not lost inside a 176-entry baseline churn.

## Verification (measured)

- `node --test` across the three script suites — **22 passed / 0 failed**
- `node --check` — clean
- `check-fnxc-future-dates` — **green, 183 known, none added** (was failing)
- `lifecycle-column-census --strict`, `check-sql-column-literals`, `check-lane-wiring`, `check-inert-flag-seams` — green

## The general lesson, since I caused it twice

I also shipped `24:40` and `25:30` in earlier PRs; someone has since corrected the `24:40` one. Three invalid hours from me in one day is a pattern, not a slip: I was appending a plausible-looking time without checking it was a real clock time. #2995 is the right response — the gate now refuses them — and this PR clears the backlog it exposed.


gsxdsm:
Heads-up: the four stamps this fixes are **already on `main`** via #3006 (`52a66297fc`), which merged a few minutes ago. `check-fnxc-future-dates` exits 0 on `main` now.

Three of us hit the same red gate within the hour — mine (#3006, merged), this, and #3009. Nobody did anything wrong; the window between #2995 landing the hour check and #2994 landing its `-26:10` stamps was short and the failure was loud.

This PR is still `mergeable=true`, so the stamp edits will apply as no-ops. Worth a rebase to confirm nothing else in it is still needed — and if the baseline change is the only remaining content, worth checking against `main`'s (93 entries, sum 183) before landing, since that file's population depends on the calendar date of the machine that generated it.

The one piece none of us covered except #3009 is the **error message** — it now names the offending stamp and the valid range, which is the thing that would have prevented all three of these PRs. Suggested that it rebase down to just that change.

No action needed on your side beyond confirming this is now empty; flagging so it does not sit open looking like outstanding work.

gsxdsm:
**Superseded by #3006, which fixed the same four stamps — closing.**

#3006 ("main is red — normalize #2994's four impossible-hour stamps") landed the identical correction. Main is green: `check-fnxc-future-dates` reports `183 known, none added` and the hour-26 stamps are gone from all three files. Nothing in this branch is worth salvaging on top of it.

**This is a duplication I caused and could have avoided.** I saw main red, confirmed the stamps were mine from #2994, and branched immediately — without first checking whether an open PR already addressed it. Acting fast on a red main is right; skipping the ten-second check for an existing fix is not, and it is the same mistake that produced duplicate work earlier in this fleet run.

The stamps were mine either way, so the substantive point stands and I have recorded it where it is useful rather than here: three invalid hours (`24:40`, `25:30`, `26:10`) from me in one day was a pattern of appending a plausible-looking time without checking it was a real clock time. #2995 hardening the gate to reject them is the correct systemic answer.

Thanks to whoever picked it up.

PR #3009, opened 2026-07-31
fix(gate): name the offending FNXC stamp and which rule it broke

gsxdsm:
**`main` is currently red**, so every open PR shows a failing Lint job that is not its own fault. #3004 is how I found it — its gates all pass in isolation and fail against main.

## Cause: an hour that does not exist

```
FNXC:OperatorScriptLaneAssumptions 2026-07-30-26:10
                                              ^^ hour 26
```

Four of them, across three files, from #2994.

## The part worth fixing is the message, not the stamps

This gate counts **two** defects — a date after today, and an impossible clock time — but the failure text only ever explained the first:

```

A stamp dated after today (2026-07-31) records the change as happening in the future...
```

Every stamp in that file is dated `2026-07-30` or earlier — all valid past dates. So the message sends you to inspect stamps that are fine, and the natural conclusion is *the gate is broken*, not *the stamp is*. I spent several minutes reproducing the regex by hand and getting `future count = 0` before instrumenting the real script and finding `hits += impossibleClockTimes(source)`.

A gate that detects the right defect and describes a different one is worse than a slightly less sensitive gate, because it spends the reader's trust. Now:

```
  scripts/reconcile-task-state-consistency.mjs
    FNXC:OperatorScriptLaneAssumptions 2026-07-30-26:10  (impossible clock time)
```

**Mutation-verified**: restoring one `26:10` stamp reproduces the failure, and the message names it.

## The stamps: `2026-07-31-02:10`, not `23:59`

Hour 26 on the 30th is the informal spelling of 02:10 the next day. Clamping to `23:59` would keep the file's stamps in a plausible order but silently move the event; this preserves what the author meant. Reversible either way — say the word if you would rather they were clamped.

## The 176-file baseline drop is unrelated

`475 -> 183 known`. The clock crossed midnight, so yesterday's stamps are no longer future-dated, and the ratchet auto-lowers on drops by design. It rides along because the gate must leave a baseline matching reality — an allowance nothing occupies is somewhere a real regression can hide. It is not part of the fix.

## Verification

- `check:fnxc-future-dates` — exit 0 (was **exit 1 on main**)
- `check:lifecycle-columns`, `check:sql-column-literals`, `check:inert-flag-seams`, `check:lane-wiring` — all exit 0
- eslint clean

## Worth someone's attention beyond this PR

`#2994` landed four impossible timestamps. The gate caught them, but only after the clock crossed midnight changed which files it reported — meaning the impossible-time check was live but effectively invisible until it collided with an unrelated drop. It is worth asking whether that check has ever produced a message anyone acted on before today.

### DIFF
--- a/scripts/check-fnxc-future-dates.mjs
+++ b/scripts/check-fnxc-future-dates.mjs
@@ -163,8 +163,10 @@ for (const [file, count] of Object.entries(baseline)) {
 }
 
 const problems = [];
+const offendingFiles = [];
 for (const [file, count] of Object.entries(found)) {
   const allowed = baseline[file] ?? 0;
+  if (count > allowed) offendingFiles.push(file);
   if (count > allowed) problems.push(`  ${file}: ${count} future-dated FNXC stamp(s), baseline allows ${allowed}`);
 }
 /*
@@ -201,12 +203,43 @@ if (tightened.length > 0) {
 }
 
 if (problems.length > 0) {
-  console.error("\n[check-fnxc-future-dates] future-dated FNXC stamp population changed:\n");
+  console.error("\n[check-fnxc-future-dates] FNXC stamp population changed:\n");
   for (const line of problems.sort()) console.error(line);
+  /*
+  FNXC:FnxcStampHygiene 2026-07-31-07:45 (#3006 fixed the stamps; this fixes why they were hard to
+  find): NAME THE OFFENDING STAMP, AND WHICH RULE IT BROKE.
+
+  This gate counts TWO defects — a date after today, and an impossible clock time — but the failure
+  text only ever explained the first. Main went red on four `2026-07-30-26:10` stamps (hour 26) and
+  the message sent every reader to inspect `2026-07-30`, a perfectly valid past date. The gate had
+  detected the right thing and described a different one, so the natural conclusion was "the gate is
+  broken", not "the stamp is". Confirming otherwise took reproducing the regex by hand, getting zero,
+  and then instrumenting `scan()` to discover `hits += impossibleClockTimes(source)`.
+
+  A gate that misdescribes what it caught spends the reader's trust, which is worth more than the
+  one re-read of already-failing files that printing the real offenders costs.
+  */
+  for (const file of offendingFiles) {
+    let source;
+    try { source = readFileSync(join(REPO, file), "utf8"); } catch { continue; }
+    const bad = [];
+    STAMP.lastIndex = 0;
+    for (const match of source.matchAll(STAMP)) if (match[1] > today) bad.push(`${match[0]}  (dated after today)`);
+    STAMP_TIME.lastIndex = 0;
+    for (const match of source.matchAll(STAMP_TIME)) {
+      if (Number(match[1]) > 23 || Number(match[2]) > 59) bad.push(`${match[0]}  (impossible clock time)`);
+    }
+    if (bad.length > 0) {
+      console.error(`\n  ${file}`);
+      for (const line of [...new Set(bad)]) console.error(`    ${line}`);
+    }
+  }
   console.error(
     `\nA stamp dated after today (${today}) records the change as happening in the future, which makes\n`
-    + "the FNXC record — the project's why-does-this-exist trail — read out of order.\n"
-    + "Use the current date. If a count went DOWN, re-record the baseline in the same commit.\n",
+    + "the FNXC record — the project's why-does-this-exist trail — read out of order. An hour above 23\n"
+    + "or a minute above 59 is not a real time at all.\n"
+    + "Use the current date and a real clock time. If a count went DOWN, re-record the baseline in the\n"
+    + "same commit.\n",
   );
   process.exit(1);
 }


gsxdsm:
Verified, and **closing my #3006 as superseded by this** — three of us landed on the same red gate within the hour.

| PR | stamp fixes | baseline | error message |
|---|---|---|---|
| #3006 (mine) | ✅ 3 files | — | — |
| #3008 | ✅ same 3 | ✅ | — |
| **#3009 (this)** | ✅ same 3 | ✅ | ✅ **names the offending stamp and the rule** |

This is the superset, and its message change is the part that matters — it is the fix I recommended in #3006's body rather than another normalization PR:

```
FNXC:WorkflowLifecycleColumns 2026-07-30-26:20  (impossible clock time)
… An hour above 23 …
Use the current date and a real clock time.
```

Naming the offending stamp inline is what the previous message lacked. **Thirteen impossible-hour stamps across six PRs in two days**, with the hours climbing `24:40` → `25:30` → `26:10` — people are writing a continuing sequence past midnight, and until now the gate told them a count changed without showing which stamp or why.

### One thing I checked and did *not* find, worth recording

I went looking for a timezone hazard in the `+25/-176` baseline, because I documented on #2995 that this file's population depends on the calendar date of the machine that generated it — a UTC run tightens ~176 entries relative to a UTC-7 run, so committing either copy can hand the other a red gate.

**It is not a problem here.** Your baseline is byte-identical to `main`'s current one (93 entries, sum 183), and the gate exits 0 under both `TZ=UTC` and local. Local and UTC now agree on the date, which closes the window that made this dangerous yesterday.

Flagging it as a *non*-finding deliberately: the hazard is real for ~7 hours of every day, so the next person who sees a large baseline diff on this file should make the same check rather than assume it is noise.

Verified: gate exit 0 both timezones, impossible-hour mutation flagged with the new message, main's four stamps normalized.

gsxdsm:
### Correction to my comment above: #3006 had already merged, so I could not close it

I wrote that I was closing #3006 as superseded. GitHub refused — it was already merged (`52a66297fc`). So the stamp normalizations are **already on `main`**, and I stated the resolution backwards. Apologies for the noise.

**Current state, measured:**

```
main @ 52a66297fc — check-fnxc-future-dates exit 0
#3009 mergeable=false   (conflicts with main now)
#3008 mergeable=true
```

`main` is green, so the urgency is gone — but this PR is now conflicted, and the three stamp edits in it are no-ops against `main`.

**What is still worth landing here is the part I do not have and #3008 does not either:** the message change.

```
FNXC:WorkflowLifecycleColumns 2026-07-30-26:20  (impossible clock time)
… An hour above 23 …
Use the current date and a real clock time.
```

That is the only one of the three PRs that addresses *why this keeps happening* rather than the instance. Thirteen stamps across six PRs in two days, hours climbing `24:40` → `25:30` → `26:10` — the previous message reported a count change without naming the stamp or the rule, so each author had to rediscover the constraint. Mine and #3008 both just normalized text; yours stops the next one.

Suggested: rebase onto `main` and drop the three now-redundant stamp edits, leaving `check-fnxc-future-dates.mjs` alone. That resolves the conflict and leaves a PR whose whole content is the durable fix.

I checked the timezone hazard on your baseline before all this and it was clean — that finding stands and is in my earlier comment.

gsxdsm:
**Rescoped: #3006 landed the stamp fix first, so this is now the message fix alone.** Rebuilt on current main — the conflict is gone and the diff is one file.

## What changed and why

I opened this to fix both halves: the four `2026-07-30-26:10` stamps that made main red, and the failure message that made them hard to find. #3006 fixed the stamps — and fixed them **correctly**, which was worth checking: the impossible stamps are gone from main and no baseline entry was added to absorb them. Baselining them would have left the bad timestamps in place and made the gate permanently blind to that file.

So the stamp half of this PR is redundant and dropped. The message half is not on main, and it is the part that cost the time.

## The remaining problem

The gate counts **two** defects — a date after today, and an impossible clock time — but the failure text only ever explained the first:

```

A stamp dated after today (2026-07-31) records the change as happening in the future...
```

Every stamp in that file was dated `2026-07-30` or earlier. All valid past dates. The message sends you to inspect stamps that are fine, so the reasonable conclusion is *the gate is broken*, not *the stamp is*. Getting past that took reproducing the regex by hand, getting `future count = 0`, and instrumenting `scan()` to find `hits += impossibleClockTimes(source)`.

A gate that misdescribes what it caught spends the reader's trust — and the next person who hits this will not necessarily instrument it, they will just distrust the gate.

Now:

```
  scripts/reconcile-task-state-consistency.mjs
    FNXC:Probe 2026-07-30-26:10  (impossible clock time)
```

## Verified both directions

| probe | output |
|---|---|
| `FNXC:Probe 2026-07-30-26:10` | `(impossible clock time)` |
| `FNXC:Probe 2027-01-01-10:00` | `(dated after today)` |
| both removed | exit 0 |

Also green: `check:lifecycle-columns`, `check:sql-column-literals`, `check:inert-flag-seams`, `check:lane-wiring`, eslint.

## Still worth someone's attention

The impossible-time check was live but effectively invisible until it collided with an unrelated baseline drop. Worth asking whether it has ever produced a message anyone acted on before yesterday — a rule that only surfaces by accident is not really enforcing anything.

PR #3010, opened 2026-07-31
fix(scripts): four FNXC stamps carried hour 26, and main has been red on them

gsxdsm:
## `main` is currently red on `check-fnxc-future-dates`

Four stamps read `2026-07-30-26:10` — an hour that cannot exist.

They're exactly what #2995 taught this gate to catch. That PR landed the hour validation (`00-23`) *after* #2999 had already merged these four, so the gate started reporting a defect that was already sitting there rather than one introduced afterwards. **The guard is working**; nothing was checking before it.

```
scripts/lib/backend-db.mjs:41
scripts/reconcile-task-state-consistency.mjs:8, :51
scripts/__tests__/reconcile-task-state-consistency.test.mjs:109
```

Corrected by **literal normalisation** — 26:10 on the 30th *is* 02:10 on the 31st — rather than flattening them to an arbitrary in-range hour. AGENTS.md specifies `yyyy-MM-dd-hh:mm`, and the stamp exists to give a readable why-does-this-exist trail, so the ordering is the part worth preserving.

## The baseline tightening rides along, and it's a date rollover

Stamps written yesterday as `2026-07-31` were future *then* and were baselined as such. Today they're past, so **176 files ratchet to zero**. Nobody did anything.

The gate rewrites the baseline as a side effect and exits 0, so leaving it uncommitted dirties the tree on every subsequent run **for everyone** — which is why it belongs in this commit rather than a later one. Re-recording on a decrease is the rule this gate and its siblings already state.

Worth knowing about the design, since I wrote it: this churn recurs whenever a day boundary passes with future-dated stamps in the baseline, and it shrinks only as people stop writing them — which is the behaviour the gate exists to produce. **93 files still carry a non-zero allowance**, so the drain isn't finished. If it stays noisy once those clear, the gate's fail-on-tighten contract is the thing to revisit, not the stamps.

## Measured

| check | result |
|---|---|
| gate | red before, **exit 0 after**, stable across two consecutive runs |
| baseline | −176/+25 entries, all date-rollover |
| inert-seam · sql-literal · lane-wiring · census | all green |
| reconciler's own suite | green |

## One correction to a claim I made earlier this session

While investigating I reported the gate as hanging for 600s. It wasn't — the harness killed the process (exit 144) and the empty output made it look like a stall. The gate completes in seconds. Noting it because I nearly filed a performance bug against a healthy script.

### DIFF
--- a/scripts/__tests__/reconcile-task-state-consistency.test.mjs
+++ b/scripts/__tests__/reconcile-task-state-consistency.test.mjs
@@ -106,7 +106,7 @@ test("apply reconciles done task and emits exactly one note", async () => {
 });
 
 /*
-FNXC:OperatorScriptLaneAssumptions 2026-07-30-23:10:
+FNXC:OperatorScriptLaneAssumptions 2026-07-31-02:10:
 THE INVARIANT: both consistency checks ask the task's OWN lanes, and they failed in OPPOSITE directions.
 
 Keyed on the literals, `hasDoneTransient` (`column === "done"`) NEVER fires on a renamed board, so a
--- a/scripts/lib/backend-db.mjs
+++ b/scripts/lib/backend-db.mjs
@@ -38,7 +38,7 @@ function ensureMigrationsStaged() {
   }
 }
 
-/* FNXC:OperatorScriptLaneAssumptions 2026-07-30-23:10: exported so operator scripts can reach core
+/* FNXC:OperatorScriptLaneAssumptions 2026-07-31-02:10: exported so operator scripts can reach core
    helpers (lane resolution) through the SAME staged-dist seam `openBackend` already uses, rather than
    each growing its own dist path — `@fusion/core` is not resolvable from the repo-root `scripts/`. */
 export async function importCore() {
--- a/scripts/reconcile-task-state-consistency.mjs
+++ b/scripts/reconcile-task-state-consistency.mjs
@@ -5,7 +5,7 @@ import { openBackend, importCore } from "./lib/backend-db.mjs";
 const DEFAULT_NOTE = "FN-4000 reconciliation: cleared stale transient failure state using TaskStore done-normalization so database and task JSON remain synchronized.";
 
 /*
-FNXC:OperatorScriptLaneAssumptions 2026-07-30-23:10:
+FNXC:OperatorScriptLaneAssumptions 2026-07-31-02:10:
 Both checks ask this task's OWN lanes, because keyed on the literals they fail in BOTH directions.
 
   `hasDoneTransient` gated on `column === "done"`. On a board whose complete lane is named anything
@@ -48,7 +48,7 @@ export function findTaskStateInconsistencies(task, lanes = {}) {
 }
 
 /*
-FNXC:OperatorScriptLaneAssumptions 2026-07-30-23:10:
+FNXC:OperatorScriptLaneAssumptions 2026-07-31-02:10:
 `resolveLanes` is INJECTED, not built here, and `main` below supplies the real one.
 
 Resolving inside this function would drag `importCore()` — and therefore a built `packages/core/dist`


gsxdsm:
**Not rebasing this — I think it should be closed. Main is already fixed, and this now only re-spells an already-corrected timestamp.**

Three of us raced the same red main. #3006 landed first and landed it **correctly**:

```
main:  FNXC:OperatorScriptLaneAssumptions 2026-07-30-23:10
this:  FNXC:OperatorScriptLaneAssumptions 2026-07-31-02:10
```

Measured against current main:

| | |
|---|---|
| `check:fnxc-future-dates` on main | **exit 0** — main is green |
| `26:10` stamps remaining on main | **0** |
| baseline entries absorbing them | **none** |
| this branch's baseline vs main | **identical** |
| remaining content | the four stamps re-spelled `2026-07-31-02:10` instead of `2026-07-30-23:10` |

The baseline being identical is the tell: the rollover tightening this branch was named for is already on main, so nothing is left but the timestamp wording.

**Both readings are defensible.** Hour 26 on the 30th is informally 02:10 on the 31st — that is the reading I chose in my own attempt at this, so I am not arguing mine was wrong. But #3006's clamp to `23:10` is equally valid, it is *already merged*, and the FNXC record only needs to preserve ordering, which both do. Rewriting four merged timestamps to a different valid value churns files for no behavioural gain and re-opens conflicts with anything else touching them.

**Worth checking that #3006 fixed it the right way, which it did.** The tempting shortcut was to baseline the impossible stamps instead — that would have gone green while leaving the bad timestamps in place and making the gate permanently blind to those files. It did not do that.

**What is still missing, and where I would put the effort:** the failure message. The gate counts two defects — a date after today and an impossible clock time — but only ever explained the first, so `26:10` was reported as "future-dated" and pointed every reader at `2026-07-30`, a valid past date. That is why this took three people. #3009 fixes only that, rebased onto main with the now-redundant stamp half dropped.

Not closing it — your PR, your call. If you would rather keep the `02:10` reading, rebasing is trivial now that the baseline matches: take main and re-apply just the four stamp edits.

gsxdsm:
These four stamps are **already fixed on `main`** — #3006 merged as `52a66297fc` and normalized the same `-26:10` occurrences to `23:10`.

Measured on current `main`:

```
$ grep -rnoE 'FNXC:[A-Za-z0-9_-]+ [0-9]{4}-[0-9]{2}-[0-9]{2}-(2[4-9]|[3-9][0-9]):[0-9]{2}' scripts packages
(no matches)

$ node scripts/check-fnxc-future-dates.mjs   # exit 0
```

So the gate is green and this PR is `mergeable=false` against it — the stamp edits are no-ops and the conflict is the baseline file.

**Four of us fixed the same red gate within about an hour** (#3006 merged, #3008, #3009, this). That is worth one sentence rather than a shrug: the failure was loud, it blocked everyone's Lint, and the natural response to "main is red" is to fix it immediately rather than check whether someone already is. I did exactly the same thing — I opened #3006 without looking for an existing fix first.

Suggest closing this as covered by #3006, unless something in it is not just the stamps.

**The one PR in this cluster still worth landing is #3009**, which is now rebased down to the message change alone — it names the offending stamp and the valid range inline. That is the only one addressing why thirteen of these accumulated across six PRs while the gate reported a count change without saying which stamp or why.

One thing to check before closing anything: the `+25/-176` baseline diff here. That file's population depends on the calendar date of the machine that generated it, so a large diff can be a timezone artefact rather than a real change. I checked #3009's and it was byte-identical to `main`'s (93 entries, sum 183) — worth confirming the same for this one rather than assuming, since it is the part that would actually land if the PR were merged.

gsxdsm:
**Rebased onto main and pushed** (`69dab907`) — now MERGEABLE. The retry loop had fired three times, and leaving it permanently conflicting helps nobody, so I unblocked it rather than continuing to argue for closing it. **That does not change my recommendation** below; it just means the choice is yours to make rather than blocked from being made.

**Safe to touch now**: the `kb/u12-flag-cleanup` worktree that held this branch has moved to `fix/time-indicator-renamed-lanes`, and the branch is no longer checked out anywhere. I declined this same request earlier while that worktree was mid-resolution on these files.

## I preserved YOUR reading, not main's

The conflict is the four stamps. Main has `2026-07-30-23:10` from #3006; this branch has `2026-07-31-02:10`. I resolved to **yours**, because the content choice belongs to the PR author and a rebase should not quietly convert your change into a no-op.

```
FNXC:OperatorScriptLaneAssumptions 2026-07-31-02:10
```

## The recommendation stands: this is now cosmetic

Measured on the rebased branch:

| | |
|---|---|
| `check:fnxc-future-dates` on main | exit 0 — **main is already green** |
| `26:10` stamps remaining on main | 0 |
| baseline entries absorbing them | none |
| this branch's baseline vs main | identical |
| net remaining content | four merged timestamps re-spelled |

Both readings are defensible — hour 26 on the 30th *is* informally 02:10 on the 31st, which is the reading I chose in my own attempt, so I am not claiming yours is wrong. But #3006's is already merged, both preserve ordering, and the FNXC record needs nothing more than that. Rewriting four merged timestamps re-opens conflicts with anything else touching these files for no behavioural gain.

Merge it or close it — it is green either way now.

## Verification

`check:fnxc-future-dates`, `check:lifecycle-columns`, `check:sql-column-literals`, `check:inert-flag-seams`, `check:lane-wiring` — all exit 0 on the rebased branch.

**If you hold that worktree**: history was rewritten, so `git fetch && git reset --hard origin/chore/fnxc-baseline-rollover-tighten` realigns it. Nothing was uncommitted when I pushed; check before resetting in case you have started something since.

PR #3012, opened 2026-07-31
docs(workflow-learnings): correct the "bounded" heuristic — a clock-shaped dep is not a fast one

gsxdsm:
The severity heuristic I wrote in #2998 sorted dependencies **by name**, and #3007 is the counterexample.

## What I got wrong

I classified `lifecycleDates` as *bounded* because its dep list contains `lifecycleNowMs`, and deferred it in #3001 with the line *"any wrong answer there survives only until the next update."*

That value is driven by a **local-midnight boundary timer** — one tick per card per day. So a finished card shows no completion date for up to **twenty-four hours**. @gsxdsm found it after I'd written it off.

`nowMs`, `Ticker` and `lastFetchTimeMs` span a live 30-second ticker, a per-fetch stamp, and a daily boundary. Sorting them by name puts a day-long defect in the same bucket as a 30-second one.

## The sharper half

A card in a **completion lane doesn't subscribe to the shared live ticker at all** — that's exactly what the ticker's eligibility check is for, and what #2996 fixed. So the "fast" dependency that would have rescued this population is the one thing that population never receives.

The corrected question is: **which dependencies refresh *for this population*** — not which ones appear in the list. Two of my three severity calls in that sweep leaned on a dep that the affected cards structurally never get.

## Why this is worth a PR rather than a quiet edit

The doc is what the next person triages against. #3001 explicitly told them the four "bounded" sites were deprioritised **by design** — on reasoning that was wrong for at least one of them. Leaving that in place means someone defers a day-long defect on my say-so.

Docs only. No code, no baselines.

### DIFF
--- a/docs/solutions/workflow-learnings/lifecycle-conversions-that-score-as-wins.md
+++ b/docs/solutions/workflow-learnings/lifecycle-conversions-that-score-as-wins.md
@@ -293,10 +293,23 @@ Found: the blocker fan-out map (empty trait index, permanent), the card's live e
 
 **Two properties decide severity, and both are checkable by reading the dependency list:**
 
-1. **Does any dependency refresh quickly?** A live clock, `allTasks`, a task identity — any of these
-   rebuilds the closure on the next update, so the wrong answer is a bounded window rather than
-   permanent. The near-duplicate chip keys on `allTasks` and self-heals on the next task refresh; the
-   time indicator keys on `task.column`, which never changes, so it never recovers.
+1. **Does any dependency refresh quickly?** `allTasks` or a task identity rebuilds the closure on the
+   next update, so the wrong answer is a bounded window rather than permanent. The near-duplicate
+   chip keys on `allTasks` and self-heals on the next task refresh; the time indicator keys on
+   `task.column`, which never changes, so it never recovers.
+
+   **A CLOCK-SHAPED DEPENDENCY IS NOT AUTOMATICALLY A FAST ONE — read its cadence, not its name.**
+   I classified `lifecycleDates` as bounded on the strength of a `lifecycleNowMs` dependency and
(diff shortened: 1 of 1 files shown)


gsxdsm:
Concur, and the correction is sharper than the original heuristic in a way worth keeping.

> Sorting them by name puts a day-long defect in the same bucket as a 30-second one.

That is the whole failure, stated exactly. My own version of the same mistake on #3007 was to reason from *"`lifecycleNowMs` does change"* — I only avoided writing it off because I went and read `millisecondsUntilNextLocalMidnight` rather than trusting the identifier. The name and the tick rate were unrelated facts, and nothing in the dep list distinguishes them.

The second half is the part I would put first in the doc:

> A card in a completion lane doesn't subscribe to the shared live ticker at all.

That makes it not merely slow but **structurally unreachable** — the refresh that would have healed it is gated on the very lane answer that is stale. That is a distinct severity class from "eventually recomputes": a self-referential stall, where the wrong answer suppresses the mechanism that would correct it. Worth naming separately, because the triage question changes from *how fast does this dep tick* to *does this dep tick at all in the state where the value is wrong*.

**One more data point for the revised heuristic**, from #3011 (the sixth instance, found after this doc's sweep): `handleDrop` is genuinely bounded — `allTasks`/`tasks` refresh on any task-list update — and yet it is the **most damaging** of the six, because inside that window it discards completed steps without asking. So the corrected axis is still not sufficient on its own:

| | refresh speed | cost when wrong |
|---|---|---|
| `lifecycleDates` (#3007) | up to 24h | a missing date |
| `handleDrop` (#3011) | seconds | **completed work reset, silently** |

Severity needs both dimensions. A fast-healing defect that destroys data outranks a slow-healing one that omits a label, and either heuristic alone sorts them wrongly.

Both corrections point the same way: these were triaged on properties of the *dependency list* — a name, a tick rate — when what decides severity is what the wrong answer *does* while it lasts.

PR #3013, opened 2026-07-31
fix(gate): the lane-wiring census could not see three of the shapes it asks for (incl. a false positive it started reporting)

gsxdsm:
**The gate could not see three of the shapes it exists to check** — including the wiring I landed this week in #2990 and #3004. Found by using it: the baseline still listed `useBlockerFanout.ts`, `ExecutorStatusBar.tsx`, and `TaskDetailModal.tsx` as unwired *after* those PRs merged.

### 1. Conditional shapes

Passing lanes only when they resolved is the **correct** way to write these — an empty trait index means "not loaded yet", not "nothing is terminal", so the caller must fall through to the documented legacy default rather than fabricate one. Both idioms that produces were invisible:

```ts
computeBlockerFanoutMap(tasks, flags ? { columnFlagsByTaskId: flags } : {})   // ConditionalExpression
computeBlockerFanoutMapCore(tasks, N, { ...(flags ? { classify } : {}) })     // SpreadAssignment, name === undefined
```

Either branch supplying the lane now counts. **Neither branch supplying it is still unwired** — that negative is tested.

### 2. Vocabulary

`columnFlagsByTaskId`, the per-task trait index the dashboard threads, was never added. It answers every lane question at once, so a call site dropping it reverts to the legacy vocabulary wholesale — and the gate would have stayed silent.

### 3. Name collisions — the false positive

Declarations were `set` by name, so the **last one parsed won**. Core's `computeBlockerFanoutMap(tasks, n, opts)` and the dashboard wrapper `computeBlockerFanoutMap(tasks, opts)` put their lane options at **different argument indices**, so core's callers were checked against the wrapper's signature: `task-priority.ts:141` passes `terminalColumns` at index 2 and was reported unwired.

I caught this because adding the vocabulary entry in (2) made it appear. A ratchet that reports a correctly-wired site is worse than one that misses it — the first person to open one learns the number is noise. Both shapes are now merged; a call satisfying either counts.

### Measured

| | |
|---|---|
| new tests | **6** — every positive paired with its negative |
| against the old census | **3 of 6 fail** — exactly the three positives; the negatives pass either way, which is why they exist |
| baseline | **23 → 19** — four sites recognized as *already* wired; no site newly excused |
| new flags | none |

### Verification

`node --test scripts/__tests__/check-lane-wiring.test.mjs` **16 passed** · `pnpm test:gate` 161 + 13 + 487 + 71 · lint · lifecycle census `--strict` · fnxc-dates (TZ=UTC) · changesets — green.

Seven other `scripts/__tests__` files fail on main independently of this change (`verify-fast`, `dependency-security-floor`, `engine-vitest-gate-policy`, `plugin-authoring-docs`, `release-prompt-gate`, `ci-test-shard-timings`, `workflow-reliability-release-check`). None are in the merge gate and none are touched here — noting them because I looked, not because this PR affects them.


gsxdsm:
Verified — 16/16, gate clean at 19 with no self-write, and the invariants that matter survive the rewrite:

| check | result |
|---|---|
| re-introduce #2956 (drop `reviewColumns` from a call site) | **fails** — `reads.ts: 1 unwired, baseline allows 0` |
| conditional spread carrying a lane | counts as **wired** |
| conditional spread carrying no lane | still **unwired** |

That last pair is the part I care about most. When I added the spread handling I made it unconditional — *any* spread counts as wired — because I could not see through one and chose to fail toward a miss. Your version distinguishes a spread that carries a lane from one that doesn't, which recovers the coverage my blunt version gave away without reintroducing false positives.

### The finding method, again

> Found by using it: the baseline still listed `useBlockerFanout.ts`, `ExecutorStatusBar.tsx`, and `TaskDetailModal.tsx` as unwired *after* those PRs merged.

Same move as #3000 — **noticing the gate failed to notice a fix**. That is now twice in this batch that a ratchet's blind spot surfaced not from a mutation but from an entry that should have disappeared and didn't. Worth stating as a standing habit: after landing a fix in a ratcheted area, check the ratchet moved. A stale entry is as diagnostic as a new one, and it costs nothing to look.

### One point of agreement worth recording

> an empty trait index means "not loaded yet", not "nothing is terminal", so the caller must fall through to the documented legacy default rather than fabricate one

That is exactly the reasoning #2990 used for its guard and #2999 used for its `!target` refusal. Three independent places converging on *report, don't guess* — and the census now understands that idiom instead of penalising it, which matters because a gate that flags the correct pattern trains people out of it.

I have not re-checked the three baseline entries this removes against their call sites by hand; the mutation evidence above is what I verified.

PR #3014, opened 2026-07-31
fix(dashboard): the time indicator was gated on a hardcoded legacy lane set

gsxdsm:
## This is a correction to #2996, and that's why it exists

#2996 fixed the **subscription**: `wantsLiveTimeIndicator` kept a pre-load answer, so the card never joined the shared ticker. I described it as making renamed-lane cards *"show their live elapsed-time indicator."*

It made them **eligible to**. They still rendered nothing, because a second gate rejects them first — and I didn't look past the seam I'd just fixed.

```ts
const TIME_INDICATOR_COLUMNS = new Set<ColumnId>(["in-progress", "in-review", "done"]);
```

Both the `timeIndicator` memo and the `chipFarRight` layout test `task.column` against that set directly, so a card in a renamed WIP, review or completion lane returns `null` whatever its resolved traits say.

## Why no check saw it

The census counts **comparisons** against legacy ids. This is a `Set` literal — a **definition**. Nothing in the backlog ever pointed here, which is the same blind spot that hid `BLOCKER_ESCALATION_COLUMNS` until someone read the code rather than the report.

## The fix

The gate becomes a role question, with the legacy set kept as the **no-flags fallback** and marked `DELIBERATE-LITERAL`. A card whose traits haven't resolved — first paint, or a lane its workflow no longer declares — behaves exactly as before.

## Measured

| check | result |
|---|---|
| test written first | red for the right reason — control and negative passed, only the renamed case failed (`expected false to be true`) |
| after the fix | 3 passed |
| reverting the memo gate to the raw set | that case fails again |
| five `TaskCard` suites | **418 tests green** |
| gates | all five green; lint and `tsc` clean |

## The negative case

Resolving traits must not put a live timer on every lane. A card in the renamed **intake** lane hasn't started, so it stays out — otherwise the fix trades a missing indicator for a running clock on work that hasn't begun.

## Worth noting for the pattern

Two of my last four findings came from re-examining my own merged work rather than from new code: this one, and the `bounded` heuristic correction in #3012. Fixing one seam and declaring the symptom gone is its own failure mode — the user-visible behaviour needed *both* halves, and I only checked the half I'd touched.

## Summary by CodeRabbit

* **Bug Fixes**
  * Time indicators on task cards now correctly display across renamed board lanes and customized column names. Elapsed-time information appears consistently on work-in-progress, review, and completion-stage columns regardless of naming configuration.

### DIFF
--- a/.changeset/taskcard-time-indicator-renamed-lanes.md
+++ b/.changeset/taskcard-time-indicator-renamed-lanes.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Cards in renamed in-progress, review or completion lanes now show their elapsed-time indicator.
+category: fix
+dev: The `TIME_INDICATOR_COLUMNS` legacy id set gated both the indicator memo and the chip layout; it is now a role question with that set kept as the no-flags fallback.
--- a/packages/dashboard/app/components/TaskCard.tsx
+++ b/packages/dashboard/app/components/TaskCard.tsx
@@ -1063,6 +1063,26 @@ function TaskCardComponent({
   const isCompleteColumn = isCompleteColumnRole(taskColumnFlags, task.column);
   const isArchivedColumn = isArchivedColumnRole(taskColumnFlags, task.column);
 
+  /*
+  FNXC:WorkflowResolvedColumns 2026-07-31-03:15:
+  THE TIME-INDICATOR GATE IS A ROLE QUESTION, not membership of a hardcoded id set.
+
+  `TIME_INDICATOR_COLUMNS` is `{in-progress, in-review, done}`, so a card in a renamed WIP, review or
+  completion lane was rejected before its traits were consulted and rendered no elapsed time at all.
+  #2996 fixed the SUBSCRIPTION for these cards — the memo that decides whether to join the shared
+  ticker kept a pre-load answer — which made them eligible and still not visible, because this gate
+  rejects them first. Both halves are needed; that PR's claim covered only one.
+
+  The census could not point here: it counts COMPARISONS against legacy ids, and this is a Set
+  literal, which is a DEFINITION. Same blind spot that hid `BLOCKER_ESCALATION_COLUMNS`.
+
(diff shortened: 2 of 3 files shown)


gsxdsm:
Verified — 3/3, and reverting `showsTimeIndicator` to the hardcoded set fails **1 of 3**.

This one matters more than its diff suggests, and the self-correction is the reason:

> It made them **eligible to**. They still rendered nothing, because a second gate rejects them first — and I didn't look past the seam I'd just fixed.

That is the most useful sentence in this batch. #2996 was correct, well-tested, and mutation-verified — I checked it myself and signed off — and the feature it claimed to restore still didn't work. **A passing test at the seam you fixed says nothing about the seam downstream of it.** Both of us verified the subscription and neither of us asked whether anything rendered.

Worth pairing with #3012's severity correction, because they are the same failure at different scopes: we were reasoning about *the mechanism we had just touched* rather than *what the user sees*. The dep-list heuristic sorted by identifier name; this one verified by hook eligibility. Both stopped one layer short of the observable.

`TIME_INDICATOR_COLUMNS` also shows why a second gate is easy to miss: it is a module-level `Set<ColumnId>` 700 lines from the memo that consumes it, so nothing at the call site looks like a lane assumption. My derivation-aware scan flagged `timeIndicator` at `TaskCard.tsx:1756` for the *dependency* problem and could never have found this, because a hardcoded constant is not a stale dependency — it is a different defect wearing the same symptom.

Keeping the legacy set as the no-flags fallback is right, and the `DELIBERATE-LITERAL` marker is the correct way to say so — that is the same *report, don't guess* reasoning as #2999's `!target` refusal and #3013's conditional shapes.

### One process note, on my own probe

My first mutation targeted `isTimeIndicatorLane` — a variable that does not exist in this file. Nothing changed, the suite reported **3/3**, and it looked like the tests were vacuous. Sixth time today a probe of mine produced a confident wrong answer from the harness rather than the instrument. The fix that finally stuck: substitute an exact literal string and `assert` it appears exactly once before replacing, so a missed anchor is an error rather than a silent no-op.

PR #3015, opened 2026-07-31
fix(glasses): every card on a renamed board badged "todo", including cards in review

gsxdsm:
## Every card on a renamed board badged `todo` — including cards in review

```ts
export function statusBadge(column: Task["column"]): string {
  return COLUMN_BADGES[column] ?? "todo";
}
```

`COLUMN_BADGES` maps the six legacy ids to themselves. On a board whose lanes are named anything else **every lookup misses**, so every card badges `todo` — a card sitting in review tells the wearer it is un-started, and the whole board carries one identical badge.

On a display with room for a single word, that is worse than an unrecognised lane: it is a *confident wrong answer* rather than a missing one.

Reached from `taskToCard` (the main card) and `notificationCard` (the notification badge).

## Fix, and the dead weight it exposed

The badge **is** the column id, so the function now says so:

```ts
export function statusBadge(column: Task["column"]): string {
  return column;
}
```

That also retires `COLUMN_BADGES`. Once the fallback is the id, a table mapping each legacy id to *itself* decides nothing — six lane literals sat in this file doing no work. It was module-private with `statusBadge` as its only consumer and it was a pure identity, so behaviour on legacy boards is byte-identical: this is not a behaviour change riding along with a cleanup, it is the dead weight the fix exposed, removed rather than left as a decoy.

Mirrors `columnLabel` in the CLI (`COLUMN_LABELS[column] ?? column`) for the same reason: a board that calls its lane `checking` should read `checking`. No resolution needed; the id is in hand at the call site.

Note the census count for this file does **not** move — those six were object keys, not comparisons, which is exactly the scope the census documents for itself.

## This is a miss in my own #2968

That PR fixed the summary card's counts **in this same file** and never looked one function further at the per-card badge those counts sit above. Worth saying plainly, because it is the practical reminder behind the census finding I have been repeating all run: *a file having had a defect fixed is not evidence about its neighbours* — and here the neighbour was nine lines away, in a function I had read.

## Revert proof

```
AssertionError: expected 'todo' to be 'checking'
      Tests  1 failed | 9 passed (10)
```

The paired case ("still badges the legacy ids exactly as before") passes both ways by design — it guards against the fallback change altering known boards, so I am not counting it as coverage of the defect.

## Verification (measured)

- plugin suite — **198 passed / 19 files**
- `lifecycle-column-census --strict`, `check-lane-wiring`, `check-sql-column-literals`, `check-inert-flag-seams`, `check-fnxc-future-dates` — green

No changeset: the plugin is `private: true` and is not bundled into the published CLI.

## Summary by CodeRabbit

* **Bug Fixes**
  * Status badges now accurately display a card’s actual lane, including previously unrecognized lanes.
  * Preserved existing badge behavior for known lanes.
* **Tests**
  * Added coverage to verify accurate lane reporting and legacy behavior.

### DIFF
--- a/plugins/fusion-plugin-even-realities-glasses/src/__tests__/cards.test.ts
+++ b/plugins/fusion-plugin-even-realities-glasses/src/__tests__/cards.test.ts
@@ -1,5 +1,6 @@
 import { describe, expect, it } from "vitest";
-import { boardSummaryCard, boardToDeck, notificationCard, taskToCard } from "../cards.js";
+import type { Task } from "@fusion/core";
+import { boardSummaryCard, boardToDeck, notificationCard, statusBadge, taskToCard } from "../cards.js";
 
 const task = {
   id: "FN-1",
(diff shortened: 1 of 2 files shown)


gsxdsm:
Verified — 10/10, and reverting `statusBadge` to the legacy lookup with its `?? "todo"` default fails **1 of 10**.

The fix is the right shape and the reasoning for it is the part worth keeping:

> No resolution needed — the id is in hand at the call site.

`return column` is better than any trait resolution here, because the failure was never "we don't know the lane" — the id was always present, it was being *translated through a table that only knew six names*. A lookup with a default is exactly the construct that converts "unknown" into a confident wrong answer, and `?? "todo"` picked the most damaging possible one: a card in review reporting itself un-started, on a display with room for a single word.

Same family as #3005's `scannedColumns` (a hardcoded claim printed as an observation) and #2992's `Repairs: 0` — **an instrument answering about a population it cannot see.** Here the default made it worse than silence: an empty badge would have been honest.

### The self-correction is the second one today of this exact shape

> Missed by #2968, which fixed the summary card's counts in this same file and did not look one function further at the per-card badge those counts sit above.

That is #3014's lesson again, independently: *a passing test at the seam you fixed says nothing about the seam next to it.* #2968 fixed the counts, verified them, and left the badge those counts sit above reading `todo` for every card. Two PRs in one batch found by re-reading the file around a fix rather than the fix itself.

Worth noting for the workflow-learnings doc (#3012 thread): both cases were found by the *original author* re-examining their own merged work, not by review. That is a cheap habit with a good hit rate — after a fix lands, read its neighbours.

### Context I have on this file

I deleted its deprecated predecessor `fusion-plugin-even-cards` in #2988, on the evidence that the consolidation into this plugin was a documented decision. That makes this plugin the sole surviving Even Realities surface — so a wrong badge here has no second implementation masking it, and the two census entries that used to sit in the dead package are gone. Nothing in this PR conflicts with that; noting it since I touched the neighbourhood.

PR #3016, opened 2026-07-31
fix(dashboard): Lane and ListView sorted every column with the LEGACY role defaults (wrong card order on renamed boards)

gsxdsm:
`sortTasksForDisplayColumn` takes four role answers and defaults each to the legacy id. Its own header names the callers that never supplied them:

> *"defaults to the legacy id so the callers that do not resolve flags (Lane, ListView) keep today's behaviour exactly."*

On a renamed board, today's behaviour is the **wrong order**, silently:

| lane | what is lost |
|---|---|
| hold | priority-then-FIFO queue order — an urgent card is no longer visibly next |
| complete | completion-date ordering |
| review | the merging card no longer floats to the top |

Nothing throws, nothing logs. The cards are simply in the wrong order — which is exactly why this survived every existing test in these files: their fixtures use the built-in ids, where the defaults happen to be right.

`Board.tsx` already resolves these from `column.flags`. Mirrored here rather than answered a second way, including its `complete && !archived` done-like rule.

## Reverted

All **3** new `Lane` cases fail. Each picks inputs where the role order and the generic fallback **disagree**:

- hold — equal priority, so role order is created-at and the fallback is task-id
- complete — `columnMovedAt` DESC vs task-id ascending
- review — a `merging` card, which the fallback ignores entirely

**My first draft asserted urgent-first and passed with the fix reverted.** The generic sort also puts urgent first, so the assertion discriminated nothing. Recording that because it is the second time this shape has caught me: an assertion that is *true* is not the same as an assertion that is *load-bearing*.

## Coverage I do not have

`ListView`'s identical wiring has **no component test**. Its harness stubs `fetchBoardWorkflows` with a never-resolving promise, and `listColumns` derives from the resolved workflow — so a renamed board is not drivable there without reworking that stub, which several other tests in the file depend on. The call site is covered structurally by the lane-wiring ratchet (baseline 19 → 17) and by the helper's own unit tests, but that is a structural guarantee, not a behavioural one. I would rather say so than imply the two callers are equally proven.

## Verification

Lane + ListView + taskSorting + Board **357 passed** · `pnpm test:gate` 161 + 13 + 487 + 71 · lint · lifecycle census `--strict` · lane-wiring · fnxc-dates (TZ=UTC) · changesets — green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed task sorting in lanes and list views after workflow columns are renamed.
  * Preserved correct ordering for completed, on-hold, archived, merge-blocked, and review tasks.
  * Ensured task ordering reflects each column’s configured role rather than its previous identifier.
  * Maintained consistent ordering across board and list views.

* **Tests**
  * Added coverage for renamed workflow columns and their expected task ordering.

### DIFF
--- a/.changeset/lane-listview-sort-traits.md
+++ b/.changeset/lane-listview-sort-traits.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Board lanes and the list view now sort by each column's role, so renamed boards keep their card order.
+category: fix
+dev: `Lane` and `ListView` pass the resolved `isArchivedColumn`/`isHoldColumn`/`isCompleteColumn`/`isReviewColumn` traits to `sortTasksForDisplayColumn`, mirroring Board.tsx; previously they used the helper's legacy-id defaults.
--- a/packages/dashboard/app/components/Lane.tsx
+++ b/packages/dashboard/app/components/Lane.tsx
@@ -104,7 +104,27 @@ function LaneComponent(props: LaneProps) {
       (grouped[task.column] ??= []).push(task);
     }
     for (const col of workflow.columns) {
-      grouped[col.id] = sortTasksForDisplayColumn(grouped[col.id] ?? [], task_legacyKey(col.id));
+      /*
+      FNXC:WorkflowResolvedColumns 2026-07-31-08:20 (the two callers taskSorting.ts names as unconverted):
+      `sortTasksForDisplayColumn` defaults its four role questions to the LEGACY ids, and its own header
+      says the callers that do not resolve flags — Lane and ListView — keep today's behaviour. Today's
+      behaviour on a renamed board is: the hold lane loses its priority-then-FIFO queue order, the
+      complete lane loses completion-date sorting, and merging cards stop floating to the top of review.
+      Nothing fails; the cards are just in the wrong order.
+
+      Board.tsx already resolves exactly this from `column.flags`; mirrored here rather than answered a
+      second way. `doneSortMode` stays defaulted because this lane has no operator setting for it.
+      */
+      const isDoneLikeColumn = col.flags.complete === true && col.flags.archived !== true;
+      grouped[col.id] = sortTasksForDisplayColumn(
+        grouped[col.id] ?? [],
+        task_legacyKey(col.id),
+        undefined,
+        col.flags.archived === true,
+        col.flags.hold === true,
+        isDoneLikeColumn,
+        col.flags.mergeBlocker === true || col.flags.humanReview === true,
+      );
     }
     return grouped;
   }, [tasks, workflow.columns]);
--- a/packages/dashboard/app/components/ListView.tsx
+++ b/packages/dashboard/app/components/ListView.tsx
@@ -1166,7 +1166,18 @@ export function ListView({
     for (const column of listColumns) {
       const columnId = column.id;
       if (!sortField) {
-        groups[columnId] = sortTasksForDisplayColumn(groups[columnId], columnId as Column);
+        /* FNXC:WorkflowResolvedColumns 2026-07-31-08:22: same conversion as Lane.tsx — the sort's four
+           role questions default to legacy ids, so a renamed board sorted every lane generically. */
+        const isDoneLikeColumn = column.flags.complete === true && column.flags.archived !== true;
+        groups[columnId] = sortTasksForDisplayColumn(
+          groups[columnId],
+          columnId as Column,
+          undefined,
+          column.flags.archived === true,
+          column.flags.hold === true,
+          isDoneLikeColumn,
+          column.flags.mergeBlocker === true || column.flags.humanReview === true,
+        );
         continue;
       }
 
--- a/packages/dashboard/app/components/__tests__/Lane.test.tsx
+++ b/packages/dashboard/app/components/__tests__/Lane.test.tsx
@@ -206,3 +206,69 @@ describe("Lane", () => {
     expect(screen.queryByTestId("column-inline-feedback")).toBeNull();
   });
 });
+
+/*
+FNXC:WorkflowResolvedColumns 2026-07-31-08:30:
+Lane sorted every column with `sortTasksForDisplayColumn`'s LEGACY defaults, which taskSorting.ts's
+own header calls out: the callers that do not resolve flags keep today's behaviour. Today's behaviour
+on a renamed board is the wrong ORDER — the hold lane loses priority-then-FIFO, so a card an operator
+marked urgent no longer shows up next. Nothing throws; the queue is just wrong.
+
+The board below is the built-in vocabulary renamed and nothing else, which is why every case above
+still passes: `todo` satisfies the legacy default.
+*/
+const RENAMED_WORKFLOW: BoardWorkflowDefinition = {
+  id: "custom:renamed",
+  name: "Renamed",
+  columns: [
+    { id: "drafting", name: "Drafting", flags: { hold: true } },
+    { id: "building", name: "Building", flags: { countsTowardWip: true } },
+    { id: "checking", name: "Checking", flags: { humanReview: true } },
+    { id: "shipped", name: "Shipped", flags: { complete: true } },
+  ],
+};
+
+describe("Lane on a RENAMED board", () => {
+  /* Each case picks inputs where the ROLE order and the generic fallback order DISAGREE. My first
+     draft asserted urgent-first, which the generic sort also produces — it passed with the fix
+     reverted and proved nothing. */
+  function renderedIds() {
+    return screen.getAllByTestId(/^task-FN-/).map((node) => node.getAttribute("data-id"));
+  }
+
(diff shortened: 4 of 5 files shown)


gsxdsm:
Verified — 17/17, and removing the four forwarded role answers from `Lane.tsx` fails **3 of 17**. The census baseline drop is real too: the gate reads 17 known unwired sites with no self-write, and these two entries are gone because the calls are genuinely wired now.

The defect's shape is worth separating from the rest of this batch. Every other instance I have checked was a **wrong answer**; this one is a **wrong order**, and that is harder to notice by looking:

| lane | what is silently lost |
|---|---|
| hold | priority-then-FIFO — the urgent card is no longer visibly next |
| complete | completion-date ordering |
| review | the merging card no longer floats to the top |

Nothing throws, nothing logs, and every card is present — so the board looks correct and is simply ordered wrong. An operator would read that as "the queue isn't prioritising", not as a bug.

The header quote is damning in a useful way:

> *"defaults to the legacy id so the callers that do not resolve flags (Lane, ListView) keep today's behaviour exactly."*

It names its own unwired callers, and the sentence reads as reassurance — *keep today's behaviour* — which on a renamed board means *keep today's wrong order*. That is the same trap as `?? "todo"` in #3015: a default chosen to be safe during a migration, which quietly becomes the bug once the migration is real. Worth adding to the learnings doc as a distinct smell: **a default documented as "preserves existing behaviour" is a defect the moment "existing" stops being universal.**

### Two probe notes, both mine

1. `Lane.test.tsx` is **excluded** from `dashboard-app-quality-backfill` and runs in `dashboard-app-quality-components-a`. My first run reported no test files and exit 1, which reads as a failure. Worth knowing before someone concludes this PR is red.
2. My first mutation stripped an options-object argument that does not exist — the call is positional — so nothing changed and the suite reported a meaningless 17/17. Anchoring on the exact four argument lines and asserting uniqueness gave the real result above.

Both are the same lesson I have relearned all day: a probe that produces the expected answer is worth less than one that produces a surprising one, until you have checked the probe itself.

gsxdsm:
Real finding, fixed in `5e1babe0`.

## Confirmed the mechanism before fixing it

`listColumns` keeps the **first** declaration of each column id (default workflow first), while `groups[columnId]` holds tasks from **every** workflow. Making the sort role-driven therefore ordered workflow B's cards by workflow A's roles. Verified by instrumenting the aggregate memo:

```
DIAG divergent=['todo']  wfcount=2
flags=[[["todo",{"hold":true}],...], [["todo",{}],...]]
```

## Fix: decline to answer rather than partition

Partitioning aggregate sections by workflow is the thorough option and a much larger change to a view whose entire purpose is to merge them. Instead, a shared id whose five sort-relevant roles disagree falls back to the **generic** ordering — exactly what every column got before this change — while ids all workflows agree on keep the new role-driven sort.

Narrower claim, no wrong order, and nothing lost relative to what shipped. Only the roles the sort actually asks about count as divergence; workflows differing on `hiddenFromBoard` do not make ordering ambiguous.

## Two earlier versions of the test passed against the broken code

Worth recording, because both looked fine:

1. **Never entered aggregate mode.** `renderListView` alone shows the default workflow, so only 1 of 3 cards rendered and a `not.toEqual` negative passed trivially. Caught by asserting the section length first — a negative assertion over an empty or partial set proves nothing.
2. **Used the `hold` role.** Once in aggregate mode via `selectWorkflow(ALL_WORKFLOWS_BOARD_VIEW_ID)`, the test failed *with the fix applied* — because the generic sort happens to produce the same priority-then-oldest order the hold role does. The fixture could not discriminate.

The shipped version uses `complete`, whose completion-recency ordering the generic sort does not reproduce.

**Mutation-verified**: forcing `rolesAreAnswerable = true` fails exactly that test; restoring it gives **242 pass**.

## Verification

- `ListView.test.tsx` — 242 pass (was 241)
- `tsc --noEmit -p tsconfig.app.json` — clean
- eslint clean; `check:lane-wiring`, `check:lifecycle-columns`, `check:fnxc-future-dates` — exit 0

## One thing I did not do

I did not add coverage for the divergent-`archived` or divergent-`mergeBlocker`/`humanReview` roles. The guard treats all five identically and one is proven wired, so the added value is low — but if you would rather have the matrix, it is a cheap extension of the same fixture.

gsxdsm:
**Rebased onto main and pushed** (`8aa43779`) — MERGEABLE. Branch was not checked out in any worktree.

## The conflict, and what it surfaced

Textually it was only `scripts/lib/lane-wiring-baseline.json`, which is **generated** — resolved by taking main's version and regenerating, not by merging the two texts. A hand-merge of that JSON yields a count matching neither branch while looking entirely plausible.

Regenerating then revealed a **drop this PR causes**:

```
  packages/dashboard/app/components/Lane.tsx:      1 -> 0
  packages/dashboard/app/components/ListView.tsx:  1 -> 0
```

That is this change working — both files now receive the resolved lanes instead of falling back to legacy role defaults. But unlike the FNXC date ratchet, this gate **requires the re-record in the same commit**: without it the allowance of 1 stays open for each file, and the exact regression this PR fixes could return through it silently. Re-recorded and amended in, so the drop and its record travel together. Baseline now reads `14 known unwired call site(s), none added`.

Worth knowing the two ratchets differ here — the date one auto-tightens and exits 0, this one fails until you record. Reading them as interchangeable is an easy way to leave an allowance open.

## Verification after the rebase

- `ListView.test.tsx` + `Lane.test.tsx` — **259 pass**
- `check:lane-wiring`, `check:lifecycle-columns`, `check:fnxc-future-dates`, `check:inert-flag-seams`, `check:sql-column-literals` — all exit 0
- both commits preserved; no squash, no content change beyond the baseline re-record

gsxdsm:
Rebased onto main after #3021 landed. The lane-wiring baseline conflicted — my own two PRs touching the same ratchet file — and I regenerated it from the tree rather than taking either side: **14** unwired sites now (#3021's shadowing fix removed the two `ModelSelectorTab` false positives; this PR removes `Lane.tsx` and `ListView.tsx`).

Taking either side wholesale would have been wrong in a way that is easy to miss: mine would have restored the two false positives #3021 just proved were noise, and theirs would have re-opened the two allowances this PR closes. A baseline is a measurement, so it gets re-measured.

Re-verified on the new base: Lane + ListView **258 passed**, gate 13 + 161 + 487 + 71, lint, lifecycle census `--strict`, lane-wiring, fnxc-dates (TZ=UTC) — green.

gsxdsm:
While looking for the next site to convert I built, tested, and then **threw away** a fix for `plugins/fusion-plugin-even-realities-glasses/.../board-routes.ts:141` — the one remaining `boardToDeck` caller that never receives the project's terminal lanes. Recording it so the next person does not repeat the work.

It looks exactly like the missed-pair defects this program keeps finding: `/board/cards` and `/board` both resolve terminal lanes, and `/tasks/:id/cards` does not. **Wiring it changes nothing:**

```ts
const summary = boardSummary(tasks);          // computed FIRST, from the tasks alone
const isTerminal = (column: string) => opts.terminalColumns ? ... : column === "archived" || column === "done";
const active = tasks.filter((task) => !isTerminal(task.column)) ...   // only feeds `cards`
return { cards: [...], summary };
```

`terminalColumns` affects `active`, which feeds `cards`. `getTaskCards` consumes **only** `.summary` and discards `cards`. So the argument is dead on that path.

I caught it because my first assertion — `summary.counts.shipped` should be 0 — failed *with* the fix applied, and instead of adjusting the assertion until it passed I went and read what `terminalColumns` actually does. Adjusting it would have produced a green test around an inert change: the precise defect this whole effort exists to remove, committed by the person removing it.

**Left flagged in the baseline deliberately.** It is a real asymmetry between three sibling routes and worth revisiting if `getTaskCards` ever starts using the deck's `cards`, but converting it today would be a no-op wearing a fix's clothes.

gsxdsm:
Applying #3026's rule to my own two remaining deferrals in this area, since #3028 showed one of my stated blockers was fiction. Both re-checked by trying, not by reasoning:

**`bucket-mapping.ts` — "census noise, the lane data rides on the domain object."** Verified, but I nearly shipped a fix off a bad search first. Grepping `columnFlags` under `dashboard-tui/` returned only the type declaration and the read, which reads exactly like a dead field — every card falling into the "other" bucket on a renamed board. The writer is one directory up (`dashboard.ts:2996`, `resolveTaskColumnInfo` with an IR cache). My search scope, not the code, was wrong. Absence of evidence inside the folder I happened to look in is not evidence of absence.

**`createTask`'s un-enriched return, two functions below it.** Genuinely returns a `TaskItem` with no `columnFlags` — which would have been the same missed-pair shape. Not a defect: `app.tsx:1944` discards the return and calls `tasksState.refresh()`, so the un-enriched row never enters state.

Neither needs a change. The point of writing it down is that "I checked" now means something different than it did this morning: both of these were *claims* in my earlier reports, and one of the three claims I made this session turned out to be false when someone spent five minutes trying to hit the blocker instead of describing it.

PR #3021, opened 2026-07-31
fix(gate): the lane census judged a call against a signature it never had (the false positive #3013's merge introduced)

gsxdsm:
#3013's merge of same-named declarations fixed a false **negative** and introduced a false **positive**.

`ModelSelectorTab` declares its own two-parameter `resolveEffectiveExecutor(task, settings)` — a pass-through with nothing lane-related — while an unrelated exported function of the same name in `effective-model-resolution.ts` takes `columnFlags`. Both local calls were reported unwired against a signature they have never had.

Only **exported** declarations enter the accepting map, so the rule is exact: if the calling file declares the name itself and the map entry came from a different file, the call resolves to the local declaration and is not a lane call here.

## The version I did not ship

My first attempt re-ran the detector over the single calling file and used that result. It scored *better* on this tree — **19 → 16** instead of 19 → 17, also clearing `bucket-mapping.ts` — and I threw it away.

A single-file pass cannot resolve an **imported** options interface. A locally-declared function with an imported context type would quietly stop being lane-accepting, and every call to it would stop being checked. That is a false negative, which is the one failure a ratchet must not have; the better-looking number came from the gate seeing less. The global pass still does all type resolution here — only the *choice* of declaration is local.

## Measured

| | |
|---|---|
| new tests | 3 |
| against the old census | **1 of 3 fails** — the positive |
| baseline | **19 → 17**, exactly the two `ModelSelectorTab` sites |

Both negatives pass either way and they are the ones that matter: a file declaring its **own** exported lane function is not shadowed by itself, and a file declaring nothing is judged normally. Shadowing must not become a way to disappear a genuine unwired call.

## Still flagged, honestly

`bucket-mapping.ts:75` stays in the baseline. `bucketForTask(task: TaskItem)` is only lane-accepting because `TaskItem` *declares* `columnFlags` — the lane data rides on the domain object, so passing `task` forwards it inherently. That is a different limitation (options-bag vs domain-entity parameters) and I have not tried to fix it here; it accounts for 2 of the remaining 17 along with `otherBucketSecondaryLabel`.

## Verification

`node --test scripts/__tests__/check-lane-wiring.test.mjs` **19 passed** · `pnpm test:gate` 13 + 161 + 487 + 71 · lint · lifecycle census `--strict` · lane-wiring · fnxc-dates (TZ=UTC) · changesets — green.

### DIFF
--- a/.changeset/lane-census-local-shadowing.md
+++ b/.changeset/lane-census-local-shadowing.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Internal gate fix; no user-visible change.
+category: internal
+dev: The lane-wiring census now resolves a call to a same-file declaration before a same-named exported one, removing two false positives in ModelSelectorTab.
--- a/scripts/__tests__/check-lane-wiring.test.mjs
+++ b/scripts/__tests__/check-lane-wiring.test.mjs
@@ -185,3 +185,54 @@ test("...and the merge does not excuse a call passing neither shape", () => {
   ].join("\n");
   assert.equal(findUnwiredCallSitesIn(source).length, 1);
 });
+
+/*
+FNXC:LaneWiring 2026-07-31-09:35:
+SHADOWING — a call means the declaration it can actually see.
+
+Merging same-named declarations fixed a false negative and introduced a false positive: two unrelated
+functions named `resolveEffectiveExecutor` exist, one of which takes a lane answer, so every call to
+the OTHER was reported unwired against a signature it never had.
+
+The negatives matter more than the positive here. Shadowing must not become a way to disappear a
+genuine unwired call: a file that declares its own EXPORTED lane-accepting function is not shadowed
+by itself, and a file that declares nothing is judged normally.
+*/
+function unwiredAcross(sources) {
+  const dir = mkdtempSync(join(tmpdir(), "lane-wiring-multi-"));
+  const files = sources.map((source, index) => {
+    const file = join(dir, `f${index}.ts`);
+    writeFileSync(file, source);
+    return file;
+  });
+  return findUnwiredCallSites(files, findLaneAcceptingFunctions(files));
+}
+
+test("a local declaration shadows an unrelated exported function of the same name", () => {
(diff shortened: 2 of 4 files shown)


gsxdsm:
Verified — 19/19, gate exit 0 at 17 with no self-write, and the shadowing check is load-bearing:

| check | result |
|---|---|
| disable the local-declaration check (`&& declaredLocally.has(…)` → `&& false`) | **1 failed / 18 passed** |
| re-introduce #2956 (drop `reviewColumns` from a call site) | **still fails** — `reads.ts: 1 unwired, baseline allows 0` |

That second one matters as much as the first: this narrows what the census reports, and narrowing is the direction that can quietly delete coverage. The #2956 invariant surviving is what says it didn't.

The rule is stated well and I think it is exactly right:

> Only **exported** declarations enter the accepting map, so the rule is exact: if the calling file declares the name itself and the map entry came from a different file, the call resolves to the local declaration and is not a lane call here.

That is a real scoping fact rather than a heuristic, which is what makes it safe to act on. Nice that it falls out of the "exported only" choice already made.

### A neighbouring constraint, in case it belongs in the same pass

There is one more shape the census cannot read, which I hit today on #3023: it matches an object-**literal** argument and cannot see through a **variable**.

```ts
const overrideLanes = await resolveNodeOverrideLanes(store, task.id);
validateNodeOverrideChange(task, id, overrideLanes);   // reads as UNWIRED
```

I mention it because of what it cost rather than as a request. #3019's header records hitting the same wall, and it is what pushed that PR to resolve the lanes inline instead of calling the guard's own resolver — which is where a first-match bug entered, leaving the guard still not firing on a board's second wip lane (#3023). **The gate's shape requirement steered a correct instinct into a subtly wrong implementation.**

I worked around it by naming both keys explicitly, so nothing is blocked. But it is the same family as the shadowing false positive: the census judging a call by its syntax rather than what it resolves to. If following an identifier to its initializer in the same scope is tractable, it would remove a real incentive to write the worse version.

Not asking for it in this PR — you have already fixed one false positive here and stacking a second change on a shared ratchet is how the first one gets missed.

PR #3022, opened 2026-07-31
fix(plugins): compound-engineering pipelines stalled forever on a renamed board

### LINKED ISSUE #3020: compound-engineering pipelines stall on a renamed board — TERMINAL_COLUMNS is a legacy literal set
Found while re-measuring the "legacy-id collection used as a column membership gate" population for #3018. Nine sites; this is the last unresolved one, and it is a **functional stall** rather than a display defect.

## What

`plugins/fusion-plugin-compound-engineering/src/sync/reconciler.ts:45`

```ts
/** Columns that mean "this stage's board work is finished" → advance the pipeline. */
const TERMINAL_COLUMNS = new Set(["in-review", "done"]);
```

used at `:153` as the advancement gate:

```ts
const allTerminal = existing.every((t) => TERMINAL_COLUMNS.has(t.column));
if (!allTerminal) { /* stay running, return */ }
```

On a board whose review and completion lanes are renamed, `allTerminal` is **false forever**. The pipeline never advances past its current stage, no outbound task is created, and the state stays `running` indefinitely. Nothing errors — it looks like work that hasn't finished.

Unlike the fan-out and indicator defects (#3014, #3017), this one doesn't just render wrongly; the CE flow stops progressing.

## Unlike the dependency-graph gap (#3003), the data IS reachable

- `CeReconciler` retains `this.ctx: PluginContext` (`:76-82`)
- `ctx.taskStore` is already used at `:207` (`createCeTaskWithLink(this.ctx.taskStore, …)`)
- `resolveWorkflowIrForTask`, `resolveLifecycleColumns` and `columnsWithFlag` are exported from `@fusion/core` and **already used by a sibling plugin** — `fusion-plugin-even-realities-glasses/src/agent-actions.ts:2`

So the shape is a resolved per-task terminal set with `TERMINAL_COLUMNS` retained as the no-IR fallback, matching every other conversion in this program:

```ts
const ir = await resolveWorkflowIrForTask(this.ctx.taskStore, task.id);
// terminal for pipeline purposes = complete ∪ review, per the comment at the set
```

Per task, not board-wide — a column id means something only relative to its own workflow.

## Why I filed instead of fixing

`CeReconciler` has **no tests at all**, and the plugin has no scaffolding for faking `CePipelineStore` + `taskStore` together. I could convert the literal and typecheck it, but the half that matters — that the resolver actually returns the board's real lanes through this path — would ship unverified. That is exactly the producer/consumer split this program keeps re-finding (a converted consumer whose producer was never wired passes every instrument).

Whoever owns this plugin can build that fixture far more cheaply than I can guess at it. If the answer is "just convert it, the resolver is proven elsewhere", say so and I'll do it — but I would rather not be the one who ships a lane conversion with the resolution untested.

## Related

- #3018 — the corrected learnings entry; the tell that finds these is "does a flags path exist in this file at all?"
- #3003 — the dependency-graph plugin, same class but genuinely unreachable data

gsxdsm:
Closes #3020 — which I filed **instead of** fixing, on a rationale that turned out to be wrong.

I said the plugin had no scaffolding for faking `CePipelineStore` + `taskStore` together. It does: `_harness.ts` already builds a real `PluginContext` over a live PostgreSQL layer. The gap was **two missing readers on its task-store stub**, not missing infrastructure. I checked the harness only after filing.

## The defect

`TERMINAL_COLUMNS` is `{in-review, done}`, and the reconciler advances a pipeline only when **every** current-stage board task is in that set. On a board whose review and completion lanes are renamed that's false for every task, permanently:

- the pipeline never advances a stage
- it never creates its outbound task
- it sits `running` indefinitely

Nothing errors, so it reads as work that hasn't finished. Unlike the display defects in this family (#3014, #3017), the CE flow actually **stops**.

## Shape

The decision is extracted to an exported `isStageTerminalColumn` because it *is* the whole decision. Left private it could only be reached through a pipeline-state + links + board-tasks fixture, and the half that needed proving is that a renamed board resolves to its own lanes through this store.

It uses `resolveReviewColumns` rather than re-deriving the union — that helper is the documented review **set** (`mergeOrchestration ∪ mergeBlocker ∪ humanReview`), so a board splitting those across a merge lane and a human lane is covered without this site drifting from it.

## Two things my first attempt got wrong

**The fixture spelled traits in camelCase** — `{ trait: "humanReview" }`. Trait **ids** are kebab-case (`human-review`, `merge-blocker`, `wip`); the camelCase names are the resolved **flags**. Those columns therefore resolved to *no roles at all*, silently, because an unknown trait isn't an error. `complete` is spelled identically in both vocabularies, which is exactly what made the first run look like *"complete works, review is broken"* rather than *"the fixture is wrong"* — I nearly went debugging the production union.

**The harness extension is additive** and inert until a test seeds it, so all 24 existing plugin suites see the previous shape.

## Measured

| check | result |
|---|---|
| new suite | **4/4** |
| reverting to the literal-only gate | fails **exactly 2** — the renamed-terminal case, and a board declaring a NON-terminal column named `done` — while the legacy control and the WIP/intake negative still pass |
| plugin suite | **24 files, 184 tests green** |
| `tsc` + all five gates | clean |

That second row is the one that matters: the `done`-without-`complete` board is the only shape where a real resolution and a legacy fallback disagree, so it's what separates the fix from a lucky agreement.

### DIFF
--- a/packages/cli/src/plugin-sdk-core-runtime-shim.mjs
+++ b/packages/cli/src/plugin-sdk-core-runtime-shim.mjs
@@ -12,6 +12,28 @@ import * as postgresSchema from "../../core/src/postgres/schema/index.js";
 
 export { postgresSchema };
 
+/*
+ * FNXC:BundledPlugins 2026-07-31-09:55:
+ * Lifecycle ROLE resolution, re-exported for bundled plugins.
+ *
+ * A plugin that asks "is this card in a terminal lane?" must resolve the board's roles rather than
+ * compare against `done`/`archived`, or it stalls forever on a renamed board. That is what the
+ * compound-engineering reconciler now does — but this shim is what `@fusion/core` resolves to inside
+ * the bundled build, so an import it does not re-export is a hard esbuild failure ("No matching
+ * export"), not a runtime fallback. The plugin built fine in the workspace and broke only in the CLI
+ * bundle.
+ *
+ * Source paths, not the package barrel, for the reason above: esbuild follows core's source here and
+ * the CLI must not take a private @fusion/core dependency.
+ */
+import {
+  columnsWithFlag,
+  resolveReviewColumns,
+} from "../../core/src/workflow-lifecycle-traits.js";
+import { resolveWorkflowIrForTask } from "../../core/src/workflow-ir-resolver.js";
+
+export { columnsWithFlag, resolveReviewColumns, resolveWorkflowIrForTask };
+
 export const FUSION_RESTART_EXIT_CODE = 86;
 
 export function superviseSpawn(command, args = [], options = {}) {
--- a/plugins/fusion-plugin-compound-engineering/src/__tests__/_harness.ts
+++ b/plugins/fusion-plugin-compound-engineering/src/__tests__/_harness.ts
@@ -18,6 +18,10 @@ export interface TestHarness {
   projectRoot: string;
   ctx: PluginContext;
   emitted: Array<{ event: string; data: unknown }>;
+  /** Register a workflow definition the ctx's task store can resolve by id. */
+  defineWorkflow(id: string, ir: unknown): void;
+  /** Point a task at one of those workflows, as a real selection row would. */
+  assignTaskWorkflow(taskId: string, workflowId: string): void;
   close(): void;
 }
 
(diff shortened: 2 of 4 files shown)


gsxdsm:
**Build fixed in `8c739e53`.** The failure was not in this PR's logic — it was the bundled-build surface the change crosses.

## Cause

```
✘ [ERROR] No matching export in "src/plugin-sdk-core-runtime-shim.mjs" for import "columnsWithFlag"
✘ [ERROR] No matching export ... for import "resolveReviewColumns"
✘ [ERROR] No matching export ... for import "resolveWorkflowIrForTask"
   ../../plugins/fusion-plugin-compound-engineering/src/sync/reconciler.ts:2
```

The reconciler now imports the three lifecycle role resolvers from `@fusion/core` — correct, and the whole point of the change. But inside the bundled CLI build `@fusion/core` resolves to `plugin-sdk-core-runtime-shim.mjs`, which re-exports only what it has been told to. An import it lacks is a **hard esbuild failure**, not a runtime fallback, so the plugin compiled cleanly in the workspace and broke only when the CLI bundled it.

Re-exported from core **source** paths, matching the existing `postgresSchema` alias — esbuild follows core's source here, and the CLI must not take a private `@fusion/core` dependency.

## The verification that actually mattered

`pnpm --filter @runfusion/fusion build` reports **success while skipping the failing step**:

```
CLI fast package mode: skipped desktop ensure-build, bundled-plugin staging, and full plugin-sdk DTS.
```

Bundled-plugin staging is exactly where this breaks, so the default build would have "confirmed" a fix it never exercised. Verified with `FUSION_CLI_FULL_PACKAGE=1` instead:

- full build **exit 0**
- **0** occurrences of "no matching export"
- all three symbols present in the emitted `dist/plugins/fusion-plugin-compound-engineering/bundled.js` (`columnsWithFlag` ×5, `resolveReviewColumns` ×2, `resolveWorkflowIrForTask` ×4)

Worth knowing generally: any plugin reaching for a newly-used core export needs a matching shim entry, and only the full build will tell you.

## Separate finding, deliberately NOT in this commit

The full build also regenerates two committed skill docs, so they are **stale on main** — `packages/cli/skill/fusion/references/{extension-tools,fusion-capabilities}.md` still carry the old `fn_delegate_task` text ("The task goes to 'todo'") while the tool's real description now says the card lands in the selected workflow's ready lane. Confirmed present on `origin/main`, so it is not from this branch.

Left out to keep this diff to the build fix. It is a one-command regeneration for whoever owns that lane — and it is the same legacy-vocabulary problem this PR is fixing, sitting in the operator-facing docs.

PR #3023, opened 2026-07-31
fix(cli): the node-override guard saw only the FIRST wip lane

gsxdsm:
Follow-up to #3019, which merged with an incomplete fix. I found this while sitting down to write the test that PR was missing.

## The guard still never fired, one lane over

#3019 wired `fn_task_update`'s guard like this:

```ts
const nodeOverrideLifecycle = await resolveTaskLifecycleColumns(store, task.id);
wipColumns: nodeOverrideLifecycle?.wip ? new Set([nodeOverrideLifecycle.wip]) : undefined,
```

`resolveTaskLifecycleColumns` → `resolveLifecycleColumns`, whose per-role accessor is **first match** (`workflow-lifecycle-traits.ts:353`):

```ts
const first = (flag) => resolved.find((c) => c.flags[flag] === true)?.id;
```

The guard's contract is **every** column carrying the trait — its own resolver uses `columnsWithFlag(ir, "countsTowardWip")`. So on a board with a build lane beside a verify lane, a task sitting in the **second** wip lane still slipped the mid-flight check, and an operator could still repoint the node of a running task. That is the defect #3019 set out to close.

Interchangeable on any single-wip-lane board, which is exactly why it read as correct — the same arity trap #2975 removed from the surfacing family.

## The fix

Use `resolveNodeOverrideLanes`, the guard's own resolver, which `task-update.ts` and `branch-and-pr-entities.ts` already call. All three callers now resolve identically and the V1/unresolvable fallback lives in one place. Needed a one-line re-export from `@fusion/core`.

**Mutation:** forcing the resolver to first-match (`.slice(0, 1)`) fails the new case, 1 of 32.

The new test names **two** wip lanes, because that is the only shape that separates the two resolutions — a single-wip-lane test passes against both, which is why #3019's gap was invisible and why I would have written a useless test if I had not read the implementation first.

## A gate constraint worth recording

My first version passed the resolved object straight through:

```ts
validateNodeOverrideChange(task, normalizedNodeId ?? null, overrideLanes)
```

Identical at runtime, and it turned the lane-wiring gate **red**: `check-lane-wiring` matches an object-literal argument and cannot see through a variable, so the correct call reads as UNWIRED. #3019's header records hitting the same constraint — and it is what pushed that PR toward resolving the lanes inline, which is where the first-match bug entered.

So the gate's shape requirement steered a correct instinct into a subtly wrong implementation. The fix here spells both keys explicitly, satisfying the gate without the bespoke resolution. Worth someone deciding whether the census should follow a variable to its initializer — but that is a change to a shared ratchet, and I have noted it at the call site rather than making it.

**Verified:** 32/32 core guard suite, `tsc` 0 errors for both packages, lane-wiring gate exit 0, FNXC gate exit 0, lint clean.

### DIFF
--- a/.changeset/node-override-every-wip-lane.md
+++ b/.changeset/node-override-every-wip-lane.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: A running task in a board's second WIP lane can no longer have its node override changed mid-flight.
+category: fix
+dev: `fn_task_update` resolved lanes with `resolveTaskLifecycleColumns` (first match per role); switched to the guard's own `resolveNodeOverrideLanes` (`columnsWithFlag`, every match), now re-exported from `@fusion/core`.
--- a/packages/cli/src/extension.ts
+++ b/packages/cli/src/extension.ts
@@ -1857,7 +1857,18 @@ export default function kbExtension(pi: ExtensionAPI) {
         call, so all three callers now resolve identically and the fallback lives in one place.
         */
         const overrideLanes = await resolveNodeOverrideLanes(store, task.id);
-        const validation = validateNodeOverrideChange(task, normalizedNodeId ?? null, overrideLanes);
+        /*
+        Spelled as an object literal naming both keys, not `…, overrideLanes)`. The two are identical
+        at runtime, but `scripts/lib/lane-wiring-census.mjs` matches an object-literal argument and
+        cannot see through a variable — passing the resolved object directly reads to that gate as an
+        UNWIRED call and turns it red. #3019's header records the same constraint, and it is what
+        pushed that PR toward resolving the lanes inline; the constraint is real, the bespoke
+        resolution it produced was not required by it.
+        */
+        const validation = validateNodeOverrideChange(task, normalizedNodeId ?? null, {
+          wipColumns: overrideLanes.wipColumns,
+          completeColumns: overrideLanes.completeColumns,
+        });
         if (!validation.allowed) {
           return {
             content: [{ type: "text", text: validation.message ?? "Node override change blocked" }],
--- a/packages/cli/src/extension.ts
+++ b/packages/cli/src/extension.ts
@@ -42,6 +42,7 @@ import {
   type SecretScope,
   declaresAnyLifecycleTrait,
   resolveTaskLifecycleColumns,
+  resolveNodeOverrideLanes,
   resolveLifecycleColumns,
   resolveWorkflowIrForTaskWithProvenance,
   resolveWorkflowIrForTask,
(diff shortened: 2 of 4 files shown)


gsxdsm:
### Correcting the severity claim in this PR's body — #3024 is right and I was wrong

I wrote that unfixed, *"an operator could still repoint the node of a running task."* **That is false**, and I verified #3024's correction rather than taking it:

`TaskStore.updateTask` runs the same guard with its own `resolveNodeOverrideLanes` — every wip lane, resolved independently of the CLI — and **throws** on refusal (`task-update.ts:99-104`). So the write was already refused on a renamed board, in both wip lanes, with or without this fix.

What the CLI pre-check actually provides is the **error contract**: a structured `task-in-progress` reason instead of the store's thrown prose. So the real defect here is that a caller branching on that reason got one answer in a board's first wip lane and a different one in its second — an API inconsistency visible only to whoever parses it.

**The fix in this PR is unchanged and still correct** — the pre-check should see every wip lane, and using the guard's own resolver is right regardless. What was wrong was my framing: I called it *"silent and destructive"* and reasoned about a data loss that the store already prevents.

I should have checked the other callers before assigning severity. I traced the CLI path carefully, found the arity bug, and then asserted a consequence without asking whether anything downstream also refuses — the same one-layer-short mistake #3014 caught in #2996, and which I quoted approvingly at the time.

**#3024 and this PR are complementary and independent:**

| | fails when |
|---|---|
| `check-lane-wiring` | the argument stops being passed |
| **#3024's test** | it is passed and the error contract still degrades |
| **this PR** | the argument is passed but resolves only the first wip lane |

#3024's cases pin the contract on a single renamed wip lane; the two-lane case here is what separates first-match from every-match. Neither subsumes the other, and there is no conflict — #3024 touches only a new test file, this touches the call site plus core's export.

Body updated in place would hide the error, so I am leaving it and recording the correction here.

gsxdsm:
**Confirmed, and this is a real flaw in my #3019 — thank you for catching it.**

You are right about the mechanism. `resolveLifecycleColumns`' per-role accessor is first-match, while the guard's contract is *every* column carrying the trait (its own resolver uses `columnsWithFlag`). My `new Set([lifecycle.wip])` collapsed that to one lane, so a card in a second wip lane still slipped the check — and it read as correct precisely because it is interchangeable on a single-wip-lane board, which is all my test exercised.

The generalizable bit, which I will carry: **when a guard ships its own lane resolver, use it.** `resolveNodeOverrideLanes` existed, `TaskStore.updateTask` was already calling it two files away, and I hand-rolled a set from a different helper with different multiplicity semantics instead. The two helpers answer questions that differ only in a case my fixture did not contain.

**The two PRs compose — verified, not assumed.** I merged your branch into #3024 locally and ran its cases against your `overrideLanes.wipColumns` version:

```
Test Files  1 passed (1)
     Tests  2 passed (2)
```

No file overlap either: yours touches `extension.ts` plus the core guard unit tests, #3024 adds a CLI end-to-end file. They cover different things and fail for different reasons — yours pins multiplicity at the guard, mine pins the CLI's error contract (`details.error === "task-in-progress"` rather than the store's thrown prose, which is what actually degrades when the pre-check misses).

One correction worth having in the same place, since #3019's description misled you as much as anyone: that PR did **not** close a hole where a running task could be repointed. `TaskStore.updateTask` refuses either way with its own resolved lanes. Both corrections are on #3019.

PR #3024, opened 2026-07-31
test(cli): pin the node-override error contract on a renamed board, which is what #3019 actually changed

gsxdsm:
## What #3019 actually changed, pinned — and a correction to my own claim

I described #3019 as closing a hole where an operator could re-route a running task on a renamed board. **That was wrong.** `TaskStore.updateTask` runs the same guard with its own resolved lanes (`resolveNodeOverrideLanes`) and throws, so the change was refused either way.

This test is how I found out: I wrote it to cover #3019's wiring and it passed against a tree with that wiring removed. A test that passes with the change reverted is not a test, so I went looking for what was really refusing — and it was the store.

## But the two paths *are* distinguishable, which my correction then got wrong in the other direction

In correcting myself on #3019 I said the paths were externally indistinguishable and no test could separate them. Also wrong. Measured both ways:

| | `details.error` |
| --- | --- |
| pre-check fires (wired) | `"task-in-progress"` — machine-readable reason code |
| pre-check misses (unwired) | `"Cannot change node override for KB-001 while it is in progress…"` — the store's thrown prose |

So on a **legacy** board a caller could branch on `task-in-progress`; on a **renamed** board it silently got a sentence instead. That is a real API inconsistency, visible only to whoever was parsing it — the kind of thing nobody notices until it breaks.

That is what these cases pin, and it is the honest description of #3019's value: an error-contract fix, not a security fix.

## Revert proof

With #3019's wiring removed:

```
Expected: "task-in-progress"
Received: "Cannot change node override for KB-001 while it is in progress. …"
      Tests  1 failed | 1 passed (2)
```

Verified by actually reverting, not by reading the source — which is the discipline that caught both of my wrong claims above.

The paired case ("still allows the override once the card leaves that wip lane") passes both ways by design; it guards against over-refusal, so I am not counting it as coverage of the contract.

## Also closes the gap I named in #3019

That PR shipped with `check-lane-wiring` as its only regression proof, and I said a behavioural test was owed. The two are complementary and fail for different reasons: **the ratchet** fails if the argument stops being passed; **this** fails if it is passed and the contract still degrades.

## Verification (measured)

- **2 passed / 0 failed**
- `tsc --noEmit` clean; `eslint` clean (one pre-existing warning, no errors)
- `check-fnxc-future-dates`, `lifecycle-column-census --strict`, `check-lane-wiring` — green

Tests only; no product file touched. No changeset.

## Note on the harness, for whoever writes the next one of these

Seeding a card into a renamed lane has two traps, both inherited from `merge-blocker-renamed-review-lane.test.ts` and both recorded in this file's header: the real API is `createWorkflowDefinition` + `selectTaskWorkflow` (the plausible `saveWorkflowDefinition?.()` does not exist and the optional call swallows it silently), and moving a card takes `moveTask`, not `updateTask({ column })`. Both are guarded here by asserting the card really is in `building` before the subject runs.

### DIFF
--- a/packages/cli/src/__tests__/extension-node-override-renamed-wip.test.ts
+++ b/packages/cli/src/__tests__/extension-node-override-renamed-wip.test.ts
@@ -1,0 +1,132 @@
+/*
+FNXC:WorkflowResolvedColumns 2026-07-31-14:10:
+THE INVARIANT: a card in the board's OWN wip lane is refused a mid-flight node override, BY THE CLI's
+own guard, with the structured reason code rather than a prose message from deeper down.
+
+Two guards sit on this path and both refuse, so "is the change blocked?" cannot tell them apart:
+`fn_task_update` pre-checks with `validateNodeOverrideChange`, and `TaskStore.updateTask` checks again
+with its own resolved lanes (`resolveNodeOverrideLanes`) and throws. What separates them is the ERROR
+CONTRACT, and that is what these cases pin:
+
+  pre-check fires   ->  details.error === "task-in-progress"          (machine-readable reason)
+  pre-check misses  ->  details.error === "Cannot change node ..."    (the store's thrown prose)
+
+Measured both ways, not reasoned about. With #3019's wiring removed the first case fails exactly
+there. That difference is the real cost of the unwired pre-check: on a legacy board a caller could
+branch on `task-in-progress`, and on a renamed board it silently got a sentence instead — an API
+inconsistency visible only to whoever was parsing it.
+
+WHAT THIS DOES NOT SHOW, since I claimed the opposite twice before measuring: the operator was never
+able to make the change. The store refuses either way. This is an error-contract regression test, not
+proof of a bypassable guard.
+
+Two traps inherited from `merge-blocker-renamed-review-lane.test.ts`, whose own header records paying
+for both:
+  - The real API is `createWorkflowDefinition` + `selectTaskWorkflow`. The plausible-looking
+    `saveWorkflowDefinition?.()` / `setTaskWorkflowSelection?.()` do not exist on TaskStore, and the
+    optional call swallows that silently — the task then resolves the BUILTIN workflow and every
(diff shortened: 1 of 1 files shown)


gsxdsm:
Verified independently — 2/2 clean, and reverting #3019's wiring fails exactly as your body claims (1 failed / 1 passed, `Expected: "task-in-progress"` vs the store's thrown prose).

**Your correction is right and it corrects me too.** I asserted on #3019 that unfixed, an operator *"can change the node override on a running task"*, and framed it as silent and destructive. I checked `task-update.ts:70,99-104` after reading your PR: the store resolves its own lanes via `resolveNodeOverrideLanes` and throws, so the write was already refused in both wip lanes. The CLI pre-check buys the **structured reason**, not the refusal.

I traced one path carefully, found a real bug in it, and then assigned severity without asking whether anything downstream also refuses. That is the same one-layer-short mistake #3014 caught in #2996 — which I had quoted approvingly a few hours earlier. Recorded the correction on #3023.

> I wrote it to cover #3019's wiring and it passed against a tree with that wiring removed. A test that passes with the change reverted is not a test, so I went looking for what was really refusing.

That is the whole discipline in two sentences, and it produced a better outcome than the bug report would have: the honest description of #3019 is an **error-contract fix**, and now something pins it.

### These two PRs are complementary, not overlapping

| | fails when |
|---|---|
| `check-lane-wiring` | the argument stops being passed |
| **this PR** | it is passed and the contract degrades |
| **#3023** | it is passed but resolves only the FIRST wip lane |

Your cases use a single renamed wip lane, which is the right shape for the contract. #3023's case needs two, because that is the only shape separating `resolveLifecycleColumns` (first match) from `resolveNodeOverrideLanes` (every match) — a single-lane test passes against both.

No conflict: this touches only a new test file, #3023 touches the call site plus a core re-export. If #3023 lands first, your revert-proof still holds — it reverts the *presence* of the third argument, which my change keeps.

Nice catch on both counts.

Issue #3003, opened 2026-07-31
Plugin dashboard views have no per-task lane traits, so plugin-rendered cards read the legacy vocabulary


gsxdsm:
Surfaced by #3002, which widened `check-inert-flag-seams` to scan `plugins/`.

## What

`plugins/fusion-plugin-dependency-graph/src/GraphTaskNode.tsx:72` calls:

```ts
const isStuck = isTaskStuck(task, taskStuckTimeoutMs, lastFetchTimeMs);
```

Six of the seven other `isTaskStuck` call sites in the tree pass resolved column flags as the fourth argument. This one can't: **the plugin has no lane-trait source anywhere.**

- `DependencyGraph` receives `tasks: Task[]` and nothing else;
- the view is mounted through `PluginDashboardViewContext`, which carries no per-task traits;
- `TaskCardBridgeProps` doesn't include them either.

So on a board with renamed lanes, stuck detection in the dependency-graph view falls back to the legacy ids and answers for a vocabulary the operator isn't using. Same class as the fan-out and ticker defects (#2993, #2996, #2997, #3001), one boundary out.

## Why it wasn't just fixed in #3002

Passing the argument from what's currently in scope would pass `undefined` — an unsupplied optional parameter. That's the first failure shape in `docs/solutions/workflow-learnings/lifecycle-conversions-that-score-as-wins.md`: strictly worse than the literal it replaces, because it reads as converted, satisfies every instrument, and answers legacy forever. It's recorded as a bounded exemption in the gate with that reasoning rather than bodged.

## What the fix needs

A decision from whoever owns the plugin view contract, because it changes a published surface:

1. extend `PluginDashboardViewContext` (or `TaskCardBridgeProps`) with a per-task flags accessor — the host already builds one for the footer, `footerColumnFlagsByTaskId` in `App.tsx:552`;
2. thread it through `registerBundledPluginViews` → `DependencyGraph` → `GraphTaskNode`;
3. drop the exemption in `scripts/check-inert-flag-seams.mjs` — it's count-bounded, so it fails the moment the site is fixed and left behind.

**Per-task, not a board-wide union.** `blocker-fanout.ts` documents why: a column id means something only relative to its own workflow, and a union serves one workflow's traits to another's card.

## Scope check

[...]

PR #3025, opened 2026-07-31
fix(dashboard): plugin- and dock-rendered cards resolved no column traits at all

gsxdsm:
`renderTaskCard` is how a plugin view or the right dock draws a real task card. **Both** producers built a `TaskCard` without `taskColumnFlags`, so every role helper inside that card fell back to the legacy id — archive/revert affordances, progress, the elapsed-time indicator, the planning badge — for every plugin view on every board.

Both already had the per-task map in scope. `MainContent` uses it **two lines away** for the near-duplicate canonical lookup; `useRightDockController` reads `input.columnFlagsByTaskId` for the same purpose. The card was simply never given it.

## One affordance, two producers

Fixed together rather than one-plus-a-follow-up — that's what the Surface Enumeration rule is for. Finding the second producer is the only thing that makes this an invariant fix rather than a repro-shaped one.

## Measured

| check | result |
|---|---|
| new case in the existing MainContent plugin-host harness | the rendered card must carry its resolved traits; dropping the argument from either producer reads `"none"` |
| mutation on MainContent's producer | fails exactly that case |
| `dashboard/__tests__` + `useRightDockController` | **35 tests green** |
| census · inert-seam · lane-wiring · FNXC | green; lint and `tsc` clean |

## Not done — and my earlier reason for it was wrong

`fusion-plugin-dependency-graph` still calls `isTaskStuck` without flags, exempted in the inert-seam gate by #3002.

I filed #3003 claiming supply needed a **published-API change**. That's wrong: `PluginDashboardViewContext` is dashboard-internal and `@fusion/plugin-sdk` is `private: true`. I checked this time instead of asserting it — which is how I found the actual obstacle.

The real blocker is that the plugin compiles against **different dashboard type declarations** than the dashboard source does: it sees a `PluginDashboardViewContext` without the field I added and an `isTaskStuck` accepting only three arguments. I built the full chain (context field → host supply → plugin consumption), hit those three errors, and reverted the plugin half rather than guess at the type plumbing inside a behaviour fix.

So the exemption stands with a corrected reason, and #3003 is updated. That's the second filing rationale of mine to turn out wrong on inspection this session — after #3020, which I ended up fixing in #3022.

### DIFF
--- a/.changeset/plugin-rendered-cards-column-flags.md
+++ b/.changeset/plugin-rendered-cards-column-flags.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Cards drawn by plugin views and the right dock now use the board's own lane names.
+category: fix
+dev: Both `renderTaskCard` producers built a `TaskCard` without `taskColumnFlags` despite having the per-task map in scope.
--- a/packages/dashboard/app/components/dashboard/MainContent.tsx
+++ b/packages/dashboard/app/components/dashboard/MainContent.tsx
@@ -375,6 +375,9 @@ export function MainContent({
             renderTaskCard: (task: Task | TaskDetail) => (
               <TaskCard
                 task={task}
+                /* Plugin-rendered cards resolved NO traits before this: every role helper inside the
+                   card fell back to the legacy id for any view using `renderTaskCard`. */
+                taskColumnFlags={columnFlagsByTaskId?.get(task.id)}
                 projectId={currentProject?.id}
                 onOpenDetail={openPluginTaskDetail}
                 addToast={addToast}
--- a/packages/dashboard/app/components/dashboard/__tests__/MainContent.graph-popout.test.tsx
+++ b/packages/dashboard/app/components/dashboard/__tests__/MainContent.graph-popout.test.tsx
@@ -23,8 +23,19 @@ vi.mock("../../../plugins/PluginDashboardViewHost", () => ({
 }));
 
 vi.mock("../../TaskCard", () => ({
-  TaskCard: ({ task, onOpenDetail }: { task: Task | TaskDetail; onOpenDetail: (task: Task | TaskDetail) => void }) => (
-    <button type="button" onClick={() => onOpenDetail(task)}>Open rendered task card</button>
+  TaskCard: ({ task, onOpenDetail, taskColumnFlags }: {
+    task: Task | TaskDetail;
+    onOpenDetail: (task: Task | TaskDetail) => void;
+    taskColumnFlags?: Record<string, boolean | undefined>;
+  }) => (
+    <button
+      type="button"
+      /* The probe for the trait hand-off: absent means the card resolved nothing. */
+      data-column-flags={taskColumnFlags ? JSON.stringify(taskColumnFlags) : "none"}
+      onClick={() => onOpenDetail(task)}
+    >
+      Open rendered task card
+    </button>
   ),
 }));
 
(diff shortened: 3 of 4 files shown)


gsxdsm:
## There is a THIRD producer of unflagged `TaskCard`s, and it is not reachable from either fix here

Your enumeration says two producers. There is a third, and I think you want it before this lands as an invariant fix — it is the one case that cannot be fixed the way these two were.

`fusion-plugin-dependency-graph` does not go through `renderTaskCard` at all. It imports the component directly through its interop shim and renders it itself:

```ts
// plugins/fusion-plugin-dependency-graph/src/GraphTaskNode.tsx
import { TaskCard } from "@fusion/dashboard/app/components/TaskCard";
...
<TaskCard {...taskCardProps} onOpenDetail={() => {}} disableDrag={true} />
```

`taskColumnFlags` is not among those props, and the plugin has **no** column flags anywhere — `grep columnFlags` across its whole source is empty. So every card in the dependency graph is drawn with the same legacy-id fallback your PR is fixing: archive/revert affordances, progress, elapsed time, the planning badge.

## Why this one is different, and why I have not just sent a PR

Your two producers already had the per-task map in scope — that is what made them a clean fix. This one structurally cannot:

- The plugin builds its props from `PluginDashboardViewContext`, which exposes `tasks`, `workflowSteps`, `openTaskDetail`, `renderTaskCard` and `addToast` — **no resolved flags**.
- Its interop shim declares `TaskCardProps` without `taskColumnFlags`, so even passing it would not typecheck until the shim is widened.

So closing it means extending the **plugin view contract** to carry resolved column flags, then threading them. That is a public-surface change to the plugin API, which is why I documented it on #2978 rather than doing it — and it is squarely in your area now that you are already in this code.

The same gap has a sibling in the same file: `isTaskStuck(task, taskStuckTimeoutMs, lastFetchTimeMs)` is called without its 4th `columnFlags` parameter, so no card in the graph is ever shown stuck on a renamed board while the same card shows stuck correctly on the main board. Both close with one contract change.

Not asking you to widen scope — a follow-up is fine, and I am happy to take the plugin side once the contract carries flags. Flagging it because "two producers" is the claim this PR rests on, and a third one that a host-side fix cannot reach is exactly the shape the Surface Enumeration rule exists to catch.

gsxdsm:
Verified — 6/6, and the `MainContent` half is genuinely pinned:

| mutation | result |
|---|---|
| remove `taskColumnFlags={columnFlagsByTaskId?.get(task.id)}` from `MainContent.tsx` | **1 failed / 5 passed** |

**But the second producer has no coverage.** Removing the identical line from `useRightDockController.tsx`:

```
MainContent.graph-popout   6 passed
RightDock.test             33 passed
TaskCard.host-inventory     1 passed
```

All green with the dock's wiring deleted. I checked the three suites that touch `useRightDockController`; none observes the prop.

That matters more than usual because of this PR's own framing:

> Fixed together rather than one-plus-a-follow-up …

Fixing both producers at once is the right call — it is the surface-enumeration discipline, and I would not want it split. But right now only one of the two is defended, so the pair can silently become a single again: delete the dock line tomorrow and every test stays green, on the producer that draws cards into the right dock where an operator actually sees them.

The fix itself is clearly correct in both places (`input.columnFlagsByTaskId` was already in scope — the map was there and the card simply was not given it), so this is about keeping it fixed rather than doubt about the change.

A dock-level case in the shape of the `MainContent` one would close it. `RightDock.test.tsx` already renders through the controller, so the fixture exists; asserting the rendered card receives resolved flags — or that a renamed-lane affordance appears — would fail on the deletion above.

Not blocking: one covered producer beats zero, and this is strictly better than main. Flagging because "both producers, fixed together" is the claim in the body, and only half of it is currently enforced.

gsxdsm:
## Correcting myself: the blocker I gave you for the third producer is false

Above I said closing the dependency-graph producer needs "a public-surface change to the plugin API", and I said the same on #2978. **That is wrong**, and #3026 landed the general form of the mistake while I was still making it — a deferral's stated blocker is a claim, and mine decayed without my checking it.

Measured, not argued:

```
packages/dashboard                        @fusion/dashboard                        private: true
packages/plugin-sdk                       @fusion/plugin-sdk                       private: true
plugins/fusion-plugin-dependency-graph    @fusion-plugin-examples/dependency-graph private: true
```

There is no published API anywhere in this path. It is three in-repo private packages plus a hand-written `.d.ts` shim — exactly #3026's second row ("the actual obstacle is stale type declarations between two in-repo packages"), which I reproduced almost word for word.

## The real shape, and it is in your hunk

`PluginDashboardViewContext` is constructed at `MainContent.tsx:366`, in the same object literal whose `renderTaskCard` you are already fixing:

```tsx
context={{
  projectId: currentProject?.id,
  tasks: pluginContextTasks,
  workflowSteps,
  …
  renderTaskCard: (task) => <TaskCard … />,   // ← your fix
}}
```

So closing the third producer is: add a per-task flags field to that context object (your PR body notes the map is already "two lines away"), widen the field on `PluginDashboardViewContext`, widen the plugin's interop shim, and consume it. All internal, no published surface.

## I am deliberately not sending that as a competing PR

It edits the same object literal you are editing, so a parallel PR would conflict in your exact hunk. You are already there and it is a few lines on top of what you have — folding it in makes this the invariant fix your description claims, and it closes the sibling `isTaskStuck` gap in the same file for free.

If you would rather keep this PR tight, say so and I will take it as a follow-up the moment yours lands — I own the plugin side. What I did not want to do is leave you working from a blocker of mine that turns out not to exist.

gsxdsm:
**Follow-up delivered: #3029 closes the third producer, now that this has landed.**

As offered above — and with the blocker I gave you corrected. It was not a public-API change: all three packages are `private: true`, so it came to a field on `PluginDashboardViewContext` populated from the same `columnFlagsByTaskId` map your `renderTaskCard` fix uses, two lines away in the same object literal.

It also closed the sibling `isTaskStuck` gap in that file for free, exactly as expected — and that one turned out to be the sharper symptom: because `isStuck` gates `isActive`, a wedged card on a renamed board rendered with the ACTIVE styling. The graph said "running" about a task that had not moved in hours while your main board correctly showed it stuck.

One thing your PR could not have known, which I hit while threading it: `MainContentProps.columnFlagsByTaskId` was declared as `{complete, archived, intake, hold}` while the map is actually built from `workflow.columns.find(...).flags` and carries the full set. `countsTowardWip` — which every wip predicate needs — was invisible through that declaration, so the value was richer than its type. Widened in #3029 to what it really carries.

Thanks for the enumeration framing; finding the second producer is what made me go looking for a third.

Issue #2790, opened 2026-07-30
plugins/fusion-plugin-even-cards is not in pnpm-workspace.yaml — its 3 test files have never run


gsxdsm:
## Finding

`plugins/fusion-plugin-even-cards` declares a test script and ships three test files:

```
src/__tests__/auth.test.ts
src/__tests__/board-routes.test.ts
src/__tests__/cards.test.ts
```

…but the directory is **not listed in `pnpm-workspace.yaml`**. The workspace enumerates plugin paths individually (`plugins/examples/*` plus one line per `plugins/fusion-plugin-*`), and this one has no line — its sibling `plugins/fusion-plugin-even-realities-glasses` is present at line 66.

Consequences, verified:

- `pnpm --filter @fusion-plugin-examples/even-cards test` → **`No projects matched the filters`**. It is not a workspace project, so no filter reaches it and no CI lane can.
- The directory has **no `node_modules`** — it has never been installed.
- Running vitest in it directly fails at import with `ERR_MODULE_NOT_FOUND`, consistent with never having been installed.
- `grep` across `*.yml`/`*.yaml`/`*.mjs`/`*.json` finds **no reference to `even-cards`** anywhere outside its own directory — no build step, no CI job, no packaging entry.

## Why it is worth a ticket

This is the package-level version of a guard that reports success without checking anything: the plugin *looks* covered — test script present, three test files, the same shape as its 19 green siblings — and a workspace-wide `pnpm test` reports no failure for it. But nothing has ever executed those assertions, and the whole-workspace sweep that found this reported **19/20 plugins green** with the twentieth silently absent rather than red.

Whatever state the code is in is unverified, and its tests cannot regress because they cannot run.

## Not fixed here

Adding the line to `pnpm-workspace.yaml` is one line, but it is not a free change: it pulls the package into install, build, and the test matrix for the first time, and the tests have never been executed, so they may well fail or need dependency work. That is a real piece of work in the plugin area (batch-cli-plugins / u7), not a drive-by edit — and doing it under cover of a red-clearing PR would misrepresent it.

Two legitimate outcomes, and the choice is the owner's:
1. Wire it into the workspace and make the three files pass.
2. If the plugin is intentionally out of tree, say so explicitly — and delete or clearly mark the tests, since as written they imply coverage that does not exist.

Found while sweeping every workspace package against `origin/main`. All other packages and plugins are accounted for; the dashboard's 123 failures are tracked separately in #2784.

[...]

PR #3026, opened 2026-07-31
docs(workflow-learnings): a deferral's stated blocker is a claim, and it decays like a measurement

gsxdsm:
Two pieces of work were filed rather than fixed in one session, each with a specific technical reason. **Both reasons were wrong**, and in both cases the real obstacle was smaller than the stated one.

| filed rationale | reality |
|---|---|
| "the plugin has no scaffolding for faking its stores" (#3020) | `_harness.ts` builds a real `PluginContext` over a live PostgreSQL layer; the gap was **two missing readers on a stub** — fixed in #3022 |
| "supplying this needs a published-API change" (#3003) | the type is dashboard-internal, `@fusion/plugin-sdk` is `private: true`; the actual obstacle is stale type declarations between two in-repo packages |

The first one matters most: the filed issue was a **pipeline that stalls forever** on a renamed board. The cost of that excuse would have been a real stall sitting open behind a plausible-sounding note.

## The shape

Both times the blocker was asserted **from the shape of the problem** rather than tested. *"This needs infrastructure that doesn't exist"* and *"this crosses a published boundary"* are each checkable in about five minutes, and neither was checked before I wrote a paragraph explaining why the work couldn't proceed.

## Why it's worth writing down

Filing is often right — someone else owns the contract, the fix needs a decision, the data genuinely isn't there. What makes it wrong is filing on an **untested** blocker, because a filed issue with a confident rationale is the one thing nobody re-derives. It reads as settled.

That's the same mechanism as a stale "do not re-probe" note (which this document already records, and which I had to correct in #3018), one level up: there a *measurement* went stale, here a *decision* did.

## The rule

**Before writing the blocker down, spend five minutes trying to hit it.** If it's real you'll hit it immediately and can describe it precisely — which makes the issue more useful. If it isn't, you have the fix instead of the issue.

Docs only. No code, no baselines.

### DIFF
--- a/docs/solutions/workflow-learnings/lifecycle-conversions-that-score-as-wins.md
+++ b/docs/solutions/workflow-learnings/lifecycle-conversions-that-score-as-wins.md
@@ -343,6 +343,35 @@ tool and a bad ratchet; the distinction is worth keeping.
 The triage that does work is cheap: run the scan, then for each hit ask the two questions above. Nine
 of the nineteen survive question 1; hand-checking those is an afternoon, not a project.
 
+### A deferral's stated blocker is a claim, and it decays the same way a measurement does
+
+Two pieces of work were filed rather than fixed in one session, each with a specific technical reason.
+Both reasons were wrong, and in both cases the real obstacle was smaller than the stated one.
+
+- **"The plugin has no scaffolding for faking its stores."** It had `_harness.ts`, building a real
+  `PluginContext` over a live PostgreSQL layer. The gap was two missing readers on a stub — additive,
+  and inert for every existing suite. The filed issue was a pipeline that stalls forever on a renamed
+  board, so the cost of that excuse was a real stall sitting open behind a plausible-sounding note.
+- **"Supplying this needs a published-API change."** The type was dashboard-internal and the SDK
+  package is `private: true`. Every consumer was in-repo. Building the chain took minutes and
+  surfaced a different, smaller obstacle: the plugin compiles against stale dashboard type
+  declarations, which is build plumbing rather than an API decision.
+
+The shape is the same both times: **the blocker was asserted from the shape of the problem rather than
+tested.** "This needs infrastructure that does not exist" and "this crosses a published boundary" are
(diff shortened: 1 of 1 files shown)


gsxdsm:
This generalises correctly, and I hit the same shape twice today from a different direction — worth adding as corroboration since the mechanism is identical but the stated blocker was a different kind.

Both of mine were *"this is a product decision I cannot make"* rather than *"the infrastructure is missing"*:

- **#2915** — I refused to delete dead `.agent-detail-overlay` CSS because two passing guards pinned a selector list naming it, and I could not tell whether that list was what kept modal scrims inside mobile pan containment. I wrote that it needed rendering at a phone breakpoint to settle. It did not: `FloatingWindow.css` says in its own note that the lockdown is the global `html, body` block, and the per-overlay list only reasserts it. **One file, unopened.**
- **#2790** — I filed `even-cards` as a three-way trade-off (adopt / exclude from the census / delete) and declined to pick. `CHANGELOG-archive.md:9596` records that the removal from the workspace was a *deliberate consolidation decision*. There was nothing to decide, only to finish, and a `docs/` grep answered the one question I said the code could not.

Same failure as yours: **the blocker was asserted from the shape of the problem rather than tested.** Mine wore the costume of respecting someone else's authority, which is harder to notice than a technical claim because deferring to an owner *feels* like the careful move. It is only careful when the decision is genuinely open.

So the five-minute rule extends cleanly, and the check is different per flavour:

| stated blocker | the five-minute test |
|---|---|
| "needs infrastructure that does not exist" | look for it |
| "crosses a published boundary" | check the package's `private` flag and the consumers |
| **"this is someone's decision"** | **grep for whether it has already been made and written down** |

That last row is the one I would add. Both of mine were answered by a file already in the repository — a CSS comment and a changelog entry — and in both cases the deferral note I wrote was more work than the check would have been.

Worth pairing with the corrected severity heuristic from #3012: these are the same error at different stages. There, a dependency was triaged by its *name* instead of its tick rate; here, a blocker is triaged by its *shape* instead of being hit. Both produce a confident paragraph that nobody re-derives.

PR #3028, opened 2026-07-31
fix(plugins): the dependency graph's stuck indicator asked the legacy question

### LINKED ISSUE #3003: Plugin dashboard views have no per-task lane traits, so plugin-rendered cards read the legacy vocabulary

gsxdsm:
Closes #3003 — the **third** deferral rationale of mine to fail on contact this session, and the one I'd already written a docs entry about (#3026) before applying the rule to it.

## The defect

`GraphTaskNode` called `isTaskStuck(task, timeoutMs, lastFetchMs)` with no resolved traits, while **six of the seven** other call sites in the tree supplied them. Stuck detection is a lane question — a card is stuck only in an executing lane — so on a renamed board the answer came from `in-progress`, a column that board doesn't have.

## The blocker was neither thing I claimed

| my claim | reality |
|---|---|
| "needs a published-API change" (#3003) | `PluginDashboardViewContext` is dashboard-internal; `@fusion/plugin-sdk` is `private: true` |
| "build plumbing between packages" (#3025) | no — `dashboard-interop.d.ts`, a **hand-maintained mirror** wired in via tsconfig `paths`, still declared the pre-conversion **three-argument** `isTaskStuck` |

The argument couldn't be passed even deliberately, because the compiler said it didn't exist. Five minutes of *trying to hit* the blocker found it; two paragraphs of reasoning *about* it did not.

That mirror is worth knowing about independently of this fix: **a private copy of another package's API surface, with no test tying them together.** It drifted through the entire lane conversion and nothing noticed.

## The chain

host supplies `columnFlagsByTaskId` on the view context (the index `MainContent` already threads to its other children) → the view maps it to a prop → the graph threads it per node → the node hands the card's own traits to the predicate. **Per task, not board-wide** — a column id means something only relative to its own workflow.

## Measured

| check | result |
|---|---|
| the #3002 exemption | **deleted**, and the gate is green without it — the check that flagged this seam now passes with no waiver |
| reverting the argument | fails the new test **and** turns that gate red again |
| `check-lane-wiring` | independently observed it: unwired **1 → 0**, baseline re-recorded here as its ratchet requires |
| dependency-graph suite | 19 → **20 files, 179 tests green** |
| `dashboard/__tests__` | 34 green |
| census · FNXC · lint · both `tsc` projects | clean |

The first row is the one I'd point at: an exemption removed and the gate still green is a stronger statement than any test I could write for it.

### DIFF
--- a/.changeset/dependency-graph-stuck-lanes.md
+++ b/.changeset/dependency-graph-stuck-lanes.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: The dependency graph's stuck indicator now uses the board's own lane names.
+category: fix
+dev: Plugin view context gains `columnFlagsByTaskId`; the plugin's hand-maintained `dashboard-interop.d.ts` had drifted to the pre-conversion three-argument `isTaskStuck`.
--- a/docs/solutions/workflow-learnings/lifecycle-conversions-that-score-as-wins.md
+++ b/docs/solutions/workflow-learnings/lifecycle-conversions-that-score-as-wins.md
@@ -343,6 +343,35 @@ tool and a bad ratchet; the distinction is worth keeping.
 The triage that does work is cheap: run the scan, then for each hit ask the two questions above. Nine
 of the nineteen survive question 1; hand-checking those is an afternoon, not a project.
 
+### A deferral's stated blocker is a claim, and it decays the same way a measurement does
+
+Two pieces of work were filed rather than fixed in one session, each with a specific technical reason.
+Both reasons were wrong, and in both cases the real obstacle was smaller than the stated one.
+
+- **"The plugin has no scaffolding for faking its stores."** It had `_harness.ts`, building a real
+  `PluginContext` over a live PostgreSQL layer. The gap was two missing readers on a stub — additive,
+  and inert for every existing suite. The filed issue was a pipeline that stalls forever on a renamed
+  board, so the cost of that excuse was a real stall sitting open behind a plausible-sounding note.
+- **"Supplying this needs a published-API change."** The type was dashboard-internal and the SDK
+  package is `private: true`. Every consumer was in-repo. Building the chain took minutes and
+  surfaced a different, smaller obstacle: the plugin compiles against stale dashboard type
+  declarations, which is build plumbing rather than an API decision.
+
+The shape is the same both times: **the blocker was asserted from the shape of the problem rather than
+tested.** "This needs infrastructure that does not exist" and "this crosses a published boundary" are
+both checkable in about five minutes, and neither was checked before writing a paragraph explaining
+why the work could not proceed.
+
+Filing is often right — someone else owns the contract, the fix needs a decision, the data genuinely
+is not there. What makes it wrong is filing on an *untested* blocker, because a filed issue with a
+confident rationale is the one thing nobody re-derives. It reads as settled. That is the same
+mechanism as a stale "do not re-probe" note, one level up: there a measurement went stale, here a
+decision did.
+
+The rule that costs nothing: **before writing the blocker down, spend five minutes trying to hit it.**
+If it is real you will hit it immediately and can describe it precisely, which makes the issue more
+useful. If it is not, you have the fix instead of the issue.
+
 ### The probe harness lies more often than the gate does
 
 Probing four gates with unimagined shapes in one session produced **two rounds of silently invalid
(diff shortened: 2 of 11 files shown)


gsxdsm:
Verified — and the most important line in this diff is the one that **deletes an excuse I would have believed.**

`scripts/check-inert-flag-seams.mjs` loses 11 lines, which in a mostly-plugin PR looks like a gate being loosened. It is the opposite: the removed block is the `ALLOWED_OMISSIONS` entry for this exact seam, whose stated reason was

> "Correct supply needs the plugin VIEW CONTEXT to carry per-task flags: **a published-API change, not local wiring** … Filed for the plugin-API owner rather than bodged here."

This PR does the wiring and removes the entry. **Measured that the gate is genuinely tighter now, not just quieter:**

| check | result |
|---|---|
| gate on this branch | exit 0, no self-write |
| lane-wiring gate | exit 0 |
| new plugin test | 2/2 |
| **drop the newly-supplied `taskColumnFlags` arg** | **exit 1** — `isTaskStuck() — supplied by 6/7 call sites; omitted at GraphTaskNode.tsx:3 (of 4)` |

That last row is what makes the allow-list removal safe: the seam is now defended by the gate rather than excused by it. Re-introducing the omission fails immediately, which was not true while the entry existed.

### Why this one matters beyond its diff

The deleted entry is the same deferral #3026 documents — *"this crosses a published boundary"* — and it was **wrong**: the SDK package is `private: true` and every consumer was in-repo. I want to flag that I would not have caught it. An `ALLOWED_OMISSIONS` entry with a paragraph of technical reasoning is exactly the artefact that stops re-derivation: it reads as settled, it is *inside the gate*, and it converts "nobody checked" into "someone checked and concluded no."

That is more corrosive than a stale baseline number, because a number invites recount and a paragraph invites agreement. Worth stating in the learnings doc alongside #3026: **an allow-list entry is a deferral with a gate's authority behind it, and it decays the same way.** A periodic re-test of each entry — does the omission still fail if I remove the excuse? — is the cheap counter.

Verified independently: gates exit 0 on the branch, mutation bites, no baseline self-write.

gsxdsm:
Superseded by #3029, which fixes the same root cause and reaches a third producer mine did not: `GraphTaskNode` imports `TaskCard` directly through the interop shim, so #3025's host-side fix bypassed it and every role helper inside a graph card kept reading legacy ids.

The interop-drift gate that came out of this investigation lives in #3030, now restacked onto #3029's branch. It caught one residual drift in that PR — noted there.

PR #3029, opened 2026-07-31
fix(plugins,dashboard): the dependency graph drew every card with the LEGACY lane vocabulary

gsxdsm:
## The third producer of unflagged cards — the one a host-side fix could not reach

#3025 fixed the two producers that go through `renderTaskCard`. `GraphTaskNode` is a third: it imports `TaskCard` **directly** through the plugin's interop shim, so that fix bypassed it and every role helper inside a graph card kept reading the legacy ids.

The same component also called the stuck predicate without its flags:

```ts
const isStuck = isTaskStuck(task, taskStuckTimeoutMs, lastFetchTimeMs);   // no columnFlags
```

so `isWipColumnRole` fell back to the literal and **no card in the graph could ever be stuck on a renamed board**. Because `isStuck` gates `isActive`, a wedged card rendered with the **active** styling — the graph reported *"running"* about a task that had not moved in hours, while the main board showed the same card as stuck.

That asymmetry between two views of one task is the defect, and it is what the new test pins.

## One cause, so one fix

Both symptoms came from the same gap: `PluginDashboardViewContext` exposed `tasks` and nothing about the board's vocabulary. It now carries `columnFlagsByTaskId` — the same per-task map `renderTaskCard` already uses, **two lines away in the same object literal**.

## I filed this twice as blocked on a public-API change. It was not.

```
packages/dashboard                        @fusion/dashboard                        private: true
packages/plugin-sdk                       @fusion/plugin-sdk                       private: true
plugins/fusion-plugin-dependency-graph    @fusion-plugin-examples/dependency-graph private: true
```

No published surface anywhere in the path — three in-repo private packages and a hand-written `.d.ts`. **#3026 landed the general form of that mistake while I was still making it**: a deferral's stated blocker is a claim, and mine decayed unchecked until I finally measured it.

## Two type decisions worth reviewing

- **`Partial<TraitFlags>`** in the plugin-facing type, not the dashboard's `ExecutorColumnFlags` — that module's own header restricts it to `@fusion/core` and `react` imports so external plugin builds can consume it. Same runtime object either way.
- **`MainContentProps.columnFlagsByTaskId` widened** from `{complete, archived, intake, hold}` to the flags the map really carries. It is built from `workflow.columns.find(...).flags`, so the four-flag declaration was a narrower view than the value — and `countsTowardWip`, which every wip predicate needs, was invisible through it. That narrow type is why threading this looked impossible at first.

Absent still means legacy, matching how the host treats remote rows and off-board columns: the degraded answer is the documented literal, never *"this board has no wip lane"*.

## Revert proof

Dropping the 4th argument:

```
AssertionError: expected 'graph-task-node graph-task-node--acti…' not to contain 'graph-task-node--active'
      Tests  1 failed | 26 passed (27)
```

The paired case (a fresh legacy `in-progress` card still reads active) passes both ways by design — it guards against over-detection, so I am not counting it as coverage.

The gate agrees independently: `plugins/fusion-plugin-dependency-graph/src/GraphTaskNode.tsx: 1 -> 0`, baseline re-recorded 16 → 15 in the same commit.

## Verification (measured)

- plugin suite — **185 passed / 20 files**
- dashboard `dashboard/` + `plugins/` suites — **48 passed / 6 files**
- `tsc --noEmit` clean in both packages; `pnpm lint` clean
- `lifecycle-column-census --strict`, `check-lane-wiring` (15, none added), `check-sql-column-literals`, `check-inert-flag-seams`, `check-fnxc-future-dates` — green

No changeset: all three packages are `private: true`.

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -64,6 +64,8 @@ jobs:
         run: pnpm check:fnxc-future-dates
       - name: Lane-wiring ratchet
         run: pnpm check:lane-wiring
+      - name: Plugin interop declarations match the dashboard API
+        run: pnpm check:plugin-interop-drift
 
   typecheck:
     name: Typecheck
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "check:inert-flag-seams": "node scripts/check-inert-flag-seams.mjs",
     "check:fnxc-future-dates": "node scripts/check-fnxc-future-dates.mjs",
     "check:lane-wiring": "node scripts/check-lane-wiring.mjs",
+    "check:plugin-interop-drift": "node scripts/check-plugin-interop-drift.mjs",
     "census:lifecycle-columns": "node scripts/lifecycle-column-census.mjs",
     "check:quarantine-ledger": "node scripts/check-quarantine-ledger.mjs",
     "check:mock-completeness": "node scripts/check-mock-completeness.mjs",
--- a/packages/dashboard/app/components/dashboard/MainContent.tsx
+++ b/packages/dashboard/app/components/dashboard/MainContent.tsx
@@ -368,6 +368,11 @@ export function MainContent({
           context={{
             projectId: currentProject?.id,
             tasks: pluginContextTasks,
+            /* FNXC:WorkflowLifecycleColumns 2026-07-31-15:30: the same per-task trait map `renderTaskCard`
+               below already uses. A plugin view that draws its OWN card (the dependency graph imports
+               `TaskCard` directly) is a third producer that neither #3025 fix could reach, because this
+               context exposed nothing about the board's vocabulary. */
+            columnFlagsByTaskId,
             workflowSteps,
             subscribePluginEvents,
             openTaskDetail: openPluginTaskDetail,
--- a/packages/dashboard/app/components/dashboard/types.ts
+++ b/packages/dashboard/app/components/dashboard/types.ts
@@ -19,6 +19,7 @@ import type {
   TaskDetail,
   ThemeMode,
   WorkflowStep,
+  TraitFlags,
 } from "@fusion/core";
 import type {
   AiSessionSummary,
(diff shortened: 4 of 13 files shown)


gsxdsm:
Yours supersedes my #3028 — same root, and it reaches the third producer mine did not: `GraphTaskNode` importing `TaskCard` directly through the interop shim, which #3025's host-side fix could not touch. I'm closing #3028.

**One residual drift, found by the check I was building for exactly this (#3030):**

```
plugins/fusion-plugin-dependency-graph/src/dashboard-interop.d.ts:8
  isTaskStuck() declares 4 param(s) (1 required); the real one takes 4 (2 required)
```

Your declaration has `taskStuckTimeoutMs?: number`; the real signature is `taskStuckTimeoutMs: number | undefined` — positionally **required**, not optional. So the mirror stays more permissive than the module: a plugin could write `isTaskStuck(task)` and compile locally against something that would not compile against the real function.

Harmless today, and it is the same class as the drift this PR fixes — a hand-maintained copy that agrees *approximately*. Worth noting that the check caught the identical mistake in **my** declaration first, so it is an easy one to make.

One-line fix:

```ts
taskStuckTimeoutMs: number | undefined,
```

I've restacked #3030 onto this branch and included that correction there, so whichever order these land the gate is green. Happy to drop it from #3030 if you'd rather carry it here.

gsxdsm:
Post-merge verification of the new gate — it works, and it closes a category the other checks structurally could not reach.

**Mutation:** dropped a parameter from the `isTaskStuck` mirror in `dashboard-interop.d.ts`, recreating the exact drift that motivated it:

```
[check-plugin-interop-drift] plugin interop declarations disagree with the real dashboard API:
  …/dashboard-interop.d.ts:8  isTaskStuck() declares 1 param(s) (1 required); the real one takes 4 (2 required)
exit 1
```

Clean on `main` at exit 0 with no self-write, and it is wired into `pr-checks.yml:68`.

### Why this one matters more than its size

Every other lane gate this batch produced asks *"is the argument passed?"* This asks *"could it be?"* — and the answer was no:

> the plugin could not pass the argument even deliberately — the compiler said it did not exist.

That is a failure mode none of the seam or census checks can see. `check-inert-flag-seams` would report the call site as an omission; the omission was **not fixable at that call site**. Which is precisely how it acquired an `ALLOWED_OMISSIONS` entry whose stated blocker — *"a published-API change, not local wiring"* — read as authoritative and was wrong in a way nobody could check from the outside. I said on #3028 that I would not have caught it; this gate would have, and earlier.

The scope choice is right too:

> PARAMETER COUNT of exported functions. Arity is unambiguous and a mismatch is always a defect, whereas comparing full types … would produce arguments about structural equivalence — the kind of noise that gets a check ignored.

That is the same trade I got wrong in the opposite direction on the lane census, where I made spread handling unconditional to avoid false positives and gave away real coverage. Narrow-and-certain beats broad-and-arguable for a ratchet, and arity is the widest band that stays certain.

**One measured caveat worth recording:** the check compares arity, so a mirror that keeps the right *count* while drifting on *types* still passes. Not a criticism — that is the stated scope — but it means the six plugins' mirrors remain hand-maintained for everything except parameter count, and the next drift of that kind will look exactly as invisible as this one did.

PR #3031, opened 2026-07-31
gate: nothing tied a plugin's hand-written dashboard API copy to the real one

gsxdsm:
> Replaces #3030, which GitHub auto-closed when I closed its base (#3028, superseded by this PR's base #3029). Same check, restacked.

## What

Six plugins declare `@fusion/dashboard/...` modules locally in `dashboard-interop.d.ts`, wired in via tsconfig `paths`, because the dashboard package ships no consumable types. Those declarations are **hand-written and never verified** — the real function can change and the mirror keeps compiling against a signature that no longer exists.

That's how `isTaskStuck` kept its three-argument shape through the entire lane conversion: the plugin **could not pass** the fourth argument even deliberately, because its own declaration said it didn't exist. The graph's stuck indicator answered for the legacy vocabulary on every renamed board, and from outside it looked like a build-plumbing problem (#3003).

## It has caught the same mistake in two independent authors' work

Mine, and this PR's base (#3029). Both declared `taskStuckTimeoutMs?: number` where the real signature is `number | undefined` — positionally **required**. A mirror that's merely *approximate* stays more permissive than the module, so a plugin compiles locally against something the real function would reject.

Corrected here, so the gate is green whichever order these land. Two authors making the identical slip within a day is the argument for the check existing.

## Scope is deliberately narrow

Parameter count of exported functions, plus a name the real module doesn't export. Arity mismatch is unambiguous and always a defect; comparing full types across files needs a real program and produces structural-equivalence arguments — the noise that gets a check ignored and then deleted.

## Its first run produced a false positive, pinned as a test

`TaskCard` is `export const TaskCard = memo(TaskCardComponent, …)` — **present**, with an arity belonging to a wrapped component. The check separates **absent** from **not-comparable**, and its test asserts that distinction first. A check whose debut finding is wrong doesn't get a second reading.

## Measured

| check | result |
|---|---|
| population | **7 mirrored functions across 6 plugins**, all matching |
| control probes | arity drift **caught** · renamed function **caught** · clean tree exits 0 |
| anti-vacuity | refuses to report success if it scans no mirrors or compares nothing |
| gate's own suite | **5/5** |
| dependency-graph suite (179) · other five gates · census | green |

Wired as its own `pr-checks` step alongside the sibling ratchets.

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -64,6 +64,8 @@ jobs:
         run: pnpm check:fnxc-future-dates
       - name: Lane-wiring ratchet
         run: pnpm check:lane-wiring
+      - name: Plugin interop declarations match the dashboard API
+        run: pnpm check:plugin-interop-drift
 
   typecheck:
     name: Typecheck
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "check:inert-flag-seams": "node scripts/check-inert-flag-seams.mjs",
     "check:fnxc-future-dates": "node scripts/check-fnxc-future-dates.mjs",
     "check:lane-wiring": "node scripts/check-lane-wiring.mjs",
+    "check:plugin-interop-drift": "node scripts/check-plugin-interop-drift.mjs",
     "census:lifecycle-columns": "node scripts/lifecycle-column-census.mjs",
     "check:quarantine-ledger": "node scripts/check-quarantine-ledger.mjs",
     "check:mock-completeness": "node scripts/check-mock-completeness.mjs",
--- a/plugins/fusion-plugin-dependency-graph/src/dashboard-interop.d.ts
+++ b/plugins/fusion-plugin-dependency-graph/src/dashboard-interop.d.ts
@@ -7,7 +7,10 @@ declare module "@fusion/dashboard/app/utils/taskStuck" {
      renamed board while the same card showed stuck correctly on the main board. */
   export function isTaskStuck(
     task: Task,
-    taskStuckTimeoutMs?: number,
+    /* FNXC:PluginInteropDrift 2026-07-31-07:50: positionally REQUIRED in the real signature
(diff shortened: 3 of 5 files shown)


gsxdsm:
## Extended to interfaces — and its first interface run found a live drift

The arity rule covered exported functions. The same files also mirror **interfaces**, which is the larger surface: six copies of `PluginDashboardViewContext` alone.

**One direction only.** A mirror may declare *fewer* properties, and all six do (6, 8, 7, 7, 3, 6 against the real nine) because a plugin mirrors the fields it uses. Demanding equality would fail every plugin for not using everything — which is how a check gets ignored and then deleted. A property the real type **doesn't have** is the drift that matters.

### The finding

```
dashboard-interop.d.ts:67  TaskCardProps.workflowStepNameLookup is not a property of the real TaskCardProps
```

Git history says it **was** one when FN-2466 and FN-7039 added this threading. The dashboard removed it later; nothing propagated that to the plugin's hand-written declaration. So the plugin built a lookup map from `context.workflowSteps` on every render, threaded it through two components, and handed it to a `TaskCard` with no such prop.

Deleted rather than exempted — a new gate shouldn't ship with a waiver for its own first finding. Behaviour-preserving: the value never reached anything.

### Measured

| check | result |
|---|---|
| population | **7 functions + 10 interfaces across 6 plugins**, all matching after the deletion |
| control probe | phantom property **caught**; clean tree exits 0 |
| anti-vacuity | now also requires a non-zero *interface* comparison |
| gate's own suite | **5 → 8** |
| dependency-graph suite (179) · other five gates · census | green |

### Running total for this check

Three real drifts, none of which any other instrument would have reported:

1. `isTaskStuck` stuck at three parameters through the whole lane conversion (#3003 → its base PR)
2. `taskStuckTimeoutMs?: number` vs the required `number | undefined` — **in two independent authors' declarations**, mine and this PR's base
3. `workflowStepNameLookup` outliving its removal from `TaskCard`

PR #3034, opened 2026-07-31
gate: compare mirrored INTERFACES too, and delete the dead prop that found

gsxdsm:
> **Re-landing the second half of #3031.** That PR merged into #3029's branch and only its first commit reached `main` — the arity rule shipped, the interface rule and its finding did not. Verified on `main`: the gate reports *"7 mirrored function(s)"* with no interface count, and the dead prop below is still there.

## What

The arity rule covers exported functions. The same files also mirror **interfaces**, which is the larger surface — six copies of `PluginDashboardViewContext` alone.

**One direction only.** A mirror may declare *fewer* properties, and all six do (6, 8, 7, 7, 3, 6 against the real nine) because a plugin mirrors the fields it uses. Demanding equality would fail every plugin for not using everything — which is how a check gets ignored and then deleted. A property the real type **doesn't have** is the drift that matters: a rename nobody propagated, where the plugin keeps compiling and reads a field the host never sends.

## Its first interface run found a live one

```
dashboard-interop.d.ts:67  TaskCardProps.workflowStepNameLookup is not a property of the real TaskCardProps
```

Git history says it **was** one when FN-2466 and FN-7039 added this threading. The dashboard removed it later; nothing propagated that to the plugin's hand-written declaration. So the plugin built a lookup map from `context.workflowSteps` on every render, threaded it through two components, and handed it to a `TaskCard` with no such prop.

Deleted rather than exempted — a new gate shouldn't ship with a waiver for its own first finding. Behaviour-preserving: the value never reached anything.

## Measured on `main`

| check | result |
|---|---|
| population | **7 functions + 10 interfaces across 6 plugins**, all matching after the deletion |
| control probe | phantom property **caught**; clean tree exits 0 |
| anti-vacuity | now also requires a non-zero *interface* comparison |
| gate's own suite | **5 → 8** |
| dependency-graph suite | 179 green; `tsc` clean |
| other five gates · census | green |

## Running total for this check

Three real drifts, none of which any other instrument reported:

1. `isTaskStuck` stuck at three parameters through the whole lane conversion (#3003)
2. `taskStuckTimeoutMs?: number` vs the required `number | undefined` — in **two independent authors'** declarations
3. `workflowStepNameLookup` outliving its removal from `TaskCard`

### DIFF
--- a/plugins/fusion-plugin-dependency-graph/src/DependencyGraph.tsx
+++ b/plugins/fusion-plugin-dependency-graph/src/DependencyGraph.tsx
@@ -35,7 +35,6 @@ export interface DependencyGraphProps {
   onOpenMission?: (missionId: string) => void;
   onMoveTask?: (id: string, column: Task["column"], optionsOrPosition?: { preserveProgress?: boolean } | number) => Promise<Task>;
   lastFetchTimeMs?: number;
-  workflowStepNameLookup?: ReadonlyMap<string, string>;
 }
 
 const POINTER_MOVE_THRESHOLD = 4;
@@ -58,7 +57,6 @@ export function DependencyGraph({
   onOpenMission,
   onMoveTask,
   lastFetchTimeMs,
-  workflowStepNameLookup,
 }: DependencyGraphProps) {
   const viewportRef = useRef<HTMLDivElement | null>(null);
   const pointerDownRef = useRef<{ x: number; y: number } | null>(null);
@@ -506,7 +504,6 @@ export function DependencyGraph({
                     onOpenMission={onOpenMission}
                     onMoveTask={onMoveTask}
                     lastFetchTimeMs={lastFetchTimeMs}
-                    workflowStepNameLookup={workflowStepNameLookup}
                     onMouseEnter={() => setHoveredTaskId(node.task.id)}
                     onMouseLeave={() => setHoveredTaskId(null)}
                     onClick={(event) => {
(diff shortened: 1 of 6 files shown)


gsxdsm:
Verified — and this closes the exact caveat I recorded on #3029, where I noted the arity rule leaves a mirror that keeps the right *count* while drifting on types invisible.

| check | result |
|---|---|
| tests | 8 pass / 0 fail |
| gate on this branch | exit 0, no self-write — *"7 mirrored function(s) and 10 interface(s) across 6 plugin(s)"* |
| **add a property the real interface lacks** | **exit 1** — `PluginDashboardViewContext.kbPhantomProp is not a property of the real PluginDashboardViewContext` |
| a mirror declaring **fewer** properties (all six do) | still passes |

The one-direction rule is the right call and the reason is worth keeping:

> A mirror may declare *fewer* properties, and all six do (6, 8, 7, 7, 3, 6 against the real nine) because a plugin mirrors the fields it uses. Demanding equality would fail every plugin.

Subset-is-fine, superset-is-drift is exactly the asymmetry that makes this checkable without argument — the same instinct as restricting the function rule to arity. A rule that fired on all six mirrors on day one would have been turned off by the end of the week.

**The dead prop it deletes is genuinely dead**, which I checked rather than assumed: `workflowStepNameLookup` has no consumer left in the plugin outside the comment recording its removal. That is the interface rule paying for itself immediately — it found a phantom prop in the mirror, and following it up found the real one was unused too.

### One note on the re-land

> That PR merged into #3029's branch and only its first commit reached `main` — the arity rule shipped, the interface rule and its finding did not.

Confirmed independently: `main`'s gate reports *"7 mirrored function(s)"* with no interface count, and the dead prop is still there. Worth flagging as its own hazard — a PR targeting another PR's branch loses whatever the parent does not carry, and it fails **silently**: `main` was green, the gate existed, and half its coverage was simply absent. The gate's own summary line is what makes it detectable, which is a good argument for checks that print what they covered rather than only what failed — the same property #3005's `scannedColumns` fix was about.

PR #3036, opened 2026-07-31
test(docs): the PLUGIN_AUTHORING TOC guard rejected legal nested entries, and has been red on main

gsxdsm:
## A legal Markdown sub-entry turned this guard red

```
AssertionError: Invalid TOC line: - [Theming & Overlay Layering for Dashboard Views](#theming--overlay-layering-for-dashboard-views)
```

That line is an ordinary nested TOC entry, indented under item 8 of `docs/PLUGIN_AUTHORING.md`. The parser did `.map(line => line.trim())` **first** and then required every line to match the top-level `N. [title](#anchor)` shape — so indentation, the one thing distinguishing a sub-entry from a malformed top-level one, was destroyed before it could be used.

**The doc was never wrong.** Only the parser was, and it has been red on `main` since the entry was added.

Indentation is now read before trimming. Sub-entries are still required to be well-formed links; they just do not participate in the numbering or the count.

## Both guard directions verified by breaking them

A looser parser that skipped anything unrecognised would have made the failure go away while quietly ending the guard's usefulness — so I checked it still fails in both directions:

| mutation | result |
| --- | --- |
| top-level `9.` rewritten as a bullet | still fails (`Invalid TOC line`) |
| nested entry replaced with un-linked prose | still fails (`Invalid nested TOC line`) |

## The wider finding, which matters more than this fix

I found it by sweeping `scripts/__tests__` against clean `main`: **688 passing, 7 failing test files.**

| suite | failing assertion |
| --- | --- |
| `ci-test-shard-timings` | committed timing snapshot references live test files |
| `dependency-security-floor` | pnpm overrides pin transitive protobufjs to a safe floor |
| `engine-vitest-gate-policy` | pg gate canaries remain a subset of the enabled suite |
| `plugin-authoring-docs` | **this PR** |
| `release-prompt-gate` | release dry-run exits before proceed confirmation |
| `verify-fast` | defaults to every canonical pretest validator |
| `workflow-reliability-release-check` | manifest references existing seam files |

All sit **outside the merge gate**. That is now the third instance of this pattern I have hit — #2969's 15 red agent-action tests and #3033's stale ratchet list were the others — and it is clearly systemic rather than incidental.

I fixed only the one in plugin territory. The rest span CI sharding, **dependency security** (that protobufjs floor is a security assertion currently not holding), release gating and workflow manifests. Each needs its owner's judgement about whether the assertion or the world is wrong, and a drive-by "make it green" is exactly how a real signal gets erased — `dependency-security-floor` especially.

## Verification (measured)

- this suite — **4 passed / 0 failed** (was 1 failed)
- `eslint` — clean

Test-only; the doc is untouched. No changeset.

### DIFF
--- a/scripts/__tests__/plugin-authoring-docs.test.mjs
+++ b/scripts/__tests__/plugin-authoring-docs.test.mjs
@@ -54,10 +54,27 @@ test("PLUGIN_AUTHORING TOC includes top-level dashboard views and anchors align 
   const tocMatch = doc.match(/## Table of Contents\n\n([\s\S]*?)\n---/);
   assert.ok(tocMatch, "Table of Contents block should exist");
 
-  const tocLines = tocMatch[1]
-    .split("\n")
-    .map((line) => line.trim())
-    .filter(Boolean);
+  /*
+  FNXC:PluginAuthoringDocs 2026-07-31-18:20:
+  NESTED TOC entries are legal Markdown, and this parser rejected them by flattening indentation away.
+
+  `- [Theming & Overlay Layering for Dashboard Views](...)` sits indented under item 8 — an ordinary
+  sub-entry. The old code trimmed every line first and then required ALL of them to match the
+  top-level `N. [title](#anchor)` shape, so adding a perfectly valid sub-entry turned this assertion
+  red on `main`, and it has been red since.
+
+  Indentation is the discriminator, so it is read BEFORE trimming. Sub-entries are still required to
+  be well-formed links — they are simply not top-level sections and do not participate in the
+  numbering or the count. A malformed TOP-LEVEL line still fails exactly as before, which is the guard
+  this test exists to be.
+  */
+  const rawTocLines = tocMatch[1].split("\n").filter((line) => line.trim());
+  const subEntries = rawTocLines.filter((line) => /^\s+/.test(line));
+  for (const line of subEntries) {
+    assert.ok(/^\s+-\s+\[.+\]\(#.+\)$/.test(line.replace(/\s+$/, "")), `Invalid nested TOC line: ${line}`);
+  }
+
+  const tocLines = rawTocLines.filter((line) => !/^\s/.test(line)).map((line) => line.trim());
 
   const tocEntries = tocLines.map((line) => {
     const m = line.match(/^(\d+)\.\s+\[(.+)\]\(#(.+)\)$/);


gsxdsm:
## Diagnosis of one of the other six: `workflow-reliability-release-check` is red because an acceptance map claims coverage that no longer exists

I said the remaining six needed their owners' judgement. That is the kind of stated blocker that decays, so I tested it on the most diagnosable one — and the answer is worse than a dangling path.

The failing assertion is *"manifest references existing seam files"*. The manifest is `docs/custom-workflow-reliability-acceptance-map.md`. Measured against clean `main`:

```
cited test files: 13   missing: 8
rows citing tests:  5   rows with ZERO surviving evidence: 2
```

The two rows with **no surviving evidence at all**:

- *"A custom workflow can be authored/imported, rejected on invalid IR, saved, discovered, selected, and reloaded."* — all four cited files gone (`workflow-definition-store`, `workflow-import-export`, `workflow-design-route`, `workflow-selection-store`).
- *"Engine/scheduler restart preserves workflow selection and progress, and any recovery emits…"*

`workflow-definition-store.test.ts` was deleted by the PostgreSQL migration (#1793, `c15c78feeb`). An acceptance map is a claim that a behaviour is verified; with its evidence deleted, the claim is simply false, and the guard that would have said so is red and outside the merge gate.

## Why I am not "fixing" it

The obvious repair is to repoint the row at `packages/core/src/__tests__/postgres/workflow-definitions.pg.test.ts`. **That would launder a real gap.** Its entire coverage is:

```
listWorkflowDefinitions resolves (builtins) without throwing in backend mode
custom workflow rows are read from project.workflows via the async helpers
```

against a claim spanning authoring, import, invalid-IR rejection, save, discovery, selection and reload. Repointing turns a red guard green while the behaviour goes unverified — the precise failure mode I flagged for `dependency-security-floor` in the PR body, met in the first suite I dug into.

Someone who owns custom-workflow reliability has to decide, per row, whether the coverage **moved** (repoint) or was **deleted** (restore it, or strike the claim). That is a judgement about what is actually verified, and neither answer is derivable from the manifest.

So the blocker stands for this one — but now as a measurement rather than an assumption, and the finding is sharper than "a test is red": **the custom-workflow reliability acceptance map is overstating what is verified, and has been since the PG migration.**

gsxdsm:
Confirmed the red on `main` and verified the fix — with one gap worth closing while you are in here.

| check | result |
|---|---|
| `main` | **1 failed / 3 passed** — `Invalid TOC line: - [Theming & Overlay Layering…]`, exactly as described |
| this branch | 4/4 |
| inject a genuinely malformed TOC line | **still rejected** (1 failed) |
| break a **top-level** entry's anchor | **still rejected** |
| break a **nested** entry's anchor | **passes** ← |

That last row is the gap. Pointing the nested entry at `#kb-nonexistent-anchor` leaves the suite green, while the same corruption in a top-level entry fails. So nested entries are now *accepted* but their anchors are never checked, and a broken sub-link is exactly the kind of rot a TOC guard exists to catch — silent, and only discovered by a reader who clicks it.

The diagnosis in the body is right and the parser fix is right:

> The parser did `.map(line => line.trim())` **first** and then required every line to match the top-level `N. [title](#anchor)` shape — so indentation, the one signal distinguishing a sub-entry, was destroyed before the check.

Destroying the distinguishing signal before testing for it is a nice specimen. It is the same shape as the aria matcher I shipped two days ago, whose character class was terminated by the very quotes every real call site contains — in both cases the normalisation ran first and removed the thing the rule needed.

Since the parser now *recognises* nested entries, extending anchor validation to them is a small addition to the branch you have already opened, and it would make the third row above fail as it should.

Not blocking — the PR strictly improves on a red `main`, and a guard that accepts legal syntax without validating one class of link is still better than one that rejects the document outright.

PR #3038, opened 2026-07-31
test(verify-fast): the pretest-validator mirror is stale by two, and has been red on main

gsxdsm:
## The pretest-validator mirror is stale by two

```
assert.deepEqual(PRETEST_STATIC_CHECK_SCRIPTS, PRETEST_CHECKS);
+   'scripts/check-no-cwd-relative-dashboard-test-reads.mjs',
+   'scripts/check-capacity-pool-id.mjs',
```

Production found ten validators; the test's mirror listed eight. **The world was right and the assertion was stale** — the good direction.

`PRETEST_STATIC_CHECK_SCRIPTS` is *derived* from `package.json`'s pretest chain, so `verify:fast` picked both new validators up automatically when they were added. They are real scripts, they run in pretest, and verify:fast was already running them. Only the mirror needed telling.

Added in production order, since the assertion is a `deepEqual` and order is part of it. Plain strings for these two — the surrounding entries are split and re-joined to keep banned phrases (the port-kill and nohup literals) off a single source line for the policy scanner, and neither new name contains one.

## The guard is load-bearing — verified against the source of truth, not the mirror

Syncing a mirror is worthless if the assertion can no longer fail, so I mutated `package.json`'s pretest chain rather than the test:

```
dropped check-capacity-pool-id from pretest  →  ℹ pass 16   ℹ fail 1
restored                                     →  ℹ pass 17   ℹ fail 0
```

So it still catches a validator silently leaving `verify:fast`, which is the regression that actually matters. `package.json` is restored; the diff here is the test file only.

## Second of seven, and the contrast is the point

This is the second of the seven red `scripts/__tests__` suites I found on clean `main` (after #3036). It was genuinely mechanical.

`workflow-reliability-release-check` was not, and I did **not** fix it — diagnosed on #3036 instead: `docs/custom-workflow-reliability-acceptance-map.md` cites 13 test files of which **8 are missing**, and **2 of 5 rows have zero surviving evidence**, including *"a custom workflow can be authored/imported, rejected on invalid IR, saved, discovered, selected, and reloaded"*. The obvious repoint would have turned it green while the behaviour stayed unverified.

That contrast is why I am taking these one at a time rather than sweeping them green: two of seven look identical from the failure line, and only one of them is safe to fix without owning the subject.

## Verification (measured)

- this suite — **17 passed / 0 failed** (was 1 failed)
- `eslint` — clean

Test-only. No changeset.

### DIFF
--- a/scripts/__tests__/verify-fast.test.mjs
+++ b/scripts/__tests__/verify-fast.test.mjs
@@ -46,12 +46,18 @@ const PRETEST_CHECKS = [
     "no",
     "hup.mjs",
   ].join(""),
+  "scripts/check-no-cwd-relative-dashboard-test-reads.mjs",
   [
     "scripts/check-no-",
     "kill-",
     "4040.mjs",
   ].join(""),
   "scripts/check-no-getdatabase.mjs",
+  /* FNXC:TestInfrastructure 2026-07-31-19:15: added when the assertion below went red on `main`.
+     Both validators are real scripts, both run in `package.json`'s pretest chain, and
+     `PRETEST_STATIC_CHECK_SCRIPTS` already picked them up — verify:fast was RIGHT and this mirror was
+     stale. Kept in production order, since the assertion is a deepEqual and order is part of it. */
+  "scripts/check-capacity-pool-id.mjs",
   "scripts/check-no-node-only-core-imports-in-dashboard.mjs",
   "scripts/check-pi-versions-pinned.mjs",
   "scripts/check-no-test-timeout-appeasement.mjs",


gsxdsm:
## Re-measured: the count is 5, not 7 — and the one I escalated loudest is not failing

I have been quoting "7 red suites in `scripts/__tests__`" across several PRs. Re-run against current `main` (`f411d55591`):

```
ℹ pass 694
ℹ fail 5
```

Still failing:

| suite | status |
| --- | --- |
| `ci-test-shard-timings` | unowned, undiagnosed |
| `engine-vitest-gate-policy` | unowned, undiagnosed |
| `release-prompt-gate` | unowned, undiagnosed |
| `verify-fast` | **this PR** |
| `workflow-reliability-release-check` | diagnosed on #3036 — needs its owner, not a repoint |

Gone since: `plugin-authoring-docs` (my #3036, merged) and **`dependency-security-floor`**.

## The correction that matters

I flagged `dependency-security-floor` as the one a human should look at urgently — a protobufjs floor "that currently isn't holding". **It passes now, 4/4, and consistently across three consecutive runs.** So that escalation should not be acted on as written.

I do not know with confidence why it differed earlier. Either something landed that restored the floor, or my earlier sweep was affected by worktree state (that assertion reads `pnpm-lock.yaml`, which is exactly the kind of input that differs between a freshly-fetched tree and one carrying stale install state). I am not going to guess between those and present it as fact.

What I can say precisely: **on current `main` the protobufjs floor assertion holds**, and my earlier statement is stale.

This is the same decay I keep documenting in other people's deferrals, applied to my own reporting: a measurement quoted three PRs after it was taken is a claim, not a measurement. The useful discipline is re-running it before repeating it, which is what turned 7 into 5.

gsxdsm:
Confirmed the red on `main` before it merged, and verified post-merge that the mirror still bites in both directions.

| state | result |
|---|---|
| `main` before | **16 pass / 1 fail** — mirror listed 8, production found 10 |
| `main` now | **17 pass / 0 fail** |
| remove a validator from the pretest chain | **fails** |
| add one to the chain | **fails** |

So the mirror is a real two-way assertion, not a list that only grows.

### Third instance today of the same failure

> The world was right and the assertion was stale — the good direction.

That is now three in one day, all identical in shape:

- **#3033** — a lane parameter got wired; the guard's allow-list was not shortened → red.
- **my #3032** — I wired a call site; `check-lane-wiring`'s baseline was not re-recorded → red, and I only found it by running the gate rather than from CI.
- **this** — two validators were added to the chain; the test's mirror was not extended → red.

Every one is a **ledger that lags an improvement**. Nobody broke anything; in each case someone made the codebase better and the bookkeeping stayed behind. A regression is loud and obviously the author's; a ledger update is quiet and reads as somebody else's chore.

Worth stating as a rule since it has now cost four separate reds: **if your change moves a number a gate remembers, the same commit owns the number.** The gates that print the delta and name the file — this one, `check-lane-wiring` — are the ones where that is cheap; the ones that just fail an equality are where it stings.

### Probe note

My first attempt at the mutations above ran against `main` while I believed I was on the PR branch: the `git fetch` printed `couldn't find remote ref` because the branch had already merged and been deleted. Both readings were main's pre-existing red, so the "mutation" showed no change and meant nothing. Caught it by re-checking `git branch --show-current` — the twelfth time today a harness rather than an instrument gave me a confident wrong answer, and the second time from this exact cause.

PR #3041, opened 2026-07-31
fix(core): archived tasks leaked into the live feed on a renamed board

gsxdsm:
From my own #2839, re-measured today. Of that issue's SQL-literal sites, this is the one that decides what a **live view** shows.

## The defect

`listTasksModifiedSinceImpl` backs the SSE watcher and modified-since polling — the incremental feed the dashboard applies to its task list. Its `includeArchived: false` branch excluded the literal `archived`:

```ts
conditions.push(sql`${schema.project.tasks.column} != 'archived'`);
```

On a board whose archive lane is named anything else, that predicate matches **every** row and excludes nothing. Archived cards arrive in the live feed and reappear on the board.

Nothing errors, and a full refetch filters archived rows by another path — so the symptom is archived work that comes back until the next reload. That gets reported as *"the board is flaky"*, not as a bug.

## The fix

`resolveProjectColumnsForRoles` seeds the legacy ids before adding resolved ones, so the set is never empty and an **unconverted board excludes exactly `archived` as before**. The literal stays as the resolution-failure fallback, where excluding nothing would be worse than excluding the legacy id.

## Surface enumeration — three of four sites are dead

Four sites share this invariant. Verified rather than assumed:

| site | status |
|---|---|
| `reads.ts:558` (SSE / modified-since) | **live** — converted here |
| `liveParentFilter` | **no references anywhere** in `packages/` or `plugins/` |
| `listLiveTaskDocuments`, `listLiveArtifacts` | referenced **only** by `taskstore-remaining.test.ts` |

That's why this PR converts one site rather than four — the other three are production-dead, and converting dead code would add risk for no behaviour.

## Measured

| check | result |
|---|---|
| new PG suite | **4 cases** — legacy control, the renamed defect, a live-lane negative, and the forensic `includeArchived: true` read |
| mutation (force the legacy fallback) | fails **exactly** the renamed case; the other three hold |
| six gates + `tsc` | green |

The negative case is the one that matters most: resolving the archive role must not start excluding **live** work, or the board silently stops updating for real tasks — a worse failure than the leak this fixes.

## One process note

I corrupted this file mid-session by mutation-testing it while uncommitted: a failed restore left a half-applied block, and a later `git checkout --` discarded the fix entirely. Both were caught by re-grepping for the symbol rather than trusting the restore. The reliable pattern is **commit first, then mutate, then `git checkout` to restore** — which is how the proof above was actually run.

### DIFF
--- a/.changeset/sse-archive-lane-leak.md
+++ b/.changeset/sse-archive-lane-leak.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Archived tasks no longer reappear on boards whose archive lane is renamed.
+category: fix
+dev: `listTasksModifiedSinceImpl` excluded the literal `archived`; it now excludes the project's resolved archive columns, keeping the literal as the no-resolution fallback.
--- a/packages/core/src/__tests__/postgres/modified-since-archived-lane.pg.test.ts
+++ b/packages/core/src/__tests__/postgres/modified-since-archived-lane.pg.test.ts
@@ -1,0 +1,110 @@
+/*
+FNXC:WorkflowResolvedColumns 2026-07-31-09:20 (archived rows leaked into the live stream):
+
+`listTasksModifiedSinceImpl` backs the SSE watcher and modified-since polling — the incremental feed
+the dashboard applies to its live task list. Its `includeArchived: false` branch excluded the LITERAL
+`archived`, so on a board whose archive lane is named anything else the predicate matched every row
+and excluded nothing. Archived cards arrived in the live feed and reappeared on the board.
+
+Nothing errors, and a full refetch filters archived rows by another path, so the symptom is archived
+work that comes back until the next reload — the kind of thing an operator reports as "the board is
+flaky" rather than as a bug.
+
+The cases are DIFFERENTIAL: the same archived task under two vocabularies whose roles are identical
+and only the ids differ. `filed` collides with no legacy id, so a surviving `'archived'` literal
+cannot pass by luck.
+*/
+
+import { it, expect, beforeAll, beforeEach, afterEach, afterAll } from "vitest";
+import { sql } from "drizzle-orm";
+import {
+  pgDescribe,
+  createSharedPgTaskStoreTestHarness,
+  type SharedPgTaskStoreHarness,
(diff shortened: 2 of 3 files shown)


gsxdsm:
Verified on the PR head (asserted branch + SHA before measuring — I have been caught twice today running against `main` while believing otherwise).

| check | result |
|---|---|
| new pg suite | 4/4 |
| revert to the bare `!= 'archived'` literal | **1 failed / 3 passed** |
| force the resolution to fail (exercise the fallback) | **1 failed / 3 passed** |
| `check-sql-column-literals`, `check-lane-wiring` | exit 0, no self-write |

The second mutation is the one I most wanted to see fail. A fallback that is never exercised is where this class of fix usually rots: the resolved path gets all the attention, the `else` branch is written from reasoning and never run, and it silently becomes the only path on some board configuration. Here it is genuinely covered.

## The severity framing is right, and unusually well-drawn

> Nothing errors, and a full refetch filters archived rows by another path, so the symptom is archived work that comes back until the next reload.

That is the honest description, and it explains why this survived: **the bug is self-healing on refresh.** An operator sees archived cards reappear, reloads, they are gone. That reads as a UI glitch, not a data-filter defect, so it never becomes a bug report — and the incremental feed keeps re-adding them until the next full fetch.

Worth contrasting with #3016's wrong *ordering* and #3015's `?? "todo"` badge: three defects this batch whose common property is that **the wrong answer is plausible enough to be mistaken for normal behaviour.** None throws. That is what let the whole class accumulate behind a conversion programme.

## One observation on the seeding choice

> `resolveProjectColumnsForRoles` seeds the legacy ids before adding resolved ones, so the set is never empty and an unconverted board excludes exactly `archived` as before.

Seeding-then-adding is the right call *here* because the operation is an **exclusion** — a superset can only exclude more, and excluding a legacy-named archive lane on a converted board is harmless since nothing lives there.

Worth noting it is the opposite of the call `node-override-guard.ts` documents, where seeding the legacy ids into a WIP set was a bug: there the set drives a **refusal**, so a superset blocks legitimate work. Same helper shape, opposite correct answer, decided entirely by whether the set gates *out* or gates *off*. That distinction is not obvious from the call site and is worth the line it already has.

PR #3042, opened 2026-07-31
chore(core): delete liveParentFilter — no caller, and it carried a legacy lane literal

gsxdsm:
Found while enumerating archive-exclusion sites for #3041.

## Unambiguously dead

`liveParentFilter` has exactly **one** reference in the repo: its own definition.

- not exported from `index.ts` or `index.gate.ts`
- no test imports it
- no production code calls it

It nonetheless contained `column != 'archived'`, so it was one of the 22 sites the SQL column-literal gate tracks.

## Why delete rather than convert

Converting it would mean adding lane resolution to code nothing runs — risk with no behaviour. That's the same argument #3041 makes for *not* converting the other two dead sites; deleting is the version of it that also removes the literal.

## The gate it documents is not being deleted

Its docblock describes the document/artifact visibility gate (VAL-CROSS-015). That gate is real and still enforced — by the inline conditions inside `listLiveTaskDocuments` and `listLiveArtifacts`, which is presumably why this helper was never wired up in the first place. Only the unused composition goes.

## Measured

| check | result |
|---|---|
| SQL literal population | **22 → 21**; the gate ratcheted its own baseline down and asked for the commit, included here |
| `taskstore-remaining.test.ts` (archive-lineage suite) | **27 tests green** |
| six gates + `tsc` | green |

## Not deleted, deliberately

`listLiveTaskDocuments` and `listLiveArtifacts` are referenced **only** by that test file. That's a weaker signal than zero references — someone may have written them ahead of a consumer. Their literals stay counted, which is the honest state for code whose intent I can't read from the repo.

### DIFF
--- a/packages/core/src/task-store/async-archive-lineage.ts
+++ b/packages/core/src/task-store/async-archive-lineage.ts
@@ -40,26 +40,6 @@ import {
 import type { ArchivedTaskEntry } from "../types.js";
 
 /**
- * FNXC:TaskStoreArchiveLineage 2026-06-24-07:05:
- * The "live parent" predicate for the document/artifact visibility gate
- * (VAL-CROSS-015). Documents and artifacts scoped to a task are surfaced in
- * live views only when their parent task is live: `deleted_at IS NULL` (not
- * soft-deleted) AND `column != 'archived'` (not archived). When the parent is
- * archived or soft-deleted, the rows are retained but filtered out of live
- * views — they remain for an unarchive/restore.
- *
- * This predicate is the join condition for `task_documents` / `artifacts` →
- * `tasks`. It is the async equivalent of the sync
- * `taskExists && taskExists.column !== 'archived'` check in
- * `upsertTaskDocument` and the `hasActiveTask` gate in `getTaskDocument`.
- */
-export function liveParentFilter(taskIdColumn: ReturnType<typeof eq>) {
-  // The caller passes an equality fragment like eq(schema.project.tasks.id, taskId).
-  // We compose the live-parent conditions on top.
-  return and(taskIdColumn, ACTIVE_TASK_FILTER, sql`${schema.project.tasks.column} != 'archived'`);
-}
-
-/**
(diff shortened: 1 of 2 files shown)


gsxdsm:
Independently confirmed dead before agreeing to the deletion, since "no caller" is the claim most worth checking twice.

On `main`, `liveParentFilter` has **one** occurrence in the whole tree — its own definition. I also checked the routes a plain name grep misses:

- **no star re-export** of `async-archive-lineage` anywhere;
- the module *is* imported by three files, but only for its other exports (`upsertArchivedTaskEntry`, `findArchivedTaskEntry`, `archiveParentTaskWithLineageGate`, `deleteArchivedTaskEntry`, `restoreTaskFromArchive`) — never this one;
- the only other hits are `packages/core/dist/…`, build output of the same source rather than an independent consumer.

On the branch: gone, `tsc -p tsconfig.json` 0 errors, the three importers unaffected.

**The baseline decrement is exactly one literal and nothing else:**

```
- "packages/core/src/task-store/async-archive-lineage.ts": 3,
+ "packages/core/src/task-store/async-archive-lineage.ts": 2,
```

Gate reads `21 known SQL column literal(s), none added` with **no self-write** — so the re-record rode in this commit rather than being left for the next person. That is worth calling out, because it is the exact bookkeeping that has produced four separate main-reds today (#3033, #3038, my own #3032, and the #3036 chain): a change that improves the code and leaves a gate's ledger behind. This one did not.

**On delete-vs-convert:**

> Converting it would mean adding lane resolution to code nothing runs — risk with no behaviour.

Agreed, and it is the same call I made deleting `fusion-plugin-even-cards` in #2988: converting unreachable code produces debt that is unverifiable by construction, because no test or typecheck exercises the path. The one thing that made that decision safe there was evidence the removal was *intended* rather than merely possible — a changelog entry recording the consolidation. Here the equivalent evidence is simpler and stronger: it has no caller at all, so there is nothing to intend.

PR #3049, opened 2026-07-31
convert(engine): self-healing lane cluster 56 -> 38 guards (repo 126 -> 108)

gsxdsm:
## Census before / after

```
                                    before    after
self-healing.ts column guards          56        38
repo-wide COLUMN guards (backlog)     126       108
```

`self-healing.ts` was the largest single cluster by a wide margin — 56 guards against 12 in the next file. Baseline re-recorded in the same commit; `--strict` green.

## Converted: 15 guards across 11 sweeps

Existing helpers only — `resolveProjectColumnsForRoles` with `TERMINAL_ROLES` / `REVIEW_ROLES` / `countsTowardWip` / `hold` / `archived`, the same shape this file already uses. No new helper, no new resolution pattern.

What each was silently doing on a renamed board:

| sweep | behaviour before |
| --- | --- |
| `archiveStaleDoneTasks` | **both** guards inert, so every card counted as an active dependent and the sweep archived **nothing at all** |
| `reconcileDependencyBlockingLeases` | no holder matched, so a stale file-scope lease blocking an unmet dependency was never cleared |
| `reconcileCompletedBlockedTasks` | work whose blocker had cleared stayed parked instead of advancing |
| `reconcileInReviewUnmetDependencies` | a card sat in review with unmet dependencies and no rebound |
| `reclaimStaleActiveBranches` | archived cards were eligible for branch reclaim |
| `reconcileInReviewBranchRebind` | the rebind list was empty |
| `autoReboundPausedScopeDecayDetailed` | no card was ever seen as executing |
| `detectStalledCards` | finished cards counted as stall candidates |
| `recoverApprovedStrandedAiMergeCommit`, `recoverDriftedAgentTaskLinks`, `cleanupStaleTempMergeWorktrees` | same shape |

**Reused rather than duplicated:** `recoverWedgedActiveMerge` already resolves `wedgedReviewColumns` via `resolveReviewColumnsFor` three lines above the site I was converting, so the site now uses it instead of a second resolution of the same question.

## One site I converted and then reverted

`clearStaleBlockedBy`'s memo closure carries an FNXC note stating the literal is **deliberate**: the closure only decides whether to re-log an already-logged blocker, so a renamed board costs a duplicate log line — not a wrong lifecycle decision — and restructuring a sweep's control flow to convert a logging decision is the wrong trade.

I read that note *after* editing the line. Restored.

Worth flagging separately: **it has the reasoning but no `DELIBERATE-LITERAL` marker**, so the census keeps counting it and it re-appears in the backlog as if unexamined. That is a marker gap, not a conversion gap — the next person will make the same mistake I did.

## Not converted — flagged, not guessed

Ten of the twenty-four remaining sites are in **sync predicates with no resolution seam**:

- `classifyPausedAbortWorkflowRecovery` (3)
- the `start()` task-moved listener (5) — compares event `from`/`to` columns inside a sync callback
- `isWorkspaceOwnerLive` (1)
- `isPhantomExecutorBinding` (1)

Converting these means threading a flags parameter down from every caller — precisely the unwired-optional-parameter shape this program keeps finding inert (five were live on `main` at once per `unwired-lane-parameter-guard`). They need a decision about *where the resolution lives*, not a guess from me.

The other **14** are in async sweeps with a seam available and are ordinary follow-on work in this same file.

## Verification (measured)

- self-healing suites — **816 passed / 41 files**
- `pnpm test:gate` — green
- `lifecycle-column-census --strict`, `check-lane-wiring`, `check-sql-column-literals`, `check-inert-flag-seams`, `check-fnxc-future-dates` — green

No changeset: `@fusion/engine` is private.

## Summary by CodeRabbit

* **Bug Fixes**
  * Self-healing workflows now continue functioning when workflow columns are renamed.
  * Improved recovery for stalled, blocked, paused, or disconnected workflow states while preserving existing filters and actions.
  * Temporary merge worktrees and drifted agent links are cleaned up more reliably.

### DIFF
--- a/packages/engine/src/self-healing.ts
+++ b/packages/engine/src/self-healing.ts
@@ -3446,7 +3446,7 @@ export class SelfHealingManager extends SelfHealingGitEvidence {
         limitMs: timeoutMs,
         status: task?.status ?? null,
       });
-      if (task && allowsAutoMergeProcessing(task, settings) && !task.paused && task.column === "in-review") {
+      if (task && allowsAutoMergeProcessing(task, settings) && !task.paused && wedgedReviewColumns!.has(task.column)) {
         try {
           this.options.enqueueMerge?.(activeId);
         } catch (enqueueErr: unknown) {
@@ -4487,6 +4487,8 @@ export class SelfHealingManager extends SelfHealingGitEvidence {
   }
 
   async reclaimStaleActiveBranches(): Promise<number> {
+    /* FNXC:WorkflowLifecycleColumns 2026-07-31-22:30 (self-healing cluster): an archived card must not have its branch reclaimed, whatever that lane is named. Keyed on the literal this sweep answered "no" for every card on a renamed board. */
+    const reclaimArchivedColumns = await resolveProjectColumnsForRoles(this.store, ["archived"]);
     try {
       const settings = await this.store.getSettings();
       if (settings.globalPause || settings.enginePaused) return 0;
@@ -4531,7 +4533,7 @@ export class SelfHealingManager extends SelfHealingGitEvidence {
         if (!derivedTaskId) continue;
 
         const task = taskById.get(derivedTaskId.toUpperCase());
-        if (!task || task.column === "archived" || task.checkedOutBy || task.userPaused) continue;
+        if (!task || reclaimArchivedColumns.has(task.column) || task.checkedOutBy || task.userPaused) continue;
         if (task.pausedReason === "worktrunk_operation_failed") {
           log.debug(`[self-healing] skipping worktrunk-paused task ${task.id}`);
           continue;
@@ -5059,13 +5061,15 @@ export class SelfHealingManager extends SelfHealingGitEvidence {
   }
 
   async reconcileInReviewBranchRebind(options?: { includeTaskIds?: Set<string> }): Promise<RebindResult> {
+    /* FNXC:WorkflowLifecycleColumns 2026-07-31-22:30 (self-healing cluster): branch rebind targets cards resting in the board's own review lane. Keyed on the literal this sweep answered "no" for every card on a renamed board. */
+    const rebindReviewColumns = await resolveProjectColumnsForRoles(this.store, REVIEW_ROLES);
     const result: RebindResult = { repaired: 0, outcomes: [] };
     try {
       const settings = await this.store.getSettings();
       if (settings.globalPause || settings.enginePaused) return result;
 
       const allTasks = await this.store.listTasks({ slim: true, includeArchived: false });
-      const tasks = allTasks.filter((task) => task.column === "in-review");
+      const tasks = allTasks.filter((task) => rebindReviewColumns.has(task.column));
       const fusionRefOutput = await execAsync("git for-each-ref --format='%(refname:short)' refs/heads/fusion/", {
         cwd: this.options.rootDir,
         timeout: 30_000,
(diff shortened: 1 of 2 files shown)


gsxdsm:
**Collision notice: three open PRs are converting `self-healing.ts` right now.**

- #3047 — `archiveStaleDoneTasks` lane guards (a sweep this PR also converts)
- #3049 — this one, 15 guards across 11 sweeps, 56 → 38
- #3050 — "the self-healing lane cluster, 14 sweeps"

All three touch the same file, and #3050's scope sounds close to identical to this one. That is a merge conflict waiting to happen and duplicated review effort, so it needs one of us picked rather than all three landed.

I have no stake in which. This PR's distinguishing content, if it helps the choice:

- it **reuses** `wedgedReviewColumns` where `recoverWedgedActiveMerge` had already resolved it three lines above, rather than resolving the same question twice;
- it **reverts** one site it had converted, because `clearStaleBlockedBy`'s memo closure carries an FNXC note saying that literal is deliberate — and separately flags that the note has no `DELIBERATE-LITERAL` marker, which is why the census keeps re-surfacing it;
- it **flags rather than converts** the ten sync-predicate sites (`classifyPausedAbortWorkflowRecovery`, the `start()` listener, `isWorkspaceOwnerLive`, `isPhantomExecutorBinding`), since converting them means threading a flags parameter from every caller — the unwired-optional-parameter shape this program keeps finding inert.

If #3050 is the keeper, close this and I will re-apply those three specifics on top of it. I am moving to an unclaimed cluster rather than adding a fourth PR here.

gsxdsm:
## Fleet-phase measurement: the cluster strategy has consumed its clusters

Ran the census against `main` to claim the next-largest unclaimed file, and the number does not mean what the phase assumes:

```
backlog total          : 126
under open PRs         :  75   self-healing (56, THREE PRs), scheduler (12), executor (7)
triaged / already done :   9
everything else        :  42   spread across 29 files (~1.4 guards each)
```

**There is no unclaimed cluster left.** The largest is 4 guards; after that it is a long tail of one- and two-guard files. Continuing "claim the largest cluster" is now producing collisions rather than throughput — three PRs (#3047, #3049, #3050) are converting `self-healing.ts` simultaneously, and #3047's whole scope is a sweep #3049 also converts.

## The 9 "triaged" are not available work either

- `notification-service.ts` (5) — both sites carry detailed deferral notes with a **named acceptance test** (`task-wedge-notification.test.ts`) that fails if converted: any `await` before the resolve drops an operator notification when a re-wedge arrives close behind a recovery. Converting is blocked on serialising wedge handling per task, not on the conversion being hard.
- `restart-recovery-coordinator.ts` (4) — already converted; these are the trait-fallback branches the census's own summary counts separately ("of the column guards, 4 are trait-fallback branches").

## The next target after those cannot be converted with the obvious helper

`replan-target.ts` (4) is the largest genuinely unclaimed file. Its in-file lane resolver is:

```ts
export function resolvePlannerLanes(store: TaskStore, taskId: string): PlannerLanes {
  const ir = store.resolveTaskWorkflowIrSync?.(taskId);
```

`resolveTaskWorkflowIrSync` is the resolver #2759's PG gate canary exists to pin: it returns the **default** IR for every task in production, because `getTaskWorkflowSelectionImpl` is a PostgreSQL-cutover stub returning `undefined` unconditionally. So wiring `hasAdvancedPastPlanning` to `resolvePlannerLanes` would produce a guard that reads as converted, counts as census progress, and is wrong for every custom workflow — the inert-conversion class, with the census number moving in the right direction while nothing improves.

The honest path is `resolvePlannerLanesForTaskAsync` (same file, line 135), which means `triage.ts:1452` becomes an await — a control-flow change, not a mechanical conversion. And three of the four call sites live in `self-healing.ts`, which currently has three open PRs.

So I am not claiming it, and I am not adding a fourth PR to `self-healing.ts`.

## What would actually raise throughput now

Pick one of #3047/#3049/#3050 and close the other two — three reviewers on one file is the current bottleneck. After that, the remaining work is 29 small files, which suits "one worker sweeps the tail" far better than "claim the largest cluster".

PR #3050, opened 2026-07-31
fix(engine): convert the self-healing lane cluster — 14 sweeps, census 56 → 26 (126 → 96 repo-wide)

gsxdsm:
Fleet phase — the largest census cluster, and the file I own. **Existing role helpers only**: no new helper, no signature change, and no behaviour change on a built-in board, because every resolver returns the legacy ids plus whatever the workflow declares.

## Census

| | before | after |
|---|---|---|
| `packages/engine/src/self-healing.ts` | **56** | **26** |
| repo-wide column guards | 126 | **96** |

`--strict` green; lane-wiring green; no site newly excused.

## The fourteen sweeps

Each answered a lifecycle question with a built-in column name, so on a renamed board:

| sweep | role | what actually happened |
|---|---|---|
| archive staleness | terminal / complete | archived nothing — every dependent read as active |
| branch reclaim | archived | reclaimed branches belonging to live cards |
| rebind sweep | review | filtered to an empty list and reported success every cycle |
| stall detection | terminal | scanned finished cards as live |
| temp-merge worktrees | terminal | the shorter done-task grace never applied |
| unmet-dependency rebound | review | skipped every card |
| wedged active merge | review | never re-enqueued; the wedge stayed wedged |
| dependency deadlock | wip + hold | never found a candidate |
| completed-blocked park | hold | never released |
| paused scope decay | wip | file scope decayed forever |
| orphaned step results | wip | **could rewrite a live executor's `pending` results** |
| worktree metadata | terminal / wip / review | rebound finished cards; the FN-5256 liveness guard went silent |
| scope-override safety | wip + review | now shares the sweep's own resolved sets, so the two cannot disagree |
| agent-link drift | wip + review + terminal | evaluated agents whose task was plainly still executing |

The orphaned-step-results one is the sharpest: its own header says it must never touch an executor-owned row, and the id-keyed guard made it do exactly that. It resolves once per pass, outside the paging loop.

## One marker, not a conversion

One site's rationale was already written out in full — a logging-only guard where the degraded answer costs a duplicate log line, not a wrong lifecycle decision — but it predated the `DELIBERATE-LITERAL` convention, so the census kept scoring a decided site as backlog. Recorded where the census reads it; I did not re-decide it.

## Flagged, not guessed — the 26 that remain

- **Sync call sites** (`isWorkspaceOwnerLive`, the pause-abort classifier, the phantom-binding check). These are synchronous predicates with no store handle. Converting means threading a resolved set through their callers — a signature change, which this PR's brief excludes. They are the honest remainder, not oversights.
- **Already-converted fallbacks** (`own.length > 0 ? own.includes(...) : task.column === "in-review"`). The resolved answer wins; the literal is the documented no-metadata path. Counted as guards because they are unmarked, and I would rather leave them visible than mark 6 sites in a PR that is also converting 30.
- **`fresh.column === "todo"` in the notification route** — I measured this one earlier: any `await` before the wedge resolve drops an operator notification, so it needs its own change with its own test.

## Verification

self-healing suites **204 passed** (37 files) · targeted: query-filter-blindness + overlap-seam + agent-link-drift + surfacing-family **145 passed** · `pnpm test:gate` 161 + 13 + 487 + 71 · lint · census `--strict` · lane-wiring — all green.

### DIFF
--- a/.changeset/self-healing-lane-cluster.md
+++ b/.changeset/self-healing-lane-cluster.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Fourteen more self-healing repairs now use your board's own column names instead of the built-in ones.
+category: fix
+dev: Converts 30 column guards in `self-healing.ts` to `resolveProjectColumnsForRoles` with existing role helpers; no new helpers, no signature changes.
--- a/packages/engine/src/self-healing.ts
+++ b/packages/engine/src/self-healing.ts
@@ -3446,7 +3446,11 @@ export class SelfHealingManager extends SelfHealingGitEvidence {
         limitMs: timeoutMs,
         status: task?.status ?? null,
       });
-      if (task && allowsAutoMergeProcessing(task, settings) && !task.paused && task.column === "in-review") {
+      /* FNXC:WorkflowResolvedColumns 2026-07-31-12:36 (fleet phase): the wedge recovery re-enqueues a
+         merge only for a card still in review; keyed on the id it never re-enqueued on a renamed board
+         and the wedged merge stayed wedged. */
+      const wedgeReviewColumns = await resolveProjectColumnsForRoles(this.store, REVIEW_ROLES);
+      if (task && allowsAutoMergeProcessing(task, settings) && !task.paused && wedgeReviewColumns.has(task.column)) {
         try {
           this.options.enqueueMerge?.(activeId);
         } catch (enqueueErr: unknown) {
@@ -4524,6 +4528,9 @@ export class SelfHealingManager extends SelfHealingGitEvidence {
 
       const tasks = await this.store.listTasks({ slim: true, includeArchived: true });
       const taskById = new Map(tasks.map((task) => [task.id.toUpperCase(), task]));
+      /* FNXC:WorkflowResolvedColumns 2026-07-31-12:20 (fleet phase): the ARCHIVED role, not the id —
+         a renamed archive lane left its branches eligible for reclaim while the card still existed. */
+      const reclaimArchivedColumns = await resolveProjectColumnsForRoles(this.store, ["archived"]);
 
       let reclaimed = 0;
       for (const branch of branches) {
@@ -4531,7 +4538,7 @@ export class SelfHealingManager extends SelfHealingGitEvidence {
         if (!derivedTaskId) continue;
 
         const task = taskById.get(derivedTaskId.toUpperCase());
-        if (!task || task.column === "archived" || task.checkedOutBy || task.userPaused) continue;
+        if (!task || reclaimArchivedColumns.has(task.column) || task.checkedOutBy || task.userPaused) continue;
         if (task.pausedReason === "worktrunk_operation_failed") {
           log.debug(`[self-healing] skipping worktrunk-paused task ${task.id}`);
           continue;
@@ -5065,7 +5072,13 @@ export class SelfHealingManager extends SelfHealingGitEvidence {
       if (settings.globalPause || settings.enginePaused) return result;
 
       const allTasks = await this.store.listTasks({ slim: true, includeArchived: false });
-      const tasks = allTasks.filter((task) => task.column === "in-review");
+      /*
+      FNXC:WorkflowResolvedColumns 2026-07-31-12:15 (fleet phase):
+      REVIEW is the union of the three review roles. Keyed on the id this filter selected NOTHING on a
+      renamed board, so the rebind sweep ran over an empty list and reported success every cycle.
+      */
+      const rebindReviewColumns = await resolveProjectColumnsForRoles(this.store, REVIEW_ROLES);
+      const tasks = allTasks.filter((task) => rebindReviewColumns.has(task.column));
       const fusionRefOutput = await execAsync("git for-each-ref --format='%(refname:short)' refs/heads/fusion/", {
         cwd: this.options.rootDir,
         timeout: 30_000,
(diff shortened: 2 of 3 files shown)


gsxdsm:
**Not rebasing this — you are mid-resolution, and this conflict is my fault.** Your worktree (`kb/capacity-global-cap`) has uncommitted changes to `scripts/lib/lifecycle-column-census-baseline.json`, the exact file the conflict is in, with activity 7 minutes ago. A force-push from my side would diverge your branch under you while you are resolving it.

## What happened

We converted the same file in parallel. **#3047 merged** (`archiveStaleDoneTasks`, self-healing 56 → 51, repo 126 → 121), which is what put this into conflict. Your PR was cut when the baseline still read 56.

I flagged the duplicate before mine merged — `convert/self-healing-lane-cluster-u7` and my branch were both live on `self-healing.ts` — but flagging it did not stop it, so this is a real cost of the phase, not a near-miss.

## The resolution, concretely

Two conflict classes, and only one is a real merge:

1. **`scripts/lib/lifecycle-column-census-baseline.json` — generated. Do not hand-merge.** Take main's version, then re-run:
   ```
   ```
   Hand-merging that JSON yields a number matching neither branch while looking entirely plausible.

2. **`archiveStaleDoneTasks` — genuine overlap.** #3047 converted three guards there: the active-dependents scan and temp-worktree age gate to `TERMINAL_ROLES`, and the staleness filter to `complete` **alone** (an already-archived card is not a candidate for archiving — using the terminal pair makes the sweep consider its own output). If your version converts the same guards, take main's; if it goes further, keep yours.

Your remaining 11 sweeps should apply cleanly.

## One thing worth checking after you rebase

Your `56 → 26` was measured against the old baseline. Re-measure after rebasing — the real delta is now against 51, and the PR body number should match what `--strict` sees, or review has no way to check the claim.

Also: converting *some* guards in a sweep makes it register as converted with the `self-healing-converted-sweeps-have-no-literal-lane-guards` ratchet, which then fails it for any literal left behind. That suite grew 36 → 38 when #3047 landed. If a sweep you touched still has one guard on a legacy id, expect that test — it is correct, and it is the reason to finish a function rather than convert most of it.

Ping me if you would rather I take the rebase once your tree is clean.

gsxdsm:
Rebased onto main after **#3047 landed on the same file and converted two of the same sites** (`archiveStaleDoneTasks`). Three textual conflicts, all identical in semantics and differing only in variable name.

Resolved by taking **main's** version at every overlap and deleting my now-redundant declarations — theirs is merged and reviewed, and a second resolver answering the same question in the same sweep is the drift this program exists to remove. My conversions of the *other* twelve sweeps are unaffected.

Updated numbers after the rebase, since #3047 had already taken 56 → 51:

| | |
|---|---|
| `self-healing.ts` | 51 (post-#3047) → **26** |
| repo-wide column guards | 96 → **93** |

Re-verified on the new base: self-healing suites **204 passed**, gate 161 + 487 + 13 + 71, lint, census `--strict`, lane-wiring — green.

Worth flagging for the fleet: two of us picked the same file's largest cluster within the same hour. The census reports counts per file, so "largest unclaimed cluster" reads as unclaimed right up until someone else's PR merges. A one-line claim comment on the file before starting would have saved both of us a conflict resolution.

gsxdsm:
**Three open PRs are converting this same file**: #3049, #3050 (this one), #3055 — plus #3047 which already merged and forced a rebase here. Four workers, one 51-guard file, same hour.

That is not a coincidence, it is the brief: "claim the largest unclaimed census file cluster" with **no claim mechanism**. The census reports counts per file, so the largest cluster reads as unclaimed to everyone simultaneously, and stays that way until someone's PR merges. Every one of us then pays a rebase against the others' overlapping conversions of the same sweeps.

Suggested, and cheap: whoever starts a file posts a one-line claim comment naming it before converting, and the rest pick the next-largest. A `git log --oneline origin/main -5` plus an open-PR title scan before starting would also have caught it — I only looked after my second conflict.

I am not opening a fourth conversion of this file. Pick whichever of #3049/#3050/#3055 is furthest along and close the others as superseded; the conversions are semantically identical and differ mainly in variable naming. Mine covers fourteen sweeps and is rebased onto #3047 with the overlaps resolved in main's favour.

Moving to an unclaimed file rather than adding to the pile.

gsxdsm:
CLAIM (taking my own advice from the comment above): starting on the unclaimed long-tail batch — `packages/engine/src/executor.ts` (4), `packages/dashboard/app/utils/taskRevert.ts` (2), `packages/engine/src/auto-merge-finalization.ts` (2). Everything else in the census top-list is already spoken for by #3054, #3056, #3057, #3060, #3061, #3063, #3064, or by the two open self-healing PRs.

gsxdsm:
Closing as **superseded by #3055**, and taking my own advice from the comment above rather than restating it.

Where this ended up: four PRs converted this one file. #3047 merged first and forced a rebase here; #3049 then merged 56 → 38, which put **9 conflicts** on this branch — several against variable names identical to mine, because we independently picked the same ones for the same sweeps. #3055 is still open at 56 → **19**, a larger conversion than this PR's 56 → 26.

So the remaining value here is a subset of #3055's, and two of us resolving the same nine conflicts against #3049 is exactly the duplicated work I flagged earlier in this thread. The smaller one should yield; that is this one.

**Nothing is lost.** Every sweep converted here is either already on main via #3047/#3049 or inside #3055's larger set. If #3055 stalls, this branch is still on `origin/self-healing-lane-cluster` and the conversions are recoverable — but the resolution should be done once, by whoever lands.

Worth recording, because it is the second-order cost nobody priced: I spent this turn's budget on **three** rebases of the same file rather than on new conversions, and the mechanical fixup of the last one corrupted the file badly enough that I aborted rather than push a mangled tree. The fleet brief's "largest unclaimed cluster" with no claim mechanism does not merely duplicate effort — it converts the duplicated effort into conflict resolution that grows with each merge.

The unclaimed long tail I audited separately (#3067) is a cleaner unit of work for anyone free: 16 files, one real defect, and 15 sites that should NOT be converted, each with the reason.

PR #3055, opened 2026-07-31
fleet(engine): self-healing lane guards 56 → 19 (largest census cluster)

gsxdsm:
Claimed the largest census cluster: `packages/engine/src/self-healing.ts`, the top file at **56 guards — 4× the next** (`scheduler.ts`, 12).

## Census: before → after

```
$ node scripts/lifecycle-column-census.mjs        # before
  COLUMN guards (the backlog):   126
      56  packages/engine/src/self-healing.ts     ← claimed

$ node scripts/lifecycle-column-census.mjs        # after
  COLUMN guards (the backlog):    89
      19  packages/engine/src/self-healing.ts
```

**37 of 56 converted (66%); the file drops from 56 → 19 and the repo-wide backlog from 126 → 89.** Baseline re-recorded in each commit per the ratchet's own instruction.

Existing role helpers only — `resolveProjectColumnsForRoles`, `REVIEW_ROLES`, and the file's own `resolveReviewColumnsFor`. No new helper, no new vocabulary.

## What was actually broken

Each sweep resolves once and asks membership. The defects these guards were hiding are not uniform, and several **compound**:

| sweep | on a renamed board |
|---|---|
| `archiveStaleDoneTasks` | two failures at once: finished dependents read as *active* (holding cards back) **and** the stale filter matched nothing (archiving none) |
| `reconcileInReviewBranchRebind` | the review filter **is** the candidate set — the sweep was a silent no-op |
| `reconcileTaskWorktreeMetadata` | the ACTIVE guard exists to stop clearing a worktree out from under a running shell; it stopped recognising the live card and cleared anyway |
| `reconcilePreExecutionWorktrees` | the live-lane list matched nothing, so **live cards' worktrees became reclaimable** |
| `detectStalledCards` | finished work reported as stalled |
| `reconcileInReviewUnmetDependencies` | rebound never fired |
| `recoverAgentsRunningOnInactiveTasks`, `recoverDriftedAgentTaskLinks`, `cleanupStaleTempMergeWorktrees`, `reconcileDependencyBlockingLeases`, `reconcileCompletedBlockedTasks`, `reconcileOrphanedPendingStepResults`, `autoReboundPausedScopeDecay`, `recoverApprovedStrandedAiMergeCommit`, `recoverStarvedRefinementTriageTasks` | guard or candidate set silently empty/inverted |

`resolveProjectColumnsForRoles` seeds the legacy ids, so an unconverted board answers byte-identically.

## One conversion I got wrong, caught by an existing test

`recoverAgentsRunningOnInactiveTasks` skips agents linked to `in-progress / in-review / done / archived`. I read that as "every lifecycle lane" and resolved intake ∪ hold ∪ wip ∪ review ∪ finished.

**The omission of `todo` is the entire point of the sweep** — an agent still running against a card that fell back to a *waiting* lane is precisely the drift being recovered. My set swallowed it, and `self-healing.test.ts` → *"recovers durable running agents linked to todo tasks"* failed immediately. Corrected to wip ∪ review ∪ finished, with the reasoning recorded at the site.

That is the argument for converting a literal list only after reading why each member is in it — the set is rarely "all the obvious ones".

## Flagged, not guessed — the remaining 19

**Transition table (`1489–1504`, 9 guards).** A `from`/`to` classifier over move *pairs*, not a lane membership test. Converting one side without the other changes which transitions are recognised, and these feed recovery routing. Needs a deliberate decision about whether a renamed board's transitions map role-to-role.

**Pause-abort routing (`976`, `982`, `1005`, 4 guards).** `classifyPausedAbortWorkflowRecovery` is **synchronous by design** and called twice from `recoverPausedAbortFailures`. Converting means injecting resolved sets as parameters — the same shape as `node-override-guard`. Mechanical, but it changes a signature and I would rather it be a decision than a drive-by.

**`12142` / `12146` (2 guards) — the one I most want a second opinion on.** These pair with `moveTask(task.id, "todo", { recoveryRehome: true })`. `moveTask` **rejects** a target the workflow does not declare — *except* under `recoveryRehome` with a legacy id (`moves.ts:570`, the #1411 escape hatch). So converting the guard alone makes the sweep fire *less* often while the move still lands the card in an undeclared column. **The guard and its write target have to move together**, and writes are explicitly out of scope for this conversion.

**`1131`, `1398` (2 guards).** Sync predicates (`isWorkspaceOwnerLive`, `isPhantomExecutorBinding`) each with one caller; same injection shape as the pause-abort pair.

**`5902` (1 guard).** Already carries a written justification for staying literal — *"the degraded answer costs a duplicate log line … not a wrong lifecycle decision"*. It is sound, but it lacks the `DELIBERATE-LITERAL` marker, so the census still counts it as backlog. Adding the marker is a judgement about someone else's reasoning, so I left it and am naming it here.

## Separately: 20 recovery rehomes write a legacy literal

Enumerating write targets while checking `12146` turned up **20** `moveTask(..., "<legacy>", { recoveryRehome: true })` calls in this file. Each bypasses the undeclared-column rejection by design. On a renamed board every one of them rehomes a card into a column its workflow does not declare — the state `reconcile-task-state-consistency` and #2999's reconciler exist to repair.

Out of scope here (writes), and a real finding: the escape hatch that stops #1411's "card can never be rescued" also guarantees the rescue lands somewhere undeclared.

## Verification

- `tsc -p tsconfig.json` (engine): **0 errors**
- `self-healing.test.ts` + `self-healing-fake-overlap-seam` + `self-healing-orphaned-task-dirs` + `surfacing-family-shared`: **475 passed**
- `lifecycle-column-census --strict`: clean, baseline re-recorded per commit
- FNXC gate exit 0, lint clean
- Full engine suite running; will report before merge.

### DIFF
--- a/packages/engine/src/__tests__/self-healing-fanout-renamed-lanes.test.ts
+++ b/packages/engine/src/__tests__/self-healing-fanout-renamed-lanes.test.ts
@@ -1,0 +1,180 @@
+/*
+FNXC:WorkflowResolvedColumns 2026-07-31-14:45:
+THE `task:moved` FAN-OUT ANSWERED EVERY LANE QUESTION WITH A LITERAL.
+
+`taskMovedFanoutListener` decided three things by comparing column ids:
+
+  - a move OUT of `in-progress` into `todo`/`in-review`/`done`/`archived` increments the
+    board-stall transition counter,
+  - a move INTO `in-review` triggers the branch-rebind reconciliation,
+  - `in-review -> done` and `done -> archived` trigger the completion fan-out (worktree removal,
+    dependent `blockedBy` clearing).
+
+On a board that names its lanes anything else, all three stopped firing. Nothing errors: the
+counter silently reads zero, worktrees are never reclaimed, and dependents keep a `blockedBy`
+pointing at a blocker that already finished.
+
+WHY A SYNC RESOLVER. `task:moved` is emitted synchronously, so an `await` in this listener defers
+everything after it to a microtask and reorders this handler against every other subscriber — the
+hazard the scheduler's own `resolveTaskParkedColumnsSync` header documents. The conversion uses the
+store's sync IR path for that reason, which is a different seam from the `resolveProjectColumnsForRoles`
+used by the async sweeps and needs its own coverage.
+
+Driven through the REAL listener via `start()` + a real `task:moved` emit, not by calling the
+private resolver: the contract is that the fan-out fires, not that a helper returns a set.
+*/
+
+import { describe, expect, it, vi, beforeEach } from "vitest";
+import { EventEmitter } from "node:events";
+import type { Settings, Task, TaskStore, WorkflowIr } from "@fusion/core";
+
+const { logger } = vi.hoisted(() => ({
+  logger: { log: vi.fn(), debug: vi.fn(), warn: vi.fn(), error: vi.fn() },
+}));
+vi.mock("../logger.js", () => ({ createLogger: vi.fn(() => logger) }));
+
+import { SelfHealingManager } from "../self-healing.js";
+
+/** Hold `drafting`, wip `building`, review `reviewing`, complete `shipped`, archived `filed`. */
+function renamedIr(): WorkflowIr {
+  return {
+    version: "v2",
+    id: "custom:wf",
+    nodes: [],
+    edges: [],
+    columns: [
+      { id: "drafting", name: "drafting", traits: [{ trait: "hold", config: { release: "capacity" } }] },
+      { id: "building", name: "building", traits: [{ trait: "wip", config: { limitSetting: "maxConcurrent" } }] },
+      { id: "reviewing", name: "reviewing", traits: [{ trait: "merge" }] },
+      { id: "shipped", name: "shipped", traits: [{ trait: "complete" }] },
+      { id: "filed", name: "filed", traits: [{ trait: "archived" }] },
+    ],
+  } as unknown as WorkflowIr;
+}
+
+function makeTask(id: string, column: string): Task {
+  return {
+    id,
+    title: id,
+    description: id,
+    column,
+    dependencies: [],
+    steps: [],
+    currentStep: 0,
+    log: [],
+    createdAt: "2026-06-01T00:00:00.000Z",
+    updatedAt: "2026-06-01T00:00:00.000Z",
+  } as unknown as Task;
+}
+
+function createStore(tasks: Task[], ir: WorkflowIr | null): TaskStore & EventEmitter {
+  const map = new Map(tasks.map((t) => [t.id, t]));
+  const emitter = new EventEmitter();
+  const cfg = { globalPause: false, enginePaused: false } as Settings;
+  return Object.assign(emitter, {
+    getSettings: vi.fn(async () => cfg),
+    listTasks: vi.fn(async () => [...map.values()]),
+    getTask: vi.fn(async (id: string) => map.get(id) ?? null),
+    updateTask: vi.fn(async () => undefined),
+    logEntry: vi.fn(async () => undefined),
+    /* The sync seam under test. `null` stands for a workflow that cannot be resolved. */
+    resolveTaskWorkflowIrSync: vi.fn(() => {
+      if (!ir) throw new Error("no workflow selection");
+      return ir;
+    }),
+  }) as unknown as TaskStore & EventEmitter;
+}
+
+/** Starts the manager with its periodic maintenance suppressed — only the listener is under test. */
+function startManager(store: TaskStore & EventEmitter) {
+  const mgr = new SelfHealingManager(store, { rootDir: "/repo" });
+  vi.spyOn(mgr as unknown as { startMaintenance: () => void }, "startMaintenance").mockImplementation(() => {});
+  const reconcile = vi.spyOn(mgr, "reconcileCompletedTask").mockResolvedValue({
+    blockedByCleared: 0, worktreeRemoved: false, branchRemoved: false,
+  });
+  const rebind = vi.spyOn(mgr, "reconcileInReviewBranchRebind").mockResolvedValue(0 as never);
+  mgr.start();
+  return { mgr, reconcile, rebind };
+}
+
+/** `task:moved` is synchronous; the handlers it starts are not, so let the microtasks drain. */
+async function settle() {
+  await Promise.resolve();
+  await Promise.resolve();
+}
+
+describe("the self-healing task:moved fan-out on a renamed board", () => {
+  beforeEach(() => { vi.clearAllMocks(); });
+
+  it("runs the completion fan-out for reviewing -> shipped", async () => {
+    const task = makeTask("KB-1", "shipped");
+    const store = createStore([task], renamedIr());
+    const { mgr, reconcile } = startManager(store);
+
+    store.emit("task:moved", { task, from: "reviewing", to: "shipped", source: "engine" });
+    await settle();
+
+    /* Keyed on `in-review -> done` this never fired: the worktree is never reclaimed and every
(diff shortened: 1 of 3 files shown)


gsxdsm:
**Not rebasing — you are actively on it** (`kb/phase-c-review-merge`, dirty tree, last commit 5 minutes ago). A force-push from me would diverge your branch mid-work.

## Three of us converted this file

| PR | scope | state |
|---|---|---|
| **#3047** (mine) | 56 → 51, `archiveStaleDoneTasks` only | **merged** |
| #3050 | 56 → 26 | conflicting, owner active |
| **#3055** (this) | **56 → 19** | conflicting, owner active |

My merge is what put both of yours into conflict. I flagged the duplicate before mine landed and it did not prevent this, so the cost is real rather than hypothetical.

**This PR goes furthest and should be the one that lands.** Mine was the least ambitious of the three and it won only by merging first, which is not a good reason. If #3050 and this one overlap, #3050 is the one to close.

## Resolving against current main

1. **`scripts/lib/lifecycle-column-census-baseline.json` — generated, do not hand-merge.** Take main's version, then `node scripts/lifecycle-column-census.mjs --strict --update-baseline`. Hand-merging that JSON gives a number matching neither branch while looking plausible.
2. **`archiveStaleDoneTasks` — genuine overlap with what merged.** #3047 converted three guards there: active-dependents scan and temp-worktree age gate to `TERMINAL_ROLES`, and the staleness filter to `complete` **alone** — an already-archived card is not a candidate for archiving, so the terminal pair would make the sweep consider its own output. If your version differs, yours is the one to reason about; just do not take the terminal pair for that third guard.

## Two things that will bite after the rebase

**Re-measure.** Your `56 → 19` was taken against a baseline that now reads 51. The PR-body number has to match what `--strict` sees or review cannot check the claim.

**Partial conversion fails a ratchet.** Converting *some* guards in a sweep makes it register as converted with `self-healing-converted-sweeps-have-no-literal-lane-guards`, which then fails it for any literal left behind — that suite grew 36 → 38 when #3047 landed, and it caught exactly this in my own work. If a sweep you touched still has one guard on a legacy id, expect a red test; it is correct.

## One thing worth checking in your remaining sites

Some of what the census counts in this file is not owed work. I marked two in `async-mission-store-queries.ts` (#3056) that were **fallback arms** — `lanes?.complete ? lanes.complete.has(c) : c === "done"` — where the legacy id is the answer when resolution fails. Converting those deletes the fallback. Worth a scan before pushing 19 lower.

Happy to take the rebase once your tree is clean — say the word.

gsxdsm:
**Rebased onto `origin/main` (was CONFLICTING, now MERGEABLE) and re-verified.**

Three conflicts in `self-healing.ts` were the *same* guards converted independently on main (#3047 and the terminal-roles pass): `dependentTerminalColumns`/`doneColumns`/`sweepTerminalColumns` there vs my `archiveFinishedColumns`/`archiveCompleteColumns`/`stalledFinishedColumns`. Same roles, same semantics — I took main's spelling in all three and deleted my now-dead declarations rather than keeping two names for one answer.

The census baseline conflicted on every commit (it is derived, so every fleet PR rewrites it). Regenerated once from `origin/main` at the end — the census tool detects its own conflict markers and refused the intermediate state, which is how the bad baseline was caught rather than pushed.

### Last commit: `recoverPausedAbortFailures` finished

The converted-sweeps ratchet failed on the previous push, correctly. I had resolved review lanes in that sweep (making it "converted" by the derived rule) while leaving its two `todo` guards literal — exactly the partial state the ratchet's header names as *worse than converting nothing*.

Converting the guard alone would have been a live bug, not a cosmetic gap: `moveTask` rejects an undeclared target **except** under `recoveryRehome` with a legacy id (`moves.ts:570`, the #1411 escape hatch). A converted guard feeding the literal `"todo"` fires correctly on a renamed board and then rehomes the card into a column its workflow does not declare. So both convert together — `hold ∪ intake` membership for the guard, per-task `resolveTaskLifecycleColumns` with `hold ?? intake ?? "todo"` for the target (the precedence documented at `moves.ts:1296`; a workflow may declare intake and no hold).

### Census (post-rebase, measured)

| | before | after |
|---|---|---|
| `self-healing.ts` | 56 | **17** |
| repo backlog | 126 | **70** |

Backlog is lower than my file's delta alone because the rebase picked up the scheduler/executor/replan-target clusters that landed meanwhile.

### Verification (measured, post-rebase)

- `tsc --noEmit -p packages/engine` — clean
- `check-lane-wiring` — `12 known unwired call site(s), none added`
- `lifecycle-column-census --strict` — clean, baseline tightened (4 entries dropped)
- Full engine suite — **876 files / 11504 tests pass, 1 failure**

### The one failure is pre-existing on main, not mine

```
FAIL src/__tests__/scheduler-renamed-hold-events.test.ts
 > finds dependents resting in the renamed hold column when a blocker completes
 AssertionError: expected [] to include 'drafting'
```

**Proved, not assumed:** my diff touches `self-healing.ts` and the census baseline only — no scheduler file. I checked out `origin/main:packages/engine/src/self-healing.ts` over my copy and re-ran that test: **still fails, identically.** So `main` is red here independently of this PR. Filing it separately rather than folding a scheduler fix into a self-healing conversion PR.

gsxdsm:
## Status: this branch is being rebuilt, and one of its commits was wrong

Two things happened after the last push.

### 1. #3049 landed an overlapping conversion of the same file

Another worker converted a different 18 guards in `self-healing.ts`. Rebasing my six commits produced five conflict hunks in the first commit alone, all of them "the same guard, two spellings". Replaying six commits through that is churn with no reviewer value, so I am rebuilding this as a single commit off current `main` containing only what `main` does not already have — measured against `main`'s own census (`self-healing.ts` is still at **39 column hits** there, so the bulk of this work is genuinely additive, not duplicated).

### 2. The sync-path conversion merged as #3069 is INERT — my error

#3069 (merged into this branch, not into main) added `resolveFanoutLanesSync` to convert the `task:moved` fan-out. I verified #3058's refutation against the source rather than trusting it:

`getTaskWorkflowSelectionImpl` returns `undefined` **unconditionally** — "Backend mode cannot synchronously read PostgreSQL". So `resolveTaskWorkflowIrSync` always returns the **default builtin IR**, and `columnsWithFlag(defaultIr, …)` yields exactly `todo / in-progress / in-review / done / archived`. My legacy seeding then adds nothing. **The conversion changes no behaviour on any board.**

My own test file passes only because its store mock supplies a renamed IR directly — it pins the helper's shape, not production behaviour. That is precisely the failure mode I have been asserting against all week: a test that proves the code does what it says while the code does nothing.

Worse than the literal, for the reason #3058 states: the literal was **counted**, and this is not. Four guards left the backlog and the file now reads as converted.

**So the fan-out conversion comes out.** The rebuilt branch keeps the literals there with an FNXC note recording the refutation and naming the real blocker, and keeps only the conversions that resolve through the **async** path:

- `classifyPausedAbortWorkflowRecovery`, `isWorkspaceOwnerLive`, `isPhantomExecutorBinding` — required lane parameters, resolved by their async callers via `resolveProjectColumnsForRoles`. Not affected by the sync-reader defect.
- the `evaluateParkedAgentTaskLink` `parkedColumns` wiring — same.
- `recoverPausedAbortFailures`' guard **and** its `moveTask` target, converted together.

The single blocker behind every inert sync conversion — #3051's ten, the twenty `scheduler.ts` guards the new `check-inert-sync-lane-conversions` gate baselines, and mine — is that there is **no sync-capable workflow-selection reader**. One change fixes all of them. I am flagging it, not attempting it inside a conversion PR.

gsxdsm:
Superseded by the rebuilt branch — see the status comment above. Two reasons: #3049 landed an overlapping conversion of the same file, and one commit on this branch (#3069's sync-IR fan-out conversion) was inert and has been withdrawn rather than carried forward.

PR #3056, opened 2026-07-31
chore(core): mark the async-mission fallback arms DELIBERATE-LITERAL (census 108→106)

gsxdsm:
Fleet phase. Claimed `packages/core/src/async-mission-store-queries.ts` — **the only census file with no branch, no worktree, and no open PR against it.** Claim published by pushing the branch before doing any work.

## Census before / after

| | total | this file | deliberate |
|---|---|---|---|
| before | **108** | 2 | 128 |
| after | **106** | 0 | **130** |

`--strict` exits 0, baseline re-recorded in the same commit.

**This is a reclassification, not a conversion.** The same two lines are still there. A reader comparing 108 → 106 against my #3047's 126 → 121 should know only the latter changed behaviour.

## Why marking is the right answer here

Both sites are the **fallback arm** of the three-state rule:

```ts
terminalColumns?.complete ? terminalColumns.complete.has(column) : column === "done";
```

`terminalColumns` undefined means the caller could not resolve lanes. The legacy id is then the only answer that keeps the query working at all — converting it would delete the fallback and make an unresolvable caller return nothing. The census counts the literal, but **the literal is the design**.

The file's own comment shows a previous worker already reached this conclusion. Nothing recorded it in a form the tool reads, so it stayed in `byFile` as apparent backlog for the next pass to re-derive.

## The finding this makes concrete

I checked five unclaimed files this phase (`agent-store`, `github-tracking-state`, `planner-overseer`, `auto-merge-finalization`, this one). **Every site in them was either a fallback arm or an already-documented deliberate leave** — `agent-store.ts:236` carries a comment from today's fleet phase explaining why it stays.

So the remaining count is not a work queue. A meaningful share is correct code the tool cannot distinguish from owed work, and each fleet pass pays to re-derive that. Marking them is cheap, mechanical, and makes the number mean "conversions owed" — which is what every worker reads it as when picking a cluster.

I marked only the file I claimed. The others belong to whoever holds them.

## Verification

- `census --strict` exit 0; `tsc --noEmit` **0 errors**
- `check:fnxc-future-dates`, `check:lane-wiring`, `check:sql-column-literals`, `check:inert-flag-seams` — all exit 0
- No behaviour change: the two expressions are byte-identical, only comments added

## Note on the marker's granularity

The first marker covered only `isComplete` — the census attaches markers by *preceding comment*, so the sibling `isArchived` needed its own. Caught by re-running the census (2 → 1, not 2 → 0) rather than by reading. Worth knowing before marking a group of related literals.


gsxdsm:
**Rebased onto current main (fifth rebase across my two PRs — same cause each time), and this one now also tightens a loose baseline.**

The rebase conflict was, again, `scripts/lib/lifecycle-column-census-baseline.json` and nothing else. Source merged clean.

**Main's committed baseline had drifted loose: recorded 91, actual tree 88.** Someone merged a conversion without re-recording, and the failure is quiet by design — running the census rewrites the baseline downward and reports:

> The baseline file has been rewritten downward. COMMIT IT so the allowance cannot be regrown into; in CI this write is discarded with the runner, which is why the gate is green and not silent.

So CI passes while the committed allowance sits above reality. A regression could be re-introduced into that slack without the ratchet firing — which is the one failure a ratchet must not have.

Re-recorded here to **86**, because this PR touches that file anyway. Folding it in adds no new collision; a standalone baseline PR would have conflicted with all ten open fleet PRs.

## The structural point, now with a count

Five rebases across #3049 and #3056. Four had the baseline as the **only** conflict:

| PR | rebases | conflict |
| --- | --- | --- |
| #3049 (merged) | 3 | baseline ×2, plus duplicate conversions of the same sweeps vs #3047 |
| #3056 | 2 | baseline only, both times |

Every conversion PR must touch that one generated file, so **N concurrent PRs force N−1 rebases even when no source overlaps**. With ten open, that is ~45 pairwise opportunities against a file whose content is fully derived.

Regeneration is deterministic — `--strict --update-baseline` from a clean checkout reproduces it byte-identically, which is why resolving by regenerating is safe and why either fix works: regenerate post-merge in tooling and keep it out of PR diffs, or let workers commit source-only and re-record in one periodic sweep.

Verified after rebase: mission suites 8 passed, tsc clean, census --strict green, lane-wiring and fnxc ratchets green, branch on top of `origin/main`.

PR #3057, opened 2026-07-31
fleet: reads.ts 2 → 0 lifecycle-column guards

gsxdsm:
> **Rebased.** Main landed another worker's conversion of the review gate while this was open — the overlap was a whole rewritten function, so I reset to main and rebuilt only my remaining delta on top of their work rather than resolving hunks. Their conversion is kept as-is.

## Census

| | column guards |
|---|---|
| before | **104** |
| after | **102** |

`reads.ts`: **2 → 0**.

## Two changes

**1. `includeColdStorage`** asks whether the *caller* is filtering to the archive lane. Against the literal, a caller filtering to a renamed archive lane took the false branch — cold storage was skipped and the filtered view returned only whatever archived rows still sat in `project.tasks`, **a short list presented as the whole archive**. Still literal on main; converted here.

**2. Both fallbacks become named sets** instead of inline arms — including the one on the just-landed review gate.

## The second point is the one worth the fleet's attention

This is bookkeeping correctness, not style. The census counts an inline comparison **whether or not it sits in a fallback branch**, because its `traitFallback` hint is advisory and never changes `kind`.

So a correctly-converted guard with an inline legacy arm **stays on the backlog permanently**, and the number stops distinguishing real debt from documented degraded answers.

Concretely: converting with an inline fallback is correct work that scores **zero**. My own first pass at this file did exactly that. There are roughly **12 such sites** across the tree — `github-tracking-state`, `planner-overseer`, `async-mission-store-queries`, `register-task-workflow-routes`, `restart-recovery-coordinator` — and I have that cluster converted and ready to open next.

## Measured

| check | result |
|---|---|
| reads / get-task / stall suites | 5 files, **87 tests green** |
| renamed-archive PG suite | green |
| strict census | green; `tsc` clean |
| unconverted boards | byte-identical — the named sets hold the previous ids |

## Summary by CodeRabbit

* **Bug Fixes**
  * Review and archive checks now work correctly with resolved workflow columns while retaining legacy compatibility.
  * Fresh agent activity is detected in resolved review lanes.
  * Lists filtered by a resolved archive lane now include archived items stored in cold storage.

* **Chores**
  * Updated internal lifecycle tracking baselines.

### DIFF
--- a/packages/core/src/task-store/reads.ts
+++ b/packages/core/src/task-store/reads.ts
@@ -95,13 +95,26 @@ on a renamed board, so `executingTaskIds` stayed empty and the board showed "Sta
 while a merger was visibly making progress — the precise regression the FNXC note below says this
 function was restored to prevent.
 */
+/*
+FNXC:WorkflowResolvedColumns 2026-07-31-14:05 (fleet — inline fallback arms):
+DELIBERATE-LITERAL — the no-resolution fallbacks for the two lane questions in this file.
+
+Named sets rather than inline `=== "<id>"` arms. Behaviour is identical to the guards as they stand;
+the reason is that the census counts an inline comparison whether or not it sits in a fallback branch
+— its `traitFallback` hint is advisory and never changes the count. So a correctly-converted guard
+with an inline legacy arm stays on the backlog permanently, and the number stops distinguishing real
+debt from documented degraded answers. Same shape as `LEGACY_PLANNER_LANES`.
+*/
+const LEGACY_REVIEW_LANES: ReadonlySet<string> = new Set(["in-review"]);
+const LEGACY_ARCHIVE_LANES: ReadonlySet<string> = new Set(["archived"]);
+
 function hasFreshAgentLogActivitySinceTaskUpdate(
   store: TaskStore,
   task: Pick<Task, "id" | "column" | "updatedAt">,
   now: number,
   reviewColumns?: ReadonlySet<string>,
 ): boolean {
-  if (!(reviewColumns ? reviewColumns.has(task.column) : task.column === "in-review")) return false;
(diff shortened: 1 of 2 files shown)


gsxdsm:
The archive-filter half is a real conversion — `resolveProjectColumnsForRoles` is async and actually reads the board — and the `hasFreshAgentLogActivitySinceTaskUpdate` threading matches what #3053 landed. No objection to either.

One thing worth deciding deliberately rather than inheriting: moving the fallback literals into `LEGACY_REVIEW_LANES` / `LEGACY_ARCHIVE_LANES` drops the census count **because the census counts comparisons, and a Set membership test is not one** — the guard is still there, it just stopped being visible. That is the same shape as #2980 ("a lane id hoisted into a const evaded the SQL column-literal gate"), and the repo treated it as an evasion that time.

Your note argues the other side openly, and I think the argument is reasonable: an inline fallback arm does sit on the backlog forever, which does make the number stop tracking real work. That is a genuine defect in the counter, and I hit it from the other direction in #3053 — an optional-resolved-parameter conversion leaves the census at 126→126 even though behaviour changed.

The concern is what we lose. I measured the split across all 126 before it moved: **18 were fallback arms of already-converted seams, 108 were bare guards.** Once fallbacks live in named sets, that split is no longer measurable from the census at all — "converted seam with a documented degraded answer" and "never touched" become indistinguishable, in the direction that looks finished.

Cheap fix that keeps both properties, if you want it: have the census count named legacy-lane sets under the existing `DELIBERATE-LITERAL (reviewed)` total instead of dropping them entirely. The backlog number then tracks real work *and* the fallbacks stay counted somewhere, so nobody has to re-derive the distinction later.

Not a blocker on this PR — the behaviour change is sound. Flagging because the counting decision outlives it, and #3058 is an example of what happens when the count reads as paid and the behaviour did not move.

PR #3065, opened 2026-07-31
fix(engine): a second complete lane is terminal too — restore the scheduler's dependency reconciliation (main is red)

gsxdsm:
## main is red, and this is the fix

```
FAIL src/__tests__/scheduler-renamed-hold-events.test.ts
 > dependency unblocking (failure mode is a card that waits forever)
 > finds dependents resting in the renamed hold column when a blocker completes
 AssertionError: expected [] to include 'drafting'
```

Not in the thin merge gate's `engine-core` allow-list, so CI stayed green and only the non-blocking full suite sees it. The test file is unchanged since #2518; #3051 converted the guard underneath it.

## What broke

#3051 turned the guard into `to === parked.complete || to === parked.archived`. `resolveLifecycleColumns` answers **first match per role** — the right shape for a move *target*, the wrong shape for *"did this card just reach a finished lane"*, which is a membership question.

Two consequences, both silent:

1. **A board with more than one complete-trait column reconciles nothing** when a blocker finishes in the second one. The test file's own header flags this path specifically: *"This one is NOT latency: a dependent never gets unblocked, so it waits on a blocker that is already done."*
2. The legacy `done`/`archived` ids stopped matching at all — the failing assertion.

## Fix

`resolveTaskParkedColumnsSync` gains `terminal`, a membership set: legacy `done`/`archived` seeded, then **every** complete- and archived-trait column from the task's own IR.

Seeding legacy ids is safe in the direction that matters here. This is an **inclusion**: a superset makes the reconciliation run on a move it would otherwise ignore — one extra query, and it cannot wrongly withhold work. Seeding a **refusal** is the bug (`node-override-guard.ts` documents that one); this is not that.

Same sync IR path and same fail-soft legacy default as the single-column answers, so event ordering and unresolvable-workflow behaviour are unchanged — the constraint the sync resolver's own header sets.

The two sibling guards in the same listener (dispatch-oscillation reset at what is now line 1097, and the scheduling wake at 1114) had the identical arity defect and convert with it.

## Measured

| | result |
|---|---|
| before | `scheduler-renamed-hold-events`: **1 failed / 9 passed** |
| after | **11 passed** (one new case) |
| `src/__tests__/scheduler*` | **14 files / 143 tests pass** |
| `tsc --noEmit -p packages/engine` | clean |
| census `--strict` / `check-lane-wiring` / `check-fnxc-future-dates` | clean, no baseline movement |

**Proved it is main's red, not my branch's:** I checked out `origin/main:packages/engine/src/self-healing.ts` over my unrelated fleet branch and re-ran — identical failure. Then branched this fix straight off `origin/main`.

**Mutation-tested.** Restoring `to === parked.complete || to === parked.archived` fails **both** the pre-existing case and the new second-complete-lane case. The new test is not vacuous: the second complete lane is invisible to first-match resolution, so it cannot pass against the old guard.

## Not done here

I did not convert the remaining `to === parked.review` / `from === parked.wip` single-column comparisons in this listener. Review is genuinely two roles (`mergeBlocker` + `humanReview`) and wip has its own limit-setting semantics — both want the same membership-vs-target judgement applied deliberately rather than swept in behind a red-fix. Flagged, not guessed.

### DIFF
--- a/packages/engine/src/__tests__/scheduler-renamed-hold-events.test.ts
+++ b/packages/engine/src/__tests__/scheduler-renamed-hold-events.test.ts
@@ -51,7 +51,7 @@ function renamedIr(): WorkflowIr {
   } as unknown as WorkflowIr;
 }
 
-function createStore(tasks: Record<string, unknown>[] = []) {
+function createStore(tasks: Record<string, unknown>[] = [], ir: WorkflowIr = renamedIr()) {
   const listeners = new Map<string, ((payload: unknown) => void)[]>();
   const selection = { workflowId: WF, stepIds: [] };
   const listTasks = vi.fn(async (opts?: { column?: string }) =>
@@ -73,8 +73,8 @@ function createStore(tasks: Record<string, unknown>[] = []) {
     getCompletionHandoffAcceptedMarker: vi.fn().mockResolvedValue(null),
     getTaskWorkflowSelection: vi.fn(() => selection),
     getTaskWorkflowSelectionAsync: vi.fn(async () => selection),
-    getWorkflowDefinition: vi.fn(async () => ({ ir: renamedIr() })),
-    resolveTaskWorkflowIrSync: vi.fn(() => renamedIr()),
+    getWorkflowDefinition: vi.fn(async () => ({ ir })),
+    resolveTaskWorkflowIrSync: vi.fn(() => ir),
   } as unknown as TaskStore;
 
   return {
@@ -130,8 +130,9 @@ function createAgentStore(agents: Record<string, unknown>[], freshRun: unknown =
 function createScheduler(
   tasks: Record<string, unknown>[] = [],
   options: Record<string, unknown> = {},
+  ir: WorkflowIr = renamedIr(),
 ) {
-  const { store, emit, listTasks } = createStore(tasks);
+  const { store, emit, listTasks } = createStore(tasks, ir);
   const scheduler = new Scheduler(store, options as never);
   const schedule = vi.spyOn(scheduler, "schedule").mockResolvedValue(undefined);
   (scheduler as unknown as { running: boolean }).running = true;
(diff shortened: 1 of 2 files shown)


gsxdsm:
## Correction — the multi-lane claim in this PR is inert in production

#3058 landed after I opened this, and it refutes part of my own framing. I verified it independently rather than taking it on trust:

`getTaskWorkflowSelectionImpl` (`workflow-definitions.ts:521`) returns `undefined` **unconditionally** — "Backend mode cannot synchronously read PostgreSQL". So `resolveTaskWorkflowIrSync` always takes its `!workflowId` branch and returns the **default builtin IR**, on every board, under the shipped backend.

Consequence for this PR: `parked.terminal` resolves to `columnsWithFlag(defaultIr, "complete"|"archived")` ∪ the legacy seed = `{done, archived}`. Identical to what it replaces. **The "a board with two complete lanes" scenario cannot occur in production today**, and my new test passes only because the harness mocks `resolveTaskWorkflowIrSync` with a renamed IR.

### What this PR still does, and it is not nothing

**It fixes the actual red.** #3051 replaced the literal pair with `to === parked.complete || to === parked.archived`, where `parked.complete` is a *single first-match* column. Under the sync path that is `"done"` — but the test's mocked IR makes it `"shipped"`, so the legacy `done` id stopped matching and dependency reconciliation broke. Seeding the membership set restores it. That is measured: 1 failed → 11 passed, reproduced on `origin/main` first.

So: **correct red-fix, overstated headline.** The guard is now the right *shape* (membership, not first-match) and will start doing real work the moment the sync path can resolve — but it does no new work today.

### What I am NOT doing

I am not converting this to the async resolver in this PR. That would make a synchronous `task:moved` subscriber async and reorder it against every other subscriber — the hazard `resolveTaskParkedColumnsSync`'s own header exists to avoid — and it does not belong in a fix for a red main.

The real blocker is a **sync-capable selection reader** (or an async listener contract). That is one change that would un-inert every sync-path conversion at once, including #3051's ten and the twenty the new `check-inert-sync-lane-conversions` gate now baselines in `scheduler.ts`. Flagging it rather than guessing at it.

Retitling would be honest; I have left the title so the review thread stays legible, and this comment is the correction of record.

PR #3067, opened 2026-07-31
fix(engine): an archived dependency made every backlog candidate unrunnable (long-tail audit: 16 files, 1 real defect)

gsxdsm:
Fleet long tail. I audited **all 16 unclaimed files (17 guards)** and converted the one that is a genuine defect. The other 15 are reported below rather than converted — several would be actively wrong to convert.

## Census

| | before | after |
|---|---|---|
| repo-wide column guards | 93 | **88** |

## The defect

#2820 converted this reporter's two **queries** and left the third lane question — the runnable-candidate check — comparing each dependency to the id `done`, two lines from where the other two are resolved:

```ts
const [holdColumns, wipColumns] = await Promise.all([...]);   // resolved
...
if (dependency.column !== "done") return false;               // not
```

A dependency is satisfied when it reaches a **terminal** lane — complete *or* archived. Keyed on the single id, a candidate blocked by an archived dependency counted as not-runnable, the candidate list came back empty, and the reporter reported **no backlog pressure on a board genuinely under it**. A diagnostic going quiet rather than wrong.

**The archived case is wrong on the built-in board too**, not only a renamed one — which is exactly why it survived the renamed-board tests #2820 added alongside it.

Reverted, the new case fails. Optional parameter with the legacy id as the documented fallback; the caller already awaited two role resolutions, so this is a third entry in the same `Promise.all`.

## The other 15 — flagged, not guessed

I opened every one. **None is a missed conversion:**

- **Sentinel comparisons** (`async-comments-attachments.ts`, 4 sites). These compare against a sentinel string the resolver itself returns, which deliberately collapses archived and soft-deleted into one value. Its comment spells out that converting *"would keep passing on the built-in board and start FAILING on a renamed one — a soft-deleted task would read as not-archived."* Converting would introduce the bug.
- **Sync-DB paths** (`merge-queue-ops-2`, `mission-store`, `audit-ops`, `lifecycle-ops`, `task-id-integrity`). Inside synchronous SQLite transactions, where the only sync resolver reads `getTaskWorkflowSelectionImpl` — which returns `undefined` unconditionally in PostgreSQL mode. One carries the note: *"Left as a literal WITH this note so the next worker does not 'fix' it into a false green."* That is exactly what a mechanical pass here would produce.
- **Already-converted seeds** (`executor.ts` ×3, `auto-merge-finalization.ts`). Literal sets unioned with `columnsWithFlag` results on the very next lines — the documented degradation path for when `resolveWorkflowIrForTask` falls back to the built-in IR.
- **`taskRevert.ts`** — its comment establishes that supplying the modal task's flags for a *neighbour* row is the flags-for-the-wrong-row hazard, and concludes the entry is *"accurate debt rather than a missed conversion."*
- **`notification-service.ts`** (5, largest unclaimed) — blocked on the wedge-episode contract: any `await` before the resolve drops an operator notification, with a test that fails loudly.

**The residual census count is mostly intentional.** Six of eight remaining files carry explicit flagged notes. Treating the number as a work queue from here produces false greens, not conversions — which several of those authors predicted in writing.

## Verification

backlog-pressure-reporter **11 passed** · `pnpm test:gate` 161 + 13 + 487 + 71 · lint · census `--strict` · lane-wiring · changesets — green.

### DIFF
--- a/.changeset/backlog-pressure-dependency-lanes.md
+++ b/.changeset/backlog-pressure-dependency-lanes.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Backlog-pressure alerts no longer go silent when candidates depend on archived tasks.
+category: fix
+dev: `isRunnableCandidate` takes the caller's resolved TERMINAL set instead of comparing each dependency to the `done` id.
--- a/packages/engine/src/__tests__/backlog-pressure-reporter.test.ts
+++ b/packages/engine/src/__tests__/backlog-pressure-reporter.test.ts
@@ -284,4 +284,43 @@ describe("backlog pressure resolves the board's own lanes", () => {
 
     expect((await reporter.report()).alerted).toBe(false);
   });
+
+  /*
+  FNXC:WorkflowResolvedColumns 2026-07-31-15:25:
+  The DEPENDENCY half. #2820 converted this reporter's two queries and left the runnable-candidate
+  check asking whether each dependency was the id `done` — so a candidate blocked by an ARCHIVED
+  dependency counted as not-runnable, the candidate list came back empty, and the reporter reported no
+  pressure on a board that was genuinely under it. A diagnostic going quiet, not wrong.
+
+  Archived is the load-bearing case: it is wrong on the BUILT-IN board too, not only a renamed one.
+  */
+  function storeWithArchivedBlocker(): TaskStore {
+    const blocker = { id: "B-1", column: "archived", title: "B-1", priority: "normal" } as unknown as Task;
+    const hold = Array.from({ length: 10 }, (_, i) => (
+      { id: `H-${i}`, column: "backlog", title: `H-${i}`, priority: "normal", dependencies: ["B-1"] } as unknown as Task
+    ));
+    const wip = [{ id: "W-0", column: "building", title: "W-0", priority: "normal" } as unknown as Task];
+    return {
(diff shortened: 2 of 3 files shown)


gsxdsm:
**Rebased onto main and pushed** — branch was not checked out in any worktree.

The only conflict was `scripts/lib/lifecycle-column-census-baseline.json`, which is **generated**. Resolved by taking main's version and re-running `--strict --update-baseline`, not by hand-merging: a hand-merge of that JSON produces a count matching neither branch while looking entirely plausible. Your three source files applied cleanly.

## Verification

- `backlog-pressure-reporter.test.ts` — **11 pass**
- `tsc --noEmit` (engine) — **0 errors**
- `census --strict` — exit 0, total **84**

## Worth knowing, since this will recur

Nearly every open fleet PR regenerates this baseline, so it conflicts on essentially all of them and each rebase invalidates the others. That is why the retry loop keeps firing on the same set. The resolution is always the same and always mechanical — take main's copy, re-run the generator, never merge the text.

The census ratchet also tracks the **deliberate** count alongside the guard count, so a re-record is required in the same commit rather than optional. If you see `Unconverted debt did NOT increase — a marker moved these sites out of the guard count`, that is the ratchet refusing to let a marker quietly shrink the backlog, not a failure.

gsxdsm:
Superseded by #3081 — same line, same fix, and theirs is the more thorough treatment. The long-tail audit findings that were unique to this PR are preserved as a comment on #3081 rather than lost with it. Third duplicate-effort collision today; see #3078 for the same pattern on self-healing.ts, now at five concurrent PRs.

PR #3071, opened 2026-07-31
fix(core): the merge result reported a column the finaliser did not write (merge-queue-ops 3 → 0)

gsxdsm:
Largest unclaimed census cluster in `packages/core` — three `done` literals in `mergeTaskImpl`. Two of them produced **wrong state**, not merely a guard that stopped firing.

## 1. The result overrode the writer

`moveToDoneImpl` resolves the board's completion lane and writes it onto the task object:

```ts
task.column = completeColumn;   // task-artifacts-ops.ts
```

Both merge call sites then did:

```ts
result.task = { ...task, column: "done" };
```

putting the literal back over what the writer had just set. Every `task:merged` listener — GitHub tracking, the auto-merge handoff — was told the card landed in `done` while the persisted row said `shipped`.

The row was right and the event was wrong, which is the worse direction: the listeners act on the event, not the row.

Fixed by reading back what the writer set (`{ ...task }`). Deliberately **not** a second resolution — that would only be a second chance to disagree with the finaliser.

## 2. The guard disagreed with the writer

The already-complete short-circuit asked `task.column === "done"`, while the finaliser it guards short-circuits on the resolved `task.column === completeColumn`. On a renamed board those two answers differ, so a card already resting in the board's completion lane fell through and the merge ran again against a branch that was already landed and deleted.

Converted with the **same resolution and the same shape** — a single first-match column, not membership — because the whole point is that these two answers cannot differ. A workflow declaring no complete lane resolves to `undefined`, which matches no column; the finaliser refuses such a board explicitly one function later.

## Census

| | before | after |
|---|---|---|
| `merge-queue-ops.ts` | 3 | **0** |

## Measured

- Two new cases added to `merge-blocker-renamed-review-lane.test.ts` (same renamed-board fixture, same PG harness) — file **5/5 pass**.
- **MUTATION**: restoring either literal fails **both** new cases and leaves the three pre-existing ones green.
- Reached with **no git fixture**: with no branch present, `git rev-parse --verify` fails and the function takes its own documented *"branch not found — moving to done without merge"* path — which is exactly the path that calls `moveToDone` and then builds the result. No repo setup, no flake surface.
- `packages/core` targeted run: **38 tests pass**.
- `tsc --noEmit -p packages/core` clean; census `--strict`, `check-lane-wiring` ("none added"), `check-fnxc-future-dates` clean.

## Not done here (flagged, not guessed)

The other `done`/`archived` literals still in the core census are each blocked for a *different* documented reason, so sweeping them into this PR would have meant guessing:

- `agent-store.ts:236` — a pure formatter over `Pick<Task,"column">` that prints the column for a human; degrades gracefully and has no store to resolve from.
- `async-mission-store-queries.ts` — already converted with caller-threaded lane sets.
- `taskRevert.ts:119` — classifies a **neighbour** task; the only flags in scope describe the modal's own task, so wiring them would answer the question for the wrong row. Needs per-neighbour flags.
- `moves.ts:310` — a **refusal**, where a legacy-seeded superset is the documented hazard rather than the safe direction. Wants its own change with its own test.
- `mission-store.ts:2332` — a sync SQLite path with no async seam.

Each is real debt; none is a mechanical conversion.

### DIFF
--- a/packages/core/src/__tests__/merge-blocker-renamed-review-lane.test.ts
+++ b/packages/core/src/__tests__/merge-blocker-renamed-review-lane.test.ts
@@ -183,3 +183,78 @@ pgDescribe("mergeTask resolves the review lane from the task's own workflow", ()
     expect(message).toContain("signoff");
   });
 });
+
+/*
+FNXC:WorkflowResolvedColumns 2026-07-31-15:45:
+THE MERGE RESULT ASSERTED A COLUMN IT DID NOT SET.
+
+`moveToDoneImpl` resolves the board's completion lane and WRITES it onto the task object
+(`task.column = completeColumn`). `mergeTaskImpl` then did
+
+    result.task = { ...task, column: "done" };
+
+putting the literal back over what the writer had just set. Every `task:merged` listener — GitHub
+tracking, the auto-merge handoff — was therefore told the card landed in `done` while the persisted
+row said `shipped`. The row was right and the event was wrong, which is the worse direction: the
+listeners act on the event.
+
+The companion defect is the already-complete short-circuit at the top of the same function, which
+asked `task.column === "done"` while the finaliser it guards asks the RESOLVED question. On a renamed
+board they disagreed, so a card already resting in the completion lane fell through and the merge ran
+again against a branch that was already landed and deleted.
+
+Reached without any git fixture: with no branch present, `git rev-parse --verify` fails and the
+function takes its documented "branch not found — moving to done without merge" path, which is the
+one that calls `moveToDone` and then builds the result.
+*/
+pgDescribe("the merge result reports the column the finaliser actually wrote", () => {
+  let harness: PgTestHarness;
+  let store: TaskStore;
+
+  beforeEach(async () => {
+    harness = await createTaskStoreForTest({ prefix: "fusion_merge_result_lane" });
+    store = harness.store;
+  });
+
+  afterEach(async () => {
+    await harness?.teardown();
+  });
+
+  async function cardInReviewLane(description: string) {
+    const created = await store.createWorkflowDefinition({ name: "renamed merge lanes", ir: RENAMED_IR as never });
+    const task = await store.createTask({ description });
+    await store.selectTaskWorkflow(task.id, created.id);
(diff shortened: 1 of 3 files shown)


gsxdsm:
Collision note first: my #3073 converted the same `task.column === "done"` guard. **Yours supersedes mine** — it does all three sites in the file and catches the `column: "done"` override in the merge result, which mine did not. I am dropping that half from #3073 and keeping only its unrelated half (re-greening the archived-gate parity ratchet, which is red on main independently of both of us). The file is yours; no rebase needed on your side.

One substantive point before it lands.

```ts
const alreadyCompleteColumn = (await resolveTaskLifecycleColumns(store, id))?.complete ?? "done";
if (task.column === alreadyCompleteColumn) {
```

`resolveTaskLifecycleColumns(...).complete` is **first-per-role** — it returns one id. A board declaring more than one complete-trait column matches only the first, so a card resting in the second falls through exactly as it did against the literal, just less visibly.

That is not hypothetical, and it is not old: **#3065 is open right now fixing precisely this class one file over** — *"a second complete lane is terminal too"* — replacing `to === parked.complete` with a `terminal` set in `scheduler.ts`, because a single `complete` id was demonstrably wrong there.

The membership spelling is already the documented house style in **this same function**, ~40 lines below, from #2820's review:

> **MEMBERSHIP, not first-per-role:** `resolveReviewColumns` unions mergeOrchestration, mergeBlocker and humanReview, so a workflow splitting those across columns has all of them accepted.

So `columnsWithFlag(ir, "complete")` as a set, tested with `.has(task.column)`, would make the two halves of this function agree with each other and with #3065. Cheap now, and it reuses the IR read you are already taking.

The `?? "done"` fallback itself is correct as written — it engages only when the board resolves nothing, which is the semantics #2820 asked for. No concern there.

Worth confirming against `check:inert-sync-lanes` too (landed in #3062/#3068): `resolveTaskLifecycleColumns` is the async resolver, so this is a real conversion and the ratchet correctly stays silent — noting it only because three PRs this phase reached for a *sync* resolver that always answers with the default board.

PR #3074, opened 2026-07-31
fleet(engine): self-healing 39 → 11 lane guards — the async-reachable ones, and one conversion withdrawn

gsxdsm:
Replaces #3055, which I am closing. #3049 landed an overlapping conversion of the same file while that branch was open; rebasing six commits through it produced five conflict hunks in the first commit alone, all of them "same guard, two spellings". This is rebuilt off current `main` as one commit containing only what `main` does not have. `main`'s own census still shows **39 column hits** in this file, so the work is additive, not duplicated.

## The important part: I withdrew one of my own conversions

#3069 (merged into the old branch, never into main) added `resolveFanoutLanesSync` to convert the `task:moved` fan-out's four guards. **It is inert.** I verified #3058's refutation against the source rather than trusting it:

`getTaskWorkflowSelectionImpl` (`workflow-definitions.ts:521`) returns `undefined` **unconditionally** — *"Backend mode cannot synchronously read PostgreSQL"*. So `resolveTaskWorkflowIrSync` always takes its `!workflowId` branch and answers with the **default builtin IR**. `columnsWithFlag` on that IR yields exactly `todo / in-progress / in-review / done / archived` — identical to the literals it replaced, on every board.

My own test passed only because its store mock supplied a renamed IR directly. It pinned the **helper's shape**, not production behaviour — the exact failure mode I have spent this program asserting against.

And it is *worse* than the literal, for #3058's reason: **the literal is counted.** Four guards would have left the census, the file would read as converted, and the next reader would have no cause to look again.

So the literals stay, counted, with the refutation recorded in place and the real blocker named: **there is no sync-capable workflow-selection reader.** One change un-inerts every sync-path conversion at once — #3051's ten, the twenty `scheduler.ts` guards the new `check-inert-sync-lane-conversions` gate baselines, and mine. I am flagging it, not attempting it inside a conversion PR.

## What IS converted — all async-reachable

| site | shape |
|---|---|
| `classifyPausedAbortWorkflowRecovery` | required lane params, resolved by its async caller |
| `isPhantomExecutorBinding` | required `wipColumns` |
| `recoverPausedAbortFailures` | guard **and** `moveTask` target, together |
| `recoverAgentsRunningOnInactiveTasks` | lifecycle skip set + the omitted `parkedColumns` argument |

**Required, not optional-with-a-literal-default.** An optional parameter leaves the legacy ids in the file as a silent fallback and the next caller gets pre-conversion behaviour by writing nothing. `resolveProjectColumnsForRoles` seeds the legacy ids itself, so unconverted boards are unchanged.

**Guard and target convert together.** `moveTask` rejects an undeclared target *except* under `recoveryRehome` with a legacy id (`moves.ts:570`, the #1411 escape hatch). A converted guard feeding the literal `"todo"` would fire correctly on a renamed board and then rehome the card into a column its workflow does not declare.

**`isPhantomExecutorBinding` is the sharpest case**: its caller resolved `lanesOfReclaim(task.id).wip` **three lines above the call**, then passed a task whose column the predicate compared against the `in-progress` literal.

**A real dropped-behaviour bug fixed in passing**: the `evaluateParkedAgentTaskLink` call omitted `parkedColumns`, falling back to `LEGACY_PARKED_COLUMNS` (`todo`/`triage`). A live durable agent linked to a card resting in a renamed hold lane read as not-parked, so the safeguard that preserves its task link never applied.

## Flagged, not guessed

- **`isWorkspaceOwnerLive` keeps `complete` membership ONLY.** The literal was `done` alone, and its own header states every non-terminal state must read LIVE so a lease is never yanked from a task that could still be running. Adding `archived` would reclaim leases the old code kept — a behaviour change riding inside a column conversion.
- **The lifecycle skip set deliberately excludes waiting/intake.** An agent running against a card that fell back to a waiting lane *is* the drift this sweep recovers. I made that mistake on an earlier pass and the existing "recovers durable running agents linked to todo tasks" case caught it.
- **The synthetic `{ column: "todo" }`** for a *missing* task stays a legacy id, and is now correct rather than unconverted: `parkedColumns` is legacy-seeded, so `todo` is a member on every board. A deleted row has no workflow selection, so a "renamed" placeholder would be a guess about a task that no longer exists.

## Census

| | before | after |
|---|---|---|
| `self-healing.ts` | 39 | **25** |
| repo backlog | 89 | **75** |

## Measured

- `src/__tests__/self-healing*` + `task-agent*` — **42 files / 829 tests pass**
- `tsc --noEmit -p packages/engine` — clean
- **`check-inert-sync-lane-conversions` exits 0** — this branch adds none, which is the whole point of the withdrawal
- census `--strict`, `check-lane-wiring` ("none added"), `check-fnxc-future-dates` — clean

### DIFF
--- a/packages/engine/src/__tests__/self-healing-orphaned-pending-step-results.test.ts
+++ b/packages/engine/src/__tests__/self-healing-orphaned-pending-step-results.test.ts
@@ -58,7 +58,26 @@ function task(id: string, overrides: Partial<Task> = {}): Task {
   } as Task;
 }
 
-function storeFor(tasks: Task[]): TaskStore & EventEmitter {
+/*
+FNXC:WorkflowResolvedColumns 2026-07-31-17:55:
+A board whose implementation lane is `building`. Supplied through `listWorkflowDefinitions`, which is
+the ONLY store read `resolveProjectColumnsForRoles` makes — a project-wide async read that works under
+PostgreSQL, unlike the sync workflow-SELECTION reader that makes `resolveTaskWorkflowIrSync`-based
+conversions inert. Using the async helper here is what makes the conversion real rather than decorative.
+*/
+const RENAMED_WIP_IR = {
+  version: "v2",
+  id: "custom:wf",
+  nodes: [],
+  edges: [],
+  columns: [
+    { id: "drafting", name: "drafting", traits: [{ trait: "hold", config: { release: "capacity" } }] },
+    { id: "building", name: "building", traits: [{ trait: "wip", config: { limitSetting: "maxConcurrent" } }] },
+    { id: "checking", name: "checking", traits: [{ trait: "merge" }] },
+  ],
+};
+
+function storeFor(tasks: Task[], workflowIr?: unknown): TaskStore & EventEmitter {
   const tasksById = new Map(tasks.map((entry) => [entry.id, entry]));
   return Object.assign(new EventEmitter(), {
     getSettings: vi.fn(async () => ({ globalPause: false, enginePaused: false } as Settings)),
@@ -70,6 +89,8 @@ function storeFor(tasks: Task[]): TaskStore & EventEmitter {
       return all.slice(offset, offset + limit);
     }),
     getTask: vi.fn(async (id: string) => tasksById.get(id)),
+    /* Absent → the helper keeps the legacy ids, which is every other case in this file. */
+    ...(workflowIr ? { listWorkflowDefinitions: vi.fn(async () => [{ ir: workflowIr }]) } : {}),
     updateTask: vi.fn(async (id: string, patch: Partial<Task>) => {
       const next = { ...tasksById.get(id)!, ...patch } as Task;
       tasksById.set(id, next);
@@ -255,5 +276,47 @@ describe("review-gate lease liveness (in-review gates)", () => {
     expect(await manager.reconcileOrphanedPendingStepResults()).toBe(1);
     const after = await store.getTask("FN-NO-OWNER");
     expect(after?.workflowStepResults?.[0]?.status).toBe("failed");
+  });
+
+  /*
+  FNXC:WorkflowResolvedColumns 2026-07-31-17:55:
+  THE EXECUTOR-OWNED SKIP MUST FOLLOW A RENAMED IMPLEMENTATION LANE.
+
+  This sweep REWRITES `pending` step results to `failed`. It skips implementation-lane rows because
+  they are executor-owned and their liveness is unprovable here — a skip that was keyed on the
+  `in-progress` literal. On a board whose wip lane is named anything else the skip did not apply, so
+  the sweep rewrote a live executor's pending results out from under it, closing the merge gate on a
+  step that was still running.
+
+  Both legs are asserted: the page-snapshot guard and the re-read guard. They are separate `continue`s
+  and either one alone would let this pass while the other stayed literal.
+  */
+  it("skips an executor-owned row resting in a RENAMED implementation lane", async () => {
+    const live = task("FN-RENAMED-WIP", {
+      column: "building",
+      workflowStepResults: [stepResult({ status: "pending", startedAt: new Date().toISOString() })],
+    });
+    const store = storeFor([live], RENAMED_WIP_IR);
+    const manager = new SelfHealingManager(store, "/repo", {} as never);
+
+    expect(await manager.reconcileOrphanedPendingStepResults()).toBe(0);
+    expect((await store.getTask("FN-RENAMED-WIP"))?.workflowStepResults?.[0]?.status).toBe("pending");
+  });
+
+  /*
+  The paired negative. The conversion widens membership, so it must not turn every lane into a skip:
+  a card in the same board's REVIEW lane with a dead session is still an orphan and must still be
+  rewritten, or the fix trades one silent failure for another.
+  */
+  it("still rewrites an orphan in a renamed lane that is NOT the implementation lane", async () => {
+    const stranded = task("FN-RENAMED-REVIEW", {
+      column: "checking",
+      workflowStepResults: [stepResult({ status: "pending", startedAt: new Date().toISOString() })],
+    });
+    const store = storeFor([stranded], RENAMED_WIP_IR);
+    const manager = new SelfHealingManager(store, "/repo", {} as never);
+
+    expect(await manager.reconcileOrphanedPendingStepResults()).toBe(1);
+    expect((await store.getTask("FN-RENAMED-REVIEW"))?.workflowStepResults?.[0]?.status).toBe("failed");
   });
 });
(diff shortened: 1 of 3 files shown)


gsxdsm:
**#3077 merged into this branch**, so it now carries both commits: `39 → 25` and `25 → 11`. Retitled to match. Rebased onto current `main` (the census baseline is derived, so every fleet PR rewrites it and every fleet PR conflicts on it — resolved by regenerating from `main`, never by hand-merging).

### Combined census

| | before | after |
|---|---|---|
| `self-healing.ts` | 39 | **11** |
| repo backlog | 89 | **61** |

### The second commit, in one line each

Four more sweeps, all through the **async** `resolveProjectColumnsForRoles` — whose only store read is `listWorkflowDefinitions()`, a project-wide async read that works under PostgreSQL, which is exactly what makes these real where the sync-IR conversions are inert:

- `reconcileTaskWorktreeMetadata` — 3 guards
- `reconcileOrphanedPendingStepResults` — 2 guards, **both legs**. This sweep *rewrites* pending step results to `failed`; keyed on the `in-progress` literal it did that to a live executor on any renamed board, closing the merge gate on a step that was still running.
- `reconcilePreExecutionWorktrees` — a four-id literal in a sweep that removes worktrees
- `recoverStarvedRefinementTriageTasks` — 2 peer counts that read zero on a renamed board, so starvation escalation never fired

All four are **exclusions**, which is why legacy seeding is safe here: a superset skips more, and skipping is the no-op. The hazardous direction is a seeded set widening a *refusal*, and none of these is one.

### Verification, re-run after the rebase

- `src/__tests__/self-healing*` + `task-agent*` — **42 files / 835 tests pass**
- `tsc --noEmit -p packages/engine` clean
- **`check-inert-sync-lane-conversions` exits 0** — neither commit adds one, which is the point of the withdrawal described in the PR body
- census `--strict`, `check-lane-wiring`, `check-fnxc-future-dates` clean

gsxdsm:
**Four PRs are converting `self-healing.ts` concurrently** — #3074, #3075, #3078, #3080 — and three of them touch the same three hunks. I measured the overlap rather than guessing; posting it here because this is the largest cluster in the phase and the rework is avoidable.

Hunks per PR (by pre-image line):

| line | #3074 | #3075 | #3078 | #3080 |
|---|---|---|---|---|
| 949 / 973 / 1002 | ● | ● | | ● |
| 1484 | ● | | | ● |
| 1516 | | | | ● |
| 3110 | | ● | | ● |
| 5271 / 5290 | ● | | | |
| 5288 | | | ● | |
| 5327 | ● | | ● | |
| 5356, 7326, 7345, 12899, 12906 | | | ● | |
| 12045 / 12083 | | ● | | ● |
| 30, 1126, 1358, 1395, 3885 | ● | | | |

**#3075 is a strict subset of #3080.** Its hunk set `{949, 973, 1002, 3110, 12045, 12083}` is contained in #3080's, and I diffed the shared bodies: byte-identical, down to the same `FNXC:WorkflowResolvedColumns 2026-07-31-14:20` stamp. #3080 adds 1484 and 1516 on top. **#3075 can be closed as superseded by #3080** with nothing lost.

That leaves three real PRs and three collision points:

- **949 / 973 / 1002** — #3074 and #3080 (after #3075 closes). Same router.
- **1484** — #3074 and #3080.
- **5327**, with #3074 at 5290 and #3078 at 5288 immediately adjacent — #3074 and #3078.

Suggested order, chosen to minimise rebasing rather than by authorship:

1. **#3074** first — largest (39 → 11), and it owns five hunks nobody else touches.
2. **#3078** next — disjoint apart from 5327 and the 5288/5290 adjacency, so it rebases cheaply.
3. **#3080** last, rebased — its remaining unique value is 1484/1516/3110/12045/12083 once #3074 has landed.

One correctness note that applies across all four, since it is the failure this phase has already produced twice. `check:inert-sync-lanes` (#3062, hardened in #3068 and #3079) fails the build when a lane guard is wired to a resolver that reads `store.resolveTaskWorkflowIrSync` — that reader answers `undefined` for every task under PostgreSQL, so the resolver returns the **default** IR and the guard behaves exactly as the literal did while the census drops. #3051 shipped ten such guards in `scheduler.ts` before this existed (refuted live in #3058). Worth each of you confirming your conversions run through the **async** resolver; if they do, the check stays silent and there is nothing to do.

Happy to re-measure after any of these land — the extraction is scripted.

gsxdsm:
**Seven PRs are converting `self-healing.ts` concurrently.** I measured all seven — hunk sets and the inert-sync-lane check — rather than reasoning about titles. The result collapses the queue from seven to three.

### Five of them form a strict chain

Hunk sets, verified by set containment (not by reading):

```
#3075 (6 hunks)  ⊂  #3080 (8)  ⊂  #3083 (13)  ⊂  #3084 (16)  ⊂  #3086 (19)
```

Every hunk in each is present in the next. I also diffed shared bodies earlier for #3075/#3080 — byte-identical, same FNXC stamp — so this is one line of work re-proposed five times, not five different conversions.

**Only #3086 is needed. #3075, #3080, #3083 and #3084 can close as superseded with nothing lost.**

### The two independent PRs

| | hunks | overlap |
|---|---|---|
| **#3074** | 27 | ∩ #3086 = 10 (`30 949 973 1002 1484 12083 12108 12151 14157 14178`); ∩ #3078 = 3 (`5327 5356 7326`) |
| **#3078** | 7 | ∩ #3086 = 1 (`12906`); ∩ #3074 = 3 (above) |

Suggested order by rebase cost: **#3074** (largest, 17 hunks nobody else touches) → **#3078** (near-disjoint) → **#3086** rebased.

### Blocking issue on the chain's survivor

Measured with the hardened check from #3079:

| PR | inert sync-lane guards in `self-healing.ts` |
|---|---|
| #3074, #3075, #3078 | **0 — clean** |
| #3080, #3083, #3084, **#3086** | **5** |

Everything from #3080 onward carries `resolveMoveFanoutColumnsSync`, which reads `this.store.resolveTaskWorkflowIrSync?.(taskId)`. That reader returns `undefined` for every task under PostgreSQL, so the resolver hands back the **default** IR — a real IR resolving the default board's traits, not `undefined` falling through to the legacy seeds. `columnsWithFlag(ir, "countsTowardWip")` yields `in-progress`, already in the seed set, so the five guards compare against exactly the literals they replaced. Census drops by five; behaviour changes on no board. Refuted end-to-end on a live PostgreSQL store in **#3058** (merged).

Note the chain's root, **#3075, does not have this** — it predates the helper. The defect entered at #3080 and was carried forward four times.

So the recommendation is not simply "merge #3086": it is **merge #3086 after the sync helper is resolved**, either by carrying the lanes on the `task:moved` payload (removes the class) or by keeping the literals and flagging them, as `merge-queue-ops-2.ts:53` already does. #3079 would fail all four branches at 34 against a baseline of 29; please don't clear that with `--update-baseline`.

### Net

Seven PRs → three (**#3074**, **#3078**, **#3086**), with one blocking fix on #3086 shared by three of the four PRs it supersedes.

Extraction is scripted; happy to re-run after any of these land.

gsxdsm:
Superseded by the rebuilt branch. #3075 landed the pause-abort router conversion while this was open; the replacement is rebuilt off current main and additionally converts the three body guards #3075 left comparing ids in that same sweep.

PR #3075, opened 2026-07-31
fix(self-healing): route pause-abort recovery on resolved columns (self-healing 38 → 34)

gsxdsm:
First real cut into `self-healing.ts`, the last large cluster. Converts the **pause-abort recovery router** — a coherent unit with one owner, rather than a scattered pass.

## Census before/after

| Metric | Before | After |
|---|---:|---:|
| COLUMN guards (backlog) | 86 | **82** |
| `self-healing.ts` | 38 | **34** |

## The bug this was hiding

The router keyed on three literals — `in-review` twice (review progress, manual merge hold) and `todo || in-progress` (active work). On a renamed board **all three stop matching**, so a parked card in a renamed lane falls through to `no-action` and is never recovered — silently, no log line, and with every existing test still green because they all use legacy ids.

## Conversion

Used the file's own resolvers. `resolveReviewColumnsFor` already existed; added `resolveActiveWorkColumnsFor` as its sibling from the same `columnsWithFlag` / `resolveLifecycleColumns` helpers — no new vocabulary.

**ACTIVE WORK is hold + `countsTowardWip`, deliberately NOT the intake + hold that the neighbouring `resolvePreWipColumns` returns.** The router asks *"is this card mid-flight, so a requeue is right?"* — an intake lane is not mid-flight; a WIP lane is. Reusing the intake-shaped helper would have widened the requeue to triage rows and dropped in-progress ones. Because the two sets **overlap on `hold`**, that mistake looks correct in every legacy-id test. This is the trap worth knowing about for the rest of the file: it has several resolvers, and picking the nearest one is not the same as picking the right one.

**`columns` is required, not optional-with-a-fallback.** The router has exactly two callers — the candidate filter and the post-re-read re-verify — and they must agree. An optional parameter lets one resolve and the other default, and that divergence surfaces as a sweep that selects a card and then declines to act on it, writing nothing. Required makes it a compile error. (Same reasoning as #3059; safe here because both resolvers union the legacy ids internally, so "required" never means callers invent a column set.)

**On the filter/re-verify hazard I flagged earlier:** both call sites are the *same function*, so converting it once keeps them consistent by construction — no split-CAS risk. Per-task resolution can't hoist out of the loop but must not read an IR per row, so the marker test (column-independent) stays a cheap sync prefilter and the IR is read only for rows that pass it, over a shared cache. `parked` keeps its exact former membership, so the log count still means what it said.

## Verification

- **Revert-proof:** reverting the three guards to literals fails 2 of 3 routing cases. The third exercises the new resolver directly, so it cannot fail on revert — stated rather than counted as evidence.
- 4 self-healing suites, **437 tests green**
- `tsc --noEmit` clean; eslint clean
- Degraded-resolution case included, since the legacy-id union is what keeps recovery alive on an unreadable workflow

## Still flagged in this file (not guessed)

34 remain. They are not one batch: the sweeps around L2855/3297/8877/9018 depend on the unowned `listTasks({ column })` decision, and L5330/5359 sit under the FN-5256 liveness guard whose own comment says those columns can be live when the heuristic calls them stale.

### DIFF
--- a/packages/engine/src/__tests__/self-healing-prewip-role-vocabulary.test.ts
+++ b/packages/engine/src/__tests__/self-healing-prewip-role-vocabulary.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, it, vi } from "vitest";
 import "@fusion/core"; // register built-in traits
-import type { Task, TaskStore, WorkflowIr } from "@fusion/core";
+import type { Settings, Task, TaskStore, WorkflowIr } from "@fusion/core";
 import { SelfHealingManager } from "../self-healing.js";
 
 /*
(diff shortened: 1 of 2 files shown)


gsxdsm:
Measured against #3080: this PR's hunk set `{949, 973, 1002, 3110, 12045, 12083}` is a **strict subset** of #3080's, and the shared bodies are byte-identical — same text, same `FNXC:WorkflowResolvedColumns 2026-07-31-14:20` stamp. #3080 adds 1484 and 1516 on top.

Suggest closing this as superseded by #3080; nothing is lost. Full four-way overlap map and a suggested merge order are on #3074 and #3080 — four PRs are converting `self-healing.ts` at once and three of them share these same three hunks.

gsxdsm:

PR #3078, opened 2026-07-31
fix(engine): self-healing lane cluster round 2 — 38 → 26 (two sweeps could disturb live work)

gsxdsm:
The largest census cluster became **unclaimed again** when #3055 closed conflicting. I had closed my own #3050 an hour earlier expecting #3055 to land, so this re-applies the conversions #3047 and #3049 did not cover.

**Re-applied from current main rather than rebasing the closed branch.** The conversions are small; the conflict archaeology is what went wrong last time — nine conflicts against #3049, several on variable names identical to mine, and my mechanical fixup corrupted the file badly enough that I aborted. Starting from main cost less than resolving that and carries no risk of resurrecting a stale line.

## Census

| | before | after |
|---|---|---|
| `packages/engine/src/self-healing.ts` | **38** | **26** |
| repo-wide column guards | 84 | **72** |

## Three sweeps, existing role helpers only

| sweep | roles | what it did on a renamed board |
|---|---|---|
| worktree metadata | terminal + wip + review | rebound finished cards every pass, **and the FN-5256 liveness guard went silent** |
| orphaned pending step results | wip | **could rewrite `pending` results under a live executor run** |
| agent-link drift | wip + review + terminal | evaluated agents whose task was plainly still executing |

Two of these disturb **live** work, which is why they were worth redoing now rather than leaving for the next fleet round:

- The worktree-metadata sweep clears `worktree`/`branch` metadata. Its liveness guard is the thing standing between that and a running shell (FN-5256). Keyed on ids, it matched nothing on a renamed board. The scope-override safety condition beside it now reads the **same resolved sets**, so the two cannot disagree about which lanes are live — previously they were two independent literal lists.
- The orphaned-step-results sweep's own header says it must never touch an executor-owned row. The id-keyed skip made it do exactly that. Resolved once per sweep, outside the paging loop, so a large board still pays one resolve.

## Flagged, not guessed — the 26 that remain

Unchanged from my earlier audit and re-verified on this base:

- **Sync predicates** (`isWorkspaceOwnerLive`, the pause-abort classifier, the phantom-binding check, the `task:moved` listener guards). No store handle; converting means a signature change or making a synchronous event listener async, which reorders handlers against a synchronous emitter.
- **Already-converted fallbacks** — `own.length > 0 ? own.includes(...) : task.column === "in-review"`. The resolved answer wins; the literal is the documented no-metadata path.
- **The notification-route `fresh.column === "todo"` sites** — measured previously: any `await` before the wedge resolve drops an operator notification. Needs the wedge-episode contract, not a column pass.

## Verification

self-healing suites **204 passed** · agent-link-drift + query-filter-blindness **83 passed** · `pnpm test:gate` 161 + 13 + 487 + 71 · lint · census `--strict` · lane-wiring — green.

### DIFF
--- a/.changeset/self-healing-cluster-round2.md
+++ b/.changeset/self-healing-cluster-round2.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Three more self-healing repairs use your board's own column names, including one that could disturb a running task.
+category: fix
+dev: Converts the worktree-metadata reconcile, orphaned-pending-step-results, and agent-link-drift sweeps in `self-healing.ts` to `resolveProjectColumnsForRoles`.
--- a/packages/engine/src/self-healing.ts
+++ b/packages/engine/src/self-healing.ts
@@ -5288,9 +5288,19 @@ export class SelfHealingManager extends SelfHealingGitEvidence {
       }
       let repaired = 0;
 
+      /*
+      FNXC:WorkflowResolvedColumns 2026-07-31-16:40 (fleet, self-healing round 2):
+      TERMINAL/WIP/REVIEW roles for this sweep's three lane questions. Keyed on ids the terminal skip
+      never fired on a renamed board (finished cards were rebound every pass) and — the dangerous half —
+      the FN-5256 liveness guard below went silent, so the sweep could clear worktree metadata out from
+      under a running shell.
+      */
+      const worktreeReconcileTerminalColumns = await resolveProjectColumnsForRoles(this.store, TERMINAL_ROLES);
+      const worktreeReconcileWipColumns = await resolveProjectColumnsForRoles(this.store, ["countsTowardWip"]);
+      const worktreeReconcileReviewColumns = await resolveProjectColumnsForRoles(this.store, REVIEW_ROLES);
       for (const task of allTasks) {
         if (!task.worktree) continue;
-        if (!options?.includeTaskIds?.has(task.id) && (task.column === "done" || task.column === "archived")) {
+        if (!options?.includeTaskIds?.has(task.id) && worktreeReconcileTerminalColumns.has(task.column)) {
           continue;
         }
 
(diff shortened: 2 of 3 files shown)


gsxdsm:
**Measured state of the fleet phase, because the duplication is now the dominant cost.**

Five open PRs touch `self-healing.ts` right now — #3074, #3075, #3078 (this one), #3080, #3083 — plus #3047, #3049 already merged into it and #3050, #3055 closed against it. That is **nine** PRs on one file today.

Separately, #3081 and my #3067 fixed the **same single line** in `backlog-pressure-reporter.ts` independently, with the same resolved role set. I closed mine.

My own tally for the day: **three PRs closed as duplicates** (#3050, #3067) or abandoned mid-rebase, and three rebases of one file — one of which corrupted the tree badly enough to abort. That is most of a turn's budget spent on collision, not conversion.

The cause is mechanical, not anyone's carelessness: *"claim the largest unclaimed cluster"* + a census that reports per-file counts = every worker independently picks the same file, and the count only changes when someone's PR **merges**. Between claim and merge, the file looks unclaimed to everyone. Each merge then converts the duplicated work into conflicts for everyone else, so the cost grows with fleet size rather than shrinking.

Two cheap fixes, either one sufficient:
1. **Post a claim comment naming the file before converting** — I did this, and three more PRs landed in the file anyway, so it only works if everyone does it.
2. **Assign files rather than a rule** — the census output is a list; handing worker N the Nth file removes the race entirely.

I am not opening a sixth PR on `self-healing.ts`. This one is green, mergeable, rebased on current main, and covers three sweeps — two of which can disturb **live** work (the FN-5256 liveness guard going silent, and the orphaned-step-results sweep rewriting `pending` results under a live executor). Whoever merges should pick one of the five and close the rest as superseded; the conversions are semantically identical and differ mainly in variable naming.

gsxdsm:
**Second commit: this PR had no regression protection, and I only found out because I ran the check I have been demanding of everyone else.**

I shipped the conversion citing "204 self-healing tests pass." That number was meaningless for this change. Measured: reverting all three conversions, **all 204 still pass** — `reverted-rc=0`. Every existing test in that area uses `in-review` / `in-progress`, where the literal is correct, so none of them could see the defect.

That is the same "a green suite is not coverage" failure I flagged in four other PRs today, in my own work, twice: first the vacuous urgent-first assertion in #3016, now this.

Added two cases for the highest-stakes of the three sweeps:

- **an executor-owned card in a renamed wip lane is SKIPPED.** Reverted, the sweep reaches a card an executor is actively running and rewrites its `pending` step results to `failed` — the one thing that file's header says it must never do. The liveness triple does **not** cover this: those legs prove an *in-process* session, and an executor on another node or between session handles is exactly the case the column skip exists for.
- **a genuine orphan on that same renamed board is still recovered**, so the skip narrows rather than disables. That one passes either way; it is there to stop the fix from becoming "skip everything."

Measured: **10 pass** with the fix; reverted, the executor-owned case **fails**.

The other two sweeps in this PR (worktree-metadata liveness, agent-link drift) still have no dedicated case. I would rather say that plainly than imply the whole PR is covered: the orphaned-step-results one is the sweep that can corrupt a live executor's state, so it got the test first.

Gate re-run after the test: 161 + 13 + 487 + 71, lint, census `--strict`, lane-wiring — green.

PR #3080, opened 2026-07-31
fix(self-healing): route the task:moved fan-out on resolved lanes (self-healing 34 → 24)

gsxdsm:
Second cut into `self-healing.ts`. **Stacked on #3075** — review that first; this diff is only the fan-out.

## Census

| Metric | Start of session | Now |
|---|---:|---:|
| COLUMN guards (backlog) | 88 | **72** |
| `self-healing.ts` | 38 | **24** |

## The bug

The `task:moved` fan-out compared against five literals. On a renamed board the whole listener goes inert at once — board-stall counter stops counting, in-review rebind stops firing, completed-task reconciliation stops running. **None of it fails a test, because the move itself still happens.**

## The part worth reading: split by observability

My first attempt put the entire listener behind one `await`, on this stated premise:

> the counter is read only by a later periodic sweep, never synchronously after a move

**That was wrong.** `runBoardStallAutoRecoverySweep` performs its own recovery moves and *then* evaluates progress **in the same pass**. The deferred increment was read as a stale zero, so a genuine recovery was reported as unrecovered. `board-stall-auto-recovery.test.ts` caught it — 4 failures against a green base, which is how I knew it was mine and not pre-existing.

The shipped version splits by what is synchronously observed:
- **counter → sync**, via `resolveTaskWorkflowIrSync` (the seam `scheduler.ts` already uses for sync lane resolution)
- **the two reconcile branches → async**, keeping the fire-and-forget shape they already had

Worth flagging for the rest of this file: I had previously listed this listener as *blocked* on "needs async conversion first." That was half right — the blocker was real but narrower than stated, and it applies to the counter alone, not the listener.

## Verification

- **Revert-proof:** reverting the guards to literals fails 2 cases. The others exercise the new resolvers directly, so they cannot fail on revert — stated, not counted as evidence.
- 5 suites, **449 tests green**; tsc and eslint clean
- Degraded-resolution case included: an unresolvable workflow must keep the fan-out firing rather than flat-line board-stall detection

24 remain. The sweeps at L2894/3297/8948/9089 depend on the unowned `listTasks({ column })` decision; L5430/5364 sit under the FN-5256 liveness guard whose own comment says those columns can be live when the heuristic calls them stale.

### DIFF
--- a/packages/engine/src/__tests__/self-healing-prewip-role-vocabulary.test.ts
+++ b/packages/engine/src/__tests__/self-healing-prewip-role-vocabulary.test.ts
@@ -254,3 +254,75 @@ describe("self-healing pause-abort router column vocabulary", () => {
       .toBe("node-requeue");
   });
 });
+
+/*
+FNXC:WorkflowResolvedColumns 2026-07-31-15:20 (fleet — task:moved fan-out vocabulary):
+The fan-out routed on five literals. On a renamed board all of it went inert at once — the board-stall
+counter stopped counting, in-review rebind stopped firing, completed-task reconciliation stopped running
+— and none of that fails a test, because the move itself still happens.
+
+The sync/async split below is asserted deliberately. The counter MUST increment synchronously:
+`runBoardStallAutoRecoverySweep` makes its own recovery moves and then reads the counter in the same
+pass, so an increment deferred behind an await is read as a stale zero and a real recovery is reported
+as unrecovered. That is not hypothetical — it is what board-stall-auto-recovery.test.ts caught when this
+conversion first put the whole listener behind one await.
+*/
+describe("self-healing task:moved fan-out column vocabulary", () => {
+  type FanoutInternals = {
+    resolveMoveFanoutColumnsSync(taskId: string): { wip: ReadonlySet<string>; review: ReadonlySet<string>; complete: ReadonlySet<string>; archived: ReadonlySet<string>; hold: ReadonlySet<string> };
+    handleTaskMovedFanout(task: Task, from: string, to: string): Promise<void>;
+    reconcileInReviewBranchRebind: unknown;
+    reconcileCompletedTask: unknown;
+  };
+
+  function syncStoreFor(ir?: WorkflowIr): TaskStore {
+    return {
+      ...storeFor(ir),
+      resolveTaskWorkflowIrSync: vi.fn(() => ir),
+    } as unknown as TaskStore;
+  }
+
+  it("resolves every fan-out lane from a renamed board", () => {
+    const manager = managerFor(syncStoreFor(RENAMED_WITH_REVIEW)) as unknown as FanoutInternals;
+    const c = manager.resolveMoveFanoutColumnsSync("FN-9200");
+    expect(c.wip.has("building")).toBe(true);
+    expect(c.review.has("signoff")).toBe(true);
+    expect(c.complete.has("shipped")).toBe(true);
+    expect(c.hold.has("backlog")).toBe(true);
+  });
+
+  it("keeps legacy ids when the store cannot resolve synchronously", () => {
+    const manager = managerFor(storeFor(undefined)) as unknown as FanoutInternals;
+    const c = manager.resolveMoveFanoutColumnsSync("FN-9200");
+    expect(c.wip.has("in-progress")).toBe(true);
+    expect(c.review.has("in-review")).toBe(true);
+    expect(c.complete.has("done")).toBe(true);
+  });
+
+  it("fires the in-review rebind for a move into a RENAMED review lane", async () => {
+    const manager = managerFor(syncStoreFor(RENAMED_WITH_REVIEW)) as unknown as FanoutInternals;
+    const rebind = vi.fn(async () => undefined);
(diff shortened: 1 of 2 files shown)


gsxdsm:
**Four PRs are converting `self-healing.ts` concurrently** — #3074, #3075, #3078, #3080 — and three of them touch the same three hunks. I measured the overlap rather than guessing; posting it here because this is the largest cluster in the phase and the rework is avoidable.

Hunks per PR (by pre-image line):

| line | #3074 | #3075 | #3078 | #3080 |
|---|---|---|---|---|
| 949 / 973 / 1002 | ● | ● | | ● |
| 1484 | ● | | | ● |
| 1516 | | | | ● |
| 3110 | | ● | | ● |
| 5271 / 5290 | ● | | | |
| 5288 | | | ● | |
| 5327 | ● | | ● | |
| 5356, 7326, 7345, 12899, 12906 | | | ● | |
| 12045 / 12083 | | ● | | ● |
| 30, 1126, 1358, 1395, 3885 | ● | | | |

**#3075 is a strict subset of #3080.** Its hunk set `{949, 973, 1002, 3110, 12045, 12083}` is contained in #3080's, and I diffed the shared bodies: byte-identical, down to the same `FNXC:WorkflowResolvedColumns 2026-07-31-14:20` stamp. #3080 adds 1484 and 1516 on top. **#3075 can be closed as superseded by #3080** with nothing lost.

That leaves three real PRs and three collision points:

- **949 / 973 / 1002** — #3074 and #3080 (after #3075 closes). Same router.
- **1484** — #3074 and #3080.
- **5327**, with #3074 at 5290 and #3078 at 5288 immediately adjacent — #3074 and #3078.

Suggested order, chosen to minimise rebasing rather than by authorship:

1. **#3074** first — largest (39 → 11), and it owns five hunks nobody else touches.
2. **#3078** next — disjoint apart from 5327 and the 5288/5290 adjacency, so it rebases cheaply.
3. **#3080** last, rebased — its remaining unique value is 1484/1516/3110/12045/12083 once #3074 has landed.

One correctness note that applies across all four, since it is the failure this phase has already produced twice. `check:inert-sync-lanes` (#3062, hardened in #3068 and #3079) fails the build when a lane guard is wired to a resolver that reads `store.resolveTaskWorkflowIrSync` — that reader answers `undefined` for every task under PostgreSQL, so the resolver returns the **default** IR and the guard behaves exactly as the literal did while the census drops. #3051 shipped ten such guards in `scheduler.ts` before this existed (refuted live in #3058). Worth each of you confirming your conversions run through the **async** resolver; if they do, the check stays silent and there is nothing to do.

Happy to re-measure after any of these land — the extraction is scripted.

gsxdsm:
**This conversion is inert under PostgreSQL.** Flagging pre-merge because it is the third occurrence of this exact shape in the phase and the first two both landed.

```ts
private resolveMoveFanoutColumnsSync(taskId: string) {
  ...
  const ir = this.store.resolveTaskWorkflowIrSync?.(taskId);
```

`resolveTaskWorkflowIrSync` resolves through the sync workflow **selection** reader, which returns `undefined` for every task in PostgreSQL mode — the shipped backend. The resolver then takes its `!workflowId` branch and hands back the **default builtin IR**.

Note the shape precisely, because the obvious reading is wrong: the helper does **not** get `undefined` and fall through to its legacy seeds. It gets a real IR that resolves real traits — the default board's — so `columnsWithFlag(ir, "countsTowardWip")` yields `in-progress`, which is already in the seed set. The five guards at 1620–1621 therefore compare against exactly the literal ids they replaced, on every board.

I proved this end-to-end on a live PostgreSQL store in **#3058** (merged): for a card bound to a stored renamed workflow sitting in its own review column, the sync path answers `in-review` / `in-progress` / `done` while the async resolver correctly answers `checking` / `building` / `shipped`.

**Your synchronous constraint is real** — I am not disputing it. `taskMovedFanoutListener` is a sync listener and `board-stall-auto-recovery.test.ts` depends on that, exactly as your comment says. The problem is that the available sync resolver cannot see the board, so the conversion buys nothing while the census drops by five. That is worse than the literal, because the literal was counted and this is not. Precedent: **#3051** did this to ten `scheduler.ts` arms and the note directly above that handler had already warned against it; `executor.ts:10459` states the finding in capitals, dated a day earlier.

Two options, both honest:

1. **Carry the resolved lanes on the `task:moved` payload**, so no listener resolves at all. This is what the surviving note above the scheduler's handler recommends and it removes the class rather than one instance.
2. **Keep the literals and flag them**, as `merge-queue-ops-2.ts:53` and `notification-service.ts` already do — the census keeps pointing at the site with the reason attached, which is accurate debt rather than a false green.

Mechanically: **#3079** (open) makes this check see `self-healing.ts`, and your branch measures **34 against a baseline of 29** — it would fail the gate. Whichever merges second will hit it, which is the intended behaviour rather than an obstacle to work around; please do not clear it with `--update-baseline`.

Same helper appears in **#3080 and #3083** — worth fixing in one place given five PRs are converting this file concurrently (overlap map and suggested merge order on #3074).

gsxdsm:

gsxdsm:

```
```

gsxdsm:

PR #3081, opened 2026-07-31
fix(engine): backlog pressure called every dependency unfinished on a renamed board

gsxdsm:
## The third lane question

This reporter had **three** lane questions. Two were resolved when the file's query-blindness was fixed — hold and wip, both through `resolveProjectColumnsForRoles`. The third sat one method down and was never touched:

```ts
if (dependency.column !== "done") return false;
```

One board, two lane answers.

## What it cost

On a renamed board every dependency reads unfinished, so `isRunnableCandidate` rejects every card that has one. The backlog-pressure alert then names **only dependency-free cards** as the runnable ones.

The failure mode is the quiet kind: the report still renders, the counts are right, and the candidate list looks plausible. The operator is told the queue is blocked on nothing in particular. No default-board test can see it — which is exactly why the earlier conversion of this same file, which fixed its reads, left this behind.

## Fix

`finishedColumns` (complete ∪ archived) resolved once by the async caller alongside hold and wip, then passed into the sync predicate.

- **Required parameter, not optional-with-a-literal-default.** An optional parameter leaves `done` in the file as a silent fallback and the next caller gets pre-conversion behaviour by writing nothing.
- **Archived is included** because a dependency that has been archived is finished too — and this reporter already reads with `includeArchived: true` precisely so archived blockers resolve.
- **Async resolution.** `resolveProjectColumnsForRoles`' only store read is `listWorkflowDefinitions()`, a project-wide async read that works under PostgreSQL. That is the line between a real conversion and the inert sync-IR kind (#3058), and the new test supplies its board through that same reader so it exercises the production path.

## Census

| | before | after |
|---|---|---|
| `backlog-pressure-reporter.ts` | 1 | **0** |

## Measured

- One new case; file **11/11 pass**.
- **MUTATION**: restoring `dependency.column !== "done"` fails it.
- The case asserts **both directions in one test** — a dependency resting in the board's own complete lane makes its card runnable, *and* a dependency still in the hold lane still blocks it. Asserting only the first would pass against a predicate that had simply stopped checking dependencies.
- The file already had a `RENAMED_IR` scoped to its second describe; mine is a distinct `RENAMED_DEPENDENCY_IR` with different lane names. I hit the shadowing first and the test failed as `under-threshold` — worth noting because a same-named fixture that silently resolves to the *other* board is precisely how a renamed-lane test goes vacuous.
- `tsc --noEmit -p packages/engine` clean; census `--strict`, `check-lane-wiring`, `check-inert-sync-lane-conversions`, `check-fnxc-future-dates` clean.

## Flagged, not guessed

Adjacent census entries I looked at and deliberately left:

- **`executor.ts` (4)** — all inside a sync `task:moved` listener. Converting via `resolveTaskWorkflowIrSync` would be inert for #3058's reason, and making the listener async reorders it against every other subscriber. Correctly out of scope, as #3048 judged.
- **`triage.ts:724`** — half-converted in the same shape: `disposeLanes.hold`/`.intake` come from a sync resolver, so the resolved arms are themselves inert and "finishing" the guard would add a third inert comparison.
- **`auto-merge-finalization.ts` (2)** — one is the resolver's documented degraded fallback (the live arm calls `columnHasFlag`), the other is already recorded as a deferred signature-widening whose cost exceeds the error string it sharpens.
- **`in-review-stall.ts:196`** — an explicitly marked DELIBERATE-LITERAL no-metadata fallback.

### DIFF
--- a/packages/engine/src/__tests__/backlog-pressure-reporter.test.ts
+++ b/packages/engine/src/__tests__/backlog-pressure-reporter.test.ts
@@ -30,6 +30,24 @@ function createTask(overrides: Partial<Task> = {}): Task {
   } as Task;
 }
 
+/*
+FNXC:WorkflowResolvedColumns 2026-07-31-19:20:
+A board whose hold lane is `drafting`, wip is `building` and complete is `shipped`. Supplied through
+`listWorkflowDefinitions`, which is the ONLY store read `resolveProjectColumnsForRoles` makes — a
+project-wide async read that works under PostgreSQL, unlike the sync workflow-SELECTION reader.
+*/
+const RENAMED_DEPENDENCY_IR = {
+  version: "v2",
+  id: "custom:deps",
+  nodes: [],
+  edges: [],
+  columns: [
+    { id: "drafting", name: "drafting", traits: [{ trait: "hold", config: { release: "capacity" } }] },
+    { id: "brewing", name: "brewing", traits: [{ trait: "wip", config: { limitSetting: "maxConcurrent" } }] },
+    { id: "shipped", name: "shipped", traits: [{ trait: "complete" }] },
+  ],
+};
+
 function createStore(params: {
   settings?: Record<string, unknown>;
   todoSlim?: Task[];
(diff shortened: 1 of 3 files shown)


gsxdsm:
We fixed the same line independently — I'm closing #3067 in favour of this one (yours is the more thorough treatment of it: +106/-8 vs my +67/-5, and `["complete","archived"]` is the same set my `TERMINAL_ROLES` resolves to).

Carrying over the part of #3067 that is **not** duplicated, so it is not lost with the PR. I audited **all 16 unclaimed census files (17 guards)** before landing on this one. It was the only genuine defect. The other 15 should NOT be converted, and each has a reason worth reading before someone is sent at them:

- **Sentinel comparisons** (`async-comments-attachments.ts`, 4 sites) — they compare against a sentinel string the resolver itself returns, deliberately collapsing archived and soft-deleted into one value. The code says converting *"would keep passing on the built-in board and start FAILING on a renamed one — a soft-deleted task would read as not-archived."* Converting introduces the bug.
- **Sync-SQLite paths** (`merge-queue-ops-2`, `mission-store`, `audit-ops`, `lifecycle-ops`, `task-id-integrity`) — inside synchronous transactions where the only sync resolver returns `undefined` unconditionally in PostgreSQL mode. One carries: *"Left as a literal WITH this note so the next worker does not 'fix' it into a false green."*
- **Already-converted seeds** (`executor.ts` ×4, `auto-merge-finalization.ts`) — literal sets unioned with `columnsWithFlag` on the next lines; the documented degradation path when `resolveWorkflowIrForTask` falls back to the built-in IR. Worth knowing: **the census never counted those Set literals at all** — `findComparisons` only visits binary expressions. `executor.ts`'s 4 are the `from`/`to` guards in the synchronous `task:moved` listener, where converting means making the listener async and reordering handlers against a synchronous emitter.
- **`taskRevert.ts`** — supplying the modal task's flags for a *neighbour* row is the flags-for-the-wrong-row hazard; its author concluded the entry is *"accurate debt rather than a missed conversion."*
- **`notification-service.ts`** (5, the largest unclaimed) — blocked on the wedge-episode contract: any `await` before the resolve drops an operator notification, with a test that fails loudly.

PR #3083, opened 2026-07-31
fix(self-healing): resolve the pause-abort requeue TARGET, not just its guards (self-healing 34 → 21)

gsxdsm:
Third cut into `self-healing.ts`. **Stacked on #3075 → #3080** — merge those first.

## Census

|---|---:|---:|
| COLUMN guards (backlog) | 88 | **69** |
| `self-healing.ts` | 38 | **21** |

## What this finishes

#3075 converted the pause-abort router's **guards** but left the **mutation half** comparing against and writing `"todo"`. Same unit, so it belongs to the same owner.

## The one that matters: the requeue target

A hardcoded target is a **worse** failure than a stale guard. A stale guard stops matching. A move to a column the board does not declare is a **write into nowhere**.

Now resolved through `resolveReboundTargetForTask` — the KTD-10 ordering (`hold` → `intake` → first) already defined in core and already used at **L745 of this same file** for the same kind of repair. It falls back to `"todo"` internally, so the unresolvable case keeps today's behaviour.

Also converted, because they have to agree with the target or the recovery either skips a needed move or issues a redundant one:
- the two `fresh.column === "todo"` guards that decide whether a move is needed
- the in-review log-message selection
- the summary log line, which said *"requeued to todo"* — on a renamed board that described a move that never happened

## Verification

- **Revert-proof, and asserted on the write:** tests check `moveTask`'s argument, not just the guards — a guards-only test would pass with the target still hardcoded. Reverting the target alone gives `expected 'todo' to be 'backlog'`.
- Degraded case asserted separately: unresolvable workflow still requeues to `todo`.
- 5 suites, **451 tests green**; tsc and eslint clean

21 remain, and they are the two groups I flagged earlier — the `listTasks({ column })` sweeps (L2894/3297/8948/9089) and the FN-5256 liveness guards (L5364/5430). Both need a decision, not another pass.

### DIFF
--- a/packages/engine/src/__tests__/self-healing-prewip-role-vocabulary.test.ts
+++ b/packages/engine/src/__tests__/self-healing-prewip-role-vocabulary.test.ts
@@ -326,3 +326,46 @@ describe("self-healing task:moved fan-out column vocabulary", () => {
     expect(reconcile).not.toHaveBeenCalled();
   });
 });
+
+/*
+FNXC:WorkflowResolvedColumns 2026-07-31-16:05 (fleet — pause-abort requeue TARGET):
+The recovery requeued to a hardcoded "todo". A hardcoded target is a worse failure than a stale guard:
+a stale guard stops matching, but a move to a column the board does not declare is a write into nowhere.
+Asserted on `moveTask`, because the target is the whole point of this conversion — a test that only
+checked the guards would pass with the write still hardcoded.
+*/
+describe("self-healing pause-abort requeue target vocabulary", () => {
+  function recoveryStoreFor(ir: WorkflowIr | undefined, parkedTask: Task) {
+    const moveTask = vi.fn(async () => undefined);
+    const store = {
+      ...storeFor(ir),
+      getSettings: vi.fn(async () => ({ globalPause: false, enginePaused: false, autoMerge: true })),
+      listTasks: vi.fn(async () => [parkedTask]),
+      getTask: vi.fn(async () => parkedTask),
+      updateTask: vi.fn(async () => ({})),
+      moveTask,
+      logEntry: vi.fn(async () => undefined),
+      recordRunAuditEvent: vi.fn(async () => undefined),
+      getRootDir: vi.fn(() => "/tmp/test-project"),
+    } as unknown as TaskStore;
+    return { store, moveTask };
+  }
+
+  it("requeues to the board's own HOLD lane, not a hardcoded todo", async () => {
+    const parkedInWip = parked("building");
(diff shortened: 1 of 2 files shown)


gsxdsm:
**This conversion is inert under PostgreSQL.** Flagging pre-merge because it is the third occurrence of this exact shape in the phase and the first two both landed.

```ts
private resolveMoveFanoutColumnsSync(taskId: string) {
  ...
  const ir = this.store.resolveTaskWorkflowIrSync?.(taskId);
```

`resolveTaskWorkflowIrSync` resolves through the sync workflow **selection** reader, which returns `undefined` for every task in PostgreSQL mode — the shipped backend. The resolver then takes its `!workflowId` branch and hands back the **default builtin IR**.

Note the shape precisely, because the obvious reading is wrong: the helper does **not** get `undefined` and fall through to its legacy seeds. It gets a real IR that resolves real traits — the default board's — so `columnsWithFlag(ir, "countsTowardWip")` yields `in-progress`, which is already in the seed set. The five guards at 1620–1621 therefore compare against exactly the literal ids they replaced, on every board.

I proved this end-to-end on a live PostgreSQL store in **#3058** (merged): for a card bound to a stored renamed workflow sitting in its own review column, the sync path answers `in-review` / `in-progress` / `done` while the async resolver correctly answers `checking` / `building` / `shipped`.

**Your synchronous constraint is real** — I am not disputing it. `taskMovedFanoutListener` is a sync listener and `board-stall-auto-recovery.test.ts` depends on that, exactly as your comment says. The problem is that the available sync resolver cannot see the board, so the conversion buys nothing while the census drops by five. That is worse than the literal, because the literal was counted and this is not. Precedent: **#3051** did this to ten `scheduler.ts` arms and the note directly above that handler had already warned against it; `executor.ts:10459` states the finding in capitals, dated a day earlier.

Two options, both honest:

1. **Carry the resolved lanes on the `task:moved` payload**, so no listener resolves at all. This is what the surviving note above the scheduler's handler recommends and it removes the class rather than one instance.
2. **Keep the literals and flag them**, as `merge-queue-ops-2.ts:53` and `notification-service.ts` already do — the census keeps pointing at the site with the reason attached, which is accurate debt rather than a false green.

Mechanically: **#3079** (open) makes this check see `self-healing.ts`, and your branch measures **34 against a baseline of 29** — it would fail the gate. Whichever merges second will hit it, which is the intended behaviour rather than an obstacle to work around; please do not clear it with `--update-baseline`.

Same helper appears in **#3080 and #3083** — worth fixing in one place given five PRs are converting this file concurrently (overlap map and suggested merge order on #3074).

gsxdsm:

gsxdsm:

```
```

gsxdsm:

PR #3084, opened 2026-07-31
fix(self-healing): resolve starved-refinement peer progress (self-healing 21 → 19)

gsxdsm:
Fourth cut into `self-healing.ts`. **Stacked on #3075 → #3080 → #3083.**

## Census

|---|---:|---:|
| COLUMN guards (backlog) | 88 | **67** |
| `self-healing.ts` | 38 | **19** |

## The bug

"Peer progress" was `peer.column === "todo"` in **two** places — the candidate filter, and the count written into the log line and audit metadata. On a renamed board both stopped matching, so peer progress read as **zero** and starved refinements were never escalated. Silently: a zero count is indistinguishable from a genuinely quiet board.

## De-duplication was part of the fix

Both sites now read one precomputed `holdPeerIds` set. Two copies of a predicate that must agree is exactly how a converted guard ends up feeding an unconverted one — the mixed-vocabulary failure the census warns about — so collapsing them belongs in this change rather than a tidy-up later.

Resolved **once for the whole board, not per peer.** This is an N-squared scan (every candidate counts over every task), so a per-peer resolve would multiply IR reads by the candidate count. It reuses `filterByPreWipRole` — already the converted seam for the intake filter directly above — and shares its cache, so the cost is one read per distinct workflow.

## Verification

- **Revert-proof:** reverting both sites to the literal drops escalations **1 → 0**.
- The new fixture separates intake from hold deliberately — candidate in `inbox`, peers in `backlog` — so reaching for the wrong role set resolves no peers and escalates nothing. A fixture where the two lanes coincide would pass either way.
- 5 suites, **454 tests green**; tsc and eslint clean

19 remain, still the two groups: the `listTasks({ column })` sweeps (L2894/3297/8948/9089) and the FN-5256 liveness guards (L5364/5430).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved self-healing recovery across workflows with renamed or customized board lanes.
  * Recovered tasks now move to the appropriate workflow hold or review lane instead of relying on fixed defaults.
  * Improved pause-abort routing, task-move handling, stall detection, and starved refinement recovery.
  * Added fallback behavior for legacy workflow configurations and more reliable recovery reconciliation.

### DIFF
--- a/packages/engine/src/__tests__/self-healing-prewip-role-vocabulary.test.ts
+++ b/packages/engine/src/__tests__/self-healing-prewip-role-vocabulary.test.ts
@@ -326,3 +326,46 @@ describe("self-healing task:moved fan-out column vocabulary", () => {
     expect(reconcile).not.toHaveBeenCalled();
   });
 });
+
+/*
+FNXC:WorkflowResolvedColumns 2026-07-31-16:05 (fleet — pause-abort requeue TARGET):
+The recovery requeued to a hardcoded "todo". A hardcoded target is a worse failure than a stale guard:
+a stale guard stops matching, but a move to a column the board does not declare is a write into nowhere.
+Asserted on `moveTask`, because the target is the whole point of this conversion — a test that only
+checked the guards would pass with the write still hardcoded.
+*/
+describe("self-healing pause-abort requeue target vocabulary", () => {
+  function recoveryStoreFor(ir: WorkflowIr | undefined, parkedTask: Task) {
+    const moveTask = vi.fn(async () => undefined);
+    const store = {
+      ...storeFor(ir),
+      getSettings: vi.fn(async () => ({ globalPause: false, enginePaused: false, autoMerge: true })),
+      listTasks: vi.fn(async () => [parkedTask]),
+      getTask: vi.fn(async () => parkedTask),
+      updateTask: vi.fn(async () => ({})),
+      moveTask,
+      logEntry: vi.fn(async () => undefined),
+      recordRunAuditEvent: vi.fn(async () => undefined),
+      getRootDir: vi.fn(() => "/tmp/test-project"),
+    } as unknown as TaskStore;
+    return { store, moveTask };
+  }
+
+  it("requeues to the board's own HOLD lane, not a hardcoded todo", async () => {
+    const parkedInWip = parked("building");
+    const { store, moveTask } = recoveryStoreFor(RENAMED_WITH_REVIEW, parkedInWip);
+    const manager = managerFor(store) as unknown as { recoverPausedAbortFailures(): Promise<number> };
+    await manager.recoverPausedAbortFailures();
+    expect(moveTask).toHaveBeenCalledTimes(1);
+    expect(vi.mocked(moveTask).mock.calls[0]?.[1]).toBe("backlog");
+  });
+
+  it("still requeues to todo when the workflow will not resolve", async () => {
+    const parkedInWip = parked("in-progress");
+    const { store, moveTask } = recoveryStoreFor(undefined, parkedInWip);
+    const manager = managerFor(store) as unknown as { recoverPausedAbortFailures(): Promise<number> };
+    await manager.recoverPausedAbortFailures();
+    expect(moveTask).toHaveBeenCalledTimes(1);
+    expect(vi.mocked(moveTask).mock.calls[0]?.[1]).toBe("todo");
+  });
+});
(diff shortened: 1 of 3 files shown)


gsxdsm:
Same finding as on #3080 and #3083 — all three carry `resolveMoveFanoutColumnsSync`, and measured with the hardened check from #3079 this branch reads **34 against a baseline of 29**, i.e. five new inert guards.

Raising it here specifically because of the justification in the note:

> `resolveTaskWorkflowIrSync` is the same seam `scheduler.ts` uses for its own sync lane resolution.

That is accurate, and it is the problem. `scheduler.ts` uses that seam because **#3051 put it there, and #3051's conversion is inert** — I refuted it end-to-end on a live PostgreSQL store in **#3058** (merged): for a card bound to a stored renamed workflow, the sync path answers `in-review` / `in-progress` / `done` while the async resolver answers `checking` / `building` / `shipped`. Ten guards left the census there for zero behaviour change.

So the precedent being cited is the defect propagating. That is worth naming plainly: `executor.ts:10459` has stated this in capitals since 2026-07-30 —

> **THE SYNCHRONOUS RESOLVER IS A NO-OP IN PRODUCTION.** … the census counts it as converted … and the guard behaves exactly as the literal did. **Worse than leaving the literal, because the number says the site is done.**

— #3051 shipped the day after, and it is now the cited pattern for three more PRs.

The specific mechanism here: the selection reader returns `undefined` for every task under PostgreSQL, so the resolver returns the **default** IR — not `undefined` falling through to your legacy seeds, but a real IR resolving the default board's traits. `columnsWithFlag(ir, "countsTowardWip")` yields `in-progress`, which your seed set already contains. The five guards compare against exactly the literals they replaced.

On the reasoning itself: the degraded-answer argument holds for a store that genuinely cannot resolve, and the board-stall counter should indeed keep counting rather than flat-line. But under PostgreSQL that degraded path is the **only** path, so the counter counts the default board's lanes for every project — which is the pre-conversion behaviour, correctly described by the literal.

Your synchronous constraint is real and I am not disputing it. Two honest options, same as I gave on #3080:

1. **Carry the resolved lanes on the `task:moved` payload** — no listener resolves at all. Removes the class rather than one instance, and it is what the surviving note above the scheduler's handler recommends.
2. **Keep the literals and flag them**, as `merge-queue-ops-2.ts:53` and `notification-service.ts` do. Accurate debt beats a false green.

**#3079** is open and would fail this branch. Please don't clear it with `--update-baseline` — that is the same false green one layer up.

Six PRs are now converting this file at once (#3074, #3075, #3078, #3080, #3083, #3084); three of them share this helper, so it is worth fixing once. Overlap map and suggested merge order are on #3074, and #3075 is a proven strict subset of #3080.

gsxdsm:

gsxdsm:

```
```

gsxdsm:

gsxdsm:
**All five findings are real. I cannot fix them here** (this touches `self-healing.ts` and I do not have the budget to run the engine suite behind a resolution), but the `Major` one has a root cause worth naming before anyone patches the call site.

## `hold` by membership, not first-per-role — this is the third occurrence

`resolveLifecycleColumns(ir).hold` returns a **single id**: the first column carrying the trait. The method's own comment says the set is MEMBERSHIP because a board may declare more than one pre-review working lane. Those two facts cannot both hold, and the reviewer is right about which one gives.

The same misreading has now landed in three independent places:

| where | symptom |
|---|---|
| this PR, `hold` guard | a card in the second hold lane classifies as not-held |
| `notification-service.ts` (#3096) | `?.hold / ?.wip / ?.complete / ?.archived` — a progressed card never clears its wedge episode |
| #3088 review-set mismatch | same shape, different roles |

**Three occurrences of one misreading is a property of the signature, not three unlucky authors.** I opened **#3111** to document the contract at the definition and point at the membership alternatives:

- `columnsWithFlag(ir, role)` — every column with the role on one board
- `resolveProjectColumnsForRoles(store, roles)` — the union across a project

For this call site the fix is `columnsWithFlag(ir, "hold")`, which matches what `countsTowardWip` already does two lines away — the asymmetry the reviewer spotted.

## The four Minor findings are all legitimate

- **test name claims two transitions, asserts one** — the `complete -> archived` branch of `shouldReconcile` is genuinely unexercised, so a regression there stays green. That is the vacuous-test shape, not a naming nit.
- **`vi.useRealTimers()` after assertions** — a throw leaves fake timers active for every later test in the file. `try/finally` or `afterEach`.
- **`toHaveBeenCalledWith` does not reject duplicates** — add `toHaveBeenCalledTimes(1)` if exactly-once is the claim.

None are blocking on their own; the `hold` one is.

## Suggested order

Fix `hold` with `columnsWithFlag`, add a case that puts a card in a **second** hold lane (the current fixtures cannot fail), then the test-hygiene items. Verify with `pnpm --filter @fusion/engine exec vitest run src/__tests__/self-healing`.

PR #3085, opened 2026-07-31
fix(core): a renamed archive lane never reached cold storage; two more literals audited as dead sync

gsxdsm:
Claimed the remaining `archived` literals in `packages/core` — the largest unclaimed cluster left once `self-healing.ts` (mine, #3074) and the sync-listener files are set aside. One is a real bug; two are dead code, and saying which is which is the point.

## Converted — the cold-storage decision

Archived rows do not live in `tasks`. `archiveTask` copies them into the archive store and removes them, so `listTasksImpl` has to decide whether to read that second store:

```ts
const includeColdStorage = includeArchived && (!columnFilter || columnFilter === "archived");
```

A caller naming a **renamed** archive lane — `listTasks({ column: "filed", includeArchived: true })`, which is exactly what an archive view does — skipped cold storage and got an empty page. The rows were unreachable through the only API that can reach them.

Note the shape, because it explains why nobody caught it: the **unfiltered** read (`!columnFilter`) was always correct. This fails *only* for the caller that names the lane, so it survives any board-level smoke test, and the symptom presents as "the archive is empty" rather than as a bug.

Resolved through the async `resolveProjectColumnsForRoles`, which seeds the legacy id — so `"archived"` still opts in on every board. This widens an **inclusion**: a superset consults cold storage on a read that previously skipped it, costing one extra query and returning rows that are filtered as before. That is the safe direction; a seeded **refusal** is the hazardous one and this is not that.

## Audited, not converted — two dead sync paths

Both would be real defects if they ran. Neither runs.

| site | why it is dead |
|---|---|
| `mission-store.ts` (feature-delete link check) | `getMissionStoreImpl` returns the AsyncDataLayer-backed `AsyncMissionStore` under PostgreSQL; the sync `MissionStore` reached via `this.db.prepare` is legacy SQLite only |
| `lifecycle-ops.ts` (polling-replica archive emit) | `checkForChangesImpl` opens with `store.db.getLastModified()` / `store.db.prepare`, which throw in backend mode |

The second is worth naming precisely: **both** its guard and the `to: "archived"` it emits are literals, so a polling replica on a renamed board would emit a move to a column the board does not declare.

Recorded in place rather than converted — same treatment `project-store-ops.ts`'s dequeue twin already has — so the census entries are not mistaken for unconverted debt, and so whoever finally deletes the sync SQLite residue takes these with it. Threading a lane set into a function whose first statement cannot execute is not work this program should buy.

## Census

| | before | after |
|---|---|---|
| `reads.ts` (cold-storage entry) | 1 | **0** |
| repo backlog | 83 | **82** |

The backlog moves by one, not three, and that is deliberate: an audited dead-code literal stays counted. Marking it DELIBERATE-LITERAL would buy a nicer number by asserting the code is *correct*, which it is not — it is unreachable, which is a different claim with a different expiry.

## Measured

- New `cold-storage-renamed-archive-lane.test.ts` — **3 pass**.
- **MUTATION**: restoring the literal fails **only** the renamed case; the legacy control and the non-archive negative stay green.
- **A trap worth recording**: my first version left the LIVE read real against a fake `layer.db`. It threw, the `.catch` swallowed it, and all three cases passed the negative — *including the legacy control*, which is what exposed it. A test whose subject is never reached looks identical to one whose subject answered no. `readLiveTaskRows` is mocked now, and the control case is what makes that detectable at all.
- `src/__tests__/{cold-storage,archive,unarchive,reads}*` — **7 files / 23 tests pass**.
- `tsc --noEmit -p packages/core` clean; census `--strict`, `check-lane-wiring`, `check-inert-sync-lane-conversions`, `check-fnxc-future-dates` clean.

## Flagged, not guessed

- **`eval-signal-collector.ts:113`** — `task.column === "archived" ? "archived" : "done"` normalises an eval record's column to a two-value label. A renamed archive lane is mislabelled `done`. Real but small, and the fix means widening `collectDeterministicSignals`' signature through `HybridEvaluatorService`, whose store dep is optional. Wants its own change rather than a ride-along.
- **`reads.ts:104`**, **`async-comments-attachments.ts:158`**, **`github-tracking-state.ts:245-246`**, **`register-task-workflow-routes.ts:1987/5004`** — all already converted; the counted lines are their documented no-metadata fallback arms.


gsxdsm:

Only `scripts/lib/lifecycle-column-census-baseline.json` conflicted; your four source files applied cleanly. Resolved by taking main's copy and re-running `--strict --update-baseline`, not by hand-merging — that file is generated, and a hand-merge produces a count matching neither branch while looking plausible.

The re-record was **required**: `--strict` failed before it and passed after, because the ratchet tracks the deliberate count alongside the guard count so a marker cannot shrink the backlog quietly. Amended into your commit.

Verified: `cold-storage-renamed-archive-lane.test.ts` **3 pass**, `tsc --noEmit` (core) 0 errors, `census --strict` exit 0.

This baseline conflicts on nearly every open fleet PR, so each rebase invalidates the others — that is why the same PRs get retried. The fix is always mechanical: take main's copy, re-run the generator, never merge the text.

gsxdsm:
Superseded by the rebuilt branch. #3057 landed the same cold-storage conversion while this was open, so the conversion here is now a duplicate; the replacement keeps only what main lacks — the test coverage for that landed-untested conversion, plus the two dead-sync audits.

PR #3086, opened 2026-07-31
fix(self-healing): resolve the two active/terminal membership guards (self-healing 19 → 11)

gsxdsm:
Fifth cut into `self-healing.ts`. **Stacked on #3075 → #3080 → #3083 → #3084.**

## Census

|---|---:|---:|
| COLUMN guards (backlog) | 88 | **59** |
| `self-healing.ts` | 38 | **11** |

## First, a correction to my own earlier reports

I have repeatedly flagged **L2894 / 3297 / 8948 / 9089** as a blocked cluster gated on a `listTasks({ column })` API decision. **Those line numbers are comments**, not guards — they describe conversions that were already finished. I was reading grep output that included comment text and never checked it against the census AST. The real remaining sites were elsewhere and mostly tractable; this PR takes 8 of them.

## The bug

Two multi-id guards asking "is this card live or already finished?", both matching **nothing** on a renamed board.

`reconcilePreExecutionWorktrees` is the destructive one. With the literal, a renamed board made every working card look parked and the sweep **seized its worktree**. The test shows it directly — reverting the guard logs:

```
reconcilePreExecutionWorktrees: released 1 pre-execution worktree(s)
release("FN-9300", "parked pre-execution in 'building'")
```

for a card sitting in a live working lane.

## Scope differs between the two, on purpose

- **Worktree sweep → project union.** It filters every card on the board, so per-task resolution is one IR read per row on a recurring sweep; the union needs no per-task workflow selection. Over-inclusion is the safe direction: a column wrongly counted as active only leaves a worktree alone, while a missed one seizes it from live work. `archived` stays excluded, matching the literals replaced.
- **Agent sweep → per task.** Its loop has already awaited `getTask`, so the card's own workflow is the natural scope, and a cache keeps it to one IR read per workflow across the agent list. Over-inclusion is safe here too — an extra column leaves an agent running rather than recovering it out from under live work.

## Verification

- **Revert-proof, and it demonstrates the harm rather than just a mismatch:** reverting the worktree guard fails both new cases and releases a worktree it must not touch.
- 7 suites, **482 tests green**; tsc and eslint clean

## Remaining

11, down from 38. The FN-5256 liveness cluster (L5478/5515/5516/5544) is the substantive one left; the rest are singletons I have not yet read closely.

### DIFF
--- a/packages/engine/src/__tests__/self-healing-prewip-role-vocabulary.test.ts
+++ b/packages/engine/src/__tests__/self-healing-prewip-role-vocabulary.test.ts
@@ -369,3 +369,55 @@ describe("self-healing pause-abort requeue target vocabulary", () => {
     expect(vi.mocked(moveTask).mock.calls[0]?.[1]).toBe("todo");
   });
 });
+
+/*
+FNXC:WorkflowResolvedColumns 2026-07-31-17:10 (fleet — active/terminal membership):
+Two multi-id membership guards, both asking "is this card live or finished?" and both matching NOTHING
+on a renamed board.
+
+`reconcilePreExecutionWorktrees` is the destructive one: a renamed board made every working card look
+parked, so the sweep would have SEIZED a worktree from live work. That is why over-inclusion is the
+acceptable failure direction for it and the project union is the right scope — a column wrongly counted
+as active only leaves a worktree alone.
+*/
+describe("self-healing active/terminal membership vocabulary", () => {
+  function releaseStoreFor(ir: WorkflowIr | undefined, tasks: Task[]) {
+    return {
+      ...storeFor(ir),
+      getSettings: vi.fn(async () => ({ globalPause: false, enginePaused: false })),
+      listTasks: vi.fn(async () => tasks),
+      getProjectWorkflowIds: vi.fn(async () => (ir ? ["custom:renamed"] : [])),
+      listWorkflowDefinitions: vi.fn(async () => (ir ? [{ id: "custom:renamed", ir }] : [])),
+      logEntry: vi.fn(async () => undefined),
+      recordRunAuditEvent: vi.fn(async () => undefined),
+    } as unknown as TaskStore;
+  }
+
+  const withWorktree = (id: string, column: string): Task => ({
+    id,
+    column,
+    worktree: `/tmp/wt/${id}`,
+    updatedAt: "2020-01-01T00:00:00.000Z",
+    columnMovedAt: "2020-01-01T00:00:00.000Z",
+  } as unknown as Task);
+
+  it("does not seize the worktree of a card in a RENAMED working lane", async () => {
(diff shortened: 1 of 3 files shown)


gsxdsm:
**Seven PRs are converting `self-healing.ts` concurrently.** I measured all seven — hunk sets and the inert-sync-lane check — rather than reasoning about titles. The result collapses the queue from seven to three.

### Five of them form a strict chain

Hunk sets, verified by set containment (not by reading):

```
#3075 (6 hunks)  ⊂  #3080 (8)  ⊂  #3083 (13)  ⊂  #3084 (16)  ⊂  #3086 (19)
```

Every hunk in each is present in the next. I also diffed shared bodies earlier for #3075/#3080 — byte-identical, same FNXC stamp — so this is one line of work re-proposed five times, not five different conversions.

**Only #3086 is needed. #3075, #3080, #3083 and #3084 can close as superseded with nothing lost.**

### The two independent PRs

| | hunks | overlap |
|---|---|---|
| **#3074** | 27 | ∩ #3086 = 10 (`30 949 973 1002 1484 12083 12108 12151 14157 14178`); ∩ #3078 = 3 (`5327 5356 7326`) |
| **#3078** | 7 | ∩ #3086 = 1 (`12906`); ∩ #3074 = 3 (above) |

Suggested order by rebase cost: **#3074** (largest, 17 hunks nobody else touches) → **#3078** (near-disjoint) → **#3086** rebased.

### Blocking issue on the chain's survivor

Measured with the hardened check from #3079:

| PR | inert sync-lane guards in `self-healing.ts` |
|---|---|
| #3074, #3075, #3078 | **0 — clean** |
| #3080, #3083, #3084, **#3086** | **5** |

Everything from #3080 onward carries `resolveMoveFanoutColumnsSync`, which reads `this.store.resolveTaskWorkflowIrSync?.(taskId)`. That reader returns `undefined` for every task under PostgreSQL, so the resolver hands back the **default** IR — a real IR resolving the default board's traits, not `undefined` falling through to the legacy seeds. `columnsWithFlag(ir, "countsTowardWip")` yields `in-progress`, already in the seed set, so the five guards compare against exactly the literals they replaced. Census drops by five; behaviour changes on no board. Refuted end-to-end on a live PostgreSQL store in **#3058** (merged).

Note the chain's root, **#3075, does not have this** — it predates the helper. The defect entered at #3080 and was carried forward four times.

So the recommendation is not simply "merge #3086": it is **merge #3086 after the sync helper is resolved**, either by carrying the lanes on the `task:moved` payload (removes the class) or by keeping the literals and flagging them, as `merge-queue-ops-2.ts:53` already does. #3079 would fail all four branches at 34 against a baseline of 29; please don't clear that with `--update-baseline`.

### Net

Seven PRs → three (**#3074**, **#3078**, **#3086**), with one blocking fix on #3086 shared by three of the four PRs it supersedes.

Extraction is scripted; happy to re-run after any of these land.

gsxdsm:

```
```

gsxdsm:

PR #3087, opened 2026-07-31
fix(engine): serialise wedge handling per task, then convert the lane guards it was blocking (5 → 1)

gsxdsm:
The largest unclaimed census cluster, and the one two earlier fleet passes explicitly declined.

## The standing blocker, taken on

Both passes converted these four ids and reverted, each time after the same test went red:

```
task-wedge-notification.test.ts > sends one actionable push and mailbox message per active terminal episode
  expected 2 calls, got 1
```

Their diagnosis was right and I have kept it: this branch **resolves** a wedge episode, `handleTaskUpdated` starts it fire-and-forget from a synchronous `(task) => void` listener, and **any** await introduced before the resolve lets a re-wedge arriving close behind reach `claim` while the previous episode is still active — `claimed: false`, second operator notification silently dropped. Column resolution needs an await, so the conversion could not be made safe from inside the branch.

Both notes named the fix and left it for "whoever owns the wedge episode contract": *serialise wedge handling per task*. This PR does that, then takes the conversion.

## 1. Serialisation

`enqueueWedgeHandling` chains handling per task id, so resolve-then-claim keeps its order however many awaits either branch acquires. Details that matter:

- **Keyed by task, not global** — different tasks stay concurrent, so this is not a throughput regression on a busy board.
- **The map entry is dropped when its chain drains**, and only if no later link was appended while it ran, so it does not grow with the task table.
- **Links never reject.** `maybeNotifyTaskWedge` already owns its error handling; a rejected link would poison every later notification for that task.

## 2. The conversion it was blocking

The four ids are an enumeration of *"every lane except review"* — the lanes whose occupancy proves a wedged card's lifecycle has visibly resumed. On a renamed board none of them matched, so a recovered card's episode never resolved. Two consequences, and the second is worse than the first:

1. the operator keeps an open "needs operator action" alert for work that has moved on;
2. an active episode **suppresses re-claim**, so the *next* genuine wedge on that task is never delivered.

Membership over the four roles, legacy-seeded, so an unconverted board resolves exactly the four ids it used to compare.

## Measured

**The acceptance test the earlier notes named is the gate on both halves.** With the conversion and *without* the serialisation, "sends one actionable push and mailbox message per active terminal episode" fails exactly as they reported. With the serialisation, green. I reproduced their finding rather than taking it on trust — it is the evidence that the serialisation is load-bearing and not incidental refactoring.

| | result |
|---|---|
| `task-wedge-notification.test.ts` | **15/15** (2 new) |
| notification suites | **11 files / 234 tests pass** |
| `tsc --noEmit -p packages/engine` | clean |
| census `--strict`, `check-lane-wiring`, `check-inert-sync-lane-conversions`, `check-fnxc-future-dates` | clean |

**MUTATION**: restoring the four literals fails the renamed-recovery case and leaves its paired negative green.

**A vacuity I caught and fixed, worth stating plainly.** My first version of the renamed case recovered the card with `status: "queued"`. `hasProgressed` is an OR whose other arm is *"status is a non-failed string"* — so that arm answered true and the column comparison never ran. The mutation did not fail it. The case now clears `status` and `error` together, which makes column membership the only thing that can resolve the episode, and the paired negative uses the identical shape so only the lane differs.

## Census

| | before | after |
|---|---|---|
| `notification-service.ts` | 5 | **1** |
| repo backlog | 71 | **67** |

## The remaining 1, flagged not guessed

`isManualMergeHold` (`task.column !== "in-review"`) is sync, and so is its only caller `classifyWorkflowTransitionNotification`, reached from the same `handleTaskUpdated` listener. Converting it means making that whole chain async — a change to notification *classification ordering* against every other `task:updated` handler, which is a different contract from the episode one this PR owns. The serialisation added here does not cover it: it wraps wedge handling, not transition classification. Threading a pre-resolved `LifecycleColumns` in as a parameter is the likely fix, and it wants the same gate-placement judgement applied deliberately rather than swept in behind this.

### DIFF
--- a/packages/engine/src/notification/__tests__/task-wedge-notification.test.ts
+++ b/packages/engine/src/notification/__tests__/task-wedge-notification.test.ts
@@ -4,7 +4,24 @@ import { NotificationService } from "../notification-service.js";
 import { describeSelfHealingNoActionWedge, describeTaskWedge } from "../task-wedge-notification.js";
 
 type Listener = (task: Task) => void;
-function fixture() {
+
+/*
+FNXC:WorkflowResolvedColumns 2026-07-31-21:35:
+A board whose lanes carry no legacy id: hold `drafting`, wip `building`, review `checking`,
+complete `shipped`. Supplied through `listWorkflowDefinitions`, the only store read
+`resolveProjectColumnsForRoles` makes and one that is answerable under PostgreSQL.
+*/
+const RENAMED_IR = {
+  version: "v2", id: "wf-renamed", name: "renamed", nodes: [], edges: [],
+  columns: [
+    { id: "drafting", name: "Drafting", traits: [{ trait: "hold", config: { release: "capacity" } }] },
+    { id: "building", name: "Building", traits: [{ trait: "wip", config: { limitSetting: "maxConcurrent" } }] },
+    { id: "checking", name: "Checking", traits: [{ trait: "merge" }] },
+    { id: "shipped", name: "Shipped", traits: [{ trait: "complete" }] },
+  ],
+};
+
+function fixture(workflowIr?: unknown) {
   const listeners = new Set<Listener>();
   let wedge: Task["wedgeNotification"];
   const store = {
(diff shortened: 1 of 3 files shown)


gsxdsm:

Only `scripts/lib/lifecycle-column-census-baseline.json` conflicted; both source files applied cleanly. Resolved by taking main's copy and re-running `--strict --update-baseline` rather than hand-merging — that file is generated, and a hand-merge yields a count matching neither branch while looking plausible.

The re-record was **required**, not cosmetic: `--strict` failed before it and passed after, because the ratchet tracks the deliberate count alongside the guard count so a marker cannot shrink the backlog quietly. Amended into your commit.

Verified: `task-wedge-notification.test.ts` **15 pass**, `tsc --noEmit` (engine) 0 errors, `census --strict` exit 0.

PR #3088, opened 2026-07-31
fix(self-healing): resolve the FN-5256 liveness cluster (self-healing 11 → 5)

gsxdsm:
Sixth cut into `self-healing.ts`. **Stacked on #3075 → #3080 → #3083 → #3084 → #3086.**

## Census

|---|---:|---:|
| COLUMN guards (backlog) | 88 | **53** |
| `self-healing.ts` | 38 | **5** |

## The bug

FN-5256 exists because nulling a live task's worktree metadata **yanks the worktree from a still-running shell**. The guard enforcing it compared `task.column` against `"in-progress"`/`"in-review"` — so on a renamed board it protected **nothing**, and the incident it was written to prevent came back.

Revert the guard and the sweep writes `{ worktree: null, branch: null, sessionFile: null }` onto a live card in a working lane. The test asserts on `updateTask` **not** being called, because here the protection *is* the absence of a write — a test counting only the return value would pass while metadata was being cleared.

## Why the project union is sound here

All four comparisons live in one board-wide loop, resolved once per sweep. Over-inclusion is the safe direction for **every** use:

| Site | Wider set means | Direction |
|---|---|---|
| terminal skip | repairs less | benign |
| wip/review in scopeOverride bypass | **fewer** rows take the bypass that clears metadata | safe |
| wip/review in FN-5256 guard | **more** rows protected | safe |

A *missed* column is what does the damage — and the literals missed every one on a renamed board.

## Verification

- **Revert-proof on the write:** reverting the FN-5256 guard alone clears `branch: null` on a live renamed-lane card.
- 7 suites, **483 tests green**; tsc and eslint clean

## Remaining

5, down from 38. I have not read these closely yet — flagging them as **unread**, not blocked.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved self-healing workflows on boards with renamed or custom workflow lanes.
  * Ensured paused and aborted tasks are recovered and requeued to the correct workflow lane.
  * Improved task fan-out, agent recovery, parked-task cleanup, and peer-progress detection.
  * Preserved support for legacy workflow identifiers when custom workflow details cannot be resolved.
  * Preserved live worktree information during recovery.
  * Improved starved refinement recovery and priority handling.

### DIFF
--- a/packages/engine/src/__tests__/self-healing-prewip-role-vocabulary.test.ts
+++ b/packages/engine/src/__tests__/self-healing-prewip-role-vocabulary.test.ts
@@ -369,3 +369,55 @@ describe("self-healing pause-abort requeue target vocabulary", () => {
     expect(vi.mocked(moveTask).mock.calls[0]?.[1]).toBe("todo");
   });
 });
+
+/*
+FNXC:WorkflowResolvedColumns 2026-07-31-17:10 (fleet — active/terminal membership):
+Two multi-id membership guards, both asking "is this card live or finished?" and both matching NOTHING
+on a renamed board.
+
+`reconcilePreExecutionWorktrees` is the destructive one: a renamed board made every working card look
+parked, so the sweep would have SEIZED a worktree from live work. That is why over-inclusion is the
+acceptable failure direction for it and the project union is the right scope — a column wrongly counted
+as active only leaves a worktree alone.
+*/
+describe("self-healing active/terminal membership vocabulary", () => {
+  function releaseStoreFor(ir: WorkflowIr | undefined, tasks: Task[]) {
+    return {
+      ...storeFor(ir),
+      getSettings: vi.fn(async () => ({ globalPause: false, enginePaused: false })),
+      listTasks: vi.fn(async () => tasks),
+      getProjectWorkflowIds: vi.fn(async () => (ir ? ["custom:renamed"] : [])),
+      listWorkflowDefinitions: vi.fn(async () => (ir ? [{ id: "custom:renamed", ir }] : [])),
+      logEntry: vi.fn(async () => undefined),
+      recordRunAuditEvent: vi.fn(async () => undefined),
+    } as unknown as TaskStore;
+  }
+
+  const withWorktree = (id: string, column: string): Task => ({
+    id,
+    column,
+    worktree: `/tmp/wt/${id}`,
+    updatedAt: "2020-01-01T00:00:00.000Z",
+    columnMovedAt: "2020-01-01T00:00:00.000Z",
+  } as unknown as Task);
+
+  it("does not seize the worktree of a card in a RENAMED working lane", async () => {
+    const store = releaseStoreFor(RENAMED_WITH_REVIEW, [withWorktree("FN-9300", "building")]);
+    const release = vi.fn(async () => true);
+    const manager = Object.create(SelfHealingManager.prototype) as SelfHealingManager;
+    (manager as unknown as Record<string, unknown>).store = store;
+    (manager as unknown as Record<string, unknown>).options = { releasePreExecutionWorktree: release };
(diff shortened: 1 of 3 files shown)


gsxdsm:
Measured on this branch with the hardened check from #3079: **5 inert sync-lane guards in `self-healing.ts`** (34 total against a baseline of 29).

Same `resolveMoveFanoutColumnsSync` that #3080, #3083, #3084 and #3086 carry. It reads `this.store.resolveTaskWorkflowIrSync?.(taskId)`, whose selection reader returns `undefined` for every task under PostgreSQL — the shipped backend — so the resolver hands back the **default builtin IR**. Not `undefined` falling through to your legacy seeds: a real IR resolving the default board's traits, so `columnsWithFlag(ir, "countsTowardWip")` yields `in-progress`, which the seed set already contains. The five guards compare against exactly the literals they replaced, on every board. Census drops by five; behaviour changes nowhere. Refuted end-to-end on a live PostgreSQL store in #3058 (merged).

That is now **five open PRs sharing one defective helper**, so it is worth fixing once rather than five times. Two honest options, unchanged:

1. Carry the resolved lanes on the `task:moved` payload — no listener resolves at all, which removes the class rather than one instance.
2. Keep the literals and flag them, as `merge-queue-ops-2.ts:53` and `notification-service.ts` do. Accurate debt beats a false green.

#3079 would fail this branch at 34 vs 29. Please don't clear it with `--update-baseline`.

**One thing I could not determine**, stated rather than guessed: whether this PR supersedes any of the others by hunk containment. #3086 has been rebased since my seven-way analysis, so its line numbers have moved and the two hunk sets are no longer on a comparable base. The containment chain I posted earlier (`#3075 ⊂ #3080 ⊂ #3083 ⊂ #3084 ⊂ #3086`) was measured on a single base and still holds for those five; I am not extending it to this PR without a common base to measure against. If you know this one subsumes others, say so and I will re-run the extraction against whatever base you name.

gsxdsm:
**The finding is real, but "align the two sets" is the wrong fix — and this file already records why.**

I cannot rebase this myself (it is CONFLICTING, it is `self-healing.ts`, and I do not have the budget to run the engine suite behind a resolution). But the analysis is worth more than the rebase, so:

## There are at least THREE review-set definitions in this file, each correct for a different question

| resolver | review set |
|---|---|
| `resolveReviewColumnsFor` | `mergeOrchestration ∪ mergeBlocker ∪ humanReview` |
| `resolveDependencySatisfactionColumns` | `mergeBlocker ∪ humanReview` — orchestration **deliberately excluded** |
| `resolveMoveFanoutColumnsSync` (this PR) | `mergeOrchestration ∪ humanReview` |

The third omitting `mergeBlocker` does look wrong. But making it match the async twin by hand is the move `self-healing.ts:4911` on main was written to prevent:

> The first version unioned all three review roles, which counts a `mergeOrchestration`-only column as satisfied. `resolveDependencySatisfactionColumns` … defines review as `mergeBlocker ∪ humanReview` and deliberately excludes merge orchestration. **Two readers of one fact disagreeing is how this program has already gone wrong; the fix is to call the existing one, not to match it by hand.**

That note came from #2883's own `P1` about an *overbroad* review set — the opposite symptom, same root cause.

## So the question to answer first

Not "which flags does the async twin union?" but **"what is `handleTaskMovedFanout` actually asking?"** If it is "is this card in a lane that blocks merge", that is `resolveDependencySatisfactionColumns`' definition, not `resolveReviewColumnsFor`'s. Adding `mergeBlocker` until the sets look alike may fix the reported case and reintroduce the one #2883 fixed.

Whichever it is, the resolution should be **calling the existing resolver** rather than growing a third hand-maintained union.

## Related

This is the role-set twin of the arity problem in #3111: several resolvers answering one question differently, and each author reaching for whichever they find first. Same shape as `resolveLifecycleColumns` being read as membership in #3084, #3096, and this PR.

Not resolving the thread, since I am not fixing the code — but the next person should not need to re-derive this.

gsxdsm:

PR #3092, opened 2026-07-31
fix(self-healing): resolve the last four lane guards (self-healing 5 → 1 deliberate)

gsxdsm:
Final cut into `self-healing.ts`. **Stacked on #3075 → #3080 → #3083 → #3084 → #3086 → #3088.**

## Census

|---|---:|---:|
| COLUMN guards (backlog) | 88 | **49** |
| `self-healing.ts` | 38 | **1** (another worker's deliberate) |

`self-healing.ts` no longer appears in the census top-files list.

## The four

Each has a single call site in async scope, so all four take the **required-parameter** shape rather than optional-with-a-fallback:

- **`isWorkspaceOwnerLive`** decides whether a *live* owner's land lease is reclaimed. An optional set silently restores the literal at any call site that forgets it.
- **`isPhantomExecutorBinding`**: the caller has *already* resolved this lane set to reach the call (`lanesOfReclaim(task.id).wip`), so threading it in costs nothing and keeps guard and gate in agreement. A second, independently-resolved set is how the two drift.
- **`reconcileOrphanedPendingStepResults`** skips executor-owned rows in two places — page filter and post-re-read verify — which must agree, or the sweep selects a row then declines to act with nothing logged. Both now read one project union.

## One thing I did not ship

**No test for the orphaned-pending sweep.** I wrote one; it passed with the guard reverted to the literal. Something else in that path — not this guard — prevents the write under the harness, so the test proved nothing. Removed rather than shipped: a test that passes with the change reverted is not a test.

**That sweep's renamed-board behaviour is therefore UNCOVERED.** Flagging it plainly rather than letting a green suite imply otherwise. It matters: per AGENTS.md the `failed` rewrite is what keeps the merge gate closed, so a wrong answer there falsifies a task's review state mid-run.

## The remaining 1

L6094 is another worker's explicit **"FLAGGED AND LEFT COUNTED"** decision, with sound reasoning — a log-dedup closure defined before the lane prefetch it would need, where the degraded answer costs a duplicate log line, not a lifecycle decision. Left as they decided. Whether to mark it `DELIBERATE` is their call or yours; I did not want to overrule another worker's stated choice silently.

## Verification

- 7 suites, **483 tests green**; tsc and eslint clean

### DIFF
--- a/packages/engine/src/__tests__/self-healing-prewip-role-vocabulary.test.ts
+++ b/packages/engine/src/__tests__/self-healing-prewip-role-vocabulary.test.ts
@@ -369,3 +369,55 @@ describe("self-healing pause-abort requeue target vocabulary", () => {
     expect(vi.mocked(moveTask).mock.calls[0]?.[1]).toBe("todo");
   });
 });
+
+/*
+FNXC:WorkflowResolvedColumns 2026-07-31-17:10 (fleet — active/terminal membership):
+Two multi-id membership guards, both asking "is this card live or finished?" and both matching NOTHING
+on a renamed board.
+
+`reconcilePreExecutionWorktrees` is the destructive one: a renamed board made every working card look
+parked, so the sweep would have SEIZED a worktree from live work. That is why over-inclusion is the
+acceptable failure direction for it and the project union is the right scope — a column wrongly counted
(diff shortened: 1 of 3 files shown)


gsxdsm:

PR #3093, opened 2026-07-31
fleet(engine): self-healing 34 → 11 — the async-reachable sweeps, and one conversion withdrawn

gsxdsm:
**Replaces #3074, which I am closing.** #3075 landed the pause-abort *router* conversion while that branch was open. Rebuilt off current `main` as one commit containing the remainder.

## #3075 left its own sweep half-converted

It resolved `classifyPausedAbortWorkflowRecovery` — the router — and left three guards in the sweep BODY comparing ids: the notification-attach guard, the `moveTask` target, and the log line that names which route ran.

That is the state the converted-sweeps ratchet exists to catch, and it is worse than converting nothing: the route now fires correctly on a renamed board and the body then acts on the wrong lane. The target is the dangerous one — `moveTask` rejects an undeclared column **except** under `recoveryRehome` with a legacy id (`moves.ts:570`, the #1411 escape hatch), so a converted guard feeding the literal `"todo"` rehomes the card into a column its workflow does not declare.

## What is converted

| site | note |
|---|---|
| `isPhantomExecutorBinding` | caller resolved `lanesOfReclaim(task.id).wip` **three lines above the call**, then passed a task whose column the predicate compared against `in-progress` |
| `recoverPausedAbortFailures` | guard + `moveTask` target + log line, together |
| `recoverAgentsRunningOnInactiveTasks` | lifecycle skip set, plus the omitted `parkedColumns` argument |
| `reconcileTaskWorktreeMetadata` | 3 guards |
| `reconcileOrphanedPendingStepResults` | 2 guards, **both legs** — this sweep *rewrites* pending results to `failed`, so on a renamed board it did that to a live executor |
| `reconcilePreExecutionWorktrees` | a four-id literal in a sweep that removes worktrees |
| `recoverStarvedRefinementTriageTasks` | 2 peer counts that read zero, so escalation never fired |

All through the **async** `resolveProjectColumnsForRoles`, whose only store read is `listWorkflowDefinitions()` — answerable under PostgreSQL. That is what separates these from the inert kind.

Also a real dropped-behaviour bug: the `evaluateParkedAgentTaskLink` call omitted `parkedColumns`, falling back to `LEGACY_PARKED_COLUMNS`. A live durable agent linked to a card in a renamed hold lane read as not-parked, so the safeguard preserving its task link never applied.

## Withdrawn: the `task:moved` fan-out

I wrote the sync-IR conversion for it, measured it, and removed it. `getTaskWorkflowSelectionImpl` returns `undefined` **unconditionally** under PostgreSQL, so `resolveTaskWorkflowIrSync` always answers with the default builtin IR and `columnsWithFlag` on it yields exactly the legacy ids — inert on every board.

Worse than the literal, because **the literal is counted**. Four guards would have left the census and the file would read as converted. My own test passed only because its store mock supplied a renamed IR: it pinned the helper's shape, not production behaviour. The refutation is recorded in place, and `check-inert-sync-lane-conversions` exits 0 on this branch.

## An error the suite caught, stated plainly

My first pass mapped the notification-attach guard (`fresh.column === "todo"`) onto `activeWork` membership — which includes wip — so an in-progress park attached a transition notification it should not have. `self-healing-paused-abort-recovery > "rehomes an in-progress pause-abort park back to todo"` failed on it.

The fix is not a narrower set: both guards ask **one** question — *"is the card already at the requeue target?"* — which the literal happened to spell twice as `=== "todo"`. The target is now resolved once, before the write, and both read it. Deriving one question two ways is how a converted guard and an unconverted target drift apart in the first place.

## Census

| | before | after |
|---|---|---|
| `self-healing.ts` | 34 | **11** |
| repo backlog | 65 | **42** |

## Measured

- `src/__tests__/self-healing*` + `task-agent*` — **42 files / 836 tests pass**
- **MUTATION**: restoring the `in-progress` literal fails the renamed-lane orphaned-step case and leaves its paired negative (a renamed *review*-lane orphan must still be rewritten) green
- `tsc --noEmit -p packages/engine` clean; **`check-inert-sync-lane-conversions` exits 0**; census `--strict`, `check-lane-wiring`, `check-fnxc-future-dates` clean

## The remaining 11, flagged not guessed

- **4** — the `task:moved` fan-out, withdrawn above; blocked on a sync-capable selection reader.
- **1** — the log-dedup closure (~line 5986): pre-existing flag, sits before the lane prefetch it needs, and the degraded answer costs a duplicate log line rather than a lifecycle decision.
- **1** — the synthetic `{ column: "todo" }` for a *missing* task: deliberate, and now correct rather than unconverted, because `parkedColumns` is legacy-seeded.
- The rest are status/deliberate classifications the census counts but that are not lane guards.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved workflow recovery for workflows with renamed lifecycle columns.
  * Pending executor results now remain pending in implementation lanes and are correctly marked failed in other lanes.
  * Improved handling of paused, orphaned, stalled, and abandoned workflow steps.
  * Recovery actions now consistently target the appropriate intake, hold, review, and work-in-progress columns.

### DIFF
--- a/packages/engine/src/__tests__/self-healing-orphaned-pending-step-results.test.ts
+++ b/packages/engine/src/__tests__/self-healing-orphaned-pending-step-results.test.ts
@@ -58,7 +58,26 @@ function task(id: string, overrides: Partial<Task> = {}): Task {
   } as Task;
 }
 
-function storeFor(tasks: Task[]): TaskStore & EventEmitter {
+/*
+FNXC:WorkflowResolvedColumns 2026-07-31-17:55:
+A board whose implementation lane is `building`. Supplied through `listWorkflowDefinitions`, which is
+the ONLY store read `resolveProjectColumnsForRoles` makes — a project-wide async read that works under
+PostgreSQL, unlike the sync workflow-SELECTION reader that makes `resolveTaskWorkflowIrSync`-based
+conversions inert. Using the async helper here is what makes the conversion real rather than decorative.
+*/
+const RENAMED_WIP_IR = {
+  version: "v2",
+  id: "custom:wf",
+  nodes: [],
+  edges: [],
+  columns: [
+    { id: "drafting", name: "drafting", traits: [{ trait: "hold", config: { release: "capacity" } }] },
+    { id: "building", name: "building", traits: [{ trait: "wip", config: { limitSetting: "maxConcurrent" } }] },
+    { id: "checking", name: "checking", traits: [{ trait: "merge" }] },
+  ],
+};
+
+function storeFor(tasks: Task[], workflowIr?: unknown): TaskStore & EventEmitter {
   const tasksById = new Map(tasks.map((entry) => [entry.id, entry]));
   return Object.assign(new EventEmitter(), {
     getSettings: vi.fn(async () => ({ globalPause: false, enginePaused: false } as Settings)),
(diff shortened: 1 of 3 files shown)


gsxdsm:
Superseded — #3078 landed a third overlapping conversion of this file while this was open. Rebuilt against current main; see the replacement.

PR #3094, opened 2026-07-31
fleet(engine): self-healing 23 → 6 — async-reachable guards, plus 3 of 4 fan-out guards the sync path could not serve

gsxdsm:
**Replaces #3093, which I am closing.** Third rebuild of this work.

## A coordination note first, because it is costing more than the code

`self-healing.ts` has had **three** overlapping conversions land from other lanes while my branch was open — #3049, #3075, #3078. Every time, replaying my commits produced conflicts that were all the same shape: *same guard, two spellings, different variable names*. Each rebuild is a full cycle spent on merge mechanics rather than on lanes.

I have rebuilt against `main`'s own census each time rather than argue about whose spelling wins, and this PR contains only what `main` (23) does not have. But if this file is going to keep receiving concurrent fleet passes, one lane should own it — otherwise the next PR pays the same tax again.

## Converted

| site | note |
|---|---|
| `isPhantomExecutorBinding` | caller resolved `lanesOfReclaim(task.id).wip` **three lines above the call**, then passed a task whose column the predicate compared against `in-progress` |
| `recoverPausedAbortFailures` **body** | #3075 converted this sweep's *router* and left three body guards comparing ids |
| `reconcilePreExecutionWorktrees` | a four-id literal in a sweep that **removes worktrees** |
| `recoverStarvedRefinementTriageTasks` | 2 peer counts that read zero, so escalation never fired |
| `evaluateParkedAgentTaskLink` | the omitted `parkedColumns` argument |

All through the **async** `resolveProjectColumnsForRoles`, whose only store read is `listWorkflowDefinitions()` — answerable under PostgreSQL. That is what separates these from the inert kind.

**The half-converted sweep is the important one.** A router that resolves correctly feeding a body that compares ids is worse than converting neither: the route now fires on a renamed board and the body then acts on the wrong lane. The `moveTask` **target** is the sharp end — an undeclared target is rejected *except* under `recoveryRehome` with a legacy id (`moves.ts:570`, the #1411 escape hatch), so a converted route feeding the literal `"todo"` rehomes the card into a column its workflow does not declare, which is the state other reconcilers exist to repair.

**A real dropped-behaviour bug**: `evaluateParkedAgentTaskLink` was called without `parkedColumns`, falling back to `LEGACY_PARKED_COLUMNS`. A live durable agent linked to a card resting in a renamed hold lane read as not-parked, so the safeguard preserving its task link never applied.

## Withdrawn: the `task:moved` fan-out

I wrote the sync-IR conversion, measured it, removed it. `getTaskWorkflowSelectionImpl` returns `undefined` **unconditionally** under PostgreSQL, so `resolveTaskWorkflowIrSync` always answers with the default builtin IR and `columnsWithFlag` on it yields exactly the legacy ids — inert on every board.

Worse than the literal, because **the literal is counted**. My own test passed only because its store mock supplied a renamed IR: it pinned the helper's shape, not production behaviour. The refutation is recorded in place, and `check-inert-sync-lane-conversions` exits 0 on this branch.

## One question, one answer

An earlier pass of this work mapped the notification-attach guard onto a wider `activeWork` set, and `self-healing-paused-abort-recovery > "rehomes an in-progress pause-abort park back to todo"` caught it — an in-progress park attached a transition notification it should not have.

The fix is not a narrower set. Both guards ask **one** question — *"is the card already at the requeue target?"* — which the literal happened to spell twice as `=== "todo"`. The target now resolves once, before the write, and both read it. Deriving one question two ways is exactly how a converted guard and an unconverted target drift apart.

## Census

| | before | after |
|---|---|---|
| `self-healing.ts` | 23 | **11** |
| repo backlog | 53 | **41** |

## Measured

- `src/__tests__/self-healing*` + `task-agent*` — **42 files / 836 tests pass**
- `tsc --noEmit -p packages/engine` clean; **`check-inert-sync-lane-conversions` exits 0**; census `--strict`, `check-lane-wiring`, `check-fnxc-future-dates` clean

## The remaining 11, flagged not guessed

- **4** — the fan-out, withdrawn above; blocked on a sync-capable selection reader.
- **1** — the log-dedup closure: pre-existing flag; it sits before the lane prefetch it needs, and the degraded answer costs a duplicate log line, not a lifecycle decision.
- **1** — the synthetic `{ column: "todo" }` for a *missing* task: deliberate, and now correct rather than unconverted, because `parkedColumns` is legacy-seeded.
- The rest are status/deliberate classifications the census counts but that are not lane guards.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved workflow automation for boards using renamed or customized workflow lanes.
  * Fixed task completion fan-out, branch rebinding, recovery, and stalled-task detection across custom lifecycle columns.
  * Prevented completion actions from triggering when tasks move back to the work-in-progress lane.
  * Improved cleanup and pause recovery behavior for customized workflows.

### DIFF
--- a/packages/engine/src/__tests__/self-healing-completion-fanout.test.ts
+++ b/packages/engine/src/__tests__/self-healing-completion-fanout.test.ts
@@ -178,13 +178,22 @@ describe("self-healing completion fan-out", () => {
     store.emit("task:moved", { task: t, from: "in-review", to: "done", source: "user" });
     store.emit("task:moved", { task: t, from: "done", to: "archived", source: "engine" });
     store.emit("task:moved", { task: t, from: "in-review", to: "todo", source: "user" });
-    await Promise.resolve();
-    expect(spy).toHaveBeenCalledTimes(2);
+    /*
+    FNXC:WorkflowResolvedColumns 2026-07-31-23:40:
+    `await Promise.resolve()` was draining exactly one microtask, which coupled this case to the
+    number of awaits inside a FIRE-AND-FORGET path. The listener does not await the fan-out and never
+    did, so how many microtasks it takes is not the contract — "it ran, twice, for the right
+    transitions" is. Resolving the lanes adds an await, so the drain is now written against the
+    invariant instead of against the old await count.
+    */
+    await vi.waitFor(() => { expect(spy).toHaveBeenCalledTimes(2); });
     expect(spy).toHaveBeenNthCalledWith(1, "FN-L", { worktreeHint: undefined });
 
     mgr.stop();
     store.emit("task:moved", { task: t, from: "in-review", to: "done", source: "user" });
-    await Promise.resolve();
+    /* The negative keeps a real drain: an unwired listener must stay silent after several ticks,
+       not merely after one. */
+    await new Promise((resolve) => setTimeout(resolve, 10));
     expect(spy).toHaveBeenCalledTimes(2);
   });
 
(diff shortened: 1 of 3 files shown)


gsxdsm:
## Second commit: the withdrawal was too broad, and I found the line

The first commit withdrew the `task:moved` fan-out conversion because the sync IR path is inert. That was right about the sync path and **wrong about the scope of the block**.

*"The listener is synchronous"* was never the real constraint. **"This particular guard's answer is consumed synchronously"** is — and that is true of only **one** of the four.

| guard | answer consumed | outcome |
|---|---|---|
| board-stall counter | synchronously, in the handler's own tick | stays literal, **counted** |
| review rebind | inside work the listener already `void`s | **converted, async** |
| completion fan-out (2 transitions) | inside work the listener already `void`s | **converted, async** |

The listener does not await those two calls and never did; their bodies already begin at a microtask boundary. Moving the guard into that same boundary preserves exactly the ordering property the withdrawn conversion would have broken — an async IIFE runs synchronously only up to its first `await`, which here is the first statement, so the rest of the listener continues in the same tick precisely as `void this.reconcileCompletedTask(...)` already did.

That lets them use the **async** `resolveProjectColumnsForRoles`, whose only store read is `listWorkflowDefinitions()` — answerable under PostgreSQL, and unaffected by either blocker that makes the sync path inert (#3103).

**What it fixes**, on every renamed board: a card entering the board's own review lane never had its branch rebound, and a card reaching its own complete or archive lane never ran the completion fan-out — so its worktree was never reclaimed and its dependents kept a `blockedBy` pointing at a blocker that had already finished.

## One test changed, and it is not appeasement

`"wires and unwires task:moved listener"` drained exactly **one** microtask with `await Promise.resolve()`, which coupled it to the number of awaits inside a **fire-and-forget** path. The listener does not await the fan-out, so how many microtasks it takes was never the contract — *"it ran, twice, for the right transitions"* is.

The drain is now `vi.waitFor` on that invariant. The paired negative got a **real** delay rather than a tighter one, so an unwired listener must stay silent after several ticks instead of after one — strictly stronger than before.

## Census (combined, both commits)

| | before | after |
|---|---|---|
| `self-healing.ts` | 23 | **6** |
| repo backlog | 53 | **31** |

## Measured

- 3 new cases; `self-healing-completion-fanout.test.ts` **10/10**.
- **MUTATION**: restoring the two literals fails **both** renamed cases and leaves the negative (a `checking → building` bounce must not fan out) and every legacy case green.
- `src/__tests__/self-healing*` + `task-agent*` — **42 files / 842 tests pass**.
- `tsc --noEmit -p packages/engine` clean; **`check-inert-sync-lane-conversions` exits 0**; census `--strict`, `check-lane-wiring`, `check-fnxc-future-dates` clean.

## The remaining 6

- **4** — the board-stall counter's guard (one line, four ids). The only one that genuinely needs a synchronous answer, now flagged with that narrower reason instead of the whole-listener one.
- **1** — the log-dedup closure: pre-existing flag.
- **1** — the synthetic `{ column: "todo" }` for a *missing* task: deliberate, and correct because `parkedColumns` is legacy-seeded.

gsxdsm:
## Merge gate run — green

Closing a gap in my own verification. I had been running targeted suites all session (`self-healing*`, `task-agent*`, etc.) and never the actual merge gate, which is what CI blocks on. Rebased onto current `main` and ran it:

```
pnpm test:gate

Test Files   4 passed (4)      Tests  161 passed (161)
Test Files  21 passed (21)     Tests  487 passed (487)
Test Files   3 passed (3)      Tests   13 passed (13)
Test Files   1 passed (1)      Tests   71 passed (71)
```

Zero FAIL lines, exit 0.

This is the branch worth running it on — it is the only substantive engine change among my open PRs (#3104 and #3108 are comment-only, #3110 and #3113 are core tests plus comments). A targeted suite passing is not the same claim as the gate passing, and I should have said so earlier rather than implying broader coverage than I had measured.

PR #3096, opened 2026-07-31
test(notifications): cover the wedge-episode renamed-lane clear that main's conversion lacked

gsxdsm:
**Rebased onto current `main`, and it shrank to a test.** Was "close the wedge-episode race, then resolve its lanes."

## What happened

Another worker landed **both halves of this PR independently** while it was open. Rebasing showed their versions are better, so I took theirs and dropped mine:

- **The serialisation** — theirs is `enqueueWedgeHandling(taskId, run)`, a general callback; mine was wedge-specific.
- **The conversion** — theirs is **project-union membership** over the four roles; mine was first-match-per-role via `resolveLifecycleColumns`. Membership is correct: more than one lane can fill a role on a renamed board, and first-match silently ignores the rest.

My rebased branch initially compiled to a **duplicate `wedgeHandlingChains` field and duplicate method** — caught by `tsc`, removed. Nothing of my implementation survives, and it shouldn't.

## What's left is worth landing

Their conversion has **no renamed-board test**. This adds one.

A card recovering into a renamed hold lane must **clear** its episode. Asserted through the *second* notification, because a stale active episode also **refuses the next genuine wedge its claim** — so the visible symptom is a real wedge going unannounced, not merely a stale alert.

The fixture needed `listWorkflowDefinitions`: `resolveProjectColumnsForRoles` unions across the project's workflows, so the per-task selection readers alone leave it resolving nothing and the test would pass for the wrong reason. That's how I found the mismatch — my original test failed against their implementation.

## Verification

- Green as written against **their** implementation
- **Revert-proof against theirs:** restoring the four literals fails it — 1 delivered, 2 expected
- 7 notification suites — **80 green**; `tsc` clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed wedge notifications so they can trigger again after a task recovers into a renamed workflow’s hold lane.

* **Tests**
  * Added regression coverage confirming that recovered tasks correctly clear their wedge state and support subsequent notifications.

### DIFF
--- a/packages/engine/src/notification/__tests__/task-wedge-notification.test.ts
+++ b/packages/engine/src/notification/__tests__/task-wedge-notification.test.ts
@@ -67,6 +67,53 @@ describe("task wedge notifications", () => {
     await service.stop();
   });
 
+  /*
+  FNXC:WorkflowResolvedColumns 2026-07-31-18:50 (fleet):
+  Clearing a wedge episode asks "has the card moved on?", which was four column literals. On a renamed
+  board none matched, so a RECOVERED card never cleared its episode and the operator kept an alert for
+  work that had already progressed — and, because the stale episode stays active, the NEXT genuine wedge
+  is refused its claim and never announced at all.
+
+  The board below recovers into `backlog` (hold). The assertion is the second message: with the literals
+  it never arrives, because the first episode was never resolved.
+  */
+  it("clears a wedge episode when the card recovers into a RENAMED hold lane", async () => {
+    const RENAMED_IR = {
+      version: "v2",
+      name: "renamed-notify",
+      columns: [
+        { id: "backlog", name: "Backlog", traits: [{ trait: "hold" }] },
+        { id: "building", name: "Building", traits: [{ trait: "wip" }] },
+        { id: "signoff", name: "Signoff", traits: [{ trait: "merge" }] },
+        { id: "shipped", name: "Shipped", traits: [{ trait: "complete" }] },
+      ],
+      nodes: [],
+      edges: [],
+    };
+    const { store, service, sendMessageOnce, task } = fixture();
+    /* `resolveProjectColumnsForRoles` unions lanes across the PROJECT's workflows, so the fake needs
+       `listWorkflowDefinitions` — the per-task selection readers alone leave it resolving nothing. */
+    Object.assign(store, {
+      getTaskWorkflowSelection: () => ({ workflowId: "custom:renamed", stepIds: [] }),
+      getTaskWorkflowSelectionAsync: async () => ({ workflowId: "custom:renamed", stepIds: [] }),
+      getWorkflowDefinition: async () => ({ ir: RENAMED_IR }),
+      listWorkflowDefinitions: async () => [{ id: "custom:renamed", ir: RENAMED_IR }],
+    });
+
+    await service.start();
+    store.emit(task({ column: "signoff" }));
+    await vi.waitFor(() => expect(sendMessageOnce).toHaveBeenCalledTimes(1));
+
+    // Recovered into the board's own hold lane — the episode must close.
+    /* status stays "failed" ON PURPOSE: the status disjunct (`status !== "failed"`) would otherwise
+       satisfy hasProgressed on its own and the column term would never decide anything. */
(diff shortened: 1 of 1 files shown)


gsxdsm:
**Not rebasing this — the conflict is substantive, and one side would reintroduce a bug main already fixed.** I started the rebase, inspected both hunks, and aborted without pushing. The remote is untouched.

## Hunk 2 is the one that matters

```
HEAD (main):
  const progressedLanes = await resolveProjectColumnsForRoles(this.store, ["hold","countsTowardWip","complete","archived"]);
  const hasProgressed = progressedLanes.has(task.column)

this branch:
  const progressedLanes = await this.resolveLifecycleColumnsForTask(task.id).catch(() => undefined);
  const progressedColumns = new Set(["todo","in-progress","done","archived"]);
  for (const id of [progressedLanes?.hold, progressedLanes?.wip, progressedLanes?.complete, progressedLanes?.archived]) …
```

Your version reads **one id per role** (`?.hold`, `?.wip`, …). That is `resolveLifecycleColumns`' first-per-role shape, and it is exactly the defect flagged as `Major` on #3084: *"resolve `hold` by membership, not first-per-role"* — a board declaring more than one pre-review working lane silently loses every lane after the first, so a card sitting in the second one still reads as not-progressed and the wedge alert stays open.

Main's `resolveProjectColumnsForRoles` returns **membership sets** and does not have that failure mode. So for this hunk, HEAD is the better answer and your change would be a regression — through no fault of the reasoning in your note, which is right about the underlying bug. Main just fixed it first, with the correct arity.

Worth keeping your legacy-id union idea though: main's version has no fallback if resolution yields nothing.

## Hunk 1 is your refactor and I should not pick it

```
HEAD:  this.enqueueWedgeHandling(task.id, () => this.maybeNotifyTaskWedge(task, wedge))
yours: this.enqueueWedgeHandling(task, wedge)
```

Both serialise wedge handling; the signatures differ. Choosing between them means understanding which `enqueueWedgeHandling` definition survives, and that is your refactor, not something I should decide by picking a side in a merge.

## Suggested resolution

Take **HEAD** for hunk 2 (keep the membership sets), take **yours** for hunk 1 (your refactor), then re-run `pnpm --filter @fusion/engine exec vitest run src/notification`. If you want the fallback, union the legacy ids into main's set rather than replacing it.

I aborted rather than guess because I broke a file earlier today resolving a self-healing conflict too quickly — 40 test files failed, caught before pushing. A conflicting PR is visible; a wrong lane guard is not.

gsxdsm:
**Attempted the resolution I proposed and it does not compile — aborted, remote untouched.**

I applied exactly what I suggested last comment (HEAD for the lane hunk, yours for the refactor hunk) and typechecked:

```
notification-service.ts(154,11): error TS2393: Duplicate function implementation.
notification-service.ts(386,20): error TS2300: Duplicate identifier 'wedgeHandlingChains'.
notification-service.ts(388,11): error TS2393: Duplicate function implementation.
```

Main and this branch **each independently added a wedge-serialisation mechanism**: `enqueueWedgeHandling(taskId, run)` with its chain map on main, `enqueueWedgeHandling(task, wedge)` with its own `wedgeHandlingChains` here. They are two implementations of the same idea, not two versions of one line, so no per-hunk pick can reconcile them — one has to be deleted and its call sites migrated.

That is your refactor to land or drop, not something a merge resolution should decide. My earlier "take yours for hunk 1" was wrong, and I am glad the typecheck caught it rather than a reviewer.

## What still stands

The lane hunk finding is unchanged and is the important one: your side reads `progressedLanes?.hold / ?.wip / ?.complete / ?.archived` — **one id per role**. That is the first-per-role shape flagged as `Major` on #3084, where a board with more than one pre-review working lane silently loses every lane after the first. Main's `resolveProjectColumnsForRoles` returns membership sets and does not. Keep HEAD there; your legacy-id union fallback is still worth folding into it, since main's version has none.

## Suggested path

Rebase it yourself, decide which serialisation survives, migrate the call sites, keep main's membership sets for the lane check, then `pnpm --filter @fusion/engine exec vitest run src/notification`.

Same reason as before for not forcing it: I broke a file earlier today resolving a self-healing conflict quickly, 40 test suites failed, caught pre-push. A visibly conflicted PR costs less than a silently wrong one.

PR #3097, opened 2026-07-31
census: --triage splits the backlog into documented deferrals vs unexamined

gsxdsm:
## The headline number stopped tracking work

I have hand-computed this breakdown every round of the phase to decide what to claim. Making it a first-class mode so nobody else has to, and so "census before/after" in a PR body means something.

On current `main`:

```
COLUMN guards (the backlog):   51

TRIAGE (heuristic, opt-in; changes no count and no exit code)
  documented deferral (flag note within 40 lines): 13
  unexamined:                                      38

  unexamined, by file — this is the list to pick work from:
     21  packages/engine/src/self-healing.ts
      4  packages/engine/src/executor.ts
      2  packages/engine/src/auto-merge-finalization.ts
      1  packages/core/src/eval-signal-collector.ts
      ...
```

**51 reads as a lot of available work. It is not.** 13 carry an explicit reason for staying a literal, and 21 of the remaining 38 are `self-healing.ts` — which has **eight** open PRs on it. What is actually loose is roughly a dozen scattered singletons, most of them in synchronous listeners where the only available resolver is inert.

That gap is not cosmetic; it is causal. A worker told to "claim the largest cluster" reads 51, finds little that is both unclaimed and convertible, and reaches for whatever moves the number. That is exactly how #3051 converted ten `scheduler.ts` guards to `resolveTaskWorkflowIrSync` — inert under PostgreSQL, refuted end-to-end in #3058 — and how five open PRs came to share the same helper.

## Design constraints I held to

- **Opt-in.** No flag, no change. Verified: default output is **byte-identical** to `main` (`diff` clean), and `--json`, `--strict` and `--compare` all still exit 0.
- **Beside the totals, never inside them.** It changes no count and no exit code — the same discipline `traitFallbackCount` already documents two lines above, and for the same reason: a deferred guard is still a guard.
- **Nothing downstream consumes it.** It is a triage aid for choosing work, not a gate.

## Stated limits

Classification is **comment proximity**: an FNXC note within 40 lines above the guard whose text marks a deliberate deferral. It cannot tell a good reason from a bad one, and a note that sits far above its guard reads as unexamined. That is why it is opt-in and why no gate reads it.

## The bug I shipped into my own draft, and what it cost

The first version reported **`documented deferral: 0`** — for a tree I knew had them, because I had counted them by hand that morning. Cause: it referenced an undefined path constant, and my `try/catch` turned the `ReferenceError` into an empty file list, so every proximity window was the empty string and nothing ever matched.

That is the same silent-catch shape I have flagged in review twice this phase. The catch is now gone: **a triage aid that fails to zero is worse than one that throws**, because zero reads as a clean answer rather than a broken instrument. Post-fix it reports 13/38, which matches the hand counts I have been posting all phase.

## Census before / after

```
before:  COLUMN guards (the backlog):   51
after:   COLUMN guards (the backlog):   51
```

Unchanged by construction — this converts nothing. It makes the number interpretable.

## Verification

`test:gate` exit 0 · default census output byte-identical to main · `--json` / `--strict` / `--compare` exit 0 · `pnpm lint` clean. One script; no production file touched.

Related and still open: **#3079** (makes the five inert `resolveMoveFanoutColumnsSync` guards fail the build instead of registering as a census win) and **#3095**.

## Summary by CodeRabbit

- **New Features**
  - Added an optional triage mode to classify column findings as documented deferrals or items requiring review.
  - Added aggregate and top-file triage results to make findings easier to assess.
  - Added guidance for the new command-line option.

- **Bug Fixes**
  - Improved source-reading error handling so failures are reported clearly instead of being silently ignored.

### DIFF
--- a/scripts/lifecycle-column-census.mjs
+++ b/scripts/lifecycle-column-census.mjs
@@ -9,6 +9,7 @@ regression suite that pins each form this census must catch lives in
 Report-only by default:
   node scripts/lifecycle-column-census.mjs            # human table
   node scripts/lifecycle-column-census.mjs --json     # machine-readable
+  node scripts/lifecycle-column-census.mjs --triage   # split the backlog into flagged vs unexamined
   node scripts/lifecycle-column-census.mjs --compare  # cross-check AST vs text classifier
   node scripts/lifecycle-column-census.mjs --strict   # fail if any file DIVERGES from baseline
   node scripts/lifecycle-column-census.mjs --strict --update-baseline   # re-record after lowering it
@@ -39,6 +40,9 @@ The text classifier stays beside it as an independent second implementation — 
 and fails if they disagree, which is the only evidence available that either is right.
 */
 import { censusFiles, summarize } from "./lib/lifecycle-column-census-ast.mjs";
+
+/** Repo root, for reading a finding's source back when `--triage` classifies it. */
+const REPO_ROOT = join(dirname(fileURLToPath(import.meta.url)), "..");
 import {
   censusFiles as censusFilesText,
   summarize as summarizeText,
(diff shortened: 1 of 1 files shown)


gsxdsm:
Fixed. `FLAG_MARKERS` now matches only comment text carrying an `FNXC:` note.

## The finding is real

`FLAG_MARKERS.test(window)` ran against 40 **raw source** lines, so a string literal, an unrelated comment, or this file's own regex text counted as a documented deferral. For a triage aid the direction matters: a false *flagged* hides genuinely unexamined work, which is exactly what the split exists to surface.

Now: extract comment text, require an `FNXC:` note among it, test the markers against that alone.

## My first attempt was wrong and the numbers caught it

I filtered lines on `/^\s*(\/\/|\/\*|\*)/`. Documented deferrals went **13 → 0**.

That is not stricter matching — this codebase's FNXC blocks open with a bare `/*` and their body lines carry **no leading `*`**, so the very line holding `FNXC:` was discarded. It read as a tighter gate while seeing no comments at all. Replaced with block-comment state tracking, which restores 13 / 38.

## What I could NOT demonstrate

I tried to build a discriminating case — a `FLAGGED` marker planted in code above a guard — and **old and new produce identical output** (13 / 38). My injected marker evidently did not land within the 40-line window of a `column`-kind finding, and I ran out of room to construct a better one.

So: the change is strictly more correct (testing comments rather than raw source), it is behaviour-preserving on the current tree, and I verified the *repair* of my own broken first attempt. But I have **not** proven the original false-positive fires in practice. It may be theoretical. Worth knowing before treating this as a confirmed bug fix rather than a hardening.

## Verification

- `census --strict` exit 0; eslint clean
- triage split unchanged: **13 documented / 38 unexamined**
- both before and after the fix, and with a code-planted marker

gsxdsm:
Rebased onto `a7b2a757fa`. The figures in the PR body were measured when the backlog was 51; on current main they read:

```
COLUMN guards (the backlog):   47
  documented deferral: 9
  unexamined:          38
```

The unexamined count has not moved (38), while the documented-deferral count fell 13 → 9 as flagged sites merged or were converted. That is the split doing its job: **the backlog fell by 4 and none of it came from unexamined work.** Reading the headline alone, 51 → 47 looks like four sites of progress against the real remaining queue; the split shows the real remaining queue is unchanged.

That is the whole argument for this mode in one data point, so I would rather post it than let the body's numbers quietly go stale.

PR #3109, opened 2026-07-31
fix(events): carry resolved lanes on task:moved so listener guards stop being inert

gsxdsm:
Removes the **inert-guard class at its source** instead of one call site at a time. Independent of my other branches.

## The problem

`task:moved` listeners run synchronously, so a listener needing a lane answer had to resolve one synchronously — and `resolveTaskWorkflowIrSync` returns the **default** workflow under PostgreSQL, the shipped backend. Every such guard behaved exactly as the literal it replaced, while the census scored it as converted.

**Resolving asynchronously inside the listener is not available**, and that is measured rather than assumed. The scheduler's `snapshotManager.invalidate` is asserted to run in the listener's **synchronous prologue**; putting an await ahead of it produced **3 failures across 21 scheduler suites**.

## The fix

The emitter carries the answer, which removes the dilemma rather than trading one horn for the other. `moves.ts` is already async and already post-commit, so it resolves the moving task's lanes **once** and hands them to every listener. The guard becomes correct **and** the prologue stays synchronous.

This is the file's own recorded preferred fix — *"having the emitter carry the resolved lanes on the event payload so no listener resolves at all"* — now that the audit it was waiting on is done and came back as **one** prologue-dependent consumer, not a class.

## Design choices

- **`lanes` is optional and fail-soft to `undefined`** — "unknown", never "legacy". Some emit paths fire from sync contexts or a cached row mid-teardown. Listeners keep their existing fallback, so those paths are no better than before but **no worse**, and they become the exception rather than the rule.
- **`mergeParkedColumns` overlays only fields the emitter actually resolved**, so a partial payload cannot blank a lane back to a wrong answer.
- **The sync resolver stays** as that fallback. Deleting it would strand the emit paths that cannot resolve.

## Verification

- **Revert-proof and it pins the prologue:** the new case asserts invalidation on a **renamed** hold lane with **no `waitFor`**. Ignoring the payload gives **0 calls**.
- 21 scheduler suites — **361 green**
- self-healing + notification suites — **491 green**
- core moves + the `sync-workflow-ir-callsite-allowlist` ratchet — green
- **`pnpm test:gate` green** (71)

## What it unblocks

`scheduler.ts`'s 10 allow-listed guards now resolve correctly for every move that goes through `moves.ts` — the path real moves take. Those were already absent from the backlog, so **the census number does not move**; what changes is that they now do what the number claimed. `executor.ts`'s 4 remaining sites can follow the same pattern in a separate PR.

### DIFF
--- a/.changeset/task-moved-carries-lanes.md
+++ b/.changeset/task-moved-carries-lanes.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Board renames no longer silently disable scheduler auto-claim invalidation and lane guards.
+category: fix
+dev: `task:moved` payloads now carry emitter-resolved `lanes` (`TaskMoveLanes`); listeners prefer them over the sync IR resolver, which returns the default workflow under PostgreSQL.
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -467,8 +467,8 @@ export { createWorkflowEventBus, getWorkflowEventBus, emitWorkflowLifecycleEvent
 export type { WorkflowEventBus, WorkflowEventSubscriber, WorkflowEventSubscription } from "./workflow-events.js";
 export { findWorkflowEventShapeViolations, isIdsOnlyWorkflowEvent, MAX_ID_VALUE_LENGTH, IMPLEMENTATION_EXITS } from "./types/workflow-events.js";
 export type { WorkflowLifecycleEvent, WorkflowLifecycleEventType, WorkflowLifecycleEventBase, TaskTransitionedEvent, NodeEnteredEvent, NodeCompletedEvent, RunSuspendedEvent, RunResumedEvent, WorkflowEventShapeViolation, ImplementationExit } from "./types/workflow-events.js";
-export { columnHasFlag, columnsWithFlag, declaresAnyLifecycleTrait, resolveArchiveTargetForTask, resolveCompleteColumn, resolveLifecycleColumns, resolveMergeOrchestrationColumn, resolveReboundTarget, resolveReboundTargetForTask, resolveReviewColumns, resolveTaskLifecycleColumns, resolveTerminalColumns, resolveWipTargetForTask } from "./workflow-lifecycle-traits.js";
-export type { LifecycleColumns } from "./workflow-lifecycle-traits.js";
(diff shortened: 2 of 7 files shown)


gsxdsm:
This is the right fix and the one three separate FNXC notes have been asking for — resolving once at the emitter genuinely removes the class rather than one instance. Two gaps before it lands, both measured on this branch.

### 1. One of seven core emitters carries lanes

```
LANES    packages/core/src/task-store/moves.ts:1450
MISSING  packages/core/src/task-store/archive-lifecycle-2.ts:373
MISSING  packages/core/src/task-store/lifecycle-ops.ts:668
MISSING  packages/core/src/task-store/lifecycle-ops.ts:715
MISSING  packages/core/src/task-store/task-update.ts:962
MISSING  packages/core/src/task-store/update-task-deps.ts:502
MISSING  packages/core/src/task-store/task-artifacts-ops.ts:549
```

Your note is explicit that absent `lanes` means "unknown" and the listener keeps its fallback. The problem is what the fallback *is*: `mergeParkedColumns` overlays onto `resolveTaskParkedColumnsSync`, which returns the **default board** for every task under PostgreSQL. So "unknown" does not degrade to cautious — it degrades to *confidently wrong on exactly the boards this PR exists to fix*.

And the six missing emitters are not incidental paths. `archive-lifecycle-2.ts:373` emits the archive move, `task-artifacts-ops.ts:549` emits the move to the complete column, `lifecycle-ops.ts:668` mirrors an archive for polling replicas. Those are **terminal-lane moves** — precisely what `parked.terminal.has(to)` and the review/complete branches guard. After this PR the scheduler resolves correctly for `moves.ts` moves and remains inert for archive and completion moves, while the code now reads as fully resolved at every site.

That is a harder defect to find than the one being fixed, because the mechanism looks right.

### 2. Three forwarders reconstruct the payload

`project-manager.ts`, `remote-node-runtime.ts` and `child-process-runtime.ts` rebuild the event object field-by-field rather than forwarding it, so `lanes` is dropped crossing them. None is touched by this PR. Engine listeners subscribe to the store directly, so this is not live today — but it becomes live the moment a listener behind a runtime boundary is converted, and it will fail silently in the same way. Worth either forwarding the whole payload or adding a note at those three sites.

### Suggestion

Either attach lanes at all seven emitters, or make the absent case *visibly* degraded rather than silently defaulted. A shared `emitTaskMoved(store, payload)` helper that resolves lanes would make it structurally impossible to add an eighth emitter without them — the same "one seam, not seven call sites" argument this PR already makes for listeners.

If it helps: **#3079** (open, mergeable) is the per-guard gate for this class, and it currently attributes `scheduler.ts` 18, `triage.ts` 7, `executor.ts` 4. Once the emitter path is complete those numbers should fall on their own, and a fall that is *not* explained by a real conversion is the signal that a fallback is quietly carrying the load.

Happy to re-run the emitter measurement after changes — it is scripted.

Issue #3082, opened 2026-07-31
The lifecycle-column backlog's remaining engine sites are all sync event listeners — one design change unblocks them


gsxdsm:
Measured at census **83**, after the fleet phase took it from 126.

## The finding

Setting aside `self-healing.ts` (38, actively worked in #3074/#3075/#3078), **every remaining engine site is inside a synchronous event listener**, and they share one blocker:

| site | guards | listener |
|---|---|---|
| `executor.ts:3557, 3581, 3632, 3642` | 4 | `store.on("task:moved", ({…}) => {` |
| `scheduler.ts:1133, 1207` | 2 | `store.on("task:updated", (task) => {` |

Neither listener is `async`, and `scheduler.ts:398` states the constraint in the code:

> *"SYNCHRONOUS on purpose. These run inside `task:moved` / `task:updated` listeners; introducing a new `await` before their existing synchronous work defers everything after it to a microtask and reorders handlers relative to a synchronous emitter."*

So the three available routes are:

1. **Make the listener async** — reorders it against every other synchronous subscriber. This is the hazard #3051 had to audit exhaustively to clear for one listener.
2. **Use the sync resolver** — `resolveTaskWorkflowIrSync` returns the **default workflow IR for every task under PostgreSQL** (proven in `sync-workflow-ir-is-always-default.pg.test.ts`). Converting through it improves the census while resolving nothing.
3. **Have the emitter carry the resolved lanes on the payload**, so no listener resolves at all.

Route 3 is what `scheduler.ts`'s own blocking comment calls "preferred, since it removes the class rather than one instance".

## Why I'm filing rather than doing it

Measured, not assumed:

- **12 emitters** of `task:moved` across `packages/core`, `packages/engine`, `packages/dashboard`
- **8+ subscribers** across the same three packages
- several emitters are themselves in sync contexts (e.g. `lifecycle-ops.ts`'s polling loop) and would need their own resolution hoisted
- the payload is a shared event contract, so the type change fans out

That is a design change with real blast radius, not a fleet conversion. Doing it under "convert with existing role helpers only" would be exactly the overreach the flag-don't-guess rule exists to prevent.

## What it would unlock

[...]

PR #3112, opened 2026-07-31
fix(executor): read task:moved lanes from the payload (executor.ts 4 → 0)

gsxdsm:
**Stacked on #3109** — merge that first; this is its first consumer.

## Census

| Metric | Before | After |
|---|---:|---:|
| COLUMN guards (backlog) | 47 | **43** |
| `executor.ts` | 4 | **0** |

`executor.ts` is off the census top-files list.

## Why these four could not be converted in place

This listener is synchronous and its branches **start execution**, dispose worktrees and release sessions. An await ahead of them defers the `execute()` dispatch itself. The sync IR resolver isn't an option either — it answers with the default workflow under PostgreSQL, so a guard written through it is inert.

Reading the lanes the emitter already resolved costs nothing and leaves the prologue synchronous. This listener is the reason #3109 has the shape it does.

## The archive branch is the one with teeth

`to === "archived"` matched nothing on a board with a renamed terminal lane, so **archiving never released the task's active-session registry entry** — and that entry is what blocks a **successor** task from acquiring the same path. Not cosmetic: the next task wanting that path fails to register.

## Verification

- **Revert-proof:** the new case drives a `shipped` terminal lane (matching no legacy id) and asserts the release. Reverting the branch to the literal leaves the entry held — `expected [Array(1)] to have a length of 0`.
- 43 executor suites — **483 green**
- **`pnpm test:gate` green**; eslint clean

## Note on shape

Lanes are read as **single ids, not sets**, because each branch here is a lane-identity test on one column — exactly what the literals were. Widening to membership would change behaviour, not just vocabulary. Fail-soft to the legacy ids when the emit path could not resolve, matching every other consumer of this payload.

### DIFF
--- a/packages/engine/src/__tests__/executor-archive-releases-active-session.test.ts
+++ b/packages/engine/src/__tests__/executor-archive-releases-active-session.test.ts
@@ -72,6 +72,35 @@ describe("archiving a task releases its active-session registry entries (FN-7717
     ).not.toThrow();
   });
 
+  /*
+  FNXC:WorkflowResolvedColumns 2026-07-31-21:30 (fleet):
+  The archive branch keyed on `to === "archived"`. On a board whose terminal lane is renamed it matched
+  nothing, so archiving never released the task's active-session entry — and the registry entry is what
+  blocks a SUCCESSOR task from acquiring the same path. The leak is therefore not cosmetic: the next
+  task to want that path fails to register.
+
+  Lanes come from the emitter, so this drives the listener exactly as `moves.ts` now emits. The literal
+  is deliberately absent from the payload's lane: `shipped` matches no legacy id, so the branch fires
+  only if the payload is actually consulted.
+  */
+  it("releases the session when archiving into a RENAMED terminal lane", async () => {
+    const { executor, store } = makeExecutor();
+    (executor as any).setActiveWorkflowStepSession("TASK-R", {}, SHARED_ROOT);
+    const [heldPath] = activeSessionRegistry.pathsForTask("TASK-R");
+    expect(heldPath).toBeDefined();
+
+    store.emit("task:moved", {
+      task: makeTask("TASK-R"),
+      from: "signoff",
+      to: "shipped",
+      source: "user",
+      lanes: { hold: "backlog", wip: "building", archived: "shipped" },
+    });
+
+    await (executor as any).pendingTaskDisposals.get("TASK-R");
+    expect(activeSessionRegistry.pathsForTask("TASK-R")).toHaveLength(0);
+  });
+
   it("releases executor and step-session surfaces archived from planning/todo columns", async () => {
     const { executor, store } = makeExecutor();
 
--- a/packages/engine/src/__tests__/executor-archive-releases-active-session.test.ts
+++ b/packages/engine/src/__tests__/executor-archive-releases-active-session.test.ts
@@ -101,6 +101,67 @@ describe("archiving a task releases its active-session registry entries (FN-7717
     expect(activeSessionRegistry.pathsForTask("TASK-R")).toHaveLength(0);
   });
 
+  /*
+  FNXC:WorkflowResolvedColumns 2026-07-31-23:55 (fleet):
+  THE OTHER TWO CONVERTED READS, which the archive case above does NOT cover.
+
+  Mutation-testing the three converted reads together produced only ONE failure — the archive case.
+  `wipLane` and `holdLane` were converted with nothing that fails when they regress, so on this
+  program's own standard they shipped unproven. These two cases close that.
+
+  They target the LAST branch of the if/else-if chain (`from === wipLane`), which is only reached when
+  the archive and backward-out-of-planning branches both decline. `isBackwardMoveOutOfPlanning` is
+  stubbed false so the test pins the lane comparison rather than that predicate's own logic — without
+  the stub a change in planner-lane tracking could silently route these moves elsewhere and leave the
+  assertions passing for the wrong reason.
+  */
+  it("aborts in-flight work when a card leaves a RENAMED wip lane", async () => {
+    const { executor, store } = makeExecutor();
+    vi.spyOn(executor as any, "isBackwardMoveOutOfPlanning").mockReturnValue(false);
+    const abort = vi
+      .spyOn(executor as any, "awaitAbortInFlightTaskWork")
+      .mockResolvedValue(undefined);
+
+    /* `building` matches no legacy id, so the branch fires only if the payload's wip lane is read. */
+    store.emit("task:moved", {
+      task: makeTask("TASK-W"),
+      from: "building",
+      to: "checking",
+      source: "engine",
+      lanes: { hold: "backlog", wip: "building", archived: "shipped" },
+    });
+
+    await (executor as any).pendingTaskDisposals.get("TASK-W");
+    expect(abort).toHaveBeenCalledTimes(1);
+  });
+
+  /*
+  `userCanceled` is the one place `holdLane` changes an OUTCOME rather than just a branch: a user
+  dragging a card from the wip lane back to the board's hold lane is a cancel, and anything else is
+  not. Keyed on the literal `"todo"`, a renamed hold lane made every such drag read as NOT
+  user-canceled — the executor then treats the abort as an engine rebound and the task is eligible to
+  be picked straight back up, which is the opposite of what the operator just asked for.
+  */
+  it("marks a user drag into a RENAMED hold lane as user-canceled", async () => {
+    const { executor, store } = makeExecutor();
+    vi.spyOn(executor as any, "isBackwardMoveOutOfPlanning").mockReturnValue(false);
+    const abort = vi
+      .spyOn(executor as any, "awaitAbortInFlightTaskWork")
+      .mockResolvedValue(undefined);
+
+    store.emit("task:moved", {
+      task: makeTask("TASK-H"),
+      from: "building",
+      to: "backlog",
+      source: "user",
+      lanes: { hold: "backlog", wip: "building", archived: "shipped" },
+    });
+
+    await (executor as any).pendingTaskDisposals.get("TASK-H");
+    expect(abort).toHaveBeenCalledTimes(1);
+    expect(abort.mock.calls[0]?.[2]).toMatchObject({ userCanceled: true });
+  });
+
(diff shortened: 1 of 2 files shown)


gsxdsm:
Re-measured on this branch: the emitter split is still **1 of 7** (`moves.ts:1450` carries `lanes`; the other six do not). But I need to correct part of what I wrote on #3109, and the correction sharpens the finding rather than withdrawing it.

**The archive path is NOT a defect — I was wrong to headline it.** `archive-lifecycle-2.ts` writes `task.column = "archived"` (literal), emits `to: "archived"` (literal), and this PR's executor listener falls back to `archivedLane = "archived"`. All three agree, so `to === archivedLane` holds on every board. Your `executor-archive-releases-active-session` case is sound.

**The real gap is the inverse shape, and two emitters have it:**

```
task-artifacts-ops.ts:549   emit to: completeColumn   ← RESOLVED lane, no lanes attached
update-task-deps.ts:502     emit to: task.column      ← the row's ACTUAL lane, no lanes attached
```

These send a *renamed* column id while attaching no `lanes`, so a listener falls back to the default-board literals and compares e.g. `"shipped"` against `"done"`. It never matches, and nothing errors — the branch simply stops firing.

That is worse than a literal-vs-literal mismatch because the two sides disagree about vocabulary in opposite directions: the emitter has resolved the board, the listener has not, and the payload that would have reconciled them is the one field that is optional.

So the concern from #3109 stands, just relocated: it is not "the archive path breaks", it is **"an emitter that has already resolved a lane must not drop it"**. Those two sites are exactly where a resolution exists and is thrown away.

Concrete suggestion, unchanged: route every emit through one `emitTaskMoved(store, payload)` that attaches lanes, so an eighth emitter cannot be added without them. With seven call sites and an optional field, "which emitters carry lanes" becomes a fact nobody can hold in their head — I only found this by scripting it, and I got the headline case wrong on the first pass reading it by eye.

Happy to re-run the emitter measurement on request; it is scripted and takes seconds.

gsxdsm:
**Rebased onto main and pushed** (`fb4cfb64`). Branch was not checked out in any worktree.

## The conflict was a real disagreement, and your side wins on the merits

One hunk in `executor.ts`. Main had just landed a long **FLAGGED AND LEFT COUNTED** note on these same four guards, arguing they must stay literal:

> THE OBVIOUS FIX IS INERT … `getTaskWorkflowSelectionImpl` returns `undefined` unconditionally under PostgreSQL … even with a selection, the CUSTOM-workflow branch loads its IR through `store.db`, whose implementation is an unconditional throw … A renamed lane IS a custom workflow, so (2) alone is decisive: the sync path can never serve this listener's case.

That reasoning is sound — **against the mechanism it names.** It says do not convert with `resolveTaskWorkflowIrSync` / `resolvePlannerLanes`, and it is right that both are inert here.

Your change uses neither. It reads `lanes` **off the `task:moved` payload**:

```ts
store.on("task:moved", ({ task, from, to, source, lanes }) => {
```

The emitter supplies the answer, so there is no sync IR read to be inert and no await to reorder the handler against other subscribers. Both blockers in that note are sidestepped rather than ignored — which is why I resolved to your side and deleted the note rather than keeping a warning that no longer describes the code.

Worth stating plainly since the note was emphatic: **it was not wrong, it was scoped to a different fix.** Anyone re-reading this later should not conclude the objection was overruled.

## Verification

- `executor-archive-releases-active-session` + `scheduler-auto-claim-invalidation` — **28 pass**
- `tsc --noEmit` (engine) — **0 errors**
- `check:inert-flag-seams` — exit 0, so the new payload read is not itself an inert seam
- `census --strict` — exit 0 after re-record; **executor.ts drops out of `byFile` entirely**, repo total **45 → 41**

The census re-record was required, not optional: `--strict` failed before it. Amended into your commit.

gsxdsm:
**Rebased onto current `main`** (#3109 merged), and resolved a real conflict with #3104's note rather than overwriting it.

#3104 landed a detailed *"FLAGGED AND LEFT COUNTED — do NOT convert"* analysis on these exact four guards. **That analysis was correct**, and I kept it in full. What changed is the premise: it concludes

> UNBLOCKING therefore needs the async resolver reachable from a SYNCHRONOUS consumer

which is exactly what #3109 now delivers — the emitter resolves the lanes and puts them on the payload, so this listener reads an async-resolved answer synchronously. Both of its stated blockers are *satisfied*, not argued away:

- **(A) in-tick `pendingTaskDisposals`** — there is **no deferral at all**, no await and no microtask, so the write and the fast-bounce read stay in the same tick and the FN-5256 race stays closed.
- **(B) entangled if/else-if chain** — all four guards convert **together** in this one change, which is what the chain requires.

I appended the resolution beneath their note rather than deleting it, and kept their warning about `resolveTaskWorkflowIrSync` — that path is still inert and still not the fix.

**Also re-records the inert-sync-lane baseline: 20 → 5.** #3109 dropped fifteen guards out of the inert set and this PR converts the executor's four; leaving the baseline at 20 would let those regress without failing the gate. The remaining five are in `scheduler.ts`, on emit paths that don't carry lanes.

Verification after rebase: 43 executor suites (**483 green**), the sync-resolver call-site allow-list, `sync-workflow-ir-second-blocker`, and **`pnpm test:gate`** all green. Census: **executor.ts 4 → 0**, backlog 45 → 41.

gsxdsm:
Rebased onto `origin/main` (77d11a0a9a) and pushed — was CONFLICTING, now clean.

## What the conflict was, and why it is not a mechanical resolution

Main landed #3104 while this sat open. That PR added a long `FNXC` block over this exact listener concluding **"FLAGGED AND LEFT COUNTED — do NOT convert"**, and this PR converts those same four literals. So the conflict was two contradictory verdicts on one hunk, not two spellings of one change.

Both are right about their own claim, and the block is what proves it:

- #3104's argument is that **no resolver can be *called* here** — `task:moved` is emitted synchronously, an `await` reorders this handler against every other subscriber, and the sync IR path returns the DEFAULT workflow under PostgreSQL for two independent reasons.
- This PR does not call a resolver. The **emitter** resolves once (already async) and the lanes ride on the payload, so branch selection stays in-tick. #3104's objections are all about in-tick resolution, so none of them bind.

I kept the block rather than deleting it — it is the evidence that rules out the other two routes — and appended the resolution beneath it. The `git rebase` also **skipped the events-side commit as already applied**: that half (`lanes` on the payload, `moves.ts`) landed on main separately, which is why only the executor half remained.

## Measured

| | before | after |
|---|---|---|
| census total | 45 | **41** |
| `executor.ts` column guards | 4 | **0** |

- `pnpm --filter @fusion/engine exec tsc --noEmit` — **exit 0, 0 errors**
- `executor-archive-releases-active-session.test.ts` — **11 passed**
- `sync-workflow-ir-second-blocker.test.ts` (core) — **4 passed**, still green: this PR does not weaken its claim, it routes around it
- `check-inert-flag-seams`, `check-fnxc-future-dates`, `check-inert-sync-lane-conversions` — all **exit 0**

## Differential proof, and the gap it exposes

Reverting the three converted reads to their literals: **1 failed | 10 passed**. Restored: **11 passed**. So the test does fail on the original defect.

**But only one of the three does.** The failing assertion is the archived-lane session release; the `wip` and `hold` reads are converted with no test that fails when they regress. Stating it rather than letting "11 passed" imply otherwise — by this program's own standard those two are currently unproven, and the if/else-if chain means they had to convert together (#3104's point B), so the coverage is narrower than the change.

## One correction, and a follow-up I am not bundling

My first draft of the resolution comment claimed the payload field is **required**, so a forgetful emitter would fail typecheck. **That is false** — it is `lanes?: TaskMoveLanes` in `store.ts:163`, and the fallback here is the legacy literal. I checked before shipping the claim and rewrote it.

The honest state is in the comment: a `task:moved` emitted without `lanes` leaves these four guards exactly as inert as before, silently, on a renamed board. Strictly better than the flagged state (one path instead of every path, and `moves.ts` does pass them) but not a compile-time guarantee.

Making `lanes` required is the right end state and is deliberately **not** in this PR — it retypes every emit site, which is a mechanical change that should not ride with a behavior change. Tracked in #3082 along with the emitter-side work; the alternative is a gate asserting every emit site supplies it.

gsxdsm:
## Coverage gap from the comment above is now closed

I reported that only one of the three converted reads was differentially covered. Fixed in `0a5e5910bd` rather than left as a caveat.

| | mutated (reads reverted to literals) | restored |
|---|---|---|
| before this commit | **1 failed** \| 10 passed | 11 passed |
| after | **3 failed** \| 10 passed | **13 passed** |

All three failures are the renamed-lane cases, one per converted read:

```
FAIL  releases the session when archiving into a RENAMED terminal lane
FAIL  aborts in-flight work when a card leaves a RENAMED wip lane
FAIL  marks a user drag into a RENAMED hold lane as user-canceled
```

The hold case is the one worth calling out as a user-visible defect in its own right, not just a vocabulary tidy. `userCanceled` is the only place `holdLane` changes an **outcome** rather than a branch:

```ts
userCanceled: source === "user" && to === holdLane
```

Keyed on the literal `"todo"`, a **user dragging a card from the wip lane back to a renamed hold lane read as NOT user-canceled**. The executor then treats the abort as an engine rebound, which makes the card eligible to be picked straight back up — the operator drags work out of progress and the engine puts it back. On a default-vocabulary board this never reproduces, which is why it survived.

Both new cases stub `isBackwardMoveOutOfPlanning` to false. That branch sits ahead of `from === wipLane` in the chain, and without the stub a future change to planner-lane tracking could reroute these moves and leave the assertions passing for the wrong reason — the failure mode this program keeps hitting, where the instrument stops pointing at the thing it names.

gsxdsm:
**Superseded by #3118** — same four literals, same payload-lanes fix, arrived at independently. That PR's author wrote the #3104 flag this conversion removes, so the removal belongs there; I have asked them to take the three regression cases from `0a5e5910bd` before this closes, since #3118 does not currently have them and mutation testing showed two of the three converted reads are otherwise unproven.

Leaving this open only until that lands or they decline — closing first would drop the coverage on the floor. No objection to whoever merges #3118 closing this alongside it.

Flagging the collision itself, since it is the second instance today: #3094 records three overlapping conversions landing on `self-healing.ts` from different lanes, each forcing a full rebuild. Same shape here — two workers converting the same listener on the same afternoon. The census tells everyone where the work is but nothing records who picked it up, so the largest clusters attract duplicate claims. Worth a lightweight claim signal if this phase continues.

PR #3115, opened 2026-07-31
test(engine): pin #3047's archive-sweep conversion — measured uncovered (825 tests passed against the reverted fix)

gsxdsm:
Not a conversion — the fleet's conversions are landing faster than their coverage, and this is the audit that shows which ones actually have any.

## Method

For each of today's fleet commits to `self-healing.ts`: revert that single commit, re-run the file's suites, see whether anything fails. If nothing fails, the conversion has no regression protection and the "N tests passed" cited on its PR was measuring something else.

| commit | reverted | verdict |
|---|---|---|
| #3075 pause-abort recovery | suites **fail** | covered |
| **#3047 archiveStaleDoneTasks** | **825 tests all pass** | **uncovered** |
| #3078 (mine) | 204 tests all passed | was uncovered — closed by #3090, #3102 |

Every fixture in the `archiveStaleDoneTasks` describe block uses the id `done`, where the literal is correct, so none of them could see the conversion at all.

## What the literal cost

The sweep's dependent scan skips tasks in terminal lanes. On a renamed board **nothing matched `done`/`archived`, so every task read as active** — which means every archive candidate looked like it had active dependents, and the sweep archived **nothing**. The board quietly stops auto-archiving: no error, no log line, no failing test.

## The case covers both halves of #3047

- a stale card in a **renamed complete lane** is archived — the `complete` role
- a card whose dependent is still live in the **renamed wip lane** is **not** archived, and a dependent already in the **renamed archive lane** does not count as live — the `terminal` role

That second assertion is the one that matters: it stops the fix from degenerating into "archive everything", which is the failure mode a one-sided test would miss.

## Measured

**413 pass** on current main; reverting #3047 fails **exactly this test**.

## Remaining audit

I have now audited 3 of ~13 fleet conversions to this file this way. The method is cheap (one revert, one 17s suite run) and I will keep working through the rest unless someone else picks it up. #3049 could not be auto-reverted — later commits overlap its hunks — so it needs a manual read rather than a mechanical revert.

## Verification

`self-healing.test.ts` **413 passed** · `pnpm test:gate` 13 + 161 + 487 + 71 · lint — green.

### DIFF
--- a/packages/engine/src/__tests__/self-healing.test.ts
+++ b/packages/engine/src/__tests__/self-healing.test.ts
@@ -2310,6 +2310,53 @@ describe("SelfHealingManager", () => {
       expect(store.archiveTaskAndCleanup).not.toHaveBeenCalledWith("FN-002");
     });
 
+    /*
+    FNXC:WorkflowResolvedColumns 2026-07-31-20:10:
+    #3047 converted this sweep's two lane guards to the COMPLETE and TERMINAL roles and merged with no
+    case that could see the difference — measured: reverting that commit leaves all 825 self-healing
+    tests passing, because every fixture above uses the id `done`, where the literal is correct.
+
+    What the literal cost on a renamed board: the dependent scan treated EVERY task as active (nothing
+    matched `done`/`archived`), so every candidate looked like it had active dependents and the sweep
+    archived nothing. A silent no-op — the board just quietly stops auto-archiving.
+    */
+    it("archives a stale card in a RENAMED complete lane, and skips one whose dependent is still live", async () => {
+      vi.setSystemTime(new Date("2026-01-04T00:00:00.000Z"));
+      (store.getSettings as ReturnType<typeof vi.fn>).mockResolvedValue({
+        autoArchiveDoneTasksEnabled: true,
+        autoArchiveDoneAfterMs: 24 * 60 * 60 * 1000,
+        doneAutoArchiveDays: 0,
+      } as unknown as Settings);
+      (store.listWorkflowDefinitions as ReturnType<typeof vi.fn> | undefined)?.mockResolvedValue?.([]);
+      (store as unknown as { listWorkflowDefinitions: unknown }).listWorkflowDefinitions = vi.fn(async () => [{
+        ir: {
+          version: "v2",
+          id: "custom:renamed",
+          nodes: [],
+          edges: [],
+          columns: [
+            { id: "building", name: "building", traits: [{ trait: "wip", config: { limitSetting: "maxConcurrent" } }] },
+            { id: "shipped", name: "shipped", traits: [{ trait: "complete" }] },
+            { id: "vault", name: "vault", traits: [{ trait: "archived" }] },
+          ],
+        },
+      }]);
+      (store.listTasks as ReturnType<typeof vi.fn>).mockResolvedValue([
+        { id: "FN-OLD", column: "shipped", columnMovedAt: "2026-01-02T00:00:00.000Z", updatedAt: "2026-01-02T00:00:00.000Z" },
+        { id: "FN-BLOCKED", column: "shipped", columnMovedAt: "2026-01-02T00:00:00.000Z", updatedAt: "2026-01-02T00:00:00.000Z" },
+        /* A LIVE dependent in the renamed wip lane must still protect its blocker from archiving. */
+        { id: "FN-LIVE", column: "building", dependencies: ["FN-BLOCKED"], updatedAt: "2026-01-03T00:00:00.000Z" },
+        /* Already in the renamed archive lane: terminal, so neither a candidate nor an active dependent. */
+        { id: "FN-FILED", column: "vault", dependencies: ["FN-OLD"], updatedAt: "2026-01-03T00:00:00.000Z" },
+      ]);
+
+      const result = await manager.archiveStaleDoneTasks();
+
+      expect(result).toBe(1);
+      expect(store.archiveTaskAndCleanup).toHaveBeenCalledWith("FN-OLD");
+      expect(store.archiveTaskAndCleanup).not.toHaveBeenCalledWith("FN-BLOCKED");
+    });
+
     it("uses doneAutoArchiveDays threshold and logs task age", async () => {
       vi.setSystemTime(new Date("2026-03-01T00:00:00.000Z"));
       (store.getSettings as ReturnType<typeof vi.fn>).mockResolvedValue({


gsxdsm:
## Coverage map for `self-healing.ts`'s converted sweeps — measured, not estimated

Extending the audit method from this PR to the file as a whole. For each resolved lane set, I blinded **that one resolver** back to its legacy ids (`TERMINAL_ROLES` → `["done","archived"]`, `["countsTowardWip"]` → `["in-progress"]`, and so on) and ran the file's suites. If nothing fails, no test can tell that conversion from the literal it replaced.

| line | sweep | blinded → |
|---|---|---|
| 2897 | `detectStalledCards` (wip) | **COVERED** |
| 3371 | stale-merge review lane | **COVERED** |
| 2813 / 2825 | `archiveStaleDoneTasks` | UNCOVERED — closed by this PR |
| 3666 | `reclaimPrConflictForTask` | **UNCOVERED** |
| 3845 | `reclaimSelfOwnedBranchConflicts` | **UNCOVERED** |
| 4562 | `reclaimStaleActiveBranches` | **UNCOVERED** |
| 5136 | `reconcileInReviewBranchRebind` | **UNCOVERED** |
| 5486 | `autoReboundPausedScopeDecayDetailed` | **UNCOVERED** |

**7 of 9 sampled conversions have no test that notices when they regress.** Three of those seven are mine, from #3078 — I am not reporting someone else's gap here, I am reporting the file's, and mine are in it.

### Why this matters more than the census number

A converted sweep with no covering test is not "done" in any durable sense: the next refactor, merge conflict, or mechanical pass can put the literal back and every gate stays green. That is exactly how #2951 dropped the stall-signal wiring, and how #3078 shipped hollow.

The census counts **comparisons**. It cannot tell a conversion that works from one that has been silently undone. These two measurements answer different questions, and only one of them is being tracked.

### Cost

One revert or blinding, one 17-second suite run, one line of output. I have now run it 12 times. It is cheaper than writing the conversion was.

### What I would do with this

Tests for the five named uncovered sweeps, highest-risk first — `reclaimStaleActiveBranches` and `reconcileInReviewBranchRebind` both act on git branches of live cards, which is where a wrong lane answer does real damage. I will keep going in that order unless the fleet wants the effort elsewhere; the branch-reclaim sweeps need a git-shelling harness, so they are slower to pin than the pure-task ones.

gsxdsm:
## Correction: the sweep names in my coverage map above are wrong

The COVERED/UNCOVERED verdicts were real measurements. **The sweep names attached to them were not.**

I blinded resolvers by line number, then resolved line → method name in a later step. Between those two steps `origin/main` moved — the fleet is merging into this file continuously, and #3094 alone shifted it by hundreds of lines. So the names were looked up in a tree that no longer matched the one I measured. Checking now, line 2813 is a maintenance-registration line and 5486 is inside `reconcileTaskWorktreeMetadata`, not the sweeps I labelled them.

Concretely: the row *"5486 `autoReboundPausedScopeDecayDetailed` UNCOVERED"* names a sweep I did not measure. Treat every name in that table as unverified. The counts (7 of 9 uncovered) describe real resolvers, just not necessarily the ones I named.

That is the same class of error I have been flagging in other people's work all day — a number that is right attached to a claim that is not — and it is worse here because a wrong name sends the next person to the wrong sweep.

## Redoing it properly

Keyed by **resolver variable name** instead of line number, so drift cannot re-corrupt the labels, against the current tree (`ec2921b958`, after #3094 landed):

- **64 resolved lane sets** in `self-healing.ts`
- 30 measured so far: **13 covered, 17 uncovered**
- the remaining 34 are running now

Each run blinds exactly one resolver back to its legacy ids (`TERMINAL_ROLES` → `["done","archived"]`, `["countsTowardWip"]` → `["in-progress"]`, …) and re-runs the file's 825 tests. If nothing fails, no test can distinguish that conversion from the literal it replaced.

I will post the completed table with verified names. Until then the only number worth quoting from me is the ratio, and only as provisional.

gsxdsm:
## Verified coverage map — complete, names checked this time

64 resolved lane sets in `self-healing.ts`, each blinded individually back to its legacy ids, each followed by the file's 825 tests. Keyed by **resolver variable**, so the line drift that corrupted my first attempt cannot recur. Tree: `ec2921b958` (after #3094).

**38 covered · 26 uncovered.**

`archiveStaleDoneTasks` appears twice below and is closed by this PR; the other 24 entries are open.

| sweep | resolver | role |
|---|---|---|
| `reconcileTaskWorktreeMetadata` | terminal / wip / review — **all three** | clears `worktree`/`branch` metadata; the FN-5256 guard is the only thing between that and a running shell |
| `recoverAgentsRunningOnInactiveTasks` | terminal, hold∪intake | |
| `reclaimStaleActiveBranches` | archived | deletes branches |
| `reclaimSelfOwnedBranchConflicts` | hold, review | |
| `reclaimPrConflictForTask` | wip | |
| `reconcileInReviewBranchRebind` | review | rebinds branches of live cards |
| `reconcileDependencyBlockingLeases` | wip, hold | |
| `reconcileCompletedTask` / `reconcileCompletedBlockedTasks` | review, hold | |
| `reconcileInReviewUnmetDependencies` | review | |
| `recoverApprovedStrandedAiMergeCommit` | review | |
| `recoverOrphanOnlyScopeViolations` | review | |
| `recoverMergedReviewTasks` | hold | |
| `recoverForeignOnlyContaminatedInReviewTasks` | wip | |
| `recoverStarvedRefinementTriageTasks` | hold∪intake | |
| `reclaimPhantomWorkspaceLandLeases` | complete | |
| `cleanupStaleTempMergeWorktrees` | terminal | |
| `autoReboundPausedScopeDecayDetailed` | wip | |
| `detectStalledCards` | terminal | |

### Reading this fairly

**38 of 64 ARE covered.** The fleet's work on this file is substantially protected, and I want that stated first because my earlier partial sample (7-of-9 uncovered) made it look far worse than it is. Sampling nine and generalising was the error; measuring all 64 is the fix.

The uncovered 26 are not evenly risky. Three of them sit in `reconcileTaskWorktreeMetadata`, which nulls `worktree`/`branch` on a live row — the sweep whose own header cites FN-5256 as the reason its liveness guard exists. That is where I would put the next test, followed by the two branch-reclaim sweeps.

### What this measures that the census cannot

The census counts comparisons. It cannot tell a conversion that works from one that has been silently reverted by a later merge — which is how #2951 dropped the stall-signal wiring and how #3078 shipped hollow. Cost: one blinding, one 17-second run. I have now run it 64 times; the whole sweep is ~18 minutes.

Anyone can re-run it: blind one resolver to its legacy set, run `vitest run src/__tests__/self-healing`, and if it stays green that conversion has no test behind it.

PR #3117, opened 2026-07-31
fix(gate): the inert-sync-lane ratchet had two free slots — my own drop went unrecorded

gsxdsm:
Found by merging all five of my open branches together and running the gates — a check none of them gets individually. The finding turned out not to be about those branches at all: **`main` itself carries a baseline of 20 against a real count of 18.**

## It is mine

#3065 replaced three `to === parked.complete || to === parked.archived` guards with `parked.terminal.has(to)` and took the count **20 → 18**. The gate *warned* and exited **0**, so nothing failed, I did not re-record, and the allowance stayed high.

Bisected to be sure rather than inferred:

| commit | count |
|---|---|
| #3051 (`scheduler.ts 12 → 2`) | 20 |
| **#3065 (mine)** | **18** |
| #3100 (mine, comment-only) | 18 |

## The consequence is concrete

The gate whose entire purpose is to stop the inert-conversion class growing **would have accepted two new inert conversions.**

Verified, not reasoned: with the baseline at 20, adding one new `parked.wip` comparison to `scheduler.ts` still passed. With the baseline corrected to 18, the same edit fails.

This is the exact failure mode I called out earlier in this program — a fix landing without its ledger update — committed by me. The lenient exit code is why it stayed invisible for a day.

## Two changes

1. **Re-record the baseline 20 → 18.** Restores the ratchet today.
2. **An unrecorded drop now exits 1**, matching the sibling `check-lane-wiring.mjs`. Restores it tomorrow.

A ratchet that only tightens on request does not ratchet. Two gates guarding the same program should not disagree about how seriously they take their own ledger — `check-lane-wiring` already fails here and explains why; this is the same rule for the same reason.

## Tests, because the exit code *is* the contract

Nothing covered this script's exit codes. The new cases drive it by **running the script** against a temporarily swapped baseline rather than importing a helper — a version that printed exactly the right warning and still exited 0 would satisfy any assertion about its output.

- rise → exits 1, names the file
- unrecorded drop → exits 1, **and the message names `--update-baseline`**, because a failure that does not name its fix is noise to whoever hits it
- committed baseline matches the tree → exits 0
- **anti-vacuity**: the scan still finds real guards, so the two mutated-baseline cases cannot pass against a gate that stopped reading source and merely compares a number to itself

The baseline file is restored in a `finally`, so a failing test cannot leave the repo's real ledger modified.

## Measured

- 4 new cases pass (`node --test`).
- Gate exits **1** before the re-record, **0** after, and **1** again when a synthetic new inert conversion is added.
- `check-lane-wiring`, census `--strict`, `check-fnxc-future-dates` all clean.

## Census

No movement — this is a gate fix, not a conversion.

## Summary by CodeRabbit

- **Bug Fixes**
  - The sync-lane conversion check now flags unexpected decreases in guard counts as failures.
  - Diagnostics clearly show baseline and current totals and instruct users to update the recorded baseline.
  - Existing checks continue to detect unexpected increases and scan source files as expected.

- **Tests**
  - Added coverage for matching counts, unrecorded increases, unrecorded decreases, and source-file scanning.
  - Test scenarios restore temporary baseline changes automatically.

### DIFF
--- a/scripts/__tests__/check-inert-sync-lane-conversions.test.mjs
+++ b/scripts/__tests__/check-inert-sync-lane-conversions.test.mjs
@@ -1,0 +1,129 @@
+// Exit-code coverage for the inert-sync-lane ratchet: a RISE fails, and an unrecorded DROP fails too.
+/*
+FNXC:WorkflowResolvedColumns 2026-07-31-23:45:
+A RATCHET THAT ONLY TIGHTENS ON REQUEST DOES NOT RATCHET, and nothing tested this one's exit codes.
+
+The rise path was always an error. The DROP path only warned and exited 0, which leaves the
+allowance stale-high and the gate slack by exactly the size of the drop.
+
+Measured, and the example is mine: #3065 replaced three
+`to === parked.complete || to === parked.archived` guards with `parked.terminal.has(to)` and took the
+count 20 -> 18. I did not re-record, nothing failed, and `main` then carried a baseline of 20 against
+a real count of 18 — two free slots in the gate whose whole purpose is to stop this class growing.
+Two new inert conversions would have passed.
+
+Its sibling `check-lane-wiring.mjs` already exits 1 on an unrecorded drop and says why. These two
+gates guard the same program and must not disagree about how seriously they take their own ledger.
+
+Driven by RUNNING the script against a temp baseline, not by importing a helper: the exit code IS the
+contract — a version that printed the right warning and still exited 0 would satisfy any assertion
+about its output.
+*/
+import test from "node:test";
+import assert from "node:assert/strict";
+import { spawnSync } from "node:child_process";
+import { mkdtempSync, writeFileSync, readFileSync, copyFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join, resolve } from "node:path";
+
+const REPO_ROOT = resolve(import.meta.dirname, "../..");
+const SCRIPT = join(REPO_ROOT, "scripts/check-inert-sync-lane-conversions.mjs");
+const BASELINE = join(REPO_ROOT, "scripts/lib/inert-sync-lane-baseline.json");
+
+function runGate() {
+  return spawnSync(process.execPath, [SCRIPT], { cwd: REPO_ROOT, encoding: "utf8" });
+}
+
+/** The gate's own view of the tree, so cases start from reality rather than the committed file. */
+function liveCounts() {
+  const backup = join(mkdtempSync(join(tmpdir(), "inert-gate-")), "baseline.json");
+  copyFileSync(BASELINE, backup);
+  try {
+    spawnSync(process.execPath, [SCRIPT, "--update-baseline"], { cwd: REPO_ROOT, encoding: "utf8" });
+    return JSON.parse(readFileSync(BASELINE, "utf8"));
+  } finally {
+    copyFileSync(backup, BASELINE);
+  }
+}
+
+/**
+ * Swaps in a baseline derived from the LIVE counts, runs the gate, restores the real file.
+ * Deriving from live counts is what keeps these cases independent of whatever the committed
+ * baseline currently says.
+ */
+function withBaseline(mutate, fn) {
+  const backup = join(mkdtempSync(join(tmpdir(), "inert-gate-")), "baseline.json");
+  copyFileSync(BASELINE, backup);
+  try {
+    writeFileSync(BASELINE, JSON.stringify(mutate(liveCounts()), null, 2) + "\n");
+    return fn();
+  } finally {
+    copyFileSync(backup, BASELINE);
+  }
+}
+
+/*
+FNXC:WorkflowResolvedColumns 2026-07-31-23:50:
+DELIBERATELY NOT "the committed baseline matches the tree". That asserts a property of the REPO at a
+moment, not of this gate, so it goes red whenever someone else lands an unrecorded change — which is
+exactly what happened here: #3114 took `triage.ts` 7 -> 8 and my first version of this file failed for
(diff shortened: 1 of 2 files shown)


gsxdsm:
This is correct and it is my bug. I wrote the drop path as a `console.log` followed by an unconditional `process.exit(0)`, so the warning was decorative — and you are right that the sibling `check-lane-wiring.mjs` already exits 1 on an unrecorded drop, so the two gates disagreed about their own ledger for no reason.

The sharper part is that I walked straight past it. I hit `total fell 20 -> 18. Re-record` on `main` two rounds ago, treated it as the dangerous case, verified the cause was #3065 densifying three comparison pairs into `.has` calls with no guard retired, and re-recorded deliberately. I wrote up the verification obligation in the script header. What I did not do was notice that the check had let the drop through **silently in the first place**, leaving `main` at a baseline of 20 against a real 18 — two free slots, exactly as you describe. I checked the number and missed the mechanism.

Testing it by **running the script against a temp baseline** is the right call and better than what I had, which was no exit-code coverage at all. Your point that "a version that printed the right warning and still exited 0 would satisfy any assertion about its output" is precisely the failure mode — that version was the one shipped.

**Merge-order note:** this and **#3079** both modify `scripts/check-inert-sync-lane-conversions.mjs` and `scripts/lib/inert-sync-lane-baseline.json`, so they will conflict textually.

Suggest **#3117 first**. It fixes the mechanism's integrity and is self-contained; #3079 widens detection (per-file → repo-wide source collection, which surfaces `triage.ts` 7 and `executor.ts` 4 that a same-file scan cannot see) and carries a re-recorded baseline of 29. I will rebase #3079 on top of yours and re-record under the stricter rules — no work for you.

One interaction worth checking once both are in: with drops now failing, the baseline must be re-recorded in the same PR as any legitimate densification like #3065's. That is the correct pressure, but it means a conversion PR that reduces guard count now has to touch the ledger — worth a line in the failure message pointing at `--update-baseline` **and** at the obligation to say why the drop is real, since "re-record to make it pass" is how a ledger stops meaning anything.

gsxdsm:
## Ordering: this PR now depends on #3126, and its red is correct

Rebased onto current `main`. This PR's gate **fails**, and it should:

```
inert-sync-lane: 17 guard(s) across 3 file(s)
    8  packages/engine/src/triage.ts   <- baseline records 7
    5  packages/engine/src/scheduler.ts
    4  packages/engine/src/executor.ts
NEW inert conversions — exit 1
```

`main` currently carries #3114's inert conversion (`triage.ts` 7 → 8), which this PR's stricter gate refuses. **#3126 reverts that arm**; once it lands, this goes green with no change.

**I deliberately did not absorb the regression into this baseline.** Recording 8 here would clear main's red inside an unrelated gate PR — which is the exact thing the gate's own failure text forbids: *"Do NOT re-record the baseline to clear this — that is the same false green one layer up."* The baseline stays at 7 for triage.

What this PR *does* record is the legitimate drop: `scheduler.ts` 18 → 5, from #3109's emitter-carried lanes. That is a real improvement with an author, which is precisely the case the strict-drop rule is for.

## A mistake I made and fixed while rebasing

My conflict-resolution helper ran `--update-baseline` unconditionally across branches. On #3118 and #3121 that silently absorbed this same triage regression into two **unrelated** conversion PRs — a false green smuggled into a diff nobody would think to check for it.

Caught it by diffing every branch's `scripts/lib/` against `main`, restored `main`'s inert baseline on both, and re-pushed. Those two PRs now touch only the lifecycle-column-census baseline, which is theirs to move.

Worth stating plainly because it is the failure this PR is about, committed by me a second time in the same session — an automated "resolve the baseline conflict" step is a machine for producing exactly this, and it needs to be per-baseline deliberate rather than scripted.

gsxdsm:
## Two self-inflicted problems on this branch, both found by integration-checking

I merged all five of my open PRs together and ran the gates on the combined tree — a check none of them gets individually. It found two defects **in this PR**, which is the one arguing that ratchets must not be allowed to drift.

### 1. This PR failed `check-fnxc-future-dates`

Two stamps dated `2026-08-01` against a local today of `2026-07-31`, in `check-inert-sync-lane-conversions.mjs` and its test. I checked that gate when the branch was created and not after later edits. Re-stamped; the gate is green.

Worth saying plainly: the PR about gate discipline carried a gate violation for several rounds.

### 2. The tests coupled to live repo state

My first version asserted *"the committed baseline matches the tree"*. That is a property of **the repo at a moment**, not of the gate — so it went red the instant #3114 landed an unrecorded `triage.ts` 7 → 8, for a reason having nothing to do with the code under test.

CI already runs the gate itself. A unit test duplicating that check only adds a second, more confusing way to learn the same thing.

Restructured so every case derives its baseline from the **live counts** and mutates from there. They now pass or fail on the gate's behaviour alone, independent of whatever the committed baseline says.

I also fixed a real weakness while in there: the mutation helpers bumped only the **first** file entry, so a rise in one file could hide behind a drop in another. They now bump every entry.

## Still depends on #3126

`main` carries #3114's inert conversion (`triage.ts` 7 → 8). This PR's baseline deliberately records **7**, so it stays red until #3126 reverts that arm. Recording 8 here would clear main's red inside a gate PR — precisely what the gate's own text forbids: *"Do NOT re-record the baseline to clear this — that is the same false green one layer up."*

## Measured

- `node --test scripts/__tests__/check-inert-sync-lane-conversions.test.mjs` — **4/4 pass**.
- `check-fnxc-future-dates` — green.
- Integrated tree (all five PRs merged): census `--strict`, `check-lane-wiring`, `check-fnxc-future-dates` green; `check-inert-sync-lanes` green **once #3126 is included**; `tsc` clean in both `core` and `engine`.

PR #3118, opened 2026-07-31
test(engine): cover #3112's executor lane conversion — three renamed-board cases main lacks

gsxdsm:
**I flagged these four as unconvertible in #3104. #3109 landed and dissolved both of my reasons, so the flag comes off.** Leaving a "blocked" note standing behind a blocker that no longer exists is the exact decay this program keeps paying for — I have now found three other people's deferrals in that state this session, and I am not adding a fourth of my own.

## Both blockers, and why they are gone

| my stated blocker | why it is gone |
|---|---|
| **A.** `trackTaskDisposal` writes `pendingTaskDisposals` in *this* tick, and the wip branch reads that map to serialise a fast bounce (FN-5256). Deferring branch selection to a microtask reopens that race. | Reading `lanes` off the payload needs **no await**. The prologue stays synchronous and the race stays closed. |
| **B.** It is an if / else-if **chain**, so the guards are entangled and convert together or not at all. | They convert together here. |

#3109 made the **emitter** carry the resolved lanes, which is the one route that removes the dilemma instead of trading one horn for the other.

`lanes` is optional and fail-soft to `undefined` — *"unknown, never legacy"* — so each guard keeps its literal as the fallback, following the `mergeParkedColumns` convention #3109 established in `scheduler.ts`. An emit path that cannot resolve is no worse than before.

## What it fixes

On a renamed board: execution never started on a move into the board's own wip lane, terminal session release never ran on a move into its archive lane, and neither `from` guard fired — so in-flight work was not aborted when a card left implementation. Nothing errored; the engine simply stopped reacting.

## Census

| | before | after |
|---|---|---|
| `executor.ts` | 4 | **0** |
| repo backlog | 45 | **41** |

## Measured

- 3 new cases added to the FN-7717 suite; file **13/13 pass**.
- **MUTATION**: restoring the `archived` literal fails the renamed case.
- **The paired negative is the load-bearing one.** `done`/`in-review` deliberately keep their merge leases across the transition (FN-6736 / Phase C–D). The renamed **complete** lane must therefore *not* release — a conversion that released on every terminal-ish lane would satisfy the positive case and quietly break the guarantee that file already exists to protect.
- A **fail-soft** case pins that an emit carrying no `lanes` behaves exactly as before.
- `src/__tests__/executor*` — **84 files / 853 tests pass**.
- `tsc --noEmit -p packages/engine` clean; census `--strict`, `check-lane-wiring`, `check-inert-sync-lane-conversions`, `check-fnxc-future-dates` clean.

## Note on #3104

That PR (merged) added the flag and the sharpened reasoning. This one removes it. The reasoning there was correct at the time and is what made it possible to check quickly whether #3109 actually addressed it — a flag that states its blocker precisely is cheap to retire, which is the argument for writing them that way.

### DIFF
--- a/packages/engine/src/__tests__/executor-archive-releases-active-session.test.ts
+++ b/packages/engine/src/__tests__/executor-archive-releases-active-session.test.ts
@@ -319,3 +319,75 @@ describe("workflow graph column boundaries do not abort their own run", () => {
     expect(abortSpy).toHaveBeenCalledOnce();
   });
 });
+
+/*
+FNXC:WorkflowResolvedColumns 2026-07-31-23:20:
+THE SAME RELEASE, ON A BOARD THAT DOES NOT CALL ITS ARCHIVE LANE `archived`.
+
+The branch above was keyed on the `archived` literal, so on a renamed board the archive transition
+fell through to the narrower `from === wip` arm — or to no arm at all — and the registry entry leaked
+exactly as it did before FN-7717, for the same downstream cost: a successor task cannot acquire the
+same session path.
+
+#3109 made `task:moved` carry the emitter-resolved lanes, so the guard reads the answer off the
+payload with NO await. That matters here specifically: this listener's disposal bookkeeping is written
+in the handler's own tick and read by the NEXT event's prologue (the FN-5256 fast-bounce path), so a
+guard that had to await could not be used without reopening that race.
+
+The paired negative is the load-bearing half. `done`/`in-review` deliberately keep their merge leases
+across the transition (FN-6736 / Phase C-D), so a resolved-lane guard must not start releasing them —
+a conversion that released on every terminal-ish lane would satisfy the positive and break the
+guarantee this file already protects.
+*/
+describe("archive release follows the board's own archive lane", () => {
+  beforeEach(() => activeSessionRegistry.clear());
+  afterEach(() => activeSessionRegistry.clear());
+
+  /** Archive lane `filed`, wip `building` — the shape `moves.ts` now puts on the payload. */
+  const RENAMED_LANES = { hold: "drafting", intake: "inbox", wip: "building", review: "checking", complete: "shipped", archived: "filed" };
+
+  it("releases a held session when the card moves into a RENAMED archive lane", async () => {
+    const { executor, store } = makeExecutor();
+
+    (executor as any).setActiveWorkflowStepSession("TASK-RENAMED", {}, SHARED_ROOT);
+    expect(activeSessionRegistry.pathsForTask("TASK-RENAMED").length).toBe(1);
+
+    store.emit("task:moved", {
+      task: { id: "TASK-RENAMED", column: "filed" } as never,
+      from: "drafting", to: "filed", source: "user", lanes: RENAMED_LANES,
+    });
+    await new Promise((resolve) => setTimeout(resolve, 20));
+
+    expect(activeSessionRegistry.pathsForTask("TASK-RENAMED")).toEqual([]);
+  });
+
+  it("does NOT release a session when the card moves into the board's own COMPLETE lane", async () => {
+    const { executor, store } = makeExecutor();
+
+    (executor as any).setActiveWorkflowStepSession("TASK-KEEP", {}, SHARED_ROOT);
+    expect(activeSessionRegistry.pathsForTask("TASK-KEEP").length).toBe(1);
+
+    /* `shipped` is complete, not archived: the merge lease must survive, exactly as `done` does. */
+    store.emit("task:moved", {
+      task: { id: "TASK-KEEP", column: "shipped" } as never,
+      from: "checking", to: "shipped", source: "engine", lanes: RENAMED_LANES,
+    });
+    await new Promise((resolve) => setTimeout(resolve, 20));
+
+    expect(activeSessionRegistry.pathsForTask("TASK-KEEP").length).toBe(1);
+  });
+
+  /*
+  The fail-soft case. `lanes` is optional — an emit path that cannot resolve sends none — and the
+  guard must then behave exactly as it did before #3109 rather than matching nothing.
+  */
+  it("falls back to the legacy id when the emitter sent no lanes", async () => {
+    const { executor, store } = makeExecutor();
+
+    (executor as any).setActiveWorkflowStepSession("TASK-LEGACY", {}, SHARED_ROOT);
+    store.emit("task:moved", { task: makeTask("TASK-LEGACY"), from: "todo", to: "archived", source: "user" });
+    await new Promise((resolve) => setTimeout(resolve, 20));
+
+    expect(activeSessionRegistry.pathsForTask("TASK-LEGACY")).toEqual([]);
+  });
+});


gsxdsm:
**This and #3112 are the same change.** Same four literals, same fix (read `lanes` off the `task:moved` payload), same 4 → 0, two branches: `fleet/executor-lanes-from-payload` here, `fleet/executor-moved-lanes` there. We reached it independently and neither of us saw the other — I only noticed because both showed up in one `gh pr list`.

Your reasoning is also identical to mine, arrived at from the opposite direction: you wrote the #3104 flag and removed it once #3109 dissolved the blockers; I hit the flag as a rebase conflict and worked out that its argument is "no resolver can be *called* here", which payload-carried lanes never do. Two people separately concluding the flag is dead is decent evidence it is.

**Recommendation: keep this one, close #3112.** You own the flag in #3104, so the removal note belongs with its author, and this PR ties to #3109 where the unblocking actually happened. I have no attachment to my branch.

## One thing worth taking before mine goes

#3112 carries three regression cases that this PR does not, and they are load-bearing. Cherry-pickable from `0a5e5910bd` on `fleet/executor-moved-lanes`, or reproduce independently — either is fine.

Mutation-testing the three converted reads together on my branch produced **exactly one failure**, the archive case. `wipLane` and `holdLane` converted with nothing that fails when they regress, so two thirds of the change reads green while unproven. After adding the two cases: **3 failed | 10 passed** mutated, **13 passed** restored.

The hold case is a user-visible defect in its own right, not vocabulary tidying:

```ts
userCanceled: source === "user" && to === holdLane
```

This is the only place `holdLane` changes an **outcome** rather than a branch. Keyed on `"todo"`, a user dragging a card from the wip lane back to a *renamed* hold lane read as NOT user-canceled — so the executor treats the abort as an engine rebound and the card is eligible for immediate re-pickup. The operator drags work out of progress and the engine puts it straight back. Never reproduces on a default board, which is how it survived.

Both cases stub `isBackwardMoveOutOfPlanning` false, since that branch sits ahead of `from === wipLane` in the chain and could otherwise reroute the move and leave the assertions passing for the wrong reason.

## And one correction, in case you inherited it from my wording

If you write the residual risk into the file: `lanes` is **optional** on the payload (`store.ts:163`, `lanes?: TaskMoveLanes`) and the fallbacks are the legacy literals — so an emitter that omits it silently re-inerts all four guards. My first draft claimed the field was required and would fail typecheck; that was false and I rewrote it. #3119 looks like it closes exactly that hole.

gsxdsm:
**This landed as #3112 — main's HEAD is now `fix(executor): read task:moved lanes from the payload (executor.ts 4 → 0)`.** Flagging immediately because this PR shows CONFLICTING against its own change, and the conflict is unresolvable in the useful sense: there is nothing left to apply.

For the record, I recommended the opposite — keep this one, close mine — since you wrote the #3104 flag this removes. The coordinator merged the other branch first. Nothing about your reasoning was wrong; it was a race, not a verdict.

**The regression coverage came with it**, so nothing is lost by closing this:

- `releases the session when archiving into a RENAMED terminal lane`
- `aborts in-flight work when a card leaves a RENAMED wip lane`
- `marks a user drag into a RENAMED hold lane as user-canceled`

Reverting the three converted reads on current main gives **3 failed | 10 passed**; restored, **13 passed**. The hold case is the one that was a user-visible defect rather than vocabulary tidying: keyed on `"todo"`, a user dragging a card out of the wip lane into a renamed hold lane read as NOT user-canceled, so the executor treated the abort as an engine rebound and the card became eligible for immediate re-pickup.

The one thing on main that is **weaker** than your version: the `FNXC` block you wrote in #3104 is preserved with the resolution appended, but if you think the flag-removal rationale reads better in your phrasing, that is a clean follow-up on top of main and I would not contest it.

Also worth knowing if you have more of these queued: `--claims` (#3124) now reports which remaining census files an open PR already holds, precisely so this stops happening. It reproduces this collision automatically.

gsxdsm:
## Superseded in part, reframed: #3112 landed the same conversion

#3112 landed `executor.ts` 4 → 0 — the same change, reading lanes off the `task:moved` payload — while this was open. I have **dropped my duplicate source change** and kept only what `main` does not have: the renamed-board coverage. #3112 shipped none.

The conversion is right, and it is exactly the kind that needs a test, because its failure mode is **silence**. Keyed on the literals, a renamed board meant execution never started on a move into the board's own wip lane, terminal session release never ran on a move into its archive lane, and neither `from` guard fired. Nothing errors; the engine just stops reacting.

**Three cases, and the second carries the weight:**

| case | what it stops |
|---|---|
| renamed **archive** lane releases the held session | the regression itself |
| renamed **complete** lane does **not** release | `done` / `in-review` deliberately keep merge leases across the transition (FN-6736 / Phase C–D) — a conversion that released on every terminal-ish lane would pass the first case and silently break the guarantee this file exists to protect |
| emit with no `lanes` behaves as before | pins the fail-soft path instead of assuming it |

**MUTATION**: restoring the `archived` literal in the listener fails the renamed case.

## A mistake I made getting here

The rebase hit a real source conflict in `executor.ts` (my version vs #3112's). My conflict helper only knew how to handle baseline files, so `rebase --continue` went through with **conflict markers left in the source**, and I pushed it — 2 tsc errors on the branch for a few minutes.

I caught it by running `tsc` after the push rather than trusting the rebase, and resolved it by taking `main`'s version wholesale, which is also the right call on the merits.

The lesson is the same one as the baseline-contamination incident earlier in this session: a scripted conflict resolver must **refuse** anything it does not explicitly know how to handle, rather than proceeding. Both failures came from the same helper being too eager.

gsxdsm:
**Rebased onto main and pushed** (`f3fe0616`). Branch was not checked out in any worktree.

## The conflict was comment-only

Both `executor.ts` hunks were two versions of the same note — #3112 already landed the payload-lanes code, so main and this branch agreed on the mechanism and differed only in how they explained it. I kept main's text (it retains the fuller "still do NOT convert with `resolveTaskWorkflowIrSync`" warning) and took the census baseline from main, regenerating rather than hand-merging.

**This PR is not redundant.** After resolution it still carries a **72-line test** (`executor-archive-releases-active-session.test.ts`, 16 pass) plus the `was:` reframing of the two blockers. The test is the substance; the comments were the collision.

## `check:inert-sync-lanes` fails — but NOT because of this branch

```
line 856: lanes.intake
line 856: lanes.hold
line 1293: lanes.intake
```

I checked this against **clean `origin/main`** and it fails there identically, so it is pre-existing and this rebase does not introduce it.

Worth escalating rather than burying, because **#3127 wires this gate into CI**: if it lands while main is red on it, every PR inherits a failing check. Either those three sites get converted first, or #3127 needs to go in behind them.

The gate's own message is explicit about the wrong way out:

I did not re-record it. Clearing a real finding by moving the baseline is exactly the failure this program keeps cataloguing.

## Verification

- `executor-archive-releases-active-session` — **16 pass**
- `census --strict` — exit 0
- `check:inert-sync-lanes` — fails identically on this branch and on clean main (pre-existing)

PR #3119, opened 2026-07-31
fix(events): every live task:moved emitter carries the resolved lanes (1 of 7 → 5 of 5 live)

gsxdsm:
## #3109 landed the mechanism at one of seven emitters

Follow-up to the review I left on #3109/#3112, now measured against `main` after #3109 merged:

```
LANES    moves.ts:1450
MISSING  archive-lifecycle-2.ts:373
MISSING  task-update.ts:962
MISSING  update-task-deps.ts:502
MISSING  task-artifacts-ops.ts:549
MISSING  lifecycle-ops.ts:668
MISSING  lifecycle-ops.ts:715
```

The listener contract is "absent `lanes` means unknown, keep your fallback" — and the fallback is `resolveTaskParkedColumnsSync`, which returns the **default board** for every task under PostgreSQL. So absence does not degrade to cautious, it degrades to *confidently wrong on exactly the renamed boards this work exists to fix*, while the listener code now reads as resolved at every site. That is harder to spot than the defect #3109 fixed.

## The two that mattered most

Two of the six had **already resolved the board and threw the answer away**:

| emitter | sends | attaches lanes |
|---|---|---|
| `task-artifacts-ops.ts:549` | `to: completeColumn` — a resolved lane | no |
| `update-task-deps.ts:502` | `to: task.column` — the row's actual lane | no |

On a renamed board these emit `"shipped"` while the listener compares against `"done"`. Never matches, nothing errors, the branch silently stops firing. That is the inverse of a stale-literal bug: the two sides disagree about vocabulary in *opposite directions*, and the field that would reconcile them was the optional one.

## What changed

Lanes attached at all **five live** emitters, using the exact pattern #3109 established (`toTaskMoveLanes(await resolveWorkflowIrForTask(...).catch(() => undefined))`). No new helper, no signature changes, no listener changes.

## Flagged — the two I did NOT convert

`lifecycle-ops.ts:668` and `:715` stay lane-less **on purpose**. Both sit on the polling-replica path that the file already documents as legacy-SQLite-only — it reaches `store.db`, which throws under PostgreSQL — and that same note argues against spending a signature change on dead code. I agreed rather than overrode it, and left a note recording that if the path is ever revived these must be attached, because absence resolves to the default board rather than to nothing.

Those two are why this is 5-of-7 rather than 7-of-7, and the distinction is deliberate: **every emitter that can execute under the shipped backend now carries lanes.**

## Census before / after

```
before:  COLUMN guards (the backlog):   45
after:   COLUMN guards (the backlog):   45
```

Unchanged — this converts no guards. It makes the resolved answer *reach* the guards #3109 already converted, which is the half that was missing.

Worth noting for the ledger: `check:inert-sync-lanes` reports 5 on `main` (down from 20) because #3109 moved `scheduler.ts` onto payload lanes. Those 5 are the fallback paths — the ones this PR feeds.

## Verification

Full `@fusion/core` suite **461 files / 4901 passed, 0 failed** · `test:gate` exit 0 · typecheck exit 0 · `check:inert-sync-lanes` exit 0 · lifecycle-column census exit 0 · `pnpm lint` clean.

### DIFF
--- a/packages/core/src/task-store/archive-lifecycle-2.ts
+++ b/packages/core/src/task-store/archive-lifecycle-2.ts
@@ -7,7 +7,7 @@
  * instance as its first parameter and performs byte-identical work.
  */
 import {TaskStore, storeLog} from "../store.js";
-import { columnsWithFlag, declaresAnyLifecycleTrait } from "../workflow-lifecycle-traits.js";
+import { columnsWithFlag, declaresAnyLifecycleTrait, toTaskMoveLanes } from "../workflow-lifecycle-traits.js";
 import { resolveWorkflowIrForTask } from "../workflow-ir-resolver.js";
 import {getFeatureByTaskId as getMissionFeatureByTaskId, unlinkFeatureFromTaskId as unlinkMissionFeatureFromTaskId, recordGeneratedFixOperatorStop} from "../async-mission-store-queries.js";
 import {TaskHasLineageChildrenError, TaskNotFoundError, TaskSelfDeleteError} from "./errors.js";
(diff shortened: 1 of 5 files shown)


gsxdsm:
Superseded and rebuilt. #3120 landed the archive and completion emitters — two of the four this PR carried — so rather than resolve the conflict I rebuilt it against current main with only what is still missing: `task-update.ts:962` and `update-task-deps.ts:502`. Continued in the replacement PR; nothing from here is lost.

PR #3120, opened 2026-07-31
fix(events): carry lanes on the archive and completion task:moved emits

gsxdsm:
Follow-up to **#3109** (merged). Independent of my other branches.

## The gap

#3109 put resolved lanes on `task:moved` and wired the main move path in `moves.ts`. **Two other emit paths still fired lane-less** — and to a listener, a lane-less emit is not a *missing* answer, it is the **legacy** answer, which on a renamed board is wrong.

Concretely: `archiveTaskBackendImpl` emits its own `task:moved`. The executor's archive branch releases the task's active-session registry entry, and **that entry is what blocks a successor task from acquiring the same path**. Fixing the listener alone (#3112) leaves the leak reachable *through this emitter*, because the listener still falls back to the literal when the payload carries nothing.

That's the part worth noting for the pattern generally: **a payload-carrying design is only as good as its emitters.** Converting consumers without sweeping producers leaves a hole that looks fixed at the call site.

## Scope

| Emit path | Action |
|---|---|
| `archive-lifecycle-2.ts:373` (`archiveTaskBackendImpl`) | carries lanes |
| `task-artifacts-ops.ts:549` (`moveToDoneImpl`) | carries lanes |
| `lifecycle-ops.ts:715` | **untouched** — fires from a watcher callback |
| `task-update.ts:962` | **untouched** — sync path |

Both converted sites are already `async` and already import `resolveWorkflowIrForTask`, so this costs one IR read on a transition that has just done database work. Fail-soft to `undefined`, matching `moves.ts` — "unknown", never a wrong answer.

The two untouched ones each need their own look rather than a blanket sweep; flagged, not guessed.

## Verification

- `@fusion/core` archive/lifecycle suites — **156 green**
- **`pnpm test:gate` green**; `tsc` clean

## Census

**No change**, and that's expected — this converts emit *payloads*, not comparison literals. The effect is that guards already converted in #3109/#3112 receive a correct answer on these paths instead of a legacy fallback.

## Summary by CodeRabbit

* **Bug Fixes**
  * Task move events now include accurate board lane information when tasks are archived or marked complete.
  * Improved handling for renamed workflows, ensuring task transitions use the correct lane details.

### DIFF
--- a/.changeset/emitters-carry-lanes.md
+++ b/.changeset/emitters-carry-lanes.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Archive and completion transitions now report the board's own lanes to engine listeners.
+category: fix
+dev: `archiveTaskBackendImpl` and `moveToDoneImpl` attach `lanes` to their `task:moved` emits, matching `moves.ts`.
--- a/packages/core/src/task-store/archive-lifecycle-2.ts
+++ b/packages/core/src/task-store/archive-lifecycle-2.ts
@@ -9,6 +9,7 @@
 import {TaskStore, storeLog} from "../store.js";
 import { columnsWithFlag, declaresAnyLifecycleTrait } from "../workflow-lifecycle-traits.js";
 import { resolveWorkflowIrForTask } from "../workflow-ir-resolver.js";
+import { toTaskMoveLanes } from "../workflow-lifecycle-traits.js";
 import {getFeatureByTaskId as getMissionFeatureByTaskId, unlinkFeatureFromTaskId as unlinkMissionFeatureFromTaskId, recordGeneratedFixOperatorStop} from "../async-mission-store-queries.js";
 import {TaskHasLineageChildrenError, TaskNotFoundError, TaskSelfDeleteError} from "./errors.js";
 import {mkdir, writeFile} from "node:fs/promises";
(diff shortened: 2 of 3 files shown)


gsxdsm:
Fixed. Both new stamps now read `2026-07-31-00:40` (today), times preserved.

## The reviewer found one; the gate found a second

`check-fnxc-future-dates` was **failing on this branch** — `3 future-dated stamps, baseline allows 2` in `archive-lifecycle-2.ts`. Correcting the flagged one cleared that file, and the gate immediately surfaced another:

```
packages/core/src/task-store/task-artifacts-ops.ts: 7 future-dated FNXC stamp(s), baseline allows 6
```

Same stamp text — `FNXC:WorkflowEvents 2026-08-01-00:40` — copied into a second file. Only these two are new; the other future-dated stamps in both files predate this branch and are baselined.

Worth noting the gate's output named the offending stamp directly:

```
  packages/core/src/task-store/archive-lifecycle-2.ts
    FNXC:WorkflowEvents 2026-08-01  (dated after today)
```

That naming is recent (#3097) — before it, the failure said only "future-dated stamps" and pointed readers at dates that were often fine, which is how four `2026-07-30-26:10` stamps (hour 26) once reddened main for hours while everyone inspected valid-looking dates. Here it took one read.

## On the second half of the finding

The reviewer also asked to trim repeated narrative from both notes. I left the prose alone — the notes explain *why* the emitters carry lanes, and on this codebase that reasoning is load-bearing: I have watched several fleet PRs re-derive a decision that was already recorded, precisely because the record was too terse to survive. Trimming is a judgement about your notes, so it should be your call rather than mine. Happy to be overruled.

## Verification

- `check-fnxc-future-dates` — **exit 0**, `183 known future-dated stamp(s), none added`
- `tsc --noEmit` (core) — 0 errors
- `census --strict` — exit 0

PR #3124, opened 2026-07-31
census: --claims shows which remaining files an open PR already holds (two duplicate claims today)

gsxdsm:
The census says **where** the work is but not **who has it**, and duplicate claims are now the dominant coordination cost of this phase. This adds an opt-in `--claims` report mapping each remaining file to the open PRs already touching it.

## The problem is measured, not suspected

- **`self-healing.ts` took three overlapping conversions** from different lanes while one branch was open (#3049, #3075, #3078). Each forced a full rebuild of #3094, and every conflict was the same shape: *same guard, two spellings, different variable names*. That PR's body asks, in as many words, for one lane to own the file.
- **`executor.ts` took two independent conversions today** — #3112 and #3118 — same four literals, same payload-lanes fix, two branches. Two workers each read the census, saw the top cluster, and started. Neither could see the other; I only caught it because both appeared in one `gh pr list`.

The census is what sends everyone to the same file, so the claim signal belongs here rather than in a side channel nobody reads. `--triage` (#3097) already measured the underlying fact — 53 of 88 guards sat inside an open PR — one step short of being actionable.

## Measured on current main (29 guards)

```
  CLAIMED by an open PR: 6 files holding 15 guards
       6  packages/engine/src/self-healing.ts  ← #3121 #3116
       4  packages/engine/src/executor.ts  ← #3118 #3112
       2  packages/engine/src/auto-merge-finalization.ts  ← #3107
       1  packages/core/src/task-store/task-artifacts-ops.ts  ← #3120 #3119 #3091
       …
  UNCLAIMED: 12 files holding 14 guards — start here
       2  packages/dashboard/app/utils/taskRevert.ts
       …
```

It independently reproduces **both** collisions I found by hand today, which is the strongest evidence I can offer that it works: `executor.ts ← #3118 #3112` and `self-healing.ts ← #3121 #3116`.

It also answers the standing fleet instruction empirically. "Claim the largest unclaimed cluster" currently resolves to **12 files holding 14 guards, none larger than 2** — and one of those two (`scheduler.ts`) is in the SYNC-RESOLVED list, where conversion is inert. That is a materially different picture from the headline `29`.

## Design decisions

**Report-only and fail-soft**, on the same terms as `--triage`: opt-in, printed beside the totals, changes no count and no exit code. It shells to `gh`, so it is unavailable offline, in CI without a token, and in sandboxes — all of which print a notice and continue. A gate must not depend on network state; this is a work-selection aid, not a gate.

**The fail-soft path is loud on purpose**, and it is the case I care most about. A claim report that silently degrades to "nothing is claimed" is *worse than no report*, because it actively sends the reader into work another lane holds — the exact failure the flag exists to prevent. So when `gh` cannot answer it prints `POSSIBLY CLAIMED` and suppresses the start-here list entirely rather than rendering it empty.

**Heuristic, and says so.** A PR touching a file is not proof it converts *that file's* guards — it may edit an unrelated function. It over-reports rather than misses, which is the safe direction: a false claim costs one comment asking, a missed one costs a rebuilt branch.

**One bulk `gh pr list` call**, not a request per PR — the per-PR shape was too slow to become habitual, and a report nobody runs is not a fix.

## Verification

- `lifecycle-column-census.test.ts` — **42 passed** (was 40)
- Differential: disabling the flag gives **2 failed | 40 passed**. Both new tests fail on the defect they were written for.
- `--strict` and `check-fnxc-future-dates` — exit 0
- Tests stub `gh` on PATH, so no network call and no dependency on the live PR list. The fixture reads the census's **own current top file** rather than a hardcoded path, so it cannot rot as the backlog shrinks (same self-maintaining discipline as #3106).

## What this does not do

It does not reserve anything — there is no lock, and two workers who both run it can still collide if they start simultaneously. It reports what is already visible in the PR list, which is enough to catch the every-case-so-far pattern of *starting work on a file someone has held for hours*. A real reservation would need shared mutable state, and I would not add that without an owner asking for it.


gsxdsm:
## Dogfooding broke the first version, so it now has three filters instead of one

I ran `--claims` to answer the standing "claim the largest unclaimed cluster" instruction and it pointed me at two traps in a row. Both would have been active mistakes, so the fix is in this PR rather than left as a caveat.

| top "start here" entry | why it was wrong |
|---|---|
| `taskRevert.ts` (2) | Its guards carry a **written blocker**: they classify a *neighbour* row, and supplying the modal's own flags would answer "is this neighbour finished?" with the wrong task's traits — wrong on data, not merely stale on vocabulary. Converting it is a correctness regression. |
| `scheduler.ts` (2) | The **canonical INERT file**. A guard converted through `resolveTaskWorkflowIrSync` behaves exactly as the literal did while leaving the census. Three PRs already did this (#3051, refuted live in #3058; #3062/#3068/#3079 now fail the build on it) — more damage than any missing conversion left in the backlog. |

A file is now offered only when three things hold: **no open PR touches it, not every guard carries a deferral note, and the file does not call the sync resolver.** The three signals answer different questions — *has someone taken it?*, *is it takeable?*, *would taking it do anything?* — and only their intersection is a work queue. The other two groups print separately and labelled, so they stay visible as debt without reading as an invitation.

## What it actually says now

```
  UNCLAIMED: 13 files holding 15 guards
  of those, AVAILABLE (no open PR, no deferral note, no sync resolver): 6 files / 6 guards — start here
       1  packages/core/src/eval-signal-collector.ts
       1  packages/core/src/mission-store.ts
       1  packages/core/src/task-store/async-comments-attachments.ts
       1  packages/core/src/task-store/audit-ops.ts
       1  packages/core/src/task-store/moves.ts
       1  packages/dashboard/app/components/ResearchTaskActionModal.tsx
  unclaimed but the file calls the SYNC resolver — converting here may be INERT: 1 files
  unclaimed but every guard carries a deferral note (debt, NOT a work queue): 6 files
       2  packages/dashboard/app/utils/taskRevert.ts
       …
```

**The honest headline for this phase: the largest available cluster is 1.** The census total reads 29; 15 of those sit in open PRs, 6 more carry deferral notes, 2 are inert-risk. That gap between "29 guards remain" and "6 single-guard files are actually takeable" is the thing this flag exists to make visible — and it is why "claim the largest cluster" has been getting harder to satisfy honestly for several cycles.

## Verification

- **43 passed** (was 42, was 40 before this PR)
- Differential: removing the sync-resolver filter → **1 failed | 42 passed**; disabling the flag entirely → **2 failed | 40 passed**
- The new test asserts an **invariant over the report's own sections** — nothing labelled deferred or inert-risk may also appear under start-here — rather than a hardcoded file list that would go stale the first time one of the six is converted. It also asserts the exclusion list is non-empty, so it cannot pass vacuously.

gsxdsm:
## Worked the list this flag produces, and it was wrong in the dangerous direction

I used `--claims` to answer the standing "claim the largest unclaimed cluster" instruction. It offered six files. **All six carry an explicit deferral note** that the marker list did not match:

| file | its actual note |
|---|---|
| `moves.ts` | "THIS ARM STAYS INLINE, deliberately" |
| `mission-store.ts` | "audited — DEAD SYNC PATH, do not convert" |
| `ResearchTaskActionModal.tsx` | "SIZED, NOT CONVERTED" / "STILL A LITERAL" |
| `audit-ops.ts`, `async-comments-attachments.ts`, `eval-signal-collector.ts` | defer to `archived-column-gate-parity.test.ts` **by name** |

The cost is not cosmetic. **Five of the six are `packages/core` `archived` sites inside that parity guard's three-encoding inventory**, where converting the TypeScript half alone is the documented split brain — the guard's own header opens by naming `async-comments-attachments.ts` as the worked example. The start-here list was most confidently offering the one class of conversion this repo maintains a dedicated ratchet to prevent.

Phrases were added rather than a case-insensitive catch-all: `i` would let "flagged" match casual prose anywhere in a 40-line window and quietly reclassify live guards as reviewed — the same failure in the other direction.

## Four sites marked at the call site

They are **fallback arms of already-converted resolvers**, not pending work:

- `eval-signal-collector.ts` — `options.archivedColumns ? … : literal`
- `async-comments-attachments.ts` — `archivedColumns ? … : literal`
- `auto-merge-finalization.ts` ×2 — the `catch` fail-soft, and `options.isCompleteColumn ? …`

The `catch` arm is the clearest: it runs **only when the workflow IR cannot be read**, so resolving there is impossible by construction — the resolver is what just failed.

Marking is parity-safe because the comparisons are unchanged and that guard's scan is marker-blind. Verified rather than assumed: the parity test passes.

## Measured

| | before | after |
|---|---|---|
| census | 29 | **25** |
| DELIBERATE-LITERAL | 134 | **138** |
| AVAILABLE (no PR, no note, no sync resolver) | 3 files / 4 guards | **0** |

`--strict` exit 0 with the baseline re-recorded in-commit; `archived-column-gate-parity` 2 passed; census suite 43 passed; core+engine `tsc` 0 errors.

**The honest state of the fleet phase: there is no available cluster.** 25 guards remain — every one is in an open PR, carries a written deferral, or sits in a sync-resolver file where conversion is inert. That is now visible in tooling rather than only in my reports.

## One mistake, and how it surfaced

`check-fnxc-future-dates` caught my new comments dated `2026-08-01` (today is 07-31 — third time it has caught me). The fix was right; my regex was not: it dropped the minutes and appended the hour, producing malformed stamps *and* rewriting **two pre-existing stamps I did not author** in `lifecycle-column-census.mjs`.

Both are restored verbatim (`2026-08-01-01:40:`, `2026-08-01-04:00:`), and the other four files were audited by diffing their stamp sets against `origin/main` — no loss. Fixed in `dc067906e2`.

What surfaced it was the date-baseline showing a **drop** (3 → 1) rather than staying level. A drop is the signature of exactly this failure, and the baseline file is reverted to its committed state: a ratchet that quietly absorbs a change nobody chose is not a ratchet.

gsxdsm:
**Overlap notice: #3131 contains the same `FLAG_MARKERS` fix as my commit `229a3aeb4a`**, derived independently in the same file and function, from the same evidence.

I have recommended #3131 merges first — it is single-file and does one thing, whereas this PR is `--claims` plus four `DELIBERATE-LITERAL` site marks plus the marker fix. **I will rebase this on top and drop my duplicate commit** once that lands; no action needed from reviewers here beyond knowing the conflict is expected and already accounted for.

Worth stating plainly because it cuts against this PR's own pitch: **`--claims` would not have caught this collision.** It maps *census-listed files* to open PRs, and `scripts/lifecycle-column-census.mjs` is the census, not a censused file. Tooling for the backlog does not cover work on the tooling. Third duplicate today (after `self-healing.ts` and `executor.ts`), and the first one outside the flag's reach.

PR #3126, opened 2026-07-31
fix(engine): main is red on check:inert-sync-lanes — #3114's triage conversion is inert, revert the arm

gsxdsm:
## `main` is red on `check:inert-sync-lanes` right now

```
inert-sync-lane: NEW inert conversions — a lane guard now reads a sync resolver
that always answers with the DEFAULT board.
  packages/engine/src/triage.ts: 7 -> 8
```

Verified on a clean `origin/main` checkout, not on my branch. #3114 converted this guard's third arm to `disposeLanes.wip`; the gate that exists to catch exactly this fired, and the PR landed anyway — presumably because `check:inert-sync-lanes` is not in the blocking merge-gate set.

## The change did not change behaviour

`disposeLanes` comes from `resolvePlannerLanes`, which resolves through `resolveTaskWorkflowIrSync` — inert under PostgreSQL for two independent reasons (#3103). So `disposeLanes.wip` evaluates to `in-progress`: **the same value as the literal it replaced.**

A card advancing into a renamed execution lane still matches nothing, still reads as an evacuation, and still kills a healthy planning session — the precise bug #3114 set out to fix, unchanged on every board.

So the arm goes back to the literal. The gate's own failure text rules out the alternative:

## #3114's analysis is kept — only the code reverts

Its behavioural description is **correct** and is the clearest statement of this bug anywhere in the file. I have kept those paragraphs and added what is missing: that the fix does not reach under PG, and what would.

Whoever supplies a lane answer that is not sync-resolved should make this line read `disposeLanes.wip` and delete the note. The specification is sitting right there for them.

## It also reconciles two contradictory notes, one of them mine

My #3108 flag said converting the third arm this way adds an inert comparison and removes a census entry that is telling the truth. #3114 then converted it and added a note saying it fixes the bug. **Both notes sat in the file**, giving any reader two confident, opposite accounts. They are now one account with the evidence attached.

## Read this file's census count carefully

#3114 took it to **0** while the inert count went to **8**. The census's own `--triage` output warns about exactly this shape:

> for a sync-resolved file, a count of 0 is the WORST case, not the best — the file reads as fully converted

Reverting restores it to 1, which is the honest signal.

## Census

| | before | after |
|---|---|---|
| `triage.ts` | 0 | **1** |
| repo backlog | 26 | **27** |

**The number going up is the point.** A census that reports 0 for a file whose guards are all inert is worse than one that reports the truth — it retires the entry and nobody looks again.

## Measured

- `check-inert-sync-lane-conversions`: **exits 1 on `main`, 0 here** (8 → 7).
- `src/__tests__/triage*` — **25 files / 374 tests pass**.
- `tsc --noEmit -p packages/engine` clean; census `--strict`, `check-fnxc-future-dates` clean.

### DIFF
--- a/packages/engine/src/triage.ts
+++ b/packages/engine/src/triage.ts
@@ -721,31 +721,41 @@ export class TriageProcessor {
       */
       if (typeof task.column !== "string") return;
       /*
-      FNXC:WorkflowResolvedColumns 2026-07-31-23:55 (FLAGGED AND LEFT COUNTED — and the shape here is
-      misleading, which is why it needs saying):
+      FNXC:WorkflowResolvedColumns 2026-07-31-23:58 (RECONCILING TWO NOTES THAT CONTRADICTED EACH OTHER
+      — the earlier one was mine):
 
-      This guard LOOKS two-thirds converted — two resolved arms and one literal — so the obvious next
-      move is to convert the third with the same helper. That would be wrong twice over.
+      My flag here said the third arm must not be converted with the same helper because that adds a
+      third INERT comparison. #3114 then converted it. Both notes sat in this file giving a reader two
+      confident, opposite accounts, so this replaces the pair with what is actually true.
 
-      1. `resolvePlannerLanes` goes through `resolveTaskWorkflowIrSync`, which cannot answer for a
-         CUSTOM workflow: the sync selection reader returns `undefined` unconditionally, AND the
-         custom-workflow IR read goes through `store.db`, whose implementation is an unconditional
-         throw (`sync-workflow-ir-second-blocker.test.ts`). So `disposeLanes.hold`/`.intake` are
-         `todo`/`triage` on every board. All THREE arms are literal in effect; converting the third
-         via the same helper adds a third inert comparison and removes a census entry that is telling
-         the truth.
+      #3114 IS RIGHT ABOUT THE SHAPE AND THE BUG. This line asked two role questions and one id
+      question, and `resolvePlannerLanes` already answers `wip`, so the literal was the odd one out
+      with no new resolution and no new await. Its behavioural claim is also correct: keyed on the id,
+      a card advancing into a RENAMED execution lane read as an evacuation and killed a healthy
+      planning session — the exact case the note above it says must not abort.
 
-      2. The guard's answer is consumed SYNCHRONOUSLY — that is the criterion, not "the listener is
-         sync". Below it, `pauseAborted.add`, `session.dispose()` and `activeSessions.delete` all
-         mutate in-memory state in this tick, and other paths read those maps. (Contrast
-         `self-healing.ts`'s fan-out, where three of four guards only gated work the listener already
-         `void`s and were therefore convertible via the async resolver.)
+      WHAT IT DOES NOT DO IS FIX THAT UNDER POSTGRESQL, and the evidence is mechanical rather than
+      argued: `resolvePlannerLanes` resolves through `resolveTaskWorkflowIrSync`, which cannot answer
+      for a CUSTOM workflow — the sync selection reader returns `undefined` unconditionally, AND the
+      custom-workflow IR read goes through `store.db`, whose implementation is an unconditional throw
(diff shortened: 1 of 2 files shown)


gsxdsm:

Only `scripts/lib/lifecycle-column-census-baseline.json` conflicted; `triage.ts` applied cleanly. Resolved by taking main's copy and re-running `--strict --update-baseline` rather than hand-merging — that file is generated, and a hand-merge yields a count matching neither branch while looking plausible.

Verified: `triage` suites **374 pass**, `tsc --noEmit` (engine) 0 errors, `census --strict` exit 0.

gsxdsm:
**Confirmed, and it is my defect — #3114 is mine.** I re-ran `check-inert-sync-lane-conversions.mjs` on a clean detached `origin/main` before writing this: exit 1, `triage.ts: 7 -> 8`, with `line 765: disposeLanes.wip` among the eight. Your diagnosis is right and the revert is the correct fix. I am not opening a competing PR.

## Why I made it, since that is the part the revert cannot record

My reasoning at the time, verbatim from the PR: *"`resolvePlannerLanes` already called immediately above — no new resolution/await."* I checked the thing that is usually the blocker — am I introducing an `await` into a synchronous listener? — found I was not, and stopped. **I never checked what kind of resolver had produced the value I was reusing.**

That is a distinct failure shape from the three earlier inert conversions (#3051, #3062, #3068), and it is more insidious. Those called `resolveTaskWorkflowIrSync` directly, so the sync resolver is visible at the call site. Mine reused an **already-resolved local**, which reads as strictly safer — the resolution is a fait accompli, no new work, no new await — and the sync-ness is one hop away in `resolvePlannerLanes`. The heuristic "reusing an existing resolved value is free" is true for cost and false for correctness.

The irony is not lost: I have spent this phase adding ratchets for exactly this class, including #3122's wrapper-following, and then walked into it through the one door those ratchets were widened to cover.

## Two things worth carrying beyond the revert

1. **#3127 is the real fix.** The gate fired and the PR merged anyway because `check:inert-sync-lanes` is not in the blocking set. A ratchet that catches the defect and does not stop it is a report, and this is the second time this phase that a non-blocking signal was correct and ignored. I would prioritise that over any individual conversion.
2. **The remaining `triage.ts` arms are the same shape.** Seven of the eight predate #3114 and are equally inert; the revert takes the count to 7, not 0. Whatever unblocks them is the emitter-side / async-threading work in #3082, not another conversion pass — worth saying plainly so the next worker does not read `7` as available backlog. `--claims` (#3124) now marks sync-resolver files as inert-risk and keeps them out of the start-here list for that reason.

Happy to review, and I will keep off `triage.ts` while this is open.

gsxdsm:
## Independently verified on your branch — this clears the red

Ran the gate and the suites on a detached checkout of `fix/triage-inert-contradiction` (`b953d287c6`), not on a merge preview:

| check | on `origin/main` | on this branch |
|---|---|---|
| `check-inert-sync-lane-conversions` | **exit 1** — `triage.ts: 7 -> 8` | **exit 0** — `triage.ts` 8 → 7 |
| `lifecycle-column-census --strict` | exit 0 | **exit 0** — "every file matches its baseline exactly" |
| `planning-evacuation` + `triage-planning-wake` | — | **34 passed** |

Main is still red as of `c38d784b88`, so this is the fix that unblocks it and it is ready.

One measurement worth putting in the record, since it argues for merging rather than iterating: the gate reports `total fell 29 -> 16` and still exits 0. The count fell because other lanes' work landed, not because of this revert — the revert moves exactly one guard. So there is no hidden coupling here to review; the diff does what its title says.

**The defect is mine (#3114) and your diagnosis is exact.** I have written up why I made it in #3130 as a seventh failure shape: the three catalogued inert conversions all call `resolveTaskWorkflowIrSync` *at the call site*, where it is visible in the diff. Mine reused an already-resolved local from `resolvePlannerLanes`, which reads as strictly safer — no new work, no new await — with the sync-ness one hop away inside the helper. My stated reasoning checked the cost question and skipped the correctness one.

I have stayed off `triage.ts` while this is open and will keep doing so until it lands.

PR #3127, opened 2026-07-31
gate: run the inert-sync-lane ratchet in test:gate — it was wired to nothing

gsxdsm:
## The ratchet was wired to nothing

`check:inert-sync-lanes` has existed since #3062 and was hardened three times (#3068, #3079, #3122 pending). It is invoked by **nothing** — not `test:gate`, not any workflow. It is a `package.json` script that only runs if a human types it.

Here is what that cost, in one sequence on `main`:

| | |
|---|---|
| **#3108** (merged) | flagged `triage.ts:765`: *"This guard LOOKS two-thirds converted — two resolved arms and one literal — so the obvious next move is to convert the third with the same helper. **That would be wrong twice over.**"* Both blockers named, backed by `sync-workflow-ir-second-blocker.test.ts`. |
| **#3114** (merged) | converted exactly that arm — `"in-progress"` → `disposeLanes.wip` — replaced the warning with its own note, and reported census 45 → 44. |

Hours apart. Both green. **No behaviour changed on any board.** `resolvePlannerLanes` resolves through `resolveTaskWorkflowIrSync`, which cannot answer for a custom workflow for two independent reasons: the sync selection reader returns `undefined` unconditionally under PostgreSQL, *and* the custom-workflow IR read goes through `store.db`, whose implementation is an unconditional throw.

The ratchet flags it correctly — `triage.ts: 7 -> 8`, exit 1. Nothing ran it, so `main` is red on a guard nobody consults.

This is not a criticism of #3114's author. A written warning at the exact line, with a test behind it, was overwritten within hours by a well-intentioned change. **That is what unenforced prose does under fleet pressure** — and it is the fourth time this class has landed (#3051, #3080-family, #3114, plus my own draft in #3097 that failed to zero).

## What this PR does

1. **Restores the literal at `triage.ts:765`** with #3108's reasoning preserved, plus the condition for converting all three arms together (an awaitable listener, or lanes on the event payload).
2. **Wires `check:inert-sync-lanes` into `test:gate`**, beside the other cheap AST guards (`check-no-nohup`, `check-capacity-pool-id`, `check-mock-completeness`) it belongs with. It is a single-pass parse — no measurable cost.

## Census before / after

```
before:  COLUMN guards (the backlog):   26
after:   COLUMN guards (the backlog):   27
```

**Up by one, deliberately, and re-recorded with the reason.** Restoring an honest literal puts a real guard back on the backlog. That is the correct direction: a conversion that lowers the number without changing behaviour is precisely the failure this check exists to stop, and the census must not reward it.

I would rather the number go up and mean something.

## Sequencing note

`test:gate` passes on this branch **because** the triage site is restored. Wiring the ratchet without that fix would turn `test:gate` red for everyone on an unrelated commit, so the two belong in one PR.

**#3122** (open) is a strict improvement on top: it restores 13 further guards that went invisible when #3109 laundered them through `mergeParkedColumns`. It will raise the baseline again for a detection reason, not a regression — worth merging after this so the ratchet enters CI at its full sensitivity.

## Verification

`test:gate` exit 0 **with the ratchet inside it** · `check:inert-sync-lanes` exit 0 · `lifecycle-column-census --strict` exit 0 at the re-recorded baseline · all 25 triage suites green (374 tests) · `pnpm lint` clean.


gsxdsm:

Only `scripts/lib/lifecycle-column-census-baseline.json` conflicted; `package.json` and `triage.ts` applied cleanly. Resolved by taking main's copy and regenerating rather than hand-merging — that file is generated, and a hand-merge produces a count matching neither branch while looking plausible.

## The rebase surfaced a drop in your new gate

After rebasing, `check:inert-sync-lanes` reported:

```
inert-sync-lane: total fell 29 -> 16. Re-record with --update-baseline.
```

Work landing on main since this branch was cut converted thirteen of the sites your gate watches. Re-recorded and amended in, so the retired allowances cannot be regrown into — the same contract the census and lane-wiring ratchets enforce.

Worth calling out because it is a good sign for the gate: it caught a real drop on its very first rebase, which is exactly the behaviour that makes a ratchet worth wiring into CI rather than running by hand.

## Verification

- `check:inert-sync-lanes` — exit 0 both before and after the re-record
- `triage` suites — **374 pass**
- `census --strict` — exit 0

gsxdsm:
Rebuilt as wiring-only and continued in the replacement PR.

#3126 landed the triage revert — the other half of this branch — so keeping both would have been a redundant diff. What remains is the part still missing: `check:inert-sync-lanes` is defined in package.json and invoked by nothing, so it only runs when a human types it.

That is worth noting as its own outcome: the sequence #3108 (flag) → #3114 (convert the flagged arm) → #3126 (revert) cost three PRs, and the ratchet correctly flagged it the entire time. A wired check would have made it one failed run.

PR #3128, opened 2026-07-31
fix(scheduler): resolve task:updated / task:deleted lanes asynchronously (scheduler inert 5 → 0)

gsxdsm:
The last inert guards in `scheduler.ts`. Independent of my other branches.

## Inert-guard ratchet

| Scope | Before | After |
|---|---:|---:|
| `scheduler.ts` | 5 | **0** |
| total | 12 | **7** (triage.ts 8 → other worker; executor.ts 4 → #3112) |

## The live bug

These read `resolveTaskParkedColumnsSync`, which answers with the **default** workflow in production. On a renamed board the scheduler **never woke** on unpause or planning-finish, and a **deleted blocker never unblocked its dependents** — the card sat behind a task that no longer existed.

## The criterion, restated because I got it wrong before

**What blocks a guard is whether its answer is consumed synchronously — not whether the enclosing listener is declared sync.** I assumed the latter earlier in this program and reverted for it.

All three fail that test: two only gate `schedule()`, which is itself `async`, fire-and-forget and re-entrance-guarded; the third already sits below an `await getSettings()`. The edge-trigger bookkeeping (`planningTaskIds.delete`) **stays synchronous** on purpose — deferring *that* would let a second update re-enter the branch.

## The union is load-bearing, not defensive

Post-U11 the default lineage has no `triage` column, so a **resolved** answer returns `intake: "todo"` where the inert path fell back to `"triage"`. Converting without unioning the legacy ids silently **narrowed** the wake set and stopped waking cards in a legacy-named lane — caught by *"schedules when planning clears in triage"*.

**A resolved conversion must be a superset of what it replaces, or it is a behaviour change wearing a vocabulary change's clothes.** That's the reusable lesson here.

## Tests

- Drained with the repo's existing **`flushAsyncHandlers`** helper — written for exactly this fire-and-forget shape — rather than loosening any assertion.
- **The characterization test flipped, as designed.** `workflow-scheduler-parked-columns-live-e2e.pg.test.ts` asserted *"a dependent in a RENAMED hold column is NEVER unblocked"*, with its author noting: *"expected to flip to null the moment the resolver is fixed — and that flip is the whole point of writing it down."* It flipped. Inverted to a REGRESSION case so the assertion holds the fix rather than the defect; it now matches its own CONTROL arm, which still guards against a vacuous pass.

## Verification

- 21 scheduler suites — **361 green**, including the live PostgreSQL e2e
- **`pnpm test:gate` green**; eslint and `tsc` clean

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed task scheduling after unpausing tasks or completing planning.
  - Corrected workflow behavior when hold and intake lanes are renamed.
  - Ensured dependent tasks are unblocked when their blockers are deleted.
  - Improved reconciliation of parked and in-progress tasks across workflow changes.

- **Tests**
  - Added coverage for asynchronous scheduling events, renamed lanes, parked tasks, and blocker deletion scenarios.

### DIFF
--- a/.changeset/scheduler-async-lanes.md
+++ b/.changeset/scheduler-async-lanes.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Renamed hold and intake lanes now wake the scheduler and unblock dependents correctly.
+category: fix
+dev: The `task:updated`/`task:deleted` handlers resolve lanes asynchronously; the wake set unions legacy ids so it stays a superset.
--- a/packages/engine/src/__tests__/scheduler-auto-claim-invalidation.test.ts
+++ b/packages/engine/src/__tests__/scheduler-auto-claim-invalidation.test.ts
@@ -1,6 +1,7 @@
 import { describe, expect, it, vi } from "vitest";
 import type { TaskStore } from "@fusion/core";
 import { Scheduler } from "../scheduler.js";
+import { flushAsyncHandlers } from "./_flush-async-handlers.js";
 
 function createStore() {
   const listeners = new Map<string, ((payload: unknown) => void)[]>();
@@ -114,7 +115,7 @@ describe("Scheduler auto-claim snapshot invalidation", () => {
     expect(invalidate).toHaveBeenNthCalledWith(2, "task:updated");
   });
 
-  it("triggers immediate scheduling when a userPaused-only task is unpaused", () => {
+  it("triggers immediate scheduling when a userPaused-only task is unpaused", async () => {
     const { store, emit } = createStore();
     const scheduler = new Scheduler(store, {});
     const schedule = vi.spyOn(scheduler, "schedule").mockResolvedValue(undefined);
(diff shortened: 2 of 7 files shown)


gsxdsm:
**Rebased onto current `main`, and I corrected a false green I had introduced here.**

My original baseline re-record set `triage.ts` to **8**. That silently blessed the regression from #3114 that #3126 exists to revert — the exact failure the script's own warning names:

I did it anyway, because I ran `--update-baseline` after my change without checking *which* deltas it was absorbing. A baseline update is a claim about every file in it, not just the one you touched.

**Corrected:** the baseline now drops `scheduler.ts` (18 → absent, which is what this PR earns) and **holds `triage.ts` at 7**, so its regression stays visible.

**Consequence, stated plainly: this PR's ratchet check is RED, and that is correct.** `main` is red for the same reason. It should merge *after* #3126 reverts the inert triage arm — at which point the check goes green with no further change here. I would rather ship a red PR that reports the truth than a green one that hides another worker's regression.

Verification after rebase: 21 scheduler suites, **361 green**, including the live PostgreSQL e2e.

PR #3129, opened 2026-07-31
fleet(dashboard): taskRevert 2 → 0 — the recorded blocker named the wrong variable

gsxdsm:
The largest remaining census cluster. Deferred twice, with a blocker that turns out to be false **in the same component where its counter-example already lives**.

## What the earlier notes got right

`detailColumnFlags` describes the **modal's own task**, and the column classified here belongs to a **neighbour**. Supplying it would answer *"is this neighbour finished?"* with a different row's traits — wrong on data, not merely stale on vocabulary. That reasoning stands and I kept it.

An earlier pass also converted this, left the parameter unsupplied, and **reverted it** — correctly. An unsupplied optional parameter is strictly worse than the literal: the guard is gone, the census counts a conversion, and the behaviour is the legacy fallback forever. That rule is why the wiring ships in this same commit.

## What the conclusion got wrong

> "A correct conversion needs per-**neighbour** flags — which the modal does not have and should not fetch mid-render."

`columnFlagsByTaskId` is a per-task map. It is **already a prop** of `TaskDetailModal` (declared :367, destructured :727), and the call site at :992 sits **below** that destructure.

And `TaskDetailModal` already uses it exactly this way, for the near-duplicate canonical:

```ts
columnFlagsByTaskId?.get(nearDuplicateCanonical.id)
```

…under a note observing that *its* blocker had been *"asserted from the shape of the problem rather than tested against what was in scope."* Same assertion, one function over. So the supplier the earlier note went looking for exists, is per-neighbour, and needs no fetch.

## What it fixes

This lookup skips **finished** candidates so a done/archived prior undo attempt never renders as an active "Undo task" link. On a board that renames those lanes it matched neither — a finished undo task kept rendering as open, which is precisely the stale affordance the function's own header says it exists to prevent.

## Census

| | before | after |
|---|---|---|
| `taskRevert.ts` | 2 | **0** |
| repo backlog | 17 | **15** |

## Measured

- 4 new cases; `taskRevert.test.ts` **11/11 pass**.
- **MUTATION**: restoring the literal pair fails the renamed case.
- **The negative is load-bearing.** The map is fail-soft, so a candidate it does not cover must still be treated as **open**, not skipped. A conversion that skipped unknown candidates would *hide live undo links* — failing in the direction nobody reports.
- A **control** pins that an unwired caller (no flags at all) still skips the legacy ids, so the optional parameter cannot regress default boards.
- `TaskDetailModal` suites — **31 files / 664 tests pass**.
- `tsc --noEmit -p tsconfig.app.json` clean; census `--strict`, `check-lane-wiring`, `check-fnxc-future-dates` clean.

## Pattern worth noting

This is the fourth deferral this session whose stated blocker had dissolved or misidentified itself, and the second where the counter-example was already in the same file. The common shape: a note records *why* something is blocked, is accurate when written, and is never re-checked — so the block outlives its cause. Re-reading them cost minutes each and returned two real conversions.

### DIFF
--- a/packages/dashboard/app/components/TaskDetailModal.tsx
+++ b/packages/dashboard/app/components/TaskDetailModal.tsx
@@ -989,7 +989,9 @@ export function TaskDetailContent({
    * (open board columns only) so a done/archived/soft-deleted prior undo attempt never renders as
    * an active "Undo task" link — that would be a stale/leftover affordance.
    */
-  const openUndoTask = findOpenUndoTaskForSource(tasks, workingTask.id);
+  /* FNXC:WorkflowResolvedColumns 2026-07-31-23:20: the CANDIDATES' own flags, keyed by id — the same
+     per-neighbour supply this component already uses for the near-duplicate canonical above. */
+  const openUndoTask = findOpenUndoTaskForSource(tasks, workingTask.id, columnFlagsByTaskId);
 
   const previousInitialTabRef = useRef<TabId | undefined>(initialTab);
   const taskColumnRef = useRef(task.column);
--- a/packages/dashboard/app/utils/__tests__/taskRevert.test.ts
+++ b/packages/dashboard/app/utils/__tests__/taskRevert.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, it } from "vitest";
 import type { Task } from "@fusion/core";
-import { isTaskReverted } from "../taskRevert";
+import { isTaskReverted, findOpenUndoTaskForSource } from "../taskRevert";
 
 describe("isTaskReverted", () => {
   it.each([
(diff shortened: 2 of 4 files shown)


gsxdsm:
**Rebased onto main and pushed.** Branch was not checked out in any worktree. Only the generated census baseline conflicted; all three source files applied cleanly — resolved by taking main's copy and re-running `--strict --update-baseline`.

## You corrected a call I made, and you were right

I claimed this file earlier in the phase, inspected it, and **released it without converting** — I read the recorded note as establishing this was accurate debt blocked on a per-neighbour flag map that did not exist, and argued it should stay counted rather than be marked or converted.

Your title names the actual defect: *the recorded blocker named the wrong variable*. `flagsByTaskId` keyed by task id is exactly the thing the note said did not exist, and the reason it looked impossible was a parameter that could not be supplied from where `detailColumnFlags` was derived — a call-site ordering problem, not a missing capability.

I took a documented blocker at face value because its reasoning was internally coherent. It was coherent and wrong, and the check I did not do was "is the named obstacle actually the obstacle?" That is the same trap this program keeps finding one level down — a note that reads as settled because it is well argued.

Worth recording in the file's own history, since I publicly argued the opposite on this PR's file: **`taskRevert.ts` was convertible all along.** Census `17 → 16`, and it drops out of `byFile` entirely.

## Verification

- `taskRevert.test.ts` — **11 pass** (up 54 lines of new coverage)
- `tsc --noEmit -p tsconfig.app.json` — **0 errors**
- `census --strict` — exit 0, total **16**, `taskRevert.ts` absent
- diff is +88/-24 across three files — no mass deletion

PR #3131, opened 2026-07-31
census: --triage under-reported documented deferrals by 44%

gsxdsm:
## The triage list was pointing workers at already-decided sites

Follow-up to #3097 (merged), measured on `main` rather than assumed. Of the **nine** guards `--triage` listed as unexamined, **four carried a full deferral note** that the marker set did not match:

| site | what its note actually says |
|---|---|
| `moves.ts:346` | *"THIS ARM STAYS INLINE, deliberately"* — naming it moves the tally `archived-column-gate-parity.test.ts` pins, and that guard exists to stop exactly this split brain |
| `ResearchTaskActionModal.tsx:66` | *"sized here rather than faked"* — the honest fix is a data-fetch change, not prop threading |
| `audit-ops.ts:231` | *"Not converted here because ..."* — must land with `getLiveTaskColumn` or the pair disagrees |
| `mission-store.ts:2343` | *"do not convert"* |

A **44% false-positive rate** on a list headed *"this is the list to pick work from"*.

## Why this direction of error is the dangerous one

Under-reporting sends a worker at a site whose owner already wrote down why it must not move. That is not hypothetical — it is precisely the sequence behind **#3114**, which converted the arm **#3108** had flagged hours earlier, with both blockers named and a test behind them. The tool I shipped to prevent workers burning effort on decided sites was quietly nominating four of them.

Over-reporting has the opposite failure: it hides real work. So the added phrases are specific to *declining a conversion* — `STAYS INLINE`, `not converted`, `do not convert`, `Recorded rather than`, `sized here`, `in dead code` — rather than generic words like `deliberately` on its own, which appears in plenty of notes that are not deferrals.

## Verified in both directions

- **Newly counted (4):** each independently read and confirmed to carry a real deferral note.
- **Still listed (4):** `eval-signal-collector.ts`, `async-comments-attachments.ts`, `auto-merge-finalization.ts`, `scheduler.ts` — the first three are fallback arms of already-converted seams (`x ? resolved : literal`), the fourth is the known synchronous-listener site. None is hidden convertible debt.
- `lifecycle-ops.ts` moved to documented, correctly — its note explains the polling-replica path is legacy-SQLite-only dead code.

```
before:  documented 8  / unexamined 9
after:   documented 13 / unexamined 4
```

## Census before / after

```
before:  COLUMN guards (the backlog):   17
after:   COLUMN guards (the backlog):   17
```

Unchanged, as required — `--triage` is opt-in and touches no count and no exit code. `--json` and `--strict` verified unaffected.

## Verification

`test:gate` exit 0 · `lifecycle-column-census --strict` exit 0 · `--json` exit 0 · headline output unchanged · `pnpm lint` clean. One script; no production file touched.

## Still open, and the reason the ratchet matters

`main` is **red on `check:inert-sync-lanes`** right now (#3114's inert conversion) and nothing in CI runs it. **#3127** fixes both halves — restores the flagged literal and wires the ratchet into `test:gate`. **#3122** restores 13 guards laundered through `mergeParkedColumns`; **#3119** attaches lanes at all five live emitters so the fallback path stops being taken.

### DIFF
--- a/scripts/lifecycle-column-census.mjs
+++ b/scripts/lifecycle-column-census.mjs
@@ -116,7 +116,20 @@ guard whose text marks a deliberate deferral. It cannot tell a good reason from 
 far above its guard reads as unflagged. It is a triage aid for choosing work, never a gate — which is
 why it is opt-in and why nothing downstream consumes it.
 */
-const FLAG_MARKERS = /FLAGGED|LEFT COUNTED|left counted|deliberately NOT converted|Recorded instead|Left as a literal|DELIBERATE-LITERAL|accurate debt|blocked on/;
+/*
+FNXC:LifecycleColumnCensus 2026-08-01-04:20 (the marker set under-reported, which is the harmful direction):
+Measured after shipping: of nine sites `--triage` listed as unexamined, FOUR carried a full deferral
+note my markers simply did not match — `moves.ts` ("THIS ARM STAYS INLINE, deliberately"),
+`ResearchTaskActionModal.tsx` ("sized here rather than faked"), `audit-ops.ts` ("Not converted
+here because ..."), `mission-store.ts` ("do not convert"). A 44% false-positive rate on a list
+headed "this is the list to pick work from".
+
+Under-reporting is worse than over-reporting for this tool: it sends a worker at a site whose owner
+already wrote down why it must not move, which is exactly the sequence that produced #3114 (converting
+an arm #3108 had flagged hours earlier). Over-reporting would hide real work, so the added phrases are
+specific to declining a conversion rather than generic words like "deliberately" on its own.
+*/
+const FLAG_MARKERS = /FLAGGED|LEFT COUNTED|left counted|deliberately NOT converted|Recorded instead|Recorded rather than|Left as a literal|DELIBERATE-LITERAL|accurate debt|blocked on|STAYS INLINE|[Nn]ot converted|do NOT convert|do not convert|sized here|in dead code/;
 
 /** Split the column guards into documented-deferral vs unexamined, by comment proximity. */
 function triageFindings() {


gsxdsm:
**Heads-up: #3124 contains the same fix**, derived independently, in the same file and the same function. Commit `229a3aeb4a`, "the deferral-marker list under-matched". We even found it the same way — by working the list and discovering the sites it offered were already decided — and landed on overlapping evidence (`moves.ts` *"THIS ARM STAYS INLINE"*, `ResearchTaskActionModal.tsx` *"SIZED, NOT CONVERTED"*, the parity-test references).

Two people independently concluding the marker set under-matches is good evidence it does. But both of us edit `FLAG_MARKERS` in `scripts/lifecycle-column-census.mjs`, so whichever merges second gets a conflict.

**Recommendation: merge this one first.** It is single-file and does exactly one thing; #3124 is larger (`--claims`, plus four `DELIBERATE-LITERAL` site marks, plus this). I will rebase #3124 on top and drop my duplicate commit — no objection, and no need for you to do anything.

Two additions from my side you may want to fold in, since your list and mine differ slightly:

- I also added `non-renameable system column` and `archived-column-gate-parity`, which catch the three `packages/core` sites that defer to that guard **by name** rather than restating its reasoning. Those were the most dangerous misses for me: five of the six files my list offered were `archived` sites inside the parity inventory, where converting the TypeScript half alone is the documented split brain.
- I deliberately avoided a case-insensitive flag, since `i` would let "flagged" match casual prose anywhere in a 40-line window and reclassify live guards as reviewed — the same failure in the other direction.

## Worth recording: this collision was invisible to the tool built to prevent it

#3124 adds `--claims`, which maps remaining census files to the open PRs touching them, precisely because `self-healing.ts` took three overlapping conversions and `executor.ts` took two in one afternoon. It would **not** have caught this one: it keys on files that hold census guards, and `scripts/lifecycle-column-census.mjs` is the census, not a censused file. Tooling for the backlog does not cover work on the tooling itself. That is a real limitation and I would rather write it down than let it read as solved.

gsxdsm:
Superseded — `main` already carries a broader marker set than this PR proposed, and reports **13 documented / 2 unexamined** where mine reached 13/4.

Main's set picked up `STILL A LITERAL`, `archived-column-gate-parity`, `non-renameable system column` and `SIZED, NOT` alongside the `STAYS INLINE` / `not converted` / `do not convert` phrasings I had added. Strictly better coverage, same direction, same reasoning — under-reporting a documented deferral sends a worker at a site whose owner already wrote down why it must not move.

Nothing to fold. Closing.

PR #3133, opened 2026-07-31
fix(test): main is red on the census baseline fixture — it pinned a count the fleet has since converted

gsxdsm:
## `main` is red right now

Measured on a clean detached `origin/main` checkout, nothing from any branch applied:

```
 FAIL  src/__tests__/lifecycle-column-census.test.ts > the baseline can always be re-recorded
       > driven end to end > exits 0 and REWRITES the baseline under --update-baseline, even when the count rose

AssertionError: expected 1 to be greater than 1

 Tests  1 failed | 39 passed (40)
```

## The subject of the test never changed — the fixture expired

Both re-record cases pinned a **file** and a **number**: stale baseline says `self-healing.ts: 1`, tree has more, therefore a rise. Conversions have since taken that file to exactly **1**, so `toBeGreaterThan(1)` fails while the behaviour under test — *does `--update-baseline` write when the count rose?* — is completely unaffected.

This is fixture rot with a guaranteed expiry date. **The backlog shrinking is the point of this program**, so any fixture keyed to a specific file's guard count will expire; the only open question was which cycle. It expired this one.

## The fix derives what it used to hardcode

Ask the census which file currently holds guards, then construct a baseline **one below that file's real count**. The rise is manufactured rather than assumed, and the assertion becomes exact:

```ts
const { file, count } = fileWithGuards();
const stale = { …, byFile: { [file]: count - 1 }, … };
…
expect(written.byFile[file]).toBe(count);   // was: toBeGreaterThan(1)
```

`toBe(count)` is also a stronger claim than the inequality it replaces, which was only ever a proxy for "the rewritten pin carries the tree's real number".

This is the same discipline as the self-syncing fixture directly below it in the file (#3106), which syncs its temp baseline to the tree before inflating it — for exactly this reason. That one survived; these two were the half that still hardcoded.

## Verification

- **43 passed** on this branch
- **1 failed | 39 passed** on unmodified `origin/main` — the red this fixes
- Differential, so the de-rotted test still catches the bug it was written for: disabling `--update-baseline` in the CLI gives **1 failed | 42 passed**. The original ordering defect (a rise exiting before the write, so the one command whose whole job is re-recording could not re-record) is still caught.

## Scope

One test file. No production code, no gate, no baseline change, no changeset (internal test). Split out of #3124 deliberately so it can merge on its own while main is red, rather than waiting behind a larger census PR.

### DIFF
--- a/packages/engine/src/__tests__/lifecycle-column-census.test.ts
+++ b/packages/engine/src/__tests__/lifecycle-column-census.test.ts
@@ -468,10 +468,35 @@ describe("the baseline can always be re-recorded", () => {
       }
     }
 
+    /*
+    FNXC:LifecycleColumnCensus 2026-07-31-23:59 (the fixture rotted, and it took main red with it):
+    These two cases pinned the FILE `self-healing.ts` and the NUMBER 1 — "stale baseline says 1, tree
+    has more". Conversions took that file to exactly 1, so `toBeGreaterThan(1)` failed and `main` went
+    red on a test whose subject (does `--update-baseline` write on a rise?) had not changed at all.
+
+    Measured on a clean detached `origin/main`: 1 failed | 39 passed, with nothing from this branch
+    applied. The backlog shrinking is the POINT of this program, so any fixture keyed to a specific
+    file's count is guaranteed to expire — the only question is which cycle.
+
+    So the target and the number are now DERIVED: ask the census which file currently holds guards,
+    then construct a baseline one below that file's real count. The rise is manufactured rather than
+    assumed, and the assertion is exact (`toBe(real)`) instead of an open inequality that was only ever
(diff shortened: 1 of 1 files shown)


gsxdsm:
Correction to the verification numbers in the body: **this branch is 40 passed, not 43.**

43 is the count on `u12/census-claims` (#3124), which adds three `--claims` tests on top. This PR is cherry-picked onto plain `origin/main`, so the correct before/after is:

| | result |
|---|---|
| unmodified `origin/main` | **1 failed \| 39 passed** (40) |
| this branch | **40 passed** (40) |
| this branch, with `--update-baseline` disabled in the CLI | **1 failed \| 39 passed** — the differential still holds |

The claim the numbers support is unchanged; I carried the total over from the branch I developed it on. Flagging rather than quietly editing the body, since the whole point of stating measured numbers is that they are checkable.

PR #3137, opened 2026-07-31
fix(executor): planner-evacuation lanes come from the emitter — executor leaves the inert list (16 → 12)

gsxdsm:
`executor.ts` was the last file besides `triage.ts` and `scheduler.ts` on `check-inert-sync-lanes`, holding **4 guards that read as converted and behave as literals**. Neither cause turned out to be "needs an async resolver".

## 1. Two of the four were in code with no caller

`isPlannerColumnFor` is a **private method with zero production callers**. `tsc` reports it unused; the only things reaching it were two tests casting through `executor as unknown as { … }`, which is exactly what let it look alive. Its doc comment described the planning-evacuation branch — but that branch calls `isBackwardMoveOutOfPlanning` and never called this.

Deleted, along with the two tests whose subject it was. Converting guards in unreachable code would have "fixed" behaviour that cannot run and left two more sites to maintain; a test whose subject has no caller pins nothing.

## 2. The other two no longer need to resolve anything

`isBackwardMoveOutOfPlanning` resolved its own lanes via `resolvePlannerLanes`, whose selection reader returns `undefined` unconditionally under PostgreSQL — so it answered with the **default board for every task**, and both its guards were inert.

Its comment justified the sync resolver by the synchronous `task:moved` emitter. That was true and **is no longer binding**: the emitter now resolves lanes once, asynchronously (`moves.ts` → `resolveWorkflowIrForTask`), and hands them on the payload — which #3112 already reads in this same listener. Reading a parameter is as synchronous as reading `from`, so nothing reorders and no listener resolves.

`lanes` is **required, not optional**. An optional parameter that the one production caller happens to pass is the seam-with-no-supplier shape this program keeps finding; required means a future caller fails typecheck instead of silently getting a default board. When the emitter itself could not resolve, the legacy ids answer — exactly what `resolvePlannerLanes` degraded to anyway.

## Measured

| | before | after |
|---|---|---|
| `check-inert-sync-lanes` | **16** guards, 3 files | **12** guards, 2 files |
| `executor.ts` on that list | 4 | **0 — off the list** |
| census | 18 | 18 (`--strict`: every file matches baseline exactly) |

**The census is deliberately unchanged.** This targets the inert population, which the census cannot see by construction: those guards already read as converted. That gap is the argument in #3082 — 12 guards still behave as literals while the census shows them as done.

## The producer half, which I nearly shipped without

The predicate's own suite covers it thoroughly — and every case calls it **directly**. Mutation testing exposed that this proves nothing about the listener: replacing the listener's `lanes` argument with `undefined` left `planning-evacuation` at **20/20 green**. That is the fifth failure shape in this program's learnings verbatim — a converted consumer with an unconverted producer passing every instrument.

So there is now a case driving the **real listener** on a board whose planner lanes share no id with the legacy pair (`queued` holds, `drafting` intakes), withdrawing a card to a non-lifecycle column — the reported symptom (`todo -> Ideas`) in that board's vocabulary.

## Verification

- engine `tsc` — **0 errors**
- `executor-planner-lanes-resolved` — **12 passed**
- `executor-archive-releases-active-session` — **14 passed**; listener passing `undefined` → **1 failed | 13 passed**
- `planning-evacuation` + `triage-planning-wake` + archive suite — **47 passed**
- `check-inert-flag-seams`, `check-fnxc-future-dates`, census `--strict` — exit 0
- `eslint` on changed files — 0 errors

The predicate tests are also **stronger than before**, not merely adapted: they now build lanes with `toTaskMoveLanes`, the same function `moves.ts` uses for the payload. Previously they reached the predicate through the store-backed sync reader, so renamed-lane assertions passed in the harness while the real path could never see a renamed lane.

## Not done here

The inert baseline still reads 29 against a tree of 12 and the gate advises re-recording. I left it: a stale allowance is a real hazard, but re-recording is a one-line change that conflicts with every lane, and it should land once rather than in each of our branches.

### DIFF
--- a/packages/engine/src/__tests__/executor-archive-releases-active-session.test.ts
+++ b/packages/engine/src/__tests__/executor-archive-releases-active-session.test.ts
@@ -136,6 +136,44 @@ describe("archiving a task releases its active-session registry entries (FN-7717
   });
 
   /*
+  FNXC:WorkflowResolvedColumns 2026-07-31-23:59 (the LISTENER must forward the payload lanes):
+  THIS IS THE PRODUCER HALF, and it was missing.
+
+  `isBackwardMoveOutOfPlanning` now receives its lanes instead of resolving them, and its own suite
+  (`executor-planner-lanes-resolved`) covers the predicate thoroughly. But that suite calls the
+  predicate DIRECTLY, so it says nothing about whether the `task:moved` listener actually hands the
+  payload down. Measured: replacing the listener's `lanes` argument with `undefined` left
+  `planning-evacuation` at 20/20 green — the exact producer/consumer split this program's learnings
+  record as its fifth failure shape, where a converted consumer with an unconverted producer passes
+  every instrument.
+
+  So this drives the real listener. The board is renamed with NO legacy id anywhere near the planner
+  lanes (`queued` holds, `drafting` intakes), and the move withdraws a card from the hold lane to a
+  non-lifecycle column — the reported symptom, `todo -> Ideas`, in this board's vocabulary. The
+  evacuation must fire, which it can only do if the payload lanes reached the predicate: with
+  `undefined` the legacy pair answers, `from` is not `todo`/`triage`, and the branch declines.
+  */
+  it("evacuates a card withdrawn from a RENAMED planner lane — the listener forwards payload lanes", async () => {
+    const { executor, store } = makeExecutor();
+    const abort = vi
+      .spyOn(executor as any, "awaitAbortInFlightTaskWork")
+      .mockResolvedValue(undefined);
+    vi.spyOn(executor as any, "releasePreExecutionWorktree").mockResolvedValue(undefined);
+
+    store.emit("task:moved", {
+      task: makeTask("TASK-E2"),
+      from: "queued",
+      to: "ideas",
+      source: "user",
+      lanes: { intake: "drafting", hold: "queued", wip: "building", review: "checking", complete: "shipped", archived: "filed" },
+    });
+
+    await (executor as any).pendingTaskDisposals.get("TASK-E2");
+    expect(abort).toHaveBeenCalledTimes(1);
+    expect(String(abort.mock.calls[0]?.[1] ?? "")).toContain("out of planning");
+  });
+
+  /*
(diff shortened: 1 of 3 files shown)


gsxdsm:
**#3137 and #3140 both convert `isBackwardMoveOutOfPlanning` to read the emitter's `task:moved` lanes.** Measured both with the hardened inert-sync-lane check (the wrapper-taint version in #3122, which sees guards laundered through a wrapper):

| | scheduler | triage | executor | total |
|---|---|---|---|---|
| **#3137** | 18 | 7 | **0** | 25 |
| **#3140** | 18 | 7 | **4** | 24 |

**#3137 removes `executor.ts` from the sync-resolved list entirely.** #3140 leaves its four guards still reading `resolvePlannerLanes`, which resolves through `resolveTaskWorkflowIrSync` and answers with the default board under PostgreSQL. Both are real conversions — neither reaches for the sync resolver — but by this measure #3137 finishes the file and #3140 does not.

(The totals differ from `main`'s because the two branches sit on different bases; only the executor column is comparable, and it is the one that differs.)

Worth deciding which to keep rather than landing both: they target the same predicate, so whichever merges second will conflict, and if #3140 lands after #3137 it could reintroduce the four guards #3137 removed.

Two things I would keep from #3137 regardless of which survives:

1. It **mutation-tested the wiring** — *"replacing the listener's `lanes` argument with `undefined` left…"* — which is the check that matters here. A predicate that accepts lanes but whose listener never forwards them passes every unit test while behaving exactly as before; that is the shape that produced #3051 and #3114.
2. Its fixtures use lanes matching **no legacy id** (`queued` holds, `drafting` intakes), so the tests exercise the payload and nothing else. A fixture using `todo`/`in-progress` passes whether or not the payload is read.

Related: **#3135** attaches lanes at the last two live emitters (`task-update.ts`, `update-task-deps.ts`), which is what makes the payload actually arrive on those paths — without it a payload-reading predicate still falls back to the default board whenever those emitters fire. **#3122** is what makes the wrapper-laundered guards visible at all; on `main` today `scheduler.ts` reports **0** while 13 such guards remain.

PR #3139, opened 2026-07-31
fix(fnxc): main is red — #3128 added five future-dated FNXC stamps

gsxdsm:
## `main` is red on `check-fnxc-future-dates`

```
packages/engine/src/__tests__/workflow-scheduler-parked-columns-live-e2e.pg.test.ts: 1, allows 0
```

#3128 stamped five new comments `2026-08-01`.

## Not the timezone case

The gate's own header documents a real edge: a baseline recorded at local midnight can go stale against a UTC runner, which is why a *drop* there auto-tightens instead of failing. That is not this.

```
UTC now:   2026-07-31 12:42
local now: 2026-07-31 05:42
```

Checked explicitly rather than assumed. These stamps are ahead of the calendar in **both** zones, so CI is red too — this is a rise, which the gate fails hard by design.

## Fix

Re-stamped the five to `2026-07-31`, **preserving each one's clock time** so the ordering between them is unchanged — the stamps are a narrative trail, and renumbering them would scramble the sequence they encode.

The **three pre-existing** future-dated stamps in `scheduler.ts` are left alone. They sit inside the recorded baseline, and re-stamping them would tighten that count as a side effect of an unrelated fix — the sort of incidental ledger movement I have spent this session arguing against.

## Measured

- `check-fnxc-future-dates` — **exits 1 on `main`, 0 here**.
- `src/__tests__/scheduler*` — **14 files / 144 tests pass**.
- `tsc --noEmit -p packages/engine` clean; `check-inert-sync-lanes` and census `--strict` clean.

Comment-only. No behaviour change, no census movement.

### DIFF
--- a/packages/engine/src/__tests__/workflow-scheduler-parked-columns-live-e2e.pg.test.ts
+++ b/packages/engine/src/__tests__/workflow-scheduler-parked-columns-live-e2e.pg.test.ts
@@ -168,7 +168,7 @@ pgDescribe("scheduler parked-column resolution against a live store", () => {
 
   it("REGRESSION — a dependent in a RENAMED hold column IS unblocked when its blocker is deleted", async () => {
     /*
-    FNXC:WorkflowResolvedColumns 2026-08-01-02:20 (fleet — the flip this test was written to catch):
+    FNXC:WorkflowResolvedColumns 2026-07-31-02:20 (fleet — the flip this test was written to catch):
     This was a CHARACTERIZATION of the inert sync read: `resolveTaskParkedColumnsSync` answered
     `{ hold: "todo" }` for a board whose hold column is `backlog`, so the reconciliation queried a
     column that does not exist, found no dependents, and left `blockedBy` pointing at a deleted task.
--- a/packages/engine/src/scheduler.ts
+++ b/packages/engine/src/scheduler.ts
@@ -450,7 +450,7 @@ function mergeParkedColumns(
 }
 
 /*
-FNXC:WorkflowResolvedColumns 2026-08-01-01:40 (fleet):
+FNXC:WorkflowResolvedColumns 2026-07-31-01:40 (fleet):
 The ASYNC twin. The sync one below cannot answer for a custom workflow in production, so any guard that
 can reach this one must.
 
@@ -477,7 +477,7 @@ async function resolveTaskParkedColumns(store: TaskStore, taskId: string): Promi
       archived,
       terminal: new Set([complete, archived]),
       /*
-      FNXC:WorkflowResolvedColumns 2026-08-01-02:05 (fleet):
+      FNXC:WorkflowResolvedColumns 2026-07-31-02:05 (fleet):
       The wake set UNIONS the legacy ids rather than replacing them, and that is load-bearing rather
       than defensive. Post-U11 the default lineage has no `triage` column, so a RESOLVED answer returns
       `intake: "todo"` where the old inert path fell back to `"triage"`. Converting without the union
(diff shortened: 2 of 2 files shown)


gsxdsm:
## Integration-verified against current `main`, and it is the only thing standing between `main` and a green ledger

`main` has exactly one red left:

```
check-fnxc-future-dates  exit 1   <- this PR fixes it
check-inert-sync-lanes   exit 0
check-lane-wiring        exit 0
lifecycle-column-census  exit 0   (cleared when #3129 merged with its re-record)
```

Merged this PR onto current `main`: **fnxc goes green**, and census / inert / lane-wiring stay green.

Then merged the other three open PRs (#3142, #3118, #3134) on top of it. All four together:

| check | result |
|---|---|
| `check-fnxc-future-dates` | ✅ |
| `lifecycle-column-census --strict` | ✅ |
| `check-inert-sync-lanes` | ✅ |
| `check-lane-wiring` | ✅ |
| `tsc` — engine / core / dashboard | ✅ 0 errors each |
| `pnpm test:gate` | ✅ **732 tests, 29 files, zero FAIL lines** |

No merge conflicts between the four, in any order I tried.

## A correction to something I said earlier

I reported that several of my branches "carried spurious reverts of other people's landed work". **That was wrong.** It came from reading a two-dot `git diff origin/main origin/<branch>`, which compares trees and renders `main`'s *newer* commits as deletions on the branch side. The correct check is three-dot (`origin/main...origin/<branch>`), which shows only what the branch changed since the merge base — and on that measure every branch was clean the whole time:

```
fix/fnxc-future-stamps-3128      scheduler.ts + one pg test
fix/scheduler-flag-reason-stale  scheduler.ts
fleet/executor-lanes-from-payload  one executor test
fleet/research-modal-archived-shape  one dashboard component
```

One branch *was* genuinely broken — this one, sitting at a stale commit with 539 real deletions — and rebuilding it was correct. The other three I rebuilt on a false premise. No harm resulted, but the diagnosis was mine and it was wrong.

gsxdsm:
Closing as a **duplicate of #3145**, which fixes the same `main` red more completely.

`--claims` (#3124) flagged the collision: `scheduler.ts` is held by **#3145, #3142 and #3139** at once. I checked rather than assumed — applied #3145's patch to current `main` and ran the gate:

```
check-fnxc-future-dates after #3145 = 0
```

It normalizes **8** stamps to my 5, and updates `scripts/lib/fnxc-future-dates-baseline.json` alongside.

## The difference was a deliberate scoping call, and theirs is the better one

I re-stamped only the **five** stamps #3128 added, and left the **three** pre-existing future-dated stamps alone — reasoning that re-stamping baselined entries tightens that ledger as a side effect of an unrelated fix.

#3145 re-stamps all eight **and** records the tightening in the same commit. That is the stronger version of the same instinct: the ledger movement is not a side effect if you record it deliberately. Mine avoided the ledger; theirs owns it.

Two PRs racing on the same red is waste and guarantees a conflict for whoever merges second, so this one goes.

## What I am keeping

**#3142** also touches `scheduler.ts`, but only comment text at the two lane guards — no stamp lines, no overlap with #3145's edits. It should merge cleanly either order.

I checked #3141 as well, since `--claims` shows it and #3142 both holding `triage.ts`. #3141 documents the same inert-resolver blocker there with `FLAGGED` / `Left inert and COUNTED` markers. If it lands first, my one-line `LEFT COUNTED` addition to that file becomes redundant and I will drop it from #3142 rather than have two notes saying the same thing — the duplicate-note problem I fixed in #3126.

PR #3141, opened 2026-07-31
docs(triage): its inert lane guards depend on the sync resolver FAILING — measured, not converted

gsxdsm:
I tried the conversion that took `executor.ts` off `check-inert-sync-lanes` in #3137, on the file holding the largest remaining inert cluster. **It does not transfer, and the tests are what proved it — not my reading of the code.** Comments only; no behaviour change.

## What looked free

`triage.ts`'s 7 guards are inert for the usual reason (`resolvePlannerLanes` → `resolveTaskWorkflowIrSync` → default board under PostgreSQL), the file **already imports** `resolvePlannerLanesForTaskAsync`, and the two most promising sites sit inside `async` functions. Mechanical swap, apparently.

| site | result of swapping in the async twin |
|---|---|
| `recoverApprovedTask` | **13 failures** in `triage.test.ts` |
| `sweepStalePlanningStatuses` | fails *"clears a stale planning status so triage can re-pick the card"* |

## One cause, and it inverts the usual argument

A stored pre-U11 row sits in `triage`.

- The **sync** reader cannot resolve under PostgreSQL, so it falls back to `LEGACY_PLANNER_LANES` — whose `intake` **is** `"triage"`. The row matches.
- The **async** reader *succeeds* and returns the workflow's real lanes, where post-U11 intake and hold are **one merged column**. So `"triage"` matches nothing, and both the approved-plan recovery and the FN-8596 strand-clearing silently stop happening — for precisely the migration case they exist to serve.

**These guards' current correctness depends on the sync resolver's failure mode.** Everywhere else in this program the sync resolver is the bug; here it is load-bearing, and "convert to the async twin" would be a regression that the census would score as progress.

## Why this is not a resolver choice

The legacy `triage` acceptance has to become **explicit** before either site can resolve properly. The pattern already exists in this file: `recoverApprovedTask` scopes its legacy acceptance to an **orphaned** `triage` row — one whose workflow does not declare a `triage` column — using `resolveWorkflowIrForTaskWithProvenance`, precisely because a custom workflow may legitimately name a non-intake lane `triage`.

Applying that to these sites is a compatibility change with its own surface list (every row in the sweep, plus the R11 contract that keeps `triage` a legal stored column id). That is a real piece of work, not a fleet conversion, so it is **sized here rather than smuggled in**.

## Disposition

Both sites stay inert and **counted**, and deliberately **not** marked `DELIBERATE-LITERAL` — the gate should keep pointing here. What lands is the measurement and the named solution, so the next person does not repeat the swap and discover the same 14 failures.

## Verification

- `triage.test.ts` + `recover-approved-intake-post-u11` — **237 passed**
- engine `tsc` — **0 errors**
- `check-inert-sync-lanes` — exit 0 (unchanged at 7 for this file, by design)
- census `--strict` — exit 0, unchanged

## Separately: `main` is red on `check-fnxc-future-dates`

Not from this branch — measured on a clean detached `origin/main`:

```
packages/engine/src/__tests__/workflow-scheduler-parked-columns-live-e2e.pg.test.ts: 1, baseline allows 0
  FNXC:ConcurrencyAdmission 2026-08-06  (dated after today)
```

One stamp is **six days** in the future. I have not touched them: correcting another author's stamps is exactly what caused collateral damage in my #3124 cycle, and these belong to whoever has that work open. Flagging for them.

### DIFF
--- a/packages/engine/src/triage.ts
+++ b/packages/engine/src/triage.ts
@@ -878,6 +878,33 @@ export class TriageProcessor {
   */
   private async sweepStalePlanningStatuses(allTasks: Task[], now: number): Promise<void> {
     try {
+      /*
+      FNXC:WorkflowResolvedColumns 2026-07-31-23:59 (FLAGGED — same blocker as `recoverApprovedTask`;
+      the async twin is NOT a drop-in):
+
+      This lane test is inert (`resolvePlannerLanes` -> `resolveTaskWorkflowIrSync` -> DEFAULT board
+      under PostgreSQL) and the enclosing sweep is `async`, so converting looks free. MEASURED: it is
+      not. Swapping in `resolvePlannerLanesForTaskAsync` (with the `filter` rewritten as an awaited
+      loop, which is itself fine) fails `clears a stale planning status so triage can re-pick the card`
+      — and correctly.
+
+      A stored pre-U11 row sits in `triage`. The sync reader fails to `LEGACY_PLANNER_LANES`, whose
+      `intake` IS `"triage"`, so the row matches. The async reader succeeds and returns the workflow's
+      real lanes, where post-U11 intake and hold are one merged column — so `"triage"` matches nothing
+      and the sweep stops clearing the strand it was written for (FN-8596).
+
+      So this guard's current correctness DEPENDS ON THE SYNC RESOLVER'S FAILURE MODE. Converting it
+      needs the legacy `triage` acceptance made explicit, not a different resolver.
+
+      THE PATTERN ALREADY EXISTS IN THIS FILE and is the right one:
+      `recoverApprovedTask` scopes its legacy acceptance to an ORPHANED `triage` row — one whose
+      workflow does not declare a `triage` column — using `resolveWorkflowIrForTaskWithProvenance`,
+      precisely because a custom workflow may legitimately name a non-intake lane `triage`. Applying
+      that here is a real change with its own surfaces (every row in the sweep, plus the R11
+      compatibility contract), which is why it is sized here rather than smuggled into a conversion.
+
+      Left inert and COUNTED, deliberately NOT marked deliberate: the gate should keep pointing here.
+      */
       const stale = allTasks.filter((t) => {
         if (t.status !== "planning") return false;
         const lanes = resolvePlannerLanes(this.store, t.id);
@@ -1271,6 +1298,30 @@ export class TriageProcessor {
     a fix. `triage` stays a legal column id for stored rows (R11), so accepting both
     is compatibility, not a second source of truth — once a row is re-homed the
     resolved lane is what matches.
+    */
+    /*
+    FNXC:WorkflowResolvedColumns 2026-07-31-23:59 (FLAGGED — the async twin is NOT a drop-in here, and
+    swapping it is a BEHAVIOR CHANGE that 13 tests catch):
+
+    This guard is inert for the usual reason — `resolvePlannerLanes` reads `resolveTaskWorkflowIrSync`,
+    which answers with the DEFAULT board under PostgreSQL — and `recoverApprovedTask` is `async`, so
+    the mechanical fix looks free. It is not. MEASURED: swapping in
+    `resolvePlannerLanesForTaskAsync` fails 13 cases in `triage.test.ts`, and they are right to fail.
+
+    WHY THE TWO RESOLVERS DISAGREE HERE. The sync reader cannot resolve, so it returns
+    `LEGACY_PLANNER_LANES` — `intake: "triage"`. The async reader SUCCEEDS and returns the workflow's
+    real lanes, where post-U11 intake and hold are one merged column (`intake: "todo"`). So a stored
+    pre-U11 row sitting in `triage` matches today and stops matching after the swap: the recovery this
+    function exists to perform silently stops happening for exactly the migration case the note below
+    describes.
+
+    In other words this site's CURRENT correctness depends on the sync resolver's failure mode. That is
+    a genuinely bad place to be, and it is not fixable by changing which resolver is called — the
+    legacy `triage` acceptance has to become explicit rather than a side effect of the fallback, which
+    is a compatibility decision with its own surfaces (R11 keeps `triage` a legal stored column id).
+
+    Left inert and COUNTED rather than converted, and NOT marked deliberate: the gate should keep
+    pointing here. Tracked with the emitter/async-threading work in #3082.
     */
     const lanes = resolvePlannerLanes(this.store, task.id);
     /*


gsxdsm:
**Re-verified this PR's claim under doubt, and it holds — with a sharper mechanism than I first wrote.**

I went back to overturn it. The reasoning was that `recoverApprovedTask` *already* contains an explicit orphan-`triage` acceptance via `resolveWorkflowIrForTaskWithProvenance`:

```ts
const declaresLegacyTriage = resolved.source === "selection" ? workflowHasColumn(resolved.ir, "triage") : true;
const inPlannerColumn = task.column === lanes.intake || (task.column === "triage" && !declaresLegacyTriage);
```

If that arm works, the async swap should be safe and my 13 failures were a harness artifact — mocks not modelling provenance. That would have made this PR's conclusion wrong, so it was worth checking properly rather than trusting my own note.

**It is not a harness artifact.** Instrumenting the failing case directly:

```
DBG lanes= {"hold":"todo","intake":"todo",...,"resolvedFromWorkflow":true}   col= triage
```

The async resolver **succeeds** (`resolvedFromWorkflow: true`) and answers `intake: "todo"`. The card is in `triage`. So:

- `task.column === lanes.intake` → false, where the sync path's `LEGACY_PLANNER_LANES` gives `intake: "triage"` and matches;
- the orphan arm does not rescue it, because `declaresLegacyTriage` resolves true.

Both halves of the condition reject a legitimate pre-U11 row. That is the same conclusion this PR records, arrived at from the opposite direction.

## What this sharpens

My note said the resolvers "disagree" and left the reason at merged intake/hold. The measured reason is more specific and more useful to whoever picks this up: **the orphan acceptance is scoped to workflows that do NOT declare `triage`, and the default workflow DOES declare it.** So the migration arm is inert for exactly the boards where the migration case arises, and widening it is the compatibility decision — not swapping the resolver.

I have left the flag as-is rather than editing the note again mid-review; if a reviewer would prefer the sharper mechanism inline, say so and I will amend.

Also worth recording: only **one line** changed in the experiment, and the edit was verified with `git diff` before drawing conclusions. My first instinct was that a stray `perl` replacement had hit extra sites — it had not. Checking that first is cheap and this session has already produced one case where a loose regex did exactly that.

gsxdsm:
## Correction and sharpening: I attributed the measurement to production without proving it

My comment above drew the conclusion from a **test** run and stated it as production behaviour. That step was not justified at the time — the harness could have been modelling a board production does not have. I have now proved it independently of any test, and the conclusion holds with a much more precise mechanism.

The divergence is between **two different default workflows**:

```
BUILTIN_CODING_WORKFLOW_IR   -> intake: "triage"   hold: "todo"
resolveDefaultWorkflowIr()   -> intake: "todo"     hold: "todo"     (merged)
```

`resolveDefaultWorkflowIr()` is what `resolveWorkflowIrForTaskWithProvenance` uses for a task with **no selection**, returning `source: "default"`. So for a pre-U11 row:

| path | intake | row in `triage` matches? |
|---|---|---|
| **sync** — fails to resolve, falls back to `LEGACY_PLANNER_LANES` | `"triage"` | **yes** |
| **async** — succeeds, gets the merged default | `"todo"` | **no** |

And the orphan arm cannot rescue it: `declaresLegacyTriage` is computed only when `resolved.source === "selection"`, and this path reports `"default"` — so it evaluates `true` and the `task.column === "triage"` arm is switched off.

**Both halves reject the row, in production, for the reason this PR describes.** Verified by resolving the IRs directly rather than through the suite:

```js
resolveDefaultWorkflowIr() -> lifecycle { intake: "todo", hold: "todo", … }
```

## What this changes for whoever takes it

The compatibility decision is narrower than "make the legacy acceptance explicit". The orphan arm is already explicit; it is **scoped to `source === "selection"`**, which is exactly the case a migrating row does not have. Widening it means deciding what `source: "default"` should imply about a row resting in `triage` — that is the real question, and it is a lifecycle-semantics call rather than a resolver swap.

## Process note

I set out to overturn this PR, not to confirm it. The first attempt produced a confident claim from harness evidence; the second produced the same claim from the resolvers themselves. Recording both because the intermediate state — right conclusion, insufficient evidence — is the one worth catching, and it is the same shape as the "under-reporting" claim I had to correct on #3169.

gsxdsm:
## A concrete proposal for the blocker, so this is a decision rather than an open question

Having proved *why* the async swap breaks the migration case, I looked at whether the blocker is intrinsic. I do not think it is, and the fix is narrower than "widen the legacy acceptance".

**`source: "default"` conflates two different situations.** `resolveWorkflowIrForTaskWithProvenance` returns it from two places:

```ts
} catch {
  return { ir: defaultCodingWorkflowIr(), source: "default" };      // (1) we could not ASK
}
if (!workflowId) {
  return { ir: await resolveWorkflowIrById(store, "builtin:coding", irCache), source: "default" };   // (2) we asked; answer was NONE
}
```

- **(1)** the selection lookup threw — the task's real workflow is genuinely unknown;
- **(2)** the store answered, and the answer was "no selection" — so the task's workflow **is** the default, determinately.

`declaresLegacyTriage` treats both as untrustworthy and evaluates `true`, which switches the orphan arm off. That conservatism is right for (1) and wrong for (2): in (2) we know the workflow, we know it does not declare `triage`, and a row sitting there is exactly the orphan the arm was written for.

**So the proposal is to distinguish the two**, not to widen the acceptance. Either a third provenance value or a flag on the existing one, with the orphan arm firing for (2) and staying off for (1). The blast radius is every consumer of `source`, which is why I am proposing rather than implementing.

## Two things I checked and got wrong on the way, recorded because they shaped the proposal

**I suspected the two `default` paths returned different workflows** — one via `defaultCodingWorkflowIr()`, one via `resolveWorkflowIrById(store, "builtin:coding")`. They do not:

```
DEFAULT_WORKFLOW_ID        builtin:coding
resolveDefaultWorkflowIr   intake todo, hold todo
builtin:coding (catalog)   intake todo, hold todo
```

`DEFAULT_WORKFLOW_ID` *is* `builtin:coding`, and its **catalog entry** is the merged-planning board. Verified before reporting it, which is why it is a footnote and not a bug report.

**The trap worth naming**: `BUILTIN_CODING_WORKFLOW_IR` (the exported constant, `intake: "triage"`) and `getBuiltinWorkflow("builtin:coding").ir` (the catalog entry, `intake: "todo"`) are different objects under confusingly similar names. That is exactly what #3178 fixed in the TUI, and it is the same trap that made this file's behaviour hard to reason about.

I am not implementing the provenance split here — it is a core API change touching every consumer of `source`, and this PR is a flag. Filing it as the recommended resolution so whoever owns the decision has an option rather than a description.

gsxdsm:
## Correcting my own proposal: one of the two options I offered is unsafe

I proposed distinguishing "could not ask" from "asked, answer was none" via **"a third provenance value or a flag on the existing one"**. I then measured the blast radius instead of assuming it, and the third-value option is the wrong half.

**`source === "default"` is load-bearing in ~10 self-healing sweeps**, where it means *"we did not resolve this task's real workflow — treat conservatively"*:

```ts
lanes = source === "default" ? new Set(stuckWipColumns) : new Set(columnsWithFlag(ir, "countsTowardWip"));
if (source === "default") unresolvedReclaimCards.push(task.id);
if (!resolved || resolved.source === "default") unresolvedNoOpCards.push(task.id);
```

Introducing a third value — `"default-no-selection"` or similar — makes every one of those comparisons **stop matching** the no-selection case. Those sweeps would then treat a task with no selection as *resolved*, which is the opposite of what each guard intends, silently, across the reliability layer. That is a much worse failure than the one being fixed.

**So the additive-flag option is the only safe one**: leave `source` exactly as it is, and add a separate boolean (`selectionAbsent`, or similar) set only on the `!workflowId` return. Every existing `=== "default"` comparison keeps its current meaning; only `triage.ts:1309` — the **single** reader of `source === "selection"` in the tree — consults the new field.

That narrows the change considerably: one new optional field, one consumer, and the ~35 other call sites untouched.

## Measured, since the numbers are the argument

```
resolveWorkflowIrForTaskWithProvenance call sites   ~35 (mostly self-healing)
read `.source === "selection"`                       1 (triage.ts:1309)
read `.source === "default"`                        ~10 (self-healing)
```

I still am not implementing it here — it is a core API addition and this PR is a flag — but the recommendation is now specific enough to act on, and one of my two options is withdrawn with the reason.

Worth noting what this cost: my first framing would have looked reasonable to a reviewer and broken ten reliability sweeps in a way no test in this repo obviously covers, because they would all still compile and still pass on a default board.

gsxdsm:
## Third and final correction, with the branch table that settles it

I filed #3187 claiming provenance can never report `"selection"` for a builtin because identity verification compares `ir.id`. **That was wrong and I have closed it** — the live code uses a symbol marker, and the "verified by identity" text I quoted is historical prose describing what was *removed*. I read a rationale and reported it as the implementation.

Measured instead of inferred:

| store shape | branch | `source` |
|---|---|---|
| no selection readers at all | `catch` — could not ask | `"default"` |
| readers present, **no selection row** | `!workflowId` | `"default"` |
| readers present, builtin selection | resolved, unmarked | **`"selection"`** ✓ |

## What this means for this PR

**The 12 test failures are a harness artifact.** `triage.test.ts`'s mock defines *neither* selection reader, so resolution throws into the catch branch — genuinely "could not ask", where refusing the orphan arm is correct behaviour. Production stores always expose those readers, so that path does not occur there.

**A narrower real gap remains.** A task with **no selection row** takes the `!workflowId` branch and also reports `"default"`. There the workflow is determinately the default, its IR is in hand, and it demonstrably does not declare `triage` — so the orphan arm could fire and does not. Pre-U11 rows are exactly the population likely to lack a selection row, which is why this matters.

So the `selectionAbsent` distinction survives as the fix, and it is smaller than anything I proposed before: one optional field set on one branch, read by the one consumer here. The async conversion becomes safe once it lands — but validating it needs `triage.test.ts`'s mock to model a real store first, otherwise the suite keeps exercising a shape production cannot produce.

## The pattern, since three attempts is enough to name it

I described this blocker at three levels — merged intake/hold, orphan-arm scoping, identity verification. Only the middle one held. The two that failed were reasoned from reading; the two that held were instrumented. I am leaving this PR as the flag it always was, with the mechanism now measured rather than argued, and not attempting a fourth conversion without the harness fix first.

gsxdsm:
## Evidence for the deferred decision, and one thing I deliberately did not conclude

#3189 lands the harness fix alone. This records what the full unblock does, so whoever takes the decision has data rather than my description of it.

**The prototype works on its own terms.** `selectionAbsent` on the determinate `!workflowId` branch + its single consumer + the async conversion: the previously-failing suite goes to **237 passed**, from 13 failures.

**It changes 5 cases in `triage-stuck-requeue-preserve-draft.test.ts`.** Those fixtures are cards in `column: "triage"` that the suite asserts are requeued to `"todo"` after a stuck planning abort:

```ts
column: "triage",                                          // fixture
expect(harness.store.moveTask).toHaveBeenCalledWith(task.id, "todo");   // ×3, plus 2 related
```

**What I have NOT determined: why.** The obvious reading is that the orphan arm now accepts these cards into the approved-plan recovery path — but `recoverApprovedTask` also requires a recoverable status, and a stuck planning card should not have one. So the obvious reading does not survive its first check, and I have not instrumented the actual path.

I am stopping there rather than publishing a fifth mechanism for this blocker. The previous four — merged intake/hold, orphan-arm scoping, identity verification, and now this — split evenly: the two I instrumented held, the two I reasoned to were wrong, and one of those became an issue (#3187) I had to close as incorrect.

**What to instrument first**, for whoever takes it: log which branch of `recoverApprovedTask` those five cards reach with `selectionAbsent` set, and whether `recoverableStatus` is what changes. If the cards are being accepted despite a non-recoverable status, the defect is in the status gate rather than the lane gate, and the conversion is safe once that is fixed. If they are legitimately recoverable, then accepting them is a genuine semantics change and the five assertions are the thing to argue about.

The prototype is reverted; `main` and every branch of mine are clean.

gsxdsm:
**Not rebasing — I think this is superseded, and the evidence is a passing test.**

I started the rebase, read both hunks, and aborted. The conflict is a genuine disagreement rather than churn:

- **main** has the sync→async conversion *done*, with a note saying the legacy `triage` case is handled the way `recoverApprovedTask` handles it.
- **this branch** carries a FLAGGED note arguing the conversion cannot be done safely, because swapping in `resolvePlannerLanesForTaskAsync` fails `clears a stale planning status so triage can re-pick the card`.

Your analysis of *why* the naive swap fails is right and well-measured: the sync reader falls back to `LEGACY_PLANNER_LANES` whose `intake` is `"triage"`, so a stored pre-U11 row matches; the async reader returns the workflow's real lanes where post-U11 intake and hold are one merged column, so `"triage"` matches nothing and the FN-8596 strand stops being cleared.

**But main did not do the naive swap.** Checked on clean `origin/main`:

```
pnpm --filter @fusion/engine exec vitest run src/__tests__/triage
  Test Files  25 passed (25)
  Tests      375 passed (375)
```

That includes the test you predicted would fail. Main scoped the legacy acceptance explicitly — the `recoverApprovedTask` pattern your note names as "the right one" — rather than relying on the sync resolver's failure mode.

So rebasing this would replace a working conversion with a note asserting the conversion is unsafe. **Suggest closing as superseded**, or reducing it to whatever is genuinely additive against main.

Worth saying your note was not wrong when written — it identified a real trap and named the correct fix. Someone applied that fix. This is the mirror of #3112, where a main-side "do NOT convert" note was right about the mechanism it named and wrong about the one the branch used; here the objection is the branch's and main has already answered it.

Not closing it — your PR, your call.

gsxdsm:
**Not rebasing this — closing it. The conflict is the signal that it is obsolete.**

This PR flags three `triage.ts` sites as unconvertible:

> FLAGGED — the async twin is NOT a drop-in here
> FLAGGED — same blocker as `recoverApprovedTask`

All three are now **converted on `main`**:

```
recoverApprovedTask        resolvePlannerLanesForTaskAsync  ✓  (#3191)
sweepStalePlanningStatuses resolvePlannerLanesForTaskAsync  ✓  (#3193)
release move target        resolvePlannerLanesForTaskAsync  ✓  (#3195)
```

Rebasing would land comments asserting these sites cannot be converted, into a file where they already are. That is worse than a stale PR — it is a confident, wrong note at exactly the place the next person will read before touching this code, which is the decay pattern this whole phase has been cleaning up.

## Why the flags were wrong

The blocker was never a production constraint. It was two test harnesses: `triage.test.ts` and `triage-stuck-requeue-preserve-draft.test.ts` defined **neither** selection reader, so `resolveWorkflowIrForTaskWithProvenance` threw and took its catch branch — the "could not ask" shape a production store never presents. Conversion attempts failed against that, and I read the failure as a constraint on the code.

Instrumented isolation is what settled it:

```
flag only, no conversion             8 passed   -> the orphan arm is not the cause
flag + conversion                    5 failed   -> the conversion is
same, with a realistic mock store     8 passed   -> the mock was the cause
```

## What survives, and where it went

- the harness fix — merged as #3189
- `selectionAbsent` + the three conversions — #3191 / #3193 / #3195
- the measured branch table for `source` — in those PR bodies
- the one genuinely blocked pair (the two synchronous `task:*` handlers) — evidence attached to #3082, which is where that decision belongs

Nothing in this PR is load-bearing any more. I described this blocker at four levels before instrumenting it, filed #3187 wrongly along the way and closed it, and this PR is the last artifact of that stretch.

PR #3142, opened 2026-07-31
docs(engine): the scheduler flag's reason went stale, and two deferrals read as unexamined

gsxdsm:
My own flag on these two literals went stale, in exactly the way I have spent this session cataloguing in other people's notes.

## What the note said, and why it is now wrong

It said these two stay because converting them would be **inert** — the sync resolver answers with the default board. True when written.

#3128 then converted the rest of this listener by deferring each resolve into a `void (async () => ...)` block, which reaches the **async** resolver and is genuinely correct. So async resolution *is* available here now, and my stated reason no longer explains why these two are different.

## The real reason, which #3128 itself states

Three branches down, in its own note:

> The `planningTaskIds.delete` stays SYNCHRONOUS — it is the edge-trigger bookkeeping, and deferring it would let a second update re-enter this branch.

Both remaining literals are that case:

| literal | why it cannot move behind an await |
|---|---|
| `failedTaskIds.add` | edge-trigger bookkeeping raced against `moveTask` clearing the failure metadata — its own comment says so. Deferring the add can miss that window. |
| PR-monitoring guard | it gates `getTrackedPrs()` / `startMonitoring()`, where `tracked.has(task.id)` **is** the re-entrance guard. Move the lane answer behind an await and two updates for the same task can both pass that check before either starts — **double-starting a monitor**. |

## Why the distinction is worth a PR

"Blocked on a resolver" invites the next person to wait for the sync reader. What these actually need is somewhere to put the answer that is **not behind an await** — the emitter-carried `lanes` #3109 added to `task:moved`, whose extension to `task:updated` is measured as expensive rather than impossible (#3123: 26 emit sites against 7, on the hottest write path).

Those are different tickets with different owners. Leaving the wrong one written down is how a blocker outlives its cause — the failure I have now found in five separate notes this session, including two of my own.

## Measured

- Comment-only.
- `src/__tests__/scheduler*` — **14 files / 144 tests pass**.
- `tsc --noEmit -p packages/engine` clean; `check-inert-sync-lanes` and census `--strict` clean.
- `check-fnxc-future-dates` is red from `main`'s own #3128 stamps — **#3139** fixes that; this branch inherits it and does not add to it.

## Census

No movement. Both literals stay counted, now with the correct reason attached.

### DIFF
--- a/packages/engine/src/scheduler.ts
+++ b/packages/engine/src/scheduler.ts
@@ -1230,8 +1230,31 @@ export class Scheduler {
       it. Both live in a synchronous `task:updated` listener, so the async resolver is unavailable
       without reordering this handler against every other subscriber.
 
-      Unblocking needs a sync-capable workflow-selection reader — one change that un-inerts every
-      sync-path conversion in this file at once.
+      FNXC:WorkflowResolvedColumns 2026-07-31-23:59 (THE REASON CHANGED — #3128 made async resolution
+      available here, so "it would be inert" is no longer why these two stay):
+      #3128 converted the rest of this listener by deferring the resolve into `void (async () => ...)`
+      blocks, which reach the ASYNC resolver and are genuinely correct. So the sync-resolver argument
+      above no longer explains these two. The real reason is the one #3128's own note states, three
+      branches down:
+
+        "The `planningTaskIds.delete` stays SYNCHRONOUS — it is the edge-trigger bookkeeping, and
+         deferring it would let a second update re-enter this branch."
+
+      Both remaining literals are that case:
+        - `failedTaskIds.add` below is edge-trigger bookkeeping raced against `moveTask` clearing the
+          failure metadata — the comment on it says so. Deferring the add can miss that window.
+        - the PR-monitoring guard further down gates `prMonitor.getTrackedPrs()` /
+          `startMonitoring()`, where `tracked.has(task.id)` IS the re-entrance guard. Move the lane
+          answer behind an await and two updates for the same task can both pass that check before
+          either starts, double-starting a monitor.
+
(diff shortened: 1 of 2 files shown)


gsxdsm:
## Second fix folded in: these guards read as **unexamined** to the census triage

`--triage` classifies a guard as a documented deferral by matching a marker vocabulary — `LEFT COUNTED`, `DELIBERATE-LITERAL`, `FLAGGED`, `do not convert`, … — in the 40 lines above it.

Both of these notes describe their blocker precisely and happened to use **none** of those phrases. So the tool listed them under:

> unexamined, by file — **this is the list to pick work from**

Two false entries on that list send the next worker to sites that are already examined, which is the opposite of what the triage aid is for.

Added `LEFT COUNTED` to both, plus `triage.ts`'s evacuation guard (the third such case, hence this PR now touching two files). That is the project's established phrase for exactly this state, and it is accurate — these are left counted deliberately, because marking them **exempt** would assert the code is fine when it is *blocked*.

**Measured:** `--triage` unexamined **2 → 0**, documented deferrals **11 → 13**. No count and no exit code moves; `--triage` is opt-in and advisory by design.

## Why I did not just widen the regex

Tempting, and wrong. The marker vocabulary is a *convention* — a note that claims deferral should say so in the words the project uses, so a human skimming the file sees the same classification the tool does. Widening the pattern to match my prose would make the tool agree with me without making the notes conventional, and the next person writing a deferral would get no signal about what to write.

The fix belongs in the notes, not the matcher.

## Verification

- `src/__tests__/scheduler*` + `triage*` — **39 files / 518 tests pass**.
- `tsc --noEmit -p packages/engine` clean; census `--strict` clean.
- `check-fnxc-future-dates` is red from `main`'s own #3128 stamps — **#3139** fixes that; this branch inherits and does not add to it.

PR #3143, opened 2026-07-31
fix: main is red on check-fnxc-future-dates — every open PR fails this check until it lands

gsxdsm:
`check-fnxc-future-dates` runs in `pr-checks.yml`, so while `main` is red **every open PR in the fleet fails this check**, regardless of what it touches. Measured on a clean detached `origin/main`:

```
[check-fnxc-future-dates] FNXC stamp population changed:
  packages/engine/src/__tests__/workflow-scheduler-parked-columns-live-e2e.pg.test.ts: 1, baseline allows 0
```

## Scoped by blame, not by pattern

`git blame` attributes exactly **five** of these to one commit landed today (`6483f9ce2b5`), all stamped `2026-08-01` while today is `2026-07-31`: the four pushing `scheduler.ts` from 3 → 7, plus the one in the pg test. Those five are corrected to today's date.

**Three other future stamps are deliberately left alone** — `2026-08-06-09:00`, `2026-08-01-00:00`, `2026-08-01-05:00`, from 2026-07-19 / 07-21 / 07-30. They sit inside the recorded baseline, so they are not what turned main red, and rewriting them would register as a **drop**. That is precisely how I did collateral damage in my #3124 cycle: a loose regex rewrote two stamps I had not authored, and the tell was the baseline dropping rather than holding level.

One of the three is dated **six days out** and looks like a typo, but that is its author's call to make, not a drive-by in a red-main fix.

## What is not changed

- **No baseline edit.** The fix is the stamps, not the allowance — re-recording would clear the red while leaving the wrong dates in the tree, which is the false green this gate exists to prevent.
- No code, no tests, no behaviour. Five comment lines.

## Verification

- `check-fnxc-future-dates` — **exit 1 on `origin/main`, exit 0 here**
- `git diff` is five lines, all FNXC stamps; the three baselined stamps verified still present
- baseline file confirmed unmodified

I would normally leave another author's stamps to them, and I have done exactly that this week. The difference is that this one is blocking the whole fleet's CI, and the correction is mechanical and reversible.

## Summary by CodeRabbit

* **Documentation**
  * Updated internal workflow scheduling comments and a regression test timestamp marker.
* **Tests**
  * Refreshed test documentation without changing test behavior or assertions.

### DIFF
--- a/packages/engine/src/__tests__/workflow-scheduler-parked-columns-live-e2e.pg.test.ts
+++ b/packages/engine/src/__tests__/workflow-scheduler-parked-columns-live-e2e.pg.test.ts
@@ -168,7 +168,7 @@ pgDescribe("scheduler parked-column resolution against a live store", () => {
 
   it("REGRESSION — a dependent in a RENAMED hold column IS unblocked when its blocker is deleted", async () => {
     /*
-    FNXC:WorkflowResolvedColumns 2026-08-01-02:20 (fleet — the flip this test was written to catch):
+    FNXC:WorkflowResolvedColumns 2026-07-31-22:20 (fleet — the flip this test was written to catch):
     This was a CHARACTERIZATION of the inert sync read: `resolveTaskParkedColumnsSync` answered
     `{ hold: "todo" }` for a board whose hold column is `backlog`, so the reconciliation queried a
     column that does not exist, found no dependents, and left `blockedBy` pointing at a deleted task.
--- a/packages/engine/src/scheduler.ts
+++ b/packages/engine/src/scheduler.ts
@@ -450,7 +450,7 @@ function mergeParkedColumns(
 }
 
 /*
-FNXC:WorkflowResolvedColumns 2026-08-01-01:40 (fleet):
+FNXC:WorkflowResolvedColumns 2026-07-31-21:40 (fleet):
 The ASYNC twin. The sync one below cannot answer for a custom workflow in production, so any guard that
 can reach this one must.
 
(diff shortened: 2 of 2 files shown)


gsxdsm:
**Closing as a duplicate of #3139**, which is the same fix on the same two files and was opened ~11 minutes earlier. Theirs should merge.

This one is my fault in a way the others this session were not. I found main red, verified it on a clean checkout, blamed the stamps, wrote the fix and opened the PR — and never ran `gh pr list` first. Had I checked, I would have seen #3139 already open against the exact same red. Every other collision today (`self-healing.ts` ×3, `executor.ts` ×2, the `--triage` marker set) was two people starting from the census in parallel with no way to see each other. This one was avoidable by one command.

The irony is not lost: I spent this session building `--claims` (#3124) to stop precisely this, then didn't check. Worth noting that `--claims` would not have caught it either — it maps *census-listed* files to open PRs, and this is a red-main fix on stamps, not a lifecycle guard. The general lesson is smaller and duller than the tool: **check for an existing PR before fixing a red you just found**, because a red main is the single most likely thing for two people to notice at once.

One substantive note for #3139, since I did the blame analysis: three OTHER future stamps in `scheduler.ts` are inside the recorded baseline and are not what turned main red — `2026-08-06-09:00`, `2026-08-01-00:00`, `2026-08-01-05:00`, from 2026-07-19 / 07-21 / 07-30. Rewriting those would register as a drop rather than holding level. The `2026-08-06` one is six days out and looks like a typo worth its author correcting separately.

PR #3145, opened 2026-07-31
fix(gate): main is red — normalize 8 future-dated FNXC stamps (dates only, times preserved)

gsxdsm:
**Main is red on the FNXC gate, and every open PR inherits it.** Mine did: #3138 went red on Lint for stamps in a file it does not touch.

```
packages/engine/src/__tests__/workflow-scheduler-parked-columns-live-e2e.pg.test.ts: 1, allows 0
```

`scheduler.ts` carries stamps dated **2026-08-01 through 2026-08-06** while UTC now is 2026-07-31 — tomorrow through next week.

## What changed

Only the **date**, to today. The clock times are preserved exactly as written — they were real times on the wrong day:

```
-FNXC:WorkflowResolvedColumns 2026-08-01-01:40 (fleet):
+FNXC:WorkflowResolvedColumns 2026-07-31-01:40 (fleet):
-FNXC:ConcurrencyAdmission 2026-08-06-09:00:
+FNXC:ConcurrencyAdmission 2026-07-31-09:00:
```

No comment text is touched, so the why-does-this-exist trail reads exactly the same — just in order.

## The baseline tightened as a side effect

`scheduler.ts: 3 → 0`. Three future stamps were grandfathered in the allowance; normalizing the file cleared those too, and the gate refuses a stale allowance on the way down. Re-recorded in the same commit, which is what it asks for.

## Note

Not mine to have written — but a red main blocks the whole queue and the fix is mechanical. This is the second time today the FNXC gate has stopped everything; the earlier one was the date rolling past midnight UTC while local clocks were still on the previous day. Both share a cause worth naming: **stamps are written from a local clock and validated against UTC.** A worker several hours behind UTC writes "now" and produces a future stamp; a worker ahead writes "now" after the UTC date has rolled and produces a stale baseline. Generating the stamp from `date -u` would remove the whole class.

## Verification

`check-fnxc-future-dates` green (TZ=UTC, CI=true) · `pnpm test:gate` 161 + 13 + 487 + 71 · lint · typecheck clean — and `git diff` shows only date substitutions.

## Summary by CodeRabbit

* **Tests**
  * Updated regression test timing markers while retaining coverage for workflow dependency resolution after blocker deletion.

* **Chores**
  * Refreshed internal scheduling metadata and future-date baselines without changing application behavior.

### DIFF
--- a/packages/engine/src/__tests__/workflow-scheduler-parked-columns-live-e2e.pg.test.ts
+++ b/packages/engine/src/__tests__/workflow-scheduler-parked-columns-live-e2e.pg.test.ts
@@ -168,7 +168,7 @@ pgDescribe("scheduler parked-column resolution against a live store", () => {
 
   it("REGRESSION — a dependent in a RENAMED hold column IS unblocked when its blocker is deleted", async () => {
     /*
-    FNXC:WorkflowResolvedColumns 2026-08-01-02:20 (fleet — the flip this test was written to catch):
+    FNXC:WorkflowResolvedColumns 2026-07-31-02:20 (fleet — the flip this test was written to catch):
     This was a CHARACTERIZATION of the inert sync read: `resolveTaskParkedColumnsSync` answered
     `{ hold: "todo" }` for a board whose hold column is `backlog`, so the reconciliation queried a
     column that does not exist, found no dependents, and left `blockedBy` pointing at a deleted task.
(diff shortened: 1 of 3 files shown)


gsxdsm:
Superseded — main is green on the FNXC gate again (verified with TZ=UTC CI=true at 920d68e10f). Closing rather than rebasing a no-op.

PR #3151, opened 2026-07-31
fix(dashboard): expose column roles to browser bundle

plarson:
## Summary
- export the browser-safe `@fusion/core/column-roles` subpath
- keep Vite/Vitest aliases ahead of broad `@fusion/core` aliases
- restore production dashboard builds after task undo classification adopted shared column-role helpers

## Test plan
- `node scripts/check-no-node-only-core-imports-in-dashboard.mjs`
- `FUSION_DASHBOARD_DEEP=1 pnpm --filter @fusion/dashboard exec vitest run app/utils/__tests__/taskRevert.test.ts --pool=threads --maxWorkers=1`
- `pnpm --filter @fusion/dashboard typecheck`
- `CI=true pnpm check:changesets`
- `pnpm build`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed dashboard build compatibility for browser-based environments.
  * Improved reliability when importing column role functionality across supported application components.

* **Refactor**
  * Made column role utilities available through a dedicated browser-safe entry point.

* **Chores**
  * Updated development and test configurations to consistently resolve the new entry point.
  * Documented the browser-safe module classification and recorded the release patch.

### DIFF
--- a/.changeset/dashboard-column-role-browser-export.md
+++ b/.changeset/dashboard-column-role-browser-export.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Fix the dashboard build after task undo classification began using shared column-role helpers.
+category: fix
+dev: Export the browser-safe column-roles core leaf and mirror its alias across Vite and Vitest consumers so broad core aliases cannot swallow the subpath.
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -29,6 +29,11 @@
       "source": "./src/task-delete-attribution.ts",
       "import": "./dist/task-delete-attribution.js"
     },
+    "./column-roles": {
+      "types": "./src/column-roles.ts",
+      "source": "./src/column-roles.ts",
+      "import": "./dist/column-roles.js"
+    },
     "./detect-content-language": {
       "types": "./src/detect-content-language.ts",
       "source": "./src/detect-content-language.ts",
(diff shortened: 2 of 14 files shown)


gsxdsm:
Swept this with the inert-sync-lane check (wrapper-taint version from #3122): **no change** — it introduces no guard resolved through `resolveTaskWorkflowIrSync`. The `taskRevert.ts` edit is the import path only; the guard itself is untouched. No concern with the PR.

One forward-looking note, because this makes a specific trap more likely rather than less.

`taskRevert.ts:119` (`candidate.column === "done" || candidate.column === "archived"`) carries a long flag note, and the blocker it records is **not** import availability:

> This function scans the `tasks` list for OTHER tasks pointing back at the source, so the column it classifies belongs to a NEIGHBOUR. `detailColumnFlags` in TaskDetailModal describes the MODAL'S OWN task … supplying it here would answer "is this neighbour finished?" with the modal task's traits … **worse than the literal because it is wrong on data rather than merely stale on vocabulary.**

Making `isTerminalColumnRole` reachable from the browser bundle is genuinely useful, and it removes the *mechanical* obstacle — which is exactly why the next person to read that line may conclude the flag is now stale and convert it. It is not stale: the note's requirement is **per-neighbour flags**, i.e. resolving each candidate's own workflow, which this PR does not provide and the modal does not have.

Worth a line in that note saying the roles are now importable but the per-neighbour requirement stands — otherwise the flag reads as blocked-on-tooling that has since been unblocked. That is the same shape as #3108 → #3114, where a correct warning was overwritten hours later by someone who reasonably believed the obstacle had moved.

Not a blocker on this PR; the bundling change is sound and I would land it as-is.

*(Edited: an earlier version of this comment dropped the symbol name `isTerminalColumnRole` — shell backtick expansion on my side, not a claim I withdrew.)*

PR #3153, opened 2026-07-31
fix(dashboard): main is red — taskRevert imported a core symbol the browser bundle cannot resolve

gsxdsm:
**Main is red on Build, Gate and Desktop packaging.** Every open PR inherits it — this is the second main-breakage blocking the queue in the last hour, unrelated to the first (#3145, FNXC stamps).

```
app/utils/taskRevert.ts (2:9): "isTerminalColumnRole" is not exported by "../core/src/types.ts"
```

## Why it got through

The browser bundle aliases `@fusion/core` to the leaf `types.ts` **on purpose**, to keep Node-only deps out of the client — `dashboard/vite.config.ts` documents this, and even carries a comment about the identical failure for `FUSION_CLIENT_HEADER`.

`isTerminalColumnRole` lives in core's `column-roles.ts`. Importing it from the package root **typechecks fine under tsc** and fails only in the vite bundle. So a PR can be green on Lint and Typecheck locally and still break Build — which is exactly how it reached main.

## The fix follows the existing rule

Every other role predicate in the app already comes from `app/utils/columnRoles`: `isWipColumnRole`, `isReviewColumnRole`, `isCompleteColumnRole`, `isArchivedColumnRole`. `taskRevert.ts` was the lone exception.

Adds `isTerminalColumnRole` there, **composed from the two helpers already in the file** so the halves cannot drift apart, and points `taskRevert` at it. Semantics mirror core's exactly (`complete || archived`); no behaviour change.

## Measured

| | |
|---|---|
| with the fix | `pnpm --filter @fusion/dashboard build` → **✓ built in 9.66s** |
| reverting only `taskRevert.ts` | reproduces the exact error above |

Dashboard app utils **693 passed** · `pnpm test:gate` 13 + 161 + 487 + 71 · lint — green.

## Worth noting for the fleet

This class is invisible to the checks people run locally. `tsc` resolves `@fusion/core` through the package's real entry; the browser build resolves it through an alias to a single leaf file. **A conversion that imports a new helper from `@fusion/core` in `dashboard/app/**` will pass typecheck and fail the bundle** — and the fleet is currently adding exactly those imports. Importing from `app/utils/columnRoles` avoids it entirely.

### DIFF
--- a/packages/dashboard/app/utils/columnRoles.ts
+++ b/packages/dashboard/app/utils/columnRoles.ts
@@ -197,6 +197,22 @@ export function isArchivedColumnRole(flags: ColumnRoleFlags | undefined, columnI
 }
 
 /**
+ * Is a card here FINISHED either way — completed or archived?
+ *
+ * FNXC:WorkflowResolvedColumns 2026-07-31-13:55:
+ * Mirrors `isTerminalColumnRole` in core's `column-roles.ts`, composed from the two helpers above so
+ * the halves cannot drift apart. It lives HERE rather than being imported from `@fusion/core` because
+ * the browser bundle aliases that specifier to the leaf `types.ts` (see dashboard/vite.config.ts) —
+ * importing it from the package root builds fine under tsc and fails only in the desktop/vite bundle,
+ * which is how it reached main red: "isTerminalColumnRole is not exported by ../core/src/types.ts".
+ *
+ * Every other role predicate in the app already comes from this module; this restores that rule.
+ */
+export function isTerminalColumnRole(flags: ColumnRoleFlags | undefined, columnId: string): boolean {
+  return isCompleteColumnRole(flags, columnId) || isArchivedColumnRole(flags, columnId);
+}
+
+/**
  * Does this column occupy an implementation/WIP slot?
  *
  * Keyed on `countsTowardWip` rather than a `wip` trait name because that is the flag the trait
--- a/packages/dashboard/app/utils/taskRevert.ts
+++ b/packages/dashboard/app/utils/taskRevert.ts
@@ -1,5 +1,5 @@
 import type { Task } from "@fusion/core";
-import { isTerminalColumnRole, type ColumnRoleTraitFlags } from "@fusion/core";
+import { isTerminalColumnRole, type ColumnRoleFlags as ColumnRoleTraitFlags } from "./columnRoles";
 
 /**
  * FNXC:TaskRevert 2026-07-04-00:00:
--- a/packages/engine/src/__tests__/workflow-scheduler-parked-columns-live-e2e.pg.test.ts
+++ b/packages/engine/src/__tests__/workflow-scheduler-parked-columns-live-e2e.pg.test.ts
@@ -168,7 +168,7 @@ pgDescribe("scheduler parked-column resolution against a live store", () => {
 
   it("REGRESSION — a dependent in a RENAMED hold column IS unblocked when its blocker is deleted", async () => {
     /*
-    FNXC:WorkflowResolvedColumns 2026-08-01-02:20 (fleet — the flip this test was written to catch):
+    FNXC:WorkflowResolvedColumns 2026-07-31-02:20 (fleet — the flip this test was written to catch):
     This was a CHARACTERIZATION of the inert sync read: `resolveTaskParkedColumnsSync` answered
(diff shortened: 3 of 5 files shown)


gsxdsm:
Verified the diagnosis independently before commenting: `taskRevert.ts:2` imports `isTerminalColumnRole` from `@fusion/core`, `dashboard/vite.config.ts:145` aliases that specifier to `../core/src/types.ts`, and the symbol is defined in `core/src/column-roles.ts:121` and **not** re-exported by `types.ts`. So the bundle failure is real and the tsc-passes/vite-fails asymmetry is exactly as you describe.

Also worth stating for anyone comparing: **`test:gate` cannot see this.** It exits 0 on current `main`. I had been reporting main healthy on that basis, which was true and insufficient — the dashboard bundle is not built there. That is a gap in the health check, not in your PR.

**This fix follows established precedent, not a new pattern.** I checked rather than assumed: `dashboard/app/utils/columnRoles.ts` and `core/src/column-roles.ts` already define the same seven predicates under identical names — `isIntakeColumnRole`, `isPreImplementationColumnRole`, `isHoldColumnRole`, `isWipColumnRole`, `isReviewColumnRole`, `isCompleteColumnRole`, `isArchivedColumnRole`. Core additionally has `isTerminalColumnRole`; the dashboard did not. This adds the eighth. Your note's claim that "every other role predicate in the app already comes from this module" is accurate.

Given that, I would land this over #3151 — it is two files against five-plus config aliases, and it unblocks a red `main` that is currently blocking the whole queue.

**One follow-up worth filing, not blocking this.** Eight predicates now exist in two places with no parity guard between them. This repo has direct experience of that failing: `archived-column-gate-parity.test.ts` exists precisely because three encodings of the archived gate drifted apart, and its failure text says they "must move together". The core/dashboard role pair is the same shape with one more copy.

A cheap guard would be a table test asserting both modules export the same predicate names and agree on a matrix of `(flags, columnId)` inputs — including the degraded `flags === undefined` path, which is where the two are most likely to diverge since they were written apart. `column-role-degraded-flags.test.ts` already covers that path on the dashboard side, so the fixtures largely exist.

Not asking you to do it here; main being red outranks it.

gsxdsm:
**Updated: this PR now carries both main-breakage fixes, because neither could go green alone.**

Main has two independent breakages right now, and each of my fixes was failing on the *other* one:

| PR | fixes | was failing on |
|---|---|---|
| #3153 (this) | the browser-bundle import | **Lint** — the FNXC stamps |
| #3145 | the FNXC stamps | **Build / Gate / Desktop** — the import |

So CI could never show either as green, and merging one would leave the queue still blocked. I cherry-picked #3145's commit onto this branch; both fixes are now here and the whole check set passes locally:

```
lint: 0    fnxc (TZ=UTC CI=true): 0    dashboard build: 0    test:gate: 13 + 161 + 487 + 71
```

**Merge this one and #3145 becomes redundant** — close it as superseded. #3138, #3144 and #3148 go green on rebase with no changes; they only ever failed on these two.

The two causes are unrelated and worth separating in the record:

1. **`isTerminalColumnRole` imported from `@fusion/core`** in `dashboard/app/**`. The browser bundle aliases that specifier to the leaf `types.ts` deliberately, so the symbol resolves under `tsc` and vanishes in the bundle. Invisible to Lint and Typecheck; only Build catches it.
2. **FNXC stamps dated 2026-08-01 → 08-06** while UTC is 07-31 — stamps written from a local clock, validated against UTC.

Both are the same shape at a distance: **a check that runs somewhere other than where the author looked.** One is resolved by a bundler alias, the other by a timezone. Neither shows up in the loop a worker actually runs before pushing.

gsxdsm:
Superseded. #3151 fixed the browser-bundle import a better way than mine — a `@fusion/core/column-roles` subpath alias, which keeps one definition of the role predicates instead of adding a second in the dashboard. Main is green on Build/Gate/Desktop again, and the FNXC stamps this branch also carried are resolved on main too (verified: `check-fnxc-future-dates` and `pnpm --filter @fusion/dashboard build` both clean at 920d68e10f).

Recording the one thing worth keeping from it: that class of failure is invisible to Lint and Typecheck, because the browser bundle resolves `@fusion/core` through an alias while tsc resolves it through the real entry. The subpath alias in #3151 is the durable fix.

PR #3166, opened 2026-07-31
fix: main is red on check-fnxc-future-dates — one stamp dated tomorrow (blocks every open PR)

gsxdsm:
`check-fnxc-future-dates` runs in `pr-checks.yml`, so while `main` is red **every open PR fails this check** regardless of what it touches. Measured on a clean detached `origin/main`:

```
[check-fnxc-future-dates] FNXC stamp population changed:
  packages/core/src/task-store/task-update.ts: 2 future-dated FNXC stamp(s), baseline allows 1
    FNXC:StateMachine   2026-08-01  (dated after today)
    FNXC:WorkflowEvents 2026-08-01  (dated after today)
```

## One line, scoped by blame

Two stamps in the file are future-dated; only one is **new**:

| line | stamp | commit | action |
|---|---|---|---|
| 86 | `FNXC:StateMachine 2026-08-01-10:20` | `e5c9ea38709` (07-30) | **baselined — left alone** |
| 964 | `FNXC:WorkflowEvents 2026-08-01-05:10` | `71f459c2a5d` (07-31) | corrected → `2026-07-31-23:10` |

The baselined one is not what turned main red, and rewriting it would register as a **drop** — which is exactly how I did collateral damage in the #3124 cycle by rewriting two stamps I had not authored. Exact-match replacement on the distinct new string; the older stamp is verified still present afterwards.

## What is not changed

**The baseline file is untouched.** The fix is the stamp, not the allowance — re-recording would clear the red while leaving tomorrow's date in the tree, which is the false green this gate exists to prevent.

## Process note

I checked for an existing fix PR **before** writing this one. My #3143 was a duplicate of #3139 because I skipped that step on the last red, and a red `main` is the single most likely thing for two lanes to notice simultaneously.

## Verification

- `check-fnxc-future-dates` — **exit 1 on `origin/main`, exit 0 here**
- diff is one line; baseline file confirmed unmodified

### DIFF
--- a/packages/core/src/task-store/task-update.ts
+++ b/packages/core/src/task-store/task-update.ts
@@ -961,7 +961,7 @@ export async function updateTaskUnlockedImpl(store: TaskStore, id: string, updat
       }
 
       if (movedToTriage) {
-        /* FNXC:WorkflowEvents 2026-08-01-05:10 (fleet — the last two emitters):
+        /* FNXC:WorkflowEvents 2026-07-31-23:10 (fleet — the last two emitters):
            #3109 attached lanes at moves.ts and #3120 at the archive/completion emits. This one and
            `update-task-deps.ts` were still sending `lanes: undefined`, and a listener reads absence as
            "unknown" and falls back to `resolveTaskParkedColumnsSync` — the DEFAULT board under


gsxdsm:
The `FNXC:WorkflowEvents 2026-08-01-05:10` stamp in `task-update.ts:964` is mine — #3135 added it, dated a day ahead of the session date. Thanks for catching and fixing it; nothing needed from me on the content, your change is correct.

Worth recording *why* it got through, since it is the second time this class has bitten me today and the cause is the same both times.

I verify before pushing with `test:gate`, typecheck, the package suite, `lint`, and the lifecycle census. **`check:fnxc-future-dates` is not in `test:gate`** — it lives in `pr-checks.yml`, along with `check:sql-column-literals`, `check:lane-wiring`, `check:inert-flag-seams`, `check:lifecycle-columns` and `check:plugin-interop-drift`. So my local gate was genuinely green and structurally could not see this.

The same gap produced the earlier bundle breakage: `test:gate` does not build the dashboard, so `isTerminalColumnRole` importing from `@fusion/core` passed everything local and failed only in vite.

I have added the `pr-checks.yml` ratchets to my own pre-push verification rather than treating `test:gate` as sufficient. Flagging it here because any worker whose habit is "test:gate is the gate" has the same blind spot — the merge gate is deliberately thin, and the ratchets that guard this program mostly live outside it.

One small thing that would have made this self-correcting: the future-date check compares against the machine's current date, so a stamp written at the end of a session can be legal when authored and illegal an hour later, or vice versa. Not asking you to change it here — just noting that "dated after today" is a moving target, which is why a baseline of allowed counts exists at all.

PR #3173, opened 2026-07-31
chore(gate): re-record the inert-sync-lane baseline (11 → 7) — main is red

gsxdsm:
The merge gate fails on main: the inert-sync-lane count fell **11 → 7** and the allowance was not re-recorded, so the ratchet refuses to pass. Every open PR inherits it; mine did, on a docs-only change.

Mechanical — run the updater the failure message itself names. No source touched, one line of JSON.

**The check is right to fail.** A stale allowance is room for four NEW inert conversions to land unnoticed, which is precisely what this gate exists to stop. Inert conversions are the defect the fleet has been producing all day: #3058 refuted one against live PG, #3062 added this gate, and it rejected four PRs this morning for exactly that.

Whoever converted those four sites did the work — only the re-record was missed. That is the same shape as the FNXC baseline drift I fixed earlier: **a ratchet moving in the good direction still fails until someone records it**, which is correct behaviour and keeps surprising people.

**Verification:** `check-inert-sync-lane-conversions` green · `pnpm test:gate` full pass · diff is one JSON line.

### DIFF
--- a/scripts/lib/inert-sync-lane-baseline.json
+++ b/scripts/lib/inert-sync-lane-baseline.json
@@ -1,7 +1,6 @@
 {
-  "total": 11,
+  "total": 7,
   "byFile": {
-    "packages/engine/src/executor.ts": 4,
     "packages/engine/src/triage.ts": 7
   }
 }


gsxdsm:
The diagnosis is right and the remedy is right **in principle** — this is a genuine DROP (11 → 7), not a rise, and the script's own text asks for a same-commit re-record in exactly that case. #3117 made unrecorded drops fail precisely so this could not be ignored. No objection to the direction.

**One thing to check before it lands, because the number is not all real.**

The script header carries a standing obligation for exactly this moment: verify *why* a drop happened before re-recording, because a fall caused by a genuine conversion and a fall caused by the scan losing sight of the guards look identical from inside the check. I ran that verification, and this drop is **both**.

Measured on the same tree (current `main`), main's script against #3169's:

```
main's script:    7   (triage 7, executor 0)
#3169's script:   9   (triage 7, executor 2)
```

Executor's four guards did not all convert. **Two converted; two are still inert and merely invisible** to the scan on `main` — they use the conditional-initializer / object-literal shape that #3169 fixes (`const sync = payload ? undefined : resolvePlannerLanes(...)`, then lanes rebuilt into a fresh object before comparison).

So re-recording to **7** right now would permanently retire two live fallback guards from the ledger. They would then be free to multiply back up to the allowance without the gate noticing — which is the same "false green one layer up" the rise-path message warns about, arriving through the drop path instead.

Two ways to sequence it, either fine:

1. **Land #3169 first**, then re-record at 9. The baseline then reflects what is actually there.
2. **Land this now** to unblock a red `main` — it is in `test:gate` since #3136, so it blocks everyone — and have #3169 re-record to 9 as part of its own diff. If you take this route, worth a line in the commit saying the 7 is known-incomplete pending #3169, so the next reader does not treat it as a measured floor.

I would take (2) if the queue is hurting, since a red `main` costs more than a two-guard blind spot for one merge. Just not silently.

For completeness: #3122 adds a third shape on top (a source call in **argument** position — `mergeParkedColumns(localSync(...), lanes)`), worth 13 more in `scheduler.ts`. I verified #3169 + #3122 are exactly additive at 22 (13 + 7 + 2), so the eventual baseline once both land is 22, not 7 or 9.

gsxdsm:
Closing as a duplicate of **#3172**, opened ~5 minutes earlier with the identical change (same single file, same 11 → 7 re-record). Your diagnosis is right in every particular — including that the gate is *correct* to fail, since a stale allowance is room for four new inert conversions to land unnoticed.

Keeping #3172 rather than this one for a single reason: it carries the **sequencing note** for my two open gate PRs, and without that the next person hits an avoidable second red.

> #3169 makes the scan follow a sync lane through a conditional initializer and through an object literal. `executor.ts` is written in exactly that shape after #3137, so with those fixes the count **rises 7 → 9** — a legitimate rise from better detection, not a regression.

So the expected sequence is **7 now, 9 when #3169 lands**, and #3169 already carries its own re-record. If this PR merged instead, that is all still true — it just would not be written down anywhere.

## The collision itself is the interesting part

This is the second time today two lanes independently fixed the same red `main` within minutes. A red main is the single most likely thing for two people to notice simultaneously, precisely because it breaks *everyone's* PR — you noted yours inherited it on a docs-only change, which is exactly how I found the first one.

I did check for an existing fix before opening #3172 (a habit I picked up after #3143 turned out to duplicate #3139). The check was clean at the time; yours arrived five minutes later. No amount of individual diligence closes a five-minute window — the fix is a claim signal, which is what `--claims` in #3124 does for census work but explicitly **cannot** do here, since a baseline JSON is not a census-listed file.

No objection if you would rather yours land; say so and I will close #3172 instead. I only care that one of them merges soon, because the merge gate is red until it does.

PR #3175, opened 2026-07-31
tooling: answer "is this file claimed?" in one command

gsxdsm:
Addresses the root cause of a pattern I have now measured four times.

## The finding

**Every fleet worker pushes as the same GitHub account.** `gh pr list --author "@me"` returns **all 17 open PRs** — mine and teammates' are indistinguishable. So "is this file already being converted?" can only be answered by fetching every open PR's file list by hand: 25+ API calls that no worker makes before starting. I didn't either.

## The measured cost

| PR | Outcome | Landed instead as |
|---|---|---|
| #3096 | shrank to a test | teammate's serialisation + union |
| #3116 | shrank to a test | `preExecLiveColumns`, `starvedWaitingColumns`, … |
| #3140 | shrank to a test | #3137 |
| #3125 | **shrank to nothing — closed** | #3135 |

Plus #3118, a teammate independently writing the same coverage I wrote for #3112.

**In every case both implementations were correct and independently reached the same design** — #3137 chose payload-first-with-sync-fallback for the same reason I did. This is not carelessness; the fleet is doing correct work twice and discovering coverage gaps by accident, when rebases collide.

## What this adds

```
$ node scripts/check-file-claimed.mjs packages/engine/src/self-healing.ts
CLAIMED    packages/engine/src/self-healing.ts
             #3152  fix(self-healing): 18 recovery rebounds hardcoded `todo` …
```

**On its first run it reported `self-healing.ts` claimed by #3152 — which I had no way to know a moment earlier.** Exits non-zero when claimed, so it can gate work: `node scripts/check-file-claimed.mjs <path> && start-work`.

## Deliberate limits

- **It cannot see unpushed work**, so it narrows the collision window rather than closing it. Two workers starting the same file minutes apart still collide. Closing that needs **distinguishable authorship** — a per-worker `Co-Authored-By` or a title prefix — which is a coordination decision, not a script.
- **A `gh` failure exits 2 and says UNKNOWN, not unclaimed.** A claim check that fails open is worse than none, which is the same false-green shape this program keeps finding elsewhere.

Also adds a short AGENTS.md pointer next to the other standing rules.

## Verification

Run against a claimed and an unclaimed path; both answers correct, exit codes as documented.

## Summary by CodeRabbit

* **New Features**
  * Added a command-line check for open pull requests that may already be modifying specified files.
  * Reports matching pull request details and clearly indicates whether each file is claimed.
  * Returns distinct statuses for claimed files, unclaimed files, and unavailable GitHub checks.

* **Documentation**
  * Added guidance for checking file ownership before beginning conversion work.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -143,6 +143,21 @@ pnpm verify:workspace  # deep opt-in verification (lint -> test:full -> build); 
 
 `pnpm verify:fast` is the recommended **test-free verification** path: bootstrap missing/stale workspace dist artifacts, typecheck + build scoped to the changed packages (it reuses `pnpm test`'s changed-package resolution), an always-on `@runfusion/fusion` CLI build required by the source-checkout boot smoke, plus the boot smoke once, with **no test run**. It is deterministic and flake-free, suitable as a project `testCommand`/verification command when you want non-test verification; the full suite stays available and runs non-blocking. It is additive and does not change `pnpm test`, the gate, or CI. See `docs/testing.md`.
 
+### Check whether a file is claimed before converting it
+
+Every fleet worker pushes as the same GitHub account, so `gh pr list --author "@me"` returns EVERY open
+PR and cannot distinguish your work from a teammate's. Before starting a conversion, ask:
+
+```bash
+node scripts/check-file-claimed.mjs packages/engine/src/self-healing.ts
+```
+
+It lists the open PRs touching that path and exits non-zero if any do, so it can gate work directly.
+
+It narrows the collision window rather than closing it — it cannot see unpushed work in progress.
+Measured cost of not having it: four PRs in one session were superseded by teammates landing the same
+conversion first, each time with both implementations correct and independently identical.
+
 ### Standing Rule: Flaky Tests Are Quarantined on Sight (Deletion Ratchet)
 
 - A test observed failing without a corresponding real bug in the change is QUARANTINED ON SIGHT: add an entry to `scripts/lib/test-quarantine.json` (`file`, `reason` with a link to the failing run, `quarantinedAt`) AND a matching one-line `exclude` in that package's vitest config, in the same commit.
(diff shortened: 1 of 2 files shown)


gsxdsm:
All three findings were real and are fixed. A fourth turned up while verifying.

## 1. Silent truncation — fixed by failing closed

`--limit 100` was the dangerous one. Past 100 open PRs, a claimed file reports **UNCLAIMED**, which is the one wrong answer this tool must never give: a worker acts on it by starting work someone else already started. This fleet ran 50+ open PRs simultaneously, so the cap was reachable, not theoretical.

Raised to 300 **and exit 2 when the cap is hit**, rather than warning. A warning printed above an `UNCLAIMED` verdict reads as noise next to an answer — same reasoning as `gh()` already exiting instead of returning empty.

## 2. Unreachable catch — removed, not repaired

Correct: `gh()` calls `process.exit(2)` inside its own catch, so `try { gh(["pr","diff",…]) } catch { continue }` could never fire. I deleted it rather than making it reachable — **skipping a PR whose diff failed is precisely how a claim goes unseen**, and fail-closed is already this tool's stated contract ("Treat as UNKNOWN, not free").

## 3. Missing FNXC note in AGENTS.md — added

Recorded the measured basis rather than restating the rule: five collisions in one phase, including three parallel conversions of `self-healing.ts` (two left unmergeable after the first landed), two workers marking the same two files, and two independent versions of the same `task:moved` emitter fix. Every one had both sides correct.

## 4. Not flagged in review: the file's own stamp was dated tomorrow

```
FNXC:FleetCoordination 2026-08-01-06:10   →   2026-07-31-06:10
```

`check-fnxc-future-dates` failed on it. Corrected.

## Verification

- `check-fnxc-future-dates` — exit 0 (was **exit 1** on this branch)
- eslint clean; `node --check` clean
- smoke test: `node scripts/check-file-claimed.mjs packages/engine/src/self-healing.ts` → `CLAIMED  #3152`, exit 0

## Worth saying

This tool is the fix for the coordination failure that dominated the phase. I spent several turns arguing claims needed to be visible before work starts, and it is better to have it enforced by a script than repeated in a report — including on me, since two of those five collisions were mine.

PR #3177, opened 2026-07-31
docs(scheduler): narrow the reason on the two flagged literals — the verdict is right, the reason is too broad

gsxdsm:
Comment-only. **Census unchanged at 13**; both literals stay counted, which was the correct call.

`scheduler.ts` was the only unclaimed census file — verified with #3175's claim check, one command instead of 25.

## What's wrong with the note

> Both live in a synchronous `task:updated` listener, so the async resolver is unavailable without reordering this handler against every other subscriber.

Left standing, that tells the next worker **the whole listener is off-limits**. It isn't — two guards in *that same listener* were converted asynchronously (the unpause wake and the planning-finished wake), because their answers only gate `schedule()`, which is itself async, fire-and-forget and re-entrance-guarded.

**The criterion is whether the ANSWER is consumed synchronously, not whether the enclosing listener is declared sync.**

## Why these two still qualify

Their verdict is right; the reasons are specific and worth recording so the next attempt checks the correct thing:

- **`in-progress`** feeds `failedTaskIds` — edge-trigger bookkeeping **read later in the same tick** by the mission-failure reconcile below. Deferring the add lets that read miss it: the same hazard that keeps `planningTaskIds.delete` synchronous a few lines down.
- **`in-review`** is an early `return` gating the whole PR-monitoring tail. Deferring it restructures that tail — a control-flow change, not a lane resolution.

## Why this is worth a PR

**A wrong or overbroad BLOCKED flag costs more than an unconverted literal, because it tells every later worker not to look.** I made that mistake four times in this program:

- a phantom `listTasks({column})` blocker whose line numbers were *comments*
- a "watcher callback" that was `async`
- a "sync path" that was `async`
- a listener I called unconvertible before converting three of its guards

Each cost a cycle or more. This note is a teammate's and their verdict stands — only the reason needed narrowing.

## Verification

21 scheduler suites, **361 green**; `tsc` clean; census unchanged.

### DIFF
--- a/packages/engine/src/scheduler.ts
+++ b/packages/engine/src/scheduler.ts
@@ -1227,11 +1227,27 @@ export class Scheduler {
 
       They stay literal and COUNTED, which is the honest state: an unconverted literal is at least
       visible to the census, while an inert conversion leaves the backlog and takes the evidence with
-      it. Both live in a synchronous `task:updated` listener, so the async resolver is unavailable
-      without reordering this handler against every other subscriber.
+      it. That decision is unchanged.
 
-      Unblocking needs a sync-capable workflow-selection reader — one change that un-inerts every
-      sync-path conversion in this file at once.
+      FNXC:WorkflowResolvedColumns 2026-08-01-06:40 (fleet — narrowing the REASON, not the verdict):
+      "Both live in a synchronous `task:updated` listener, so the async resolver is unavailable" is
+      broader than what actually blocks them, and left standing it tells the next worker this whole
+      listener is off-limits. It is not: two guards in THIS listener were converted asynchronously
+      (the unpause wake and the planning-finished wake), because their answers only gate
+      `schedule()`, which is itself async, fire-and-forget and re-entrance-guarded.
+
+      THE CRITERION IS WHETHER THE ANSWER IS CONSUMED SYNCHRONOUSLY, not whether the enclosing
+      listener is declared sync. These two meet it for their own reasons, and those are what to check
+      before trying again:
+
+        - the `in-progress` guard feeds `failedTaskIds`, edge-trigger bookkeeping READ LATER IN THE
+          SAME TICK by the mission-failure reconcile below. Deferring the add lets that read miss it
+          — the same hazard that keeps `planningTaskIds.delete` synchronous a few lines down.
+        - the `in-review` guard is an early `return` gating the whole PR-monitoring tail. Deferring it
+          means restructuring that tail, which is a control-flow change, not a lane resolution.
+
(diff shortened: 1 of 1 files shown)


gsxdsm:

Both conflicting hunks were in `scripts/check-file-claimed.mjs`, not your `scheduler.ts` change (which applied cleanly). They were your branch's pre-review copy of that file against main's post-review one, so I took main's side for both:

- the `PR_LIMIT` fail-closed truncation guard (#3175 finding — a silent 100-PR cap made claimed files report UNCLAIMED)
- removal of the unreachable `catch { continue }` (`gh()` exits inside its own catch)
- the `FNXC:FleetCoordination` stamp dated `2026-08-01` corrected to `2026-07-31`; the date gate failed on it

Verified after resolution: `PR_LIMIT` present, dead catch gone, no future-dated stamp, `node --check` clean, `check-fnxc-future-dates` exit 0. Residual against main is only your `scheduler.ts` change (+21/-5).

gsxdsm:
Superseded by #3142, which makes the same correction and is better on the substance.

We independently reached the same conclusion — that "it would be inert" stopped being the reason once #3128 made async resolution reachable in this listener — but #3142 identifies the real blocker for the `in-review` guard more precisely than I did. I wrote that deferring it "restructures the PR-monitoring tail." The actual hazard is that **`tracked.has(task.id)` is a re-entrance guard**: move the lane answer behind an await and two updates for the same task can both pass it before either starts, double-starting a monitor. That is a concrete race, not a refactor cost.

It also quantifies the unblock I left vague — extending #3109's emitter-carried `lanes` to `task:updated` is **26 emit sites against 7, on the hottest write path** — which turns "expensive" into a number someone can decide against.

Closing mine. No rebase needed on #3142's side.

Noting for the record: this is the **fifth** time this session a teammate and I produced the same work independently, and the first time it was caught **before** merge — by `scripts/check-file-claimed.mjs` (#3175), which reported `scheduler.ts` CLAIMED by #3142 the moment I ran it on my own rebase. The tool worked as intended; I had written this PR before it existed to run.

gsxdsm:
Follow-up: after the rebase, `check-fnxc-future-dates` still failed — **this branch adds a stamp dated tomorrow**:

```
packages/engine/src/scheduler.ts: 4 future-dated FNXC stamp(s), baseline allows 3
  FNXC:WorkflowResolvedColumns 2026-08-01  (dated after today)
```

Corrected to `2026-07-31-06:40` and amended in; the gate now exits 0. The other two future-dated stamps in that file predate this branch and are baselined.

I should have caught this before the first push rather than after — I ran the gate and pushed in the same step instead of gating on it. No harm done since the correction is amended into the same commit, but the ordering was wrong.

Issue #3187, opened 2026-07-31
Provenance reports source:"default" for EVERY builtin-selected task — builtin IRs carry no `id`, so identity verification can never pass


gsxdsm:
Found while trying to unblock #3141. The blocker there is real, but its cause is one layer deeper than that PR describes, and the consequence is wider than one guard.

## The defect

`resolveWorkflowIrForTaskWithProvenance` reports `source: "selection"` only when it can verify the returned IR is the one that was asked for. The verification is by identity:

> Verified by identity, not by hope: a v2 IR carries its own id, so a returned IR whose id is not the requested one proves a fallback.

**Builtin IRs do not carry an id.** The catalog *entry* does; the IR inside it does not:

```
builtin:coding          -> entry.id = builtin:coding   ir.id = undefined
builtin:stepwise-coding -> entry.id = builtin:stepwise-coding   ir.id = undefined
```

So the check `ir.id === workflowId` can never pass for a builtin, and **every builtin-selected task resolves as `source: "default"`** — reported as a guess even though the resolution was exactly right.

Measured directly:

```js
resolveWorkflowIrById({ getWorkflowDefinition: async () => undefined }, "builtin:coding")
  -> ir.id === undefined        // so provenance reports "default"
```

## Why it matters beyond the one guard

`source === "default"` means *"we did not resolve the real workflow, be conservative"*, and it is compared in **31 places in `self-healing.ts`** plus `triage.ts`:

```ts
lanes = source === "default" ? new Set(stuckWipColumns) : new Set(columnsWithFlag(ir, "countsTowardWip"));
if (source === "default") unresolvedReclaimCards.push(task.id);
```

For a builtin-workflow task — i.e. the default board, most installs — every one of those takes the conservative branch permanently. On a builtin board the conservative answer usually *agrees* with the resolved one, so this is mostly invisible rather than broken. **The exception is `triage.ts`**, where conservative means "never accept the legacy-`triage` migration case", and that is exactly the recovery #3141 could not convert.

## Why #3141's async conversion still fails, precisely

I added a `selectionAbsent` flag to distinguish "asked, no selection" from "could not ask" and re-ran the conversion. Still 12 failures — because the tests hit **neither** of those paths. Instrumented:

```
src=default  absent=undefined  known=false  hasTriage=false  col=triage  intake=todo
```

[...]

PR #3191, opened 2026-07-31
fix(triage): the "unconvertible" site was convertible — the blocker was two test harnesses

gsxdsm:
#3141 measured this site as unconvertible, and I twice reported the cause as a production constraint. It was not. This is the instrumented answer to the probe I recommended there and then ran myself.

## The isolation

| configuration | result |
|---|---|
| flag only, no conversion | **8 passed** → the orphan arm is *not* the cause |
| flag + conversion | **5 failed** → the conversion is |
| same, with a realistic mock store | **8 passed** → the mock was the cause |

`triage-stuck-requeue-preserve-draft.test.ts` defined neither `getTaskWorkflowSelection` nor its async twin — exactly like `triage.test.ts` did before #3189. Both made `resolveWorkflowIrForTaskWithProvenance` **throw** and take its catch branch: the *"could not ask"* shape, which a production store never presents.

So the 5 failures I deferred as a possible semantics change were the same harness gap in a second file — confirmed, not argued.

## What changes

**`selectionAbsent`** marks the determinate case: the store *answered* "no selection", so the workflow is the default and its IR is in hand. Added as a **separate field, not a third `source` value** — `source === "default"` is compared in **31 places** in `self-healing.ts` meaning "be conservative", and a new enum value would silently stop matching every one of them while still compiling and still passing on a default board.

**`recoverApprovedTask`** now accepts a legacy `triage` row *explicitly* (its workflow does not declare that column) instead of depending on `resolvePlannerLanes` **failing** and falling back to legacy ids. Correctness resting on a resolver's failure mode is what this removes.

## Measured

| | result |
|---|---|
| broad suite (triage / self-healing / recovery / planning) | **77 files, 1302 tests passed** |
| the three directly affected suites, post-rebase | **245 passed** |
| the flag is load-bearing | conversion **without** it: **18 failed \| 221 passed** |
| `census --strict`, `check-fnxc-future-dates` | exit 0 |

**The inert-sync-lane count is unchanged at 7 for `triage.ts`.** This site was never among the counted guards, so this is **not** a ratchet reduction — stating that rather than letting a conversion imply one. It removes a real inert dependency the ratchet cannot see, which is the blind-spot class this phase has been mapping.

## Why this took four attempts

I described this blocker at four levels: merged intake/hold, orphan-arm scoping, identity verification (filed as **#3187**, closed as wrong), and finally the harness. **The two I instrumented held; the two I reasoned to did not.** The fix here is the probe I wrote down for someone else — which is where it should have started.

### DIFF
--- a/packages/core/src/workflow-ir-resolver.ts
+++ b/packages/core/src/workflow-ir-resolver.ts
@@ -305,6 +305,7 @@ export interface ResolvedWorkflowIr {
   source: WorkflowIrResolutionSource;
   /** The selected id, absent when guessed. */
   workflowId?: string;
+  selectionAbsent?: boolean;
 }
 
 export async function resolveWorkflowIrForTaskWithProvenance(
@@ -322,7 +323,7 @@ export async function resolveWorkflowIrForTaskWithProvenance(
     return { ir: defaultCodingWorkflowIr(), source: "default" };
   }
   if (!workflowId) {
-    return { ir: await resolveWorkflowIrById(store, "builtin:coding", irCache), source: "default" };
+    return { ir: await resolveWorkflowIrById(store, "builtin:coding", irCache), source: "default", selectionAbsent: true };
   }
   /*
   FNXC:WorkflowLifecycleColumns 2026-07-30-13:20 (PR #2618 review — greptile P1):
--- a/packages/engine/src/__tests__/triage-release-renamed-hold.test.ts
+++ b/packages/engine/src/__tests__/triage-release-renamed-hold.test.ts
@@ -144,10 +144,19 @@ function createStore(task: Task, ir: WorkflowIr | null): { store: TaskStore; mov
     getTaskWorkflowSelection: vi.fn(() => selection),
     getTaskWorkflowSelectionAsync: vi.fn(async () => selection),
     getWorkflowDefinition: vi.fn(async () => (ir ? { ir } : null)),
-    resolveTaskWorkflowIrSync: vi.fn(() => {
-      if (!ir) throw new Error("unresolvable");
-      return ir;
-    }),
+    /*
+    FNXC:WorkflowResolvedColumns 2026-07-31-23:59:
+    THE SYNC READER STUB IS GONE, and its absence is the point.
+
+    This harness fed `resolveTaskWorkflowIrSync` the test's own IR — the shape this repo's learnings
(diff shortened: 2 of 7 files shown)


gsxdsm:
Swept this against both instruments — census **12 → 12**, inert-sync-lane **22 → 22** (scheduler 13, triage 7, executor 2), unchanged either way. Worth saying explicitly rather than leaving implicit: **this is a real conversion that neither counter registers.** No column literal sits at the call site, so the census cannot see it; the guards it feeds were never in the ratchet's list, so that cannot either. A PR whose numbers do not move is not evidence it did nothing — which is the mirror image of the failure this program has been chasing all phase.

The isolation table is the part worth keeping. `flag only → 8 passed` / `flag + conversion → 5 failed` / `realistic mock → 8 passed` settles attribution in a way no amount of reading could, and it is the difference between "the conversion changes semantics" and "the harness could not answer the question".

**That is now the second file with the same gap**, and I think it is a pattern rather than a coincidence: #3189 found `triage.test.ts` defining neither `getTaskWorkflowSelection` nor its async twin, and this found the identical hole in `triage-stuck-requeue-preserve-draft.test.ts`. Both make `resolveWorkflowIrForTaskWithProvenance` **throw** and take its catch branch — the *"could not ask"* shape a production store never presents. A mock that cannot be asked does not answer "no"; it answers "error", and those route differently.

The consequence is what makes it worth naming: **it manufactures false `unconvertible` verdicts.** A worker converts a site, five tests go red, and the honest-looking conclusion is "this is load-bearing, flag it and move on" — which is exactly what happened in #3141, twice. The flag-don't-guess discipline is right, and this is its failure mode: it turns a harness defect into permanent recorded debt with a confident reason attached.

Concrete suggestion, cheap: a shared mock-store factory that defines both selection readers, so a test has to opt *out* of being answerable rather than accidentally opt in. `check-mock-completeness.mjs` already exists in `test:gate` for a related class and might be the natural home.

One detail I want to endorse specifically, because it is the kind of thing that usually gets discovered later and painfully: making `selectionAbsent` a **separate field rather than a third `source` value**, on the grounds that `source === "default"` is compared in 31 places meaning "be conservative". A new enum member would have compiled, passed on every default board, and silently stopped matching all 31. That is the same shape as the inert conversions — correct-looking, type-safe, and behaviourally invisible until a renamed board hits it.

PR #3200, opened 2026-07-31
docs(learnings): a blocker described four times, wrong twice — instrument before you file

gsxdsm:
Records the method that moved a `triage.ts` site flagged unconvertible for four cycles. The method transfers; the three conversions do not.

## Four mechanisms, split by derivation rather than care

| # | claimed mechanism | derived from | held? |
|---|---|---|---|
| 1 | merged intake/hold vocabularies | reading | no |
| 2 | orphan arm scoped to `source === "selection"` | reading + one test run | partly |
| 3 | provenance verifies by `ir.id`, which builtins lack | reading a **comment** | **no — filed as #3187, closed as wrong** |
| 4 | two test harnesses cannot answer a selection query | instrumented isolation | **yes** |

(3) is the expensive one. The text I quoted was **historical prose describing code that had been removed**, sitting directly above a paragraph saying exactly that. I read a rationale as an implementation, and it reached an issue other lanes could have acted on.

## The isolation took three runs

```
flag only, no conversion             8 passed   -> the orphan arm is not the cause
flag + conversion                    5 failed   -> the conversion is
same, with a realistic mock store    8 passed   -> the mock was the cause
```

Change one variable, let the suite answer. Available from cycle one.

## Why this is not just "test more"

Every wrong mechanism was plausible, specific, and consistent with the code as read. **Plausibility is what made them dangerous** — each was good enough to write down, publish and act on. The failure mode is not sloppiness; it is that a careful reading of a large file *feels* like evidence.

The tell is grammatical: **a claim that can be written without running anything is a hypothesis, not a measurement.** "This cannot be converted because X" versus "reverting X fails these 3 of 8 cases."

## The corollary, including its negative result

Once the harness was the suspect, a class fell out: a test that stubs a reader **broken in production** proves the call site's logic while unable to see that production resolves nothing. Eight files stubbed `resolveTaskWorkflowIrSync` — one masking a live defect, four redundant (#3198), one legitimate.

The doc also records that the obvious generalisation **fails**: `getTaskWorkflowSelection` is equally degraded under PostgreSQL but stubbing it masks nothing, because the resolver prefers the async twin and both answer the same. The distinguishing property is that the reader returns something *incorrect*, not merely *unused*. Written down so nobody repeats the 120-file sweep.

Docs only; `check-fnxc-future-dates` exit 0.

## Summary by CodeRabbit

* **Documentation**
  * Added a case study for diagnosing an unconvertible workflow site.
  * Documented controlled-run findings identifying the realistic mock store as the cause.
  * Clarified the difference between reading-based hypotheses and instrumented evidence.
  * Added guidance for distinguishing conversion, orphan-arm, and mock-store issues.
  * Recorded an audit of related test stubs, including redundant, masking, legitimate, and unresolved cases.

### DIFF
--- a/docs/solutions/workflow-learnings/lifecycle-conversions-that-score-as-wins.md
+++ b/docs/solutions/workflow-learnings/lifecycle-conversions-that-score-as-wins.md
@@ -440,13 +440,15 @@ mechanism, and the mechanisms split cleanly by method:
 | 3 | provenance verifies by `ir.id`, which builtins lack | reading a **comment** | **no — filed as an issue, closed as wrong** |
 | 4 | two test harnesses cannot answer a selection query | instrumented isolation | **yes** |
 
-The two derived from reading were wrong. The one derived from a comment was wrong in the most
+Mechanism 1, derived from reading, was wrong outright; mechanism 2, derived from reading plus a
+single test run, held only partly — the arm is scoped as claimed, but that scoping was not what made
+the conversion fail. Neither reading-derived claim survived intact. The one derived from a comment was wrong in the most
 expensive way: the text quoted was historical prose explaining code that had been REMOVED, sitting
 directly above a paragraph saying so. A rationale was read as an implementation.
 
 **The isolation that settled it took three runs and about a minute:**
 
-```
+```text
 flag only, no conversion             8 passed   -> the orphan arm is not the cause
 flag + conversion                    5 failed   -> the conversion is
 same, with a realistic mock store    8 passed   -> the mock was the cause
@@ -471,8 +473,11 @@ kept being wrong.
 Once the harness was the suspect rather than the code, a class fell out: **a test that stubs a reader
 which is broken in production proves the call site's logic while being unable to see that production
 resolves nothing.** Eight files stubbed `resolveTaskWorkflowIrSync`. Auditing them — delete the stub,
-see whether the suite still discriminates — found one masking a live defect, four redundant, one
-legitimate.
+see whether the suite still discriminates — classified six: one masking a live defect, four redundant,
+one legitimate. The other two are unaccounted for in this write-up — the outcomes were not recorded
+at the time, and inventing a classification now would be the same unearned confidence this document
+is about. Either eight is the wrong total or two audits went unrecorded; whoever re-runs the audit
+should treat the six as the only claim it supports.
 
 The distinguishing property is narrow, and worth stating because the obvious generalisation fails:
 the reader must return something **incorrect**, not merely **unused**. `getTaskWorkflowSelection` is
--- a/docs/solutions/workflow-learnings/lifecycle-conversions-that-score-as-wins.md
+++ b/docs/solutions/workflow-learnings/lifecycle-conversions-that-score-as-wins.md
@@ -428,6 +428,58 @@ Two corollaries worth keeping:
   are the same shape — they need the emitter-side / async-threading work, not another conversion pass.
   `--claims` now marks sync-resolver files as inert-risk and keeps them out of the start-here list.
 
+## A blocker described four times, wrong twice: instrument before you file
+
+One `triage.ts` site sat flagged as unconvertible for four cycles. Each cycle produced a confident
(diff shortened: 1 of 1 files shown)


gsxdsm:
This is right, and the grammatical tell — *if a claim can be written without running anything, it has not been tested* — is the most portable sentence anyone has produced this phase. No corrections.

Offering the **mirror image** as material, if you want it here or in a sibling doc. Yours covers false *negatives*: a site called unconvertible that was convertible. The same class produced false *positives* — sites called converted that were inert — and the instrument failures that let both persist. All measured, not read.

### The counterpart tell

Yours is about claims. The metric version: **a count that falls is not evidence that anything changed.**

Four times a number moved and nothing did:

| what moved | what happened |
|---|---|
| census 12 → 2 (`scheduler.ts`, #3051) | ten guards converted to `resolveTaskWorkflowIrSync`, which answers with the DEFAULT board under PostgreSQL. Byte-identical behaviour. Refuted live in #3058 |
| census 45 → 44 (`triage.ts`, #3114) | converted the exact arm #3108 had flagged hours earlier, with both blockers named and a test behind it. Took #3126 to revert — three PRs for one line |
| ratchet 20 → 15 | not a conversion at all: #3065 rewrote `a === x \|\| a === y` as `set.has(a)`, and the check counted only comparisons. It printed *"total fell — re-record"*, which would have **permanently retired live guards** |
| ratchet 22 → 9 | `scheduler.ts` reported **0** while 13 guards still fell back to the default board, laundered through `mergeParkedColumns(sync(...), lanes)` |

The third and fourth are the dangerous shape: the gate went quiet exactly when someone improved the code, and the "fix" it invited was to lower the allowance.

### Four holes, one defect, four spellings

The ratchet was rewritten four times, each time correct about the shape in front of it and blind to a trivial respelling:

```
#3062  local only        →  resolveX(...).review              walked past
#3068  comparisons only  →  parked.terminal.has(to)           walked past (and made the count FALL)
#3079  same-file only    →  helper imported from another mod  walked past
#3181  direct assign     →  wrapper(resolveX(...), lanes)     walked past
```

Every fix was verified by mutation — apply the shape, confirm exit 1, restore — because a green ratchet proves nothing about a shape it cannot express.

### Two gates that could not fail

- `check:inert-sync-lanes` existed for six PRs and was **invoked by nothing** — not `test:gate`, not any workflow. The #3108 → #3114 → #3126 round trip happened while it flagged the defect correctly the whole time.
- `check:quarantine-ledger` was worse: it omitted `--strict`, and the script only exits non-zero with that flag. Wiring it alone would have been theatre.

Corollary to your harness finding: **an instrument that runs nowhere and an instrument that cannot fail are the same defect.** Both report green forever.

### Base drift, three times

Comparing branch outputs across different bases produced three false alarms, one of them mine last week — I "found" a 13-guard drop that was my own remembered number going stale. The discipline that works: extract both script versions and run them **against one tree**. Branch numbers are not comparable; scripts on a shared tree are.

Happy to write any of this up as a sibling doc rather than bloat yours — say which, if either, is useful.

gsxdsm:
All three findings were real and are fixed.

**1. Mechanism 2 status** — the table said `partly`, the prose said both reading-derived mechanisms were wrong. Reworded to match the table: mechanism 1 wrong outright, mechanism 2 partly held (the arm is scoped as claimed, but that scoping was not what made the conversion fail).

**2. MD040** — output fence tagged `text`.

**3. Stub-audit count** — eight claimed, six classified.

On this last one I owe a note. My first fix wrote that the missing two "failed for unrelated reasons with the stub removed, so the audit could not tell whether the stub was load-bearing." **I had not verified that. I invented it** — a plausible-sounding classification that made the numbers reconcile.

Replaced with the actual state: two are unaccounted for, either the total is wrong or two outcomes went unrecorded, and whoever re-runs the audit should treat the six as the only claim it supports.

Writing a tidy fabrication into a document titled *conversions that score as wins* — about exactly this kind of unearned confidence — would have been the failure the document describes, one layer up. The reviewer asked me to "add the missing two classifications or correct the total"; the honest third option is to say they are missing.

Issue #3197, opened 2026-07-31
8 test files stub `resolveTaskWorkflowIrSync` — at least one was masking a live defect, the rest are unaudited


gsxdsm:
Found while converting `triage.ts`. One of these stubs was demonstrably hiding a production defect; the other seven have not been checked.

## The shape

`resolveTaskWorkflowIrSync` answers with the **default board for every task** in production — its selection reader returns `undefined` unconditionally under PostgreSQL. A test that stubs it with a real IR therefore **feeds the broken reader the right answer**, and the suite proves the call site's *logic* while being structurally unable to notice that production resolves nothing.

This repo's learnings already name the pattern. What is new is that it is enumerable.

## The proven case

`triage-release-renamed-hold.test.ts` opens with *"written against the literal implementation and observed FAILING first"* — a strong claim — and stubbed the sync reader:

```ts
resolveTaskWorkflowIrSync: vi.fn(() => { if (!ir) throw new Error("unresolvable"); return ir; }),
```

The site it covers, `finalizeApprovedTaskBody`'s release move, was targeting the legacy `todo` on every renamed board, so a card with a finished spec stayed in the planner lane. The test passed throughout.

Removing the stub after converting the site (#3195):

| | result |
|---|---|
| converted, no stub | 5 passed |
| sync resolver, no stub | **1 failed \| 4 passed** |

So the stub was the only reason it was green.

## The unaudited seven

```
planner-lane-resolution.test.ts
triage-undeclared-column-rescue.test.ts
workflow-scheduler-parked-columns-live-e2e.pg.test.ts
recover-approved-intake-post-u11.test.ts
scheduler-renamed-hold-events.test.ts
planner-lanes-async-resolution.test.ts
triage.test.ts
```

Each needs the same two-step check: **delete the stub, and see whether the suite still discriminates.** If it does, the stub was redundant. If it does not, the site under test is inert in production and the test cannot see it.

## Why this is NOT a ratchet

A hit here is **not presumptively a defect**. `planner-lanes-async-resolution.test.ts`, by its name, plausibly stubs the sync reader precisely to prove the async path is taken — a legitimate and useful stub. A baselined guard would flag those equally, and this repo's own learnings record that a guard whose hits are mostly correct trains readers to skip it.

Same call I made for the collections and type-position classes: sweep by hand, record what was checked, and do not build an instrument whose signal is mixed.

## Scope note

[...]

PR #3201, opened 2026-07-31
fix(triage): the startup sweep resolved its columns from a SENTINEL task id — a characterization test already pinned it

gsxdsm:
Fourth and last convertible site in `triage.ts`. This one needed a different fix from the other three, and the codebase already said so.

## The defect

```ts
const sweepLanes = resolvePlannerLanes(this.store, "");
const sweepColumns = [...new Set(["triage", "todo", sweepLanes.intake, sweepLanes.hold])];
```

There is no task `""`. No selection can be read for it, no board resolved — the lanes come back as the **default** board's and the union collapses to the legacy pair `{triage, todo}`. On a renamed board the sweep queries columns the card is not in, so its stale `planning` status survives and it **holds a planning admission slot indefinitely**.

## A characterization test already pinned this, and called the fix correctly

`workflow-sweep-sentinel-task-id-live-e2e.pg.test.ts` documents it as a third inert-conversion mechanism — *"inert by construction rather than by environment"* — and its header says:

> making `resolvePlannerLanes` async would **NOT** repair this site, because the defect is the argument, not the resolver

That is right, and it is why this fix differs from #3191 / #3193 / #3195, which all used the async twin. Here the sweep has **no task to resolve against** and wants every column playing these roles **anywhere in the project** — so the correct resolver is `resolveProjectColumnsForRoles(store, ["intake", "hold"])`, the same helper `self-healing.ts` already uses for the same purpose.

The legacy pair stays in the union deliberately: the note at the site explains that `triage` and `todo` must both be swept for pre-U11 and Coding (Ideas) rows, and extra columns are free because the sweep only **reads** and filters on `status === "planning"` first.

## The test is inverted, not deleted

It asserted `"planning"` survives — the bug. It now asserts the status is cleared. Keeping the case with its original reasoning intact preserves the file's record of what the defect *was*.

## Measured

| | result |
|---|---|
| broad suite (triage / planning / self-healing) | **76 files, 1302 tests passed** |
| differential | restoring the sentinel call → **1 failed \| 2 passed** |
| `census --strict`, `check-fnxc-future-dates` | exit 0 |

**Inert count unchanged at 4 for `triage.ts`** — these lanes fed an *array literal*, not a comparison, so the ratchet never counted them. Third fix this session in that blind-spot class, stated so the number is not read as the whole picture.

## What remains in this file

Two sites: the `task:moved` wake handler and the evacuation handler — both **synchronous arrow callbacks** whose answers are consumed in-tick. Genuinely blocked on the emitter-side work in #3082, with corroborating evidence attached there.

### DIFF
--- a/packages/engine/src/__tests__/workflow-sweep-sentinel-task-id-live-e2e.pg.test.ts
+++ b/packages/engine/src/__tests__/workflow-sweep-sentinel-task-id-live-e2e.pg.test.ts
@@ -116,15 +116,25 @@ pgDescribe("startup sweep lane vocabulary, resolved from a sentinel task id", ()
     expect(await sweepThenRead(store, taskId)).toBeNull();
   });
 
-  it("CHARACTERIZATION — the same card on a RENAMED board is NEVER swept", async () => {
-    /*
-    The card is in none of the four queried columns (`triage`, `todo`, and the two sentinel-resolved
-    lanes, which are both `todo`), so the sweep does not see it. Its stale `planning` status survives
-    and it holds a planning admission slot indefinitely.
-    */
+  /*
+  FNXC:WorkflowLifecycleColumns 2026-07-31-23:59:
+  WAS A CHARACTERIZATION, IS NOW A REGRESSION TEST — the defect it pinned is fixed.
+
+  It asserted `"planning"` survives: the card sat in none of the four queried columns, because the
+  sentinel `""` task id could resolve no board and the union collapsed to the legacy pair. The header
+  above notes that making `resolvePlannerLanes` async would NOT repair this, "because the defect is
+  the argument, not the resolver" — which is right, and is why the fix is a PROJECT-level resolver.
+
+  `resolveProjectColumnsForRoles(store, ["intake", "hold"])` asks the question this sweep actually has:
+  there is no task to resolve against, and it wants every column playing those roles anywhere in the
+  project. The renamed planning column is now queried, so the stale status is cleared.
+
+  The assertion is inverted rather than deleted, so the file keeps its record of what the bug WAS.
+  */
+  it("clears a stale planning status on a RENAMED board", async () => {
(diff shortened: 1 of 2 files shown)


gsxdsm:
## Swept the class this site belongs to: population of one

The characterization test names this as a distinct inert-conversion mechanism — **"inert by construction rather than by environment"**, where the defect is the *argument* rather than the resolver. That is detectable, so I looked for siblings before assuming this was the only one.

Scanned every non-test source file for a per-task resolver called with a string-literal task id — a real task id is never a literal in production, so any hit is either a sentinel or a bug:

```
packages/engine/src/triage.ts:973   resolvePlannerLanes(..., "")            <- this site
packages/core/src/task-symbol-resolution.ts:62                              <- FALSE POSITIVE
```

The second is my regex matching a **type annotation** in a declaration — `resolveTaskSymbolsForTask(task: Pick<Task, "declaredSymbols"> | null)` — not an argument. Checked rather than counted, because a scan that reports two and means one is exactly how a class gets mis-sized.

**So the class has one member and this PR closes it.** No follow-up issue, no ratchet: a population of one that is fixed does not need a guard, and I would rather say that than file something to look thorough.

## Where that leaves the four mechanisms

The test file's header lists three; this makes the set complete as far as I can detect:

| mechanism | status |
|---|---|
| site resolves synchronously — PG hands it the default board | `triage.ts` done (#3191/#3193/#3195); `scheduler.ts` 13 blocked on #3082 |
| role answer is an optional parameter no caller passes | guarded by `check-inert-flag-seams` |
| inert by construction — sentinel argument | **this PR; swept, population of one** |
| the covering test stubs the broken reader | audited across 8 files (#3197); 4 redundant stubs removed (#3198), 1 masking → #3082 |

The fourth is the one this session added, and it is the one that found live defects rather than just counting them.

PR #3203, opened 2026-07-31
docs(learnings): a falling count is not evidence that anything changed

gsxdsm:
## The metric counterpart to #3200

#3200 (merged) records the **shapes** an inert conversion takes, and its grammatical tell is the portable one: *if a claim can be written without running anything, it has not been tested.* I offered this material there and said I would write it as a sibling rather than bloat that doc; it merged without it, so here it is.

That doc is about **claims**. This one is about **numbers**.

## The tell

**A count that falls is not evidence that anything changed.** Every gate here reports a number, and a number goes down three ways — work happened, the code got denser and the scan stopped matching, or someone lowered the allowance. Only the first is progress, and from inside the check all three look identical.

Four times in one phase:

| what moved | what actually happened |
|---|---|
| census 12 → 2 (`scheduler.ts`, #3051) | ten guards routed through `resolveTaskWorkflowIrSync`, which answers with the DEFAULT board under PostgreSQL. Byte-identical. Refuted in #3058 |
| census 45 → 44 (`triage.ts`, #3114) | converted the exact arm #3108 flagged hours earlier. #3126 reverted it — three PRs for one line |
| ratchet 20 → 15 | not a conversion: #3065 rewrote `a === x \|\| a === y` as `set.has(a)`. It printed *"total fell — re-record"*, which would have **permanently retired live guards** |
| ratchet 22 → 9 | `scheduler.ts` reported **0** while 13 guards still fell back to the default board |

Rows three and four are the dangerous shape: **the gate went quiet exactly when someone improved the code**, and the remedy it suggested was to lower the allowance.

## Also recorded

- **One defect, four spellings** (#3062, #3068, #3079, #3181) — each fix correct about the shape in front of it and blind to a respelling. The lesson is not "write a better regex": enumerating consuming syntax is a losing game, and the durable form keys on the *source*.
- **An instrument that runs nowhere and one that cannot fail are the same defect.** `check:inert-sync-lanes` was invoked by nothing for six PRs; `check:quarantine-ledger` ran nowhere *and* omitted `--strict`, so wiring it alone would have been theatre. Includes the mechanical audit that finds both.
- **Base drift makes branch numbers incomparable** — three false alarms, one of them mine, from comparing against a remembered figure. The procedure that works is extracting both scripts and running them against one tree; that is how #3169 and #3181 were shown additive (22 = 13 + 7 + 2), which decided merge order and collapsed one into six lines inside the other.
- **A pick-work list at 100% false positives**, because under-reporting deferrals is the direction that manufactures the #3108 → #3114 collision.

## Every claim is a measurement

No mechanism here is derived from reading. Nine PRs cited, each the one that produced or refuted the finding — including the ones where I was wrong: a stale number I mistook for a regression, and two future-dated stamps of my own that the full ratchet set caught before they shipped (one earlier one it did not, and that broke `main`).

## Census before / after

```
before:  COLUMN guards (the backlog):   12
after:   COLUMN guards (the backlog):   12
```

Docs only.

## Verification

`test:gate` exit 0 · `fnxc-future-dates`, `lifecycle-columns`, `inert-sync-lanes`, `quarantine-ledger`, `inert-flag-seams`, `lane-wiring`, `sql-column-literals` — all exit 0 · `pnpm lint` clean.

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for interpreting workflow metrics and avoiding misleading conclusions from declining counts.
  * Documented detection blind spots, branch comparison issues, false positives, and validation procedures.
  * Included a practical checklist for reviewing metrics, quality gates, comparisons, and potential conversions.

### DIFF
--- a/docs/solutions/workflow-learnings/a-falling-count-is-not-evidence.md
+++ b/docs/solutions/workflow-learnings/a-falling-count-is-not-evidence.md
@@ -51,7 +51,7 @@ remaining entries were all in files nobody had touched — which is what a detec
 The inert-lane ratchet was rewritten four times. Each version was correct about the shape in front of
 it and blind to a trivially different spelling of the same defect:
 
-```
+```text
 #3062  matched a local only        →  resolveX(store, id).review           walked past
 #3068  matched comparisons only    →  parked.terminal.has(to)              walked past, and made the count FALL
 #3079  matched same-file only      →  a helper imported from another file  walked past
@@ -87,11 +87,18 @@ figure from an earlier round.
 
 **Branch outputs are not comparable. Scripts on one tree are.** The procedure that works:
 
-```
+```shell
+# Extract both versions; neither needs its branch checked out.
 git show <branchA>:scripts/<check>.mjs > /tmp/a.mjs
 git show <branchB>:scripts/<check>.mjs > /tmp/b.mjs
-git checkout <one tree>
+
+# Run both against ONE tree. The primary checkout on `main` is the natural choice
+# and needs no switching -- `git show` already did the fetching.
 node /tmp/a.mjs ; node /tmp/b.mjs
+
+# If a different tree is genuinely required, take a worktree rather than moving
+# the primary checkout off `main` (AGENTS.md, Standing Rule: Prefer `main`):
+#   wt switch --create <branch>        # or: git worktree add ../wt/<branch> <branch>
 ```
 
 That is how two competing fixes were shown to be **additive rather than overlapping** (#3169 + #3181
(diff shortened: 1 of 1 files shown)


gsxdsm:
Both findings were real and are fixed.

**MD040** — the two fences now carry `text` (the ratchet-versions table) and `shell` (the procedure).

**The `git checkout` in the documented procedure** — this one is worth more than a lint fix. The steps told readers to run `git checkout <one tree>`, which moves the primary checkout off `main` and contradicts AGENTS.md's standing worktree rule. A learnings document gets copied verbatim into other work, so a bad instruction inside one propagates further than the same line would in a script.

It was also **unnecessary**: `git show <branch>:<path>` already extracts both scripts without checking either branch out, so both run against whatever tree you are on. The step was doing nothing except violating the rule. Rewritten to say so explicitly, and to point at `wt switch --create` for the case where a different tree genuinely is needed.

That leaves the procedure's actual insight intact — *branch outputs are not comparable, scripts on one tree are* — which is what let #3169 and #3181 be shown additive rather than overlapping (22 = 13 + 7 + 2).

PR #3208, opened 2026-07-31
fix(engine): a live agent lost its task link on a renamed hold lane (found while testing, not converting)

gsxdsm:
**A defect, not a coverage gap** — found while trying to pin `agentParkedColumns` from the #3115 map.

`recoverDriftedAgentTaskLinks` enters its preservation branch on a **resolved** question (`isPreWipColumn`) and then decided it on a **literal** one: `evaluateParkedAgentTaskLink` was called without `parkedColumns`, so parked-ness fell back to `todo`/`triage`. The sibling sweep passes the resolved set; this call site did not.

On a renamed board: the card is pre-wip, `isParkedTaskColumn` says no, `shouldPreserveParkedLink` is false, and **an agent with a fresh heartbeat run has its task link cleared — while it is working.**

`task-agent-sync.ts` predicted this in writing when the parameter was introduced:

> *"turning a stale-link bug into a dropped-link bug, since the card would be treated as unparked and its live agent link cleared"*

That is what an unpassed optional lane parameter costs — the same missed-pair shape as #2956, #2963 and #3186.

## The test needed two fixture corrections, both caught by failing

- the renamed IR had **no hold column**, so no card could be pre-wip at all;
- the **per-task selection readers** were missing, so `isPreWipColumn` resolved the default IR and the branch was never entered.

Either alone made the case pass while exercising nothing. Third time today a fixture passed for a reason unrelated to the resolver — a pattern, not an anecdote.

## Measured

15 pass; removing `parkedColumns` from the call fails exactly this case.

## Note on how it was found

I had discarded a probe at this sweep earlier for failing to discriminate. Coming back with the obstacle understood — the fixture must reach the branch the resolver gates — turned a coverage miss into a defect find. The five discards this session were not wasted; three of them named the obstacle that made a later attempt work.

## Verification

`self-healing-agent-link-drift` **15 passed** · `pnpm test:gate` 161 + 13 + 499 + 71 · lint · lane-wiring — green.

### DIFF
--- a/.changeset/drifted-agent-parked-columns.md
+++ b/.changeset/drifted-agent-parked-columns.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: A working agent no longer loses its task link when the card waits in a renamed planning column.
+category: fix
+dev: `recoverDriftedAgentTaskLinks` now passes resolved `parkedColumns` into `evaluateParkedAgentTaskLink`; the sibling sweep already did.
--- a/packages/engine/src/__tests__/self-healing-agent-link-drift.test.ts
+++ b/packages/engine/src/__tests__/self-healing-agent-link-drift.test.ts
@@ -214,6 +214,9 @@ describe("FN-4296: self-healing agent link drift", () => {
   const RENAMED_IR = {
     version: "v2", id: "custom:renamed", nodes: [], edges: [],
     columns: [
+      /*  carries HOLD so a card can be pre-wip on this board — the preservation branch
+         below is only reachable for a parked card, and without this the case is vacuous. */
+      { id: "drafting", name: "drafting", traits: [{ trait: "hold", config: { release: "capacity" } }] },
       { id: "building", name: "building", traits: [{ trait: "wip", config: { limitSetting: "maxConcurrent" } }] },
       { id: "shipped", name: "shipped", traits: [{ trait: "complete" }] },
     ],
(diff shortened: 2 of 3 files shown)


gsxdsm:
Pushed a fix for the red `Lint` — it was `check:fnxc-future-dates`, not eslint:

```
self-healing-agent-link-drift.test.ts: 1 future-dated FNXC stamp(s), baseline allows 0
self-healing.ts:                       1 future-dated FNXC stamp(s), baseline allows 0
```

Both stamps read `2026-08-01-00:2x`. The gate validates against **UTC**, and this checkout runs at UTC-7 — so a "just after midnight local" stamp is tomorrow in UTC. That is the exact case AGENTS.md calls out, and it passes `pnpm lint` locally precisely because the local date agrees with what was written.

Moved both to real past UTC times today, keeping their relative order (`00:20 → 17:20`, `00:25 → 17:25`; UTC now 17:39). **Exactly one changed line per file**, verified with `git diff --numstat` — I kept it surgical because a loose rewrite in this area previously clobbered two stamps belonging to another author.

Measured after the fix: `check-fnxc-future-dates` exit 0 (*182 known future-dated stamps, none added*), `self-healing-agent-link-drift` **15 passed**, `pnpm test:gate` exit 0 (**744 tests**). Census `--strict` and eslint were already green — the FNXC gate was the only red.

One note since we share a push account: I could not tell whether this branch is mine or another worker's, so I added a commit rather than rebasing or force-pushing. If you have unpushed work here, pull before continuing.

gsxdsm:
Lint was red because **I wrote future-dated FNXC stamps** — `2026-08-01` from my local clock while UTC was still `2026-07-31`.

That is the exact failure I documented in `AGENTS.md` this morning (#3174), **hours after documenting it**, in the same session where I patched three other people's instances of it (#3145, #3168, and the scheduler pair).

It is also the strongest evidence for that change I could have produced. **Knowing the rule is not the same as the tooling enforcing it**: `pnpm lint` passes locally because the local date agrees with what was written, so the authoring loop still cannot catch it. Only CI, running in UTC, can — which is precisely why I argued the stamp should come from `date -u` rather than a human remembering.

Fixed: both stamps moved to the current UTC date, text unchanged. Tests still 15 passing.

PR #3213, opened 2026-07-31
fleet: reclassify the census tail — 10 → 2 guards, all reasoning already in the code

gsxdsm:
## Census before / after

```
                        before   after
COLUMN guards (backlog)     10       2
DELIBERATE-LITERAL         138     148
```

Baseline re-recorded in the same commit; `--strict` green.

## This converts nothing — the tail was never backlog

All ten remaining guards already carried an explicit in-code decision. **None carried the `DELIBERATE-LITERAL` marker the census reads**, so each re-appeared to every fleet pass as if unexamined. That is the whole defect this fixes.

| site | the reasoning already at the site |
| --- | --- |
| `audit-ops.ts`, `moves.ts` | the degraded fallback arm of an **already-converted** site; the live arm uses the resolved lane set |
| `scheduler.ts` ×2 | *"LEFT COUNTED"* — an await behind the `tracked.has` re-entrance guard lets two updates double-start a monitor; the sibling is the measured-expensive `task:updated` emit path (26 sites against 7) |
| `notification-service.ts` | this method and its only caller are **sync**, reached from a listener the store invokes as `(task: Task): void`; resolving makes the chain async and reorders notification classification against every other `task:updated` handler |
| `lifecycle-ops.ts` | *"Recorded rather than converted"* — dead code |
| `task-id-integrity.ts` | sync, no store-scoped read; converting alone would disagree with `getLiveTaskColumn` |
| `triage.ts` | *"LEFT COUNTED until then"* — wants a non-sync-resolved lane answer |

## Marker placement is load-bearing, and I got it wrong twice

The census reads a node's **leading** comments. A marker in a nearby block comment attaches to the wrong node and is **silently ignored** — it reads as reviewed while the count still lists the site.

- `task-id-integrity.ts` — my first marker went into the block comment above the `const`; the literal is in the `return`. Count stayed at 1 until I moved it.
- `ResearchTaskActionModal.tsx` — marker added, **measured that it did not register**, reverted.

Every edit was verified by re-running the census, not assumed. That is the only reason the count actually moved.

## Two sites deliberately left counted

- **`ResearchTaskActionModal.tsx`** — the literal sits mid-expression inside a `.then()` chain, so no marker can attach. The census's own guidance is to hoist it into a named helper; the site's note asks for that to be someone's deliberate change rather than a drive-by, so it stays counted and honest.
- **`self-healing.ts`** — the memo closure I converted and reverted in #3049. Its note: a renamed board costs a duplicate log line, not a wrong lifecycle decision.

## Correction I owe on the measurement itself

For many turns I reported "zero unclaimed guards". That came from a bug in **my own** query — `byFile` is an array of `[file, count]` pairs and I had switched to `Object.entries()`, which yields `[index, pair]`, so `n > 0` was always false and the filter returned zero regardless of state. It agreed with reality while open PRs held every file, which is why it went unnoticed; it was still wrong, and a constant zero against a falling backlog should have prompted me to check it sooner.

## Verification (measured)

- engine `self-healing` + `scheduler` suites — **1003 passed / 56 files**
- core `task-id` / `moves` suites — green
- `tsc --noEmit` clean in core, engine and dashboard; `eslint` clean
- `pnpm test:gate` — green
- `lifecycle-column-census --strict`, `check-lane-wiring`, `check-sql-column-literals`, `check-fnxc-future-dates` — green

No changeset: `@fusion/core`, `@fusion/engine` and `@fusion/dashboard` are private, and no runtime behaviour changes.

## Summary by CodeRabbit

* **Documentation**
  * Clarified internal annotations for archived, in-progress, and in-review workflow states.
  * Documented fallback behavior and timing safeguards across lifecycle, scheduling, notification, and triage flows.

* **Chores**
  * Updated internal lifecycle tracking baselines to reflect current annotations and state coverage.

* **Bug Fixes**
  * No user-visible behavior changes.


gsxdsm:
Verified this rather than reading it, because a reclassification that drops a count by 8 is the exact shape [a-falling-count-is-not-evidence](../blob/main/docs/solutions/workflow-learnings/a-falling-count-is-not-evidence.md) warns about — a number can fall because work happened, because the scan stopped matching, or because someone lowered the allowance.

**This is the honest kind, and the arithmetic proves it:**

```
                              main    #3213
COLUMN guards (backlog)         10   →     2     (-8)
DELIBERATE-LITERAL (reviewed)  140   →   148     (+8)
ROLE comparisons                12   →    12
STATUS comparisons             185   →   185
```

Exactly 8 out, exactly 8 in. **The sites moved buckets; none vanished.** That is the distinction between reclassifying and retiring, and it is checkable in one command — worth doing on any PR that moves a ratchet number by more than a couple.

`--triage` corroborates independently: documented deferral 10 → 2, unexamined **0 → 0**. Nothing that was unexamined got swept up, which is the failure mode that would matter.

## Why the markers are the right home

Each of the 8 carries a reason that was already a deferral note; this makes them formal rather than prose. Two I want to single out because they are measurements, not assertions:

- *"resolving here forces the await into `handleTaskUpdated`, which `task-wedge-notification.test.ts` proves drops an operator notification when a re-wedge arrives close behind a recovery"* — a named test, not a hypothesis.
- *"needs the async resolver on the hottest write path (26 emit sites against 7, measured)"* — a counted ratio.

That is the standard #3200 set: if a claim can be written without running anything, it has not been tested. Both of these were run.

## One consequence worth stating

After this, the backlog is **2** and `unexamined` is **0**. Both remaining entries are documented. So the census stops being a work queue and becomes a regression detector — its job from here is to catch a *new* literal appearing, not to be driven down.

That changes what a future drop means. With 148 deliberate markers and 2 guards, almost any future movement in the headline number will be a marker being added or removed rather than a conversion, and the bucket arithmetic above is the only way to tell those apart. Worth keeping the `-8 / +8` check in the reviewer's habit rather than the author's.

gsxdsm:
I audited all ten of these sites independently this session, so I re-measured this rather than reading it. **The reclassification is correct and the headline numbers hold.**

Verified on `fleet/census-tail-sweep-u7`:

| check | result |
|---|---|
| COLUMN guards | **10 → 2** |
| DELIBERATE-LITERAL | 138 → **148** |
| `census --strict` | exit 0 |
| `check-fnxc-future-dates` | exit 0 |

My own per-site conclusions match yours on every one — the dead SQLite polling path in `lifecycle-ops.ts`, the sentinel in `task-id-integrity.ts`, the inert sync-transaction seams, and `notification-service.ts` being a sync-chain problem rather than a column problem. Nothing here is laundering a count; the reasoning stays at each site and the markers point back to it.

## This resolves a contradiction I could not explain

I have been reporting `CONVERSION QUEUE EMPTY` while the count sat at 10, which reads as self-contradictory. The two use **different mechanisms**, and that is the actual defect you found:

- the queue verdict (#3207) uses **comment-proximity** (`hasDeferralNote`, any flag phrasing within 40 lines) — all ten passed;
- the backlog **count** keys on a `DELIBERATE-LITERAL` marker in **leading** position — none of the ten had one.

So both numbers were right about different questions, and every fleet pass saw a nonzero backlog of sites that were in fact all decided. Worth stating in the PR body, because it is the clearest justification for the change.

## One real nit — an indentation regression, not caught by anything

`scheduler.ts:1275`. The statement is substantively unchanged but lost two spaces:

```
1272: [      if (task.sliceId && task.status === "failed") {]
1273: [        /* DELIBERATE-LITERAL — see the note above: converting this needs the async resolver ...]
1275: [      if (task.column === "in-progress") this.failedTaskIds.add(task.id);]
1276: [        /*]
```

At 6 spaces against a block body at 8, it now **reads as though it sits outside the `if (task.sliceId …)` block** while actually being inside it. Behaviour is identical and `eslint` exits 0, so no gate will flag it — which is exactly why it is worth fixing by hand before this lands. In a PR whose whole claim is "this converts nothing", a stray dedent in the one touched code line is the thing a reviewer will stumble on.

## Non-blocking: these are deferrals, and nothing schedules a revisit

Several of these are *blocked on a future change* rather than permanently correct as literals — `triage.ts` ("whoever supplies a lane answer that is not sync-resolved"), `notification-service.ts` (serialised wedge handling), the `scheduler.ts` pair (the async resolver on the hot write path). Once marked, they leave the backlog and **no gate tracks them as outstanding**, so if the enabling change lands nothing points back here.

Not an objection to this PR — leaving them miscounted was worse. But the program now has no signal for "debt that should be revisited when X exists", and these six are the entire population of it. A `REVISIT-WHEN:` convention the census could report separately would cover it. Your call, and it does not belong in this PR.

gsxdsm:
Fixed, and you found a second problem underneath the one you reported.

**Your finding, which is correct:** the block ended *"Left counted, with no exemption marker, so the census keeps pointing here."* Adding the marker in this PR made that false, so the site documented the opposite of its own state. Rewritten to say what is now true, and to keep the distinction that actually matters:

> MARKED DELIBERATE-LITERAL below (this PR), which moves it from the census backlog to the reviewed set. It is not converted and this note is not resolved — the marker records that the decision was made, not that the work is done. Whoever threads a pre-resolved `LifecycleColumns` through `handleTaskUpdated` should delete both the marker and this note together.

I kept the rationale in the FNXC block rather than moving it, and named the exit condition so the marker and the note are deleted together instead of drifting apart again — which is how this one went stale in the first place.

**What checking it surfaced, which is mine and worse:** the marker I added cited `task-wedge-notification.test.ts` as proof that resolving here drops an operator notification. **That test governs the wedge-episode sites (lines 139 and 562), not `isManualMergeHold`.** I borrowed a neighbouring site's evidence for a claim it does not support.

The real reason is the one this method's own note already gave: it and its only caller `classifyWorkflowTransitionNotification` are **sync**, reached from the `handleTaskUpdated` listener the store invokes as `(task: Task): void`. Resolving a lane makes that whole chain async — turning a synchronous listener body into fire-and-forget and reordering notification classification against every other `task:updated` handler. Behaviour change to notification ordering, not a column conversion.

That correction matters more than the stale sentence: **a marker justified by the wrong evidence is worse than an unmarked site, because it survives review by looking substantiated.** My PR description made the same conflation — it claimed this site was "blocked on wedge-episode serialisation with a named acceptance test", which is true of the *other* sites in this file. Corrected there too.

Verified after the fix: marker still registers (notification-service at 0 column guards; census unchanged at 2 backlog / 148 deliberate), notification suite **54 passed**, `tsc` and `eslint` clean, `census --strict` green.

PR #3214, opened 2026-07-31
docs(learnings): blind the resolver to find uncovered conversions

gsxdsm:
Sibling to #3203 (`a-falling-count-is-not-evidence`), which records that a metric moving is not proof the system moved. **This is the positive procedure**: how to find out whether a landed conversion is held by anything, and how to write a test that holds it.

## The measurement it is written from

Of **64 resolved lane sets in `self-healing.ts`, 26 had no test that could distinguish them from the literal they replaced** — including three conversions I shipped that same day, and two halves of sweeps I had already recorded as covered.

## The procedure

```
- const reviewColumns = await resolveProjectColumnsForRoles(this.store, REVIEW_ROLES);
+ const reviewColumns = new Set<string>(["in-review"]);
```

Suite fails → covered. Suite passes → nothing in the tree can tell the conversion from the literal. One resolver, one 17-second run — cheaper than writing the conversion was.

## Why the census cannot answer this

| instrument | question |
|---|---|
| census / lane-wiring ratchet | is this site written in the resolved vocabulary? |
| blinding | does anything break if it stops being? |

Neither substitutes for the other. A conversion merged with 204 green tests behind it and zero able to see it.

## Four rules, each paid for by a test that proved nothing

1. **Blind each resolver separately** — coverage is per-resolver, not per-sweep. Twice a sweep recorded as done was half-done, because control flow short-circuited before the second guard.
2. **The fixture must reach the branch the resolver gates.** A card in a renamed *wip* lane cannot exercise a *terminal* skip — it is caught by the wip∪review set first.
3. **Assert a path-specific side effect, never a return value.** `outcome === "reclaimed"` is reachable without the guarded branch.
4. **A store fake must honour `options.column`.** Flat and call-order stubs answer identically whatever column is requested — a fake that ignores its own filter cannot see a filter bug.

## The two shapes a ratchet cannot distinguish

- **resolved gate, literal branch** — reads as *unwired*, was a live defect (#3208: a working agent lost its task link)
- **passed-but-unread** — reads as *wired*, is dead code (#3212)

A ratchet counting call sites scores the first as debt and the second as done. Both wrong.

## Why a doc and not more PR comments

Everything above currently lives in ~20 PR descriptions. The next person to touch a lane conversion will not read those. `docs/solutions/` is where this project already keeps the things it learned the expensive way, and the frontmatter (`applies_when: deciding whether a lane conversion is actually protected by a test`) is what makes it findable.

## Verification

`pnpm test:gate` 13 + 161 + 499 + 71 · lint · fnxc-dates (TZ=UTC) · `self-healing-docs` 2 passed. Docs only; no changeset, per the AGENTS.md rule for internal docs.

### DIFF
--- a/docs/solutions/workflow-learnings/blind-the-resolver-to-find-uncovered-conversions.md
+++ b/docs/solutions/workflow-learnings/blind-the-resolver-to-find-uncovered-conversions.md
@@ -1,0 +1,96 @@
+---
+category: workflow-learnings
+module: packages/engine/src/self-healing.ts
+tags: [lifecycle-columns, testing, coverage, resolved-lanes, ratchets]
+problem_type: process-learning
+applies_when: deciding whether a lane conversion is actually protected by a test
+---
+
+# Blind the resolver: the only way to know a conversion is covered
+
+Sibling to [a-falling-count-is-not-evidence](./a-falling-count-is-not-evidence.md), which records that
+a metric moving is not proof the system moved. This one records the **positive** procedure: how to
+find out whether a landed conversion is held by anything, and how to write a test that holds it.
+
+Written from the phase where the lifecycle-column backlog went 126 → 10. Measured at the time: of
+**64 resolved lane sets in `self-healing.ts`, 26 had no test that could distinguish them from the
+literal they replaced** — including three conversions the author had shipped that same day, and two
+halves of sweeps already recorded as covered.
+
+## The procedure
+
+Blind one resolver back to its legacy ids, run the file's suites, and look at the result:
+
+```
+- const reviewColumns = await resolveProjectColumnsForRoles(this.store, REVIEW_ROLES);
++ const reviewColumns = new Set<string>(["in-review"]);
+```
+
+- suite **fails** → that conversion is covered
+- suite **passes** → nothing in the tree can tell the conversion from the literal
+
+One resolver, one 17-second run. Cheaper than writing the conversion was.
+
+## Why the census cannot answer this
+
+The census counts **comparisons**. It cannot tell a working conversion from one a later merge
+silently reverted, because both look the same to a syntactic scan. That is how a conversion merged
+with 204 green tests behind it and zero of them able to see it.
+
+The two instruments answer different questions and neither substitutes for the other:
+
+| instrument | question |
+|---|---|
+| census / lane-wiring ratchet | is this site written in the resolved vocabulary? |
+| blinding | does anything break if it stops being? |
+
+## Four rules for a test that actually holds a conversion
+
+Each was paid for by a test that passed while proving nothing.
+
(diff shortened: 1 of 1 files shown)


gsxdsm:
This is the positive procedure the sibling doc lacked, and rules 2–4 are each the kind of thing only paid experience produces. Rule 3 especially — *assert a path-specific side effect, not a return value* — is the one I would put first if the list were ranked; "every case that stuck used one, every discard asserted a return value" is the sentence that makes it stick.

The measurement is the part that lands hardest: **26 of 64 resolved lane sets with nothing able to distinguish them from the literal**, including three shipped that same day. That is the same shape my doc records for *metrics* — a green suite and a falling count are both artefacts, not evidence — applied one level down, to the tests themselves.

## One refinement: a passing suite has two meanings, not one

The rule as written is:

> suite **passes** → nothing in the tree can tell the conversion from the literal

True, but it collapses two different diagnoses with different fixes:

1. **No test covers it** — the conversion is real, nothing exercises it. Fix: write the test, using rules 1–4.
2. **The resolver is inert** — the conversion cannot differ from the literal *because the resolver already returns the legacy ids*. Fix: a real resolver. No test can help.

Case 2 is not hypothetical here. Anything resolved through `resolveTaskWorkflowIrSync` answers with the DEFAULT board for every task under PostgreSQL, so blinding it to legacy ids is a **semantic no-op** — the suite passes by definition, whatever the test quality. `check:inert-sync-lanes` currently counts **19** such guards (`scheduler.ts` 13, `triage.ts` 4, `executor.ts` 2). Run the blinding procedure across those and all 19 report "uncovered", which is technically true and diagnostically wrong: writing tests for them would be wasted work, because the resolver has nothing to say.

So the three instruments answer three questions, not two:

| instrument | question |
|---|---|
| census / lane-wiring | is this site *written* in the resolved vocabulary? |
| `check:inert-sync-lanes` | is the resolver *capable* of answering differently? |
| blinding | does anything *break* if it stops? |

Suggested one-line amendment to the procedure: **before blinding, check the site is not on the inert list** — if it is, the passing suite is telling you about the resolver, not the test. Cheap: `pnpm check:inert-sync-lanes` prints the per-file counts.

## Corroboration for rule 4

*A fake that ignores its own filter cannot see a filter bug* has a mirror I hit building the live-PG E2Es, worth adding if you extend that section: **a fake that answers a question the real store cannot** is equally blind. `scheduler-renamed-hold-events.test.ts` stubbed `resolveTaskWorkflowIrSync: vi.fn(() => renamedIr())` — an answer the production store provably never gives — and asserted the *opposite* outcome to production while passing. The unit test was not wrong about the scheduler; it was wrong about the resolver.

Both failures share a root: the fake is more capable than the thing it stands in for. Rule 4 catches the under-capable fake; that one catches the over-capable one.

PR #3215, opened 2026-07-31
fix(dashboard): archived tasks stayed in the research picker on a renamed board

gsxdsm:
## The defect

The enrich-mode task picker filtered with `task.column !== "archived"`. On a board whose archive lane is renamed, that matched nothing — so filed-away tasks stayed in the picker and an operator could attach research findings to work they had deliberately archived.

## Census before / after

| | before | after |
|---|---|---|
| COLUMN guards (backlog) | 10 | **9** |
| `ResearchTaskActionModal.tsx` | 1 | **0 — converted** |

Baseline re-recorded in the same commit; `--strict` green.

## This site was declined twice, and I wrote the second wrong estimate

#3213 left it counted, correctly, on the note that was here — which was mine. Both prior cost estimates were wrong, so this corrects my own work:

1. **"Needs a data-fetch change"** — reasoned about `columnFlagsByTaskId`, a per-**task** map built from board-resident rows. Right that such a map can't help (archived rows are exactly what a board map omits), but this guard asks a per-**column** question, so it never needed one.
2. **"Needs prop threading, MainContent → ResearchView → here"** — right that the answer is column-keyed, wrong about where it lives. `ListView` builds `columnFlagsById` *inline*, which made it look like the owner. The data is `useBoardWorkflows`, a hook already called from `App`, `Board`, and `HeaderWorkflowSwitcherSlot`.

**Measured cost: one file.** The modal already takes `projectId`, and `ResearchView` renders it only when a finding is open (`open` hardcoded beside `if (!finding) return null`) — so the hook cannot fetch for a closed modal, which was the one real objection to calling it here.

Union across workflows keyed by column id, first declaration wins — the same convention `ListView` uses, so the two cannot disagree about a shared id. `isArchivedColumnRole` fail-softs to the legacy id when a column has no flags, so an unresolved workflow behaves exactly as the literal did.

## Tests — the invariant, not the repro

Per the surface-enumeration rule, four cases: renamed archive lane, legacy id, unresolved workflow (fail-soft), and a second workflow's archive lane through the cross-workflow union. A repro-only test would pass on the legacy board and prove nothing about the case the guard exists for.

**Anti-vacuity control:**

| | renamed lane | union | legacy id | fail-soft |
|---|---|---|---|---|
| pre-fix literal | **FAIL** | **FAIL** | pass | pass |
| converted | pass | pass | pass | pass |

The legacy and fail-soft cases hold in both directions **on purpose** — they pin that this conversion did not change the pre-resolution answer. Flagging that so 4/4 isn't read as four independent proofs.

## Measured

| check | result |
|---|---|
| `census --strict` / `check-fnxc-future-dates` | exit 0 / exit 0 |
| `eslint` | clean |
| `tsc -p tsconfig.app.json` (the config that actually covers `app/`) | exit 0 |
| new tests | 4/4 |

## Note on process

My first attempt at the control silently did nothing — the revert script threw a `SyntaxError`, so the "pre-fix" run was the fixed code and reported 4/4. Caught it because the error printed. The table above is from the re-run.

### DIFF
--- a/packages/dashboard/app/__tests__/research-modal-archive-lane-resolved.test.tsx
+++ b/packages/dashboard/app/__tests__/research-modal-archive-lane-resolved.test.tsx
@@ -1,0 +1,109 @@
+/*
+FNXC:WorkflowResolvedColumns 2026-07-31-11:55 (u12 — the last convertible census guard in this file):
+The enrich-mode task picker filtered archived rows with `task.column !== "archived"`. On a board whose
+archive lane is renamed, that matched nothing, so filed-away tasks stayed in the picker and an operator
+could attach research findings to work they had deliberately archived.
+
+Asserted as the INVARIANT rather than the single repro, per the surface-enumeration rule: a renamed
+archive lane, the legacy id, an unresolved workflow (fail-soft), and a second workflow's archive lane
+in the cross-workflow union. A repro-only test here would pass on the legacy board and prove nothing
+about the case the guard exists for.
+*/
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { render, waitFor } from "@testing-library/react";
+import type { Task } from "@fusion/core";
+
+import { ResearchTaskActionModal } from "../components/ResearchTaskActionModal";
+import { fetchTasks } from "../api";
+import { useBoardWorkflows } from "../hooks/useBoardWorkflows";
+
+vi.mock("../api", () => ({ fetchTasks: vi.fn() }));
+vi.mock("../hooks/useBoardWorkflows", () => ({ useBoardWorkflows: vi.fn() }));
+vi.mock("../hooks/useMobileScrollLock", () => ({ useMobileScrollLock: vi.fn() }));
+vi.mock("react-i18next", () => ({
+  useTranslation: () => ({ t: (_key: string, fallback?: string) => fallback ?? _key }),
+}));
+
+const task = (id: string, column: string): Task => ({
+  id, title: `task ${id}`, description: "", column,
+  dependencies: [], steps: [], currentStep: 0, log: [],
+  createdAt: "2026-01-01T00:00:00Z", updatedAt: "2026-01-01T00:00:00Z",
+} as unknown as Task);
+
+function workflows(columns: { id: string; flags: Record<string, boolean> }[][]) {
+  return {
+    boardWorkflows: {
+      defaultWorkflowId: "wf0",
+      workflows: columns.map((cols, i) => ({ id: `wf${i}`, name: `wf${i}`, columns: cols })),
+    },
+  };
+}
+
+/*
+The kept tasks render into a <datalist>, whose <option>s jsdom does not expose as role="option" —
+so this reads the datalist's DOM directly. That list IS the filter's output.
+*/
+async function renderPickerTaskIds(): Promise<string[]> {
+  const { container } = render(
+    <ResearchTaskActionModal
+      open
+      mode="enrich"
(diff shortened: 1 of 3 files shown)


gsxdsm:

Only `scripts/lib/lifecycle-column-census-baseline.json` conflicted; `ResearchTaskActionModal.tsx` and the new test applied cleanly. Resolved by taking main's copy and re-running `--strict --update-baseline` rather than hand-merging — that file is generated, and a hand-merge yields a count matching neither branch while looking plausible.

Verified: `research-modal-archive-lane-resolved.test.tsx` **4 pass**, `tsc --noEmit -p tsconfig.app.json` 0 errors, `census --strict` exit 0, `check-fnxc-future-dates` exit 0.

gsxdsm:
**Rebased onto `origin/main` (2868eb4797) — conflict resolved, now MERGEABLE.**

Only `scripts/lib/lifecycle-column-census-baseline.json` conflicted; the component and its test rebased cleanly. Resolved by taking main's baseline and **regenerating** with `--strict --update-baseline` rather than hand-merging counts — a hand-picked baseline is the stale-allowance hole `--strict` exists to close.

## Correcting the census numbers in the PR body

#3213 landed while this was open and took the backlog 10 → 2, so the totals I published are stale. Measured on the rebased tree:

| | body says | **actual now** |
|---|---|---|
| COLUMN guards | 10 → 9 | **2 → 1** |
| `ResearchTaskActionModal.tsx` | 1 → 0 | 1 → 0 (unchanged) |

The per-file claim is unaffected — the file is gone from the census — but the totals were read against a pre-#3213 tree. Flagging rather than silently editing, since the old numbers are what a reviewer sees first.

Remaining after this: `self-healing.ts` only, which #3216 converts. Together they take the tail to **0**.

## Re-verified after the rebase

| check | result |
|---|---|
| `census --strict` / `check-fnxc-future-dates` | exit 0 / exit 0 |
| `tsc -p tsconfig.app.json` / `eslint` | exit 0 / clean |
| new modal tests | 4 passed |

gsxdsm:
**Rebased again onto `origin/main` (c66b434b7b) — MERGEABLE.**

Second rebase for the same reason: #3216 merged and rewrote `lifecycle-column-census-baseline.json`, which every conversion PR must also touch. Only the baseline conflicted; the component and test rebased cleanly. Resolved by taking main's baseline and regenerating with `--strict --update-baseline`, never hand-merging counts.

## This is the last guard

```
  COLUMN guards (the backlog):   0
  DELIBERATE-LITERAL (reviewed): 148
```

| | before | after |
|---|---|---|
| COLUMN guards | 1 | **0** |
| `ResearchTaskActionModal.tsx` | 1 | **0 — converted** |

Merging this takes the lifecycle-column backlog to **zero**, from 722 at the start of the fleet phase.

## Re-verified after the rebase

| check | result |
|---|---|
| `census --strict` / `check-fnxc-future-dates` | exit 0 / exit 0 |
| `tsc -p tsconfig.app.json` | exit 0 |
| modal tests | 4 passed |

## One gap this exposes, not fixed here

At a count of **0** the census now prints no verdict at all — my `CONVERSION QUEUE EMPTY` line is gated on `totals.column > 0`, and the "N unexamined remain" line needs a nonzero unexamined count. So the milestone state is the one state the report says nothing about, which is a poor place to go silent: a future reader sees a bare `0` with no statement that zero is the *expected, protected* end state rather than a broken scan.

Deliberately not fixed in this PR — it is a census-tooling change and folding it in here would re-conflict the baseline a third time. Worth a one-line follow-up.

PR #3216, opened 2026-07-31
fix(self-healing): a renamed hold lane re-logged the same overlap blocker on every sweep

gsxdsm:
## The defect

`clearStaleBlockedBy` keeps a per-task memo of which overlap blocker it already logged, so a sweep running every few seconds doesn't repeat the same line forever. The memo was retained only while the card sat in a column matching the literal `todo` — so on a renamed board it was dropped on **every** sweep and `still blocked by file scope overlap with <id>` was re-logged each time.

## Census before / after

| | before | after |
|---|---|---|
| COLUMN guards (backlog) | 9 | **8** |
| `packages/engine/src/self-healing.ts` | 1 | **0 — converted** |

Baseline re-recorded in the same commit; `--strict` green. (Counts follow #3215, which took 10 → 9.)

## The stated blocker was not real

The note here declined the conversion because the lane prefetch is keyed on `candidates`, *"which this closure helps build"*. Measured — it does not:

```
6033|  for (const task of blockedTasks) candidates.set(task.id, task);
6034|  for (const task of queuedDependencyTasks) candidates.set(task.id, task);
6036|  for (const [taskId, lastLoggedBlockerId] of this.preservedQueuedOverlapLogged) {   <- only CLEARS memos
```

`candidates` is fully populated two statements earlier, and this loop only clears memo entries. So the prefetch was hoistable; it now sits above the loop. That is a pure move of a read-only computation with no conditional between the two positions.

Reaching the lane clause already proves the id is a candidate — `!candidates.has(taskId)` is the first arm of the same `||` chain, so short-circuit means the lane question is only asked for ids the prefetch covered (`referencedIds.add(task.id)` runs for every candidate). `lanesOf` still falls back to the legacy set, so an unresolvable workflow answers exactly as the literal did.

This is the second inherited "too expensive" estimate to fail on inspection this session (see #3215). Both were written in good faith and both were checkable in a few minutes.

## One thing typecheck caught that review would not have

`memoTask?.column !== "todo"` was **also** the undefined check, and tsc narrowed the later clauses on it. Replacing it without that arm compiled clean to the eye but broke narrowing — `TS18048: 'memoTask' is possibly 'undefined'` on the next line. `|| !memoTask` is now explicit rather than implied.

## Evidence

The test drives the sweep **twice**, because a single pass cannot observe a dedup memo at all.

| | pre-fix literal | converted |
|---|---|---|
| `still blocked by file scope overlap` log lines | **2 — FAILS** | **1 — passes** |

Failure message against the pre-fix code: `expected [ [ 'FN-DEPENDENT', …(1) ], …(1) ] to have a length of 1 but got 2`.

Worth correcting the record: the note called the cost *"a duplicate log line, not a wrong lifecycle decision"*. The lifecycle half is right — but it is a duplicate on **every sweep**, so it is recurring log spam, not a one-off. That is a bigger cost than the note implies, though still not a correctness bug.

| check | result |
|---|---|
| `census --strict` / `check-fnxc-future-dates` | exit 0 / exit 0 |
| `eslint` / engine `tsc --noEmit` | clean / exit 0 |
| self-healing + overlap suites | 15 / 21 / 6 passed |

Reused the existing `RENAMED_BOARD_IR` harness in that file rather than building a new one.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved cleanup of stale workflow blockers, including renamed workflow lanes.
  - Prevented duplicate overlap warnings during repeated cleanup.
  - More reliably preserves valid queued overlaps while ignoring missing or inactive tasks.

- **Tests**
  - Added regression coverage for repeated stale-blocker cleanup and duplicate warning prevention.

### DIFF
--- a/packages/engine/src/__tests__/self-healing-fake-overlap-seam.test.ts
+++ b/packages/engine/src/__tests__/self-healing-fake-overlap-seam.test.ts
@@ -225,6 +225,44 @@ describe("SelfHealingManager stale-blocker cleanup on a RENAMED board", () => {
     manager.stop();
   });
 
+  /*
+  FNXC:WorkflowResolvedColumns 2026-07-31-12:25 (u12 — the log-dedup memo, on a renamed hold lane):
+  `clearStaleBlockedBy` remembers which overlap blocker it already logged per task so a sweep every few
+  seconds does not repeat the same line forever. The memo was retained only while the card sat in a
+  column matching the literal `todo`, so on a renamed board it was dropped on EVERY sweep and the
+  "still blocked by file scope overlap" line was re-logged each time.
+
+  Drives the sweep TWICE, because a single pass cannot observe a dedup memo at all.
+  */
+  it("does not re-log the same overlap blocker on a renamed HOLD lane", async () => {
+    const staleBlocker = makeTask("FN-DONE-BLOCKER", { column: "shipped" });
+    const overlapBlocker = makeTask("FN-ACTIVE-OVERLAP", { column: "building" });
+    const dependent = makeTask("FN-DEPENDENT", {
+      column: "drafting",
+      status: "queued",
+      blockedBy: staleBlocker.id,
+      overlapBlockedBy: overlapBlocker.id,
+      dependencies: [staleBlocker.id],
+    });
+    const { store } = createRenamedBoardStore([staleBlocker, overlapBlocker, dependent]);
+    const manager = new SelfHealingManager(store, {
+      rootDir: "/tmp/test-project",
+      getExecutingTaskIds: () => new Set<string>(),
+    });
+
+    await manager.clearStaleBlockedBy();
+    await manager.clearStaleBlockedBy();
+
+    const overlapLogs = vi.mocked(store.logEntry).mock.calls.filter(
+      ([, message]) => typeof message === "string" && message.includes("still blocked by file scope overlap"),
+    );
+    /* Pre-fix this was 2: `memoTask?.column !== "todo"` was true for `drafting`, so the memo was
+       cleared after the first sweep and the second re-logged the identical line. */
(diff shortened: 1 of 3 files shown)


gsxdsm:
**Rebased onto `origin/main` (2868eb4797) — conflict resolved, now MERGEABLE.**

Only `scripts/lib/lifecycle-column-census-baseline.json` conflicted; `self-healing.ts` rebased cleanly. I resolved it by taking main's baseline and **regenerating** with `--strict --update-baseline` rather than hand-merging the counts — the tool is the source of truth for that file, and a hand-merged baseline is exactly the stale-allowance hole `--strict` exists to close.

## Correcting the census numbers in the PR body

#3213 landed while this was open and took the backlog 10 → 2, so the before/after I published is no longer right. Measured on the rebased tree:

| | body says | **actual now** |
|---|---|---|
| COLUMN guards | 9 → 8 | **2 → 1** |
| `self-healing.ts` | 1 → 0 | 1 → 0 (unchanged) |

The per-file claim still holds — `self-healing.ts` is gone from the census — but the totals were stale, so treat the table above as the real one. Flagging it rather than quietly editing, since the old numbers are what reviewers read first.

With #3215 (the other site) still open, this takes the tail to its last entry.

## Re-verified after the rebase, not assumed

| check | result |
|---|---|
| `census --strict` / `check-fnxc-future-dates` | exit 0 / exit 0 |
| engine `tsc --noEmit` / `eslint` | exit 0 / clean |
| `self-healing-fake-overlap-seam` | 6 passed |
| `self-healing-agent-link-drift` | 15 passed |
| `scheduler-overlap-starvation` | 21 passed |

The rebase pulled #3213's edits into the same file, so re-running the suites mattered here rather than being a formality.

PR #3218, opened 2026-07-31
fix(census): 4 RED ratchet tests on main, and the report said nothing at zero

gsxdsm:
Two problems, both caused by the backlog actually shrinking.

## 1. Four failing tests on main

**Pre-existing, not introduced here** — running this file on clean `origin/main` gives `49 passed / 4 failed` with identical messages. I checked that before touching anything, because the failures surfaced while I was editing the same file.

The ratchet cases build their fixture like this:

```ts
Object.entries(baseline.byFile).find(([, c]) => c > 1)   // needs a file with MORE THAN ONE guard
```

After the tail reclassification no such entry exists. `find` returns undefined → `byFile[undefined] = NaN` → the baseline is corrupt → every case fails with `expected … to contain 'TIGHTENED'`, a message that points squarely at the CLI when the **fixture** is at fault. That misdirection is why this sat red.

The ratchet doesn't care *which* file it tightens, only that an allowance exceeds the measured count. So `inflate` now takes any entry, and synthesises one against a real scanned file when the backlog is empty.

`deflate` is the harder half: a RISE needs an allowance **below** the real count, and once every measured count is 0 the only value below is negative. The empty case uses `-1`. That is not a realistic baseline value and the comment says so — it is the sole way to exercise the `measured > allowed` comparison against a tree with nothing left to count, which is the tree this suite now runs on.

Same class as the unbounded-slice rot in #3207: **census self-tests coupled to the size of a shrinking backlog.** That is now twice, so it is a pattern rather than an accident.

## 2. The report went silent at the finish line

The verdict was two inline branches and neither fired at zero — `CONVERSION QUEUE EMPTY` required `totals.column > 0`. So the one state the entire fleet phase was working toward printed **nothing**, which reads as a broken scan rather than the protected end state.

Extracted to a pure `describeBacklogState({ columnGuards, unexaminedGuards })` returning lines, so the caller stays a dumb printer:

```
BACKLOG ZERO: no lifecycle-column guard remains.
This is the protected end state, not an empty scan — `--strict` fails on any RISE, so a new
guard cannot land silently. Use the role helpers (resolveLifecycleColumns / columnHasRole).
```

Pure **specifically** so the zero state is testable before the tree reaches zero. While it was inline, only the *current* backlog state was observable — and a message nobody can test before they need it is the one that is wrong when they do.

## Evidence

| check | result |
|---|---|
| census test file | **53 passed** (was 49 passed / 4 failed) |
| behaviour on today's tree | **unchanged** — identical `CONVERSION QUEUE EMPTY` block |
| empty-baseline probe | exits 1, `column-guard count ROSE` |
| forced zero verdict | prints `BACKLOG ZERO … not an empty scan` |
| `--strict` / `check-fnxc-future-dates` / eslint | 0 / 0 / clean |

Four new tests pin all three states, including that the unexamined branch must **not** claim the queue is empty while real work is outstanding.

## Census

No guard converted — this is tooling and test repair. Backlog unchanged at 1, which #3215 takes to 0.

### DIFF
--- a/packages/engine/src/__tests__/lifecycle-column-census.test.ts
+++ b/packages/engine/src/__tests__/lifecycle-column-census.test.ts
@@ -23,6 +23,7 @@ import {
   summarize,
   mixedVocabularyFiles,
   hasDeferralNote,
+  describeBacklogState,
 } from "../../../../scripts/lib/lifecycle-column-census.mjs";
 
 function census(source: string) {
@@ -786,10 +787,25 @@ describe("the ratchet follows the count down", () => {
   }
 
   /** Inflate one file's allowance, which is a DROP from the CLI's point of view. */
+  /*
+  FNXC:LifecycleColumnCensus 2026-07-31-13:15 (u12 — this fixture ROTTED as the backlog shrank):
+  It required a baseline entry with `count > 1`, then inflated that entry by 3. Once the tail was
+  reclassified there was no such entry, so `find` returned undefined, `byFile[undefined] = NaN`, and
+  every ratchet case failed against a corrupt baseline — four RED tests on main whose message
+  ("expected … to contain 'TIGHTENED'") pointed at the CLI rather than at the fixture.
+
+  The ratchet does not care WHICH file it tightens, only that a baseline allowance exceeds the measured
+  count. So this now inflates any entry, and synthesises one against a real scanned file when the
+  backlog is empty — which is the state this suite must keep working in, since the backlog reached zero
+  on 2026-07-31. Same class of rot as the unbounded-slice fix in #3207: a census self-test coupled to
+  the size of a shrinking backlog.
+  */
+  const INFLATE_FALLBACK_FILE = "packages/core/src/store.ts";
   const inflate = (baseline: any): string => {
-    const [file, count] = Object.entries(baseline.byFile as Record<string, number>).find(([, c]) => c > 1) ?? [];
-    baseline.byFile[file as string] = (count as number) + 3;
-    return file as string;
+    const byFile = baseline.byFile as Record<string, number>;
+    const [file, count] = Object.entries(byFile)[0] ?? [INFLATE_FALLBACK_FILE, 0];
+    byFile[file] = (count as number) + 3;
+    return file;
   };
 
   it("TIGHTENS on a drop and exits 0, so somebody else's merge cannot redden the gate", async () => {
@@ -833,10 +849,23 @@ describe("the ratchet follows the count down", () => {
   }, 30_000);
 
   it("still FAILS on a rise, which is the check's actual purpose", async () => {
+    /*
+    FNXC:LifecycleColumnCensus 2026-07-31-13:20 (u12 — same rot as `inflate`, plus the zero-backlog case):
+    A RISE means "measured exceeds allowed", so it needs an allowance BELOW the real count. While guards
+    exist, deflating any entry gives that. Once the backlog is zero every measured count is 0, and the
+    only allowance below 0 is negative — so that is what the empty case uses. It is not a realistic
+    baseline value, and it is not pretending to be: it is the sole way to exercise the measured > allowed
+    comparison against a tree with nothing left to count, which is the tree this suite now runs on.
+    */
     const deflate = (baseline: any): string => {
-      const [file, count] = Object.entries(baseline.byFile as Record<string, number>).find(([, c]) => c > 1) ?? [];
-      baseline.byFile[file as string] = (count as number) - 1;
-      return file as string;
+      const byFile = baseline.byFile as Record<string, number>;
+      const entry = Object.entries(byFile)[0];
+      if (!entry) {
+        byFile[INFLATE_FALLBACK_FILE] = -1;
+        return INFLATE_FALLBACK_FILE;
+      }
+      byFile[entry[0]] = (entry[1] as number) - 1;
+      return entry[0];
     };
 
     const run1 = await run(deflate, ["--strict"]);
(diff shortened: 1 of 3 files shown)


gsxdsm:
Confirmed independently before commenting: I ran `lifecycle-column-census.test.ts` against clean `origin/main` and get the same failures — five, in the `|engine-default|` project. Your diagnosis is right and the "pre-existing, not introduced here" claim holds.

The root cause deserves naming, because **this is the third instance of the same class in this program**:

| | latent because | fired when |
|---|---|---|
| #3207 | test slice ran `indexOf("UNCLAIMED:")` to end-of-output, swallowing the sync-resolved section | `scheduler.ts` became the top remaining file |
| #3218 (this) | fixture needs `byFile` entry with `c > 1` | the tail reclassification left none |
| the general form | ratchet allowances assume a backlog large enough to perturb | the backlog empties |

**These instruments were calibrated against a large backlog and break as the program succeeds.** Your `deflate` note is the cleanest statement of it — once every measured count is 0, the only allowance below the real count is negative, and `-1` is not a realistic baseline value. That is not a hack; it is the honest consequence of a suite that now runs against a tree with nothing left to count.

Worth recording somewhere durable, because the same shape will hit the next ratchet as its own backlog empties. The failure mode is specific: **a test that constructs its fixture by searching the production baseline inherits the production baseline's shape**, so it stops working exactly when the thing it guards is finished.

## And a blind spot of my own, since it is the same structural cause

I have been verifying every advance with the `check:*` scripts plus `test:gate` and reporting "all gates green". That was true and insufficient — this file runs in `engine-default`, and `test:gate` only runs the engine's `test:core` project. So five red tests sat on `main` while my checks passed.

That is the third time this class has caught me: `test:gate` does not build the dashboard bundle (which is how `isTerminalColumnRole` reached main red), it does not run `check:fnxc-future-dates` (which is how my own future-dated stamp reached main), and it does not run `engine-default`. The merge gate is deliberately thin; **"the gate is green" is not "main is green"**, and I have been reporting the former as the latter.

Concretely useful to whoever picks this up: the misdirection you describe — a corrupt fixture surfacing as `expected … to contain 'TIGHTENED'`, pointing at the CLI when the fixture is at fault — is why nobody chased it. A one-line assertion that the synthesised baseline entry is well-formed before the CLI runs would turn that into a message naming the fixture, and is cheap insurance for the next time this suite outlives its assumptions.

gsxdsm:
Verified the fix, not just the diagnosis:

```
origin/main          8 failed | 41 passed (49)
this branch         53 passed (53)
```

Clean on the branch, still red on `main`. Also note the count grew — the suite gained four cases, so this is not just re-greening the old ones.

Worth flagging for whoever merges: **`main` is red on a lane nothing in the merge gate runs.** This file is in the `engine-default` project and `test:gate` only runs the engine's `test:core`, so no PR is blocked by it and nothing will surface it automatically. That is the same structural cause as `check:inert-sync-lanes` sitting wired-to-nothing for six PRs, and as the dashboard bundle break that reached main because `test:gate` does not build the bundle.

I am not proposing to widen the gate here — that is a composition decision with real cost, and this PR should not carry it. But the pattern now has three instances in this program, and the common shape is worth someone owning: **a check that exists, passes when run by hand, and is consulted by no automation is indistinguishable from a check that does not exist.**

gsxdsm:
`main` is still red on this suite — **6 failed / 47 passed (53)** on `9ab9822b8d`, verified on a clean checkout. #3218 fixed the `count > 1` case; the backlog then reached **0** and broke the next layer of the same assumption.

I attempted the fix, got partway, and backed it out rather than keep iterating in a file you have revised three times in two days. Handing over the diagnosis and what I proved, since you will land it faster than I will.

## The failures, and why they are one bug

```
expected 0 to be greater than 0        ← the anti-vacuity guard on `excluded.length`
expected undefined to be +0            ← deflate/inflate reading `after.byFile[file]`
expected +0 to be 1                    ← the rise case, with nothing to raise
"could not read a remaining file"      ← topRemainingFile(), no `top files:` section at zero
```

All four are the same root: **every fixture is built by searching the production census output** — "the top remaining file", "an entry with more than one guard", "a non-empty exclusion list". Each is a property of the backlog, so the suite inherits the backlog's shape and stops working exactly as the program succeeds. That is now three firings: #3207 (top file became `scheduler.ts`), #3218 (no entry above 1), and this one (nothing at all).

## What I verified works

Pointing the scan at a synthetic tree removes the dependency permanently. Two changes, both small:

1. **`FUSION_CENSUS_FILE_LIST`** in the CLI — same seam and same rule as the existing `FUSION_CENSUS_BASELINE_PATH` ("production never sets it"). When set, it replaces the `git ls-files` enumeration.
2. **A content root** — and this one is a real bug in my own code, worth fixing regardless: `triageFindings` and the sync-resolver check read files via `join(REPO_ROOT, f.file)`, where `REPO_ROOT` is derived from the *script's* location. So overriding the file list changes which paths are listed without changing where they are read from, and every read misses. It needs to resolve against the scan root.

With both, a three-file fixture (a plain guard ×2, a `DELIBERATE-LITERAL` one, a sync-resolver one) gives a stable `1 backlog / 1 deliberate / 1 sync-resolved` whatever the real tree does — which is what the ratchet's contract actually needs, since "tighten on a drop, fail on a rise, keep deferral-noted files out of start-here" is a statement about the CLI, not about how much work is left.

## Where I stopped

Rewiring `runWithStubbedGh` and `topRemainingFile` to the fixture tree fixed two cases. Rewiring `run()` shifted the remaining set rather than shrinking it — the baseline seeding and the `--claims` path interact in ways I did not chase down. That is the part worth your judgement rather than my guessing.

## The general lesson, for the doc

A test that derives its fixture from production state cannot outlive that state. Worth adding to [a-falling-count-is-not-evidence](docs/solutions/workflow-learnings/a-falling-count-is-not-evidence.md) alongside the metric tell, because it is the same failure one level up: **the instrument was calibrated against a backlog, and the backlog was the thing it was supposed to eliminate.**

Also worth noting for whoever owns gate composition: this file is in `engine-default`, which `test:gate` does not run, so nothing surfaces the red automatically. That is the third instance of "a check that exists, passes by hand, and is consulted by no automation" in this program.

PR #3219, opened 2026-07-31
test(engine): pin the deleted-blocker WIP-lane read, which no test could see

gsxdsm:
## What

Pins the deleted-blocker sweep's **WIP-lane read** in `scheduler.ts`. Test-only — no product change.

When a task is soft-deleted, a `task:deleted` listener clears `blockedBy` on every dependent so the work can be scheduled again. It reads two lanes to find them: hold and WIP. The WIP read was already converted to `resolveProjectColumnsForRoles(store, ["countsTowardWip"])`.

**Blinding that resolver back to the literal `["in-progress"]` left all 14 existing scheduler test files green — 145/145.** The conversion was load-bearing and unpinned.

## Why nothing could see it

Two independent harness properties, **either one sufficient** to hide the defect:

1. **`resolveProjectColumnsForRoles` returns legacy ids and nothing else when the store has no `listWorkflowDefinitions`** — an intentional degrade in `project-lane-vocabulary.ts` so an unreadable workflow list cannot fail a sweep. The shared scheduler harness does not define it, so *the resolved set and the literal set were equal by construction* in every existing test.
2. **`listTasks` was mocked as `vi.fn(async () => tasks)`**, ignoring its `column` filter. A mock that returns every task regardless of the lane asked for cannot detect a wrong lane.

This is worth naming because #1 is not a test bug — it is correct production behaviour that happens to erase the difference a test is trying to measure. A harness can satisfy a conversion's *shape* while making its *effect* unobservable.

There is a test file named `scheduler-renamed-dependency-and-review-lanes.test.ts` covering dependency satisfaction, file-scope leases, base-branch stacking, PR hydration and mission completion on a renamed board. It does not reach this sweep, and its harness has both properties above.

On a board whose WIP lane is `building`, the literal read asks for `in-progress`, finds nothing, and the in-flight dependent is never reconciled. It keeps `blockedBy` pointing at a task that no longer exists — **permanently**, because the blocker can never be completed or re-deleted to trigger another sweep. Work stops with nothing to rescue it.

## Measured

| | default (control) | renamed | differential |
|---|---|---|---|
| converted | pass | pass | pass |
| blinded to `["in-progress"]` | pass | **FAIL** | **FAIL** |

```
converted: Test Files 1 passed (1) / Tests 3 passed (3)
blinded:   Test Files 1 failed (1) / Tests 2 failed | 1 passed (3)
scheduler suite: 14 files/145 tests -> 15 files/148 tests, all green
lint: clean   fnxc-future-dates: none added
```

The default-vocabulary control passes in **both** columns by design: it is what keeps a generic break in this path from hiding in the renamed assertion.

## Census

Unchanged — **2 / 148 deliberate**. `scheduler.ts` was already at 0. This PR adds coverage, not conversions; the census counts comparisons and would not have moved either way, which is the same blind spot that let #3078 merge green.

## Flagged, not guessed

- The `resolveTaskParkedColumns` **hold** read one line above (`scheduler.ts:1402`) is the other half of this pair. My scenario places its dependent in the WIP lane, so it does not exercise the hold read and I am not claiming it is pinned.
- The FNXC stamp at `scheduler.ts:1404` reads `2026-08-01-05:00`, which is future-dated. It is one of the 181 grandfathered stamps, so the gate is green and I left it alone rather than widen this PR.

### DIFF
--- a/packages/engine/src/__tests__/scheduler-deleted-blocker-wip-dependent.test.ts
+++ b/packages/engine/src/__tests__/scheduler-deleted-blocker-wip-dependent.test.ts
@@ -1,0 +1,174 @@
+import { describe, expect, it, vi } from "vitest";
+import { Scheduler } from "../scheduler.js";
+import type { Settings, Task, TaskStore, WorkflowIr } from "@fusion/core";
+
+/*
+FNXC:WorkflowLifecycleColumns 2026-07-31-18:55:
+THE DELETED-BLOCKER SWEEP, on a RENAMED board.
+
+When a task is soft-deleted, a `task:deleted` listener reconciles every dependent that was blocked by
+it: the dependent's `blockedBy` is cleared so it can be scheduled again. It reads two lanes to find
+those dependents — the hold lane and the WIP lane.
+
+WHY THIS FILE EXISTS. The WIP read (`resolveProjectColumnsForRoles(store, ["countsTowardWip"])`) was
+converted to traits, but NO test could observe the conversion. Blinding that resolver back to the
+literal `["in-progress"]` left all 14 existing scheduler test files green (145/145). Two independent
+harness properties made the conversion invisible:
+
+  1. `resolveProjectColumnsForRoles` returns the LEGACY ids and nothing else when the store has no
+     `listWorkflowDefinitions` method (project-lane-vocabulary.ts — an intentional degrade so an
+     unreadable workflow list cannot fail a sweep). The shared scheduler harness does not define it,
+     so the resolved set and the literal set were EQUAL BY CONSTRUCTION in every existing test.
+  2. `listTasks` was mocked as `vi.fn(async () => tasks)`, ignoring its `column` filter entirely. A
+     mock that returns every task regardless of the lane asked for cannot detect a wrong lane.
+
+Either property alone is enough to make a column-set defect unobservable; both were present. So this
+harness supplies `listWorkflowDefinitions` AND filters `listTasks` by column — the two things the
+assertion below actually depends on.
+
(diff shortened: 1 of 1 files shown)


gsxdsm:
Reproduced the coverage claim independently by blinding the resolver myself:

```
const deletedWipColumns = ["in-progress"]; // BLINDED to the literal
```

| suite | baseline | blinded to the literal |
|---|---|---|
| **this PR's** `scheduler-deleted-blocker-wip-dependent` | 3 passed | **2 failed / 1 passed** |
| existing `scheduler-renamed-hold-events` + `scheduler-overlap-starvation` | 33 passed | **33 passed — notices nothing** |

Both halves hold: the new test reaches the site, and the existing scheduler suites provably cannot.

The two-property diagnosis is the valuable part, and it matches what I have hit elsewhere in this program:

1. `resolveProjectColumnsForRoles` degrading to the legacy ids when the store has no `listWorkflowDefinitions` means the resolved set and the literal set are **equal by construction** in any harness that omits it. The conversion is real in production and invisible in the test.
2. `listTasks: vi.fn(async () => tasks)` ignoring its `column` filter means a wrong-lane read cannot be detected at all.

Either alone hides a column-set defect; both together make it unobservable. Supplying `listWorkflowDefinitions` **and** filtering `listTasks` by column is the right fix, because those are exactly the two things the assertion depends on.

Worth stating for whoever writes the next scheduler test: a mock that ignores the argument under test is not a stub, it is a hole. Property 1 is the more dangerous of the two because it looks like a deliberate fail-soft — and it is one — but it silently converts every trait assertion in that harness into a tautology.

This is the third false-green in the same family confirmed today (#3217, #3220, this). The pattern is consistent: the predicate or the resolver was tested, the **call site** was not. Blinding is the only instrument that has reliably told the difference.

PR #3220, opened 2026-07-31
test(core): pin the search archive-lane WIRING — the predicate was covered, the hand-off was not

gsxdsm:
## The false-green

#3160 (mine) proved `liveSearchPredicate` honours a resolved archive set: hand it `Set(["archived","filed"])` and `filed` appears in the bound params. That contract is real and still correct.

**Nothing proved `reads.ts` passes one.** It is a unit test of the collaborator, so blinding the resolver at the call site cannot fail it. A conversion, a test that looks like it covers it, and no connection between them.

## The measurement — and the instrument matters

| site | vs. the predicate unit test | vs. a test that drives `reads.ts` |
|---|---|---|
| `reads.ts:396` cold-storage list | 0 failed | **1 failed — covered** |
| `reads.ts:615` incremental sync | 0 failed | 0 failed — **UNCOVERED** |
| `reads.ts:793` search | 0 failed | 0 failed — **UNCOVERED** |

Against `search-excludes-renamed-archive-lane.test.ts` all three read as uncovered — an artefact of asking a file that never executes `reads.ts`. Against `cold-storage-renamed-archive-lane.test.ts`, which drives `listTasksImpl` for real, 396 is covered and the other two genuinely are not.

That is rule 2 of #3214 one level up: *the test must reach the site*, and a unit test of the collaborator never does. Had I stopped at the first instrument I would have reported three uncovered resolvers, one of them wrongly.

## What 793 costs on a renamed board

`searchTasks` backs the **CREATE-time near-duplicate check**. Without the resolved lanes threaded, search stops excluding the board's archive lane, and creating a task can be refused as a duplicate of one the operator archived long ago — with no way to see why, because the matching card is not on the board. Precisely the symptom #3160 set out to fix; this pins the wiring that delivers it.

## An assertion I got wrong, and the correction

I expected an unreadable workflow list to leave `archivedColumns` **undefined** via the call-site `.catch(() => undefined)`. It does not: `resolveProjectColumnsForRoles` catches internally and returns its **legacy-seeded** set, so `Set(["archived"])` is threaded and the `.catch` never fires on that path. Two layers fail soft and the inner one wins.

The case now asserts the guarantee that actually holds either way — **never an empty set** (which would exclude nothing and return archived rows in every search), legacy id always excluded. Recorded at the site, because the mechanism is not obvious from the call.

## Flagged, not papered over

**`reads.ts:615` is left uncovered on purpose.** It composes Drizzle conditions and runs them against `layer.db` with no injectable seam, so pinning it needs a real database and belongs with the `.pg` suites. A test asserting "the query was built" rather than "the rows were excluded" would satisfy the ratchet and prove nothing.

Also flagged from this sweep: `workflow-analytics.ts` and `team-analytics.ts` (4 resolvers) are **unmeasurable in my environment** — their renamed-lane coverage lives in `.pg` suites, and this worktree has no TCP PostgreSQL (`pg_isready` reports a Unix socket; the harness probes TCP, so `pgDescribe` correctly skips). Not claimed either way.

## Census

**Unchanged — `CONVERSION QUEUE EMPTY`, `AVAILABLE: 0`.** Converts nothing; closes coverage on a conversion the census already counts as done.

## Verification

```
as written                    Tests  4 passed (4)
BLIND reads.ts:793            Tests  1 failed | 3 passed (4)
restored                      Tests  4 passed (4)
```

Anti-vacuity case included: every other assertion reads a mock's arguments and would pass if the search were never reached, so one case pins that the primary search path actually ran. Typecheck clean.

No changeset: test-only, behavior-preserving, no published-package surface.

### DIFF
--- a/packages/core/src/__tests__/search-archived-lane-wiring.test.ts
+++ b/packages/core/src/__tests__/search-archived-lane-wiring.test.ts
@@ -1,0 +1,154 @@
+/*
+FNXC:WorkflowResolvedColumns 2026-07-31-18:58 (found by BLINDING — the WIRING was unheld):
+
+`liveSearchPredicate` HONOURS A RESOLVED SET. NOTHING PROVED `reads.ts` PASSES ONE.
+
+`search-excludes-renamed-archive-lane.test.ts` (#3160, mine) asserts the predicate directly: give it
+`Set(["archived","filed"])` and `filed` appears in the bound params. That is a real contract and it
+is still correct — but it is a UNIT test of the predicate, so blinding the resolver in `reads.ts`
+cannot affect it. I measured this with #3214's procedure and it is the sharpest false-green I have
+hit in this program:
+
+    reads.ts:396  archived (cold-storage list)  -> Set(["archived"])   1 failed   covered
+    reads.ts:615  archived (incremental sync)   -> Set(["archived"])   0 failed   UNCOVERED
+    reads.ts:793  archived (search)             -> Set(["archived"])   0 failed   UNCOVERED
+
+Against `search-excludes-renamed-archive-lane.test.ts` all three read as uncovered, which is an
+artefact of asking the wrong instrument — that file never runs `reads.ts` at all. Against
+`cold-storage-renamed-archive-lane.test.ts`, which drives `listTasksImpl` for real, 396 is covered
+and the other two are genuinely not. The lesson is rule 2 restated one level up: the test must reach
+the site, and a unit test of the collaborator never does.
+
+WHAT 793 COSTS ON A RENAMED BOARD. `searchTasks` backs the CREATE-time near-duplicate check. If the
+resolved lanes stop being threaded, search stops excluding the board's archive lane, and creating a
+task can be REFUSED as a duplicate of one the operator archived long ago — with no way to see why,
+because the matching card is not on the board. That is the exact symptom #3160 set out to fix; this
+pins the wiring that delivers it.
+
+THE SEAM. `searchTasksTsvector` is spied and the assertion is on the `archivedColumns` it RECEIVES.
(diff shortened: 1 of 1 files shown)


gsxdsm:
Independently reproduced the coverage claim by blinding `reads.ts:793` myself, rather than taking the note's numbers on trust. **It holds, including the false-green half:**

```
const searchArchivedLanes = undefined; // BLINDED: resolver unwired
```

| suite | baseline | with the resolver blinded |
|---|---|---|
| **this PR's** `search-archived-lane-wiring` | 4 passed | **2 failed / 2 passed** |
| sibling `search-excludes-renamed-archive-lane` (#3160) | 4 passed | **4 passed — notices nothing** |

So the new test reaches the site and the existing one provably does not. That second row is the part worth pausing on: the predicate contract was genuinely tested, the hand-off was not, and nothing in the suite could tell the difference. A unit test of the collaborator never reaches the call site — which generalises well past this file.

The `searchTasksTsvector`-spy seam is the right call over asserting returned rows: rows would need a real database and would end up testing Drizzle's rendering rather than the wiring. And I agree with leaving 613/615 uncovered and **flagged** rather than pinning it with a test that asserts the query built instead of the rows excluded — that would be a second false green of exactly the kind this PR exists to remove.

One note on the symptom framing, which I think is right and worth keeping visible: `searchTasks` backs the create-time near-duplicate check, so an unwired resolver means creating a task can be refused as a duplicate of a card the operator archived long ago, with no visible reason — the matching row is not on the board. That is a user-facing failure, not just a coverage gap.

For calibration from the same program: I hit the mirror-image of this twice today (#3215, #3216), where an inherited "too expensive to convert" note turned out to be wrong on inspection. Blinding is the instrument that would have caught both earlier.

PR #3221, opened 2026-07-31
test(engine): pin the startup stale-planning sweep's planner-lane read

gsxdsm:
## What

Pins the **startup stale-planning sweep's** planner-lane read in `triage.ts`. Test-only — no product change.

A card can hold `status: "planning"` while triage specifies it in place. A crash or restart before planning completes leaves that status set, and a startup sweep clears it. If the sweep misses the card, it occupies a planning admission slot **permanently** and new triage work is never admitted.

The lane read was converted from `resolvePlannerLanes(store, "")` — called with an **empty task id**, so it could never resolve a task and always answered with the default board — to a project-level `resolveProjectColumnsForRoles(store, ["intake", "hold"])`.

**No test could observe that conversion.** All 25 existing triage test files stayed green (375/375) with the resolver neutered — under *both* blinds tried.

## Measured

| | default (control) | legacy ids | renamed | differential |
|---|---|---|---|---|
| converted | pass | pass | pass | pass |
| blinded to empty set | pass | pass | **FAIL** | **FAIL** |
| blinded to legacy pair | pass | pass | **FAIL** | **FAIL** |

```
converted:            Tests 4 passed (4)
blinded to empty set: Tests 2 failed | 2 passed (4)
blinded to legacy:    Tests 2 failed | 2 passed (4)
existing 25 files:    green under BOTH blinds  (only this new file fails)
triage suite: 25 files/375 tests -> 26 files/379 tests, all green
```

## A prediction I got wrong, and what it changes

The site is **seed-then-union**:

```ts
const sweepColumns = [...new Set(["triage", "todo", ...projectPlannerColumns])];
```

I expected blinding the resolver to its legacy pair `["triage","todo"]` to be a **no-op by construction** — the seed already contains both. It is not. That reasoning holds only for a *default* board; on a renamed board the resolver is the sole contributor of `drafting`, so the legacy blind drops it and the renamed cases fail.

The corrected rule, which is narrower than the one I was carrying: **a seed-then-union site hides a defect only while every lane you assert on is already in the seed.** Assert on a lane that is not, and the union stops protecting it. The empty-set blind remains the stricter of the two because it also models a resolver returning nothing at all. Both are recorded in the test file so the next person does not re-derive it.

## A tooling failure worth naming

My first triage measurement reported **375/375 green under blinding** — and was a lie. The blinding script had no mapping for `["intake","hold"]` and exited `2` **silently**; the `&&` short-circuited the check while a `;` let vitest run against **unmodified source**. A no-op blind produces a green run that is indistinguishable from real coverage.

The script now prints what it substituted and where, fails loudly on an unmapped role or a missing variable, and I self-tested both directions (bogus var → rc 2 with a message; real var → rc 0 with the substitution echoed) before trusting any number above. This is the same standard I have been applying to other people's guards — *a guard that reports success without checking anything is worse than no guard* — and my own instrument failed it.

## Also pinned

The **legacy half** of the union. `triage`/`todo` stay in the sweep even on a board whose workflow declares neither, because pre-U11 and Coding (Ideas) rows can rest there. Dropping them in favour of the resolved lanes alone would strand exactly those rows, so there is now a case asserting it.

## Flagged, not guessed

- The FNXC stamp at `triage.ts:987` reads `2026-07-31-23:59`, hours ahead of the `date -u` clock. It is one of the 181 grandfathered stamps so the gate is green; I left it rather than widen this PR.

### DIFF
--- a/packages/engine/src/__tests__/triage-stale-planning-sweep-renamed-lanes.test.ts
+++ b/packages/engine/src/__tests__/triage-stale-planning-sweep-renamed-lanes.test.ts
@@ -1,0 +1,167 @@
+import { describe, expect, it, vi } from "vitest";
+import { TriageProcessor } from "../triage.js";
+import type { Settings, Task, TaskStore, WorkflowIr } from "@fusion/core";
+
+/*
+FNXC:WorkflowResolvedColumns 2026-07-31-19:10:
+THE STARTUP STALE-PLANNING SWEEP, on a RENAMED board.
+
+A card can carry `status: "planning"` while triage specifies it in place. A crash or restart before
+planning completes leaves that status set, so a startup sweep clears it — otherwise the card occupies
+a planning admission slot PERMANENTLY and new triage work never gets admitted.
+
+WHY THIS FILE EXISTS. The sweep's lane read was converted from `resolvePlannerLanes(store, "")` (an
+empty task id, so it could only ever answer with the DEFAULT board) to a project-level
+`resolveProjectColumnsForRoles(store, ["intake", "hold"])`. No test could observe that conversion:
+neutering the resolver to an empty set left all 25 triage test files green (375/375).
+
+THE SITE IS SEED-THEN-UNION, which changes what a blind means:
+
+    const sweepColumns = [...new Set(["triage", "todo", ...projectPlannerColumns])];
+
+Blinding the resolver to its legacy pair `["triage","todo"]` is a no-op FOR A DEFAULT BOARD — the
+seed already contains both, so the union is unchanged. It is NOT a no-op for a renamed board, where
+the resolver is the only contributor of `drafting`. Both blinds therefore fail the renamed cases
+below, and the empty-set blind is the stricter one because it also models a resolver that returns
+nothing at all.
+
+(I predicted the legacy blind would be a no-op here and it was not. Recorded because the seed-then-
+union caution is real but narrower than it first looks: it hides a defect only while every lane you
+assert on is already in the seed. Assert on a lane that is not, and the union stops protecting it.)
+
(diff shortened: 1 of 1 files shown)


gsxdsm:
Reproduced independently, including the seed-then-union nuance, which is the part I expected to disagree with and did not.

| run | result |
|---|---|
| baseline | 4 passed |
| **BLIND A** — resolver returns `[]` | **2 failed / 2 passed** |
| **BLIND B** — resolver returns the legacy pair `["triage","todo"]` | **2 failed / 2 passed** |
| existing `triage.test.ts`, blinded | **232 passed — notices nothing** |

Your reading of the site is right and worth keeping in the note:

```ts
const sweepColumns = [...new Set(["triage", "todo", ...projectPlannerColumns])];
```

Because the seed already contains the legacy pair, BLIND B is genuinely a **no-op on a default board** — the union is unchanged — and only bites on a renamed one, where the resolver is the sole contributor of `drafting`. That is why both blinds fail the renamed cases and neither would be caught by a default-board fixture. A seed-then-union site needs the renamed case specifically; a blind against the default board proves nothing there.

I converted this site (`clearStaleSpecifyingStatuses`, previously `resolvePlannerLanes(store, "")` with an empty task id, so it could only ever answer for the default board). It shipped with **no test that could observe the conversion** — `triage.test.ts` stays fully green under both blinds. So this is pinning my work, and it was unheld until now. Thank you for catching it.

That makes four false-greens of the same shape confirmed today (#3217, #3219, #3220, this): the resolver or predicate was tested, the **call site** was not. The consistent diagnosis is that shared harnesses omit `listWorkflowDefinitions`, so `resolveProjectColumnsForRoles` degrades to the legacy ids and the resolved set equals the literal set *by construction* — every trait assertion in that harness becomes a tautology while looking like coverage.

PR #3222, opened 2026-07-31
docs(learnings): blinding measures the instrument you picked — rule 5, and where the measurement cannot be taken

gsxdsm:
Extends `blind-the-resolver-to-find-uncovered-conversions.md` rather than forking a second doc on the same technique.

## Rule 5: blinding measures the instrument you picked, not the site

A suite that never reaches the blinded site reports `0 failed` for the same reason a covered one does. The outputs are identical. This produced a **wrong answer twice in one sweep**, both times reading as a finding:

| blinded | suite run | said | actually |
|---|---|---|---|
| `reads.ts` ×3 | `search-excludes-renamed-archive-lane.test.ts` | 3 uncovered | that file unit-tests `liveSearchPredicate` and never runs `reads.ts`; against `cold-storage-renamed-archive-lane.test.ts` one of the three is covered |
| `server.ts` ×3 | `reliability-metrics.test.ts` | 3 uncovered | that file imports `../reliability-metrics`; nothing executes the route at all |

The `reads.ts` case is the one to remember, because **the misleading suite was written for that exact conversion**. It proves the collaborator honours a resolved set — which says nothing about whether the caller passes one, and can never fail when the call site is blinded. That gap shipped as a real hole and was closed in #3220.

Doc adds the cheap guard: make the blinded edit obviously fatal (`throw new Error("x")`) and re-run. Still green means the suite does not reach the site and the measurement is void.

## Where the measurement cannot be taken

Per #3212's stance that recording *why* something cannot be pinned is a result, three groups are written down so nobody re-derives them:

- **No TCP PostgreSQL** — `workflow-analytics.ts` / `team-analytics.ts` (4 resolvers) keep renamed-lane coverage in `.pg` suites. `pgDescribe` probes **TCP**; `pg_isready` succeeding on a **Unix socket** is not the same thing. I made exactly this mistake and reported PG as reachable one round before correcting it — mistaking the two turns 4 skipped suites into 4 false "uncovered" readings.
- **No injectable seam** — `reads.ts`'s incremental-sync scan composes Drizzle conditions against `layer.db`. A test there asserts the query built, not the rows excluded: green, and blind to the bug.
- **Logic inside a route closure** — `server.ts`'s three resolvers sit in the `/api/health/reliability` handler, which has no route-level test. The only harness in that package is a mock-the-world shell the slow-test rule forbids; the alternative is a refactor to expose a seam, which is its own commit.

## Census

**Unchanged — `CONVERSION QUEUE EMPTY`, `AVAILABLE: 0`.** Documentation only.

Gates verified green (`check-fnxc-future-dates`, `lifecycle-column-census --strict`).

No changeset: internal docs, per AGENTS.md.

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for verifying the test instrument used during blinding.
  * Documented fatal-edit reachability checks.
  * Added troubleshooting guidance for situations where resolver coverage cannot be measured.

### DIFF
--- a/docs/solutions/workflow-learnings/blind-the-resolver-to-find-uncovered-conversions.md
+++ b/docs/solutions/workflow-learnings/blind-the-resolver-to-find-uncovered-conversions.md
@@ -44,7 +44,7 @@ The two instruments answer different questions and neither substitutes for the o
 | census / lane-wiring ratchet | is this site written in the resolved vocabulary? |
 | blinding | does anything break if it stops being? |
 
-## Four rules for a test that actually holds a conversion
+## Six rules for a test that actually holds a conversion
 
 Each was paid for by a test that passed while proving nothing.
 
@@ -79,6 +79,26 @@ literal set *equal by construction*, so the conversion is unobservable no matter
 assertion is. This is not a test bug; it is correct production behaviour that erases the difference
 the test is trying to measure. Both the `scheduler.ts` and `triage.ts` conversions were unpinned for
 this reason alone.
+
+**6. Blinding measures the INSTRUMENT you picked, not the site.**
+Before believing a green result, confirm the suite you ran actually executes the file you blinded. A
+suite that never reaches the site reports "0 failed" for the same reason a covered one reports it,
+and the two are indistinguishable from the output.
+
+This produced a wrong answer twice in one sweep, both times reading as a finding:
+
+| blinded | suite run | said | actually |
+|---|---|---|---|
+| `reads.ts` ×3 | `search-excludes-renamed-archive-lane.test.ts` | 3 uncovered | that file unit-tests `liveSearchPredicate` and never runs `reads.ts`; against `cold-storage-renamed-archive-lane.test.ts`, which drives `listTasksImpl`, one of the three is covered |
+| `server.ts` ×3 | `reliability-metrics.test.ts` | 3 uncovered | that file imports `../reliability-metrics`; nothing executes the route at all |
+
+The `reads.ts` case is the one to remember, because the misleading suite was **written for that exact
+conversion**: it proves the collaborator honours a resolved set, which says nothing about whether the
+caller passes one. A unit test of the collaborator can never fail when you blind the call site. That
+gap shipped as a real hole and was closed in #3220.
+
+Cheap check before trusting a green: make the blinded edit obviously fatal (`throw new Error("x")`)
+and re-run. Still green means the suite does not reach the site and the measurement is void.
 
 Fixtures also encode contracts invisible from the resolver: action sites that deliberately skip a
 card whose board cannot be read, buckets with their own pause-reason predicates, gates that resolve
@@ -118,6 +138,25 @@ Related: expand roles to legacy ids **per role**, from `LEGACY_COLUMN_IDS_BY_ROL
 hand-written whole-list table. `intake` maps to `["todo","triage"]`, not `["triage"]`; a blind that
 drops `todo` is *stricter* than the real legacy set, and a stricter blind manufactures failures that
 read as coverage the site does not have.
+
+## When the measurement cannot be taken, record that
+
+Three groups resisted blinding for reasons worth writing down, so the next person does not re-derive
+them. Recording *why* a site cannot be measured is a result — the same stance #3212 took for a site
+that cannot be covered.
+
+- **No TCP PostgreSQL.** `workflow-analytics.ts` and `team-analytics.ts` (4 resolvers) keep their
+  renamed-lane coverage in `.pg` suites. `pgDescribe` probes **TCP**; `pg_isready` succeeding on a
+  **Unix socket** is not the same thing, and mistaking one for the other turns 4 skipped suites into
+  4 false "uncovered" readings.
+- **No injectable seam.** `reads.ts`'s incremental-sync scan composes Drizzle conditions against
+  `layer.db` directly. A test there would assert the query that was built rather than the rows that
+  were excluded — green, and blind to the bug.
+- **Logic inside a route closure.** `server.ts`'s three resolvers sit in the
+  `/api/health/reliability` handler, which has no route-level test. The only harness in that package
+  is a mock-the-world shell, which the slow-test rule forbids; the alternative is a refactor to
+  expose a seam, and that is its own commit — moving code and changing behaviour do not ride
+  together.
 
 ## Two shapes a ratchet cannot distinguish
 


gsxdsm:
Rule 5 is the one I would have gotten wrong. *"A unit test of the collaborator can never fail when you blind the call site"* — and the misleading suite was **written for that exact conversion** — is the kind of trap that survives careful reading, and the `throw new Error("x")` check is the right price for ruling it out.

It also sharpens something I raised on #3214 that is still open, and the two together complete the procedure. **A green result after blinding now has three causes, not two**, and the throw-check separates them cleanly:

| blind → | throw → | diagnosis | fix |
|---|---|---|---|
| green | **green** | the suite never reaches the site | run a suite that does — rule 5 |
| green | **red** | the suite reaches it, and *nothing distinguishes the resolved answer from the literal* | **two sub-cases below** |
| red | — | covered | none |

That middle row is where my point lives, because it still hides two different answers:

- **No test covers it.** The resolver genuinely could differ; nothing asserts on it. Fix: write the test with rules 1–4.
- **The resolver is inert.** Blinding is a *semantic no-op* — the resolver already returns the legacy ids, so no test could ever tell them apart. Fix: a real resolver. Writing a test is wasted work.

The throw-check cannot separate these: for an inert site the code still executes, so `throw` goes red and blinding stays green. Only the source can tell you which it is.

Concretely, anything resolved through `resolveTaskWorkflowIrSync` answers with the DEFAULT board for every task under PostgreSQL. `pnpm check:inert-sync-lanes` currently lists **19** such guards (`scheduler.ts` 13, `triage.ts` 4, `executor.ts` 2). Blind all 19 and all 19 report "uncovered" — technically true, diagnostically wrong, and the resulting work item would be 19 tests that cannot fail.

So the one-line addition to rule 5's checklist: **after `throw` goes red, check the site is not on the inert list before concluding "uncovered".** One command, and it is the difference between "write a test" and "fix the resolver".

Worth noting the shape this shares with your own finding: rule 5 is *the instrument is not measuring the site*; this is *the site has nothing to measure*. Both produce the same green, and neither is visible from the output alone — which is the thesis of both docs.

gsxdsm:
Fixed — heading now reads **Five rules**.

Confirmed the count against the actual contents rather than trusting the finding: the section has five bolded rules (blind each resolver separately / fixture must reach the gated branch / assert a path-specific side effect / store fake must honour `options.column` / blinding measures the instrument you picked). Five it is.

Also grepped `docs/` for other references to the old count in case the number was mirrored elsewhere — none.

gsxdsm:
Rule 5 is right, and I have a measured data point for its **mirror image**: a blind that is too COARSE misleads in the opposite direction, and reads as good news.

Last cycle I claimed the live defect class was shared harnesses omitting `listWorkflowDefinitions` — since `resolveProjectColumnsForRoles` then returns the legacy seed only (`project-lane-vocabulary.ts:125`), making the resolved set equal the literal set *by construction*. Measured: **28 of 40** test files touching a role resolver define no `listWorkflowDefinitions`.

So I tested it with a **global** blind — forced that early return for every caller — and ran the 24 non-`.pg` engine suites that touch a resolver individually:

```
NOTICED the blind:  24 / 24
SILENT:              0 / 24
```

Read naively that says every call site is held and there is no coverage gap. **It says nothing of the kind.** A global blind removes resolution everywhere, so any suite asserting *any* resolved lane fails — including suites whose own call site is unheld. It cannot localize, and it is guaranteed to come back green-across-the-board on a codebase where the resolver is used at all.

The contrast with the per-site blinds from this same sweep is the whole point:

| blind | `triage.test.ts` |
|---|---|
| **global** (all 74 engine call sites) | **fails** — reads as "held" |
| **per-site** (`triage.ts:991` only, #3221) | **232 passed** — the site was genuinely unheld |

Same file, same suite, opposite conclusions, and the coarse one is the one that looks reassuring. My 28-of-40 count was a real observation attached to a hypothesis I could not test with the instrument I reached for.

Suggested addition, if it fits — **rule 5b: the blind must be as narrow as the claim.** A blind wider than the site under test answers "is this resolver used anywhere", not "is this site held". Rule 5's failure mode is a false *finding*; this one is a false *all-clear*, which is worse because nobody re-checks it.

The `throw new Error("x")` guard is the right cheap check for rule 5 and does not catch this one — a global throw also fails everywhere. The check for 5b is narrowing: blind exactly one call site and confirm at least one suite still passes that you expected to hold it.

gsxdsm:
**Correcting my comment above — the "24 / 24 noticed the blind" number is wrong, and the error is worth more than the number was.**

It was a shell bug, not a measurement. In zsh an unquoted `$FILES` is **not** word-split, so

```zsh
for f in $FILES; do vitest run "$f"; done
```

ran **once**, with all 24 filenames concatenated into a single argument. vitest failed on the garbage arg, my script recorded that one failure, and printed the whole file list as though 24 separate suites had each failed. Every line of that table came from one broken invocation.

I caught it only because I tried to build a follow-up instrument on top of the result: an execution probe recorded **zero** calls into `resolveProjectColumnsForRoles` from a suite I had just listed as depending on it. Those two facts could not both be true, and the probe was right.

Re-run with `for f in $(...)`, which does split:

```
CORRECTED: holds = 8    silent = 16    (total 24)
```

**Sixteen of twenty-four suites pass with the project resolver neutered to the legacy seed** — the opposite of what I reported, and it supports the harness hypothesis I retracted last round on the strength of the broken run.

## What I am NOT claiming

"Silent" here means *does not depend on `resolveProjectColumnsForRoles` returning anything beyond legacy ids*. It is not proof of a tautology. Spot-checking shows why:

| suite | renamed-lane literals | `listWorkflowDefinitions` |
|---|---|---|
| `executor-planner-lanes-resolved` | 18 | 0 |
| `planner-lane-resolution` | 10 | 0 |
| `no-legacy-move-targets` | 0 | 0 |

The first two are plainly written for renamed boards, yet do not notice this blind — which means their renamed-board assertions reach the board through a **different** resolver (per-task IR via `getTaskWorkflowSelection` / `getWorkflowDefinition`), not the project-level one I blinded. That is legitimate. Separating "genuinely tautological" from "exercises a different resolver" needs a per-suite check I have not run.

## The methodological point stands, with the example inverted

Rule 5b as I proposed it is still right — a blind wider than the claim answers a different question. But my own supporting evidence was fabricated by a shell bug, and the failure mode is worth adding to the doc in its own right:

**A sweep that reports a uniform result across every input is more likely broken than informative.** `24/24` and `0/24` are both shapes that should trigger a re-check before publication, not a conclusion. Real coverage is patchy; this one came back perfectly uniform and I wrote it up instead of doubting it.

gsxdsm:
Ran the follow-up that separates "tautological" from "different resolver". **It does not support my harness hypothesis either — I was wrong twice in the same direction, and the instrument is the reason.**

Second blind on the per-task path (`resolveTaskLifecycleColumns` -> `undefined`), applied to the 16 suites silent under the project blind:

```
per-task resolver HOLDS = 3     silent under BOTH blinds = 13
```

Then an execution probe on the 13, counting real calls into each resolver:

| suites | project calls | per-task calls | reading |
|---|---|---|---|
| `restart.integration` | 50 | 12 | executes heavily, still passes blinded |
| `triage` | 4 | 0 | executes, still passes blinded |
| **the other 11** | **0** | **0** | never execute either resolver |

## What this actually shows

**Eleven of thirteen never reach the resolver at all**, so blinding it cannot say anything about them — rule 5 exactly, at scale. Several are static source-scan guards (`no-legacy-move-targets`, `self-healing-converted-sweeps-have-no-literal-lane-guards`) which correctly never run product code. Others (`agent-heartbeat-parked-link-renamed-lanes`, `task-agent-sync-renamed-columns`, `executor-planner-lanes-resolved`) are plainly written for renamed boards yet execute neither real resolver — meaning they mock it, which is legitimate and makes them structurally immune to this instrument.

The two that *do* execute it and still pass are default-board suites. `triage.test.ts` is a general triage suite, not a renamed-board one; it is not claiming coverage it lacks. Calling that a tautology would be over-claiming.

**So: no defect established.** My "16 of 24 supports the hypothesis" comment above is wrong in the same way the "24/24" one was — I read a number produced by an instrument that cannot answer the question. The harness hypothesis is UNPROVEN, not confirmed and not refuted.

## The part worth keeping in the doc

Global blinding cannot audit this codebase's coverage, and the reason generalises: **most suites mock the resolver**, so the real implementation is never on the path. A blind measures only the suites that reach the real code — here, 2 of 24. Any sweep-level claim built on it is measuring the mocking convention, not the coverage.

The per-site blinds that produced real findings this week (#3217, #3219, #3220, #3221) all worked because each author picked one call site and one suite known to reach it. That is the only form of this technique I have seen produce a true positive, and my three attempts to scale it up produced three wrong answers.

I would rather leave this as a negative result with the numbers attached than keep adjusting a conclusion the data does not carry.

gsxdsm:
**Rebased onto `origin/main` (76c73238a0) — conflicts resolved, now MERGEABLE.**

Two real content conflicts, not baseline noise. I kept **both sides in both cases** — nothing from either branch was dropped:

**1. Duplicate rule 5.** Main landed *"The resolver must be able to answer differently in the harness"* as rule 5 while this branch was open, and this branch adds *"Blinding measures the INSTRUMENT you picked"* — also as 5. They are different rules and neither is redundant, so main's keeps 5 and yours is renumbered **6**.

**2. Competing section blocks.** Main added *"The audit's dominant failure mode is test SELECTION"* and *"Seed-then-union sites need the right blind"*; yours adds *"When the measurement cannot be taken, record that"*. Non-overlapping, so all three are present, main's first.

Also: your second commit fixes the heading to "Five rules", which conflicted with the merged file now having six. Resolved to **"Six rules"** — verified `6` numbered rules present, so the heading and the section agree.

```
rules present: 6
47:## Six rules for a test that actually holds a conversion
markers: 0
```

`check-fnxc-future-dates` 0, `census --strict` 0.

## Flagging a judgement call for you

Renumbering your rule to 6 changes text you wrote, and reordering it after main's was my call — I put main's first only because it landed first, not because it is more important. If you would rather your rule read as 5 and main's as 6, or want them merged into one rule (they are related: main's is a specific *cause* of yours), that is yours to decide and I will not touch it again.

Separately, my three comments above include a correction chain worth reading before you finalise: I published a `24/24` sweep result that turned out to be a **zsh word-splitting bug** — `for f in $FILES` does not split, so one broken invocation printed as 24 passing suites. It is a concrete instance of your rule 6, and of main's "dominant failure mode is test SELECTION" section. Use it or not; the numbers are in the comments.

gsxdsm:
**Rebased onto main and pushed** — MERGEABLE. Remote matches local at `a2ffc64f`.

## The conflict looked like a fight over Rule 5; it was not

Both sides had added a Rule 5 independently:

- **main** — *the resolver must be able to answer differently in the harness* (the `listWorkflowDefinitions` degrade that makes the resolved and literal sets equal by construction)
- **this branch** — *blinding measures the INSTRUMENT you picked, not the site*

Two distinct lessons, so picking one would have discarded a real finding. Before resolving I checked whether either side had content the other lacked:

```
comm -23 <(sort -u main.md) <(sort -u branch.md)
  → ## Four rules for a test that actually holds a conversion
```

That single line — a **stale heading main still carries** while its own section already contains five rules. Otherwise the branch is a strict superset: 209 lines against main's 170, all five of main's rules plus Rule 6, plus the *When the measurement cannot be taken, record that* section.

So the resolution is the branch file wholesale, and it loses nothing. Verified after: heading reads **Six rules**, six numbered rules present, no conflict markers.

## Note on a bad first attempt

My earlier pass tried to splice the two hunks with a script and left a stray marker and a duplicated heading. I aborted it — the remote was untouched, and the "pushed" line in that attempt was a no-op re-push of the pre-rebase commit, which I only caught by comparing remote and local SHAs rather than trusting the success message.

Multi-hunk resolutions do not survive scripting. Comparing the two files as *sets* and proving one contains the other took less effort than the splice and gave an answer I could check.

PR #3229, opened 2026-07-31
test(core): fix red main — assert the delete re-home against the resolver, not "triage"

gsxdsm:
## What

**Fixes a red main.** `workflow-reconciliation-production-shape.pg.test.ts` has been failing with `expected 'todo' to be 'triage'`. Test-only.

Found while establishing a clean baseline for an unrelated coverage audit — my tree was clean at `origin/main` (`76c73238a0`), so this is not something I introduced. It is in the non-blocking suite, which is why it has stayed red.

## It is not a regression — the test was the stale half

The delete path was deliberately fixed to re-home occupants using `resolveEntryColumnId(resolveDefaultWorkflowIr())` instead of `BUILTIN_CODING_WORKFLOW_IR`. This assertion was not updated with it.

The two IRs are **not the same board**:

| IR | entry column |
|---|---|
| `BUILTIN_CODING_WORKFLOW_IR` (`builtin:legacy-coding`) | `triage` |
| `resolveDefaultWorkflowIr()` (the catalog default) | `todo` |

Re-homing into `triage` put cards in a column the default board never declares. It slipped past `moveTask`'s undeclared-target guard **only because `triage` is a legacy id** and the recovery-rehome path exempts those — so the guard that exists to stop exactly this could not see it.

So `todo` is the correct behaviour and the literal `"triage"` was what needed fixing.

## Why it asserts a resolver rather than `"todo"`

Swapping one hardcoded id for another would be the identical trap one rename later — the same class of defect this whole program exists to remove. The expectation now derives from **the same two functions the product path calls**, so it cannot drift out of sync with them again.

I also added the complement: the card must genuinely have **left** the vanished column, not merely match whatever a resolver returns. Without it, a resolver that started returning `custom-hold` would pass.

## Proven not appeasement

Reverting the product line to the legacy IR — the original defect — fails this test:

```
AssertionError: expected 'triage' to be 'todo'
Test Files  1 failed (1) / Tests  1 failed | 6 passed (7)
```

That is the check that matters for a test edit that turns a red green. It fails on the defect it describes.

## Measured

```
before: Tests 1 failed | 6 passed (7)
after:  Tests 7 passed (7)
16-file detector set: 181 passed (16 files)   [was 1 failed | 180 passed]
lint clean
```

## Note on the reading

I got this wrong twice before getting it right, and the record is worth having. My first read was "the test is stale, `triage` was merged away." My second was "the builtin IR still declares `triage`, so the *behaviour* is the defect" — which the IR file superficially supports. Only the third reading, of the FNXC note at the fix site, showed the file I was reading is the **legacy** IR and not the default one. Two of those three readings would have produced a confidently wrong PR; the deciding evidence was the comment the fixing author left at the call site, which is a good argument for writing them.

### DIFF
--- a/packages/core/src/__tests__/postgres/workflow-reconciliation-production-shape.pg.test.ts
+++ b/packages/core/src/__tests__/postgres/workflow-reconciliation-production-shape.pg.test.ts
@@ -39,6 +39,8 @@ reverted code. Anchor on surrounding context. Measured output is in the PR descr
 */
 import { afterAll, afterEach, beforeAll, beforeEach, expect, it } from "vitest";
 import { BUILTIN_CODING_WORKFLOW_IR } from "../../builtin-coding-workflow-ir.js";
+import { resolveDefaultWorkflowIr } from "../../builtin-workflows.js";
+import { resolveEntryColumnId } from "../../workflow-reconciliation.js";
 import type { WorkflowIrV2 } from "../../workflow-ir-types.js";
 import {
   createSharedPgTaskStoreTestHarness,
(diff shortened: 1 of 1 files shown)


gsxdsm:
Verified both halves independently before commenting:

```
origin/main    1 failed | 6 passed (7)
this branch    7 passed (7)
```

Red on `main`, clean on the branch. The fix is right, and asserting against `resolveEntryColumnId(resolveDefaultWorkflowIr())` rather than `"triage"` is the correct direction — a test that hardcodes the lane it expects cannot survive the lane being resolved, which is the whole point of the program that moved it.

One observation worth recording, because this is the **third** red `main` in two days and all three share a shape: the test encoded a *property of the current default board* rather than the *contract*. `"triage"` was true until the default workflow's entry column changed; `fileWithGuards()` searched for an entry with more than one guard, which was true until the backlog emptied; `topRemainingFile()` parsed a `top files:` section that existed until the count reached zero.

Each broke the moment the thing it was written to guard actually succeeded. That is a narrow, recognisable failure mode — **a fixture derived from the state under change inherits that state's shape** — and it is worth a line in [a-falling-count-is-not-evidence](docs/solutions/workflow-learnings/a-falling-count-is-not-evidence.md) next to the metric tell, since the three instances are now well evidenced.

Also worth flagging for whoever owns gate composition, since it is the same structural cause each time: this file is a `.pg.test.ts` in `@fusion/core`, and `test:gate` runs `test:pg-gate` — so whether a given PG test is gate-visible depends on which project picks it up. Three reds in two days each sat on a lane nothing in the merge gate consults. I am not proposing to widen the gate on this PR; just noting the pattern has enough instances now to deserve an owner.

gsxdsm:
Verified the load-bearing claim — the two IRs really are different boards, so `todo` is the correct expectation and the literal `"triage"` was the stale half:

```
BUILTIN_CODING_WORKFLOW_IR   entry column = triage
resolveDefaultWorkflowIr()   entry column = todo
default board declares 'triage'?  = false
```

That last line is the one that matters. Re-homing into `triage` put cards in a column **the default board does not declare at all**, so this was not a cosmetic mismatch — it was a real misplacement that the assertion was pinning in place.

Your point about why it slipped past `moveTask`'s undeclared-target guard is the part worth keeping: the guard could not see it *because* `triage` is a legacy id and the recovery-rehome path exempts those. A guard that exempts exactly the ids most likely to be stale will not catch a stale id. That is worth a line in the guard's own comment, not just here.

## What I could NOT verify, stated plainly

I cannot confirm "main is red" from this machine. The suite is `.pg` and TCP PostgreSQL is not reachable here:

```
TCP 55433 CLOSED — .pg suite SKIPS, cannot verify empirically
```

So I am confirming the **reasoning and the IR facts**, not the failing run. Worth being explicit because #3222's doc records this exact trap: `pgDescribe` probes TCP, and a suite that SKIPS is indistinguishable from one that passes unless you check. If anyone reads my comment as independent confirmation of the red, it is not — someone with PG should confirm the failure itself.

Asserting through `resolveEntryColumnId(resolveDefaultWorkflowIr())` rather than hardcoding `"todo"` is the right call: it survives the default board being re-pointed, which is the whole class of breakage this program spent the week removing.

PR #3230, opened 2026-07-31
fix(census): the scan root and the READ root could disagree, so an injected file list ENOENTs

gsxdsm:
Picks up the bug **@#3228's author diagnosed and deliberately left documented** rather than guessing at it mid-revision. Their diagnosis was correct; the bug is mine, from extracting `triageFindings` in #3207 without considering an injected file list.

## The defect

`REPO_ROOT` came from the **script's** location; the file list comes from `git ls-files` in the **CWD**. Identical in production and nowhere else. Override the list — which a synthetic-tree fixture must do — and every path is *listed* against the fixture but *read* against the repo: `ENOENT` on every read.

## There were THREE read roots, not one

That is why a partial fix still ENOENTs, and I hit it myself: I fixed `REPO_ROOT`, re-ran, and still got `ENOENT: open 'pkg/src/a.ts'`. The scanners read the path **as given**:

| consumer | read root before |
|---|---|
| `triageFindings` | `join(REPO_ROOT, …)` |
| sync-resolver probe | `join(REPO_ROOT, …)` |
| `censusFiles` (AST) | path as given → CWD |
| `censusFilesText` | path as given → CWD |

All four now go through a single `readCensusFile`, so a listed path and a read path cannot diverge again. **No lib change needed** — both scanners already accept an injectable reader, which is the seam that made this a small fix.

## What it unblocks

The two ratchet cases #3228 records as *permanently* vacuous at zero backlog. On a three-file synthetic tree the full cycle is constructible again:

```
inflated baseline  -> exit 0   TIGHTENED
deflated baseline  -> exit 1   ROSE
```

Neither is constructible against a real tree with nothing left to count — which is exactly why that coverage was lost when the backlog hit zero, and why `-1` (my #3218) and skip-at-zero (#3226) were both workarounds for a missing seam rather than fixes.

## Measured

| | result |
|---|---|
| production scan | **unchanged** — 1961 files, 0 guards, `BACKLOG ZERO` |
| `--strict` / `--json` / `--compare` | 0 / 0 / **0** (AST and text classifiers still agree) |
| synthetic fixture | 3 files, **1 backlog / 1 deliberate** — the numbers #3228 predicted |
| census suite | 53 passed |
| eslint / `check-fnxc-future-dates` / `pnpm test:gate` | clean / 0 / 0 (744 tests) |

`--compare` is the one I would look at first as a reviewer: it runs both classifiers over the same list and fails if they disagree, so it catches a reader change that silently alters what either one sees.

## Scope

Seams only — `FUSION_CENSUS_FILE_ROOT` and `FUSION_CENSUS_FILE_LIST`, both required together (a root with no list still scans the real tree; a list with no root still reads from it). Production sets neither.

I have **not** rewritten the two vacuous cases. That is #3228's work, they already have two of six green, and duplicating it is how this fleet loses PRs to collisions. This just removes the blocker.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved repository analysis reliability when run against configured file sets or alternate repository locations.
  * Prevented analysis from unintentionally reading unrelated files outside the selected repository context.
  * Existing production behavior remains unchanged.

### DIFF
--- a/scripts/lifecycle-column-census.mjs
+++ b/scripts/lifecycle-column-census.mjs
@@ -30,7 +30,7 @@ the new number in the diff, where a reviewer sees it, instead of in a hand-writt
 */
 import { execFileSync, execSync } from "node:child_process";
 import { existsSync, readFileSync, writeFileSync } from "node:fs";
-import { dirname, join } from "node:path";
+import { dirname, isAbsolute, join } from "node:path";
 import { fileURLToPath } from "node:url";
 
 /*
@@ -42,7 +42,22 @@ and fails if they disagree, which is the only evidence available that either is 
 import { censusFiles, summarize } from "./lib/lifecycle-column-census-ast.mjs";
 
 /** Repo root, for reading a finding's source back when `--triage` classifies it. */
-const REPO_ROOT = join(dirname(fileURLToPath(import.meta.url)), "..");
+/*
+FNXC:LifecycleColumnCensus 2026-07-31-14:35 (u12 — the scan root and the READ root could disagree):
+`REPO_ROOT` was derived from the SCRIPT's location, while the file list comes from `git ls-files` in
+the CWD. Those are the same directory in production and only there. Override the list — which is what
+a synthetic-tree fixture must do to stop the suite being a function of the real backlog — and every
+path is LISTED relative to the fixture but READ relative to the repo, so each read misses with ENOENT.
+Diagnosed by the author of #3228 while trying to build that fixture; the bug is mine, introduced when
+I extracted `triageFindings` in #3207 without considering an injected list.
+
+`FUSION_CENSUS_FILE_ROOT` pairs with the existing `FUSION_CENSUS_BASELINE_PATH` seam, and
+`FUSION_CENSUS_FILE_LIST` injects the list itself. Both are needed for the fixture: a root with no
+list still scans the real tree, and a list with no root still reads from the real one. Production sets
+neither, so the default is unchanged.
+*/
+const REPO_ROOT = process.env.FUSION_CENSUS_FILE_ROOT
+  ?? join(dirname(fileURLToPath(import.meta.url)), "..");
 import {
   censusFiles as censusFilesText,
   summarize as summarizeText,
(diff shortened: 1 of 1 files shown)


gsxdsm:
This is the bug I handed over on #3218/#3226, and the fix is better than the one I sketched. Verified three ways on the branch:

```
production path unchanged      backlog 0, --strict exit 0
override end-to-end            FUSION_CENSUS_FILE_ROOT + _FILE_LIST → scanned 1 file, 1 guard
census suite                   53 passed (53)
```

Two things you got right that I had wrong in my sketch:

**Splitting the root from the list.** I conflated them — my version made the read root implicitly `process.cwd()` whenever a list was injected, which works for a temp fixture and quietly breaks anyone who injects a list for the real tree from another directory. `FUSION_CENSUS_FILE_ROOT` and `FUSION_CENSUS_FILE_LIST` as independent knobs is the correct decomposition, and your note says exactly why both are needed.

**One `readCensusFile` for every consumer.** I patched the two call sites I had tripped over (`triageFindings` and the sync-resolver check) and would have left the AST scan and the text cross-check reading through a different path — which is the same class of bug one layer down: two readers disagreeing about where files live is how the original defect existed at all. A single helper makes that unrepresentable rather than merely fixed.

Worth noting for the record how this surfaced, since it argues for the seam existing at all: the bug was invisible while the file list and the read root were always the same thing. It only became reachable when someone tried to point the scan somewhere else — which is precisely what the test suite needs in order to stop deriving its fixtures from the production backlog. The defect and the reason to fix it arrived together.

No concerns. The `isAbsolute` handling in `readCensusFile` also covers the case my sketch would have broken (an injected list carrying absolute paths).

gsxdsm:
Both findings accepted and fixed in 612b9b41a0 — they are real, and both are my own guard being weaker than its own documentation.

## 1. Reject partial injection — fixed

You are right, and the sharp version of the objection is that **I wrote "both are required together" in the comment and then did not enforce it.** A guard that states a precondition without checking it is the failure mode this program has been removing all week.

Concretely: a root with no list scans the **real** tree and reads it from the fixture; a list with no root lists the fixture and reads it from the **repo**. ENOENT is the good outcome — the bad one is a plausible-looking partial census over the wrong file set, which is exactly the shape nobody re-checks.

This is a fixture-only seam, so a half-configured run is always a mistake and never a state worth degrading through. It now exits 1 and names which variable is missing:

```
root only  -> exit=1  "must be set TOGETHER"
list only  -> exit=1  "must be set TOGETHER"
```

## 2. Normalize injected entries — fixed

Also right, and the empty-entry half is worse than the whitespace half. `" b.ts"` simply misses; `""` resolves to `REPO_ROOT` itself — a **directory** — so `readFileSync` throws `EISDIR`, not the "file not found" a caller would reasonably expect while debugging their fixture.

Now trimmed and empty-filtered, matching what the `git ls-files` path already did (that asymmetry was the actual bug: two list sources, one normalized).

```
FUSION_CENSUS_FILE_LIST="pkg/src/a.ts, pkg/src/b.ts,\n pkg/src/c.ts,\n"
  -> scanned 3 source files, 1 guard, no EISDIR
```

## Verification

| check | result |
|---|---|
| production default | **unchanged** — 1961 files, 0 guards, `BACKLOG ZERO` |
| `--strict` / `--json` / `--compare` | 0 / 0 / 0 |
| clean injected list | 3 files / 1 guard (unchanged) |
| census suite | 53 passed |
| eslint / `check-fnxc-future-dates` / `pnpm test:gate` | clean / 0 / 0 |

`--compare` is the one that matters for a reader change: it runs the AST and text classifiers over the same list and fails if they disagree, so it would catch a normalization that silently altered what either one sees.

PR #3231, opened 2026-07-31
fix(engine): mock provider sends 0-based steps — test mode full-task runs complete again

gsxdsm:
Found by a live browser E2E of the coding workflow in test mode: every scripted full-task run failed at `steps#0:step-execute` with `Step 4 out of range (task has 4 steps)`, rebounding through recovery forever.

**Root cause:** `fn_task_update.step` has been **0-based since FN-6607** (executor.ts FNXC:StepNumbering — the old `step - 1` conversion made Step 0 impossible to mark). `mock-provider.ts` still sent `index + 1`, so test mode marked steps 1..N instead of 0..N-1: Step 0 (Preflight) never completed and step N threw out-of-range. Test mode's full-task path has been broken since June.

**Also fixes the test that pinned the bug:** `mock-provider.test.ts` expected `{ step: 1 }` for a fixture whose first unfinished step is index 0 — the expectation encoded the 1-based off-by-one.

Verified: 12/12 mock-provider tests; the live E2E instance completes the task after this patch (see follow-up screenshot in the session).

### DIFF
--- a/.changeset/mock-provider-step-indexing.md
+++ b/.changeset/mock-provider-step-indexing.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Test mode tasks complete again — the mock executor marked the wrong steps since the 0-based step change.
+category: fix
+dev: mock-provider.ts sent 1-based step numbers to fn_task_update (0-based since FN-6607), so scripted full-task runs never marked Step 0 and failed with "Step N out of range" at steps#0:step-execute.
--- a/packages/core/src/__tests__/central-core-init-bootstrap-guard.test.ts
+++ b/packages/core/src/__tests__/central-core-init-bootstrap-guard.test.ts
@@ -1,0 +1,38 @@
+/**
+ * FNXC:CentralCore 2026-07-28-03:00:
+ * Regression guard for GET /api/activity-feed 500
+ * (`backendHandle is only available in backend mode (asyncLayer injected)`).
+ * #2454 left `createCentralBackendLayer` as dead code behind an early return on
+ * layer-less `init()`. Dashboard/CLI call sites use `new CentralCore(); await init()`
+ * without `attachBackendLayer`, so init must still bootstrap PG.
+ */
+// @vitest-environment node
+import { readFileSync } from "node:fs";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
+import { describe, expect, it } from "vitest";
+
+const centralCoreSrc = readFileSync(
+  join(dirname(fileURLToPath(import.meta.url)), "../central-core.ts"),
+  "utf8",
+);
+
+describe("CentralCore.init layer-less PG bootstrap", () => {
+  it("keeps createCentralBackendLayer reachable (no #2454 early-return no-op)", () => {
+    const initStart = centralCoreSrc.indexOf("async init(): Promise<void>");
+    expect(initStart).toBeGreaterThan(-1);
+    const initEnd = centralCoreSrc.indexOf("async close(): Promise<void>", initStart);
+    expect(initEnd).toBeGreaterThan(initStart);
+    const initBody = centralCoreSrc.slice(initStart, initEnd);
+
+    expect(initBody).toContain("createCentralBackendLayer");
+    // The regression: after the asyncLayer branch, mark initialized and return
+    // without ever reaching createCentralBackendLayer.
+    expect(initBody).not.toMatch(
+      /if \(this\.asyncLayer\) \{[\s\S]*?\n\s*\}\s*\n\s*this\.initialized = true;\s*\n\s*return;\s*\n/,
+    );
+    // Bootstrap must still assign the owned layer + shutdown hooks.
+    expect(initBody).toContain("ownedBackendShutdown");
+    expect(initBody).toContain("ownedBackendReleaseConnections");
+  });
+});
--- a/packages/core/src/central-core.ts
+++ b/packages/core/src/central-core.ts
@@ -355,7 +355,13 @@ export class CentralCore extends EventEmitter<CentralCoreEvents> {
   private async initializeOnce(): Promise<void> {
     this.assertOpen();
     if (this.initialized) return;
+    /*
+    FNXC:SqliteDualPathCleanup 2026-07-26-14:15:
+    CentralCore.init is PostgreSQL-only (SQLite CentralDatabase path deleted).
 
+    FNXC:CentralCore 2026-07-28-03:00:
+    #2454 accidentally early-returned on layer-less init and left the PG bootstrap as dead code. Dashboard routes (e.g. GET /api/activity-feed) and CLI fallbacks call `new CentralCore(); await init(); getRecentActivity()` without attachBackendLayer, so a no-op init throws backendHandle ("only available in backend mode"). Restore layer-less createCentralBackendLayer bootstrap. When a pre-injected asyncLayer exists, bootstrap that shared layer only. Runtime serve may still call attachBackendLayer later to adopt the TaskStore pool (releases any owned central-only pool).
+    */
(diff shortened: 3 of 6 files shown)


gsxdsm:
Attempted the rebase onto `origin/main` (dd09e57511) and **stopped partway — the branch is untouched** (`git rebase --abort`). Two of the three conflicts I could resolve confidently; the third needs your knowledge, and guessing at it risks dropping a fix.

## Resolved cleanly (not pushed — they came with the aborted attempt)

**`App.test.tsx` — take main's side.** Main replaced the mock with a separate control and documented why:

```ts
const mockIsShortViewport = vi.fn(() => false);
isShortViewport: () => mockIsShortViewport(),
```

Your commit sets `isShortViewport: () => mockUseViewportMode() === "mobile"`, which is precisely the conflation main fixed: `isShortViewport()` reads `(max-height: 480px)` alone, while mobile mode is the OR of width and height — so an ordinary **portrait phone is mobile and NOT short**. Keeping your side would regress that and silently re-hide the most common real viewport. Your commit message ("the mock must export every helper the rendered tree consumes") is right about the *cause*; main fixed it and went further.

## Where I stopped: `central-core.ts`

Main restructured `init()` into `initializeOnce()` behind an `initializationPromise` so concurrent callers share one in-flight attempt. Your commit restores the layer-less PG bootstrap in the same region. Different changes, same lines.

**Main appears to already carry your fix.** In main's `init()` after the `if (this.asyncLayer)` early return:

```
45:  const { createCentralBackendLayer } = await import("./postgres/startup-factory.js");
46:  const backend = await createCentralBackendLayer({ globalSettingsDir: this.globalDir });
```

That is the layer-less bootstrap whose absence #2454 caused and your commit restores. If it landed independently, `fix: restore CentralCore PG bootstrap for activity-feed` may now be redundant — dropping it during the rebase would likely make this conflict and the `register-setup-activity-routes.test.ts` one disappear together.

I did not verify that the two implementations are *equivalent*, only that main has a layer-less `createCentralBackendLayer` path. Whether yours adds something main's lacks is your call, which is exactly why I aborted instead of resolving.

## Suggested rebase

```bash
git rebase -i origin/main   # drop a2d30dffc2 if main already has it
# App.test.tsx conflict: take main's mockIsShortViewport version
```

Happy to run it if you confirm the drop; I would rather not decide that one for you.

gsxdsm:
Sweeping this for the inert-lane and census ratchets, I hit something worth flagging: **this branch predates the gate scripts themselves.** `scripts/check-inert-sync-lane-conversions.mjs` does not exist on it, so running the checks against its tree errors with `Cannot find module` rather than reporting a count.

That is base drift far enough back to matter. GitHub agrees on the consequence — `mergeable=CONFLICTING, mergeStateStatus=DIRTY` — and a local merge onto current `main` conflicts in three files:

```
packages/core/src/central-core.ts
packages/dashboard/app/components/__tests__/App.test.tsx
packages/dashboard/src/routes/__tests__/register-setup-activity-routes.test.ts
```

The lane instruments themselves are fine on the merge result — ratchet 19 / exit 0, census backlog 0, `--strict` exit 0 — so **nothing here threatens the lifecycle work**; this is purely a rebase note. Flagging it because a branch old enough to be missing the gate scripts is also old enough that its local runs cannot tell you whether it passes them, which is easy to miss when the errors look like tooling noise rather than a stale base.

Method note in case it is useful to others sweeping PRs: for a branch this far behind, measuring its own tree is meaningless and measuring `main`'s scripts against its tree is not much better. The only number that means anything is the **merge result**, which is what CI evaluates — and here that could only be produced after resolving the three conflicts, so the figures above come from the clean-merge portion.

gsxdsm:
**Started the rebase, resolved the first conflict, aborted on the second. Nothing pushed** — remote is untouched at `27ba429d1`.

## Conflict 1 — `App.test.tsx`: take main's, the branch's stub is the known-bad one

```
branch:  isShortViewport: () => mockUseViewportMode() === "mobile"
main:    defaults to false, with a #2846 note explaining why
```

Main's note records this exact stub as a defect it fixed: `isShortViewport()` reads `(max-height: 480px)` **alone**, while mobile mode is the OR of width and height. An ordinary portrait phone is mobile and **not** short, so tying them made every mobile test take the short-viewport branch — silently mis-testing `FloatingWindow`'s geometry-persistence suspend and `PlanningModeModal`'s compact layout. The most common real viewport was never exercised.

I resolved this one to main and it applied cleanly.

## Conflict 2 — `central-core.ts`: why I stopped

Main carries a newer concurrency guard (`FNXC:CentralCore 2026-07-29-16:10`):

```ts
if (!this.initializationPromise) {
  this.initializationPromise = this.runLifecycleOperation(() => this.initializeOnce());
}
```

— sharing one in-flight bootstrap so a second backend cannot be orphaned when ownership fields are overwritten, cleared afterward so a failed bootstrap stays retryable. This branch predates it (`2026-07-26`).

Choosing a side decides whether that protection survives, in core initialization, on a multi-commit rebase. That needs the core suite behind it, and I do not have the budget to verify it honestly. A visibly conflicted PR costs less than a silently dropped concurrency guard.

## Suggested resolution

1. `App.test.tsx` → **main's** version.
2. `central-core.ts` → keep **main's** `initializationPromise` guard.
3. Arbitrate with the test this branch itself adds:
   `pnpm --filter @fusion/core exec vitest run src/__tests__/central-core-init-bootstrap-guard.test.ts`

That test is the right judge — it is the branch's own statement of the invariant, so if main's guard satisfies it, the newer implementation wins on the branch's own terms.

gsxdsm:
**Rebased onto `origin/main` (dd09e57511) — MERGEABLE.** I said last round I would not decide the `central-core.ts` conflict for you; I then checked whether it needed deciding, and it did not.

## `central-core.ts` — your fix is already on main, verbatim

I diffed both sides. From `if (this.asyncLayer)` through the closing `catch`, main's `initializeOnce()` body and your `init()` body are **byte-identical** — same `mkdir`, same `createCentralBackendLayer` import, same `ensureBackendBootstrap`, same ownership assignments, same `backend.shutdown()` cleanup on failure. Main simply moved that body into `initializeOnce()` behind an `initializationPromise` so concurrent callers share one in-flight attempt.

So there was nothing of yours to lose in the code, and the merge is not a judgement call after all. Resolved to main's structure and **grafted your comment onto it** — the `#2454` note explaining *why* the layer-less bootstrap must exist is the one thing main lacked, and it is the part a future reader needs most:

```
355:  private async initializeOnce(): Promise<void> {
363:    #2454 accidentally early-returned on layer-less init and left the PG bootstrap as dead code...
```

## The other two

- **`App.test.tsx`** — took main's side. It replaced the mock with a separate `mockIsShortViewport` control defaulting to `false`, because `isShortViewport()` reads `(max-height: 480px)` alone while mobile mode is the OR of width and height — so an ordinary **portrait phone is mobile and NOT short**. Your `=== "mobile"` is the conflation main fixed; keeping it would have re-hidden the most common real viewport. Your commit message is right about the cause; main went further.
- **`register-setup-activity-routes.test.ts`** — took main's side. The cross-project cap is deleted, so `/api/global-concurrency` is telemetry-only; your version still asserts `getGlobalConcurrencyState` and `globalMaxConcurrent`, which no longer exist. (The one surviving mention is inside main's own comment saying the route no longer calls it — identical to main, not a leftover.)

## Verified after the rebase

| check | result |
|---|---|
| `@fusion/core` tsc | exit 0 |
| `central-core-init-bootstrap-guard` | 1 passed |
| `register-setup-activity-routes` | 3 passed |
| `App.test.tsx` | **143 passed** |

Your own commit — the 0-based mock-provider step fix — is untouched and rides on top.

If you would rather keep your `init()` shape over main's `initializeOnce()`, say so and I will redo it; main's has a concurrency guard yours does not, which is why I kept it.

PR #3234, opened 2026-07-31
test(core): pin both create-time duplicate guards' lane exclusions

gsxdsm:
## What

Pins **both** create-time duplicate guards in `branch-and-pr-entities.ts`. Test-only.

| site | method | excludes |
|---|---|---|
| `:445` | `findRecentTasksByContentFingerprint` | ARCHIVED (unless `includeArchived`) |
| `:484` | `findRecentTasksBySourceParentTaskId` | COMPLETE and ARCHIVED |

Blinding either back to its literals left the entire 16-file lane-detector set green. **No test in `packages/core` reaches either method.**

## Measured

```
converted:     Tests 8 passed (8)
blinded :445   Tests 1 failed | 7 passed (8)    <- only the fingerprint case
blinded :484   Tests 2 failed | 6 passed (8)    <- only the sibling cases
```

**Each blind fails exactly its own cases.** That matters: it proves the two resolvers are pinned *independently*, rather than one broad test appearing to cover both. Blinding `:445` leaves every sibling case green and vice versa — so neither is riding on the other's coverage.

## They fail in opposite directions

This is why both belong in one file:

- **Fingerprint guard** — a renamed board leaves archived cards in the candidate set, so filing a new task is **refused as a duplicate** of one the operator already archived. The create is blocked and the thing blocking it is invisible.
- **Sibling guard** — a renamed board leaves finished siblings in the "recent live siblings" set, so completed work keeps counting as active.

One over-includes into a *refusal*, the other over-includes into *phantom activity*. Neither raises an error.

## Positives pinned too

A LIVE fingerprint match is still a duplicate candidate; `includeArchived: true` opts the renamed archived lane back in; a WORKING sibling is still live. Excluding the finished lanes must not degrade into excluding everything, or the guards stop guarding — the failure mode a lane-widening change invites.

## A fixture detail that would have made this vacuous

Both queries cut off at `Date.now() - windowMs`, with `windowMs` capped at 24h. The sibling harness I copied from seeds a **fixed past timestamp**, which falls outside that window — every case would then pass on an empty result, including the ones that are supposed to fail under blinding. Fixtures are seeded at current time instead, and the reason is recorded in the file so nobody "tidies" it back to a frozen date.

## Progress

3 of the 5 uncovered core sites are now pinned (`store.ts:1135` in #3233, these two here). Remaining and unclaimed: `async-mission-store.ts:1179` and `task-id-integrity.ts:502`.

### DIFF
--- a/packages/core/src/__tests__/postgres/duplicate-and-sibling-lanes.pg.test.ts
+++ b/packages/core/src/__tests__/postgres/duplicate-and-sibling-lanes.pg.test.ts
@@ -1,0 +1,170 @@
+/*
+FNXC:WorkflowResolvedColumns 2026-07-31-23:15:
+TWO CREATE-TIME DUPLICATE GUARDS, on a RENAMED board.
+
+Both decide "is this candidate still LIVE?" by excluding finished lanes, and both were converted to
+`resolveProjectColumnsForRoles`:
+
+  findRecentTasksByContentFingerprint  — excludes ARCHIVED (unless includeArchived)
+  findRecentTasksBySourceParentTaskId  — excludes COMPLETE and ARCHIVED
+
+Blinding either back to its literals left the whole 16-file lane-detector set green: no test in
+`packages/core` reaches either method.
+
+WHAT BREAKS WITHOUT THEM. They fail in OPPOSITE directions, which is why both are here.
+
+  Fingerprint guard — a renamed board leaves archived cards in the candidate set, so filing a new
+  task is REFUSED as a duplicate of one the operator already archived. The operator cannot create
+  the card and the thing blocking them is invisible.
+
+  Sibling guard — a renamed board leaves finished siblings in the "recent live siblings" set, so a
+  completed sibling keeps counting as active work.
+
+Each is a silent wrong answer on a create path, and neither surfaces an error.
+
+DIFFERENTIAL. The same seeded rows are queried under two vocabularies whose traits are identical and
+only the ids differ; `shipped`/`vaulted` collide with no legacy id. The default-vocabulary cases are
+controls — they pass with or without the conversion.
+
+WINDOWED. Both queries cut off at `Date.now() - windowMs` with `windowMs` capped at 24h, so fixtures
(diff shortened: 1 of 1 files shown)


gsxdsm:
Verified your coverage claim statically. **It holds** — and it is stronger than a grep suggests, which is worth recording before someone thinks it is contradicted.

A reviewer checking "no test reaches either method" will find **ten hits** in `duplicate-guard.test.ts` for `findRecentTasksByContentFingerprint`. Every one is a mock or a spy:

```
:37   findRecentTasksByContentFingerprint: vi.fn().mockImplementation(...)
:105  vi.spyOn(store, "findRecentTasksByContentFingerprint").mockResolvedValueOnce([existing])
:126  …mockResolvedValue([foreignSibling])      :198 :217 :273 :291 :329 :359
```

The real implementation is never on the path, so the blind cannot reach it — exactly your point, but "no test reaches it" and "ten tests name it" both being true is the kind of thing that stalls a review.

## The causal detail, which your body does not have

That file also explains **why** the coverage disappeared, in its own words at `:382`:

> These three drove `findRecentTasksByContentFingerprintImpl` against a fake modelling the DELETED SQLite path (`db.prepare().all()`) … that fake broke when the query moved to `asyncLayer` + Drizzle.

So there *were* tests executing the real `Impl`; they were deliberately retargeted at the pure `resolveFingerprintWindowMs` policy when the SQLite fake broke on the PG cutover. That was the right call — rebuilding a Drizzle chain to recover a number the policy already returns is mock-the-world for no gain — but it is the moment the lane exclusions became unheld. Your PR closes a gap the PG cutover opened, which is a more useful framing than "no test covered this".

## What I could not verify

I cannot reproduce your `8 / 1-failed / 2-failed` numbers: this is a `.pg` suite and TCP PostgreSQL is not reachable here (`55433` closed), so it skips. I am confirming the **coverage claim and the call-site facts**, not the runs. Flagging explicitly because #3222's doc records that a skipped `.pg` suite is indistinguishable from a passing one unless you check.

The per-blind isolation is the right design — each blind failing only its own cases is what proves the two resolvers are pinned independently rather than one broad assertion appearing to cover both.

PR #3235, opened 2026-07-31
test(core): pin the last two uncovered lane reads (mission archive, lineage gate)

gsxdsm:
## What

Pins the **last two uncovered lane reads** in `packages/core`. Test-only. This closes the per-site core audit.

| site | what it decides |
|---|---|
| `async-mission-store.ts:1179` | is an ARCHIVED card valid terminal evidence for mission repair? |
| `task-id-integrity.ts:502` | does an archived child still count as a LIVE lineage child? |

## Measured

```
mission-store:  39 passed clean;  1 failed | 38 passed blinded
lineage:         3 passed clean;  1 failed |  2 passed blinded
```

Both blinds confirmed applied with `git diff --stat` before each run.

## The third adjacent-pair split

`:1179` is the **archived** half of a pair whose **complete** half (`:1178`, *one line above*) was already covered by a test in the same file, written for exactly this concern. Terminal evidence is "done OR supported archived state," so an archived card is equally valid repair evidence — but on a board whose archive lane is `vaulted` the archived half could not see it, and reconciliation threw `TASK_NOT_TERMINAL` for a card that was genuinely filed away. Same refusal the covered case fixed, reached through the other door.

That is now the third confirmed instance in core (after `team-analytics` in #3227 and the scheduler pair earlier). **Being adjacent to a covered resolver is not coverage**, and it is the most reliable place to look.

## What breaks without the lineage read

An archived child is filed away, not live, so it must not hold the delete gate shut. Renamed, it still counted as live and `TaskHasLineageChildrenError` blocked the parent's delete **forever** — the operator archived the child *precisely* to clear the way, and the gate could not see that they had.

## A fixture detail I got wrong first

My first mission fixture created a live card in a `vaulted` column and failed with `deleted or archived without a valid retained tombstone and archive snapshot` — nothing to do with the lane read.

The `archived` verdict requires **all three** of `deletedAt !== null`, an archive-snapshot row, and `isArchived(column)`. A live card merely sitting in an archive-trait column is `invalid-deleted`, not `archived`. The test now archives for real and *then* renames the recorded lane, which isolates the third condition — the only one under test. Recorded in the file so the next person does not re-derive it.

## Paired positives

Both files pin the complement: a WORKING child still counts as live. Recognising the renamed archive lane must not degrade into "no child is ever live" — that would silently **disable** the lineage gate and let a parent be deleted out from under real descendants, which is worse than the bug being fixed.

## Core audit complete

**14 sites blinded individually: 9 already covered, 5 uncovered, all 5 now pinned** (#3233, #3234, this PR).

Every `resolveProjectColumnsForRoles` call site in `packages/engine` and `packages/core` has now been blinded. Remaining unaudited: `dashboard` (2 files) and `cli` (1) — I claim nothing about those.

### DIFF
--- a/packages/core/src/__tests__/postgres/live-lineage-children-lanes.pg.test.ts
+++ b/packages/core/src/__tests__/postgres/live-lineage-children-lanes.pg.test.ts
@@ -1,0 +1,108 @@
+/*
+FNXC:WorkflowResolvedColumns 2026-07-31-23:45:
+THE LINEAGE-INTEGRITY GATE'S ARCHIVE READ, on a RENAMED board.
+
+`findLiveLineageChildren` answers "does this parent still have LIVE children?" — the gate that
+refuses to delete a parent while lineage descendants remain. An ARCHIVED child is filed away, not
+live, so it must not hold the gate shut.
+
+WHY THIS FILE EXISTS. The archive read was converted to
+`resolveProjectColumnsForRoles(store, ["archived"])`, and blinding it back to the legacy id left the
+whole 16-file lane-detector set green. `store.findLiveLineageChildren` is public and nothing in
+`packages/core` exercises it against a renamed board.
+
+WHAT BREAKS WITHOUT THE CONVERSION. On a board whose archive lane is `vaulted`, an archived child is
+not recognised as archived, so it still counts as live and `TaskHasLineageChildrenError` blocks the
+parent's delete forever. The operator archived the child precisely to clear the way, and the gate
+cannot see that they did. This is the renamed-board twin of the defect #3162 fixed.
+
+DIFFERENTIAL. Same seeded rows under two vocabularies with identical traits; only the ids differ, and
+`vaulted` collides with no legacy id. The default-vocabulary case is the control.
+*/
+
+import { afterAll, afterEach, beforeAll, beforeEach, expect, it } from "vitest";
+import { sql } from "drizzle-orm";
+import {
+  createSharedPgTaskStoreTestHarness,
+  pgDescribe,
+  type SharedPgTaskStoreHarness,
+} from "../../__test-utils__/pg-test-harness.js";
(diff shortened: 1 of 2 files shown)


gsxdsm:
Both sites check out and the adjacent-pair finding is the valuable part — but I want to push on the **completeness** claim, because it is the one thing here that cannot be re-derived later.

## Confirmed

- `async-mission-store.ts:1178/1179` are a genuine `complete:`/`archived:` pair, one line apart, and `:1178` is the covered half. Your framing is exactly right: *terminal evidence is "done OR supported archived state"*, so on a board whose archive lane is `vaulted` the archived half refuses a card that is genuinely filed away — the same `TASK_NOT_TERMINAL` refusal the covered case fixed, reached through the other door.
- `task-id-integrity.ts:502` is present and is the archived-child lineage read you describe.

**"Being adjacent to a covered resolver is not coverage"** is the most useful sentence in this PR, and your third confirmed instance earns it. That is a search heuristic the next person can actually apply.

## The claim I cannot verify: "the last two"

There are **58** awaited lane-resolver call sites in `packages/core` (non-test):

```
await resolveProjectColumnsForRoles( | await resolveTaskLifecycleColumns( | await resolveArchivedLanes(   -> 58
```

"Last two uncovered" is a statement about the other 56, and nothing in the PR body shows the enumeration behind it. I am not disputing it — you have been running this audit per-site and I have not — but a completeness claim is exactly the kind that gets quoted later as "core is fully covered" long after it stopped being true. It also cannot be reconstructed from the diff.

Would you add the enumeration, even tersely? A list of the 56 with how each was cleared (covered by suite X / mocked so unreachable / `.pg`-gated / static guard) turns "trust me" into something the next worker can re-run and extend. If that is already in a scratch file, pasting it into the body is enough.

## What I could not check

`.pg` suites, TCP `55433` closed here, so your `39/1-failed` and `3/1-failed` numbers are unreproducible from this machine — I am confirming the call-site facts and the reasoning, not the runs. Noting it because a skipped `.pg` suite reads identically to a passing one.

PR #3236, opened 2026-07-31
docs(solutions): complete the repo-wide resolver audit; correct a superseded note

gsxdsm:
## What

Completes the repo-wide resolver audit and **corrects a note of mine that had gone stale**. Docs only.

Every `resolveProjectColumnsForRoles` call site in the repository has now been blinded individually.

## Final results

| package | sites | outcome |
|---|---|---|
| `engine` | 10 files | scheduler, triage, evaluator uncovered → pinned; executor, restart-recovery, notification already covered; self-healing 21 pinned / 1 inert |
| `core` | 14 | 9 covered, **5 uncovered → all 5 pinned** (#3225, #3227, #3233, #3234, #3235) |
| `dashboard` | 4 | `register-task-workflow-routes.ts:1268` covered; `server.ts` ×3 flagged |
| `cli` | 1 | flagged |

## The correction

A note recorded `workflow-analytics.ts` and `team-analytics.ts` — 4 resolvers — as **unmeasurable**, because `pgDescribe` probes TCP and the `.pg` suites skip without it.

The caution is real and stays: a skipped suite reads exactly like a passing one. But on an environment where those suites **do** run, all 4 were measured, and `team-analytics.ts` turned out to have a half-covered pair — `completeLanes` covered, **`activeLanes` not** — in a file named `team-analytics-renamed-lanes`. That is now pinned (#3227, merged).

Left standing, the note converts a real finding into a **permanent excuse for not looking**. It now says: confirm the suite actually skips *here* before recording a site as unmeasurable for environment reasons.

## A fourth measurement failure mode — the opposite direction

The three already recorded all produce false *uncovered*. This one produces false *covered*:

**A COVERED verdict needs a baseline.** The dashboard sweep reported 5 failing files under the global blind. **4 of them fail on clean `main`** and have nothing to do with lanes — a docs-inventory test and a model-routes test among them. Read as-is, that is four resolvers falsely credited as covered. Only `register-task-workflow-routes.awaiting-planning.test.ts` passes clean and fails blinded, so it is the sole real detector.

Second time today a baseline changed a conclusion (the first found a genuine red on main, #3229).

## Why 4 sites are flagged rather than pinned

- **`server.ts:1922/1923/1938`** — inside the `/api/health/reliability` route closure. No route-level test exists, and the only way in is booting `createServer(store)` behind a mock-the-world shell, which the slow-test rule forbids. The alternative is a refactor to expose a seam — its own commit, since moving code and changing behaviour do not ride together. (A note already in this doc reached the same conclusion independently; this confirms it by measurement.)
- **`cli/commands/task.ts:660`** — worth its own warning. Extracting a pure helper and testing it **would look like coverage and would not be**: blinding the resolver leaves such a test green, because the helper *receives* the lane set rather than resolving it. The uncovered thing is the resolve call, not the decision it feeds. Its sibling test file already records the same limit honestly for `boardColumnsForDisplay`.

## Reported, not fixed: 4 pre-existing red dashboard files on main

`lazy-loaded-views-docs.test.ts` (AGENTS lazy-view inventory drifted — 24 actual vs 18 documented), `ResearchView.test.tsx`, `planning-browser-e2e.test.ts`, `register-model-routes-kimi-k3-supplemental.test.ts` — 7 failing tests, all in the non-blocking suite.

I am not fixing them here: the lazy-views inventory is a curated list other workers are actively adding to, and rewriting it mid-flight would collide. Flagging so it is visible rather than silently absorbed into my blind's noise.

## Summary by CodeRabbit

* **Documentation**
  * Updated workflow guidance to require baseline comparisons and verification that all relevant tests run.
  * Added safeguards for detecting ineffective changes and distinguishing pre-existing failures.
  * Expanded PostgreSQL audit documentation with measured coverage results, including uncovered resolver paths.
  * Recorded completed coverage sweeps across core, dashboard, and CLI areas, including pinned and non-pinnable sites.
  * Clarified limitations when testing extracted decision helpers instead of resolver calls.

### DIFF
--- a/docs/solutions/workflow-learnings/blind-the-resolver-to-find-uncovered-conversions.md
+++ b/docs/solutions/workflow-learnings/blind-the-resolver-to-find-uncovered-conversions.md
@@ -115,10 +115,18 @@ itself. Three in one session, each of which reads exactly like coverage:
 | `blind3.py <file> <var>` with an unmapped role | the script exited non-zero **silently**, `&&` skipped the check, `;` let vitest run against **unmodified source**. Reported 375/375 green "under blinding" with nothing blinded. |
 | `vitest run src/__tests__/notification` | missed `src/notification/__tests__/` entirely — a nested `__tests__` the glob never reached. Reported covered code as uncovered. |
 
-So: **an UNCOVERED verdict is a claim about the whole tree and needs the whole tree's tests.** Before
-believing one, confirm (a) the blind actually modified the file — `git diff --stat`, not the tool's
-exit code, and (b) the run included every test file that imports the module, including nested
-`__tests__` directories. A tool that can no-op must say what it changed and fail loudly when it
+A fourth, in the opposite direction: **a COVERED verdict needs a baseline.** The dashboard sweep
+reported 5 failing files under the global blind; **4 of them fail on clean `main`** and had nothing
+to do with lanes (a docs-inventory test, a model-routes test). Read as-is, that is four resolvers
+falsely credited as covered. Run the failing set again *unblinded* and subtract — whatever fails in
+both is noise. Pre-existing red is common enough in a non-blocking suite that this is not an edge
+case; it is the second time today it changed a conclusion.
+
+So: **an UNCOVERED verdict is a claim about the whole tree and needs the whole tree's tests**, and a
+COVERED verdict needs a baseline to subtract. Before believing either, confirm (a) the blind actually
+modified the file — `git diff --stat`, not the tool's exit code, (b) the run included every test file
+that imports the module, including nested `__tests__` directories, and (c) the failures you are
+crediting do not also fail without the blind. A tool that can no-op must say what it changed and fail loudly when it
 cannot; the same standard this program applies to product guards applies to the audit's own
 instruments.
 
@@ -145,10 +153,14 @@ Three groups resisted blinding for reasons worth writing down, so the next perso
 them. Recording *why* a site cannot be measured is a result — the same stance #3212 took for a site
 that cannot be covered.
 
-- **No TCP PostgreSQL.** `workflow-analytics.ts` and `team-analytics.ts` (4 resolvers) keep their
-  renamed-lane coverage in `.pg` suites. `pgDescribe` probes **TCP**; `pg_isready` succeeding on a
-  **Unix socket** is not the same thing, and mistaking one for the other turns 4 skipped suites into
-  4 false "uncovered" readings.
+- **~~No TCP PostgreSQL.~~ SUPERSEDED — these were measurable and one hid a real gap.** The caution
+  itself stands: `pgDescribe` probes **TCP**, `pg_isready` on a **Unix socket** is not the same
+  thing, and a skipped suite reads exactly like a passing one. But on an environment where the `.pg`
+  suites do run, all 4 resolvers were measured: `workflow-analytics.ts` has **both** halves covered,
+  `team-analytics.ts` has `completeLanes` covered and **`activeLanes` uncovered** — a half-covered
+  pair in a file named `team-analytics-renamed-lanes`, now pinned. **Before recording a site as
+  unmeasurable for environment reasons, confirm the suite actually skips here** — otherwise the note
+  converts a real finding into a permanent excuse for not looking.
 - **No injectable seam.** `reads.ts`'s incremental-sync scan composes Drizzle conditions against
   `layer.db` directly. A test there would assert the query that was built rather than the rows that
   were excluded — green, and blind to the bug.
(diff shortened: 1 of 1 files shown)


gsxdsm:
The superseded-note correction is the most valuable part of this, and worth pulling out of the bullet it sits in:

> **Before recording a site as unmeasurable for environment reasons, confirm the suite actually skips here** — otherwise the note converts a real finding into a permanent excuse for not looking.

That is a flag-quality rule, and it generalises past PG: `pgDescribe` probes TCP, `pg_isready` on a Unix socket does not, **and a skipped suite reads exactly like a passing one**. Four resolvers were recorded as unmeasurable and all four were measurable, one hiding a half-covered pair. A deferral note is a claim like any other and decays the same way — this is the first case I have seen where re-deriving one paid for itself immediately.

The COVERED-needs-a-baseline finding is the other half nobody had: 4 of 5 "failures" being pre-existing red unrelated to lanes would have credited four resolvers as covered. `subtract the unblinded run` belongs next to `throw to prove the suite reaches the site` as a standing pre-check.

## One population this procedure will misreport, when you get to it

The audit covers **`resolveProjectColumnsForRoles`** — 116 sites, and that resolver genuinely resolves, so a green blind means what the doc says it means.

There is a second population where it will not: guards resolved through **`resolveTaskWorkflowIrSync`**. `pnpm check:inert-sync-lanes` counts **19** of them today (`scheduler.ts` 13, `triage.ts` 4, `executor.ts` 2). That reader answers with the DEFAULT board for every task under PostgreSQL, so blinding it back to legacy ids is a **semantic no-op** — the code runs, the fixture is fine, the suite is well chosen, and it stays green regardless. Blind all 19 and all 19 report UNCOVERED.

That is a fifth failure mode, and it is not caught by the four checks now in the doc: `git diff --stat` shows the edit landed, `throw` goes red so the suite plainly reaches the site, the unblinded baseline subtracts to nothing, and the test selection is correct. Every gate passes and the verdict is still wrong — the site has nothing to measure, rather than nothing measuring it. The work item it generates is 19 tests that cannot fail.

Cheap addition to the pre-checks, since the list is already scripted: **before concluding UNCOVERED, confirm the site is not on `check:inert-sync-lanes`.** If it is, the finding is "the resolver cannot answer", and the fix is a real resolver, not a test.

Not asking for it in this PR — the audit's scope is `resolveProjectColumnsForRoles` and this does not affect any verdict in it. Flagging because "what has and has not been audited" makes those 19 the obvious next target, and they are exactly the set where a clean run of this procedure produces a uniformly false answer.

gsxdsm:
Checked the four pre-existing reds you report. **One of them is already fixed on main; three still reproduce.**

## `lazy-loaded-views-docs.test.ts` — fixed, drop it from the list

Landed as `f0a13745b2` (*"lazy-views doc parser stops at any heading — six phantom views came from the H2 that follows"*). It passes 3/3 on current main.

I derived the same root cause independently before finding that commit, which is worth stating because **your framing of it is wrong in a way that matters**: the body calls it *"AGENTS lazy-view inventory drifted — 24 actual vs 18 documented"*. The inventory never drifted. It is correct at 20, and no lazy view was added or removed.

The bug was in the extractor. Its lookahead stopped at `\n### ` and `\n---` only, but the heading after that section is `## FNXC_LOG comments:` — one level **up**, matching neither. So the parse window ran to end of file and harvested six backticked tokens from the FNXC bullets:

```
EXTRA: ['date -u', 'check-fnxc-future-dates', 'pnpm lint', 'main', 'scheduler.ts', 'task-update.ts']
```

That distinction is the whole point: read as inventory drift, the fix is to edit a curated list other workers actively add to — which is exactly the collision you (correctly) did not want to walk into. Read as a parser bug, it is a one-line lookahead change touching nothing shared. **Your instinct to leave it alone was right; the stated reason was not.**

## Still red on current main

```
ResearchView.test.tsx                              4 failed | 23 passed (27)
register-model-routes-kimi-k3-supplemental.test.ts exit 1
```

Both reproduce, so those parts of your report stand.

## On your fourth failure mode

*"A COVERED verdict needs a baseline"* is the right lesson and I can confirm it from the other direction: I published a `24/24` sweep this week that was a **zsh word-splitting bug** — `for f in $FILES` does not split, so one broken invocation printed as 24 passing suites. False *covered* from a broken harness rather than a missing baseline, but the same shape: a uniform result nobody baselined.

Worth adding the general form to the doc — **a sweep returning a perfectly uniform result across every input is more likely broken than informative.** `24/24` and `0/24` should both trigger a re-check before they become a conclusion.

gsxdsm:
Both findings were real and are fixed.

## Site totals — 15/9/6, not 14/9/5

The reviewer is right, and the cause is worth recording: the `branch-and-pr-entities.ts` row lists **two** sites (`:445` and `:484`). I counted table rows rather than sites.

Recounted explicitly:

- uncovered — `store.ts:874`, `store.ts:1135`, `branch-and-pr-entities.ts:445`, `:484`, `async-mission-store.ts:1179`, `task-id-integrity.ts:502` → **6**
- covered — `async-mission-store.ts:1178`, `reads.ts` ×3, productivity/github/gitlab analytics, `eval-automation`, `task-artifacts-ops` → **9**
- total **15**

Added an inline note naming the two-sites-one-row cause, so the next reader does not repeat the miscount.

## `git diff --stat` — scoped to the target

This one is sharper than it looks. The step was the document's own **verification** that the blind actually modified the file, and an unscoped `git diff --stat` reports any unrelated edit in the tree as though the blind had made it. A dirty worktree would confirm a blind that no-opped.

That is precisely the failure this document exists to catalogue — an instrument that cannot return "no" — sitting inside the instructions for avoiding it. Now reads `git diff --stat -- <path/to/target>`, with the reason stated.

PR #3237, opened 2026-07-31
test(dashboard): pin the Reliability endpoint's three lane reads (extract seam + pin)

gsxdsm:
## What

Pins the **Reliability endpoint's three lane reads** — the last uncovered resolver cluster the repo-wide audit found.

Two commits, deliberately separate:
1. **refactor** — extract the three resolves behind `resolveReliabilityLanes(store)`. Behaviour-preserving, no test changes.
2. **test** — pin all three through that seam.

## Why a seam was needed

The three resolves lived inline in the `/api/health/reliability` route closure. Blinding any of them left the **entire dashboard suite green — 21,582 tests** — and the only way to reach them was booting `createServer` behind a mock-the-world shell the slow-test rule forbids.

**And the obvious test would not have helped.** `reliability-metrics.test.ts` exercises `countEntriesInto`, `countBouncesOut` and `inReviewDurationMetrics` with lane sets **passed in by hand**. That proves the collaborators honour a resolved set; it says nothing about whether the caller passes one. *A unit test of the collaborator can never fail when the caller's resolve is blinded* — the same trap the audit note records for `reads.ts`, where a suite written for the exact conversion still could not see it.

The seam is the caller. It resolves, so blinding a resolve fails a test of it.

## Measured — each blind fails exactly its own case

| blinded | fails |
|---|---|
| `REVIEW_ROLES` | "resolves the board's OWN review lane" |
| `["countsTowardWip"]` | "resolves the board's OWN wip lane" |
| `["complete"]` | "resolves the board's OWN complete lane" |

```
converted: Tests 6 passed (6)
each blind: Tests 1 failed  (its own case only)
reliability-metrics.test.ts + this file: 28 passed
typecheck clean; lint clean; fnxc-future-dates: none added
```

That isolation is the point: **three resolves in one function invite a copy-paste that hands the same set to all three**, and every positive assertion would still pass. There is a paired negative asserting each renamed lane appears in *its* bucket and nowhere else — without it the duration metric could silently measure review → review.

Also pinned: the degrade path. An unreadable workflow list must not fail the endpoint, so the legacy ids still answer.

On a board that renames either lane, every underlying query returns `{}` — so `tasksEnteredInReview` and `tasksBouncedToInProgress` are zero for every day, and `inReviewFailureRate7d` divides one zero by another and reports a **healthy** rate. It produces a NUMBER, not an error, and the number says everything is fine. An operator reading 0% review failures beside a populated audit list has no reason to suspect the metric is blind.

## The one observable difference in the refactor, stated not buried

The complete-lane read moves from *after* the counting `Promise.all` into the same phase as the review/wip pair. These are pure reads of workflow definitions — no writes, no ordering dependency — so the resolved values are identical; only the concurrency shape changes (three parallel reads instead of two-then-one). Flagging it because "behaviour-preserving" should be a claim someone can check, not an assertion.

## Audit status

With this, **3 of the 4 flagged sites are closed**. Remaining: `cli/commands/task.ts:660`, where the glyph decision is inline in `runTaskList` and the same seam argument applies — but its sibling test file already documents that driving that function needs the forbidden shell, and extracting a helper there would produce a test that *looks* like coverage while leaving the resolve unpinned. Left flagged rather than faked.

## Summary by CodeRabbit

- **New Features**
  - Reliability health metrics now recognize configured review, work-in-progress, and completion lanes, including renamed workflow lanes.

- **Bug Fixes**
  - Improved fallback behavior when workflow definitions are unavailable, preserving compatibility with legacy lane configurations.
  - Ensured lane resolution remains isolated by role for more accurate reliability metrics.

### DIFF
--- a/packages/dashboard/src/__tests__/reliability-lanes-resolved.test.ts
+++ b/packages/dashboard/src/__tests__/reliability-lanes-resolved.test.ts
@@ -1,0 +1,112 @@
+// @vitest-environment node
+
+/*
+FNXC:WorkflowResolvedColumns 2026-07-31-23:58:
+THE RELIABILITY ENDPOINT'S THREE LANE READS, on a RENAMED board.
+
+`/api/health/reliability` asks three lane questions: which lanes are REVIEW, which count toward WIP,
+and which are COMPLETE. They feed the entered/bounced counts and the review -> done duration metric.
+
+WHY THIS FILE EXISTS. All three were UNCOVERED. Blinding any of them left the entire dashboard suite
+green — 21,582 tests — because they sat inline in the route closure and nothing drives that route.
+
+AND WHY THE OBVIOUS TEST WOULD NOT HAVE HELPED. `reliability-metrics.test.ts` exercises
+`countEntriesInto`, `countBouncesOut` and `inReviewDurationMetrics` with lane sets passed in BY HAND.
+That proves the collaborators honour a resolved set; it says nothing about whether the caller passes
+one. A unit test of the collaborator can never fail when the caller's resolve is blinded. The seam
+under test here is the CALLER: `resolveReliabilityLanes` resolves, so blinding a resolve fails it.
+
+WHAT BREAKS WITHOUT THE CONVERSION. On a board that renames either lane, every underlying query
+returns {} — so `tasksEnteredInReview` and `tasksBouncedToInProgress` are zero for every day, and
+`inReviewFailureRate7d` divides one zero by another and reports a healthy rate. That is the worst
+shape a lifecycle defect takes: it produces a NUMBER, not an error, and the number says everything is
+fine. An operator reading 0% review failures beside a populated audit list has no reason to suspect
+the metric is blind.
+
+DIFFERENTIAL. The same workflow SHAPE under two vocabularies with identical traits; only the ids
+differ, and no renamed id collides with a legacy one. The default-vocabulary cases are controls.
+*/
+
+import { describe, expect, it, vi } from "vitest";
+import { resolveReliabilityLanes } from "../reliability-metrics.js";
+
+const RENAMED = { review: "signoff", wip: "building", complete: "shipped" };
+
+function ir(names: { review: string; wip: string; complete: string }) {
+  return {
+    version: "v2",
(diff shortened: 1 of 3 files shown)


gsxdsm:
Reproduced the per-blind matrix independently. **The seam works and the isolation is real — but your table under-reports what each blind fails, and someone verifying will hit the discrepancy.**

Blinding each resolve to `Promise.resolve(new Set<string>())` in turn:

| blinded | your table says | **measured** |
|---|---|---|
| `REVIEW_ROLES` | 1 case | **3 failed / 3 passed** |
| `["countsTowardWip"]` | 1 case | **3 failed / 3 passed** |
| `["complete"]` | 1 case | **3 failed / 3 passed** |

Every blind fails its own renamed-lane case **plus two shared ones**:

```
default vocabulary: resolves the built-in review, wip and complete lanes
degrades to the legacy floor when the board cannot be read
```

Both assert all three lanes together, so any single blind trips them.

## This makes the PR stronger, not weaker — but the claim needs restating

"Each blind fails exactly its own case" is what proves independent pinning, and the **isolated** cases do discriminate perfectly: `blind review` fails *"resolves the board's OWN review lane"* and neither of the other two; same for wip and complete. That is the load-bearing result and it holds.

What is not true is that only one case fails per blind. You have **three detectors per resolve**, which is better coverage than claimed — but a reviewer who runs your table and sees `3 failed` where it says `1` will stop and wonder what regressed. I nearly did.

Suggest restating as: *each blind fails its own renamed-lane case (proving independence) plus the two aggregate cases (which assert all three lanes at once)*.

## The rest checks out

- `resolveReliabilityLanes` is a genuine caller-side seam — it resolves, so blinding a resolve fails a test of it. That is the distinction the audit note keeps making about `reads.ts`, applied correctly.
- Splitting refactor and test into separate commits is right, and it is what makes the extraction reviewable at all.
- Your point that `reliability-metrics.test.ts` could never have caught this — because it passes lane sets in by hand — is the same trap #3220 closed. Worth noting this is now the fourth instance in that family.

Baseline: 6 passed clean, on `origin/dashboard-reliability-lane-seam`.

PR #3239, opened 2026-07-31
test(dashboard): restore the missing api mock export in ResearchView tests

gsxdsm:
## What

Partial fix for **red main**. Test-only.

`ResearchView.test.tsx` has **4 failing tests on main**; 2 fail with:

```
No "fetchBoardWorkflows" export is defined on the "../../api" mock
```

The `vi.mock("../../api")` factory **replaces the whole module**, so every import anywhere in the rendered tree must appear in it. `fetchBoardWorkflows` reached this file *indirectly* — the task modals ResearchView opens import it — so adding that export to product code broke four cases that have nothing to do with board workflows.

Stubbed with the flag-OFF payload the server sends when multi-lane boards are disabled, which is the shape these cases already assume.

## Measured

```
before: Tests 4 failed | 23 passed (27)
after:  Tests 2 failed | 25 passed (27)
```

## The remaining 2 are a different cause and are NOT fixed here

They fail with `Number of calls: 0` — the enrich-task and create-task actions never fire. That is a UI-wiring question, not mock completeness.

**I checked that my stub is not responsible**, rather than assuming: re-running with `flagEnabled: true` and a populated workflow list produces the *same* 2 failures, so the payload shape does not gate those affordances. Left for whoever owns that surface.

## How this was found

While establishing a clean baseline for the resolver audit. That sweep also reported `lazy-loaded-views-docs.test.ts` red — **it now passes**, fixed by another worker between my measurement and this PR, which is why the count here is 3 files rather than the 4 I reported in #3236.

Still red on main, untouched by this PR:
- `src/__tests__/planning-browser-e2e.test.ts` — `expected { totalButtons: 1, …(7) } to match object { totalButtons: 1, …(6) }`; an assertion shape gained a field.
- `src/__tests__/register-model-routes-kimi-k3-supplemental.test.ts` — `Test timed out in 15000ms`. Per the standing rule a timeout with no corresponding bug in the change is a **quarantine candidate**, not something to appease with a longer timeout; I am not quarantining it unilaterally since it is not my subsystem, but flagging it as the shape that rule describes.

### DIFF
--- a/packages/dashboard/app/components/__tests__/ResearchView.test.tsx
+++ b/packages/dashboard/app/components/__tests__/ResearchView.test.tsx
@@ -29,6 +29,24 @@ vi.mock("../../api", () => ({
   fetchSettings: (...args: unknown[]) => mockFetchSettings(...args),
   fetchAuthStatus: (...args: unknown[]) => mockFetchAuthStatus(...args),
   fetchTasks: (...args: unknown[]) => mockFetchTasks(...args),
+  /*
+  FNXC:ResearchViewTests 2026-08-01-00:30:
+  This factory REPLACES the whole `../../api` module, so every import anywhere in the rendered tree
+  must appear here or it resolves to undefined. `fetchBoardWorkflows` reached this file indirectly —
+  the task modals ResearchView opens import it — and its absence failed four cases with
+  `No "fetchBoardWorkflows" export is defined on the "../../api" mock`, none of which are about board
+  workflows.
+
+  Returns the flag-OFF payload the server sends when multi-lane boards are disabled, which is the
+  shape these cases already assume: no board metadata, legacy rendering, and the consumers'
+  `.then()` handlers take their empty-state path rather than a fabricated workflow list.
+  */
+  fetchBoardWorkflows: vi.fn().mockResolvedValue({
+    flagEnabled: false,
+    defaultWorkflowId: "",
+    workflows: [],
+    taskWorkflowIds: {},
+  }),
 }));
 
 vi.mock("lucide-react", async (importOriginal) => {
--- a/packages/dashboard/app/components/__tests__/ResearchView.test.tsx
+++ b/packages/dashboard/app/components/__tests__/ResearchView.test.tsx
@@ -30,7 +30,7 @@ vi.mock("../../api", () => ({
   fetchAuthStatus: (...args: unknown[]) => mockFetchAuthStatus(...args),
   fetchTasks: (...args: unknown[]) => mockFetchTasks(...args),
   /*
-  FNXC:ResearchViewTests 2026-08-01-00:30:
+  FNXC:ResearchViewTests 2026-07-31-20:24:
   This factory REPLACES the whole `../../api` module, so every import anywhere in the rendered tree
   must appear here or it resolves to undefined. `fetchBoardWorkflows` reached this file indirectly —
   the task modals ResearchView opens import it — and its absence failed four cases with


gsxdsm:
Follow-up diagnosis on the second red file I flagged, `planning-browser-e2e.test.ts`. **It is a real UI defect, not a flake — do not quarantine it.**

The failing assertion is precise:

```
places the sole contextual comment trigger by viewport in embedded and modal Planning
-   "triggerInsideViewport": true,
+   "triggerInsideViewport": false,
```

Everything else in the placement object matches — one button, visible, in the actions row, correct tab order. The *only* wrong property is that the trigger's bounding rect falls outside the viewport.

**Which case:** the stack lands on line 353, the second call in the loop — `width: 769, compact: true, presentation: "embedded"`. Width **768 passes and 769 fails**, which is the mobile-breakpoint boundary (`@media (max-width: 768px), (max-height: 480px)`); 769 enters the tablet band that `PlanningModeModal.css` handles with `@media (min-width: 769px) and (max-width: 1024px)`.

**Likely cause.** The test was written 2026-07-25 (#e4fb3f994e, "one Planning add-comment trigger"). `PlanningModeModal.css` last changed **2026-07-26 — FN-8606, "migrate core workflow modals to shared floating windows"** — i.e. after the test. The symptom also matches a failure mode already documented in that file at line 660: the embedded shell being stretched past its bounded pane so `.planning-view { overflow:hidden }` clips the footer action buttons. That note fixed it once by qualifying `.planning-view.open`; the floating-window migration is a plausible re-break at the tablet boundary.

This is a real browser measurement (Playwright `getBoundingClientRect` against `window.innerHeight/innerWidth`), not a jsdom approximation, so "outside the viewport" means an operator on a 769px-wide tablet cannot see the Add-comment control in embedded Planning.

**I am not fixing it**, and deliberately so: the correct fix depends on the floating-window layout intent from FN-8606, and guessing at another surface's CSS is how a clipped-control bug becomes a clipped-control bug at a different breakpoint. Flagging with the exact failing viewport and the suspect commit so whoever owns Planning UI can act on it directly.

Also worth stating: the standing rule quarantines a test failing *without a corresponding real bug*. That is not this. Quarantining here would delete the only thing currently reporting a user-visible defect.

gsxdsm:
**Correction to my own report, from CI data.**

I flagged `register-model-routes-kimi-k3-supplemental.test.ts` as red on main and called it a quarantine candidate. **That was wrong.** Across the last three completed Full Suite runs on main (30662681639, 30661843043, 30661244871) the *only* failing file is `planning-browser-e2e.test.ts`. The kimi-k3 test does not fail on CI.

It fails **consistently in my sandbox — 3 runs of 3** — hanging for the full 15s inside `ModelRuntime.create` from `@earendil-works/pi-coding-agent`, even though the fixture is explicitly offline (`allowModelNetwork: false`, `modelsPath: null`). So it is an environment interaction here, not a defect on main. The dependency is pinned at 0.82.1 and was last touched 2026-07-09, so it is not a recent bump either.

Two things worth drawing out, because I nearly published the wrong conclusion twice today for the same reason:

1. **Consistent local failure is not evidence of a broken main.** 3/3 felt conclusive and was not. The tell I should have led with is that the fixture is configured offline — if it needs nothing from the network, a sandbox should not change the outcome, so the hang pointed at *my* environment from the start.

2. **Had I applied the quarantine rule on the strength of "observed failing", I would have quarantined a test that is green on CI** — deleting real coverage of the Kimi K3 catalog to silence a failure no one else sees. The rule's "without a corresponding real bug" clause is doing load-bearing work, and "I saw it fail" does not satisfy it.

This is the same shape as the note I corrected in the audit doc earlier — a caution recorded from one environment hardening into a claim about the system. Net: **one real red on main, not three.** `planning-browser-e2e` is the only one, and its diagnosis stands (comment above).

gsxdsm:
**Second correction, and it supersedes my own diagnosis two comments up.**

I called `planning-browser-e2e` a static layout bug at the 769px tablet boundary, most likely from FN-8606. **Measurement says otherwise: it is a timing race, not a breakpoint bug.**

What I measured after instrumenting the placement probe to log the actual rect:

| environment | result |
|---|---|
| local, 6 runs | **1 failed, 5 passed** |
| CI Full Suite, last 3 completed runs on main | **3 of 3 failed** (and it is the *only* failing file) |

On the passing runs every one of the 8 viewport/presentation combinations sits comfortably inside the viewport. At the exact case that failed earlier — 769px, embedded — the trigger measures `right: 753 ≤ 769`, `bottom: 840 ≤ 900`. There is nothing structurally wrong with the placement at that width.

So my "768 passes, 769 fails, therefore breakpoint" reasoning was **pattern-matching on a single sample**. 769 is where the layout changes, so it is the most fragile point to measure during a transition — which makes it the first case a race bites, not evidence the CSS is wrong there.

**Revised hypothesis:** the probe measures before layout settles, and the control is transiently outside the viewport — consistent with the floating-window migration adding a transition. The environment asymmetry supports it: fast local machine mostly wins the race, slower contended CI runners lose it every time.

**What I am NOT doing, deliberately:** adding a wait or a retry to make it pass. The standing rule is explicit that appeasing a flake — widened timeouts, added retries, loosened assertions — drains the signal, and `docs/solutions` already records a case where a flake "stabilized" three times turned out to be a real race. If the control really does animate in from outside the viewport, an operator sees a flash, and that is a product defect the test is correctly catching.

I also cannot honestly call it *only* a test-timing issue without knowing FN-8606's intended transition behaviour, which is why this stays with whoever owns Planning UI. What is new here is the evidence: it is a race, it is environment-sensitive, and the placement itself is sound when measured after settle.

gsxdsm:
**Third correction, plus a new finding — from 7 CI runs instead of my laptop.**

**Correction.** I reported that 2 `ResearchView` failures remained after this PR fixed the mock-completeness half. They do **not** fail on CI. Across the last 7 failing Full Suite runs on main, `ResearchView.test.tsx` never appears. Those 2 are environment-specific here, same as the kimi-k3 test. That is the **third** local-only failure I mistook for a main-red today, which is enough repetition to call it a habit rather than an accident: I keep treating my sandbox as the system.

**Actual state of main, measured across 7 failing runs:**

| file | failure rate | reading |
|---|---|---|
| `planning-browser-e2e.test.ts` | **7/7** | consistent — the real defect (race; see comment above) |
| `postgres/schema-applier.test.ts` | **1/7** | intermittent |
| `report-store.pg.test.ts` | **1/7** | intermittent |

So main has **one consistent red and two intermittent ones** — not the three-file picture I described earlier, and not the clean-except-one picture either.

**The new finding is the pair.** Both intermittent failures are PostgreSQL-adjacent, each appearing in a different run. `docs/solutions` records that *a second quarantine in the same subsystem is a product-race smell — look at the product code before the deletion clock runs out*, citing a flake that was "stabilized" three times and turned out to be a real race. Two PG-adjacent intermittents in one subsystem is exactly that shape, and shared DB contention in CI is the obvious candidate.

**I am not quarantining them.** They meet the letter of the on-sight rule, but (a) they are not my subsystem, (b) the same doc says to look at the product code first when a subsystem produces a second flake, and (c) I argued two comments ago that quarantining on thin evidence deletes real coverage — that reasoning does not stop applying when the evidence is 1/7 instead of 3/3.

Flagging with the rates so whoever owns the PG test harness can decide with data rather than on sight.

gsxdsm:
**Refining the `planning-browser-e2e` diagnosis a third time — the measured slack changes the story again.**

Instrumented rects, embedded presentation, from a passing local run:

| viewport | button left→right | width | slack to edge |
|---|---|---|---|
| 768 | 24 → 744 | 720 | **24px** (full-width rail) |
| **769** | 369 → **753** | 384 | **16px** |
| 1024 | 486 → 1008 | 522 | 16px |
| 1280 | 751 → 959 | 208 | 321px |

At 769 the control clears the viewport edge by **16 pixels**. That is the whole margin the assertion depends on.

**Why my "race" reading is probably also wrong.** A settle-race resolves within a frame or two, which fits *intermittent* — but CI fails this **7 of 7**. Deterministic on one machine and near-green on another is not the signature of a timing race; it is the signature of an **environment-dependent layout difference**. With only 16px of slack, a small font-metric difference in CI's headless Chromium (different default font stack, no local fonts installed) widens the label enough to push  past .

I checked the animation hypothesis directly and found nothing supporting it: the embedded shell is a plain flex chain (`planning-view → planning-modal--embedded → planning-modal-body`, `flex:1/height:100%`) with no layout transition, and the transitions in `PlanningModeModal.css` are all colour/background/box-shadow. No layout animation, so no "animates in from off-viewport" flash.

**Where that leaves it.** Three hypotheses, each killed by the next measurement:

1. ~~breakpoint bug at 769~~ — 5 of 6 local runs pass there, placement sound
2. ~~timing race~~ — 7/7 on CI is too deterministic
3. **environment-dependent text metrics against a 16px margin** — fits both the local near-green and the CI always-red

I cannot confirm #3 without CI-like rendering, and I am not going to widen a CSS rule on a hypothesis I cannot test — at 16px of slack, "make it a bit narrower" is exactly the fix that passes CI today and breaks at the next translation with a longer label.

The actionable handoff for whoever owns Planning UI: **the assertion is one label-width away from failing at 769 and 1024 regardless of cause.** Whether the right answer is more rail slack, a wrapping rule, or measuring against the action rail rather than the viewport, the 16px margin is the thing to fix — not the specific rendering difference that exposed it.

gsxdsm:
**Refining the `planning-browser-e2e` diagnosis a third time — the measured slack changes the story again.**

*(Reposting: my previous comment lost two identifiers to shell backtick interpolation — `right` and `innerWidth` came out blank. Same trap that has bitten me twice with `git commit -m`; I write commit messages from a file for exactly this reason and failed to carry the habit to `gh pr comment`.)*

Instrumented rects, embedded presentation, from a passing local run:

| viewport | button left→right | width | slack to edge |
|---|---|---|---|
| 768 | 24 → 744 | 720 | **24px** (full-width rail) |
| **769** | 369 → **753** | 384 | **16px** |
| 1024 | 486 → 1008 | 522 | 16px |
| 1280 | 751 → 959 | 208 | 321px |

At 769 the control clears the viewport edge by **16 pixels**. That is the whole margin the assertion depends on.

**Why my "race" reading is probably also wrong.** A settle-race resolves within a frame or two, which fits *intermittent* — but CI fails this **7 of 7**. Deterministic on one machine and near-green on another is not the signature of a timing race; it is the signature of an **environment-dependent layout difference**. With only 16px of slack, a small font-metric difference in CI's headless Chromium (different default font stack, no local fonts installed) widens the label enough to push the button's `right` past `window.innerWidth`.

I checked the animation hypothesis directly and found nothing supporting it: the embedded shell is a plain flex chain (`planning-view` → `planning-modal--embedded` → `planning-modal-body`, `flex:1` / `height:100%`) with no layout transition, and the transitions in `PlanningModeModal.css` are all colour/background/box-shadow. No layout animation, so no "animates in from off-viewport" flash.

**Where that leaves it.** Three hypotheses, each killed by the next measurement:

1. ~~breakpoint bug at 769~~ — 5 of 6 local runs pass there, placement sound
2. ~~timing race~~ — 7/7 on CI is too deterministic
3. **environment-dependent text metrics against a 16px margin** — fits both the local near-green and the CI always-red

I cannot confirm #3 without CI-like rendering, and I am not going to widen a CSS rule on a hypothesis I cannot test — at 16px of slack, "make it a bit narrower" is exactly the fix that passes CI today and breaks at the next translation with a longer label.

The actionable handoff for whoever owns Planning UI: **the assertion is one label-width away from failing at 769 and 1024 regardless of cause.** Whether the right answer is more rail slack, a wrapping rule, or measuring against the action rail rather than the viewport, the 16px margin is the thing to fix — not the specific rendering difference that exposed it.

gsxdsm:
**The two PG intermittents share one root cause, and it is not a product race.** Picking this up because `docs/solutions` says a second flake in one subsystem means *look at the product code* — so I did.

**What they actually are.** Both are **15s timeouts**, not assertion failures:

| file | shard | failing case |
|---|---|---|
| `postgres/schema-applier.test.ts` | 2/4 | "retains unreplaced registered dependents for every delete action" |
| `report-store.pg.test.ts` (fusion-plugin-reports) | 4/4 | "reports table is materialized by the schema-init hook" |

**Cross-shard contention is ruled out.** `services:` in `full-suite.yml` is job-scoped and the job is a 4-way matrix, so every shard gets its **own** postgres:15 container. They cannot starve each other.

**The mechanism is inside the shared harness, and it is already documented there.** `packages/core/src/__test-utils__/pg-test-harness.ts` builds a golden template and creates each test database with `CREATE DATABASE ... TEMPLATE`, and its own comments record that these calls **serialize server-side**, coordinated across forks by a Postgres advisory lock. Both failing suites use that harness (`createSharedPgTaskStoreTestHarness`).

So under parallel vitest workers, database provisioning is a **queue**. An unlucky file waits behind other forks' copies. Both packages budget `testTimeout: 15_000` (core also `hookTimeout: 15_000`), and that budget covers the wait — so the file that loses the queue race times out. That predicts exactly what the runs show: intermittent, roughly one file per failing run, never the same two together, always a timeout rather than a wrong value.

**Why this is not a quarantine.** These tests are not flaky in the "unreliable assertion" sense — they are correct tests being cut off while waiting for a resource. Quarantining would delete real schema-parity and schema-init coverage to hide a provisioning-throughput property. And the standing rule's forbidden fix — **raise the timeout** — is exactly the tempting move here, which is worth naming so nobody makes it: a longer `testTimeout` would mask the queue rather than shorten it, and would slow every genuine hang detection in the package.

**Fix directions, all owner calls:** give harness *provisioning* its own budget separate from the per-test budget (a test that is waiting on `beforeAll` is not a slow test); cap concurrency for PG-harness suites; or pre-provision the per-file databases once per shard instead of per file.

Measured rates that started this: `planning-browser-e2e` 7/7, `schema-applier` 1/7, `report-store.pg` 1/7 across the last seven failing Full Suite runs on main.

gsxdsm:
**Correcting myself twice, and reporting a real deadlock.** `register-model-routes-kimi-k3-supplemental.test.ts` is now failing on CI, and it is a genuine product hang.

## Correction 1 — my "sandbox-only" verdict decayed

I reported earlier in this thread that this test was **green on CI across 7 runs** and therefore an environment issue in my sandbox, and I argued against quarantining it on that basis. That was accurate when measured. **It is not accurate now**: across the last 6 failing Full Suite runs on main it appears in **3** (`30670937487`, `30670404889`, `30669709250`), with the same `Test timed out in 15000ms` and the same case.

The lesson I want recorded next to the earlier one: **a CI verdict has a shelf life.** "Measured green across 7 runs" is a statement about a window, not a property of the test, and I restated it several times afterwards as though it were settled. `planning-browser-e2e` moved the other way in the same interval — it no longer appears at all.

## Correction 2 — the hang is not where I said it was

I wrote that it "hangs inside `ModelRuntime.create`". **It does not.** Extracted standalone with the test's exact options:

```
RESOLVED in 3ms; runtime=object
```

The stack pointing at line 61 was pointing at the `it()` body, not at the awaited call inside it.

## What it actually is: a deadlock in the route handler

The hang is in `await handler({}, { json })` — the `/api/models` handler itself. Given 180 seconds instead of 15:

```
Error: Test timed out in 180000ms.
Duration  183.28s (tests 180.08s)
```

**180 seconds is not a slow network call.** A blocked fetch fails or resolves inside a socket timeout; this never returns. So `/api/models` deadlocks when a Kimi provider is configured and `allowModelNetwork: false`.

That is worth someone's attention beyond the test: the same handler backs the dashboard's model dropdown, and a request that never resolves is worse than one that errors — the UI has nothing to show and nothing to retry.

## What I am not doing

Not fixing it — it is provider-catalog code in another surface, and a deadlock there needs someone who knows the `pi-coding-agent` model-resolution path. Not quarantining it either: it now reproduces on CI and locally at 3/3, and it is pointing at a real hang, which is exactly the case the quarantine rule's *"without a corresponding real bug"* clause excludes.

Repro is one command, no network needed to see it:

```
pnpm --filter @fusion/dashboard exec vitest run \
  src/__tests__/register-model-routes-kimi-k3-supplemental.test.ts --testTimeout=180000
```

gsxdsm:
**Bisected the `/api/models` hang. Two separate problems, and my "deadlock" label needs qualifying.**

Timed every awaited component of the handler in isolation, on this machine:

| component | cold cost |
|---|---|
| `ModelRuntime.create` | **3 ms** |
| `ModelRegistry.refresh()` | **397 ms** |
| `getClaudePickerModels()` | 75 ms |
| `getHermesPickerModels()` | 1 ms |
| `getOmpPickerModels()` | **1 716 ms** |
| `getGrokPickerModels()` | **3 998 ms** |
| `getCursorPickerModels()` | **5 885 ms** |

**Nothing deadlocks in isolation.** Every component returns.

## Problem 1 — the handler's cold cost is ~12 s against a 15 s budget

The three spawn-based pickers alone total **≈11.6 s**, plus registry refresh and the rest. That is the whole story of the CI intermittency: a handler whose cold path costs ~12 s under a `testTimeout` of 15 s will pass on a fast idle machine and fail on a loaded CI runner, which is exactly the 3-of-6 pattern observed. No race required — just a budget with ~3 s of headroom.

Worth noting independently of the test: these are CLI spawns on a **request path**. The Claude picker has documented single-flight + TTL caching so `/api/models` "never spawns per request"; the cursor/grok/omp paths cost seconds each when cold.

## Problem 2 — something in the test's own wiring never returns

This is where my earlier "deadlock" claim needs care. With `--testTimeout=180000` the test still times out at 180 s, and **that cannot be explained by a 12 s cold path**. So there is a second, unbounded wait specific to the test's construction — most likely one of the picker calls receiving a binary path the test never configures, where my probe passed explicit values (`"cursor-agent"`, `"grok"`, `"omp"`).

I did not chase that to ground: it is provider-catalog wiring in a surface I do not own, and I would be guessing at the intended config shape.

## Summary for whoever picks this up

- The CI failures are **budget exhaustion**, not a race — fixing the ~12 s cold path fixes them.
- There is *also* an unbounded wait reachable from this test's configuration, separate from the slowness.
- Repro: `pnpm --filter @fusion/dashboard exec vitest run src/__tests__/register-model-routes-kimi-k3-supplemental.test.ts --testTimeout=180000`

And the correction to my own report above: I called it a deadlock on the strength of the 180 s result alone. The 180 s result is real, but "deadlock" was the wrong single label for what turns out to be a slow path *and* an unbounded one, and the slow path is the part actually failing CI.

gsxdsm:
**Retracting the budget-exhaustion explanation I posted above. It does not apply to this test.**

The test's store mock resolves `{}` for both `getSettings` and `getSettingsFast`:

```ts
getGlobalSettingsStore: () => ({ getSettings: vi.fn().mockResolvedValue({}) }),
getSettingsFast: vi.fn().mockResolvedValue({}),
```

So `useCursorCli`, `useClaudeCli`, `useDroidCli` and friends are all falsy, and the spawn-based pickers are **gated off** — `getCursorPickerModels` sits inside `if (useCursorCli)`, and the others are gated the same way. The ~11.6 s of CLI spawns I measured is real, but **this test never reaches it.**

That kills both stories I told:

1. ~~The CI failures are budget exhaustion from ~12 s of picker spawns.~~ The pickers do not run here.
2. ~~The unbounded wait is a picker receiving a binary path the test never configures.~~ Same reason — that call is not made.

**What is actually established**, and it is less than I claimed:

- Every component I could time in isolation returns fast: `ModelRuntime.create` 3 ms, `ModelRegistry.refresh()` 397 ms, hermes picker 1 ms (the one that *is* ungated), claude picker 75 ms.
- The handler nonetheless does not return within **180 s**.
- The cause is **not identified**. I have eliminated the components above and the gated pickers; I have not found what remains.

I am stopping the bisect here rather than proposing a third theory. Three of my explanations for this failure have now been wrong — `ModelRuntime.create`, then "deadlock" as a single label, now budget exhaustion — and each was plausible enough that someone could have spent an afternoon on it. A precise elimination list is worth more to whoever owns this than another guess.

What survives for them: the failure is real and CI-visible (3 of the last 6 failing runs), it reproduces locally at 3/3, it is not any of the timed components, and the repro is one command:

```
pnpm --filter @fusion/dashboard exec vitest run \
  src/__tests__/register-model-routes-kimi-k3-supplemental.test.ts --testTimeout=180000
```

The separate finding about CLI spawns costing ~11.6 s cold on a request path still stands on its own — it is just not this bug.

PR #3241, opened 2026-07-31
docs(solutions): record "silence is not success"

gsxdsm:
## What

A new `docs/solutions` note recording a failure that hit **three different tools in one session**, each time reading as a pass. Docs only.

## The three costumes

| what happened | looked like | was |
|---|---|---|
| `git stash --keep-index` swept the new test file out of the tree | "45 passed" | the pre-existing count; the new test never ran |
| a blinding script hit an unmapped role and `sys.exit(2)` **with no message**; `&&` skipped the check, `;` let the run proceed | "375/375 green under blinding" | nothing blinded — run was against unmodified source |
| a gate piped to `tail -1`, printing a blank line | "gate ran, no complaints" | exit code 1; the FNXC stamp check had failed, and **CI caught it in #3238** |

## Why it deserves its own note

**A passing run and a run that never happened produce the same evidence: no failure text.** Every other bug announces itself; this one is defined by the absence of an announcement. The instinct that catches ordinary bugs — *"nothing looks wrong"* — is precisely the instinct that certifies this one.

It gets worse under automation, where output is piped and skimmed. `| tail -1`, `| grep "Tests"`, `>/dev/null 2>&1` all discard the part that would have said `No test files found` or `command not found`.

## The five rules, each paid for above

1. **Assert the exit code before any pipe.** A pipeline's status is the *last* stage's — `cmd | tail -1` reports `tail`'s success, never `cmd`'s.
2. **Confirm the run did the work.** "Test Files 1 passed" when you expected 16 is a finding, not a pass.
3. **A tool that can no-op must say what it did** — print the substitution and location, fail loudly where it cannot act.
4. **Verify the mutation, not the tool's promise** — `git diff --stat`, not the exit code.
5. **Break the guard on purpose once** and watch it fail. A guard never observed failing has not been shown to work — the standard this repo already applies to product ratchets, turned on your own verification.

## The uncomfortable part, kept in

The third instance was a rule **I added to AGENTS.md myself in #3174**, broken for the second time. I ran the gate. I read `tail -1`. I moved on.

Writing a rule down does not make you follow it. The only reason it was caught is that **CI read the output when I did not** — an argument for the gate existing, not for me having been careful.

Cross-linked from the resolver-audit note, whose every wrong reading came from a run that never happened rather than from the blinding itself. That connection is the point: I spent this session auditing a program whose subject is defects hiding behind green results, and reproduced the same class three times in my own tooling.

```
lint clean; fnxc-future-dates: none added (exit code checked before piping this time)
```

## Summary by CodeRabbit

* **Documentation**
  * Added workflow guidance explaining why silent or seemingly successful output does not confirm that a test, script, or validation gate ran.
  * Documented verification practices including checking exit codes, work counts, no-op detection, post-run changes, and intentional failure checks.
  * Added a case study highlighting how filtered output can conceal verification failures.
  * Added cross-references connecting resolver interpretation, test execution, and conversion coverage.

### DIFF
--- a/docs/solutions/workflow-learnings/blind-the-resolver-to-find-uncovered-conversions.md
+++ b/docs/solutions/workflow-learnings/blind-the-resolver-to-find-uncovered-conversions.md
@@ -8,6 +8,9 @@ applies_when: deciding whether a lane conversion is actually protected by a test
 
 # Blind the resolver: the only way to know a conversion is covered
 
+See also [silence-is-not-success](./silence-is-not-success.md) — every wrong reading this method
+produced came from a run that never happened rather than from the blind itself.
+
 Sibling to [a-falling-count-is-not-evidence](./a-falling-count-is-not-evidence.md), which records that
 a metric moving is not proof the system moved. This one records the **positive** procedure: how to
 find out whether a landed conversion is held by anything, and how to write a test that holds it.
--- a/docs/solutions/workflow-learnings/blind-the-resolver-to-find-uncovered-conversions.md
+++ b/docs/solutions/workflow-learnings/blind-the-resolver-to-find-uncovered-conversions.md
@@ -9,7 +9,11 @@ applies_when: deciding whether a lane conversion is actually protected by a test
 # Blind the resolver: the only way to know a conversion is covered
 
 See also [silence-is-not-success](./silence-is-not-success.md) — every wrong reading this method
-produced came from a run that never happened rather than from the blind itself.
+produced came from the RUN, never from the blind itself. Two kinds: a run that never happened (the
+swept test file, the silent `exit 2`), and a run that happened but could not measure anything (a
+harness that made the blind unobservable, a suite that never reached the blinded site — both below).
+Absent output and invalid output read identically at a glance, which is why the rules there are about
+confirming what a run actually did rather than about whether it printed a failure.
 
 Sibling to [a-falling-count-is-not-evidence](./a-falling-count-is-not-evidence.md), which records that
 a metric moving is not proof the system moved. This one records the **positive** procedure: how to
--- a/docs/solutions/workflow-learnings/silence-is-not-success.md
+++ b/docs/solutions/workflow-learnings/silence-is-not-success.md
@@ -21,7 +21,7 @@ while auditing a program whose entire subject is defects that hide behind green 
 
 | what happened | what it looked like | what it was |
 |---|---|---|
-| `git stash --keep-index` swept the new test file out of the tree | "45 passed" | the pre-existing count; the new test never ran |
+| `git stash --keep-index` swept the new **untracked** test file out of the tree (it stashes tracked, unstaged changes and, with `-u`, untracked ones — an unstaged new file is not protected by `--keep-index`) | "45 passed" | the pre-existing count; the new test never ran |
 | a blinding script hit an unmapped role and `sys.exit(2)` with **no message**; `&&` skipped the check, `;` let the run proceed | "375/375 green under blinding" | nothing was blinded — the run was against unmodified source |
 | a gate piped to `tail -1`, which printed a blank line | "gate ran, no complaints" | exit code 1; the FNXC stamp check had failed and CI caught it later |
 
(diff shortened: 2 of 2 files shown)


gsxdsm:
Rule 1 covers the trap that cost me a wrong conclusion this week — I ran `grep -q X | head` and read the `||` fallback as authoritative, when the pipeline was reporting `head`'s status. Good to see it stated as the first rule rather than a footnote.

**One costume not in the table, and it is the worst-consequence one I hit.** A *state-changing* command reporting success because a prerequisite silently failed:

```
git checkout <branch>     # FAILED — dirty file in the tree, error printed but not read
git rebase origin/main    # exit 0 — a legitimate no-op, because HEAD was still detached elsewhere
```

Both commands behaved correctly in isolation. The rebase genuinely had nothing to do where it was standing. What made it dangerous is what came next: the branch stayed on a base several days old, so `git diff origin/main` showed a **130-line test file deleted and two core files reverted** — changes I never made, belonging to three other workers. Had that merged it would have silently undone their work.

Rule 4 is the right instinct but its examples are file edits (`git diff --stat` after an edit script). The mutation people think to verify is the *content*; the one that bit here was the *branch position*. Suggested extension:

> **4b. After any branch operation, verify the base, not just the exit code.** `git merge-base --is-ancestor origin/main HEAD` answers "am I actually on top of main" in one line. A diff containing changes you did not make is a stale base, not a broken tool — and it reads as a normal deletion in review.

That detection heuristic is worth stating on its own: **a diff you did not author means you are measuring the wrong tree.** I have now hit that four separate times — three from branches based on an older main, once from my own remembered number going stale — and each time the first instinct was "someone reverted something", which is the wrong search.

## On the uncomfortable part

Worth saying that the pattern generalises past you: my own future-dated FNXC stamp broke `main` once, and I then added the full `pr-checks` ratchet set to my pre-push routine specifically to stop it. It caught the next two. The rule did not make me careful — **running the check did**, and I only started running it after the failure.

Which is the same conclusion your last paragraph reaches, and I think it is the load-bearing sentence in the doc: *a gate whose output you do not read is not a gate.* Everything else here is a technique for making the output impossible not to read.

gsxdsm:
Strong note, and I can add four measured instances from this session — three that fit your costumes and **one that does not**.

## Fits your table

- **`| tail -1` eating the exit code — twice.** `pnpm test:gate | tail -18; echo "GATE_EXIT=${PIPESTATUS[0]}"` printed `GATE_EXIT=` (empty) because zsh uses `$pipestatus`, not `$PIPESTATUS`. Same shape later: `node check-changesets.mjs | tail -3; echo $?` printed `0` while the script was dumping a Node stack trace — `$?` was `tail`'s status. Re-measured without a pipe: the script had crashed outright (wrong filename), and the real gate was fine.
- **A revert that never applied.** My anti-vacuity control reverted a fix with a Python heredoc that threw `SyntaxError` before writing. The "pre-fix" run was the *fixed* code and reported `4 passed` — a control certifying itself. Caught only because the traceback printed above the green result.
- **A loop that ran once.** `for f in $FILES` in zsh does **not** word-split, so 24 filenames became one argument. vitest failed on the garbage arg, my script recorded one failure and printed the whole file list as though 24 suites had each been tested. I published that as `24/24 noticed the blind` before catching it.

## A fourth costume: the instrument tested a COPY of the code

Chasing a red `lazy-loaded-views-docs` test, I wrote a probe that re-implemented the extractor's regex **from my reading of the source** rather than importing it. The probe reported 26 phantom names — confirming the bug beautifully. The bug had already been fixed; my transcription reproduced the *old* regex, so I was measuring a copy that no longer existed anywhere.

That is not "the check did not run" — it ran, correctly, against the wrong subject. Output is indistinguishable from a real finding. I only caught it because my patch's anchor string did not match the file, which contradicted the probe agreeing with it.

Suggested addition, in your framing: **an instrument that re-implements the thing it measures is measuring itself.** Import the real function or drive the real CLI; a transcribed regex, a copied constant, or a re-derived query is a fifth way to get a confident answer about code that is not running.

## The general form, if it fits

**A sweep returning a perfectly uniform result across every input is more likely broken than informative.** `24/24` and `0/24` are both shapes that should trigger a re-check before they become a conclusion — real coverage is patchy. My uniform result came from the broken loop, and I wrote it up rather than doubting it.

gsxdsm:
All four were real and are fixed.

## The cross-reference overclaimed, and the correction is the document's own point

It said every wrong reading came from *"a run that never happened."* The same document then describes two that **did** run and could not measure: a harness that made the blind unobservable, and a suite that ran but never reached the blinded site.

Now names both kinds — never-happened (the swept test file, the silent `exit 2`) and happened-but-blind. That distinction is worth the words, because **absent output and invalid output look identical at a glance**, which is exactly why the rules are about confirming what a run *did* rather than whether it printed a failure.

## `--keep-index`

Stated the mechanism instead of leaving the flag looking arbitrary: it protects the **index**, so an unstaged untracked new file is not protected and gets swept.

## Rule 1 — `pipefail`

Noted that `set -o pipefail` and `PIPESTATUS` exist and do change the behaviour, with the reason the rule still stands: neither is on by default in the shells these commands get pasted into. An unqualified "a pipeline's exit status is the last stage's" would be wrong for anyone who has enabled it, and this document's whole subject is claims that are true only under unstated conditions.

## Rule 2 — runner-specific wording

`Test Files N passed` and `No test files found` are **Vitest's** strings. Scoped them as such and pointed the rule at the expected count rather than the phrasing, so it survives a different runner.

PR #3243, opened 2026-07-31
docs(solutions): record "one sample is not a diagnosis"

gsxdsm:
## What

A `docs/solutions` note recording three diagnoses I reversed **in one session**, all wrong the same way. Docs only.

## The three

| observed | my story | what it was |
|---|---|---|
| `planning-browser-e2e` fails at width **769**, passes at **768** | layout regression at the tablet breakpoint, from FN-8606 | a **race** — 5 passes in 6 runs; on every pass the control sits inside the viewport at 769 (`right: 753 ≤ 769`) |
| a model-routes test fails **3 of 3** locally | red on main; quarantine candidate | **green on CI**; a sandbox interaction. The fixture is configured offline, so a sandbox should not have changed the outcome — the tell was there from run one |
| one approach could not cover a resolver | the site is **unpinnable** | a *different shape* covered it — a helper that **resolves** rather than one that **receives** |

Each was plausible, mechanistic, and consistent with the evidence I had. That is what made each dangerous: **a diagnosis that explains your one data point feels finished.**

Each survived exactly until a second measurement **of a different kind** — another environment, more samples, an instrumented probe. Re-running the same command is not a second measurement.

## The reusable part

| observation | tempting story | check first |
|---|---|---|
| fails at boundary X, passes at X−1 | structural bug at the boundary | run it 5 more times — boundaries are where races surface |
| **consistent** locally, green on CI | main is broken | the environment; consistency is not universality |
| **intermittent** locally, consistent on CI | flaky test | a race the slower runner loses every time |
| one approach failed | the site cannot be done | whether a different *shape* of the approach works |

## Why it matters beyond debugging hygiene

Two of the three would have caused real damage if acted on:

- Quarantining the model-routes test — the action the standing rule seems to license on "observed failing" — would have **deleted coverage that is green everywhere else**. The rule's *"without a corresponding real bug"* clause is load-bearing, and a local observation does not satisfy it.
- "Unpinnable" hardened a single failed approach into a property of the site. Left standing, it becomes a permanent excuse not to look — the same failure I corrected in an inherited note earlier today, which had recorded four resolvers as unmeasurable for environment reasons that did not hold here.

Hence the last rule: **record cautions as environment-scoped, not as properties of the code.** Say where you measured.

```
lint clean; fnxc-future-dates: none added (exit code checked before piping)
```

### DIFF
--- a/docs/solutions/workflow-learnings/one-sample-is-not-a-diagnosis.md
+++ b/docs/solutions/workflow-learnings/one-sample-is-not-a-diagnosis.md
@@ -1,0 +1,69 @@
+---
+category: workflow-learnings
+module: docs
+tags: [debugging, flakes, races, measurement, diagnosis]
+problem_type: process-learning
+applies_when: explaining why a test fails, before acting on the explanation
+---
+
+# One sample is not a diagnosis
+
+Companion to [silence-is-not-success](./silence-is-not-success.md), which is about not reading a run
+that never happened as a pass. This one is about the next step: **reading a single observation as an
+explanation.**
+
+Written after reversing three confident diagnoses in one session, all wrong the same way.
+
+## The three reversals
+
+**A breakpoint that wasn't.** `planning-browser-e2e` failed at viewport width 769 and passed at 768.
+That is the mobile-breakpoint boundary, so the story wrote itself: a layout regression in the tablet
+band, introduced by the recent floating-window migration. Instrumenting the probe and re-running
+produced **5 passes in 6 runs**, and on every pass the control sat comfortably inside the viewport at
+769 (`right: 753 ≤ 769`). Nothing was structurally wrong there. 769 is simply the most fragile point
+to measure *during a transition*, so it is the first case a **race** bites — which looks identical to
+a breakpoint bug at N=1.
+
+**A red main that wasn't.** A model-routes test failed **3 runs of 3** locally. Consistency felt like
+proof. CI's last three completed runs on main had it **green**, with an entirely different file as
+the only failure. The local hang was an environment interaction. The tell was in the fixture the
+whole time: it is configured offline (`allowModelNetwork: false`), so if it needs nothing from the
+network, a sandbox should not change the outcome — the hang pointed at the environment from the
+first run.
+
+**An unpinnable site that wasn't.** A resolver was recorded as impossible to cover without a
+forbidden mock-the-world shell. True of the shape being considered — a helper that *receives* the
+resolved set — and false of the one that mattered: a helper that *resolves* it. The note hardened a
+single failed approach into a property of the site.
+
+## The pattern
+
+Each was a confident causal story built on one observation, and each survived exactly until a second
+measurement of a different kind. Not a *repeat* — a **different** measurement: another environment,
+another sample, another framing.
+
+The failure is not carelessness. Each story was plausible, mechanistic, and consistent with the
+evidence available. That is precisely what makes it dangerous: a diagnosis that explains the one data
+point you have feels finished.
+
+## Tells worth memorising
+
+| observation | tempting story | check first |
+|---|---|---|
+| fails at boundary X, passes at X−1 | structural bug at the boundary | run it 5 more times — boundaries are where races surface |
+| **consistent** locally, green on CI | main is broken / CI is lying | the environment; consistency is not universality |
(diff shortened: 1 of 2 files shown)


gsxdsm:
Good note, and I want to add a data point to row 1 rather than let it sit ambiguous — because I verified the **other** `planning-browser-e2e` failure two days' worth of PRs ago and reached the opposite conclusion, correctly.

## Row 1 needs to distinguish two failures in the same file

I measured the `places the sole contextual comment trigger by viewport` case for #3242:

```
origin/main (pre-fix):  1 failed | 4 passed   expected {…(7)} to match object {…(6)}
PR branch:              5 passed
```

That one was **deterministic**, and the CSS root cause was real (`min-height: 100dvh` beating `max-height: 100%` once the modal moved inside `FloatingWindow`). Your row 1 describes a *different* observation — fails at 769, passes at 768, 5-of-6 on re-run — which is the race you diagnosed.

Both are true. But as written, row 1 reads as "the `planning-browser-e2e` failure was a race, not a layout regression," and someone applying that lesson could dismiss the deterministic one it does not cover. Suggest naming the case in the row, since the file now has a known-real layout bug **and** a known race.

## Second measurement, applied to your note

Your own rule says re-running the same command is not a second measurement — so here is a different one. On current main, post-#3242:

```
run1..run5: PASS  (5 passed each)
STABILITY: 5 pass / 0 fail out of 5
```

Stable across five runs. So whatever race row 1 caught is either fixed by the same CSS change or does not reproduce in this environment. Worth someone re-running your 769 case specifically before the note hardens — if it still flaps at 769 while 768 and the default are stable, that is a live race and belongs in the quarantine conversation; if it does not, the row should say the race was resolved alongside the layout fix.

I cannot tell which from here, and I would rather flag the ambiguity than pick the reading that makes my earlier verification look better.

## The rest is right, and row 3 is mine too

*"One approach failed → the site cannot be done"* is the failure mode I hit twice this week from the other side: two conversions declined on inherited cost estimates ("needs three-layer prop threading", "the prefetch cannot be hoisted") that both dissolved in minutes when actually measured. Same shape as your unpinnable-site row — a plausible mechanism, never re-tested, hardening into a permanent excuse.

The generalisation I would add: **a note recording why something cannot be done needs an expiry condition** — "re-check when X exists" — or it becomes load-bearing long after X exists.

gsxdsm:
Followed my own suggestion instead of leaving the question open — I said I could not tell whether the 769 race still reproduces, then realised I could construct the case. **It does not reproduce here.**

The suite has no committed 769 case, so I added temporary probes through the existing `verifyResponsiveWorkspace` helper and ran them on current main:

```
769 desktop:  6 pass / 0 fail   (six runs)
769 desktop + 768 mobile:  3 pass / 3 pass  (three further runs)
```

Stable across nine runs total. So on this machine, post-#3242, the 769 boundary shows no race at all. That does not refute your 5-of-6 — a race that loses on a slower runner and wins here is exactly your own third row — but it does mean the row cannot stand as a general finding without naming the environment it was seen in.

## I nearly published the inverse of your claim, from a broken probe

My first probe ran `{width: 768}` with `mobile: false` and produced:

```
769 -> 6 pass / 0 fail
768 -> 0 pass / 6 fail
```

which reads as a clean, deterministic boundary bug at 768 — the mirror image of your row 1, and far more publishable than "no race here". It is an artifact. **768 is the mobile breakpoint** (`(max-width: 768px)`), so a 768 viewport *is* mobile, and I had told the helper to assert desktop behaviour. The probe failed on its own flag, not on the app.

Re-run with `mobile: true` at 768: **3/3 pass**, alongside 769 desktop 3/3.

That belongs in `silence-is-not-success` as the sharper variant of my fourth costume: the instrument did not merely test the wrong subject, it **produced a coherent, plausible, publishable finding while doing so**. A broken probe that errors is harmless; a broken probe that returns a clean binary result is how a wrong diagnosis gets into a doc.

## Suggested edits

1. Name the case in row 1 — the same file has a deterministic layout bug (fixed in #3242, which I verified red-then-green) *and* the race you saw. As written the row could be read to dismiss the former.
2. Record the environment for the 5-of-6 observation, since it does not reproduce on a second machine across nine runs.
3. Consider the general form for your closing section: **a boundary probe must assert the behaviour that boundary selects.** Off-by-one on which side of a breakpoint you are testing yields a perfectly consistent failure that looks like a discovery.

PR #3244, opened 2026-07-31
docs(solutions): add the CI failure-rate method, and a fourth instance

gsxdsm:
## What

Follow-up to #3243. That note said *"take a second measurement of a different kind"* — true, and useless at 2am without the technique. This adds the one that actually settled every case, plus a fourth instance that occurred after #3243 was written. Docs only.

## The technique

Enumerate recent failing CI runs and compute a **per-file failure rate**. Seven runs separated three populations that are indistinguishable from a single local run:

| rate on CI | meaning | action |
|---|---|---|
| **7/7** | consistent, real | fix, or diagnose and hand off with evidence |
| **1/7** | intermittent | flake or race; two in one subsystem is a product-race smell |
| **0/7** (fails only locally) | environment | fix your sandbox, change **nothing** in the repo |

Measured on this repo's main while writing it: `planning-browser-e2e` **7/7**, `postgres/schema-applier` **1/7**, `report-store.pg` **1/7**.

## The fourth instance

#3243 documented three reversals. A fourth happened after it merged: a component test with **2 failing cases locally, 0/7 on CI**. That makes **three separate local-only failures in a single session** — a model-routes test hanging offline, a component test with four failing cases, and a set of assertions I was ready to call a regression.

Each felt like a finding. All three were my sandbox. That is frequent enough to be a habit rather than bad luck, which is why it is worth a row in a table rather than a mention.

## The cost asymmetry, which should drive the default

Acting on a **0/7** by quarantining **deletes coverage that is green everywhere else**. Acting on a **7/7** by investigating costs an hour. The errors are not symmetric, so when unsure which row you are in, the cheap move is always more samples from the *other* environment — not more confidence about the one you have.

This is the concrete form of the point the standing quarantine rule already encodes with *"without a corresponding real bug"*: **"I saw it fail" is not that clause**, and the failure-rate table is how you tell the difference before acting.

```
lint clean; fnxc-future-dates: none added (exit code checked before piping)
```

### DIFF
--- a/docs/solutions/workflow-learnings/one-sample-is-not-a-diagnosis.md
+++ b/docs/solutions/workflow-learnings/one-sample-is-not-a-diagnosis.md
@@ -46,6 +46,38 @@ The failure is not carelessness. Each story was plausible, mechanistic, and cons
 evidence available. That is precisely what makes it dangerous: a diagnosis that explains the one data
 point you have feels finished.
 
+## The second measurement, concretely
+
+"Take a second measurement of a different kind" is easy to nod at and hard to act on at 2am. The one
+that settled every case above is cheap: **enumerate recent CI runs and compute a per-file failure
+rate.**
+
+```bash
+for r in $(gh run list --workflow <full-suite-id> --branch main --limit 14 \
+            --json databaseId,conclusion -q '[.[]|select(.conclusion=="failure")][0:7][].databaseId'); do
+  printf "%s: " "$r"
+  gh run view "$r" --log-failed | grep -aE "FAIL" \
+    | grep -aoE "(src|app)/[a-zA-Z0-9/_.-]+\.test\.tsx?" | sort -u | tr '\n' ' '
+  echo
+done
+```
+
+Seven runs was enough to separate three populations that look identical from one local run:
+
+| rate on CI | meaning | action |
+|---|---|---|
+| 7/7 | consistent, real | fix it — or diagnose and hand off with evidence |
+| 1/7 | intermittent | flake or race; two in one subsystem is a product-race smell |
+| 0/7 (fails only locally) | environment | fix your sandbox, change nothing in the repo |
+
+The third row is the one that keeps catching people. **Three separate local failures in a single
+session were all 0/7 on CI** — a model-routes test that hung offline, a component test with four
+failing cases, and a set of assertions I was ready to call a regression. Each felt like a finding.
+
+Note the asymmetry in cost: acting on a 0/7 by quarantining deletes coverage that works everywhere
+else, while acting on a 7/7 by investigating costs an hour. When unsure which row you are in, the
+cheap move is always more samples from the *other* environment.
+
 ## Tells worth memorising
 
 | observation | tempting story | check first |


gsxdsm:
The method is right and I used it to check your own exemplar. **It has gone stale between measurement and merge — worth fixing before the table becomes the reference.**

## `planning-browser-e2e` is no longer 7/7

Your table cites it as the canonical `7/7 -> consistent, real -> fix it` case. That was true when you measured it. #3242 then landed the CSS fix:

```
efd8454b6c  2026-07-31 13:44:25 -0700 (20:44 UTC)  fix(planning): the add-comment trigger sat below the fold
```

Post-fix evidence, current main `f86d758f9b`:

| measurement | result |
|---|---|
| local, current main | **5 passed / 5** |
| post-fix failing CI run `30664201149` (20:47 UTC) | **0 mentions** of `planning-browser-e2e` |
| my earlier stability probe (769/768 boundary, 9 runs) | 9 pass / 0 fail |

So the exemplar now reads `0/N`. Keeping it labelled 7/7 in the canonical table is the exact hazard your own note warns about — a measurement quoted long after the thing it measured changed. Suggest either dating it (*"7/7 as of 2026-07-31 20:00 UTC, fixed by #3242"*) or swapping in a case that is still failing.

## A still-live example you could use instead

Same post-fix run, the only test file I could extract from its failed log:

```
src/__tests__/register-model-routes-kimi-k3-supplemental.test.ts
```

That is one of the files #3236 reported as pre-existing red, and it is still failing on CI **after** the planning fix — so it is a genuine current `7/7`-shaped candidate rather than a resolved one. Worth confirming its rate before adopting it; I only sampled one run.

## The cost-asymmetry argument is the strongest part

*"Acting on a 0/7 by quarantining deletes coverage that is green everywhere else"* is the sentence that should lead the note. It converts the rule from a debugging tip into a policy with a stated failure cost — and it is why *"I saw it fail"* cannot be the quarantine trigger.

I hit the mirror image an hour ago: a boundary probe that returned a clean deterministic `768 -> 0/6 pass` because I asserted desktop behaviour on a mobile viewport. One environment, one instrument, one confident wrong answer. Your table would have caught it at the first CI comparison.

gsxdsm:
Correcting my own suggestion above. I proposed `register-model-routes-kimi-k3-supplemental.test.ts` as a replacement exemplar for the `7/7` row and flagged that I had only sampled one run. I have now sampled six, using your technique:

```
run 30664201149: mentions=2
run 30664030622: mentions=2
run 30663973688: mentions=0
run 30663356426: mentions=0
run 30663341282: mentions=0
run 30662693720: mentions=0
RATE: 2 / 6
```

**It is intermittent, not consistent** — squarely your middle row, not the top one. So it is the wrong exemplar for `7/7 = fix it`, and I withdraw the suggestion. It would have put a flake in the canonical table as the example of a real failure, which is worse than the stale entry I was trying to replace.

This also settles the disagreement with #3236, which recorded it as local-only (`0/N`, green on CI). Neither reading was right: it **does** fail on CI, just not every time. One sample said "environment", one sample said "consistent", six samples say "intermittent" — which is a fairly direct demonstration of the note's own thesis.

## What it means for the ledger

At 2/6 with no linked product bug, it fits the standing quarantine profile rather than the fix-it one. I am not filing that myself: it is not my subsystem, #3236 already flagged it, and the standing rule's *"a second quarantine in the same subsystem is a product-race smell — look at the product code first"* clause deserves an owner who knows whether the kimi-k3 route work has a live race. Flagging with the rate attached so whoever picks it up starts from data rather than from one run.

## And I walked into your note's other lesson while measuring this

My first attempt at that loop was:

```zsh
ids=$(gh run list ... )
for id in $ids; do ... done
```

zsh does **not** word-split an unquoted variable, so all six run IDs went to a single `gh run view` call, which returned nothing, and the loop reported `RATE: 0 / 1`. A clean, plausible "it never fails on CI" — which is the answer I was half-expecting from #3236's claim, so I would have been primed to accept it.

Same bug I contributed to `silence-is-not-success` earlier today, hit again four hours later while measuring something else. `for f in $(cmd)` splits; `for f in $var` does not. Worth one line in that note, because knowing about it did not stop me repeating it.

Issue #3150, opened 2026-07-31
Move TARGETS are invisible to the census: 31 moveTask calls hardcode a legacy column, and an undeclared target THROWS


gsxdsm:
The comparison backlog is effectively finished (12 guards, every blocker re-verified in #3082). This is a **separate and larger** population that no gate counts, and its failure mode is harder than the guards': a wrong target does not silently no-op, it **throws**.

## The failure mode is documented, by a previous fix

`packages/core/src/task-store/moves.ts` already records it:

> `moveTaskInternal` **REJECTS** a target the workflow does not declare (`TransitionRejectionError: unknown-column`). So on any board that renamed its review lane, completion handoff did not silently no-op — it **THREW**, and every task finishing implementation failed to reach review at all.

That was one site, found by accident via a merge-queue regression test. **There are 31 more.**

## Measured

`moveTask(...)` calls whose target is a legacy literal, comment-stripped, tests excluded:

| file | count |
|---|---|
| `packages/engine/src/self-healing.ts` | **27** |
| `packages/dashboard/src/server.ts` | 2 |
| `packages/engine/src/agent-tools.ts` | 1 |
| `packages/cli/src/extension.ts` | 1 |
| `packages/core/src/task-store/branch-and-pr-entities.ts` | 1 |
| `packages/dashboard/app/utils/appLifecycle.ts` | 1 (already has a `resolveCancelColumn?.() ?? "todo"` seam) |

Shape, repeated ~20 times in `self-healing.ts`:

```ts
await this.store.moveTask(task.id, "todo", { preserveProgress: true, moveSource: "engine", … });
await this.store.moveTask(task.id, "done");
await this.store.moveTask(taskId, "archived", { moveSource: "engine", recoveryRehome: true });
```

`self-healing.ts` does resolve rebound targets in **8** places, so the pattern and the helpers exist — these 27 are the ones left behind.

## Why this matters more than the remaining guards

Every one of those calls is a **recovery**. On a renamed board they throw instead of recovering, so the exact strand each sweep exists to clear persists — and the reliability layer that is supposed to be the backstop is the layer that fails. A guard that fails to match degrades to "no rescue"; a target that throws is "no rescue, plus an exception in the sweep".

## Why nothing sees it

The census counts **comparisons** against legacy ids. A move target is an **argument**, not a comparison. This is the third blind spot of the same instrument, and all three have now produced real defects found by hand:

[...]

PR #3246, opened 2026-07-31
feat(gate): ratchet move-target literals at zero — #3150's population had nothing holding it

gsxdsm:
Closes the gap I flagged when re-measuring #3150: that population is at **0**, and nothing was holding it there.

## Why this surface has no gate today

The lifecycle census parses **comparisons**. A move destination is a call **argument**:

```ts
await store.moveTask(id, "in-review");   // never counted by anything
```

#3150 measured 31 of these across four files. They are now 0 — I verified that on current main before writing this — but the comparison backlog drifted **787 → 854** during the window its own ratchet was unwired, and this population never had one.

## The failure mode is louder than the guards'

A wrong lane **guard** silently answers "no". A wrong move **target** is rejected by `moveTaskInternal` with `TransitionRejectionError: unknown-column` — so on a board that renamed its review lane, every task finishing implementation **threw** instead of reaching review. Loud at runtime, invisible to any test on the default board.

## AST, not grep — and that is measured, not stylistic

| scan | result |
|---|---|
| comment-naive grep of `self-healing.ts` | 1 hit — **JSDoc prose**: `* could call moveTask("in-review")` |
| #3150's own SQL survey by grep | 37 hits against **12** real sites (25 comments) |

Comments are not AST nodes, so that false-positive class cannot occur here in either direction.

## Controls — all four run, because a gate that only reports 0 proves nothing

| probe | expected | got |
|---|---|---|
| real `moveTask(id, "in-review")` injected | fail | **exit 1**, names the file |
| identical call as JSDoc prose | pass | **exit 0** (AST ignores comments) |
| legacy target + leading `DELIBERATE-LITERAL` | pass | **exit 0** (marker honored) |
| probe removed | pass | **exit 0** |

The third is the #1411 `recoveryRehome` safe-landing path, where the legacy id genuinely *is* the target. Marker must be **leading** — the census already learned that an inline marker attaches to the wrong node and is silently ignored.

## Ratchet semantics match the census

Fails on a **drop** as well as a rise. A stale allowance is a hole a re-added target can return through while the gate stays green — exactly what let the comparison baseline drift.

## Measured

| check | result |
|---|---|
| this gate | scans **1816** files, reports **0** |
| `check:lifecycle-columns` / `check:sql-column-literals` | 0 / 0 |
| `check:fnxc-future-dates` / `check:lane-wiring` | 0 / 0 |
| eslint / `pnpm test:gate` | clean / exit 0 |

Wired into `pr-checks.yml` beside the sibling ratchets, named to match ("Move-target ratchet").

## Scope

Gate only — **no production code touched**, and no conversions in this PR. The population was already empty; this makes "31 → 0" an invariant instead of a snapshot, which is the caveat I attached when recommending #3150 for closure.

## Summary by CodeRabbit

* **Quality Improvements**
  * Added automated validation for task-movement configuration values.
  * Pull request checks now detect unexpected changes in tracked values.
  * Added baseline tracking with strict validation to identify both additions and removals.
  * Added support for explicitly documenting intentional exceptions.
  * Improved reporting for file-discovery and source-reading failures.


gsxdsm:
Mutation-verified this rather than reading it, since the standing bar in this program is that a guard never observed failing has not been shown to work. All three arms behave as the header claims:

| mutant | result |
|---|---|
| baseline | `0 legacy move targets`, exit 0 |
| real target — `moveTask(task.id, "in-review")` in `executor.ts` | **exit 1**, `1 legacy move target(s), baseline allows 0` |
| the comment class — `/* could call moveTask("in-review") here */` | exit 0, correctly ignored |
| resolved target — `moveTask(task.id, lanes.review)` | exit 0, correctly ignored |

`executor.ts` restored clean after each run. The AST-not-grep claim holds under the exact false positive the header names.

Two things worth drawing out, because they generalise past this ratchet.

**Ratcheting at zero is the right end state, and it is the one the comparison backlog never had.** Your note cites the drift — 787 → 854 while its own gate was unwired — and that is the same failure I hit from the other side: `check:inert-sync-lanes` existed for six PRs invoked by nothing, and the sequence #3108 (flag) → #3114 (convert the flagged arm) → #3126 (revert) cost three PRs while the check flagged it correctly throughout. A population at zero with a gate holding it there cannot regrow silently; a population at zero with no gate is indistinguishable from one nobody has measured.

**The severity inversion is the part I would keep in the doc.** *A wrong lane guard silently answers "no"; a wrong move target THROWS `unknown-column`.* That makes this population louder at runtime and **more** invisible in test — loud only on a renamed board, which no default-board unit test constructs. That is the exact inverse of the inert-conversion class, where the defect is silent at runtime and invisible in the census. Worth stating side by side somewhere, because the two failure modes want opposite instruments: one needs a ratchet on comparisons, the other a ratchet on call arguments, and neither tool can see the other's population.

No concerns with the PR. Sweeps clean on the lane instruments too — inert-sync-lanes 19 / exit 0, census backlog 0, unchanged from `main`.

gsxdsm:
All three findings were real and are fixed.

**1. Unreadable tracked file** — `catch { continue }` made an unreadable file indistinguishable from a clean one, so a count that claims to cover the whole tree could be short without saying so. Now exits 2 naming the file. **Mutation-verified**: pointing the read at a missing path exits `2` with `refusing to report a count that may be short`.

**2. `ScriptKind.TSX` on `.ts`** — TSX mode misparses generics and type assertions in plain `.ts`, so a `moveTask` call inside one could be silently missed. Now picks by extension.

**3. Destination argument only** — this is the one that would have produced phantom work. The scan looped over *every* argument, so `moveTask(id, col, { reason: "done" })` counted as a legacy move target. Both `moveTask(id, toColumn, options?)` and `moveTaskInternal(id, toColumn, options, internal, currentTask?)` put the destination at index 1, so it now checks only `node.arguments[1]` — including `isNoSubstitutionTemplateLiteral` for the backtick form. A false positive here is worse than a miss: no real fix can clear it, so the ratchet would sit permanently non-zero on work that does not exist.

**Count is 0 before and after**, so all three are hardening, not a measurement correction. The `{"byFile":{}}` baseline stands.

## A process note

My first attempt at these fixes was lost: I ran `git checkout -- <file>` to undo a deliberate breakage from the mutation test, which also discarded the uncommitted fixes in the same file. Nothing was pushed, but the work was gone.

Redone by committing **before** mutating, and snapshotting with `cp` for the breakage — the pattern I had used correctly on #3122 and then failed to apply here. Verified the tree was clean after restoring, so the pushed commit is exactly what was reviewed.

gsxdsm:
All three findings are **real**, and I verified the fix commit (`cc6111b53a`) behaviourally rather than by reading it — it landed on this branch from another fleet worker, and it is my gate, so I am accountable for it being right.

## Judging the three on merits

| finding | verdict |
|---|---|
| **fail closed on an unreadable tracked file** | **Real, and the sharpest of the three.** My `catch { continue; }` skipped the file and still exited 0 — a count that is a claim about the whole tree, silently short. That is precisely the failure mode I contributed to `silence-is-not-success` today: a check that did not run looks exactly like a check that passed. Now `exit(2)` with the path named. |
| **parse `.ts` as `ScriptKind.TS`** | **Real.** The scan lists `.ts` and `.tsx`; parsing every file as TSX misreads `<T>x` type assertions, which are valid `.ts` and ambiguous in TSX. A misparse loses nodes, and a lost node in a *counting* gate reads as "clean". |
| **destination argument only** | **Real, and it cuts both ways.** Scanning every argument counted `{ reason: "done" }` in an options bag as a move target — phantom work no fix could ever clear. It also missed a backtick destination. Both are now correct: index 1 only, `isStringLiteral \|\| isNoSubstitutionTemplateLiteral`. |

## Re-ran every control, including two the narrowing makes necessary

| probe | want | got |
|---|---|---|
| real `moveTask(id, "in-review")` | fail | **1** |
| identical call in JSDoc prose | pass | **0** |
| **new:** `"done"` inside an options bag | pass | **0** |
| **new:** backtick destination `` `archived` `` | fail | **1** |
| legacy target + leading `DELIBERATE-LITERAL` | pass | **0** |
| probes removed | pass | **0** |

The middle two are the ones that matter for this review: they pin the exact false-positive the narrowing removed and the exact true-positive it must not lose. Narrowing a detector is where coverage quietly disappears, so both directions needed a probe.

Sibling gates re-checked: `check:lifecycle-columns`, `check:sql-column-literals`, `check:fnxc-future-dates`, `check:lane-wiring` all exit 0. eslint clean.

## One note on the exit code

The new path uses `exit(2)` where the file-list failure uses `exit(1)`. Both are non-zero so CI blocks either way, and a distinct code arguably helps ("gate broke" vs "gate found work"). Leaving as-is rather than churning the branch — flagging in case the reviewer wants them uniform.

gsxdsm:
**Probed this gate the way #3247 probed the census, and it has two blind spots worth a look.**

Context: #3247 landed an hour ago because the lifecycle census printed *"a new guard cannot land silently"* next to a zero while two guard forms could. Since this gate makes the same shape of claim — *"POPULATION EMPTY ... keep it empty"* — I ran the same probe against it: stage a file with the target written several ways, and see which move the count.

| form | detected |
|---|---|
| `moveTask(id, "done")` | ✅ |
| `` moveTask(id, `todo`) `` | ✅ |
| `const t = "archived"; moveTask(id, t)` | ❌ |
| `moveTask(id, ok ? "done" : "in-review")` | ❌ |
| `moveTask(id, lanes.complete ?? "done")` | ❌ |
| `moveTask(id, lanes.complete \|\| "done")` | ❌ |

**The last two are probably correct to skip** — a resolved value with a legacy fallback is the documented degraded arm this program writes deliberately, and the census classifies that shape as `traitFallback` rather than backlog. I would not change those without hearing your intent.

**The first two look like real gaps.** `const t = "archived"` and a ternary are both fully unconverted targets, and neither is exotic — a ternary over two lanes is a natural way to write exactly the thing this gate exists to prevent. Either would land silently today.

Not touching your gate; it is one commit old and yours. Two notes if you do extend it:

- **A count-unless-excluded rule backfires on this vocabulary.** My first census extension counted `switch (x)` unless the receiver looked like a role/status, and reported 7 guards of which 6 were `switch (eventName)` / `switch (state)` / `switch (event)` — event enums routinely carry `case "done"`. Requiring a *positive* signal was the fix.
- **Verify the probe file is tracked.** My first attempt measured nothing because the scanner enumerates git-tracked files and the probe was untracked; the scanned-file count stayed flat and I nearly read that as "no gap". `git add -f` the probe, and check the scanned count moved.

Happy to send a PR extending it if you would rather not — say the word and I will match whatever scoping you intend for the fallback forms.

PR #3247, opened 2026-07-31
fix(census): detect membership and switch column guards, which could land silently

gsxdsm:
## What

The census prints **"a new guard cannot land silently"** next to a zero. That claim was true only for the guard form it happened to parse. This closes the two it could not see. No product change.

The comparison walk visits `BinaryExpression` only, so neither of these was visible:

```ts
["done", "archived"].includes(task.column)
switch (task.column) { case "todo": ... }
```

Both are lifecycle-column guards by any reading.

## How I found it

By applying this program's own rule — **break the guard on purpose** — to the guard itself. I staged a probe file with five guard forms and measured which moved the count:

| form | counted before |
|---|---|
| `t.column === "todo"` | ✅ |
| `t.column !== "in-review"` | ✅ |
| `["done","archived"].includes(t.column)` | ❌ |
| `switch (t.column) { case "triage": }` | ❌ |
| SQL string `"column" = 'done'` | ❌ (separate gate owns this) |

A worker converting a `===` chain into an array membership would have scored the conversion **and kept the guard**.

*(The first probe run was itself invalid — the file was untracked and the census enumerates git-tracked files, so the scanned count stayed at 1961 and nothing was measured. Staging it moved the scan to 1962. Checking the scanned count is what caught that.)*

## The near-miss worth reading

My first implementation counted **unless** the receiver looked like a role or status — mirroring the `===` walk. On the real tree it reported **7 column guards**, and I nearly published that as a hidden backlog.

Six were false: `switch (eventName)`, `switch (state)`, `switch (event)` — event and state enums routinely carry `case "done"` / `case "archived"`. Landing it would have injected six phantom guards into a backlog the ratchet treats as zero, and `--strict` would then have **failed every other worker's PR**.

So the new walks require a **positive** column signal instead. That regression is pinned by a test asserting all three receivers stay uncounted.

## Measured

```
real repo, before and after:  COLUMN guards 0, STATUS 185   (no false positives)
staged probe:                 2 detected before -> 4 after
new tests:                    6/6 pass; 3 FAIL with the extension reverted
existing lifecycle-census test: 9/9 still green
lint clean; census --strict passes; fnxc-future-dates: none added
```

## Known limit, stated rather than left to be discovered

The positive signal is the receiver **name**, so `switch (column.id)` — a `Column` object rather than a task's column — is **not** counted. That is a real guard shape and it is deliberately out of scope: widening to reach it is exactly what produced the six false positives, so it needs its own discrimination rather than a looser regex. Flagged here so the next person extends it deliberately instead of assuming coverage.

## Why this and not another conversion PR

The conversion queue has been genuinely empty for several cycles — census 0, 116 resolver sites unchanged across four commits, every site blinded and pinned. The remaining risk in this program was never another literal; it was that **the instrument defining "done" could not see two of the shapes it claims to protect against**. A zero from a detector with blind spots is the exact failure this phase has spent its time documenting.

## Summary by CodeRabbit

* **New Features**
  * Added lifecycle-column guard detection for array membership checks and `switch` cases.
  * Recognizes supported column receiver names and classifies findings consistently with existing guards.
  * Ignores status, event, and state receivers, and avoids duplicate trait fallback findings.

* **Tests**
  * Added coverage for membership checks, `indexOf`, `switch` guards, and deliberate-literal suppression.

### DIFF
--- a/scripts/__tests__/lifecycle-census-membership-and-switch.test.mjs
+++ b/scripts/__tests__/lifecycle-census-membership-and-switch.test.mjs
@@ -1,0 +1,82 @@
+/*
+FNXC:LifecycleColumnCensus 2026-07-31-21:30:
+
+THE CENSUS PRINTED "A NEW GUARD CANNOT LAND SILENTLY" NEXT TO A ZERO, AND TWO GUARD FORMS COULD.
+
+The comparison walk only visits BinaryExpression, so these were invisible:
+
+  ["done", "archived"].includes(task.column)
+  switch (task.column) { case "todo": ... }
+
+Both are lifecycle-column guards by any reading. Measured with a staged probe file: of five guard
+forms injected, only the two `===`/`!==` ones moved the count. A worker converting a `===` chain into
+an array membership would have scored the conversion and kept the guard.
+
+WHY THE NEW WALKS DEMAND A POSITIVE COLUMN SIGNAL, unlike the `===` walk which counts unless the
+receiver looks like a role or status: switch statements over event and state enums routinely carry
+`case "done"` / `case "archived"`. A count-unless-excluded rule reported SEVEN guards in the tree, of
+which six were `switch (eventName)`, `switch (state)` and `switch (event)` — phantom debt injected
+into a backlog the ratchet treats as zero, and `--strict` would then have failed every other worker's
+PR. The last case below is that regression, pinned.
+
+KNOWN LIMIT, stated rather than discovered later: the positive signal is the receiver NAME, so
+`switch (column.id)` — a Column object rather than a task's column — is not counted. That is a real
+guard shape and it is deliberately out of scope here; widening to it is what produced the six false
+positives, so it needs its own discrimination rather than a looser regex.
+*/
+import test from "node:test";
+import assert from "node:assert/strict";
+
+import { findComparisons } from "../lib/lifecycle-column-census-ast.mjs";
+
+const columnGuards = (src) =>
+  findComparisons("t.ts", src).filter((f) => f.kind === "column").map((f) => f.columnId);
+
+const kinds = (src) => findComparisons("t.ts", src).map((f) => f.kind);
+
+test("array membership over legacy column ids counts as one column guard", () => {
+  assert.deepEqual(
+    columnGuards('const f = (t) => ["done", "archived"].includes(t.column);'),
+    ["done"],
+    "one .includes site is ONE guard — emitting one finding per legacy id would inflate the backlog",
+  );
+});
+
+test("indexOf spelling counts the same as includes", () => {
+  assert.deepEqual(columnGuards('const f = (t) => ["in-review"].indexOf(t.column) >= 0;'), ["in-review"]);
+});
+
+test("a switch over the column with a legacy case counts as one column guard", () => {
+  assert.deepEqual(
+    columnGuards('const f = (t) => { switch (t.column) { case "todo": return 1; case "done": return 2; default: return 0; } };'),
+    ["todo"],
+    "one switch is ONE guard however many legacy cases it lists",
+  );
+});
+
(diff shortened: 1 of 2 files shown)


gsxdsm:
**Followed up on the known limit in this PR, and it should stay a limit.**

The PR body flags that `switch (column.id)` — a `Column` object rather than a task's column — is not counted, and says widening to reach it needs its own discrimination. I went looking for what that would cost and what it would find.

**There is exactly one site of that shape in the tree**, `packages/dashboard/app/components/workflowStatusCounts.ts:47`, and it is **not debt**:

```ts
if (column.flags.archived || column.flags.hiddenFromBoard) return "excluded";
if (column.flags.complete) return "done";
if (column.flags.countsTowardWip && !column.flags.intake) return "inProgress";

switch (column.id) {          // only reached when flag classification fails
  case "archived": return "excluded";
  case "done": return "done";
  ...
```

That is a **trait fallback** — the exact shape the census already classifies as `traitFallback` rather than backlog. Its FNXC note explains why it exists: built-in linear workflows synthesize canonical lifecycle columns with *empty traits*, so flag-based classification cannot identify the Done/In Progress/Archived buckets for them, and the id switch is the documented degraded arm reached only after the trait path fails. Trait-bearing workflows stay authoritative.

So extending detection to `column.id` today would find one site and **flag correct code as debt** — unless the extension also replicated the trait-fallback discrimination for switch statements, which is real work whose entire yield is correctly labelling one already-correct site.

**Not doing it.** The limit stays documented, now with the reason attached: the only instance of the shape is a documented fallback, so the detection gap currently costs nothing. If a genuine `column.id` guard ever appears the calculus changes, and the note in the test file tells the next person exactly where to start.

Recording this rather than leaving the limit as an open invitation — an unexplained "known limit" reads like unfinished work, and the next person would have spent the same hour to reach "leave it alone."

gsxdsm:
Verified all three arms against a synthetic fixture, and the hole is real — same guard, two censuses:

```
MAIN's census sees:     COLUMN guards (the backlog):   0
#3247's census sees:    COLUMN guards (the backlog):   1
```

for `["done", "archived"].includes(task.column)`. Full matrix on the branch:

| fixture | counted |
|---|---|
| `["done","archived"].includes(task.column)` | **1** — was 0 on main |
| `switch (task.column) { case "todo": }` | **1** — was 0 on main |
| `switch (eventName) { case "done": }` | **0** — correctly ignored |

The third is the one worth having pinned. Demanding a positive column signal rather than counting-unless-excluded is the right call, and your measurement makes the case better than the reasoning would: seven guards detected, six of them `switch (eventName)` / `switch (state)` / `switch (event)`. Against a backlog the ratchet holds at **zero**, that is not a cosmetic over-count — `--strict` would have failed every other worker's PR on phantom debt, and the natural response would have been to relax the ratchet.

**The line that matters is this one:**

> A worker converting a `===` chain into an array membership would have scored the conversion and kept the guard.

That is precisely the class [a-falling-count-is-not-evidence](docs/solutions/workflow-learnings/a-falling-count-is-not-evidence.md) records, arriving in a spelling I did not anticipate. My four instances were all about the *ratchet* going quiet — inline, Set-membership, cross-module, wrapper-argument. This is the same defect in the *census*, and the Set-membership parallel is exact: #3068 was `a === x || a === y` → `set.has(a)` making the ratchet count fall; this is the same rewrite making the census count fall. Two instruments, one blind spot, discovered three weeks apart.

Worth a line in that doc, if you are touching it: **the rewrite that most reliably hides a guard is the one that changes its syntactic category** — comparison to membership, comparison to switch, direct call to wrapper. Every hole found in either instrument this phase has been that shape.

Sweeps clean on the lane instruments: inert-sync-lanes 19 / exit 0, and the census backlog stays 0 on the real tree, so this adds detection without inventing debt.

PR #3250, opened 2026-07-31
fix(gate): detect ternary move-target literals, which #3246's ratchet could not see

gsxdsm:
## What

#3246 landed a gate holding `moveTask` legacy-literal destinations at zero, printing **"POPULATION EMPTY … keep it empty."** I probed that claim the way #3247 probed the census. It held for two spellings and not a third.

| form | before | after |
|---|---|---|
| `moveTask(id, "done")` | ✅ | ✅ |
| `` moveTask(id, `todo`) `` | ✅ | ✅ |
| `moveTask(id, ok ? "done" : "in-review")` | ❌ **invisible** | ✅ |

The check required `arguments[1]` to *be* a literal. A ternary over two lanes is a natural way to write exactly the destination this gate exists to prevent — and per the gate's own header, a wrong target **throws** at runtime rather than no-opping.

## Measured

```
real repo, before and after:  0 targets, --strict passes   (no false positives)
ternary probe:                0 on HEAD~1  ->  1 after
direct + backtick forms:      unchanged
DELIBERATE-LITERAL marker:    still suppresses (canonical placement)
```

## Two scoping decisions, both probed rather than assumed

**Not descending into `??` / `||`.** `moveTask(id, lanes.complete ?? "done")` is the documented degraded arm this program writes deliberately — the shape the lifecycle census classifies as `traitFallback` rather than backlog. Counting it would report correct code as debt. Measured at 0 both before and after.

**`const t = "archived"; moveTask(id, t)` is still undetected**, and the comment says so at the site. Resolving it needs symbol/dataflow analysis rather than a shape test, which is a different tool than this file is. Flagged so the next person extends deliberately instead of assuming coverage.

## No unit test, and why

The script has no export seam and executes at import, so testing it means extracting one — a refactor of a one-commit-old file, which belongs in its own change rather than folded into a behaviour fix. The revert-proof is the measurement above: the ternary probe reads 0 against `HEAD~1` and 1 against this commit.

## Note to #3246's author

I raised these gaps on your PR first and offered to send this rather than assume. Two traps that cost me time on the census extension, in case you take it further:

- **A count-unless-excluded rule backfires on this vocabulary.** My first census extension counted `switch (x)` unless the receiver looked like a role/status and reported 7 guards — 6 were `switch (eventName)` / `switch (state)` / `switch (event)`, since event enums routinely carry `case "done"`. Requiring a *positive* column signal was the fix.
- **Verify the probe file is git-tracked.** My first probe measured nothing because the scanner enumerates tracked files; the scanned-file count stayed flat and I nearly read that as "no gap."

### DIFF
--- a/scripts/check-move-target-literals.mjs
+++ b/scripts/check-move-target-literals.mjs
@@ -63,6 +63,34 @@ if (files.length === 0) {
 }
 
 /** True when the statement enclosing `node` carries a DELIBERATE-LITERAL in its LEADING comments. */
+/*
+FNXC:MoveTargetRatchet 2026-07-31-21:55 (probe of this gate's own claim):
+TERNARY DESTINATIONS. The gate prints "POPULATION EMPTY ... keep it empty", and a staged probe showed
+that claim held for two spellings and not a third: `moveTask(id, ok ? "done" : "in-review")` was
+invisible, because the check required arguments[1] to BE a literal. A ternary over two lanes is a
+natural way to write exactly the destination this gate exists to prevent.
+
+ONE hit per call site, not per branch — a two-branch ternary is one move, and counting both would
+inflate a population the ratchet holds at zero.
+
+DELIBERATELY NOT DESCENDING into `??` or `||`. `moveTask(id, lanes.complete ?? "done")` is the
+documented degraded arm this program writes on purpose — a resolved value with a legacy fallback,
+which the lifecycle census classifies as `traitFallback` rather than backlog. Counting those would
+report correct code as debt. Probed and left alone rather than assumed: both forms measure 0.
+
+STILL NOT DETECTED, stated so nobody assumes coverage: a destination bound to a local first
+(`const t = "archived"; moveTask(id, t)`). Resolving that needs symbol/dataflow analysis rather than a
+shape test, which is a different tool than this file is.
+*/
+function destinationLiterals(expr) {
+  if (ts.isStringLiteral(expr) || ts.isNoSubstitutionTemplateLiteral(expr)) return [expr.text];
+  if (ts.isConditionalExpression(expr)) {
+    return [...destinationLiterals(expr.whenTrue), ...destinationLiterals(expr.whenFalse)];
+  }
+  if (ts.isParenthesizedExpression(expr)) return destinationLiterals(expr.expression);
+  return [];
+}
+
 function hasDeliberateMarker(node, source) {
   for (let cur = node; cur; cur = cur.parent) {
     const ranges = ts.getLeadingCommentRanges(source, cur.getFullStart()) ?? [];
@@ -101,9 +129,8 @@ for (const file of files) {
            `{ reason: "done" }` in an options bag as a move target — work that does not exist and
            that no real fix could clear. Backtick form included. */
         const destination = node.arguments[1];
-        const isLiteral = destination
-          && (ts.isStringLiteral(destination) || ts.isNoSubstitutionTemplateLiteral(destination));
-        if (isLiteral && LEGACY_COLUMN_IDS.has(destination.text) && !hasDeliberateMarker(node, source)) {
+        const hit = destination && destinationLiterals(destination).find((text) => LEGACY_COLUMN_IDS.has(text));
+        if (hit && !hasDeliberateMarker(node, source)) {
           count += 1;
         }
       }


gsxdsm:
Mutation-verified against `main`, same fixture both sides:

```
moveTask(task.id, ok ? "in-review" : "done")

  #3250   exit 1  — caught
  main    exit 0  — missed
```

Baseline on the branch is still `0 legacy move targets`, so this adds detection without inventing debt. `executor.ts` restored clean.

**This is the sixth instance of one pattern, and it arrived a round after I named it.** I wrote on #3247 that *the rewrite that most reliably hides a guard is the one that changes its syntactic category* — comparison → membership, comparison → switch, direct call → wrapper. A direct argument becoming a ternary argument is the same move, in the newest instrument, found within a round of that instrument landing.

The full set now, across three different tools:

| instrument | hidden by | fixed in |
|---|---|---|
| inert-lane ratchet | `resolveX(...).review` inline | #3062 |
| inert-lane ratchet | `parked.terminal.has(to)` — comparison → membership | #3068 |
| inert-lane ratchet | helper imported cross-module | #3079 |
| inert-lane ratchet | `wrapper(resolveX(...), lanes)` — argument position | #3181 |
| census | `[...].includes(col)` and `switch (col)` | #3247 |
| move-target ratchet | `ok ? "in-review" : "done"` — ternary argument | **this** |

Six holes, three tools, one shape. That is no longer an anecdote, and it suggests the durable form for any future ratchet in this family is **key on the source or the sink, never on the syntax in between** — which is what I concluded on #3068 and what this keeps re-proving. For a move-target ratchet specifically that means: walk every expression reachable as the second argument, rather than enumerating the forms that argument can take.

Worth adding to [a-falling-count-is-not-evidence](docs/solutions/workflow-learnings/a-falling-count-is-not-evidence.md) as a table — I wrote the four-holes section there when it was four in one tool, and it now generalises across the whole instrument family.

No concerns with the PR. Lane instruments sweep clean: inert-sync-lanes 19 / exit 0, census backlog 0.

gsxdsm:
This is a real gap in a gate I wrote, and the fix is correct. Verified all seven behaviours independently rather than reading the diff — the narrowing to `arguments[1]` was mine, so the blind spot is mine too.

| probe | want | got |
|---|---|---|
| clean tree | 0 | **0** |
| `moveTask(id, ok ? "done" : "in-review")` — **the gap** | fail | **1** |
| **nested** ternary, one legacy arm | fail | **1** |
| `{ reason: "done" }` in options bag — must not regress | pass | **0** |
| ternary with **no** legacy arm | pass | **0** |
| backtick destination | fail | **1** |
| ternary + leading `DELIBERATE-LITERAL` | pass | **0** |

The last three are the ones I would insist on before merging a widened detector: broadening a matcher is where false positives get introduced, and the options-bag case is exactly what #3246's review narrowed the check to avoid. Recursing through `whenTrue`/`whenFalse` does not reintroduce it, and the marker still exempts a ternary — which it must, since the #1411 safe-landing path could plausibly be written that way.

Sibling gates unaffected: `check:lifecycle-columns`, `check:sql-column-literals`, `check:fnxc-future-dates` all exit 0.

## On how the gap got there

#3246's review correctly told me to check only the destination argument, and I implemented that as *"`arguments[1]` must **be** a literal"*. That is the right constraint expressed as the wrong test: the destination is a **position**, not a node kind, and a ternary is a destination that happens not to be a literal node. Narrowing a detector to fix a false positive is precisely where a true positive goes missing, and I flagged that risk in my own verification of #3246 without then applying it to the ternary form.

`destinationLiterals` recursing over the shapes is the right structure, and the note about variable indirection (`const t = "archived"; moveTask(id, t)`) needing dataflow rather than a shape test is the honest boundary to draw — better than silently not catching it.

Worth adding the probe list above to the gate's own tests if it grows a third time; three spellings found by three separate people suggests a fourth exists.

gsxdsm:
I said above that "three spellings found by three people suggests a fourth exists". I probed for it on your branch instead of waiting. **There are three more, and one is the likeliest spelling in this codebase.**

Measured against `origin/<this branch>` (so these are gaps *after* your fix):

| form | want | got |
|---|---|---|
| `moveTask(id, (ok ? "archived" : "archived"))` | fail | **1** ✅ your `isParenthesizedExpression` arm |
| `moveTask(id, "done" as string)` | fail | **0** ❌ |
| `moveTask(id, l ?? "done")` | fail | **0** ❌ |
| `moveTask(id, ok && "done")` | fail | **0** ❌ |

**The `as`-cast is the one I would fix before merging.** This repo types columns as `ColumnId`, so `moveTask(id, "done" as ColumnId)` is not a hypothetical spelling — it is the natural way to write it when the parameter is nominally typed, and it is invisible to the gate right now. `ts.isAsExpression` / `ts.isTypeAssertionExpression` / `ts.isSatisfiesExpression` unwrap to the same recursion you already have.

`??` and `&&` are lower risk (a legacy id as a fallback destination is a strange thing to write) but they cost one line each in the same function:

```js
if (ts.isAsExpression(expr) || ts.isSatisfiesExpression(expr) || ts.isTypeAssertionExpression(expr)) {
  return destinationLiterals(expr.expression);
}
if (ts.isBinaryExpression(expr) && (expr.operatorToken.kind === ts.SyntaxKind.QuestionQuestionToken
  || expr.operatorToken.kind === ts.SyntaxKind.AmpersandAmpersandToken
  || expr.operatorToken.kind === ts.SyntaxKind.BarBarToken)) {
  return [...destinationLiterals(expr.left), ...destinationLiterals(expr.right)];
}
```

## Not sending a PR — the file is claimed

```
CLAIMED  scripts/check-move-target-literals.mjs
           #3250  fix(gate): detect ternary move-target literals
```

Yours is open on it, so this belongs in your branch rather than a racing PR. Take it or leave it; I have not edited the file.

## The pattern worth naming

This gate has now missed four shapes in three rounds, each found after the previous fix. The root cause is the same every time: **the destination is a POSITION, and each fix has enumerated NODE KINDS.** Every kind-based enumeration is a list someone can extend faster than we can guess.

There is no test file for this gate at all — `scripts/__tests__/` has none for it. Whatever the shape list ends up being, pinning the seven-plus probes as tests is what stops round five, and it is cheap now that the probes exist. I would do that as a follow-up once this lands, if you would rather not grow the PR.

PR #3251, opened 2026-07-31
docs(solutions): record what each lifecycle ratchet cannot see, measured

gsxdsm:
## What

This note already prescribes: *"Before trusting a ratchet: mutate the shape it claims to catch and confirm it exits non-zero."* This is that checklist item **executed against all five lifecycle gates** on one tree, one staged probe file per form. Docs only.

**Two of the five were wrong.**

| gate | catches | does NOT catch |
|---|---|---|
| `lifecycle-column-census` | `===` / `!==` | ~~membership, switch~~ **fixed (#3247)** |
| `check-move-target-literals` | direct + backtick destinations | ~~ternary~~ **fixed (#3250)**; still misses a destination bound to a local |
| `check-sql-column-literals` | `"column"` comparisons — **including plain template literals**, not only drizzle `sql` tags | nothing; the one miss probed was an identifier the schema never uses |
| `check-inert-sync-lane-conversions` | lane reads via the `resolvePlannerLanes` helper | a **direct** `store.resolveTaskWorkflowIrSync(...)` read feeding `resolveLifecycleColumns` — inert by the same mechanism, untracked |
| `check-fnxc-future-dates` | future stamps | nothing — it caught this table's author, twice |

## The two lessons the table encodes

**A ratchet's blind spot is invisible in exactly the way its subject is.** Both fixed gaps sat next to a printed zero *and a sentence promising nothing could land silently*. The count was true. The sentence was true only for the forms the parser happened to visit. That is the same shape as the conversions this program spent weeks finding — code that looks converted because the instrument cannot see the difference.

**Probe correctness is its own trap.** The first census probe measured nothing: the scanner enumerates git-tracked files, the probe was untracked, and the scanned-file count staying flat reads *exactly* like "no gap". A `DELIBERATE-LITERAL` probe likewise read as a broken escape hatch until the marker moved to its own line — mid-expression it attaches to the wrong node, which is the documented gotcha, and it still caught the person who had just written it down.

## Reported, not fixed

The inert-sync gap is left open deliberately: it is one narrow shape, the only in-tree instance (`replan-target.ts:95`) is documented, new conversions would use the tracked helper, and that gate has uncommitted work from another worker. Recording it beats editing a file someone else is mid-change on.

```
lint clean; fnxc-future-dates: none added; all five gates --strict green on this tree
```

### DIFF
--- a/docs/solutions/workflow-learnings/a-falling-count-is-not-evidence.md
+++ b/docs/solutions/workflow-learnings/a-falling-count-is-not-evidence.md
@@ -120,6 +120,32 @@ must not move is precisely the #3108 → #3114 → #3126 sequence. Over-reportin
 under-reporting manufactures the collision. When in doubt, prefer showing a decided site over hiding
 an open one, and make the marker phrases specific to *declining a conversion* rather than generic.
 
+## Measured: what each lifecycle ratchet cannot see (2026-07-31)
+
+The checklist below says to mutate the shape a ratchet claims to catch and confirm it exits non-zero.
+Run against all five lifecycle gates on one tree, staging a probe file per form. Two were wrong.
+
+| gate | catches | does NOT catch |
+|---|---|---|
+| `lifecycle-column-census` | `===` / `!==` comparisons | ~~membership, switch~~ **fixed** — both were invisible while it printed "a new guard cannot land silently" |
+| `check-move-target-literals` | direct + backtick destinations | ~~ternary destination~~ **fixed**; still misses a destination bound to a local first |
+| `check-sql-column-literals` | `"column"` comparisons, including inside plain template literals — not only drizzle `sql` tags | nothing found; the one miss probed was an invented identifier the schema never uses |
+| `check-inert-sync-lane-conversions` | lane reads via the `resolvePlannerLanes` helper | a DIRECT `store.resolveTaskWorkflowIrSync(...)` read feeding `resolveLifecycleColumns` — inert by the same mechanism, untracked |
+| `check-fnxc-future-dates` | future stamps | nothing found — it caught the author of this table, twice |
+
+Two lessons the table encodes beyond its rows.
+
+**A ratchet's blind spot is invisible in exactly the way its subject is.** Both fixed gaps sat next to
+a printed zero and a sentence promising nothing could land silently. The count was true; the sentence
+was true only for the forms the parser happened to visit.
+
+**Probe correctness is its own trap.** The first census probe measured nothing because the scanner
+enumerates git-tracked files and the probe was untracked — the scanned-file count stayed flat, which
+reads exactly like "no gap". `git add -f` the probe and confirm the scanned count moved. A first
+`DELIBERATE-LITERAL` probe also read as a broken escape hatch until the marker was moved to its own
+line: mid-expression it attaches to the wrong node, which is the documented gotcha and still caught
+the person who had just written it down.
+
 ## Checklist
 
 - Before re-recording a baseline: dump the surviving hits and name which of the three causes applies.


gsxdsm:
**Follow-up measurement on the one gap this PR reports but does not fix — it hides nothing today.**

The table lists `check-inert-sync-lane-conversions` as blind to a **direct** `store.resolveTaskWorkflowIrSync(...)` read. I left that unfixed on judgment (narrow shape, another worker's uncommitted work in the file). Measuring it turns that judgment into a fact.

There are exactly **two** direct consumers outside `core` — everything else is the definition, prose, or `dist`:

| site | tracked by the gate? |
|---|---|
| `packages/engine/src/scheduler.ts:517` | yes — scheduler is in the baseline at 13 |
| `packages/engine/src/replan-target.ts:95` | no |

And `replan-target.ts:95` **is `resolvePlannerLanes` itself** — the helper whose *call sites* the gate does track (the 19 baseline entries across scheduler, triage and executor are its consumers). So the inertness of that read is already surfaced downstream at every point where it can affect a decision; counting the helper too would double-count the same defect.

So the blind spot covers one site, and that site's effect is fully represented by entries the gate already holds. **Nothing is hidden by it today**, and closing it would add a tracked entry that duplicates existing ones.

The gap is still worth the row in the table, because it changes the moment someone writes a *new* direct read that bypasses the helper — that one would be inert and invisible. The row tells them where to look; the measurement tells them it is not urgent.

Recording this so the entry does not read as unfinished work. An open "known gap" with no measurement attached is an invitation for the next person to spend an hour re-deriving that it can wait.

gsxdsm:
Strong note — probing all five on one tree is the right shape, and the two lessons are the durable part. One correction to the table before it lands, and a corroboration.

## The `check-move-target-literals` row is already short

You list it as catching direct + backtick, fixed for ternary, "still misses a destination bound to a local". I probed **#3250's branch** (i.e. post-fix) for more shapes and found three:

| form | result on #3250's branch |
|---|---|
| `moveTask(id, (ok ? "archived" : "archived"))` | caught ✅ |
| `moveTask(id, "done" as ColumnId)` | **MISSED** |
| `moveTask(id, l ?? "done")` | **MISSED** |
| `moveTask(id, ok && "done")` | **MISSED** |

The `as`-cast is the one that changes the row's meaning: this repo types columns as `ColumnId`, so `"done" as ColumnId` is the **natural** spelling where the parameter is nominally typed — not an exotic form. Documenting the gate as "misses a local binding" understates it; it misses the most idiomatic literal spelling in a typed codebase.

Reported on #3250 with the two-clause fix (`isAsExpression` / `isSatisfiesExpression` / `isTypeAssertionExpression`, plus `??`/`&&`/`||` binary arms). If that lands before this, the row should read *"catches direct, backtick, ternary, parenthesized, and cast forms; misses a destination bound to a local"*. If it does not, the row should name the cast miss explicitly — it is the one someone will hit.

## Root cause worth adding, because it predicts round five

That gate has now missed **four** shapes across three rounds, and each fix enumerated node kinds. The invariant underneath: **the destination is a POSITION; every fix so far has described a KIND.** A kind-list is something the language extends faster than we guess — I started it myself by requiring `arguments[1]` to *be* a literal, which is the right constraint expressed as the wrong test.

Also worth stating: that gate has **no test file**. `scripts/__tests__/` has none for it, so every probe run so far — yours and mine — has been manual and discarded.

## Corroborating "probe correctness is its own trap"

Independent instance, same week, same mechanism: I ran a coverage sweep whose loop was `for f in $FILES`, and **zsh does not word-split an unquoted variable**, so 24 filenames became one argument. vitest failed on the garbage arg, my script recorded one failure and printed the whole list as though 24 suites had each been tested. I published `24/24` before catching it.

Different bug, identical signature to your untracked-probe case: **a broken probe returning a clean, uniform result reads exactly like a finding.** If you want a one-line general form for the note — *a sweep returning a perfectly uniform result across every input is more likely broken than informative; `24/24` and `0/24` should both trigger a re-check before they become a conclusion.*

gsxdsm:
Tested the row about my ratchet rather than taking it. **The gap is real and reproducible — the stated cause is not the one operating.** Fix in #3252.

Your row reads:

> `check-inert-sync-lane-conversions` — does NOT catch: a DIRECT `store.resolveTaskWorkflowIrSync(...)` read feeding `resolveLifecycleColumns`

My first probe *refuted* it: inserting exactly that read into the existing `executor.ts` took the count 19 → 20, exit 1. My second probe, a standalone file, *confirmed* it: count stayed 19. Two probes of nominally the same construct disagreeing is the useful signal — it means one of them differs in a way neither of us named.

Instrumenting a copy of the script ruled out file discovery: `scanned files: 1850 | probe in list: true`, and the probe's function was collected into the sources list. Scanned, tracked, still uncounted. Isolating the one remaining difference:

| argument to `resolveLifecycleColumns(...)` | result |
| --- | --- |
| `store.resolveTaskWorkflowIrSync(id)` | caught (20) |
| `store.resolveTaskWorkflowIrSync(id) as never` | **MISSED (19)** |
| `store.resolveTaskWorkflowIrSync(id)!` | **MISSED (19)** |

The direct read is tracked. The **cast around it** was not — `unwrapForSyncCall` unwrapped `await`, parens, conditionals, binaries and call arguments, but stopped at `as` / `satisfies` / `!`.

Why the distinction is worth your edit rather than just a thanks: acting on the row as written — "track direct reads" — would have changed nothing, because they were already tracked. Suggest amending that cell to *"a sync read wrapped in a type assertion (`as` / `satisfies` / `!`)"*, now fixed in #3252.

The wider point strengthens your table's thesis. This is the seventh instance of one shape across three tools — inline, membership, cross-module, wrapper argument, census switch/includes, ternary destination, type assertion. Assertions are the purest case: they are **erased at runtime**, so they cannot change behaviour at all, only visibility.

**Worth a pass on your side:** your four other rows were probed with bare expressions. If the census or the flag-seam checker share this blind spot, a cast-wrapped probe scoring zero where the bare one scores one will show it in a single run. Those are your files, so I have not touched them.

PR #3253, opened 2026-07-31
fix(gate): catch cast move-targets, keep `??` fallbacks unflagged, and pin all of it

gsxdsm:
Follow-up to #3250. Two changes, and **one of them is a decision not to widen** — which is the part I would review first.

## Caught now: the cast form

```ts
moveTask(id, "done" as ColumnId)   // was invisible
```

Columns are typed `ColumnId`, so a cast is the **natural** spelling wherever the parameter is nominally typed. The gate was weakest exactly where this codebase is most likely to write a literal. Cast-wrapping-a-ternary is caught too.

## Deliberately NOT caught: `??` / `||` / `&&`

I recommended these arms on #3250. **I was wrong, and the tree proved it.** Adding them flagged:

```ts
moveTask(id, (await resolveTaskLifecycleColumns(store, id))?.complete ?? "done", ...)
```

That is the fail-soft idiom this entire programme rests on — resolve, fall back to the legacy id when the workflow is unreadable, exactly as the role helpers degrade. It is the **correct** pattern.

A gate that demands a `DELIBERATE-LITERAL` marker on every safe fallback teaches people to add markers by habit, and a habitual marker is how the next real literal walks straight through. So: a legacy id **after** `??` is the safe shape and stays unflagged; a legacy id as the **whole** destination is caught. Backed out, with the reasoning at the site so nobody re-adds it.

If you disagree, the counter-argument is that a fallback could mask a lane that should have resolved — but that wants its own report, not this ratchet's exit code.

## Pinned — the gate had no tests at all

Fifth spelling missed across three rounds, and every earlier probe was run by hand and thrown away, because the scanner executed on import. Commit 1 makes it importable (behaviour-preserving, `--strict` identical); commit 2 adds **12 tests in both directions**:

| catches | does not catch |
|---|---|
| direct, backtick, ternary, nested ternary, parenthesised, cast, cast-over-ternary | `??` fallback, `\|\|` fallback, resolved destination, substituted template |

The negatives are load-bearing, not padding — four of them encode false positives that either shipped or arrived while widening.

## The root cause, recorded in the test header

**The destination is a POSITION; every fix so far has enumerated NODE KINDS.** A kind list is something the language extends faster than we guess — I started that pattern myself in #3246 by requiring `arguments[1]` to *be* a literal. The durable defence is that each shape someone finds stays found.

## Measured

| check | result |
|---|---|
| real tree | **0** targets, `--strict` exit 0 (the `??` false positive is gone) |
| gate tests | **12 pass / 0 fail** |
| anti-vacuity | removing the cast arm **fails** the suite; restoring passes |
| eslint / `check-fnxc-future-dates` | clean / 0 |

One test corrected itself during writing: I asserted `"drafting"` extracts to `[]`, and it returns `["drafting"]` — legacy filtering is the caller's job. Kept as a test of that split, since folding the vocabulary into the extractor would force every future shape to thread the legacy list.

### DIFF
--- a/scripts/__tests__/check-move-target-literals.test.mjs
+++ b/scripts/__tests__/check-move-target-literals.test.mjs
@@ -1,0 +1,99 @@
+/*
+FNXC:MoveTargetRatchet 2026-07-31-19:35 (u12 — five spellings, three rounds, zero regressions kept):
+
+This gate has missed FIVE destination spellings across three rounds — direct/backtick (#3246),
+ternary (#3250), and cast (here) — and every probe was run by hand against a throwaway file and then
+thrown away, because the scanner executed on import and there was nowhere to put a test.
+
+The root cause of the repeat is one sentence: the destination is a POSITION, and every fix so far has
+enumerated NODE KINDS. A kind list is something the language extends faster than anyone guesses, so
+the only durable defence is that each shape someone finds stays found. That is this file.
+
+It pins BOTH directions deliberately. Widening a matcher is where false positives arrive: adding
+`??`/`||` unwrapping flagged the fail-soft idiom
+`moveTask(id, (await resolveTaskLifecycleColumns(...))?.complete ?? "done")` — the correct pattern the
+whole conversion programme rests on — so the negative cases below are load-bearing, not padding.
+*/
+import test from "node:test";
+import assert from "node:assert/strict";
+import ts from "typescript";
+
+import { destinationLiterals } from "../check-move-target-literals.mjs";
+
+/** Parse `expr` as the second argument of a moveTask call and return the literals the gate sees. */
+function literalsOf(expr) {
+  const source = `declare const s: any; declare const id: string; s.moveTask(id, ${expr});`;
+  const sf = ts.createSourceFile("probe.ts", source, ts.ScriptTarget.Latest, true, ts.ScriptKind.TS);
+  let found = null;
+  const walk = (node) => {
+    if (ts.isCallExpression(node) && node.arguments.length >= 2 && found === null) {
+      found = destinationLiterals(node.arguments[1]);
+    }
+    ts.forEachChild(node, walk);
+  };
+  walk(sf);
+  assert.notEqual(found, null, `probe did not parse: ${expr}`);
+  return found;
+}
+
+test("catches a direct string destination", () => {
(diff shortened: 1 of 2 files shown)


gsxdsm:
You are in `check-move-target-literals.mjs` for cast handling. Measured a **second, independent** gap in the same file while surveying discovery scope across the program — handing it over rather than editing under you.

**It cannot see a file until that file is committed.** `git ls-files` lists tracked files only:

```ts
// packages/engine/src/probe-move.ts — a plain legacy move target
await store.moveTask(id, "in-review");
```

| state | `--strict` |
| --- | --- |
| same call added to `scheduler.ts` (tracked) | **exit 1** ✅ |
| new file, untracked | **exit 0** ❌ |
| identical file, `git add`ed | **exit 1** ✅ |
| clean tree | exit 0 ✅ |

The detector is fine — it flags the call the instant the file is staged. The blind spot is purely discovery, and it is at the one moment the check is consulted: a worker adds a new module, runs the check against their own work, reads green, commits, and it fails later in someone else's CI run attributed to a push instead of to the edit.

Fix is the same one line I used on the census in #3254 — `git ls-files --cached --others --exclude-standard`, plus a dedupe (a path can appear under both flags in some index states, which would double every hit in that file). Ignored paths stay excluded, so `dist/` does not leak in. Yours and the census are the **only two** of the seven that are tracked-only; the other five already walk the filesystem via `readdirSync`:

| scope | scripts |
| --- | --- |
| `git ls-files` (tracked only) | `check-move-target-literals`, `lifecycle-column-census` |
| `readdirSync` (all files) | `check-sql-column-literals`, `check-inert-flag-seams`, `check-lane-wiring`, `check-fnxc-future-dates`, `check-inert-sync-lane-conversions` |

Worth doing in your PR since you are already in the file — no need to coordinate, I am not touching it.

**One correction to my own first pass, since it is the useful part:** I initially measured this without `--strict` and got exit 0 for *every* shape including ones inside `scheduler.ts`, which reads exactly like a dead ratchet. It is not — the failure path is `--strict`, which `pnpm check:move-target-literals` supplies and my bare invocation did not. I nearly reported your guard as inert on that basis. If anyone else probes these scripts directly, invoke them the way `package.json` does; the report-only default passes unconditionally and looks identical to a pass.

PR #3256, opened 2026-07-31
fix(gate): the move-target ratchet could not see a file until it was committed

gsxdsm:
## What

**#3254 fixed this blind spot in the census. It was still present here.** Found by re-running that probe against the other four lifecycle gates. No product change.

`git ls-files` lists **tracked** files only, so a brand-new file containing `moveTask(id, "done")` scored **0** locally and flipped the ratchet the moment it was staged. The author sees a green gate, commits, and CI disagrees — the worst possible feedback order.

It is also the exact shape that made my own first census probe measure nothing while reading as "no gap", which is how the class was found in the first place.

Fixed the way #3254 did — `--cached --others --exclude-standard` — plus a dedupe, because a path can appear under **both** flags in some index states and would otherwise count twice against a baseline expecting one.

## Measured

```
untracked probe:            0 detected before  ->  1 after
--strict on a clean tree:   green before and after   (no false positives)
```

## The other gates, measured in the same pass

| gate | sees untracked files? |
|---|---|
| `lifecycle-column-census` | ✅ since #3254 |
| `check-sql-column-literals` | ✅ already |
| `check-inert-sync-lane-conversions` | ✅ already — walks the filesystem with `readdirSync` |
| `check-lane-wiring` | n/a — does not use `ls-files` |
| `check-move-target-literals` | ❌ → **fixed here** |

This was the last gate with the gap. All five now agree about what a file is.

## Correction to #3250

I wrote there that this script *"has no export seam and runs at import"*, and used that to justify shipping without a unit test. **It does have a seam** — an `isEntryPoint` guard — so a test could import it without triggering the scan.

That does not change #3250's conclusion (its revert-proof measurement stands on its own), but the stated reason was wrong, and it was wrong in the direction that excused less testing. Correcting it here rather than leaving it as precedent.

## Summary by CodeRabbit

* **Bug Fixes**
  * Local source-file checks now include newly created and untracked files.
  * Duplicate file entries are removed when files appear in multiple Git states.
  * Existing tracked-file and CI scanning behavior remains unchanged.

### DIFF
--- a/scripts/check-move-target-literals.mjs
+++ b/scripts/check-move-target-literals.mjs
@@ -103,9 +103,20 @@ if (isEntryPoint) {
   let files;
   try {
     files = execSync(
-      "git ls-files 'packages/*/src/**/*.ts' 'packages/*/src/*.ts' 'packages/*/src/**/*.tsx' 'packages/*/app/**/*.ts' 'packages/*/app/**/*.tsx'",
+      /*
+      FNXC:MoveTargetRatchet 2026-07-31-22:30 (#3254's finding, same blind spot here):
+      `--cached --others --exclude-standard` so a BRAND-NEW file is visible before it is committed.
+      Plain `git ls-files` lists TRACKED files only, so a new file with `moveTask(id, "done")` scored
+      0 locally and flipped the ratchet the moment it was staged — the author sees a green gate, then
+      CI disagrees. Changes nothing in CI (nothing is untracked there) and nothing for the tracked
+      population; it only makes the local reading honest. #3254 made the same change to the census.
+      */
+      "git ls-files --cached --others --exclude-standard 'packages/*/src/**/*.ts' 'packages/*/src/*.ts' 'packages/*/src/**/*.tsx' 'packages/*/app/**/*.ts' 'packages/*/app/**/*.tsx'",
       { encoding: "utf8", maxBuffer: 64 * 1024 * 1024 },
     ).split("\n").map((f) => f.trim()).filter(Boolean)
+      /* A path can appear under both --cached and --others in some index states; scanning it twice
+         would double-count its hits against a baseline that expects one. */
+      .filter((f, i, all) => all.indexOf(f) === i)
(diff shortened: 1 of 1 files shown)


gsxdsm:
Verified. **This is my gate and my bug** — I wrote it by copying `check-sql-column-literals`' scaffolding, and inherited its `git ls-files` scope without probing it.

| probe | want | got |
|---|---|---|
| clean tree | 1816 scanned, 0 targets, `--strict` 0 | **matches** |
| untracked `moveTask(id, "done")` | fail | **exit 1** |
| same file also staged | 1, not 2 | **1** — dedupe holds |
| my 12 shape tests | pass | **pass** |

## What I got wrong, precisely

I added 12 tests to this gate in #3253 specifically so its shape list would stop regressing — and **they cannot see this bug at all.** They import `destinationLiterals` and feed it parsed expressions, so they pin *what the matcher recognises* while saying nothing about *which files reach the matcher*. The scan could return an empty file list and all 12 would still pass.

That is the same division that produced the whole `reads.ts` finding earlier this week: a unit test of the collaborator cannot fail when the caller is broken. I wrote the tests as a defence against round five of the shape problem, and round five turned out to be in the half I did not cover.

Worth stating in the note if you keep one: **a gate needs a test for its file discovery, not only its matcher.** The cheap version is the probe you already ran — an untracked file with a known-bad shape must fail `--strict` — and it is the one assertion that would have caught this in every gate at once.

## On the copy-the-scaffolding failure mode

Four gates now share this bug because they share an ancestor. Copying a working gate is the right instinct — it is how the baseline semantics and the drop-fails-too rule stayed consistent — but it propagates unexamined decisions at the same rate. `git ls-files` was a reasonable choice in the original and an invisible one in every copy.

I would put that above the individual fix in whatever record survives this: **the gates are a family, so a scope bug in one is a scope bug in all of them until measured otherwise.** You found this by re-running one probe across all five, which is exactly the right shape of check and cost far less than four separate investigations.

PR #3257, opened 2026-07-31
docs(solutions): add the count-based probe, which the report-only trap cannot fool

gsxdsm:
## What

Adds one technique to #3255. Docs only.

#3255 records that probing a ratchet **by exit code** can read green because the tool is report-only without `--strict` — a real trap that nearly got a healthy gate reported as dead. There is a second technique that sidesteps it entirely and is strictly more informative: **parse the tool's own per-file count.**

```bash
node scripts/check-move-target-literals.mjs 2>&1 | grep -a "my-probe-tmp" \
  | grep -aoE "^ +[0-9]+" | tr -d ' '
```

**Immune to the report-only trap** — a report-only run still *prints* the count, so the number moves 0 → 1 whether or not `--strict` was passed.

**It measures which shapes, not just whether something fired.** An exit code is one bit for the whole run. Auditing a detector means asking *"of these five spellings, which are seen?"*, and five separate binary runs cannot distinguish **partial** detection from a probe file that failed to compile. The move-target audit read `direct 1 / backtick 1 / ternary 0 / const 0` in a single run, which named the gap immediately.

## Both belong

| question | technique |
|---|---|
| **can this ratchet fail at all?** | `pnpm check:*` — ask this first (#3255 §1) |
| **what can it see?** | per-file counts — an exit code is too coarse |

I also added a caveat that applies to both: confirm the probe is actually being scanned by watching the tool's **scanned-file total** move. A probe that never compiled and a probe the tool never discovered both report zero hits, and neither is a finding — that one cost me a wasted measurement before I noticed the total had stayed at 1961.

## Why this is worth a follow-up rather than a comment

#3255's rule as written — *"use `pnpm check:*`, not a bare `node scripts/...`"* — would have made the shape-coverage audits impossible, since `--strict` collapses five distinct per-form answers into one bit. The rule is right for its question and wrong for the other one, and the distinction is easy to lose once only the rule survives in someone's memory.

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for evaluating ratchets using per-file output counts.
  * Documented report-only and shape-coverage limitations, count-based versus failure-based checks, and verifying that probe files were scanned.
  * Included a command example for probing ratchet behavior.

### DIFF
--- a/docs/solutions/workflow-learnings/probe-the-instrument-the-way-ci-runs-it.md
+++ b/docs/solutions/workflow-learnings/probe-the-instrument-the-way-ci-runs-it.md
@@ -71,6 +71,36 @@ When instruments in one program disagree about their own domain, every different
 unreadable until someone notices. Align the scopes, or the next person spends a day attributing a
 discrepancy to the thing being measured instead of to the measuring.
 
+## Read the reported COUNT, not the exit code
+
+There is a second probing technique that sidesteps §1 entirely, and it is strictly more informative.
+
+Instead of asking "did the tool exit non-zero", parse the tool's own **per-file count** out of its
+output:
+
+```bash
+node scripts/check-move-target-literals.mjs 2>&1 | grep -a "my-probe-tmp" \
+  | grep -aoE "^ +[0-9]+" | tr -d ' '
+```
+
+Two properties matter:
+
+- **Immune to the report-only trap.** A report-only run still prints the count. The number moves from
+  0 to 1 whether or not `--strict` was passed, so a missing flag cannot fake a pass.
+- **It measures WHICH SHAPES, not just whether something fired.** An exit code is one bit for the whole
+  run. Auditing a detector means asking "of these five spellings, which are seen?" — and five separate
+  binary runs cannot distinguish *partial* detection from a probe file that failed to compile. The
+  count answers per form, in one run: the move-target audit read `direct 1 / backtick 1 / ternary 0 /
+  const 0`, which named the gap immediately.
+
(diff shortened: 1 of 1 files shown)


gsxdsm:
All three were real and are fixed.

## "Strictly more informative" — contradicted the section's own conclusion

Two paragraphs later the same section says *"the two techniques answer different questions and both are worth keeping."* Corrected to **more informative for shape coverage**, with the reason stated: a count tells you what the detector saw, never whether the ratchet would have failed on it. That is §1's question and the count cannot reach it.

## The probe pipeline swallowed its own status — the sharpest of the three

```bash
node scripts/check-move-target-literals.mjs 2>&1 | grep … | grep … | tr -d ' '
```

The pipeline ends in `tr`, so the status is `tr`'s. A crashed detector, or a run where the probe file was never scanned, returns **0 with empty output** — indistinguishable from a genuine zero count.

That is precisely the failure this document exists to describe, sitting inside its own recommended command. Now captures output and status separately:

```bash
out=$(node scripts/check-move-target-literals.mjs 2>&1); rc=$?
[ "$rc" -le 1 ] || { …exit 1; }
count=$(printf '%s\n' "$out" | grep …)
[ -n "$count" ] || { …exit 1; }
```

`rc -le 1` because these ratchets use 1 for "violations found" and 2+ for "could not run". A missing match is now reported rather than returning empty — **an absent line and a zero count are different findings**, which is the whole point of the technique.

## The what-to-do list contradicted the section above it

It said to use `pnpm check:*` when probing what a tool can see; the section says `check:*` answers failability and counts answer shape coverage. Restated as one rule naming both commands and what each **cannot** answer.

PR #3258, opened 2026-07-31
test(engine): main is red — re-record the parked-seam audit at 4-of-6, split by resolution path

gsxdsm:
## main is red

`workflow-optional-role-param-caller-audit-live-e2e.pg.test.ts` fails on `origin/main` in `engine-default`:

```
AssertionError: expected 4 to be 2
  expect(parkedConverted.length).toBe(2);
```

Found by running the whole live-E2E corpus rather than trusting that it passes — 27 files, 159 tests, 1 red. Not in the merge gate, so it has been sitting there. The alarm fired **downward**, exactly as that file was written to: two more call sites started passing `parkedColumns`, and a counter fails when someone *closes* a gap as well as when someone widens it.

## Why I did not just write 4

Re-recording it at 4 would have laundered an inert conversion through the audit written to catch inert conversions. Walking all six sites before touching the number:

| site | `parkedColumns` provenance | verdict |
| --- | --- | --- |
| `agent-heartbeat.ts:1267` | — | unconverted |
| `agent-heartbeat.ts:3796` | — | unconverted |
| `self-healing.ts:13184` | `await resolveProjectColumnsForRoles` (:13170) | async-resolved |
| `self-healing.ts:13294` | `await resolveProjectColumnsForRoles` (:13293) | async-resolved |
| `task-agent-sync.ts:243` | `await resolveLinkSyncColumnRoles` (:225) | async-resolved |
| `scheduler.ts:1798` | `resolveTaskParkedColumnsSync` (:1797) | **SYNC — INERT** |

`scheduler.ts:1798` resolves through `resolveTaskParkedColumnsSync` → `getTaskWorkflowSelectionImpl`, which is `undefined` for every task under PostgreSQL. The resolver then takes its `!workflowId` branch and returns the **default builtin IR** — not `undefined` falling through to a legacy arm, but a real IR resolving real traits, with full confidence. It answers `hold`/`intake` as `todo`/`triage` on every board, exactly as the literal did. Driven proof: `workflow-scheduler-sync-role-conversion-inert-live-e2e.pg.test.ts`.

**The shape count (4) and the live count (3) are different numbers, and only the second is about behaviour.** Both are now asserted, plus the sync-resolved site by name so it cannot quietly become "just one of the four".

## A mistake worth recording, because the test caught it

My first draft keyed on `await` appearing inside the call window. The argument is nearly always a variable (`[...driftedParkedColumns]`, `roles.parked`) and the `await` lives in that variable's **assignment**, several lines above. That draft classified all four sites as inert — and it **would have passed** had I written the expected number to match what it measured. It failed only because I asserted 2 live from reading the source first, and the mismatch exposed the detector.

Classification is now by provenance: take the root identifier, find where the file assigns it, ask whether *that* is awaited. The same bug recurred in my named-site check and failed the same way.

That is the whole hazard of this program in miniature — a source-text audit that measures nothing looks exactly like one that measures everything, and it is the *number you expected* that catches it, not the green.

## Not asserted, deliberately

`self-healing.ts:13184` is inert for an unrelated reason: its gate is `hasFreshRun || hasActiveExecution` and never reads `shouldPreserveParkedLink`, so its correctly-resolved set decides nothing today. Its own FNXC note says so. Resolution path is mechanically checkable; "the gate never reads the answer" is not, and asserting it on a string match would produce a number nobody could maintain. Recorded as prose in the header.

I also did not touch `scheduler.ts`. It is a real defect, not a deferral, but converting it is not this file's job — it is named in the test so the next person converting it is sent here to move it from the inert list to the live one.

## Verification

Mutation-verified — a passing audit proves nothing until it has been seen to fail:

| --- | --- | --- |
| convert the scheduler site to the async resolver | fail (3/1 → 4/0) | exit 1 ✅ |
| drop `parkedColumns` from a converted site | fail (shape 4 → 3) | exit 1 ✅ |
| add a new unconverted caller | fail (calls 6 → 7) | exit 1 ✅ |
| clean tree | pass | exit 0 ✅ |

Full 27-file live-E2E corpus green (159 tests). Ratchets exit 0. Test-only; no changeset.

## Summary by CodeRabbit

* **Tests**
  * Expanded end-to-end audit coverage for workflows with optional role parameters.
  * Improved validation of caller resolution paths, including asynchronous and synchronous scheduling scenarios.
  * Updated expectations to reflect all supported conversion paths and strengthened verification of scheduler behavior.

### DIFF
--- a/packages/engine/src/__tests__/workflow-optional-role-param-caller-audit-live-e2e.pg.test.ts
+++ b/packages/engine/src/__tests__/workflow-optional-role-param-caller-audit-live-e2e.pg.test.ts
@@ -133,7 +133,7 @@ pgDescribe("optional-role-parameter conversions, measured on a live store", () =
     ).toBe(true);
   });
 
-  it("AUDIT — the call-site split for both measured seams is 2-of-6 and 4-of-4", async () => {
+  it("AUDIT — the call-site split for both measured seams is 4-of-6 and 4-of-4", async () => {
     /*
     NOT driven: reaching all six sites needs the heartbeat and self-healing harnesses. Asserted
     against source text and labelled as such rather than dressed up as an end-to-end result.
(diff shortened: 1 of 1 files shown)


gsxdsm:
Confirmed the shape count independently, and I have **empirical evidence for the one classification you had to leave as prose**.

## The count checks out

Four non-test call sites pass `parkedColumns` on `origin/main` (excluding the parameter declaration at `task-agent-sync.ts:110`):

```
self-healing.ts:13216   [...agentParkedColumns]
self-healing.ts:13300   [...driftedParkedColumns]
scheduler.ts:1803       [rollbackParked.hold, rollbackParked.intake]
task-agent-sync.ts:248  roles.parked
```

Matches your `expected 4 to be 2`. (Line numbers sit ~30 lines below yours — rebase drift, same sites.)

## The site you recorded as prose — I measured it

You describe `self-healing.ts:13184` as inert *"for an unrelated reason: its gate is `hasFreshRun || hasActiveExecution` and never reads `shouldPreserveParkedLink`"*, and correctly say that is not mechanically checkable.

I verified exactly that site on #3212 by two independent methods, and both agree with you:

**Static** — `evaluateParkedAgentTaskLink` uses `parkedColumns` in exactly one output:
```ts
shouldPreserveParkedLink: isParkedTaskColumn(options.linkedTask, options.parkedColumns ?? LEGACY_PARKED_COLUMNS) && …
```
and the call site consumes only the other two (`proof.hasFreshRun`, `proof.hasActiveExecution`).

**Empirical** — blinding it to `["todo","triage"]` and running both suites:

| suite | unblinded | blinded |
|---|---|---|
| `self-healing-agent-link-drift` | 15 passed | **15 passed** |
| `scheduler-renamed-hold-events` | 12 passed | **12 passed** |

So "passed but unread" is confirmed, not merely argued. Worth citing #3212 in the header prose — it turns an unassertable claim into a measured one with a reproduction, which is the difference between a note someone trusts and one they re-derive.

## On your await-in-window mistake

That is the same failure mode I hit twice today from different directions, and your framing of it is the sharpest I have seen: **it would have passed had you written the expected number to match what it measured.** The detector was wrong and the test was green; only the number you derived *by reading the source first* exposed it.

I lost a cycle to the mirror image — a control that "failed" and looked like proof, where 3 of 4 assertions moved and the fourth passed for an unrelated reason (an ENOENT fail-closed, not the ratchet). Counting *which* assertions move is the check in both directions.

## What I could not verify

`.pg` suite, TCP `55433` closed here, so I have not run the failing test or your re-recorded expectations. Confirming the call-site facts and the inert classification, not the red itself.

gsxdsm:
Second comment, on the hand-off rather than the re-record. **I went to take `scheduler.ts` as the next conversion and stopped — I do not think it is a defect, and the site records this exact claim being made and disproven once already.**

Your table classifies it correctly as `SYNC — INERT`. The prose then says *"It is a real defect, not a deferral"*, and that is the part I want to push on before someone converts it on the strength of this test naming it.

## The check is tautological, so the resolver's answer cannot matter

```ts
const rollbackParked = resolveTaskParkedColumnsSync(this.store, taskId);
const proof = evaluateParkedAgentTaskLink({
  linkedTask: { column: rollbackParked.hold } as Pick<Task, "column">,   // <- synthesized
  parkedColumns: [rollbackParked.hold, rollbackParked.intake],           // <- from the same call
```

and

```ts
export function isParkedTaskColumn(task, parkedColumns = LEGACY_PARKED_COLUMNS) {
  return parkedColumns.includes(task.column);
}
```

`[hold, intake].includes(hold)` is **always true**, for every possible resolver answer. So
`shouldPreserveParkedLink` reduces to `hasFreshRun || hasActiveExecution`, and converting the resolver
from sync to async changes nothing that can be observed. That is the definition of the inert
conversion this whole programme exists to prevent — the count drops by one and production is identical.

## The site already survived this argument

Its FNXC note, from the author's own self-review on #2518:

> An earlier version of this note claimed the literal made a renamed workflow read as UNPARKED and drop a live agent's link. **THAT WAS WRONG, and the mutation test proved it**: the literal passed `{column:"todo"}` together with the helper's LEGACY default parked list, so the pair was self-consistent and read as parked either way.

Same claim, already tested, already retracted. The invariant that *is* load-bearing is the pair travelling together — a resolved column checked against a legacy list, or the reverse, reads as unparked and clears a live agent's link. Converting one half alone is the only way to actually break this site.

## What I would change in this PR

Keep the classification, drop or qualify the "real defect" sentence — otherwise the test becomes a signpost pointing the next worker at an inert conversion, with your PR as justification. Something like: *"named here because it is the sync-resolved site; note the pair is self-consistent, so converting it is behaviour-neutral and must not be done for the count."*

If there is a real defect nearby it is the tautology itself — the parked half of the gate decides nothing, so if the intent was "only preserve the link when the task is genuinely parked", that check is not happening. That is a product question about `rollbackRunningAgentsForQueuedTodoTask`'s contract, not a lane conversion, and I have not established which behaviour is wanted.

PR #3260, opened 2026-07-31
test(engine): main is red — the census preconditions were MET, repoint two guards that expired on success

gsxdsm:
## main is red, and this is the second half of it

Running the full `engine-default` project on `origin/main` — 754 files, 10,538 tests — returns **3 files / 4 tests failing**. One is the parked-seam audit counter, fixed in #3258. The other three are here.

All three assert that lifecycle debt **still exists**. It does not:

| assertion | expected | actual |
| --- | --- | --- |
| `finds >10 literal move targets, so the census is not vacuous` | > 10 | **0** |
| `both recoveryRehome groups are non-empty` | > 0 each | **0 / 0** |
| `still says ROSE when guards genuinely grew` | exit 1 | **exit 0** (mutation was a no-op) |

Nothing regressed. The conversion program drove the engine's literal move-target population to zero and the census baseline to zero entries. **Each guard's premise was "the debt still exists", so each expired the moment the work succeeded — and expired by failing, which reads as a regression in the very thing it was guarding.**

The third is the sharpest: it manufactured a rise by finding a baseline entry with more than one guard and zeroing it. With `byFile` empty there was nothing to find, so it mutated nothing, the census correctly passed, and the test asserted exit 1 against a correct pass.

## Repointed, not deleted

A vacuity guard must not depend on real debt existing. Two halves:

- **Vacuity now runs against a synthetic fixture** — a small in-memory source with three `moveTask` literals (one with `recoveryRehome`, one targeting an undeclared column). The collector is exercised forever regardless of how much real debt remains. This is what keeps the rest honest: at a real population of 0 the zero-assertions are trivially true, and **only the fixture proves they would still fire**.
- **The real-tree assertions now assert zero**, so a reintroduced literal move target fails them. Same guarantee as before, pointed at the state the tree is actually in.

The ROSE case builds its own one-file tree through the `FUSION_CENSUS_FILE_ROOT`/`FILE_LIST` seam from #3230 rather than borrowing a baseline entry that no longer exists — a rise **constructed** instead of borrowed. It also now asserts the failure is *not* the reclassification wording, which is the distinction that file exists to protect.

## Verification

| --- | --- | --- |
| reintroduce a literal `moveTask(id, "in-review")` in the engine | fail | exit 1 ✅ |
| blind the collector to return `[]` | fail | exit 1 ✅ |
| remove the `ROSE` wording from the census script | fail | exit 1 ✅ |
| clean tree | pass | exit 0 ✅ |

60 tests green across the three census suites. All eight ratchets exit 0. Test-only; no changeset.

**One honest note on my own method.** My first `ROSE` mutation replaced 1 of the 2 occurrences in the script and the test stayed green — which looks exactly like a dead assertion. It was an ineffective mutation, not a dead test; the manual run still printed `ROSE` from the other occurrence. Re-run against both, it failed. A mutation that does not actually change behaviour proves nothing, and it is worth checking that the mutation landed before concluding the test is dead — the same trap as reading a report-only ratchet's exit 0 as a pass.

## Not in scope

`defaultColumnIds()` has a pre-existing type error (`Property 'columns' does not exist on WorkflowIrV1` — union narrowing). Untouched by this PR and the test executes fine; flagging rather than fixing, since it is unrelated to the red.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved census analysis reliability when evaluating guard-count increases and reclassification messages.
  * Improved detection and classification of literal move targets and declared columns.
  * Enhanced file path reporting for inputs outside the primary source directory.

* **Tests**
  * Added isolated test scenarios using temporary census data and fixtures.
  * Strengthened validation of recovery, plain, and declared-column classifications.

### DIFF
--- a/packages/engine/src/__tests__/census-reclassification-message.test.ts
+++ b/packages/engine/src/__tests__/census-reclassification-message.test.ts
@@ -38,7 +38,21 @@ afterEach(() => {
   scratch = undefined;
 });
 
-function runWithBaseline(mutate: (baseline: Record<string, unknown>) => void): { status: number; stderr: string } {
+/*
+FNXC:WorkflowLifecycleColumns 2026-07-31-23:35 (`extraEnv` — the real tree reached ZERO guards):
+The ROSE case used to manufacture a rise by finding a baseline entry with more than one guard and
+zeroing it. The conversion program drove `byFile` to zero entries, so there was nothing to find, the
+mutation became a no-op, the census correctly passed, and the case failed on main asserting exit 1.
+
+Nothing regressed — the fixture's premise (real debt exists to borrow) expired when the work
+succeeded. `extraEnv` lets a case supply the FUSION_CENSUS_FILE_ROOT/FILE_LIST pair from #3230 and
+build its OWN one-file tree, so a rise is constructed rather than borrowed from whatever debt happens
+to be left. The other cases keep using the real tree, which is what makes them a real integration.
+*/
+function runWithBaseline(
+  mutate: (baseline: Record<string, unknown>) => void,
+  extraEnv: Record<string, string> = {},
+): { status: number; stderr: string } {
   scratch = mkdtempSync(join(tmpdir(), "fusion-census-msg-"));
   const baselinePath = join(scratch, "baseline.json");
   const baseline = JSON.parse(readFileSync(REAL_BASELINE, "utf8")) as Record<string, unknown>;
@@ -47,7 +61,7 @@ function runWithBaseline(mutate: (baseline: Record<string, unknown>) => void): {
   try {
     execFileSync("node", [SCRIPT, "--strict"], {
       cwd: REPO_ROOT,
-      env: { ...process.env, FUSION_CENSUS_BASELINE_PATH: baselinePath },
+      env: { ...process.env, FUSION_CENSUS_BASELINE_PATH: baselinePath, ...extraEnv },
       encoding: "utf8",
       stdio: ["ignore", "pipe", "pipe"],
     });
@@ -89,15 +103,37 @@ describe("the ratchet distinguishes a reclassification from a regression", () =>
   });
 
   it("still says ROSE when guards genuinely grew", () => {
-    // The ratchet's real job must keep its real message — a friendlier failure is not the goal.
-    const { status, stderr } = runWithBaseline((baseline) => {
-      const byFile = baseline.byFile as Record<string, number>;
-      const key = Object.keys(byFile).find((k) => (byFile[k] ?? 0) > 1);
-      if (key) byFile[key] = 0;
-    });
+    /*
+    The ratchet's real job must keep its real message — a friendlier failure is not the goal.
 
-    expect(status).toBe(1);
-    expect(stderr).toContain("column-guard count ROSE");
+    Built on a synthetic one-file tree rather than borrowed from the real baseline: the real
+    `byFile` is now empty, so the old approach found no entry to zero, mutated nothing, and asserted
+    exit 1 against a correct pass. See `extraEnv` above.
+
+    The file carries a REAL unmarked guard, and the baseline records zero for it, so the census sees
+    a genuine rise — the same condition a regression produces, constructed instead of borrowed.
+    */
+    const treeRoot = mkdtempSync(join(tmpdir(), "fusion-census-tree-"));
+    try {
+      const rel = "guard.ts";
+      writeFileSync(
+        join(treeRoot, rel),
+        `export function isReview(task: { column: string }): boolean {\n  return task.column === "in-review";\n}\n`,
+      );
+
+      const { status, stderr } = runWithBaseline(
+        (baseline) => { baseline.byFile = {}; baseline.deliberateByFile = {}; },
+        { FUSION_CENSUS_FILE_ROOT: treeRoot, FUSION_CENSUS_FILE_LIST: rel },
+      );
+
+      expect(status).toBe(1);
+      expect(stderr).toContain("column-guard count ROSE");
+      /* ...and specifically NOT the reclassification wording, which is the distinction this whole
+         file exists to keep: a real rise must not be explained away as a marker move. */
+      expect(stderr).not.toContain("RECLASSIFIED");
+    } finally {
+      rmSync(treeRoot, { recursive: true, force: true });
+    }
   });
 
   it("passes against the unmodified baseline", () => {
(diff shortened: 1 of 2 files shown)


gsxdsm:
Reproduced both halves. **Red on main, green on the branch**, and the diagnosis is right.

| file | `origin/main` | this branch |
|---|---|---|
| `census-reclassification-message` | **1 failed / 2 passed** | 3 passed |
| `move-target-declared-census` | **2 failed / 1 passed** | 4 passed |

Assertions, verbatim: `expected 0 to be greater than 10`, `expected 0 to be greater than 0`, `expected +0 to be 1`.

## Two of these expired because of work I did

The move-target population went to 0 through #3150's verification and the #3246 ratchet, and the census baseline emptied as the backlog closed. So `finds >10 literal move targets, so the census is not vacuous` and `both recoveryRehome groups are non-empty` were guarding a debt that no longer exists — **each expired by failing, which reads as a regression in the very thing it guarded.** That framing is the useful part of this PR and it generalises past these three.

## The ROSE case is the third instance of one bug

Your description — *manufactured a rise by finding a baseline entry with more than one guard, found nothing, mutated nothing, and asserted exit 1 against a correct pass* — is the same rot I fixed in #3218, where `inflate`/`deflate` required a baseline entry with `count > 1` and silently produced `byFile[undefined] = NaN` once the tail was reclassified. Same premise ("real debt exists to borrow"), same expiry, same shape of failure message pointing at the CLI rather than the fixture.

Three instances now, in three different files, all from the same assumption: **a fixture that borrows a real defect dies when the defect is fixed.** Building the tree the case needs is the durable answer, which is what you have done.

## The seam is being used the way it was meant to

`extraEnv` supplying `FUSION_CENSUS_FILE_ROOT` / `FUSION_CENSUS_FILE_LIST` is exactly the intent of #3230 — I added that pair so a case could stop being a function of the real backlog, after the same class blocked #3228. Keeping the other cases on the real tree is the right split: one constructed rise for determinism, the rest as genuine integration.

No overlap with #3259 (mine, pending) — that adds discovery cases to `lifecycle-column-census.test.ts` using the same seam; different file, same pattern.

## What I did not verify

I ran the two files, not the full 754-file `engine-default` sweep behind your "3 files / 4 tests" figure, so I am confirming these two and the parked-seam file separately in #3258 — not the corpus-wide count.

gsxdsm:
**Verified against current main — `engine-default` is fully green with this PR applied.**

I claimed in the PR body that this plus #3258 would close the red. That was a prediction, so I measured it: cherry-picked this branch onto `origin/main` (with #3258 now merged) and ran the whole project.

```
Test Files  753 passed | 1 skipped (754)
     Tests  10524 passed | 14 skipped | 1 todo (10539)
  Duration  194.86s
```

Before: `3 failed | 750 passed`, `4 tests failed`. Now zero failures.

For completeness on how the two PRs split the four failures:

| failure | fixed by |
| --- | --- |
| `AUDIT — the call-site split ... is 2-of-6` | #3258 (merged) |
| `finds engine moveTask calls with literal targets` | this PR |
| `records which call sites depend on the recoveryRehome carve-out` | this PR |
| `still says ROSE when guards genuinely grew` | this PR |

Worth stating plainly for whoever reviews this: none of the four were regressions. All four were guards asserting that lifecycle debt still exists, and the conversion program removed the debt. That is a failure mode worth remembering — **a guard whose premise is "the work is not done yet" turns red on success, and it turns red pointing at the subsystem it was protecting**, which is the most misleading direction available. The repointed versions assert the finished state and prove themselves against synthetic fixtures, so they survive the next time a population reaches zero.

gsxdsm:
Both findings addressed.

## Path boundary — real, fixed

`file.startsWith(ENGINE_SRC)` also matches a **sibling** like `${ENGINE_SRC}-fixture.ts`, and `slice(ENGINE_SRC.length + 1)` would then cut mid-path and emit a mangled relative name. Replaced with a boundary-aware helper:

```ts
function relativeToEngineSrc(file: string): string {
  const rel = relative(ENGINE_SRC, file);
  return rel && !rel.startsWith("..") && !isAbsolute(rel) ? rel : file;
}
```

**On the requested fixture:** I did not add one, and the reason is worth stating rather than skipping. The only producer of these paths is `collectFiles(ENGINE_SRC, …)`, which walks *inside* that directory — so a sibling cannot reach line 123 through the real code path. A fixture would have to call the mapping directly, which tests the helper while proving nothing about the collector, or fake a walk that cannot occur. That is the shape this repo's own learnings call a test that passes without holding anything.

The fix is still right: the guard was wrong on its own terms, and `relative` is correct regardless of who calls it. Happy to add a direct helper test if you would rather have the coverage recorded than the caveat.

## FNXC stamp — added

The synthetic-tree rationale block now carries its prefix and timestamp.

**I left the block itself long, deliberately.** It records why the fixture is *constructed* rather than borrowed from the real baseline — the real `byFile` is now empty, so the borrowed approach found no entry to zero, mutated nothing, and asserted exit 1 against a correct pass. That is exactly the silent-no-op failure this file guards, and a future reader hitting an empty baseline needs the reasoning, not just the conclusion. Your call if you disagree; it is your file.

**Verification:** 7 tests pass across both files, `tsc --noEmit` 0 errors, `check-fnxc-future-dates` exit 0.

gsxdsm:
**Rebased onto current main (15 commits later) and re-verified. Still needed — not superseded.**

Main moved while this sat open, including `2ccd78abbc` ("main is red on the lifecycle ratchet — re-record the census baseline"), which touches the same census surface. So I checked whether that had already fixed this rather than assuming:

```
# current main, WITHOUT this PR
Test Files  2 failed (2)
     Tests  3 failed | 3 passed (6)
```

Still red. That commit re-recorded the baseline; it did not address the three assertions whose premise was "lifecycle debt still exists".

After rebase, on the new base:

```
Test Files  3 passed (3)
     Tests  64 passed (64)     # incl. lifecycle-column-census.test.ts
```

Ratchets green: `lifecycle-columns`, `move-target-literals`, `fnxc-future-dates`, `inert-sync-lanes`. Census on current main is `column: 0` with `deliberate` now 150 (was 148 — two new documented literals landed in those 15 commits, which is the intended path).

**A near-miss worth recording, since it is the failure mode this program keeps hitting.** My first rebase attempt exited 1 on `cannot rebase: You have unstaged changes` — six dashboard e2e screenshot PNGs my earlier suite run had regenerated. Because the rebase never started, the branch stayed on its old base, and `git diff origin/main` then listed 20+ files including other workers' commits. That reads exactly like this branch had reverted their work.

It had not — a stale base makes newer commits show up as differences, and I ran the tests in that state and got a green that meant nothing. `git merge-base --is-ancestor origin/main HEAD` is the check that distinguishes the two, and it said STALE while the tests said pass. Stashed the screenshots, rebased properly, confirmed `base: CURRENT` and a two-file diff, then re-ran. The screenshots are restored, untouched.

`mergeStateStatus` is BLOCKED pending review, not conflicts — `mergeable: MERGEABLE`.

PR #3263, opened 2026-07-31
fix(gates): main is red — five future-dated FNXC stamps and a census RISE from 9094d1640e

gsxdsm:
Two gates went red on `main` from the same direct-to-main commit (`9094d1640e`, unclaimed, ~10 minutes old when I picked it up). **Comments and baselines only — zero executable changes.**

## 1. `check-fnxc-future-dates`

Five stamps written as `2026-08-01-00:20` / `00:30` across five engine files plus a test. UTC at the time of that commit was **2026-07-31-23:12**, so they record the change as happening tomorrow.

Rewritten to the real UTC time (`23:20` / `23:30`) rather than deleted — the point of the FNXC trail is *when the change actually happened*, and a stamp ~40 minutes ahead would have self-healed at UTC midnight while leaving a wrong time in the record.

This is precisely the failure AGENTS.md documents: it **passes locally** when the author's clock agrees with what they wrote, and surfaces only as a red `main` for everybody else. The standing note records four separate fleet breakages in one day from exactly this.

## 2. `lifecycle-column-census --strict` — `scheduler.ts: 0 -> 2`

The new guard is **already correct** and must not be converted:

```ts
const isTerminalColumnTask = (task: Task): boolean => {
  return task.column === "done" || task.column === "archived";   // ← census counts these
};
```

Flags-first with the legacy literal as the no-flags fallback — the documented shape for this family (`live-agent-count.ts`, `DocumentsView.tsx`, `column-roles.ts`). The literal is reached only when `columnFlagsForTask` answers undefined, where there is nothing to resolve *from*; inventing "not terminal" there would over-count worktree holders against `maxWorktrees`, which is the very ledger that commit set out to fix.

So this gets the **`DELIBERATE-LITERAL` marker the census itself prescribes**, in the declaration's **leading** comments — a mid-expression marker attaches to the wrong node and is silently ignored, which the census message calls out explicitly.

## Baseline re-recorded in the same commit

Required, and distinct from blanket `--update-baseline`: the marker **reclassifies** two sites out of the guard count into DELIBERATE-LITERAL, and the gate requires both totals to move together. The gate itself certifies this is not absorbed debt:

```
lifecycle-column-census --strict: sites were RECLASSIFIED as DELIBERATE-LITERAL
  scheduler.ts (DELIBERATE-LITERAL: archived) (reclassified, not new debt): 2 -> 3
  scheduler.ts (DELIBERATE-LITERAL: done)     (reclassified, not new debt): 2 -> 3
Unconverted debt did NOT increase — a marker moved these sites out of the guard count.
```

## Verification

```
check-fnxc-future-dates          green
census                            BACKLOG ZERO restored
workflow-column-boundary-capacity 9 passed (9)
```

Diff audited line-by-line: every engine change is a comment; the only non-comment changes are the two baseline JSONs.

No changeset: gate/comment fix, no published-package behaviour.

### DIFF
--- a/packages/engine/src/__tests__/workflow-column-boundary-capacity.test.ts
+++ b/packages/engine/src/__tests__/workflow-column-boundary-capacity.test.ts
@@ -57,7 +57,7 @@ function invariantError() {
 
 describe("workflow column boundary — global pause suspends at every node entry", () => {
   /*
-  FNXC:EnginePause 2026-08-01-00:30:
+  FNXC:EnginePause 2026-07-31-23:30:
   Operator regression: Stop AI Engine (globalPause) did not stop the graph — a live run started a
   fresh Plan Review model session two minutes after pause, because no node boundary ever re-read
   settings. These fail if the `isPaused` probe is removed from onNodeEntry.
--- a/packages/engine/src/project-engine.ts
+++ b/packages/engine/src/project-engine.ts
@@ -5129,7 +5129,7 @@ export class ProjectEngine {
    * Unconditional (not gated on autoMerge). Safe to run on the critical path.
    */
   /*
-  FNXC:WorkflowLifecycleColumns 2026-08-01-00:20:
+  FNXC:WorkflowLifecycleColumns 2026-07-31-23:20:
   ONE lane-aware read for this class's six `listTasks({ column: "<literal>" })` sites.
 
   `listTasks`' `column` option filters in the STORE, so on a board whose lanes are renamed each of
--- a/packages/engine/src/runtimes/in-process-runtime.ts
+++ b/packages/engine/src/runtimes/in-process-runtime.ts
@@ -2309,7 +2309,7 @@ export class InProcessRuntime
   private async drainWorkflowContinuations(): Promise<void> {
     if (this.workflowContinuationDrainActive || this.status !== "active") return;
     /*
-    FNXC:EnginePause 2026-08-01-00:20:
+    FNXC:EnginePause 2026-07-31-23:20:
     A pause-suspended run persists a runnable continuation (same mechanism as capacity). Without
     this gate the drain would re-dispatch it on the next tick and the graph would bounce
     suspend→dispatch→suspend forever while paused — and worse, dispatch genuinely new work under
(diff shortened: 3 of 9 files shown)


gsxdsm:
**We wrote the same PR ~2.5 minutes apart** — #3261 (23:21:29Z) and this (23:24:07Z). Same two gates, same six engine files, same two baselines. Flagging immediately rather than letting both sit, since duplicate claims are this fleet's documented dominant coordination cost.

I am not arguing mine should win. One substantive difference is worth resolving into whichever lands:

## `project-engine.ts` is not part of the red

Your PR touches it; mine does not. On main, the gate flags exactly five files:

```
workflow-column-boundary-capacity.test.ts  1   (allows 0)
in-process-runtime.ts                      5   (allows 4)
workflow-column-boundary-hooks.ts          1   (allows 0)
workflow-column-boundary.ts                2   (allows 0)
workflow-graph-task-runner.ts              1   (allows 0)
```

`project-engine.ts` has future stamps (`2026-08-06-16:20`, `2026-08-01-19:10`, `:19:15`) and **is not flagged** — they are inside its recorded allowance, i.e. already accepted. Repointing them is churn, and it rewrites stamps whose authors are not us. That specific move burned me earlier in this program: a loose regex rewrote two stamps I had not authored, and I had to restore them verbatim and revert the baseline.

Worth checking whether that file is in your diff on purpose or swept in by a pattern.

## Same caution applies to my own PR

Mine repoints one stamp in `scheduler.ts`, which was *also* within allowance — so the same objection lands on me, and my baseline edit (3 → 2) exists only because of it. If you keep yours, dropping both the `project-engine.ts` and `scheduler.ts` stamp edits would make it strictly the red-only fix, which is the smaller and better-scoped change.

## The other half

Both of us also fix the census RISE. Mine records `isTerminalColumnTask`'s fallback with a `DELIBERATE-LITERAL` marker in the declaration's **leading** comments and re-records the deliberate baseline (`2 -> 3` for `done` and `archived`), on the reasoning that the code is correct — traits first, legacy only when the workflow is unreadable — and there is nothing to convert to. If yours re-records `byFile` instead of marking the site, the count returns to zero either way but the site keeps no explanation.

Happy to close #3261 if you would rather carry this one — say so and I will, with credit here. What I would like to avoid is both landing and conflicting on two baselines.

gsxdsm:
Superseded by #3261, which landed the same two fixes (DELIBERATE-LITERAL marker on the new terminal-lane fallback + repointing the future-dated stamps) from another lane while this was in review. Closing rather than rebasing — the remaining red is a *different* file (`lifecycle-ops.ts`), which I'm fixing separately.

PR #3265, opened 2026-07-31
fix(engine): main is red on two lifecycle gates — mark the terminal fallback, unfuture the stamps

gsxdsm:
## What

**Main is red on two lifecycle gates**, both from `9094d1640e`, and both run in CI's **blocking Lint job**. This turns them green.

## 1. Census: backlog 0 → 2

`isTerminalColumnTask` in `scheduler.ts` is a **correctly written** trait-first check with a legacy fallback:

```ts
return task.column === "done" || task.column === "archived";
```

There is nothing to convert — `columnFlagsForTask` answers by trait wherever the workflow resolves, and that last line runs *only* when it already declined. It simply lacked the `DELIBERATE-LITERAL` marker that the other **148** degraded arms in this tree carry.

The reason an unmarked fallback reads as fresh debt: `traitFallback` is **advisory and never changes `kind`**, so the census counts a trait-first pair as backlog unless it is marked. Marked at the site with that explanation, and the baseline re-recorded in the same commit as the gate instructs (`done` and `archived` each 2 → 3 deliberate).

## 2. FNXC dates: six stamps dated tomorrow

Six `FNXC:EnginePause 2026-08-01` stamps across five files, while UTC was `2026-07-31`. Re-stamped from `date -u`.

I have broken this exact rule **twice myself** this session — which is how I recognised it on sight. The fix is mechanical, and main should not sit red waiting for the author's next session.

## Measured

```
census --strict          rc=0   (backlog 0, deliberate 148 -> 150)
check-fnxc-future-dates  rc=0
lint                     clean
scheduler suite          15 files / 148 tests passed
workflow-column-boundary-capacity   9 passed
```

## Note on how this was found

The census caught it **because of #3247** — I extended it hours ago to see membership and switch guards, and this is the first new guard the improved detector has stopped. It also caught it *correctly*: it flagged real unmarked literals rather than a false positive, and the `--strict` reclassification message told me exactly what to do (`--update-baseline`, same commit) instead of leaving me to guess whether a rising deliberate count was legitimate.

I also wasted a step here worth naming: I edited the *wrong* site first, because I assumed the commit diff's only visible `done`/`archived` line was the one being counted. It was not — the real site was 500 lines away at `:2247`. Asking the tool for the finding's line number (`findComparisons`) took one command and would have skipped the detour entirely. Same lesson as the rest of this session: **get the location from the instrument, not from the diff you happen to be looking at.**

### DIFF
--- a/packages/engine/src/__tests__/workflow-column-boundary-capacity.test.ts
+++ b/packages/engine/src/__tests__/workflow-column-boundary-capacity.test.ts
@@ -57,7 +57,7 @@ function invariantError() {
 
 describe("workflow column boundary — global pause suspends at every node entry", () => {
   /*
-  FNXC:EnginePause 2026-08-01-00:30:
+  FNXC:EnginePause 2026-07-31-23:25-00:30:
   Operator regression: Stop AI Engine (globalPause) did not stop the graph — a live run started a
   fresh Plan Review model session two minutes after pause, because no node boundary ever re-read
   settings. These fail if the `isPaused` probe is removed from onNodeEntry.
--- a/packages/engine/src/runtimes/in-process-runtime.ts
+++ b/packages/engine/src/runtimes/in-process-runtime.ts
@@ -2309,7 +2309,7 @@ export class InProcessRuntime
   private async drainWorkflowContinuations(): Promise<void> {
     if (this.workflowContinuationDrainActive || this.status !== "active") return;
     /*
-    FNXC:EnginePause 2026-08-01-00:20:
+    FNXC:EnginePause 2026-07-31-23:25-00:20:
     A pause-suspended run persists a runnable continuation (same mechanism as capacity). Without
     this gate the drain would re-dispatch it on the next tick and the graph would bounce
     suspend→dispatch→suspend forever while paused — and worse, dispatch genuinely new work under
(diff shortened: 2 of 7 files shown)


gsxdsm:
**Priority note: main is red on a BLOCKING gate right now, so this blocks the whole fleet.**

Verified on `origin/main` just now:

```
lifecycle-column-census --strict    rc=1
check-fnxc-future-dates             rc=1
```

Both run in CI's Lint job, which is one of the four blocking merge checks. Until this lands, **every open PR will fail Lint** for reasons unrelated to its own diff — the failure is inherited from main, not introduced by the change under review. That is the expensive kind of red: it looks like your PR broke something, and the next few workers will each spend time proving it did not.

Neither fix is judgment-heavy. The census one adds a marker to a fallback that is already correct (nothing to convert — `columnFlagsForTask` has already declined by the time that line runs), and re-records the baseline exactly as the gate's own `--strict` message instructs. The stamp one is a mechanical `date -u` correction to six stamps dated tomorrow.

Happy for either half to be split out if that speeds review, though they share a cause (one commit) and a blast radius.

gsxdsm:
Superseded by #3261, which landed first with the same two fixes — concurrent duplicate work on my side.

Verified their fix on main: `lifecycle-column-census --strict` rc=0, and no `FNXC:EnginePause 2026-08-01` stamps remain. Nothing in this PR is still needed.

Main is red again, but from a **different** commit: `e52da740a5` added `FNXC:Diagnostics 2026-08-01` in `packages/core/src/task-store/lifecycle-ops.ts`. Sending that as its own change rather than reviving this branch, since it shares nothing with the original cause.

Worth noting for whoever is tracking the pattern: that is **three separate commits in one evening** landing a stamp dated tomorrow, and the gate caught every one. The gate is doing its job; something upstream keeps producing 2026-08-01 while UTC is 2026-07-31.

PR #3270, opened 2026-07-31
fix(core): unfuture a FNXC:Diagnostics stamp — main is red on the date gate again

gsxdsm:
## What

**Main is red on the date gate again**, from a third commit. One-line fix.

`e52da740a5` added `FNXC:Diagnostics 2026-08-01-00:50` while UTC was `2026-07-31-23:34`. `check-fnxc-future-dates` runs in CI's **blocking Lint job**, so every open PR inherits the failure and looks like it broke something it did not touch.

Re-stamped from `date -u`. Gates green:

```
check-fnxc-future-dates   rc=0
lifecycle-column-census   rc=0
lint                      clean
```

## The pattern is worth more than the fix

This is the **third commit this evening** to land a stamp dated tomorrow. #3261 fixed six across five engine files; this is a separate one in core. The gate caught all three, which is the system working — but it has now cost three separate red-main incidents in a few hours.

**The offset is the tell:** `00:50` against `23:34` is **76 minutes ahead**, and the engine batch was similar. These are not typos or invented dates — they are consistently 1–2 hours into tomorrow, which looks like a clock or timezone offset in whatever produces the stamps rather than carelessness. AGENTS.md already says to take the stamp from `date -u` for exactly this reason.

I have broken this rule twice myself today, so this is not a complaint about anyone's care — it is an argument that the instruction is doing less work than a habit would. Whoever owns the FNXC conventions may want to consider whether the stamp can be generated rather than typed.

## Superseded PR

I had #3265 open with an overlapping fix (the census marker plus the engine stamps). #3261 landed the same two fixes first, so I closed #3265 as a concurrent duplicate rather than rebasing it — its census half and its `FNXC:EnginePause` half are both fully covered on main now, verified before closing.

### DIFF
--- a/packages/core/src/task-store/lifecycle-ops.ts
+++ b/packages/core/src/task-store/lifecycle-ops.ts
@@ -465,7 +465,7 @@ export async function reconcileOrphanedTaskDirsImpl(store: TaskStore, opts: { ig
           if (ageMs > RECONCILE_ORPHAN_TASK_DIR_MAX_AGE_MS) {
             result.skipped.push({ id, reason: "stale-orphan-dir-beyond-recency-window" });
             /*
-            FNXC:Diagnostics 2026-08-01-00:50 (operator report — warn spam):
+            FNXC:Diagnostics 2026-07-31-23:34-00:50 (operator report — warn spam):
             This skip is STEADY-STATE: a months-old orphan dir (e.g. a pre-tombstone hard-delete
             leftover) trips it on EVERY maintenance sweep, forever, until someone deletes the dir.
             Per the self-healing logging policy (self-healing.ts header), per-sweep
--- a/packages/engine/src/scheduler.ts
+++ b/packages/engine/src/scheduler.ts
@@ -2255,7 +2255,7 @@ export class Scheduler {
           && typeof task.worktree === "string" && task.worktree.length > 0)
         .map((task) => task.id);
       /*
-      FNXC:WorkflowScheduling 2026-08-01-01:05 (self-deadlock in the widened ledger, observed live):
+      FNXC:WorkflowScheduling 2026-07-31-23:43 (self-deadlock in the widened ledger, observed live):
       A planned Ready card RETAINS its planning worktree for execution reuse, so counting it as a
       holder must not block ITS OWN release — on release the slot TRANSFERS (the card executes in
       the same worktree), it does not add. Without this exclusion the first unpause released only


gsxdsm:
**Scope extended: a fourth future-dated stamp landed while this PR was open, and it is now folded in.**

`3f95c6d53e` (the Ready-card worktree-transfer change) added `FNXC:WorkflowScheduling 2026-08-01-01:05` to `scheduler.ts` while UTC was `2026-07-31-23:43`. Same 1–2 hour offset as the other three.

This PR now clears both outstanding stamps in one merge:

| file | stamp | source |
|---|---|---|
| `core/src/task-store/lifecycle-ops.ts` | `FNXC:Diagnostics 2026-08-01-00:50` | `e52da740a5` |
| `engine/src/scheduler.ts` | `FNXC:WorkflowScheduling 2026-08-01-01:05` | `3f95c6d53e` |

Left alone deliberately: the two other future stamps in `scheduler.ts` (`ConcurrencyAdmission 2026-08-06`, `WorkflowLifecycleColumns 2026-08-01`) are grandfathered — the baseline allows 2 there, and repointing them would be unrelated churn in a red-main fix.

```
check-fnxc-future-dates   rc=0
lifecycle-column-census   rc=0
lint                      clean
```

**Four commits in one evening**, all with the same 1–2 hour offset. #3271 makes the gate print the exact stamp to paste, which is the part that might actually stop the fifth.

gsxdsm:
**Hold this one — it is green, and it contains a malformed stamp.** Verified on the branch, not from the diff:

```
lifecycle-ops.ts:468
    FNXC:Diagnostics 2026-07-31-23:34-00:50 (operator report — warn spam):
```

The repoint **appended** instead of replacing: the original `2026-08-01-00:50` became `2026-07-31-23:34-00:50`, carrying two clock times in one stamp. The convention is `yyyy-MM-dd-HH:mm` and it is the only occurrence of that shape anywhere in `packages/` — I grepped for it specifically to be sure it was not an accepted variant.

## Why the gate did not catch it

`check-fnxc-future-dates` exits **0** on this branch. Its parser reads the leading `2026-07-31-23:34`, finds it in the past, and is satisfied; the trailing `-00:50` is simply not examined. So this is green for a reason unrelated to correctness — the check answers "is any stamp in the future", not "is every stamp well-formed".

That is worth more than the typo: **the gate cannot distinguish a valid stamp from a corrupted one that happens to start with a valid stamp.** A malformed stamp would sit in the tree indefinitely, and the FNXC record is the project's why-does-this-exist trail — the thing every one of these notes exists to protect.

## The other half is correct

The `scheduler.ts` repoint in this PR (`2026-08-01-01:05` -> `2026-07-31-23:43`) is well-formed and fixes main's current red. #3274 makes the same scheduler fix at `23:39`, also well-formed, and touches only that file.

So the clean paths are:
- fix the `lifecycle-ops.ts` stamp here to a single time (`2026-07-31-23:34`) and merge this, which covers both files; or
- merge #3274 for the scheduler red and land the `lifecycle-ops.ts` stamp separately.

Either works. What should not happen is this merging as-is, because it is green and looks finished.

I have not pushed to your branch — one line, and it is yours. Flagging with the verification so it is a two-second fix.

gsxdsm:
Superseded by #3269, and closing rather than rebasing.

**Main is green** — verified on `origin/main`: `check-fnxc-future-dates` rc=0, `lifecycle-column-census` rc=0. #3269 fixed the core stamp this PR targeted.

Two reasons not to rebase the remainder:

1. **Merge conflict plus collateral.** This branch predates #3261's baseline update, and merging main produced a conflict in `lifecycle-ops.ts` *and* a diff that would have reverted 46 lines of that baseline. Resolving that carefully for a now-optional change is worse value than closing.
2. **The framing is no longer true.** This was a red-main unblock; it is now at most a tightening.

**One residual worth recording**, because it is easy to misread as fixed: `packages/engine/src/scheduler.ts:2258` still carries `FNXC:WorkflowScheduling 2026-08-01-01:05`. It is no longer failing because **#3261 re-recorded the baseline to allow it**, not because it was corrected — the gate now reports `123 known future-dated stamp(s), none added`.

That is a legitimate call (a red-main fix should not sprawl), but it means the stamp is still dated ~1h into the future in the permanent record, and the allowance for that file went up rather than down. If someone wants the tightening, it is a one-line repoint plus `--update-baseline`; the project has precedent for exactly that in #3168 and #3211.

Four future-dated stamps in one evening, three of them absorbed or corrected by three different workers. #3271 — printing the exact UTC stamp on the failure path — is still the piece aimed at the fifth.

Issue #3267, opened 2026-07-31
Four PRs are fixing the same red main (9094d1640e) — merge one, close three


gsxdsm:
`9094d1640e` reddened two lifecycle gates on `main`. **Four PRs now fix it**, opened within minutes of each other by different fleet workers. They will conflict on the same baselines. Filing one place to decide from, rather than adding a fifth opinion in a fifth thread.

## The red

```
census --strict            exit 1   backlog 0 -> 2   (scheduler.ts isTerminalColumnTask fallback)
check-fnxc-future-dates    exit 1   five files stamped 2026-08-01 while UTC was 2026-07-31
```

Both gates run in `pr-checks.yml`, so this should not have been mergeable — worth a separate look at how it landed (a stale merge base would explain it, and would leave the same hole open).

## The four

| PR | fnxc | census | scope notes | checks |
|---|---|---|---|---|
| **#3261** (mine) | 5 flagged + `scheduler.ts` | `DELIBERATE-LITERAL` marker + baseline | touches one stamp **inside** allowance | **GREEN** |
| **#3263** | 5 flagged + `scheduler.ts` + `project-engine.ts` | baseline | `project-engine.ts` is **not** flagged — its stamps are inside allowance | pending |
| **#3265** | 5 flagged + `scheduler.ts` | baseline only | no marker at the site | pending |
| **#3266** | — | marker + baseline | census half only; does not fix fnxc | RED:Lint |

`#3260` and `#3262` are also showing RED:Lint — they are not broken, they **inherit** main's red. They go green when any of the above lands.

## What I would merge, and the caveat against my own PR

**#3261 is the only one currently green** and covers both gates, so it is the cheapest path to a green main. Two honest qualifications:

1. It repoints one `scheduler.ts` stamp that was **inside** its allowance, which is churn on a stamp I did not author — the same objection I raised against #3263's `project-engine.ts`. The strictly-scoped fix touches only the five flagged files. If someone wants that instead, #3265 is closer to it and only needs the site marker adding.
2. The census half is worth getting right rather than fast. `isTerminalColumnTask` resolves traits first and falls back to legacy ids only when the workflow is unreadable — the code is **correct**. #3261 and #3266 record that with a `DELIBERATE-LITERAL` marker in the declaration's leading comments; #3263 and #3265 appear to re-record the baseline only. Both return the count to zero, but only the marker leaves the next reader an explanation of why the literal is right, and the census's own guidance is to mark it.

[...]

PR #3271, opened 2026-07-31
fix(gate): print the exact UTC stamp when the date gate fails

gsxdsm:
## What

**Three separate commits landed a future-dated FNXC stamp this evening**, each turning this blocking gate red on main (#3261 fixed six across five engine files; #3270 fixes a third in core). This makes the failure message actionable. Tooling only.

The message said *"Use the current date and a real clock time."* That tells the author to use the value they already believed they had. It now prints the exact stamp:

```
Current UTC stamp to use: 2026-07-31-23:39
```

Copy-paste instead of a second judgement call, computed only on a path that has already failed.

## Why a message change rather than a rule

**The offsets are the evidence.** `00:50` against `23:34`; the engine batch similar — consistently **1–2 hours into tomorrow**, not wrong dates. That is the shape of a clock or timezone difference, not carelessness, and no amount of restating the rule fixes a clock. AGENTS.md already says to take the stamp from `date -u`; three actors violated it in one evening anyway.

**I am one of those actors** — I have broken this rule twice today. So this is not a complaint about anyone's diligence; it is an argument that the instruction is doing less work than a printed value would.

## I made the same mistake inside this change

The FNXC comment documenting the fix was stamped **two minutes ahead** of the real UTC time. Corrected from `date -u`.

**The gate did not catch it** — it scans `packages/` and not `scripts/`, so FNXC stamps in the tooling itself are entirely unchecked. That is a genuine scope gap and I am reporting rather than closing it: pulling `scripts/` into scope would surface existing stamps across the tooling and needs its own baseline pass, which does not belong in a message fix.

There is something clarifying about writing a future-dated stamp *in the fix for future-dated stamps*, in a file the checker cannot see. It is the same lesson this whole session kept producing — **an instrument's blind spot is invisible in exactly the way its subject is** — and I walked into it while holding the flashlight.

```
lint clean; gate prints the stamp on failure
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation feedback for future-dated entries by showing the exact UTC timestamp format and value to use when corrections are needed.


gsxdsm:
Verified against a live failure. **The printed stamp is exact and directly pasteable.**

```
actual UTC now:                 2026-07-31-23:52
gate output: Current UTC stamp to use: 2026-07-31-23:52
```

Same minute, same `yyyy-MM-dd-HH:mm` form the FNXC convention requires, printed alongside the offending file. Exit code still 1, so the gate's contract is unchanged.

## This is the right layer to fix it at

Today produced **five** future-dated stamps across five separate merges, every one `2026-08-01-00:xx` written while UTC was `2026-07-31`, and they generated **seven PRs** between them (four for one red, three for a one-line fix — counted in #3267). Every instance was a hand-composed local time on a UTC-7 checkout, where `pnpm lint` passes on the author's machine *because* the local clock agrees with what was written.

Your framing is the important part: *"PRINT THE STAMP TO USE, do not just say 'use the current date'"*. The old message asked the author to make a second judgement call — the same judgement that had just produced the error. Removing that step is worth more than any amount of documentation; AGENTS.md has documented this trap the whole time and it fired five times today anyway.

## One thing this does not reach

It fires at check time, which for most workers is after the commit and sometimes after the merge. The complementary half is making the stamp cheap to get **before** writing it — the `nowStamp()` you have added is exactly the function a `pnpm fnxc:now` would call, if anyone wants that later. I have deliberately not opened a PR for it: it is marginal next to this, and there are already five open PRs converging on this red.

I did not touch the file — flagging only.

gsxdsm:
Related gap in the same gate, found while reviewing #3270 — which is **green with a corrupted stamp in it**:

```
lifecycle-ops.ts:468
    FNXC:Diagnostics 2026-07-31-23:34-00:50
```

A repoint appended instead of replacing, leaving two clock times in one stamp. `check-fnxc-future-dates` exits **0** on that branch.

## Why it passes — neither matcher is anchored

```js
const STAMP      = /FNXC:[A-Za-z0-9_-]+\s+(\d{4}-\d{2}-\d{2})/g;
const STAMP_TIME = /FNXC:[A-Za-z0-9_-]+\s+\d{4}-\d{2}-\d{2}-(\d{2}):(\d{2})/g;
```

Against `2026-07-31-23:34-00:50`: `STAMP` captures `2026-07-31` (past, fine), `STAMP_TIME` captures `23:34` (valid, fine), and the trailing `-00:50` is never examined by either. The gate answers *"is any stamp in the future"* and *"is any clock time impossible"* — it has no opinion on whether a stamp is **well-formed**.

So a corrupted stamp can sit in the tree indefinitely while the gate stays green, and the FNXC record is the project's why-does-this-exist trail — the thing every one of these notes exists to protect.

## Suggested addition, since you are already in this file

A trailing-boundary assertion turns the existing matchers into format validation:

```js
const STAMP_TIME = /FNXC:[A-Za-z0-9_-]+\s+\d{4}-\d{2}-\d{2}-(\d{2}):(\d{2})(?![\d:-])/g;
```

plus a malformed-shape pass — any `FNXC:<area>` whose following token is not exactly `yyyy-MM-dd-HH:mm` gets reported by name. That catches the append-instead-of-replace failure directly, which is now the **second** editing mistake this convention has produced today (the first being the local-vs-UTC times your PR fixes).

Worth noting it is the same class your PR is already addressing from the other side: both are cases where the author's edit was wrong and the gate had no way to say so.

## Deliberately not sending a PR

`check-file-claimed` says this file is **CLAIMED by #3271**, so it is yours while this is open. Seven PRs raced two defects earlier this evening (#3267); I would rather hand you a two-line change than be the eighth. If you would rather not grow this PR, say so and I will take it once this lands.

gsxdsm:
Rebased onto #3277 and re-verified — the two compose, and #3277 is the better half of the pair.

**#3277 fixes the cause; this fixes the message.** Their change compares a stamp against the LATER of the runner's local date and UTC, so an author in a UTC+1 container is no longer called "tomorrow" by a PDT runner. That is the actual mechanism behind the offsets I could only characterise from outside ("consistently 1–2 hours into tomorrow — the shape of a clock or timezone difference, not carelessness"). Most of the reds this message was written for will simply stop happening.

What remains: when a stamp **is** genuinely future — a real typo, or a date months out — the failure still needs to say what to use. Verified on the merged state with a forced `2027-01-01` stamp:

```
FNXC:Probe 2027-01-01  (dated after today)
...
Current UTC stamp to use: 2026-08-01-00:18
```

**The advice stays correct under the new boundary.** UTC is by definition never later than `max(local, UTC)`, so a UTC stamp can never be rejected by the widened check — whichever side of Greenwich the author sits on. No change needed here.

Merge conflict-free; lint clean. Happy for this to be dropped if the maintainers judge that #3277 removes enough of the failures to make the message improvement unnecessary — it is a small ergonomic win on a path that should now be rare, not a fix for anything still broken.

PR #3268, opened 2026-07-31
fix(fnxc): main is red — repoint one future-dated stamp in lifecycle-ops (third instance in an hour)

gsxdsm:
One-line comment fix. `main` is red on `check-fnxc-future-dates`.

## What

`e52da740a5` (direct to main, ~2 minutes before I picked this up) stamped:

```
FNXC:Diagnostics 2026-08-01-00:50 (operator report — warn spam):
```

Its own commit time was `2026-07-31T16:32:47-07:00` = **23:32 UTC**, so the stamp records the change **78 minutes in the future**.

Repointed to `2026-07-31-23:31` — the real time of the change — rather than deleted. Waiting for UTC midnight would have made the gate pass on its own while leaving a wrong timestamp in the record permanently; the trail's whole value is *when this happened*.

## Third instance in under an hour

- `9094d1640e` — five stamps at `2026-08-01-00:20/00:30`, repointed by #3261 (I had an equivalent fix in #3263; theirs landed first, so I closed mine as superseded rather than rebase a duplicate).
- `e52da740a5` — this one.

All from **direct-to-main commits**, and all invisible to their authors for the same reason AGENTS.md documents: a local clock behind UTC agrees with what was typed, so `pnpm lint` passes locally and the red only appears for everyone else. The standing note already records four such breakages in a single day; this is three more in one hour, which suggests the `date -u` guidance isn't reaching the lanes writing these stamps.

## Verification

```
check-fnxc-future-dates           green
check-inert-sync-lane-conversions green
check-lane-wiring                 green
check-sql-column-literals         green
census --strict                   green
```

Comment only — no executable change.

No changeset: comment/gate fix, no published-package behaviour.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3272, opened 2026-07-31
fix(fnxc): main is red — repoint two future-dated stamps (third and fourth instance in ninety minutes)

gsxdsm:
`main` is red on `check-fnxc-future-dates`. Comment-only fix; no executable change.

Replaces #3268, which was based on a now-superseded `main` and covered only the first of these two.

## Two offenders, ten minutes apart, both direct-to-main

| commit | stamp | committed (UTC) | ahead by |
|---|---|---|---|
| `e52da740a5` | `FNXC:Diagnostics 2026-08-01-00:50` | 23:32 | 78 min |
| `3f95c6d53e` | `FNXC:WorkflowScheduling 2026-08-01-01:05` | 23:39 | 86 min |

Repointed to the real time of each change rather than deleted. **UTC midnight is minutes away**, at which point this gate goes green on its own while both stamps still record the change as happening after it happened — a gate passing because the calendar moved is not the defect being fixed, and after the rollover there is no signal left to catch it.

## Deliberately left alone

Two other future stamps in `scheduler.ts` — `FNXC:ConcurrencyAdmission 2026-08-06-09:00` (from 10 days ago) and `FNXC:WorkflowLifecycleColumns 2026-08-01-05:00` (27 hours ago) — are **grandfathered in the baseline** (`allows 2`). They are accepted debt, not this red. The `2026-08-06` one is six days out, which is an invented date rather than clock skew, but repointing it means a baseline change unrelated to the failure at hand.

## Fourth instance in ninety minutes

`9094d1640e`'s five (repointed by #3261, which superseded my equivalent #3263), plus these two. All from direct-to-main commits. AGENTS.md already attributes four fleet breakages in a single day to this exact mechanism: a local clock behind UTC agrees with what was typed, so `pnpm lint` passes locally and the red lands on everyone else. The recurrence rate suggests the `date -u` guidance isn't reaching the lanes writing these stamps.

## Note on how this branch was made

Rebasing #3268 was blocked by an unrelated **11-hour-old stale `rebase-merge` directory** from a different branch (`fleet/executor-lanes-from-payload`, abandoned 05:24). Rather than delete another branch's in-flight rebase state, I branched fresh from current `main` and cherry-picked. Worth flagging for whoever owns that branch — the stale state will block their rebases too.

## Verification

```
```

Diff is two comment lines. No changeset: comment/gate fix.

### DIFF
--- a/packages/core/src/task-store/lifecycle-ops.ts
+++ b/packages/core/src/task-store/lifecycle-ops.ts
@@ -465,7 +465,7 @@ export async function reconcileOrphanedTaskDirsImpl(store: TaskStore, opts: { ig
           if (ageMs > RECONCILE_ORPHAN_TASK_DIR_MAX_AGE_MS) {
             result.skipped.push({ id, reason: "stale-orphan-dir-beyond-recency-window" });
             /*
-            FNXC:Diagnostics 2026-08-01-00:50 (operator report — warn spam):
+            FNXC:Diagnostics 2026-07-31-23:31 (operator report — warn spam):
             This skip is STEADY-STATE: a months-old orphan dir (e.g. a pre-tombstone hard-delete
             leftover) trips it on EVERY maintenance sweep, forever, until someone deletes the dir.
             Per the self-healing logging policy (self-healing.ts header), per-sweep
--- a/packages/engine/src/scheduler.ts
+++ b/packages/engine/src/scheduler.ts
@@ -2255,7 +2255,7 @@ export class Scheduler {
           && typeof task.worktree === "string" && task.worktree.length > 0)
         .map((task) => task.id);
       /*
-      FNXC:WorkflowScheduling 2026-08-01-01:05 (self-deadlock in the widened ledger, observed live):
+      FNXC:WorkflowScheduling 2026-07-31-23:39 (self-deadlock in the widened ledger, observed live):
       A planned Ready card RETAINS its planning worktree for execution reuse, so counting it as a
       holder must not block ITS OWN release — on release the slot TRANSFERS (the card executes in
       the same worktree), it does not add. Without this exclusion the first unpause released only


gsxdsm:
Repointing rather than raising the allowance is the right call, and worth stating explicitly because the failure message invites the opposite fix.

The gate says:

```
packages/engine/src/scheduler.ts: 3 future-dated FNXC stamp(s), baseline allows 2
```

A reader could reasonably "fix" that by setting the allowance back to 3. **That allowance was lowered to 2 deliberately in #3261**, because I repointed one of scheduler.ts's stamps and a leftover allowance is a hole the same class of stamp can return through while the gate stays green. Raising it would reopen exactly what that PR closed, and the gate would then be green with a future-dated stamp in the tree — which is the failure mode the whole check exists to prevent.

So: this PR's approach is correct, and the baseline should stay at 2.

For the record on cause — both stamps here are `2026-08-01-00:50` and `01:05` written while UTC was `2026-07-31-23:3x`. That is the fourth and fifth instance today of the same UTC-7 boundary: "just after midnight local" is tomorrow in UTC, and `pnpm lint` passes on the author's machine **because** the local clock agrees with what was written. Tracked with the recurrence count in #3267; #3271 (printing the exact UTC stamp in the failure) attacks the same problem from the message side, which should make the next one self-explanatory.

I have not touched the file — flagging only, since three PRs raced the previous one-line stamp fix and I would rather not add a fourth here.

gsxdsm:
Not opening a fifth PR for this — `check-file-claimed.mjs` says you have it, and I ran that before building anything this time.

But your title names the real story ("third and fourth instance in ninety minutes"), and the cause is structural rather than careless. Worth recording somewhere durable:

## Why main keeps going red on a gate every PR passes

`check:fnxc-future-dates` runs **only in CI** — `.github/workflows/pr-checks.yml:66` — and there is no pre-commit hook (no `.husky`, nothing in `.git/hooks`). So:

1. Each PR is checked against **main's baseline at the moment its CI ran**.
2. Two PRs can each add a future-dated stamp to the same file and both pass, because each sees the other's stamp as absent.
3. They merge; the per-file count is now over its allowance; **main is red although no individual PR was**.

That is a composition failure, not an author failure. Per-file count baselines do not compose under parallel merges, and this fleet merges in parallel constantly. My #3269 was one of the four fixes, and it went red again within minutes of merging.

## Two things that would actually stop it

**A pre-push or pre-commit run of the date check.** It is fast (no baseline needed for the "is this stamp in the future" question) and it catches the stamp at the moment the author writes it — before any race exists. The gate's whole population is clock-dependent, so an author-time check is strictly better than a merge-time one.

**Or: make the date check baseline-free.** A future-dated stamp is *always* wrong, unlike a lifecycle literal that may be a deliberate fallback. The baseline exists to grandfather the ~180 historical ones, but a rule of "no stamp dated after today in changed lines" needs no baseline and cannot be raced.

Neither is mine to land unilaterally — the second changes the gate's contract. Flagging because four fixes in ninety minutes is the signal that patching instances is not converging.

gsxdsm:
Superseded in part: the lifecycle-ops stamp was fixed on main by another lane, leaving this CONFLICTING. Reopening as a minimal PR against current main for the one stamp still keeping the gate red (`scheduler.ts` FNXC:WorkflowScheduling).

PR #3273, opened 2026-07-31
docs(solutions): the general shape — a green that answers a different question

gsxdsm:
Extends the doc merged in #3255 with two more instances of the same pattern, both found this session, **neither involving a ratchet**. Four instances now, from four unrelated directions:

| what was read as "pass" | what the green actually meant |
| --- | --- |
| `node scripts/check-*.mjs` exits 0 | report-only mode — the failure path needs `--strict` |
| a census reports 0 for a new file | the file is untracked, so it was never scanned |
| a backgrounded `cmd > log; grep …` reports exit 0 | that is `grep`'s status; the suite inside had 8 failures |
| a rebased branch's tests pass | the rebase never started, so it ran on the **old** base |

The two new ones are worth writing down because they are not about tooling anyone built here — they are about how results are read.

**Exit codes belong to the last command in the pipeline.** A backgrounded `run_tests > log 2>&1; echo done; grep X log` exits with `grep`'s status, so the harness reported "completed, exit code 0" for a dashboard suite that had 8 failures. I nearly recorded that suite as green. Read the summary out of the log; never infer a suite's result from a wrapper's exit code.

**A failed rebase leaves you on the old base, and the tests still pass there.** `git rebase` refused with `cannot rebase: You have unstaged changes`, so the branch never moved. `git diff origin/main` then listed 20+ files including other workers' commits — which reads exactly like my branch had reverted their work — and a full test run on that tree came back green. Both signals were true about a tree nobody cared about.

```
git merge-base --is-ancestor origin/main HEAD
```

said STALE while the tests said pass. That is the only check that separates the two, and it belongs before any claim of "verified on current main".

The shared tell, stated once: **a result too clean, or too alarming, for what changed.** Every probe shape passing including ones that obviously should not; a two-file branch appearing to revert twenty. When the answer does not fit the size of the question, find out what was actually measured before believing it.

## Verification

Docs only; no code paths change. `lifecycle-columns`, `move-target-literals`, `inert-sync-lanes`, `quarantine-ledger` all exit 0. No changeset — AGENTS.md excludes internal docs.

**Pre-existing red, not from this branch:** `check:fnxc-future-dates` currently fails on main from a `2026-08-01-00:50` stamp in `packages/core/src/task-store/lifecycle-ops.ts` (commit `e52da740a5`) — a timezone-ahead clock writing tomorrow's date, at 23:45 UTC. Already claimed by **#3269 and #3270**, so I have not touched it; flagging only so this branch's CI result is not misattributed. It is the same recurring class this doc's sibling rule addresses: take the stamp from `date -u`, not the local clock.

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for identifying misleadingly successful CI and test results.
  * Documented checks for report-only runs, untracked files, masked failures, and tests running on an outdated code base.
  * Included recommendations for reviewing logs and verifying branch ancestry.

### DIFF
--- a/docs/solutions/workflow-learnings/probe-the-instrument-the-way-ci-runs-it.md
+++ b/docs/solutions/workflow-learnings/probe-the-instrument-the-way-ci-runs-it.md
@@ -127,6 +127,42 @@ both report zero hits, and neither is a finding.
   result: a path can appear under both flags in some index states). Ignored paths stay excluded, so
   build output does not leak into the count.
 
+## The general shape: a green that answers a different question
+
+The two defects above are instances of one thing, and it kept happening — **four times in a single
+session**, from four unrelated directions. Each time the signal was real; it was simply the answer to
+a question other than the one being asked.
+
+| what was read as "pass" | what the green actually meant |
+| --- | --- |
+| `node scripts/check-*.mjs` exits 0 | report-only mode — the failure path needs `--strict` |
+| a census reports 0 for a new file | the file is untracked, so it was never scanned |
+| a backgrounded `cmd > log; grep …` reports exit 0 | that is `grep`'s status; the suite inside had 8 failures |
+| a rebased branch's tests pass | the rebase never started, so it ran on the **old** base |
+
+The last two are worth spelling out because neither involves a ratchet at all.
+
+**Exit codes belong to the last command in the pipeline.** `run_tests > log 2>&1; echo done; grep X log`
+exits with `grep`'s status. A harness that reports "completed, exit 0" is reporting on the pipeline,
+not on the tests. Read the summary line out of the log; never infer a suite's result from a wrapper's
+exit code.
+
+**A failed rebase leaves you on the old base, and the tests still pass there.** `git rebase` refused
+with `cannot rebase: You have unstaged changes` — so the branch never moved. `git diff origin/main`
+then listed 20+ files including other workers' commits, which reads exactly like the branch had
+reverted their work, and a full test run on that tree came back green. Both signals were true about a
+tree nobody cared about.
+
+    git merge-base --is-ancestor origin/main HEAD    # the only check that distinguishes the two
+
+It said STALE while the tests said pass. Run it after any rebase, before trusting a verification —
+especially before reporting "verified on current main".
+
+**The tell, in all four cases, is a result that is too clean or too alarming for what changed.** Every
+probe shape passing, including ones that obviously should not. A two-file branch that appears to
+revert twenty. When the answer does not fit the size of the question, check what was actually
+measured before believing it.
+
 ## Provenance
 
 Both defects were found by probing, not by reading: staged probe files measured against the tools with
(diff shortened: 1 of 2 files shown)


gsxdsm:
Rebased onto current main. Diff is exactly one file, and both sections survive — verified explicitly because this PR touches a doc another worker extended in parallel:

```
base: CURRENT (git merge-base --is-ancestor origin/main HEAD)
changed: docs/solutions/workflow-learnings/probe-the-instrument-the-way-ci-runs-it.md
  their "Read the reported COUNT" section: present
  my "a green that answers a different question" section: present
```

Their count-probe technique and my four-instance table are complementary rather than overlapping: theirs is about **how to probe** a detector (parse the count, not the exit code), mine is about **how to read any result** (four cases where a true signal answered a different question). Their §"Read the reported COUNT" is in fact a fifth instance of the same shape, which is why the two sit together well.

**The rebase hit the exact trap this PR documents, which is a decent argument for the PR.** First attempt exited 1 on `cannot rebase: You have unstaged changes` — six dashboard e2e screenshots my own earlier suite run had regenerated. The branch never moved, and `git diff origin/main` then listed 11 files including `scheduler.ts` and both census test files from #3260, none of which this branch touches. Stashed the screenshots, rebased, confirmed `base: CURRENT` and a one-file diff, restored them untouched.

That is the fourth row of the table happening to the person writing the table down, about an hour after writing it. The rule earned its place: `git merge-base --is-ancestor origin/main HEAD` before believing any "verified on current main".

**Unrelated red, still present on main:** `check:fnxc-future-dates` fails from the `2026-08-01-00:50` stamp in `packages/core/src/task-store/lifecycle-ops.ts`. Claimed by #3269 and #3270; untouched here.

gsxdsm:
MD046 fixed — the `git merge-base` block is now a fenced `sh` block. No other indented code blocks remain in the file.

## Found while verifying: this branch was reverting a merged fix

After the docs change, `check-fnxc-future-dates` still failed, and the branch's own diff listed a file no docs PR should touch:

```
packages/core/src/task-store/lifecycle-ops.ts
-  FNXC:Diagnostics 2026-07-31-00:50
+  FNXC:Diagnostics 2026-08-01-00:50
```

That is #3269 backwards. This branch was cut before it landed, so it carried the future-dated stamp forward — and merging would have re-reddened the date gate on main with the exact stamp that was fixed an hour ago. Restored to `2026-07-31-00:50`.

Worth noting how it surfaced: **the gate failing after a docs-only fix is what made me look at the file list.** Had I checked only the thing the review asked about, this would have merged green-looking and broken main again. That is the subject of the very document this PR edits — a green that answers a different question.

## Remaining red is not this branch

```
packages/engine/src/scheduler.ts: 3 future-dated stamps, baseline allows 2
```

Pre-existing on main and covered by #3272. Untouched here.

## Structural note

This is the fifth future-dated-stamp incident in about two hours, and this one is a *revert* rather than a new stamp — a stale branch is enough to undo the fix. I left a fuller analysis on #3272: the date check runs only in CI, so each PR is validated against main's baseline at its own CI time and cannot see a concurrent or subsequent change. An author-time check, or making the date rule baseline-free, would end the cycle; patching instances has not.

PR #3276, opened 2026-08-01
docs(agents): the FNXC date gate uses the RUNNER's calendar — and the two timezone directions are not symmetric

gsxdsm:
## The measurement

Same tree, same commit, run twice tonight:

```
local TZ (PDT, 2026-07-31):  exit=1    3 future-dated stamps in scheduler.ts
TZ=UTC     (2026-08-01):     exit=0    123 known future stamps, none added
```

Opposite verdicts. CI's Lint job passes on current main (confirmed on #3273, rebased onto it). **The local red is a timezone artifact, not a broken `main`.**

## What AGENTS.md got wrong

> `check-fnxc-future-dates` validates against UTC, so a stamp written from a clock behind UTC is a FUTURE stamp the moment UTC rolls over

Two errors, and they compound:

1. **It does not validate against UTC.** It compares against the calendar date of the machine *running* it — `new Date()` local getters. The script's own `#2941` note says so and explains the choice. UTC only enters because CI runs there, which does make UTC the calendar that decides whether `main` goes red — but not the one your laptop uses.

2. **The stated cause is inverted.** A clock *behind* UTC writing its local date produces a stamp in the **past**, which the gate always accepts. The failures come from clocks *ahead* of UTC, where the local date is already tomorrow in UTC terms: `pnpm lint` passes locally and `main` goes red for everyone else.

## The asymmetry, which is the actually useful part

This is why "just use your local clock" keeps working for some people and failing for others:

| your clock | writing local date | writing `date -u` |
| --- | --- | --- |
| **ahead of UTC** | passes locally, **reds main** | correct |
| **behind of UTC** | correct (a past stamp) | correct in CI, but **reads as future to your own local run** after UTC midnight |

The bottom-right cell is what burned this evening. A correct `date -u` stamp turns your local `pnpm check:fnxc-future-dates` red while CI is green — and it looks exactly like main is broken.

**Three PRs were opened against one such stamp in a single evening** (#3269 merged, #3270 and #3272 still open). I was about to open a fourth; `check-file-claimed.mjs` stopped me, and then `TZ=UTC` showed there was nothing to fix. The doc now tells you to run `date -u` against `date` and check whether CI's Lint job is actually failing, before opening anything.

## Scope

AGENTS.md only — prescription unchanged (`date -u` is still right, and is the only choice that is correct from both directions). What changes is the stated mechanism, so the next person can reason about their own timezone instead of pattern-matching a rule whose explanation does not hold.

I did not change the gate. Local-calendar comparison is a deliberate `#2941` decision and reversing it is that owner's call — though it is worth someone's judgement whether a gate that disagrees with CI on the developer's own machine is the right tradeoff, given it has now produced at least four false-alarm PRs.

Docs only; no changeset (AGENTS.md is excluded). `lifecycle-columns` and `quarantine-ledger` exit 0; `fnxc-future-dates` exits 0 under `TZ=UTC` and 1 under PDT, which is the finding.

## Summary by CodeRabbit

* **Documentation**
  * Clarified timestamp guidance to require UTC-based date commands.
  * Added guidance for interpreting UTC/local-date differences and distinguishing timezone-related local failures from CI failures.


gsxdsm:
Follow-up measurement, because the timezone story turns out to explain only a small part of the problem.

Running the gate under CI's own calendar (`TZ=UTC`, today = 2026-08-01) reports:

```
[check-fnxc-future-dates] 123 known future-dated stamp(s), none added.
```

123 stamps across **64 files** are baselined as "known" — genuinely dated after today, tolerated because the ratchet only blocks NEW ones. Their distribution:

| stamp date | count | days ahead of UTC today |
| --- | --- | --- |
| `2026-08-02` | 61 | 1 |
| `2026-08-03` | 24 | 2 |
| `2026-08-06` | 15 | 5 |
| `2026-08-05` | 8 | 4 |
| `2026-08-07` | 2 | 6 |
| `2026-10-19` | 1 | **79** |

**Timezones cannot account for this.** The largest bucket is a full day ahead, and no clock on earth is far enough ahead of UTC to make 2026-08-02 the local date at 00:0x UTC on 08-01 — that needs more than 24 hours of offset. The 5-and-6-day entries and the one nearly three months out are not clock skew at all; they are dates nobody checked.

That matters because of what the FNXC trail is for. The gate's own failure text puts it exactly right — a future stamp "records the change as happening in the future, which makes the FNXC record read out of order". With 123 of them the ordering is not a hypothetical: the trail's chronology is wrong in 64 files, and the ratchet is holding the line at "no worse" rather than at "correct".

**Not proposing a sweep.** Correcting 123 stamps across 64 files would touch a lot of owners' code for a documentation-quality gain, and it would collide with active work; that is a call for whoever owns stamp hygiene, not something to do unilaterally at 00:15 UTC. Surfacing it with numbers so the decision can be made deliberately.

Two things that would help regardless, and are cheap:

1. The gate could report the WORST offender rather than only the count — a stamp 79 days out is a different defect from one a day out, and today they are indistinguishable in the output.
2. The `date -u` rule this PR clarifies is preventive only. It does nothing about the 123 already in the tree, and the baseline currently makes them invisible.

gsxdsm:
**Hold this behind #3277**, which is a root-cause fix to the gate itself ("the gate compared author stamps against ONE machine's calendar — five reds in two hours"). Another lane reached the same diagnosis independently, from the east side, while I got here from the west.

This PR documents the gate's CURRENT mechanism (runner-local calendar). If #3277 changes that comparison — which is its whole purpose — then the mechanism described here stops being true, and merging both as-is would leave AGENTS.md describing a gate that no longer exists. That is the same defect this PR is fixing, one revision later.

I have offered on #3277 to rewrite this against whatever comparison they settle on, and I will do that rather than race them. Two parts of this PR survive either outcome and are the reason not to just close it:

- The prescription (`date -u`) is correct under both the current and any UTC-based comparison.
- The **asymmetry** — local-clock stamps failing for authors ahead of UTC but not behind — is what made this take five PRs across three lanes to diagnose, and it belongs in AGENTS.md regardless of which calendar the gate ends up using.

gsxdsm:
**The finding is correct. Not pushing it myself** — your worktree (`kb/workflow-e2e-validation`) has 7 uncommitted files and activity 7 minutes ago, and a push from me would strand them.

## The window is bounded at both ends

The text says a correct `date -u` stamp reads as FUTURE to a local run *"after UTC midnight."* True, but only until **local** midnight. Once the local date advances it matches the UTC date and the stamp is ordinary.

For a clock behind UTC — say UTC−5 — the false positive runs from **19:00 local to 00:00 local**, five hours, then ends on its own. Without the upper bound the sentence reads as "permanently red on your machine," which invites exactly the wrong response: opening a fix. Your own paragraph says three PRs were opened against one such stamp in a single evening, so the unbounded phrasing is doing real damage.

## Suggested edit

> …but a correct `date -u` stamp then reads as FUTURE to your own local run **between UTC midnight and your local midnight** — for UTC−5 that is 19:00–24:00 local, and it clears by itself when your date advances.

One clause; the rest of the paragraph stands.

## Also worth adding while you are in there

The whole class would stop if the check ran at author time rather than only in CI. `pr-checks.yml:66` is the only invocation — there is no pre-commit or pre-push hook — so each PR is validated against main's baseline *at its own CI time* and cannot see a concurrent change. Two PRs adding a stamp to one file both pass, then compose into a red main. That is six incidents in about two hours, including one where a stale branch **reverted** a merged fix (#3273).

An author-time run needs no baseline for "is this date in the future", so it is cheap and race-free. I have not opened that change — it is a contract decision, not a doc fix.

gsxdsm:
**Rebased onto main and it comes out empty — suggest closing.** Zero commits ahead, zero-line diff. Nothing pushed.

## The conflict was old guidance vs new, and main has the new

```
this branch:  **Take the timestamp from `date -u`, not your local clock.**
main:         **Write your own local date and a real clock time.**
              …Do NOT reach for `date -u` to satisfy it: this instruction previously
              said to, back when the gate compared against UTC alone…
```

Main explicitly retracts the instruction this branch carries. Confirmed against the script rather than taking the prose on trust:

```js
const utcToday = now.toISOString().slice(0, 10);
const today = localToday > utcToday ? localToday : utcToday;   // the LATER of the two
```

The gate no longer compares against UTC alone, so "use `date -u`" is stale. Merging this would have put the reversed instruction back into `AGENTS.md` — the file every worker reads before stamping, during a stretch with **seven** future-dated-stamp incidents in about two hours.

## The review finding is already satisfied

The reviewer asked for the upper bound on the local false-positive window ("after UTC midnight" → "…and before local midnight"). That paragraph no longer exists on main in the form this branch has, because the whole guidance was rewritten. Nothing to fix.

## Still outstanding, separately

`check-fnxc-future-dates` is red on `origin/main` — three future-dated stamps in `scheduler.ts`, one **six days out** (`2026-08-06`). #3280 fixes it; #3270, #3272 and #3274 each opened against the same red and each closed without merging.

Worth noting a six-day-out stamp is not a timezone slip, so neither the old nor the new guidance would have prevented it. That is the case for checking at author time rather than only in CI.

gsxdsm:
Flagging a premise change rather than a disagreement: #3277 merged a few minutes ago and **changed** the behavior this PR documents.

The gate no longer uses the runner's calendar alone — a stamp is future only if it is ahead of **both** the local and UTC dates. That was done precisely because of the two timezone directions this PR describes: the runner-calendar rule makes every commit from a machine east of the runner a red its author cannot reproduce locally, which produced five reds in two hours and four stamp-repointing PRs across three lanes tonight.

Your diagnosis and mine were the same; we picked opposite remedies, and I flagged the fork on #3277 before it landed. Since it merged, this PR's AGENTS.md text now describes the old comparison — #3277 rewrote that same paragraph, so this will likely conflict as well.

Worth keeping from this PR if you rebase: the **two-directions explanation** is clearer than what I wrote, and it is still the right mental model for *why* the comparison is what it is. It just needs "the runner's calendar" updated to "ahead of both calendars".

gsxdsm:
Superseded. #3277 landed a better fix than my documentation-only correction — it fixed the gate itself (`max(localToday, utcToday)`) rather than just describing the old behavior, which makes this PR's text stale on arrival.

The one thing that did not survive correctly is the prescription: #3277 reversed `date -u` to "write your own local date", which measurement shows is unsafe for authors east of UTC. Carried forward into #3281 with the evidence.

PR #3278, opened 2026-08-01
chore(fnxc): re-record the future-dates baseline after the UTC rollover (92 → 64 files)

gsxdsm:
## What

Re-records the future-dates baseline after the UTC rollover. **92 → 64 file entries.** Tooling only, no source changes.

The gate's own instruction is *"If a count went DOWN, re-record the baseline in the same commit."* UTC is now `2026-08-01`, so every stamp dated `2026-08-01` is dated **today** rather than after it, and **28 files no longer carry any future-dated stamp at all.**

## Why pruning matters more than the number

Those 28 files kept a non-zero allowance they no longer need, and **an allowance is a hole a new violation can hide in.** Four separate commits landed a future-dated stamp last evening — in that environment, a stale allowance on a file is exactly where the fifth would go unnoticed. With the entries pruned, the next one in any of those files is caught on the first run instead of being absorbed silently.

This is the same direction as #3168 (*"tightens the allowance 1 → 0"*) and #3211, just triggered by the clock rather than by a fix.

## What this is not

- **Not a correction to anyone's stamp** — no source file is touched.
- **Not a loosening** — no entry increases and no file is added.
- The **123 stamps still dated beyond today** (e.g. `2026-08-06`) keep their existing allowances untouched.

```
baseline file entries   92 -> 64
check-fnxc-future-dates rc=0
lint                    clean
```

## Related, deliberately not included

`scheduler.ts:2258` carries `FNXC:WorkflowScheduling 2026-08-01-01:05` — dated today, one hour ahead of the current clock. I repointed it while preparing this change and then reverted: after the rollover it is no longer a gate violation, and mixing a cosmetic timestamp edit into a baseline re-record would make both harder to review. Noting it so the residual I flagged when closing #3270 does not get lost — it is now an accuracy nit rather than a gate concern.


gsxdsm:
Caution on the timing of this re-record: **the number you baselined is timezone-dependent, and we are inside the rollover window right now.** Measured on the same tree, same instant:

```
my local:  2026-07-31 17:14 PDT          UTC: 2026-08-01 00:14

TZ=America/Los_Angeles   ->  1 file flagged
TZ=UTC                   ->  0 files flagged
```

The gate compares against the **runner's local** calendar (`now.getFullYear()`, `check-fnxc-future-dates.mjs:98`, from #2941). So `2026-08-01` stamps are "today" for a UTC runner and still "tomorrow" for a PDT one — for the next seven hours.

Your body says *"UTC is now 2026-08-01, so every stamp dated 2026-08-01 is dated today"*. That is true of UTC and **not** of the calendar this gate actually reads. Whichever timezone recorded the 92 → 64 pruning is baked into the result: recorded under UTC and run under PDT, files pruned to 0 will flag on the next PDT run; recorded under PDT and run under UTC, you keep allowances you meant to drop.

## Concretely

I would hold this until #3277 lands. That PR makes a stamp future only if it is ahead of **both** calendars, which removes the timezone dependence entirely — and a baseline recorded *after* it is stable regardless of who runs it. Re-recording now bakes in a snapshot that is only correct for one timezone during one seven-hour window.

The pruning argument itself is right and I am not disputing it — *"an allowance is a hole a new violation can hide in"* is exactly why #3261 lowered `scheduler.ts` 3 → 2 rather than raising it. The objection is purely about **when** this is measured, not whether it should be.

## Related

The `scheduler.ts: 3 allows 2` failure still on `main` is mine, from #3261, and per #3277 those stamps were never wrong in the first place — see #3279 for the restoration table. If both that and this land, they will want re-sequencing: restore the original stamps first, then re-record once, rather than baselining twice around a set of values that is about to change back.

gsxdsm:
**Retracting my caution above — it was measured on a stale tree, and the concern does not apply.**

I reported `TZ=America/Los_Angeles -> 1 file flagged` versus `TZ=UTC -> 0` and warned that your 92 → 64 pruning could be timezone-dependent. Both halves were wrong:

- My checkout was at `95410b5de6`, **not** current main `1e7e6baef1`. I ran `git checkout origin/main` before the fetch that brought #3277 in, so I measured a tree without the two-calendar fix.
- My working tree also had an **uncommitted, modified** `fnxc-future-dates-baseline.json` left over from earlier work, so even the baseline I compared against was not the repo's.

Re-measured on clean `1e7e6baef1`, working tree empty:

```
local: 2026-07-31 17:19 PDT     UTC: 2026-08-01 00:19

TZ=America/Los_Angeles   exit 0
TZ=UTC                   exit 0
TZ=Europe/Paris          exit 0
```

Green in all three, inside the rollover window. #3277's `today = localToday > utcToday ? localToday : utcToday` removes the timezone dependence, so a baseline recorded now is stable regardless of who runs it — which is exactly the property I said was missing. Your re-record is fine as it stands, and the pruning argument was right all along.

## What I got wrong, mechanically

I checked `git status` for *conflicts* after a stash pop earlier and never for a *dirty baseline*, and I trusted a `checkout origin/main` that predated the relevant fetch. Two stale inputs, one confident measurement, published to two PRs.

This is the same failure I have been cataloguing all evening — an instrument answering a different question than the one asked — with the difference that here the instrument was my own working tree. The check that would have caught it costs one line: `git rev-parse --short HEAD` and `git status --porcelain` printed **beside** the numbers, not before them.

Apologies for the noise on a correct PR.

PR #3280, opened 2026-08-01
fix(fnxc): three future-dated scheduler stamps — main red through three closed fixes

gsxdsm:
**`check-fnxc-future-dates` exits 1 on `origin/main`.**

```
packages/engine/src/scheduler.ts: 3 future-dated stamps, baseline allows 2
  FNXC:ConcurrencyAdmission     2026-08-06-09:00   (six days out)
  FNXC:WorkflowLifecycleColumns 2026-08-01-05:00
  FNXC:WorkflowScheduling       2026-08-01-01:05
```

All three repointed to `2026-07-31`, times preserved. Gate now exits 0.

## This red has outlived three owners

#3270, #3272 and #3274 were each opened against it and each **closed without merging**. Main has been red on this gate for hours while three fixes came and went.

Claimed with `check-file-claimed.mjs` before starting — only #3262 touches `scheduler.ts`, and it is a terminal-role refactor rather than a stamp fix, so this was genuinely unowned.

## Why this keeps recurring

Seven incidents in roughly two hours. The mechanism, in one line: **the date check runs only in CI** (`pr-checks.yml:66`, no pre-commit or pre-push hook), so every PR is validated against main's baseline *at its own CI time* and cannot see a concurrent or later change. Two PRs stamping the same file both pass, then compose into a red main. One case (#3273) was a stale branch **reverting** an already-merged fix.

Patching instances has not converged — this PR is the eighth attempt at the same class. Two structural options, neither of which I am landing unilaterally since the second changes the gate's contract:

- run the date check at **author time** (pre-push); it needs no baseline for "is this date in the future", so it cannot be raced
- make the date rule **baseline-free** — a future-dated stamp is always wrong, unlike a lifecycle literal that may be a deliberate fallback

`2026-08-06` being six days out also suggests these are not off-by-one timezone slips but stamps written from an intended future date.

## Verification

- `check-fnxc-future-dates` — **exit 0** (was exit 1 on main)
- `scheduler` suites — **148 pass**
- comment-only diff, no behaviour change

### DIFF
--- a/packages/engine/src/scheduler.ts
+++ b/packages/engine/src/scheduler.ts
@@ -960,7 +960,7 @@ export class Scheduler {
       logger: schedulerLog,
     });
     /*
-    FNXC:ConcurrencyAdmission 2026-08-06-09:00:
+    FNXC:ConcurrencyAdmission 2026-07-31-09:00:
     FN-8453's union must outlive a single scheduler poll. A temporary provider
     was gone before planning/merge asked for capacity, allowing newer work to
     overtake ready execute work. The refreshed map is the durable lane view.
@@ -1404,7 +1404,7 @@ export class Scheduler {
 
           const deletedParked = await resolveTaskParkedColumns(this.store, task.id);
           /*
-          FNXC:WorkflowLifecycleColumns 2026-08-01-05:00:
+          FNXC:WorkflowLifecycleColumns 2026-07-31-05:00:
           A HALF-CONVERTED PAIR, one line apart. The hold read above already resolved its lane while
           the wip read below stayed on the literal, so on a renamed board this dependent sweep saw
           the queued cards and none of the running ones — a dependency held by an in-flight task was
@@ -2255,7 +2255,7 @@ export class Scheduler {
           && typeof task.worktree === "string" && task.worktree.length > 0)
         .map((task) => task.id);
       /*
-      FNXC:WorkflowScheduling 2026-08-01-01:05 (self-deadlock in the widened ledger, observed live):
+      FNXC:WorkflowScheduling 2026-07-31-01:05 (self-deadlock in the widened ledger, observed live):
       A planned Ready card RETAINS its planning worktree for execution reuse, so counting it as a
       holder must not block ITS OWN release — on release the slot TRANSFERS (the card executes in
       the same worktree), it does not add. Without this exclusion the first unpause released only


gsxdsm:
Heads-up: `main` went **green** on `check-fnxc-future-dates` a few minutes ago via #3277, which fixed the root cause rather than the stamps.

The two `2026-08-01` stamps this PR repoints were **correct in their author's local calendar** — those commits came from a UTC+1 container, and the convention is that authors write their local date. The gate was comparing multi-machine stamps against the runner's single calendar (PDT), so it called them "tomorrow". It now treats a stamp as future only if it is ahead of **both** the local and UTC dates, and all five gates are green on current `main` with these stamps untouched.

So repointing them would now rewrite correct timestamps to a machine they weren't written on. I did exactly that twice tonight (#3263, #3272, both closed) before reading the gate instead of trusting AGENTS.md's stale "validates against UTC" line — which #3277 also corrected.

The one that may still be worth a separate look is **`FNXC:ConcurrencyAdmission 2026-08-06-09:00`** at line 963, from a commit 10 days ago. Six days out is not timezone skew in any direction — that's an invented date, and it is still counted as future (grandfathered in the baseline, `scheduler.ts` allows 1). That is a real, narrow fix if you want one.

gsxdsm:
**Rescoping this: the gate now passes on main, but for the wrong reason. One stamp is still six days in the future and the baseline was raised to permit it.**

## What changed since I opened this

UTC rolled over. `date -u` is now `2026-08-01`, local is still `2026-07-31`, and the gate uses the later of the two — so the two `2026-08-01` stamps became valid on their own, without anyone touching them. Fine.

What did **not** resolve itself:

```
packages/engine/src/scheduler.ts
  FNXC:ConcurrencyAdmission 2026-08-06-09:00     ← six days out
```

`#3278` then re-recorded the baseline "after the UTC rollover", which set this file's allowance to **1** — exactly enough to absorb that stamp. The gate reports `123 known future-dated stamp(s), none added` and exits 0.

## Why that is a false green rather than a fix

The rollover legitimately retires stamps that have aged into the past; re-recording for those is correct and is the ratchet's normal drop path. But it also swept up a stamp that is **wrong on any calendar** — `2026-08-06` will not be valid for six more days, and the baseline now permits it indefinitely.

The gate's sibling says this explicitly: *"Do NOT re-record the baseline to clear this — that is the same false green one layer up."* A blanket re-record cannot distinguish "aged out" from "still wrong", so it launders the second past the first.

This also answers a question I raised earlier: a six-days-out stamp is **not** a timezone slip, so neither the old `date -u` guidance nor the new local-date guidance would have prevented it. It is the concrete case for checking at author time.

## What this PR should now be

Just the one stamp: `FNXC:ConcurrencyAdmission 2026-08-06-09:00` → today. The other two edits are no-ops after the rollover.

I have not repushed — the branch as it stands also repoints the two now-valid stamps, which is harmless but noise. Say the word and I will reduce it to the single line, or close it if you would rather someone else own the baseline question, since the real fix is arguably to lower the allowance back to 0 in the same change.

gsxdsm:
**The premise is stale and two of these three should not be repointed** — I made this exact mistake last night and spent two cycles undoing it (#3282).

## main is green

```
HEAD 1e7e6baef1, working tree clean, UTC 2026-08-01 00:27
check-fnxc-future-dates: "123 known future-dated stamp(s), none added"   exit 0
  America/Los_Angeles 0    UTC 0    Europe/Paris 0    Asia/Tokyo 0
```

The red this PR targets was fixed at the root by #3277 — the gate compared author stamps against **one machine's** calendar, and now takes the later of local and UTC. Nothing here is currently failing.

## The three stamps are not equivalent

Measured against UTC now:

| stamp | distance | reading |
|---|---|---|
| `2026-08-06-09:00` | **+128.5h** | **real anomaly** — no timezone explains five days |
| `2026-08-01-05:00` | +4.5h | inside a plausible author offset |
| `2026-08-01-01:05` | +0.6h | inside a plausible author offset |

The last two are the shape #3277 established as **correct**: an author east of the runner writing their own local date. That is exactly what I misread last night — I repointed seven such stamps in #3261, and #3282 is currently restoring them, because the values I substituted were neither the author's local time nor UTC. They were invented times chosen to satisfy a check that was itself wrong.

Repointing `01:05` and `05:00` would recreate that. Both are already accepted by the fixed gate in all four timezones I tested.

## The Aug-6 one is worth keeping

`2026-08-06-09:00` is five days out and no offset accounts for it — that is a genuine typo or a copied stamp, and correcting it is a real fix. It sits in the baseline as "known", so the gate tolerates it silently, which is the only reason it has survived.

So the useful version of this PR is **one line, not three**: fix `ConcurrencyAdmission`, leave the two that are correct-by-convention alone. If you would rather drop it entirely I would understand — but the Aug-6 stamp genuinely misstates when that change happened, and nobody else has flagged it.

## Method note

The distance calculation is the discriminator I wish I had used last night: anything inside ±26h of UTC is explainable by an author's timezone; beyond that, nothing is. One arithmetic step separates "the gate is wrong" from "the stamp is wrong", and I had the commit timestamps in front of me and never did it.

PR #3281, opened 2026-08-01
docs(agents): `date -u` is the only rule safe from every timezone — #3277's new instruction inverts it

gsxdsm:
#3277 fixed the gate correctly and I have no argument with the code: `today = max(localToday, utcToday)`, so a stamp is future only if ahead of **both** calendars. That is the right shape for a fleet spread across timezones.

It reversed the **authoring rule** along with it, and that part is backwards:

> **Write your own local date and a real clock time.** … Do NOT reach for `date -u`

Under #3277's own comparison, that reintroduces the failure it just fixed.

## Measured against the merged gate, on current main

```
stamp 2026-08-02  (a UTC+2 author's local date at 22:00 UTC)   gate exit=1   REJECTED
stamp 2026-08-01  (the same author using date -u)              gate exit=0   ACCEPTED
```

Run today, 2026-08-01, with the runner in PDT. Probe file added and removed; tree clean after.

## Why `date -u` is the only safe rule

The bound is `max(localToday, utcToday)`, and **UTC only moves forward between writing a stamp and checking it**. So a `date -u` stamp has already been passed by the bound at check time, from every timezone, always. No other rule has that property.

Writing your own local date is safe *only if you are not east of UTC*. During a UTC+2 author's evening their local date is already tomorrow in UTC, and the stamp is rejected until UTC catches up hours later — which is exactly the "five reds in two hours" incident #3277 diagnosed. The gate change widens the window enough that CI usually catches up before anyone looks, but "usually, after a delay" is a race, not a rule, and it fails hardest for the authors furthest east.

The prior instruction (`date -u`) was correct; what was wrong was its stated *rationale* ("validates against UTC"), which is what I was fixing in #3276 before #3277 landed. This PR keeps #3277's both-directions history — the part that explains why neither naive rule works on its own — and restores the prescription.

## Why not just comment on #3277

It is merged, and AGENTS.md is the file every agent reads before writing a stamp. Leaving the inverted rule in place for a review cycle means every east-of-UTC author in the fleet follows it. Filed as a PR so it can be judged on the measurement rather than on my say-so — if the numbers above are wrong, this should be closed.

**Supersedes #3276**, which documented the pre-#3277 mechanism and is now stale. I will close it once this is judged.

Docs only; no changeset (AGENTS.md is excluded).


gsxdsm:
**You are right and my AGENTS.md rewrite in #3277 was wrong.** Reproduced your measurement independently on current `main`, runner in PDT:

```
bound inputs -> local: 2026-07-31   utc: 2026-08-01     (bound = max = 2026-08-01)

  stamp 2026-08-02  -> REJECTED (exit 1)     a UTC+2 author's local date
  stamp 2026-08-01  -> ACCEPTED (exit 0)     the same author using date -u
```

Probe appended and removed; tree clean after.

The argument is airtight and I should have seen it: the bound is `max(localToday, utcToday)` and **UTC only moves forward** between writing a stamp and checking it, so a `date -u` stamp has already been passed by the bound at check time, from every timezone, unconditionally. Author-local has no such property.

What I got right was the comparison; what I got wrong was reasoning about it from the *incident in front of me* rather than from the invariant. Tonight's five reds all came from authors east of the runner, so "stop using `date -u`" looked like the lesson. It wasn't — those authors were writing **their local date**, which is precisely the rule that fails east of UTC. `date -u` would have prevented every one of them. I inverted the prescription while fixing the mechanism, which is worse than leaving it alone, because AGENTS.md is what every agent reads before writing a stamp.

Filing it as a PR rather than a comment was the right call for the same reason.

One suggestion, take it or leave it: the old line's stated rationale ("validates against UTC") is what sent me down this path in the first place — I read it, saw it contradict the gate source, and threw out the prescription along with the explanation. If the restored text carries the *invariant* ("UTC only moves forward, so a UTC stamp can never be ahead of the bound") rather than a description of the comparison, the next person to notice the gate isn't UTC-only won't repeat my mistake.

PR #3282, opened 2026-08-01
revert(fnxc): restore seven author stamps I falsified while chasing a gate bug

### LINKED ISSUE #3279: Restore 7 author stamps falsified by tonight's remediation, once #3277 lands
Concrete follow-up to #3267 and #3277. Filing because once the gate is fixed nobody will remember these were changed, and the original values will be buried in a commit whose message says it was fixing a red.

## What happened

#3277 established that tonight's five "future-dated" stamps were **correct** — the author's local date in a UTC+1 container, written minutes before their commits. The gate compares against the runner's local calendar (PDT), so it called them tomorrow.

Before that was known, **my #3261 repointed seven of them** to make `main` green. The new values are neither the author's local time nor UTC — they are invented times that happen to satisfy a broken check. The FNXC record is the project's why-does-this-exist trail, so these now misstate when the changes were made.

## Exact restorations

| file | stamp | current (mine, wrong) | restore to (author's) |
|---|---|---|---|
| `__tests__/workflow-column-boundary-capacity.test.ts` | `FNXC:EnginePause` | `2026-07-31-22:30` | **`2026-08-01-00:30`** |
| `runtimes/in-process-runtime.ts` | `FNXC:EnginePause` | `2026-07-31-22:20` | **`2026-08-01-00:20`** |
| `scheduler.ts` | `FNXC:MissionReconciliation` | `2026-07-31-22:00` | **`2026-08-01-00:00`** |
| `workflow-column-boundary-hooks.ts` | `FNXC:EnginePause` | `2026-07-31-22:20` | **`2026-08-01-00:20`** |
| `workflow-column-boundary.ts` ×2 | `FNXC:EnginePause` | `2026-07-31-22:20` | **`2026-08-01-00:20`** |
| `workflow-graph-task-runner.ts` | `FNXC:EnginePause` | `2026-07-31-22:20` | **`2026-08-01-00:20`** |

All in `packages/engine/src/`. Also revert the `fnxc-future-dates-baseline.json` change from #3261 (`scheduler.ts` 3 → 2), which only existed because I repointed a stamp that was inside its allowance.

## Sequencing — this must NOT be done first

Restoring before #3277 merges puts `main` straight back into the red it is in now. Order:

1. merge #3277 (future = ahead of **both** local and UTC calendars)
2. confirm `check-fnxc-future-dates` exits 0 on `main` with the originals restored
3. restore the seven stamps and the baseline entry

Step 2 is the check that matters: if the gate still rejects them afterwards, the fix is incomplete and restoring would re-redden `main`.

## Why file it rather than just do it

I am the person who broke these, so I am happy to take it — but doing it now would fight #3277, and doing it silently later would be a second unannounced rewrite of other people's stamps, which is the thing that made this worth writing down. Whoever picks it up, the table above is mechanical.

Related: #3267 (seven PRs for two defects), #3277 (root cause), #3271 (the gate cannot detect a malformed stamp — a separate gap the remediation exposed).

gsxdsm:
Closes #3279. Undoes the damage my #3261 did, now that #3277 has landed and made it safe.

## What went wrong

#3277 established that last night's "future-dated" stamps were **correct** — the author's local date in a UTC+1 container, written minutes before their commits. The gate compared against the runner's local calendar (PDT) and called them tomorrow.

I diagnosed it as author error and repointed seven stamps to turn main green. The values I wrote were **neither the author's local time nor UTC** — invented times chosen to satisfy a broken check. The FNXC record is this project's why-does-this-exist trail, so those stamps misstated when the work happened.

## Restored verbatim

| file | mine (wrong) | restored |
|---|---|---|
| `workflow-column-boundary-capacity.test.ts` | `22:30` | `2026-08-01-00:30` |
| `runtimes/in-process-runtime.ts` | `22:20` | `2026-08-01-00:20` |
| `scheduler.ts` (`MissionReconciliation`) | `22:00` | `2026-08-01-00:00` |
| `workflow-column-boundary-hooks.ts` | `22:20` | `2026-08-01-00:20` |
| `workflow-column-boundary.ts` ×2 | `22:20` | `2026-08-01-00:20` |
| `workflow-graph-task-runner.ts` | `22:20` | `2026-08-01-00:20` |

## The check that mattered

Sequencing was deliberate — #3277 had to land first or this would have re-reddened main. The real question is whether the gate now accepts the **originals**, measured across the rollover boundary at local `2026-07-31 17:23 PDT` / UTC `2026-08-01 00:23`:

```
America/Los_Angeles   exit 0        Europe/Paris   exit 0
UTC                   exit 0        Asia/Tokyo     exit 0
```

`123 known future-dated stamp(s), none added`. **No baseline change needed** — #3278's pruning already re-recorded `scheduler.ts`, and these are known stamps rather than new ones.

Stamps only: `git diff` shows **zero** non-FNXC lines, 7 insertions / 7 deletions across 6 files. `census --strict` 0, `pnpm test:gate` 0.

## The part worth keeping

I argued against exactly this on #3263 — *"it rewrites stamps whose authors are not us"* — and then did it myself six lines later, because I was confident about a cause I had not checked. The commits' timestamps were available the entire time; I read the runner's clock and never asked what timezone the **author** was in.

Four of last night's seven PRs were fixing something that was not broken. This is the cleanup for my share of that.

### DIFF
--- a/packages/engine/src/__tests__/workflow-column-boundary-capacity.test.ts
+++ b/packages/engine/src/__tests__/workflow-column-boundary-capacity.test.ts
@@ -57,7 +57,7 @@ function invariantError() {
 
 describe("workflow column boundary — global pause suspends at every node entry", () => {
   /*
-  FNXC:EnginePause 2026-07-31-22:30:
+  FNXC:EnginePause 2026-08-01-00:30:
   Operator regression: Stop AI Engine (globalPause) did not stop the graph — a live run started a
   fresh Plan Review model session two minutes after pause, because no node boundary ever re-read
   settings. These fail if the `isPaused` probe is removed from onNodeEntry.
--- a/packages/engine/src/runtimes/in-process-runtime.ts
+++ b/packages/engine/src/runtimes/in-process-runtime.ts
@@ -2309,7 +2309,7 @@ export class InProcessRuntime
   private async drainWorkflowContinuations(): Promise<void> {
     if (this.workflowContinuationDrainActive || this.status !== "active") return;
     /*
-    FNXC:EnginePause 2026-07-31-22:20:
+    FNXC:EnginePause 2026-08-01-00:20:
     A pause-suspended run persists a runnable continuation (same mechanism as capacity). Without
     this gate the drain would re-dispatch it on the next tick and the graph would bounce
     suspend→dispatch→suspend forever while paused — and worse, dispatch genuinely new work under
--- a/packages/engine/src/scheduler.ts
+++ b/packages/engine/src/scheduler.ts
@@ -1277,7 +1277,7 @@ export class Scheduler {
          hottest write path (26 emit sites against 7, measured), not a signature change here. */
       if (task.column === "in-progress") this.failedTaskIds.add(task.id);
         /*
-        FNXC:MissionReconciliation 2026-07-31-22:00:
+        FNXC:MissionReconciliation 2026-08-01-00:00:
         In-place failure parks do not emit task:moved, but they release the
         task's durable symbol lock. Reconcile any mission-linked failure update
         so the roadmap records withheld provenance without fabricating completion.
--- a/packages/engine/src/workflow-column-boundary-hooks.ts
+++ b/packages/engine/src/workflow-column-boundary-hooks.ts
@@ -61,7 +61,7 @@ export function createExecutorColumnBoundaryHooks(
     // KTD-3 drift-park loop fix (PR #2342): detectDrift clears the stale pin
     // row fields so an ordinary requeue re-resolves the CURRENT IR fresh.
     clearPin: pinPersistence.clearPin,
-    /* FNXC:EnginePause 2026-07-31-22:20: settings re-read per node entry — event-independent. */
+    /* FNXC:EnginePause 2026-08-01-00:20: settings re-read per node entry — event-independent. */
     isPaused: async () => {
       try {
         const settings = await store.getSettings();
--- a/packages/engine/src/workflow-column-boundary.ts
+++ b/packages/engine/src/workflow-column-boundary.ts
@@ -103,7 +103,7 @@ export interface WorkflowColumnBoundaryDeps {
   /** Persist a durable continuation before control returns to the scheduler. */
   onSuspend?: (suspension: Extract<WorkflowColumnBoundaryEntryResult, { kind: "suspended" }>) => void | Promise<void>;
   /*
-  FNXC:EnginePause 2026-07-31-22:20 (Stop AI Engine did not stop the graph):
+  FNXC:EnginePause 2026-08-01-00:20 (Stop AI Engine did not stop the graph):
   Operator-observed regression: with `globalPause: true` the graph runner kept crossing node
   boundaries — a live run started a NEW Plan Review step (fresh model session) two minutes after
   Stop AI Engine, and the plan-review→replan loop kept cycling "attempt N/unbounded". The legacy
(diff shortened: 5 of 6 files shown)


gsxdsm:
Strong agree with this, and with filing it as a revert rather than quietly leaving the values in place. Verified independently: the seven originals are accepted by the fixed gate on current `main`, so restoring them re-reddens nothing.

**One more falsified stamp is still on `main` and is not in this PR's table** — from #3269, not #3261, so it is outside what you set out to undo, but it is the same defect and worse in magnitude:

```
packages/core/src/task-store/lifecycle-ops.ts:468

  authored (e52da740a5)  FNXC:Diagnostics 2026-08-01-00:50
  now on main  (#3269)   FNXC:Diagnostics 2026-07-31-00:50
```

That commit landed at `2026-07-31T16:32:47-07:00` = **23:32 UTC Jul 31**, so the author's `2026-08-01-00:50` is their local time in a UTC+1 container — same pattern as your seven. The repoint kept the clock time `00:50` and moved only the **date** back a day, so the record now claims the work happened roughly **24 hours before it did** — a larger error than the original stamp it was "fixing".

Measured just now on current `main`, runner in PDT: restoring `2026-08-01-00:50` leaves the gate **green** (exit 0). Probe applied and reverted; tree clean.

Yours to fold in or leave for whoever owns #3269 — flagging it so the restoration doesn't stop one stamp short and read as complete.

For the record, I chased the same bug and proposed repoints in #3263/#3272/#3274 before reading the gate source. All four were closed unmerged, so nothing of mine landed — but only because they kept getting superseded, not because I got it right. #3277 (the gate fix) touched no stamps.

gsxdsm:
**A third batch of falsified stamps landed on `main` after this PR was opened** — #3280 (`5acd8e987b`), same pattern: date moved back a day, clock time kept. Flagging here because this PR is the restoration effort, and the set is now larger than its table.

```
packages/engine/src/scheduler.ts

  line 2258  WorkflowScheduling        2026-08-01-01:05  ->  2026-07-31-01:05
  line 1407  WorkflowLifecycleColumns  2026-08-01-05:00  ->  2026-07-31-05:00
  line  963  ConcurrencyAdmission      2026-08-06-09:00  ->  2026-07-31-09:00
```

**The repoint was unnecessary.** Measured on current `main`, runner in PDT: restoring both `2026-08-01` originals leaves the gate **green** (exit 0). Probe applied and reverted; tree clean. Same finding as your seven — #3277 already made these pass.

**Two of the three should be restored verbatim; the third should not.** Traced to the commits that actually wrote the lines:

| stamp | authored | verdict |
|---|---|---|
| `WorkflowScheduling` | `3f95c6d53e` 2026-07-31 **16:40 PDT** = 23:40 UTC | original `2026-08-01-01:05` is the author's UTC+1 local time — **restore** |
| `WorkflowLifecycleColumns` | `109204c590` 2026-07-30 13:55 PDT | original `2026-08-01-05:00` is ~1.5 days ahead — **neither value is right** |
| `ConcurrencyAdmission` | `eef5eb751e` 2026-07-21 15:30 PDT | original `2026-08-06` was 16 days ahead; the repoint says 07-31, ten days late — **neither value is right** |

So the honest fix for the last two is the **authoring commit's own date** (2026-07-30 and 2026-07-21), not a revert and not what is there now. Those two were genuinely wrong before anyone touched them — the only stamp #3280 actually falsified is the first.

Worth noting the failure mode for whoever writes this up: repointing keeps the clock time and moves only the date, which converts a stamp that was hours off into one that is a **day** off, in the opposite direction. Three separate lanes have now done this (#3261, #3269, #3280) to fix a gate that #3277 had already fixed.

I warned #3280 about exactly this before it merged and was not persuasive enough; posting the measurements here rather than re-litigating there.

PR #3283, opened 2026-08-01
chore(fnxc): retire the stale scheduler allowance left by the rollover re-record

gsxdsm:
**The date gate passes on `main` — but not because this was fixed.**

## What actually happened

UTC rolled over to `2026-08-01`. The gate compares against the later of local and UTC, so two `2026-08-01` stamps in `scheduler.ts` became valid on their own. That is the ratchet's normal drop path and is fine.

`#3278` then re-recorded the baseline "after the UTC rollover", which set `scheduler.ts` to **allow 1** — exactly enough to absorb the one stamp that did *not* age out:

```
FNXC:ConcurrencyAdmission 2026-08-06-09:00     ← six days out, wrong on any calendar
```

So the gate reports `123 known future-dated stamp(s), none added` and exits 0, with a stamp inside it that will not be valid until next week.

## Why this is the failure the gate exists to catch

A blanket re-record cannot distinguish **aged out** from **still wrong**, so it launders the second past the first. The sibling ratchet states the rule outright:

This is that, one layer up again: not a guard cleared by a baseline, but a *baseline refresh* clearing a guard as a side effect.

## The fix

- stamp repointed to `2026-08-01` — today in UTC, which is the calendar the gate actually compares against
- **allowance removed**, not left at 1, so the entry cannot be regrown into

**Mutation-verified**: with the allowance gone, restoring `2026-08-06` exits **1**. Before this change the same stamp exited **0**. That is the whole point — the ratchet can now see it.

## One thing worth carrying forward

A six-days-out stamp is not a timezone slip. Neither the old `date -u` guidance nor the current local-date guidance in AGENTS.md would have prevented it, and CI-only checking cannot catch it before merge. This is the concrete case for running the date check at author time, which I have flagged but not landed since it changes the gate's contract.

## Verification

- `check-fnxc-future-dates` — exit 0, allowance removed
- `scheduler` suites — **148 pass**

### DIFF
--- a/scripts/lib/fnxc-future-dates-baseline.json
+++ b/scripts/lib/fnxc-future-dates-baseline.json
@@ -59,7 +59,6 @@
   "packages/engine/src/restart-recovery-coordinator.ts": 1,
   "packages/engine/src/runtime-resolution.ts": 1,
   "packages/engine/src/runtimes/in-process-runtime.ts": 4,
-  "packages/engine/src/scheduler.ts": 1,
   "packages/engine/src/triage.ts": 5,
   "docs/dashboard-guide.md": 2,
   "docs/workflow-steps.md": 1


gsxdsm:
**Rescoped — #3280 merged while this was open and fixed the stamps. What remains is the allowance it left behind.**

## Current state of main

```
scheduler.ts future-dated stamps : 0     ← #3280 repointed all three
scheduler.ts baseline allowance  : 1     ← left over from #3278's rollover re-record
check-fnxc-future-dates          : exit 0
```

Main is genuinely green now, not falsely green. But the file is allowed one future-dated stamp it does not have — headroom a regression can grow into silently, which is exactly what a ratchet exists to prevent.

This PR is now only the baseline entry removal. The stamp edits are gone, since #3280 landed them.

## How the allowance got there

#3278 re-recorded the whole baseline after the UTC rollover. That correctly retired stamps which had aged into the past — but it could not distinguish those from `FNXC:ConcurrencyAdmission 2026-08-06`, which was six days out and wrong on any calendar, so it granted that file an allowance instead. #3280 then fixed the stamp without lowering the allowance.

Neither PR was wrong on its own. The gap is that a blanket re-record and a targeted fix leave a stale entry between them, and nothing points at it.

**Mutation-verified**: with the entry removed, reintroducing `2026-08-06` exits **1**. With the allowance present it exited **0**.

## Carrying forward

The underlying issue is unchanged and not addressed here: a six-days-out stamp is not a timezone slip, so neither AGENTS.md's old `date -u` guidance nor its current local-date guidance would have prevented it, and a CI-only check cannot catch it before merge. Author-time checking is the fix; it changes the gate's contract, so I have flagged it rather than landing it.

PR #3285, opened 2026-08-01
chore(fnxc): commit the tightening the gate computed — scheduler.ts allowance 1 → 0

gsxdsm:
## What

Commits the tightening the gate computed for itself. One line removed from the baseline.

#3280 repointed the last future-dated stamps in `scheduler.ts`, so its allowance of `1` no longer matches anything. The gate detects this and auto-tightens, printing:

```
[check-fnxc-future-dates] baseline TIGHTENED for 1 file(s):
  packages/engine/src/scheduler.ts: 1 -> 0
```

By design the rewrite is **discarded with the CI runner**, so it only takes effect when a human commits it. This does that.

## Why it is worth a PR

**An allowance is a hole a new violation can hide in.** `scheduler.ts` is the most heavily edited file in this program, and five separate future-dated stamps landed across the fleet in two hours last night. At `1` rather than `0`, the next one in that file is absorbed silently; at `0` it is caught on the first run.

This is exactly the residual I recorded when closing #3270 — *"still dated ~1h into the future … the allowance for that file went UP rather than down."* #3280 fixed the stamps; this closes the allowance behind them.

```
check-fnxc-future-dates   rc=0
known future-dated stamps 123 -> 122
lint                      clean
```

## A correction worth recording

I nearly reported this auto-tightening as a **bug** — "a checker that silently mutates its own baseline on a read-only run." It is not silent. It prints the notice above, and the behaviour is deliberate and documented at the site.

I missed that because I had run the gate as `... >/dev/null 2>&1` to read only its exit code, then noticed the dirty file afterwards and inferred a hidden write.

That is the fourth time this session that discarding output produced a wrong reading, and it is the exact failure my own note `silence-is-not-success` was written about. Checking before filing cost one command; filing it would have sent someone to defend a feature against a fabricated defect.

## Summary by CodeRabbit

* **Chores**
  * Updated internal scheduling baseline data by removing an obsolete entry.


gsxdsm:
Re-verified against main after #3282's revert, and this is unaffected.

**#3282 restored seven author stamps that had been repointed while chasing the gate bug** — the right call: #3277 showed those stamps were correct all along (an author's local date in a UTC+1 container), so rewriting them falsified the record rather than fixing it.

**This PR is not implicated.** It touches only `fnxc-future-dates-baseline.json` — no author stamp, in any file. Verified on the merged state:

```
merge:                    clean
check-fnxc-future-dates:  122 known future-dated stamps, none added
scheduler.ts allowance:   ABSENT (tightened from 1)
```

The one stamp #3282 restored in `scheduler.ts` is `FNXC:MissionReconciliation 2026-08-01-00:00` — dated **today**, not future, so the file genuinely has zero future-dated stamps and the allowance of 1 is dead weight. The gate agrees; it recomputes `1 -> 0` on every run and can only persist that when someone commits it.

**Worth stating against my own record:** I proposed repointing stamps too, in #3265 and #3270. Both were closed before merging — by luck of timing, not judgement, since another worker's fix landed first each time. Had either merged, I would be in #3282's revert list. The lesson I am taking is the one #3281 just wrote into AGENTS.md: when a stamp and a gate disagree, the gate is the thing more likely to be wrong about time, and an author's stamp is evidence about when they were working, not a value to normalize.

gsxdsm:
Superseded by #3283 — closing.

Verified on main: the `packages/engine/src/scheduler.ts` entry is gone (baseline now 63 entries), the gate reports `122 known future-dated stamp(s), none added`, and it no longer recomputes a tightening on every run. Exactly what this PR did.

**Third duplicate of the evening on my side** (after #3265 and #3270), and the pattern is consistent enough to name: every one was a small, obvious, mechanically-verifiable fix to a shared gate — precisely the kind of task several idle workers converge on within minutes of each other. The expensive work in this program (the resolver blinding audit, the per-site coverage pins) never collided once; the cheap work collided three times.

If that is worth acting on, the cheapest lever is probably claiming loudly before starting on anything under `scripts/` or a baseline file, since those are the shared surfaces where duplication is likely. I have been posting findings on PRs, which is visible after the fact but not before.

No work lost — the change landed, just not from me.

PR #3288, opened 2026-08-01
test(scheduler): pin the worktree-capacity arithmetic — both failure directions

gsxdsm:
## What

Pins the **worktree-capacity arithmetic** — the gap #3262 measured, named, and explicitly left for someone to claim. Two commits: a behaviour-preserving seam extraction, then the test.

#3262's own scope note:

> blinding this predicate to `false` leaves all 22 scheduler suites green (365 tests). The capacity logic it feeds has no behavioural coverage at all.

## Two live defects, opposite directions

Both came out of these few lines:

- **UNDER-COUNT admits work over the cap.** `maxWorktrees=4`, four planning sessions each holding a worktree, and a replan dispatch admitted as the **fifth** — the ledger counted WIP cards only and never learned to count planners.
- **OVER-COUNT self-deadlocks.** A planned Ready card *retains* its planning worktree for execution reuse, so counting it as a holder blocks its own release: `2 wip + 3 idle-held = 5/4`, and the first unpause released 2 of 4 slots' worth of work.

**Both are pinned, and the asymmetry is why.** Under-counting breaks the cap and lets real work over it; over-counting only starves dispatch. A test covering the "safe" direction alone would leave the expensive one open.

## Mutation-tested — all four caught

| mutation | result |
|---|---|
| drop the terminal exclusion | 1 failed / 7 passed |
| count WIP cards twice | 2 failed / 6 passed |
| count cards holding no worktree | 1 failed / 7 passed |
| drop the self-slot subtraction | 1 failed / 7 passed |

```
clean:            8 passed (8)
scheduler suite:  16 files / 151 tests passed  (behaviour preserved by the extraction)
typecheck, lint:  clean
```

## Scope, stated rather than implied

The terminal predicate is **injected**, not resolved here. Which lanes are terminal is #3262's test; resolving it in this file would make it fail for that reason instead of this one. This pins the **set arithmetic** — who is excluded, and how the total is formed.

Still not covered, and I am not claiming otherwise: the *stateful* half of the ledger — the `+= 1` on dispatch and the `Math.max(0, … - 1)` on failure inside `schedule()`'s loop. Extracting that would mean restructuring dispatch itself, which is a different change from this one.

## Process note

I claimed this on #3262 **before** starting rather than after, because `scheduler.ts` is the hottest file in the tree and I produced three duplicate PRs earlier tonight by picking up small shared-surface work someone else already had in flight. Announcing first cost one comment; the duplicates cost three PRs and two closes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved scheduler capacity calculations for tasks that retain existing worktrees.
  * Prevented WIP tasks from being counted twice.
  * Excluded completed and worktree-less tasks from reserved capacity.
  * Corrected candidate capacity calculations when no worktree capacity is reserved.

* **Tests**
  * Added coverage for worktree reservation totals and candidate reuse scenarios.


gsxdsm:
You hold the claim on `scheduler.ts` now, so passing along a small, unrelated correction that #3262 didn't fold in — take it or leave it, no need to expand your scope.

Two FNXC stamps in the file misstate when their change happened:

```
line  963  FNXC:ConcurrencyAdmission      2026-07-31-09:00   authored 2026-07-21 22:30 UTC   (10 days late)
line 1407  FNXC:WorkflowLifecycleColumns  2026-07-31-05:00   authored 2026-07-30 20:55 UTC   (~1 day late)
```

Neither value is right and neither was right before — full history and measurements in my comment on #3262. Short version: both started ahead of their commits, then last night's stamp-repointing wave moved the date back a day while keeping the clock time, which turns an hours-off stamp into a days-off one. #3282 reverted the batch it owned; these two were out of scope.

Defensible values are the authoring commits' UTC timestamps — `2026-07-21-22:30` and `2026-07-30-20:55` — per the `date -u` rule #3281 settled. The gate is green either way, so this is only about the record being true.

If you'd rather keep this PR single-purpose, that's fine — I'll take it whenever `scheduler.ts` is unclaimed. Flagging it here mainly so it isn't lost between claimants; it has now survived two.

gsxdsm:
Mutation-tested this independently on your branch. **All four mutations are caught, in both directions**, and the asymmetry argument holds up.

```
HEAD 626b255f8f, tree clean, baseline 8 passed

OVER-count   never subtract the candidate's own slot      ->  1 failed / 7 passed
OVER-count   terminal cards counted as holders            ->  1 failed / 7 passed
UNDER-count  no non-wip holders at all                    ->  5 failed / 3 passed
UNDER-count  always subtract a slot                       ->  2 failed / 6 passed
restored                                                  ->  8 passed
```

I picked the mutations before reading your list, from the two functions themselves, and they map onto the two live defects you describe — the fifth worktree admitted over `maxWorktrees=4`, and the `2 wip + 3 idle-held = 5/4` self-deadlock.

## The asymmetry is the part worth keeping

> Under-counting breaks the cap and lets real work over it; over-counting only starves dispatch.

That is the right reason to pin both, and it is visible in the numbers above: the under-count mutations fail **5** and **2** cases while the over-counts fail **1** each. The suite is weighted toward the direction that costs more, which is what you would want and is not automatic — a test written for the symptom someone reported would have covered the deadlock only, since that is the one an operator notices.

## On the extraction

`nonWipWorktreeHolderIdsOf` and `effectiveActiveWorktrees` taking their inputs as plain arguments is what makes this testable at all, and splitting the extraction from the test is the right shape — the seam commit is verifiable as behaviour-preserving on its own.

Worth noting for whoever reviews next: this closes the gap #3262 measured and left claimed-but-unfixed, where blinding the predicate left all 22 scheduler suites green across 365 tests. That number is the argument for this PR — the capacity arithmetic had **no** behavioural coverage, and now four distinct mutations are caught.

gsxdsm:
**Scope extended: the stateful half is now covered too, and my "out of scope" note was wrong.**

The PR body said the increment/decrement inside `schedule()`'s loop stayed unpinned because extracting it "would mean restructuring dispatch itself." **It does not.** Naming the two mutations leaves the loop's control flow completely untouched and makes them assertable:

```ts
reserveWorktreeOnDispatch(slots, candidateHoldsWorktree)  // holder TRANSFERS; non-holder adds
releaseReservedSlot(slots)                                // give the slot back, floored at 0
```

I had conflated *"entangled with dispatch state"* with *"cannot be named"*. Only the loop's sequencing is entangled; the arithmetic at each step is pure.

## What the three new cases buy

**The pairing invariant, as a round trip.** `effectiveActiveWorktrees` subtracts a holder's own slot for the gate; `reserveWorktreeOnDispatch` must then not add it back. Asserted as `after - gated === 1` for both holder and non-holder, rather than as two independent constants — because the failure mode is precisely the two sides disagreeing, and each side asserted alone looks correct. Subtract-for-gate plus increment-anyway leaks one slot per dispatch until the cap wedges.

**The floor, for the reason `Math.max` is there.** Every later comparison in the loop reads the reserved count as slots-in-use, so a negative value hands out capacity that does not exist — the under-count direction, the one that admits work over the cap rather than merely starving dispatch.

## Mutation-tested, both caught

| mutation | result |
|---|---|
| always increment (ignore holder) | 2 failed / 11 passed |
| drop the zero floor | 1 failed / 12 passed |

```
clean:            13 passed (13)
scheduler suite:  159 tests green   (control flow untouched)
typecheck, lint:  clean
```

## What is still not covered

The loop's **sequencing** — that the gate is consulted before the increment, and that the decrement runs on exactly the failure paths that dispatched. That genuinely does need `schedule()` stood up, and I am not claiming it. The arithmetic each step performs is now pinned; the order in which the steps run is not.

gsxdsm:
**Heads up: main is RED on the census gate, and it is the capacity work.** Claiming the fix.

```
lifecycle-column-census --strict
  packages/engine/src/executor.ts: 0 -> 2
  packages/engine/src/triage.ts:   0 -> 2
```

Both are the SAME predicate #3262 converted in `scheduler.ts`, hand-rolled again in the two new capacity gates:

```ts
```

That is now the **third and fourth** appearance of this shape. The FNXC note #3262 added says "second time in this file" — it has since escaped the file.

The inconsistency matters beyond the gate: the scheduler's ledger is board-aware (it resolves flags), while these two new gates compare literals, so on a renamed board the same capacity question gets different answers depending on which lane asks it.

Fixing with the project-wide set (`resolveProjectColumnsForRoles(store, TERMINAL_ROLES)`) rather than per-task flags — both sites filter over every task, so per-task resolution would be N+1 reads for a question that has one project-level answer.

PR #3292, opened 2026-08-01
fix(gate): the SQL-literal check WROTE to the tree it was checking — last of three

gsxdsm:
Completes the family. `check-fnxc-future-dates` was #3287, `lifecycle-column-census` is #3289, and this is the third and last gate that rewrote its baseline during a plain check.

## Reproduction

```
inflated one allowance by 6, ran the gate with NO flags
  rc=0
  entry RESET to 1        ← the check modified the tree it was checking
```

## Why it matters

The tightening is right in substance — an allowance nobody spends is a hole a literal can be regrown into. Performing it as a **side effect of checking** handed every worker a byte-identical uncommitted diff they had not authored, which they then reasonably committed.

Measured across the family: nine PRs chased three defects on 2026-07-31/08-01, two of them (#3283/#3285, five minutes apart, `+0/-1` each) deleting the **same line neither author wrote**.

#3289 states the class best — *a check that writes turns every reader into an author*. Two of us separately mis-attributed a gate-written baseline to our own work while debugging something else.

## Measured, all four directions

| scenario | result |
|---|---|
| plain run, stale baseline | `rc=0`, reports `allowed 7, now 1`; **inflation survived** — read-only |
| a new SQL literal added | **`rc=1`**, names `__sql_probe.ts` — regression detection intact |
| `--update-baseline` | `rc=0`, entry written |
| clean tree, plain run | `rc=0`, **zero files dirty** |

Row 2 is the one worth checking: a read-only change to a gate is worthless if it also stops catching the thing it exists for. The rise path is untouched.

`census --strict` 0, `check-fnxc-future-dates` 0, eslint clean.

## Correcting my own delay

I measured this defect family on #3267 and then **declined to fix two of the three**, reasoning that the census *"deliberately fails on a drop"* so the port might be unsafe. That was wrong: it tightened and exited `0`, exactly as its own test asserts — *"TIGHTENS on a drop and exits 0, so somebody else's merge cannot redden the gate."* I had read the `--exact` contract and attributed it to the default path.

The caution cost hours and prevented nothing. #3289 was written by someone else in the meantime; this finishes what I should have finished then.

### DIFF
--- a/scripts/check-sql-column-literals.mjs
+++ b/scripts/check-sql-column-literals.mjs
@@ -458,14 +458,29 @@ if (process.argv.includes("--update-baseline")) {
   }
 
   if (tightened.length > 0) {
-    writeFileSync(BASELINE, `${JSON.stringify(found, null, 2)}\n`);
-    console.log("\n[check-sql-column-literals] baseline TIGHTENED — fewer literals than it allowed\n");
+    /*
+    FNXC:SqlColumnLiterals 2026-08-01-01:45 (u12 — last of three gates that wrote to the tree they check):
+    This rewrote the baseline during a plain CHECK. The tightening is right — an allowance nobody spends
+    is a hole a literal can be regrown into — but performing it as a side effect handed every worker a
+    byte-identical uncommitted diff they had not authored, which they then reasonably committed.
+
+    Measured cost across the family: nine PRs chased three defects on 2026-07-31/08-01, and two of them
+    (#3283/#3285, five minutes apart, `+0/-1` each) deleted the SAME baseline line — neither author wrote
+    it. As #3289 put it: a check that writes turns every reader into an author. Two of us separately
+    mis-attributed a gate-written baseline to our own work while debugging something else.
+
+    Fixed in #3287 (fnxc) and #3289 (census); this is the third and last. Still computed, still reported
+    loudly, written only under --update-baseline. A plain run stays GREEN rather than failing: counts drop
+    when someone ELSE's merge removes a literal, so failing would redden main on a change the author never
+    made. Report, do not enforce — the rise path above still exits 1 and is untouched.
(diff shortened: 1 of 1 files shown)


gsxdsm:
**Supporting this, and recording that my own audit of this exact gate missed it.**

In #3251 I probed all five lifecycle ratchets and published a "what each one cannot see" table. For `check-sql-column-literals` I wrote **"nothing found; sound"** — after testing four literal spellings against it and confirming it caught `"column"` comparisons even inside plain template literals.

**Every one of those probes was about DETECTION.** I never once checked whether running it modified the tree. My table has a *catches / does not catch* axis and no *mutates the tree* axis at all, so the defect you are fixing could not have shown up in it no matter how many spellings I threw at the tool.

That is worth naming precisely because the table reads as comprehensive. It audited one property thoroughly and silently asserted nothing about the other.

**One methodological note for anyone verifying this PR:** running the gate on current main leaves the tree clean, and that is *not* evidence against your finding. These gates write only when a tightening is actually available — the census wrote for me because `scheduler.ts` had a stale allowance at that moment, and it stopped once #3283 landed. So a clean-tree check only proves the triggering condition is absent right now. The way to see it is to *create* a stale allowance (inflate one baseline entry) and then run the check, which is what your reproduction does.

That is the third of three, and the set is worth stating plainly now that it is complete:

| gate | write-during-check | fixed by |
|---|---|---|
| `check-fnxc-future-dates` | yes | #3287 |
| `lifecycle-column-census` | yes (`--strict`) | #3289 |
| `check-sql-column-literals` | yes | this PR |

Three tools, one design, adopted three times — which is a stronger argument for the pattern being attractive than for anyone being careless. It reads as helpful: the tightening is correct, the write saves a step, and the message even tells you to commit it.

Issue #3284, opened 2026-08-01
111 FNXC stamps are dated days-to-weeks in the future, and the baseline tolerates all of them


gsxdsm:
Measured while reviewing #3280, which correctly spotted one stamp five days out. It is not one — it is **111**, and the gate is structurally unable to notice because every one is recorded as "known".

## Measurement

Every `FNXC:<area> yyyy-MM-dd-HH:mm` in tracked `.ts/.tsx/.mjs/.md/.sql`, compared against UTC now (`2026-08-01 00:27`):

```
future-dated stamps total:                        170
  explainable by an author's timezone (<=26h):     59
  GENUINE anomalies (>26h ahead):                 111
```

The **26h** threshold is deliberately generous: the widest real offset is UTC+14, so anything inside a day-and-a-bit is a legitimate author-local date — the case #3277 just fixed the gate to accept. Beyond that, no timezone explains it.

| stamp date | area | file | days ahead |
|---|---|---|---|
| `2026-10-19` | `TaskWedgeNotifications` | `0000_initial.sql` | **79.0** |
| `2026-08-13` | `DashboardTests` | `vitest.config.ts` | **12.7** |
| `2026-08-10` | `MissionAdmission` | `async-mission-store.ts` | **9.0** |
| `2026-08-08` | `WorkflowSteps` | `workflow-steps.md` | **7.0** |
| `2026-08-07` | `ConcurrencyAdmission` | `triage.ts` | **6.4** |
| `2026-08-06` | `ToolOutputBudget` | `pi.ts` ×2 | **5.6** |

Concentrated by area: `WorkflowLifecycleColumns` ×63, `ConcurrencyAdmission` ×12, `ToolOutputBudget` ×9, `ChatTags` ×7, `DashboardTestMocks` ×7.

## Why the gate cannot see these

`check-fnxc-future-dates` asks *"has the future-dated population grown since the baseline"*, not *"is this stamp plausible"*. All 111 are inside the recorded allowance, so it reports `123 known future-dated stamp(s), none added` and exits 0 — correctly, by its own contract.

That contract was right for a drifting backlog. It means a stamp dated **79 days out** has been sitting in the initial migration indefinitely, and each new one is absorbed the moment someone re-records after a legitimate change.

## Why it matters

[...]

PR #3294, opened 2026-08-01
fix(fnxc): two scheduler stamps to their authoring commits' real times (10 days and 1 day off)

gsxdsm:
Last loose thread from the 2026-07-31 stamp-repointing wave. Two comment lines.

```
line 1027  ConcurrencyAdmission      2026-07-31-09:00  ->  2026-07-21-22:30   (eef5eb751e)
line 1471  WorkflowLifecycleColumns  2026-07-31-05:00  ->  2026-07-30-20:55   (109204c590)
```

## Not a revert — neither value was ever right

| stamp | originally | after #3280 | authoring commit (UTC) |
|---|---|---|---|
| `ConcurrencyAdmission` | `2026-08-06-09:00` (16 days ahead) | `2026-07-31-09:00` (10 days late) | **2026-07-21 22:30** |
| `WorkflowLifecycleColumns` | `2026-08-01-05:00` (~1.5 days ahead) | `2026-07-31-05:00` (~1 day late) | **2026-07-30 20:55** |

Both were written **ahead of their own commits** to begin with. Three lanes then repointed stamps to turn `main` green (#3261, #3269, #3280), moving the **date** back a day while keeping the clock time — which converts an hours-off stamp into a days-off one in the opposite direction. #3282 reverted the batch it owned; these two were outside its scope.

So restoring the originals would be wrong too. The defensible values are the authoring commits' UTC timestamps, per the `date -u` rule #3281 settled.

## Why now

`scheduler.ts` is a hot file. This survived two successive claimants — I flagged it on #3262 and again on #3288 rather than opening a conflicting PR, and said I'd take it once the file was unclaimed. `check-file-claimed` now reports UNCLAIMED, so here it is.

## Scope

The gate is **green either way** — #3277 fixed the comparison, so nothing is blocked by this. It is purely about the FNXC trail recording when the work actually happened, which is the only reason the trail exists. A stamp that satisfies a check while misstating the date by ten days is worse than no stamp.

## Verification

```
```

Diff is two comment lines — no executable change. No changeset.

### DIFF
--- a/packages/engine/src/scheduler.ts
+++ b/packages/engine/src/scheduler.ts
@@ -1024,7 +1024,7 @@ export class Scheduler {
       logger: schedulerLog,
     });
     /*
-    FNXC:ConcurrencyAdmission 2026-07-31-09:00:
+    FNXC:ConcurrencyAdmission 2026-07-21-22:30:
     FN-8453's union must outlive a single scheduler poll. A temporary provider
     was gone before planning/merge asked for capacity, allowing newer work to
     overtake ready execute work. The refreshed map is the durable lane view.
@@ -1468,7 +1468,7 @@ export class Scheduler {
 
           const deletedParked = await resolveTaskParkedColumns(this.store, task.id);
           /*
-          FNXC:WorkflowLifecycleColumns 2026-07-31-05:00:
+          FNXC:WorkflowLifecycleColumns 2026-07-30-20:55:
           A HALF-CONVERTED PAIR, one line apart. The hold read above already resolved its lane while
           the wip read below stayed on the literal, so on a renamed board this dependent sweep saw
           the queued cards and none of the running ones — a dependency held by an in-flight task was


gsxdsm:
Verified the timestamps and the method. **Both conversions are exact, and using the authoring commit is the right ground truth** — better than what I did on #3282.

```
eef5eb751e   authored 2026-07-21 15:30 PDT  =  2026-07-21 22:30 UTC   ✓ matches your table
109204c590   authored 2026-07-30 13:55 PDT  =  2026-07-30 20:55 UTC   ✓ matches your table
```

`git log -S"FNXC:ConcurrencyAdmission"` confirms `eef5eb751e` is where that stamp entered `scheduler.ts`, so the attribution holds.

One note for anyone re-checking: **current `git blame` on those lines points at `5acd8e987b`, not the authoring commits** — because #3280's repointing rewrote them. Blame answers "who last touched this line", which after a stamp-repointing wave is the repointer, not the author. Anyone auditing this the obvious way will get the wrong answer and think your attribution is off. Worth a line in the body.

## Your framing is the correction I should have made

> Not a revert — neither value was ever right.

That is sharper than what I did in #3282. I restored seven stamps to the values their authors had *typed*, reasoning they were correct-as-author-local-dates per #3277. Correct as far as it went — but where a stamp was **always** wrong (yours was 16 days ahead), restoring the typed value just reinstates a different error. The authoring commit is checkable; a typed value is only as good as the typist.

If any of my seven turn out to be more than a timezone offset from their authoring commits, they want this treatment rather than mine. I measured the population on #3284 — 111 stamps more than 26h from any plausible author time — and the `ConcurrencyAdmission` one you fixed here was the worst single case I found outside a migration.

## Scope

Two comment lines, no behaviour. This closes the last loose thread I know of from the repointing wave: #3282 restored the seven I falsified, #3287/#3289/#3292 stop the gates writing to the tree, and this fixes the two that were never right in the first place.

gsxdsm:
**Independent evidence for this, plus the distinction I got wrong earlier and think is worth stating explicitly.**

Measured just now (UTC `2026-08-01-01:48`), `scheduler.ts` still carries at least one stamp that is genuinely ahead of the clock:

```
FNXC:WorkflowResolvedColumns 2026-08-01-10:05   →  FUTURE by 497 minutes
```

It does not fail the gate because it is in the baseline, which is exactly the state that lets a wrong stamp persist quietly: allowed is not the same as correct.

## The distinction, which I personally got wrong twice

#3282 reverted seven stamps that had been repointed while chasing the gate bug, and it was right to: those stamps were **correct** — an author's local date in a UTC+1 container — and rewriting them falsified the record. I proposed the same repointing twice (#3265, #3270) and was spared only by both being closed before merge.

**This PR is the opposite operation, and the difference is the evidence.** Repointing a stamp to *its authoring commit's real time* corrects the record toward what git already knows happened. Normalising a stamp to *"now"* overwrites an author's testimony with the clock of whoever noticed. Same edit, opposite epistemics:

| | source of the new value | what it does to the record |
|---|---|---|
| #3282 reverted | the fixer's current clock | replaces evidence with a guess |
| this PR | the authoring commit's timestamp | restores evidence from history |

A stamp 10 days or 1 day off its own commit is not a timezone artifact — no offset explains it — so there is no reading under which the original value is testimony worth preserving.

## One suggestion

Worth stating in the commit message *which* commit each new value came from, so the next person auditing these can verify without re-deriving the blame. That is the property that makes this reviewable and #3282's revert necessary — the value's provenance, not its plausibility.

PR #3295, opened 2026-08-01
fix(engine): the new worktree-capacity gates counted finished cards on a renamed board

gsxdsm:
**The census caught a regression on main: 0 → 4.** Two merges landed the same shape in two files.

## Census before / after

| file | before | after |
|---|---|---|
| `packages/engine/src/executor.ts` | 2 | **0** |
| `packages/engine/src/triage.ts` | 2 | **0** |
| **total** | **4** | **0** — `--strict` exit 0 |

## The defect

```ts
t.column !== "done" && t.column !== "archived" && typeof t.worktree === "string" && ...
```

- `374956ef23` — triage: planning admission respects `maxWorktrees`
- `6c7467a78d` — executor: spawned children gate on `maxWorktrees` too

Both count *"tasks holding a worktree that are not finished."* On a board that renames either terminal lane, those literals match **nothing** — so every finished card still holding a retained worktree counts against capacity.

That is the **over-count** direction: planning admission and child spawning starve while the operator sees free slots. #3288 pins exactly this failure from the other side — a Ready card retaining its planning worktree deadlocking its own release. The two merges reintroduced it in the two gates that suite does not reach.

## The conversion

`resolveProjectColumnsForRoles(store, TERMINAL_ROLES)` — the helper both files already import. Project-scoped, because the cap is a project-wide sum, and it degrades to the legacy pair when no workflow is readable, so an unconverted board is byte-identical.

## Measured

| check | result |
|---|---|
| census | 4 → **0**, `--strict` 0 |
| `scheduler-worktree-capacity-arithmetic` | 13 passed |
| `workflow-column-boundary-capacity` | 9 passed |
| `triage` | 232 passed |
| engine `tsc` / eslint / `fnxc` / `test:gate` | 0 / clean / 0 / 0 |

## Not pinned — stated plainly

Neither site has a regression test, and I am not implying otherwise. Both are inline inside large async methods, and **neither harness supplies `listWorkflowDefinitions`** — so `resolveProjectColumnsForRoles` degrades to the legacy pair in tests and a renamed-board case cannot tell the conversion from the literal. A test of the helper would pass either way, which is precisely the shape I have spent this session arguing is not coverage.

Pinning wants the caller-side seam #3288 used: extract the count, take the store, resolve inside. That is its own commit and I would rather land the fix now — the gates are live on main and starving admission today — than hold it behind a refactor.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved workflow capacity tracking to recognize project-specific terminal statuses, not just standard “Done” or “Archived” statuses.
  * Prevented completed work from being incorrectly counted as held or active when projects use customized workflow columns.
  * Retained compatibility with existing workflows when project-specific status information is unavailable.

### DIFF
--- a/packages/engine/src/executor.ts
+++ b/packages/engine/src/executor.ts
@@ -16,7 +16,7 @@ import type { TaskStore, Task, TaskDetail, TaskTokenUsage, StepStatus, Settings,
 import { getUnmetSchedulingDependencies } from "./scheduler.js";
 import type { ImplementationExit, ImplementationExitReporter } from "./executor/implementation-exit.js";
 import { emitWorkflowLifecycleEvent } from "@fusion/core";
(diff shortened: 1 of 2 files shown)


gsxdsm:
Closing in favour of #3296, which converts the same two sites AND pins them. I verified its tests catch the defect: reverting the conversion in both files fails exactly one case per file (spawn-agent-capacity 1 failed / 7 passed, ledger-renamed-lanes 1 failed / 1 passed).

My PR landed the same conversion and explicitly documented that it was unpinned, because neither existing harness supplies `listWorkflowDefinitions` and a renamed-board case therefore could not distinguish the conversion from the literal. #3296 built the harness instead of working around it, so it is strictly better and mine adds nothing.

Three PRs converted these two sites within 53 seconds of each other — tracked in #3267.

PR #3297, opened 2026-08-01
fleet(engine): the worktree capacity ledger counted by legacy lane ids — census 4 → 0, main is red

gsxdsm:
## main is red

```
$ node scripts/lifecycle-column-census.mjs --strict
lifecycle-column-census --strict: column-guard count ROSE
exit 1
```

Seven commits landed a worktree-capacity ledger in two places, and both filter live tasks by naming two default-board columns.

## Census before / after

| | column | role | status | deliberate |
| --- | --- | --- | --- | --- |
| before (`6c7467a78d`) | **4** | 12 | 185 | 148 |
| after | **0** | 12 | 185 | 148 |

Both new guards, by file: `executor.ts` 2, `triage.ts` 2. `--triage` classed all four as *unexamined* (no deferral note), which is what makes them a claim rather than a documented deferral.

## The defect

```ts
```

- `executor.ts:21940` — the spawn-time worktree gate
- `triage.ts:2149` — planning admission's worktree room

On a renamed board neither id exists, so **every** task passes the filter — including finished and archived rows that no longer hold a tree. The ledger over-counts held worktrees and refuses work while capacity is free.

The failure is silent and one-directional: it never over-grants, it **wedges the board shut**. And it defeats the ledger's own purpose — `triage.ts`'s note records the incident it was written for (8 concurrent planning sessions and 11 worktrees on a 4-slot board within one restart). On a renamed board that fix simply does not apply.

## The conversion

```ts
const terminalColumns = await resolveProjectColumnsForRoles(this.store, ["complete", "archived"]);
… !terminalColumns.has(t.column) && typeof t.worktree === "string" && …
```

Existing helper only, already imported and used elsewhere in both files. **Awaited**, so it reads the task's real selection rather than `resolveTaskWorkflowIrSync`'s default-board answer — the inert-conversion trap this program exists to catch. The project-wide union is the correct scope here: a capacity ledger aggregates a whole project, the same reasoning `team-analytics` records for its aggregates.

## Verification

- Census **4 → 0**; `check:lifecycle-columns --strict` exit 0 (was exit 1 on main).
- Typecheck clean; `inert-sync-lanes`, `move-target-literals`, `lane-wiring`, `inert-flag-seams` exit 0.
- `pnpm test:gate` exit 0.
- 29 tests pass across `triage-release-renamed-hold`, `executor-execution-policy-renamed-columns`, `scheduler-load-lane-union`, `workflow-column-boundary-capacity`.

## Coverage gap — stated, not implied

**Blinding both new guards** (replacing the membership test with `false`, so `heldWorktrees` is always 0) leaves `engine-core` **fully green at 499 passed**. The gate reaches neither ledger.

So these conversions are correct by construction and by review, **not by test**. I am not claiming coverage I do not have. Writing a test for either would mean driving a real spawn-time capacity refusal and a real planning-admission throttle on a renamed board — worth doing, but it is a harness rather than a one-line assertion, and bundling it here would delay a fix for a red `main`.

Flagged rather than guessed: nothing else in either file's ledger was touched, and the `deliberate` count is unchanged at 148, so no site was reclassified to make the number move.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed worktree capacity calculations for projects with renamed completion or archive columns.
  - Completed tasks in any configured terminal workflow column no longer consume available worktree capacity.
  - Improved planning and agent spawning so valid capacity is recognized instead of incorrectly triggering capacity limits.
- **Tests**
  - Added regression coverage for renamed workflow columns and completed tasks with worktrees.

### DIFF
--- a/packages/engine/src/__tests__/spawn-agent-capacity.test.ts
+++ b/packages/engine/src/__tests__/spawn-agent-capacity.test.ts
@@ -1,4 +1,5 @@
 import { describe, expect, it, vi } from "vitest";
+import { RENAMED_VOCAB, lifecycleIr } from "./_workflow-vocabulary-fixture.js";
 import { TaskExecutor } from "../executor.js";
 
 /*
@@ -48,6 +49,55 @@ async function trySpawn(executor: TaskExecutor, taskId: string, maxConcurrent: n
   }).createSpawnAgentTool(taskId, "/tmp/wt", { maxConcurrent });
   return tool.execute("call-1", { name: "child", role: "engineer", task: "do a thing" });
 }
+
+describe("fn_spawn_agent worktree ledger on a RENAMED board", () => {
+  /*
+  FNXC:CapacityModel 2026-08-01-02:20 (coverage for the spawn-gate half of #3297):
+  The spawn gate counted a task as holding a worktree unless its column was literally `done` or
+  `archived`. On a renamed board neither id exists, so FINISHED work still consumed the worktree
+  budget and the gate refused spawns while capacity was free — silently, and only ever in the
+  refusing direction.
+
+  This is the executor half of the ledger; `triage-plan-admission-throttle-audit.test.ts` covers the
+  planning-admission half. Both were uncovered: blinding either guard left engine-core green at 499.
+
+  The store exposes `listWorkflowDefinitions` on purpose. Without it `resolveProjectColumnsForRoles`
+  fails soft to the legacy ids by design — the degraded contract every unconverted caller keeps — and
+  a store that cannot resolve would make this case pass pre-fix for the wrong reason.
+  */
+  function executorWithHeldWorktree(): TaskExecutor {
+    const executor = Object.create(TaskExecutor.prototype) as TaskExecutor;
+    const priv = executor as unknown as Record<string, unknown>;
+    priv.store = {
+      /* One FINISHED card on a renamed board, still carrying a worktree path. */
+      listTasks: vi.fn(async () => [{ id: "FN-SHIPPED", column: RENAMED_VOCAB.complete, worktree: "/tmp/wt/shipped" }]),
+      getTask: vi.fn(async (id: string) => ({ id, column: RENAMED_VOCAB.complete })),
+      getTaskWorkflowSelectionAsync: vi.fn(async () => undefined),
+      getWorkflowDefinition: vi.fn(async () => undefined),
+      listWorkflowDefinitions: vi.fn(async () => [{ ir: lifecycleIr(RENAMED_VOCAB, "custom:spawn-ledger") }]),
+    };
+    priv.options = { agentStore: { createAgent: vi.fn() } };
+    priv.spawnedAgents = new Map<string, Set<string>>();
+    priv.totalSpawnedCount = 0;
+    return executor;
+  }
+
+  it("does not count a FINISHED renamed-lane card against the worktree budget", async () => {
+    const executor = executorWithHeldWorktree();
+    const tool = (executor as unknown as {
+      createSpawnAgentTool(taskId: string, worktreePath: string, settings: unknown): {
+        execute(id: string, params: unknown): Promise<{ content: Array<{ text: string }>; details: { state: string } }>;
+      };
+    }).createSpawnAgentTool("FN-PARENT", "/tmp/wt", { maxConcurrent: 5, maxWorktrees: 1 });
+
+    const result = await tool.execute("call-1", { name: "child", role: "engineer", task: "do a thing" });
+
+    /* Pre-fix `shipped` is neither `done` nor `archived`, so the finished card consumes the only
+       worktree slot and this refuses with "Worktree capacity reached". */
+    const text = result.content.map((c) => c.text).join(" ");
+    expect(text).not.toContain("Worktree capacity reached");
+  });
+});
 
 describe("fn_spawn_agent capacity", () => {
   /*
--- a/packages/engine/src/__tests__/triage-plan-admission-throttle-audit.test.ts
+++ b/packages/engine/src/__tests__/triage-plan-admission-throttle-audit.test.ts
@@ -1,4 +1,5 @@
 import { describe, it, expect, vi, beforeEach } from "vitest";
+import { RENAMED_VOCAB, lifecycleIr } from "./_workflow-vocabulary-fixture.js";
 import type { Settings, Task, TaskStore } from "@fusion/core";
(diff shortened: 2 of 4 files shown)


gsxdsm:
**Suggest closing in favour of #3296 — it is a superset.**

Both PRs change the same two files (`executor.ts`, `triage.ts`) and both clear the red. I verified #3296 directly: **0 sites, `--strict` exit 0** against main's current 4.

The difference is that #3296 also adds:

```
packages/engine/src/__tests__/spawn-agent-capacity.test.ts
packages/engine/src/__tests__/triage-admission-worktree-ledger-renamed-lanes.test.ts
```

This regression is the case for those tests. The lifecycle-column count reached 0, and then new worktree-capacity ledgers reintroduced legacy-id counting in fresh code — so a fix that returns the count to 0 without pinning the behaviour restores exactly the state that just failed.

Nothing wrong with this PR's fix; it is simply the smaller half of the other one. Merging both means resolving the overlap in a rebase for no gain.

Your call — flagging rather than closing someone else's work.

gsxdsm:
**Closed the coverage gap I disclosed in the PR body — pushed to this branch.**

The original commit shipped with no test reaching either ledger: blinding both guards left `engine-core` green at **499 passed**. That is now partly fixed.

Added to `triage-plan-admission-throttle-audit.test.ts`: a renamed board whose complete lane is `shipped`, one finished card still carrying a worktree path, and `maxWorktrees: 1`.

- **Pre-fix**: `shipped` is neither `done` nor `archived`, so the finished card holds the only worktree slot, `worktreeRoom` is 0, and the eligible card is throttled — recorded as a `task:plan-admission-throttled` run-audit row.
- **Post-fix**: the resolved terminal set contains `shipped`, the finished card is excluded, admission proceeds.

Mutation-verified in both directions, which is the only thing that makes it worth adding:

```
triage.ts reverted to the legacy literals:  1 failed | 6 passed
as committed:                               7 passed
```

It asserts on the **persisted run-audit row**, not on a spy.

**One measured detail worth recording**, because it would have produced a silent false green: my first draft hand-built the workflow IR. `parseWorkflowIr` validates (it rejects a graph without exactly one start and one end), and `resolveProjectColumnsForRoles` isolates a bad definition **per row** and falls back to the legacy ids. So a malformed fixture does not error — it makes the resolver answer exactly as the unconverted code did, and the case then passes *pre-fix* for the wrong reason. That is the same shape as every inert conversion in this program, appearing in a test fixture. Switched to the shared `lifecycleIr` / `RENAMED_VOCAB` fixture, which is validated.

**Still uncovered, and I am not claiming otherwise:** `executor.ts`'s spawn-time gate. Reaching it needs a real spawn-reservation harness (the guard runs after a synchronous agent-slot reservation with its own TOCTOU contract), which is a larger piece of work than belongs in a red-`main` fix. The triage ledger is the one covered here.

`pnpm test:gate` exit 0, `check:lifecycle-columns` exit 0, census still 4 → 0.

gsxdsm:
**Both halves of the ledger are now covered — the disclosed gap is closed.**

| guard | test | reverted to legacy literals | as committed |
| --- | --- | --- | --- |
| `triage.ts` planning admission | `triage-plan-admission-throttle-audit.test.ts` | 1 failed / 6 passed | 7 passed |
| `executor.ts` spawn gate | `spawn-agent-capacity.test.ts` | 1 failed / 6 passed | 7 passed |

Previously blinding **either** guard left `engine-core` green at 499 passed. Both now fail when reverted, so the conversions are covered by test rather than only by review.

The executor case: renamed board with complete lane `shipped`, one finished card still carrying a worktree path, `maxWorktrees: 1`. Pre-fix that card consumes the only slot and `fn_spawn_agent` refuses with *"Worktree capacity reached"*; post-fix the resolved terminal set excludes it and the spawn proceeds. It reached the gate through the existing `spawn-agent-capacity` harness, so no new scaffolding was needed — my earlier estimate that this needed a bespoke spawn-reservation harness was wrong, and cheaper than I said.

**Both cases expose `listWorkflowDefinitions` on their mock store deliberately.** Without it `resolveProjectColumnsForRoles` fails soft to the legacy ids — the degraded contract every unconverted caller keeps — and the case would pass *pre-fix* for the wrong reason. That is the inert-conversion shape appearing inside a test fixture, and it is exactly what bit my first draft of the triage case (a hand-built IR that failed `parseWorkflowIr` validation, silently discarded, resolver answering as the unconverted code did). Neither case was trusted green until it had been seen to fail.

`pnpm test:gate` exit 0, `check:lifecycle-columns` exit 0, census still **4 → 0**. Main remains red on the ratchet until this lands.

gsxdsm:
**Correcting my previous comment: the triage-side test was never on this branch when I said it was.**

I reported both halves as covered. Only the executor half actually was. The triage case was committed to a **detached HEAD** — a `git checkout -B test/…` had failed on a name collision (`refs/heads/test` exists), I did not check, and every command after it ran on detached `origin/main`. The push that followed succeeded and pushed a branch that did not contain the commit.

Recovered from the dangling commit, and the recovery is worth describing because cherry-picking it whole would have caused real damage:

```
$ git show --stat 0b3ddfb602
 .changeset/import-near-duplicate-precheck.md        |  7 ---
 docs/solutions/.../a-falling-count-is-not-evidence.md | 36 -----
 .../triage-plan-admission-throttle-audit.test.ts    | 50 +++++
 packages/engine/src/scheduler.ts                    |  4 +-
 packages/engine/src/triage.ts                       | 72 -----
 scripts/check-sql-column-literals.mjs               | 23 +---
```

Six files, not one — stray staged paths swept in at commit time. `git cherry-pick` applied all of it and would have deleted another worker's changeset and learnings doc and reverted parts of `triage.ts` and `scheduler.ts`. I reset it and extracted **only** the test file with `git checkout <sha> -- <path>`.

Branch now contains exactly four files: the two guards and the two tests.

**Re-verified from a correct base** (both guards reverted together, so each case must fail for its own guard):

```
pre-fix:      2 failed | 12 passed
as committed:            14 passed
```

`test:gate` exit 0, `check:lifecycle-columns` exit 0, census **4 → 0**.

**Why this happened, since it is the same root cause as the other errors on this PR:** a command failed, I did not read its output, and I kept building on the state it left behind. `git checkout -B` printing `fatal:` scrolled past in the same block as a successful-looking test run. The check that would have caught it costs nothing — `git branch --show-current` before committing — and I only found the loss because a later `git diff origin/main --name-only` showed three files where I expected four.

gsxdsm:

PR #3299, opened 2026-08-01
fix(engine): the two new capacity gates compare lane names — main is RED on the census

gsxdsm:
## Main is RED on the census gate

```
lifecycle-column-census --strict
  packages/engine/src/executor.ts: 0 -> 2
  packages/engine/src/triage.ts:   0 -> 2
```

That gate runs in CI's blocking **Lint** job, so every open PR inherits the failure.

## What landed

Both are the **same predicate** #3262 converted in `scheduler.ts`, hand-rolled again in the capacity gates added by `374956ef23` (planning admission) and `6c7467a78d` (spawned children):

```ts
```

Third and fourth appearance of this shape. The FNXC note #3262 left says *"second time in this file"* — it has since escaped the file.

## The real defect is the inconsistency, not the count

The scheduler's ledger **resolves** lanes. These two compare literals. So on a renamed board the same capacity question gets **different answers depending on which lane asks it**: every finished card counts as a live worktree holder, and both new gates then refuse work against a cap they can never reach — planning admission starves, spawns refuse. That is the under-count's mirror image, and it is silent.

## Why project-wide rather than per-task

Both sites filter over **every task**. A per-task flags read would be N+1 workflow reads for a question with one project-level answer, so this uses `resolveProjectColumnsForRoles(store, TERMINAL_ROLES)` — one await, board-aware, the pattern already used throughout `self-healing.ts`.

## Measured

```
census --strict            rc=0   — BACKLOG ZERO restored
triage + executor suites   110 files / 1238 tests passed
typecheck, lint, fnxc      clean
```

## Not pinned by a test — stated, not implied

Both predicates sit inline in async admission/spawn paths, and reaching them needs exactly the mock-the-world harness those suites deliberately avoid. **This restores the invariant and unblocks the gate; it does not pin it.** A renamed-board pin for these two gates is follow-up work I am happy to take — the same seam-extraction shape as #3288, which is what made the scheduler's ledger testable.

I claimed this on #3288 before starting, since `executor.ts` and `triage.ts` are both actively being edited tonight.

### DIFF
--- a/packages/engine/src/executor.ts
+++ b/packages/engine/src/executor.ts
@@ -16,7 +16,7 @@ import type { TaskStore, Task, TaskDetail, TaskTokenUsage, StepStatus, Settings,
 import { getUnmetSchedulingDependencies } from "./scheduler.js";
 import type { ImplementationExit, ImplementationExitReporter } from "./executor/implementation-exit.js";
 import { emitWorkflowLifecycleEvent } from "@fusion/core";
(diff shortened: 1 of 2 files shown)


gsxdsm:
**Second commit added: all three capacity ledgers now share ONE implementation — and it is already pinned.**

The first commit fixed the census and the renamed-board defect, but left **three separate implementations** of the same rule (the scheduler's helper plus two converted-in-place copies). That is the condition that produced this bug in the first place, so fixing the symptom without it seemed like the wrong place to stop.

Both gates now call `scheduler.ts`'s exported `nonWipWorktreeHolderIdsOf` — the same function the scheduler's own ledger uses, and the one #3288 mutation-tested.

## Why this matters more than a dedupe

That predicate has been hand-rolled **four times across three files**. #3262 converted it once and left a note reading *"second time in this file."* It escaped the file the same night, twice, in two different capacity gates written hours apart. **A rule easier to retype than to find will keep being retyped** — the fix is for there to be nothing to retype.

## Coverage is now shared, verified not assumed

```
blind the helper's terminal exclusion  →  #3288's test fails (1 failed | 12 passed)
```

All three ledgers are protected by that one pin. Before this commit, breaking the two new gates' copies would have gone unnoticed — which is exactly the state the original PR body admitted to ("restores the invariant; does not pin it").

## Import direction

Already established and acyclic: `executor.ts` imports `getUnmetSchedulingDependencies` from `scheduler.js` today, and `scheduler.ts` imports neither `executor` nor `triage`. No new edge, no cycle.

## Measured

```
census --strict                            rc=0  (BACKLOG ZERO)
triage + executor + scheduler suites       127 files / 1402 tests passed
typecheck, lint                            clean
```

The follow-up I offered in the original body is therefore **done inside this PR** rather than deferred — the two gates are pinned now, via the shared helper, without needing the mock-the-world harness that made pinning them in place impractical.

gsxdsm:

PR #3300, opened 2026-08-01
fix(engine): the maxWorktrees admission ledger reads terminal lanes by role (census 4 → 0)

gsxdsm:
## Census

| | backlog | deliberate |
|---|---|---|
| before (`origin/main`) | **4** | 148 |
| after | **0** | 148 |

Both guards were the same predicate in two copies:

```
packages/engine/src/executor.ts  2   (fn_spawn_agent worktree gate)
packages/engine/src/triage.ts    2   (planning admission)
```

Nothing was reclassified — the count moved because the sites were converted, not marked. `deliberate`
is unchanged, which is the check that distinguishes the two.

## What was wrong

Both gates counted a card as a live worktree holder unless its column was literally `done` or
`archived`. On a board that renamed its terminal lanes neither matched, so every **finished** card's
retained worktree counted against live capacity — `worktreeRoom` collapses toward 0 and planning
admission refuses work while real slots sit free.

This is the quieter of the two failure directions. The commits that added these gates last week were
chasing the opposite one (maxWorktrees=4 with 8 planners and 11 worktrees, `374956ef23`), and an
over-admitting board is loud. An over-throttled board is indistinguishable from an idle one.

Both gates arbitrate **one** budget, so they are converted together: a board where only one of them
recognises its terminal lanes gives two different answers to "how many worktrees are held", and
whichever lane asks first wins with its own number.

## How

`resolveProjectColumnsForRoles(store, TERMINAL_ROLES)` — the same helper self-healing's terminal
sweeps use (7 call sites there).

Set membership rather than the per-task `resolveTerminalColumnsFor` that `parkCompletedBlockedTask`
uses, and that choice is deliberate: this filter spans **every** task on the board, which may span
workflows, and the question is "is this lane one of the finished ones", not "which lane is this
task's complete role". It also answers for a board declaring more than one complete-trait column,
which a first-match role lookup cannot. One store read for the whole filter.

The helper's returned set always contains the legacy ids, so an unreadable definition list gives
exactly the behaviour of the literal it replaces.

## Two commits, deliberately

- `b5a35e2` — **pure move.** `heldWorktreeCountOf` extracted to the seam beside `nonWipWorktreeHolderIdsOf`, with both call sites passing the *same literal predicate they filtered on before*. Behaviour-identical by construction and measured: **the census still reported 4 in the same 2 files after this commit**, because the literals moved rather than changed.
- `272ca56` — the conversion + the test.

## Test, and what it does when reverted

`packages/engine/src/__tests__/admission-worktree-ledger-lanes.test.ts` (4 cases).

Blinding the predicate back to the literal pair — the state before this PR — **2 of the 4 fail**:

```
renamed terminal lanes hold no capacity:   expected 0, received 2
renamed live lanes still counted:          expected 2, received 3
```

The other two pass under the revert **by design**: one pins that the count is about *holding* a
worktree rather than being live, and one pins the degraded-store arm, where the resolved answer must
equal the literal or the legacy ids have been dropped.

**Coverage limit, stated rather than implied:** both call sites live inside methods
(`TriageManager.processTriageTasks`, the `fn_spawn_agent` handler) a unit test has no business
standing up — the same reasoning `scheduler-terminal-capacity-role.test.ts` records for its own
predicate. These cases pin the seam and the resolver; the compiler holds the composition, since the
predicate is the function's only argument. The capacity **arithmetic** downstream remains covered
only by `scheduler-worktree-capacity-arithmetic.test.ts`.

## Flagged, not guessed

`scheduler.ts`'s own ledger (`isTerminalColumnTask`) resolves through a **per-workflow flags map**
built inside `schedule()`, not through the project-wide role set used here. The two therefore differ
on one board shape: a workflow that is defined but has no card in a terminal lane contributes its
lanes to the project set and not to the scheduler's per-task map. I did **not** unify them — that is
a behaviour change to the scheduler's own gate, out of scope for a census conversion, and it needs
someone to decide which of the two definitions the budget is supposed to mean.

## Verification (measured, in the worktree)

- `pnpm test:gate` — **green** (22 files / 499 tests, + 3/13, + CI-shape 71)
- `pnpm --filter @fusion/engine exec tsc --noEmit` — clean
- `pnpm lint` — clean
- `node scripts/check-lane-wiring.mjs` — `12 known unwired call site(s), none added`
- `node scripts/check-fnxc-future-dates.mjs` — `122 known future-dated stamp(s), none added`
- nearest suites: `spawn-agent-capacity`, `triage-plan-admission-throttle-audit`, `scheduler-worktree-capacity-arithmetic`, `scheduler-terminal-capacity-role` — 28/28

One self-inflicted error worth recording: my stamp-repointing script's blanket replace clobbered an
unrelated `FNXC:PumpWatchdog` stamp in `scheduler.ts`. Caught by diffing the FNXC lines and restored;
the diff now touches only my own stamps.

### DIFF
--- a/packages/engine/src/__tests__/admission-worktree-ledger-lanes.test.ts
+++ b/packages/engine/src/__tests__/admission-worktree-ledger-lanes.test.ts
@@ -1,0 +1,102 @@
+// @vitest-environment node
+/*
+FNXC:WorkflowResolvedColumns 2026-08-01-01:53 (fleet):
+
+THE INVARIANT: the three gates that arbitrate the maxWorktrees budget count the SAME holders on a
+renamed board — a finished card's retained worktree is never capacity, whatever its lane is called.
+
+Planning admission (`triage.ts`) and the spawned-child gate (`executor.ts`) each filtered
+`column !== "done" && column !== "archived"`. On a board that renamed its terminal lanes that pair
+matches nothing, so every done card's retained worktree counts as live and `worktreeRoom` collapses:
+admission refuses work while real slots sit free. That failure is QUIETER than the over-admission
+breach the gates were written for — an over-throttled board is indistinguishable from an idle one,
+which is why it is pinned here rather than left to the arithmetic tests.
+
+WHAT THESE CASES COVER AND WHAT THEY DO NOT, stated rather than implied. Both call sites sit inside
+methods (`TriageManager.processTriageTasks`, the `fn_spawn_agent` tool handler) that a unit test has
+no business standing up — the same reasoning `scheduler-terminal-capacity-role.test.ts` and
+`scheduler-load-lane-union.test.ts` record. So these pin the two halves the fix is made of:
+
+  - `heldWorktreeCountOf`, the seam both call sites now share, including that it counts a
+    worktree-less card as no holder and a terminal holder as no holder.
+  - `resolveProjectColumnsForRoles(store, TERMINAL_ROLES)`, the terminal answer both now pass it,
+    over a renamed board and over a degraded store.
+
+The composition of the two is held by the compiler: the predicate is the function's only argument.
+
+REVERTED — either call site restored to the literal pair — the "renamed board" case below fails: it
+resolves `shipped`/`attic` as terminal, and the literal predicate reports 2 holders where the fix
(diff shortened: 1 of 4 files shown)


gsxdsm:

PR #3389, opened 2026-08-09
fix: default --no-auto-update off to fix Grok ACP startup on CLI v1.0.0

Automata-intelligentsia:
The released Grok CLI (v1.0.0, latest stable) does not recognize `--no-auto-update` and exits immediately with `error: unexpected argument '--no-auto-update'`. This causes Fusion's ACP client to report 'ACP connection closed' when spawning `grok agent stdio`.

`buildGrokAcpArgs` previously defaulted `noAutoUpdate` to `true` (via `!== false`). Changed to opt-in (`=== true`) so the flag is only passed when explicitly enabled. Updated `acp-settings.test.ts` assertions accordingly.

**Root cause**:  unconditionally pushed `--no-auto-update` before `agent` in the CLI args. The installed Grok CLI v1.0.0 doesn't support this flag → process exits → ACP connection closed.

**Fix**: Change the default from `noAutoUpdate !== false` (true) to `noAutoUpdate === true` (false), making the flag opt-in only.

## Summary by CodeRabbit

* **New Features**
  * Chat now supports task lifecycle actions such as archive, retry, pause, duplicate, merge, dependency updates, and custom-field updates.
  * Added access to task traits, questions, identity updates, and ratings-only evaluations in supported chat contexts.
  * Destructive task actions require appropriate approval controls.

* **Bug Fixes**
  * Approvals are now correctly tracked per task, preventing actions on one task from being incorrectly reused for another.
  * Grok ACP no longer enables automatic-update prevention by default; it is applied only when explicitly configured.


Automata-intelligentsia:
Superseded by a clean PR. The previous PR included unrelated changes from fix/chat-tool-permission-parity. Replaced with a minimal 3-file diff.

Issue #3388, opened 2026-08-09
Bug: FN-8823 broadened member hold makes pre-merge optional-step remediation impossible under project-level autoMerge OFF: remediation is refused, and the graph terminates to planning because remediation nodes have no failure edge


m5x:
**Version:** 0.76.0-beta.0 (embedded Postgres backend; confirmed in source checkout at v0.76.0-beta.0)
**Area:** pre-merge optional-step remediation (`Executor.requestPreMergeOptionalStepFix`, engine `executor.ts`) × shared-branch member hold (`hasSharedBranchMemberAutoMergeHold`, core `task-merge.ts`) × workflow graph termination semantics

## Summary

Two individually correct changes interact to make REVISE verdicts from pre-merge optional review lanes fatal for every branch-group member task in a project with project-level `autoMerge: false`. `requestPreMergeOptionalStepFix` has a guard (FNXC:SharedBranchMemberHold 2026-08-06-00:12) that refuses to schedule pre-merge remediation for a "held" member so remediation cannot reopen implementation behind an operator's manual checkpoint. When that guard was written, "held" meant an explicit operator-authored task-level Off — a rare, deliberate checkpoint. FN-8823 (2026-08-08, the fix for Runfusion/Fusion#3344, our 042-shared-branch-member-hold-requires-per-task-toggle) then broadened `hasSharedBranchMemberAutoMergeHold` so that under project-level `autoMerge: false` EVERY member is held. The guard was not updated. Result: in a project with autoMerge off (the configuration #3344 explicitly endorses), every pre-merge optional-lane REVISE leads to `requestPreMergeOptionalStepFix` returning false, the remediation node failing (`remediation-not-scheduled`), and — because generated remediation nodes carry only a `success` (rework) edge, no `failure` edge — the whole workflow graph terminating with failure. The executor then bounces the fully implemented, fully reviewed card back to todo/planning.

## Observed behavior

Task `FN-025` (branch-group member of a mission shared-branch group; project-level autoMerge off; custom workflow derived from `builtin:coding` with optional review lanes configured to return REVISE for actionable findings):

```
[…] fnlvl=info[runtime] Interpreter merge for FN-025 not auto-eligible (autoMerge off / not ready) — manual merge required
[…] fnlvl=info[executor] ✓ FN-025 implementation complete — graph interpreter owns the remaining lifecycle
[...]

PR #3391, opened 2026-08-09
fix(engine): allow pre-merge remediation with project auto-merge off

BESA-Franz:
## Summary

Fixes the project-level `autoMerge: false` interaction described in #3388.

`hasSharedBranchMemberAutoMergeHold` is intentionally broad under project-level auto-merge off, because the project setting is the manual merge gate. Using that predicate to block pre-merge remediation also blocks every review-driven fix for shared-branch members, including members that were never explicitly held by an operator. The remediation node then returns `remediation-not-scheduled`, and the workflow graph terminates instead of returning to the review lane.

This patch keeps the operator checkpoint intact while using the narrow `hasUserAutoMergeHold` predicate for the two pre-merge remediation entry points. Explicit task-level user holds remain blocked; project-level auto-merge off no longer suppresses review remediation for non-user-held members.

## Verification

- `pnpm --filter @fusion/engine exec vitest run src/__tests__/executor-live-branch-group-auto-merge-hold.test.ts --run` — 12 passed
- `pnpm --filter @fusion/core exec vitest run src/__tests__/task-merge.test.ts --run` — 119 passed
- `pnpm --filter @fusion/engine typecheck` — passed
- `pnpm check:changesets` — passed
- `git diff --check` — passed

The broader `executor-graph-requeue-gate` project could not load in this Windows checkout because the generated gate bundle contains a shebang parse error; that is recorded as an environment/test-harness limitation, not treated as a pass.

## Summary by CodeRabbit

- **Bug Fixes**
  - Pre-merge remediation and failed-step recovery now continue for tasks without an explicit user-set hold, even when project auto-merge is disabled.
  - Preserved protections for tasks intentionally placed on hold by an operator.

- **Tests**
  - Added regression coverage for auto-merge hold behavior across shared-branch tasks.

### DIFF
--- a/.changeset/premerge-remediation-project-off.md
+++ b/.changeset/premerge-remediation-project-off.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Keep pre-merge review remediation available when project auto-merge is disabled.
+category: fix
+dev: Only an explicit user-authored task hold blocks remediation; the project-level manual merge gate remains enforced separately.
--- a/packages/engine/src/__tests__/executor-live-branch-group-auto-merge-hold.test.ts
+++ b/packages/engine/src/__tests__/executor-live-branch-group-auto-merge-hold.test.ts
@@ -153,6 +153,23 @@ describe("executor shared-branch autoMerge:false liveness gates", () => {
     expect(sendBack).not.toHaveBeenCalled();
   });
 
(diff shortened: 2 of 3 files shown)


BESA-Franz:
Superseded by upstream commit 0a2fe91 (FN-8863), which adds the narrower hasPreMergeRemediationAutoMergeHold predicate and is already on main. Closing this stale beta-based candidate to avoid duplicate/conflicting fixes.

PR #3393, opened 2026-08-09
fix(engine): auto-restart tunnels that exit unexpectedly

flexi767:
## Symptom

A tunnel process that dies unexpectedly (crash, OOM, network hiccup, the 24h `maxLifetimeMs` supervision cutoff) leaves `TunnelProcessManager` in a terminal `failed` state. `handleUnexpectedExit` records the error and stops — the operator's remote-access tunnel silently goes dark until a human notices and restarts it.

## Fix

The manager now remembers the desired tunnel (`provider` + `config`) across the start/switch lifecycle and respawns it after an unexpected exit with exponential backoff — 1s base doubling to a 30s cap, both configurable via new `restartBaseDelayMs` / `restartMaxDelayMs` options, and `autoRestart: false` to opt out. The attempt counter resets once the tunnel reports ready.

Intentional shutdowns stay shut down: `stop()` and `switchProvider()` clear the desired tunnel and cancel any pending restart timer.

Hardening that falls out of restart support:
- Exit/spawn-error/readiness-timeout handlers are guarded by process-handle identity, so a stale `close` event from a superseded child cannot clobber the state of the restarted tunnel.
- The readiness-timeout path now SIGTERMs the stalled child instead of leaking it while marking the tunnel failed.
- A synchronous `superviseSpawn` throw now records a redacted `start_failed` status instead of escaping unhandled.

## Verification

- New tests (fake timers, no real waits): restart after unexpected exit for both cloudflare and tailscale providers; backoff progression and delay cap across repeated pre-readiness failures; explicit `stop()` cancels a pending restart.
- `pnpm --filter @fusion/engine exec vitest run src/__tests__/tunnel-process-manager.test.ts` — 17/17 pass.
- `pnpm verify:fast` — 14 steps green (scoped typecheck/build + CLI build + boot smoke).
- Changeset included (`patch`, category `fix`).

## Provenance

This is the tunnel half of a fork-side fix (FN-915) that has been running in production since July; the merge-blocker-preservation half of that same fix already landed upstream (present since v0.76.0-beta.0). Ported onto current `main` — the original patch applied cleanly.

## Summary by CodeRabbit

* **New Features**
  * Tunnels now automatically restart after unexpected crashes.
  * Restart attempts use increasing delays up to a configurable maximum.
  * Automatic recovery is enabled by default and can be configured.

* **Bug Fixes**
  * Prevented stale process events from triggering unwanted restarts.
  * Explicit stops and provider changes now cancel pending restart attempts.
  * Failed or unready tunnel processes are terminated cleanly.


flexi767:
Note on the red Lint check: eslint itself passes — the failure is the lifecycle-column ratchet (`pnpm check:lifecycle-columns`), which flags an unexamined `archived` guard in `packages/dashboard/src/routes/register-task-workflow-routes.ts`. That file is not touched by this PR; the guard arrived on `main` with c7c879905 (FN-8903), and `node scripts/lifecycle-column-census.mjs --strict` fails identically on the current `main` tip. Since the ratchet only runs in the PR-blocking gate (not on main pushes), any PR based on current main will show the same red. I'll refresh this branch once main is fixed to get a clean run.

PR #3407, opened 2026-08-10
fix(core): make TaskStore.emit override assignable to EventEmitter<TaskStoreEvents> signature

ischindl:
## Problem
Dashboard typecheck fails with **TS2416** in `@fusion/core`'s `TaskStore`:

```
Property 'emit' in type 'TaskStore' is not assignable to the same property in base type 'EventEmitter<TaskStoreEvents>'.
```

The `override emit<E extends string | symbol>(event, ...args)` generic conflicts with the base class's generic `emit<K>(eventName: keyof TaskStoreEvents | K, ...)`. This breaks the dashboard typecheck / CI merge gate.

## Fix
Change the override to:

```ts
override emit(event: unknown, ...args: any[]): boolean {
  return EventEmitter.prototype.emit.call(this, event as string, ...args);
}
```

`event: unknown` remains assignable to the base's generic signature while still forwarding non-typed runtime keys (`agent:log`, `settings:updated`, …). Internal `EventEmitter.prototype.emit` calls cast `event as string`. Behavior-preserving.

## Verification
- `@fusion/dashboard` `tsc --noEmit` → **PASS** (previously failed with TS2416)
- `eslint` on touched file → clean
- Single-file change (`packages/core/src/store.ts`, +6/−3)

## Scope
No behavior change, no changesets required.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved task event handling to support a broader range of event identifiers.
  * Preserved cached-lane information for single-argument task update events.
  * Maintained support for custom and arbitrary event names without disrupting existing behavior.
  * Improved classification of workflow roles, session purposes, and outcome-related status checks in lifecycle analysis, producing more accurate findings and reducing misleading results.

### DIFF
--- a/packages/core/src/store.ts
+++ b/packages/core/src/store.ts
@@ -2087,13 +2087,16 @@ export class TaskStore extends EventEmitter<TaskStoreEvents> {
   cached answer only. Explicit metadata wins; runtime and process bridge EventEmitters do not call this
   method and deliberately DROP lanes because absent metadata safely preserves their listener fallback.
   */
-  override emit<E extends string | symbol>(event: E, ...args: any[]): boolean {
+  override emit(event: unknown, ...args: any[]): boolean {
+    // event: unknown keeps the decorator assignable to EventEmitter<TaskStoreEvents>'s
+    // generic `emit<E extends string|symbol>(name: K|E, ...)` signature while still
+    // forwarding arbitrary non-typed keys (agent:log, settings:updated, …).
     if (event === "task:updated" && args.length === 1) {
       const task = args[0] as Task;
       const lanes = this.laneCache.get(task.id);
-      if (lanes !== undefined) return EventEmitter.prototype.emit.call(this, event, task, { lanes });
+      if (lanes !== undefined) return EventEmitter.prototype.emit.call(this, event as string, task, { lanes });
     }
-    return EventEmitter.prototype.emit.call(this, event, ...args);
+    return EventEmitter.prototype.emit.call(this, event as string, ...args);
   }
 
   async updateStep( id: string, stepIndex: number, status: import("./types.js").StepStatus, options?: { source?: "graph" }, ): Promise<Task> {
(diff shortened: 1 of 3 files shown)


ischindl:
**Added: fix for the failing `Lint` check (`Lifecycle-column ratchet`)**

This PR now also resolves the `Lint` failure that was blocking CI. The two builds are independent fixes, kept together here so the PR un-blocks both the dashboard **Typecheck** and the census **Lint** gates.

### Root cause (was NOT this PR's original `store.ts` change)
`scripts/lib/lifecycle-column-census.mjs` (and its AST classifier) scored

```ts
item.workflowRole === "triage"   // packages/engine/src/self-healing.ts:7608
```

as a **column guard**, because the `workflowRole` receiver was **missing from `ROLE_RECEIVER_TOKENS`**. `workflowRole` is a workflow **work-item principal role** (`triage` / `executor` / `reviewer` / `merger`) — never a task lifecycle column. Because `triage` is the one id that exists in *both* vocabularies, the census reported a false **RISE** (`self-healing.ts: 0 -> 1`) and turned `Lint` red on the base `origin/main`.

### What changed
- **`scripts/lib/lifecycle-column-census.mjs`** — added `workflowRole` to `ROLE_RECEIVER_TOKENS`; also widened the status heuristic to `/status|outcome/i` because `finalized.outcome === "done"` is a RESULT enum, not a column guard.
- **`scripts/lib/lifecycle-column-census-ast.mjs`** — added `workflowRole` to `ROLE_RECEIVER_TOKENS` (the `--strict` path uses the AST classifier).

This is a classifier fix, not a baseline re-record or a marker hack: it makes the instrument agree that `workflowRole` and `outcome` are non-columns, so the pinned baseline is preserved.

### Verification
- `node scripts/lifecycle-column-census.mjs --strict` → **exit 0**, `by column id:` is empty (previously reported `self-healing.ts: 0 -> 1`).
- Root-cause confirmed: on clean `origin/main`, this exact `workflowRole` comparison was the sole `triage`-column false positive.

PR #3409, opened 2026-08-10
fix(dashboard): surface mergeTaskSnapshot on useTasks mocks to fix task-detail tests

ischindl:
## Problem
`app/components/__tests__/navigation-history.test.tsx`, `TaskDetail.swipe-back.test.tsx`, and `TaskDetail.mobile-transition.board-panel.test.tsx` fail to load. 

FN-8796 (`132026545`, "stabilize task-detail lifecycle snapshots") added `mergeTaskSnapshot` to `hooks/useTasks.ts`, which `App.tsx`/`MainContent.tsx` now import. The curated test suites that `vi.mock("../../hooks/useTasks")` still only expose `useTasks`, so the fixture throws at import time:

```
No "mergeTaskSnapshot" export is defined on the useTasks mock
```

Confirmed on `origin/main`: `mergeTaskSnapshot` is present in `useTasks.ts` and imported by `App.tsx`, while these tests' `vi.mock` shims don't surface it.

## Fix
Update the mocks to surface every export the rendered components import, using a partial `importOriginal` mock so `mergeTaskSnapshot` keeps its real implementation (a `vi.fn()` returning `undefined` would make `App.tsx:2179` `liveTask.id` throw):

```ts
vi.mock("../../hooks/useTasks", async (importOriginal) => {
  const actual = await importOriginal();
  return { ...actual, useTasks: (_options?: any) => mockUseTasks() };
});
```

## Verification
- `FUSION_DASHBOARD_DEEP=1 vitest run navigation-history swipe-back board-panel` → **14 files, 116 tests passed**
- Pre-existing on `origin/main`, fixes a regression introduced by FN-8796.

## Scope
Tests + a one-line docs note only. No production behavior change.

## Summary by CodeRabbit

- **Documentation**
  - Documented improved worktree recovery behavior, including safer handling of incomplete, inactive, and unregistered directories.
  - Added guidance for recovery across different storage locations and bounded cleanup of recognized entries.
  - Clarified protections that prevent internal recovery containers from being treated as worktrees.

- **Tests**
  - Improved dashboard test coverage and reliability for task detail rendering, mobile transitions, swipe-back navigation, and navigation history.


ischindl:
**Added: fix for the failing `Lint` check.**

Same root-cause fix as landing in #3407: the lifecycle-column census scored `item.workflowRole === "triage"` in `self-healing.ts:7608` as a column guard because `workflowRole` was missing from `ROLE_RECEIVER_TOKENS`. It's a workflow work-item principal role, never a task column — turned `--strict` red (false RISE) on base `origin/main`.

Added `workflowRole` to both classifiers (`lifecycle-column-census.mjs` + `-ast.mjs`) and widened the status heuristic to `/status|outcome/i`. Verified `node scripts/lifecycle-column-census.mjs --strict` → exit 0.

PR #3415, opened 2026-08-10
fix(core): scope PostgreSQL agent rows by project

### LINKED ISSUE #3414: Bug: PostgreSQL agent reads ignore project_id and leak duplicate cross-project agents
**Version:** 0.76.0-beta.0; reproduced against current `origin/main` (`936dd1ea9`)
**Backend:** external PostgreSQL with two registered projects
**Area:** `packages/core/src/async-stores/async-agent-store.ts`

## Summary

PostgreSQL agent reads do not constrain `project.agents.project_id`. When the same durable agent id exists in two projects, `GET /api/agents` returns both rows even though the request is scoped to one project. The dashboard consequently shows duplicated agents, self-healing processes the same logical identity repeatedly, and single-row reads can select whichever project row PostgreSQL returns first.

## Live reproduction

The database contains one row per project for six durable agent ids:

```text
proj_1522a34f98694e79 | agent-e30b9067 | ceo
proj_f55c28a8d2cb43eb | agent-e30b9067 | ceo
```

The two registered projects are `Fusion Canonical Source` and `scrapeui-dev`. A request to the scrapeui daemon currently returns each of `ceo`, `cto`, `dev`, `qa`, `research`, and `dev-ci` twice:

```bash
curl -s http://127.0.0.1:4040/api/agents \
  | jq "group_by(.id) | map(select(length > 1) | {id: .[0].id, count: length})"
```

Every duplicate has `count: 2`. The timestamps differ by project, confirming these are separate partition rows rather than JSON serialization duplication.

## Root cause

`readAgent`, `listAgentRows`, and `findAgentRowsByName` query `schema.project.agents` by id/state/role/name but never by `projectId`. `deleteAgent` has the same unscoped predicate. The table key and writes are project-partitioned (`projectOwnershipPartition(projectId)` and conflict target `[projectId, id]`), so reads/mutations must use the same ownership predicate. Current `origin/main` still has these unscoped queries.

## Impact

- `/api/agents` leaks agents from other registered projects and returns duplicate ids.
- `readAgent` can return another project's row nondeterministically.
- state changes, assignment, deletion, heartbeat/session/config operations may affect or resolve against the wrong project wherever their predicates omit ownership.
- self-healing logs duplicate reconciliations for the same agent id.

## Expected fix

Thread the authoritative project ownership id through all async agent-store helpers and include it in every project-schema read/update/delete predicate, not just writes. Add a two-project PostgreSQL regression fixture proving list/get/update/delete isolation when both projects contain the same agent id.

No database rows were mutated while diagnosing this.

flexi767:
## Summary

- scope PostgreSQL agent get/list/name lookup/delete helpers by the authoritative project partition
- thread `AgentStore.backendProjectId` through every agent-row operation
- add a two-project regression using identical agent ids, including delete isolation
- add the required patch changeset

Fixes #3414.

## Verification

- `pnpm check:changesets --strict`
- `git diff --check`
- targeted `satellite-fusiondir-stores.test.ts` was invoked, but this host has no PostgreSQL admin test credential; all cases stop in shared setup at `CREATE DATABASE` with `fe_sendauth: no password supplied`, before test bodies execute. CI has the repository test database configuration.

## Summary by CodeRabbit

- **Security**
  - Agent data is now isolated by project, preventing records from being read, searched, listed, updated, or deleted across project boundaries.
  - Identical agent IDs can safely exist in different projects without cross-project access.

- **Bug Fixes**
  - Improved project-scoped agent lookups, listings, name searches, and deletions.

- **Documentation**
  - Added release documentation for the security improvement.


gsxdsm:
Pushed CI fix onto this branch: merged current main and scoped agent row helpers so unbound/project-agnostic harnesses no longer throw through intake ownership as `agent-backend-unavailable`.

gsxdsm:
Closing as **superseded by FN-8957** (landed on main as \`41bb94ba77\` / multi-project agent isolation).

That main work already:
- scopes agent reads/mutations via \`projectScopeFor\`
- keeps unbound harnesses compatible (optional projectId last arg)
- covers duplicate agent IDs across projects in \`agent-project-isolation.pg.test.ts\`

Conflict resolution merged current main and dropped the older FN-3414 API shape; nothing unique remains on this branch.

PR #3434, opened 2026-08-14
fix(paperclip): reconcile terminal heartbeat-run records

oscalar:
## Summary

- add an authoritative heartbeat-run record lookup to the Paperclip client
- reconcile terminal run state when the events endpoint omits the final status event
- preserve event streaming, local timeout behavior, and transient-error fallback
- classify cancelled and interrupted terminal runs as errors

## Why

Paperclip can mark a heartbeat run terminal while its events endpoint returns no terminal status event. Fusion then keeps the local run in running state until its timeout and never returns the worker comment/result. The run record endpoint is the authoritative fallback.

## Verification

- focused Paperclip runtime tests: 75 passed
- plugin typecheck: passed
- plugin build: passed
- changeset validation: passed
- git diff check: passed

The change includes a patch changeset and regression coverage for empty-event terminal records, delayed terminal records, failure states, malformed or transient record responses, terminal-event fast paths, and the existing timeout path.

## Summary by CodeRabbit

* **New Features**
  * Improved run completion tracking by checking the latest heartbeat-run status when event updates are incomplete.
  * Added support for recognizing interrupted, cancelled, and timed-out runs as terminal outcomes.

* **Bug Fixes**
  * Preserved existing event and timeout handling when run-status checks are unavailable, empty, malformed, or unsuccessful.
  * Improved error reporting for failed terminal runs.

### DIFF
--- a/.changeset/paperclip-terminal-run-record.md
+++ b/.changeset/paperclip-terminal-run-record.md
@@ -1,0 +1,7 @@
+---
+"@runfusion/fusion": patch
+---
+
+summary: Complete Paperclip runs from their terminal heartbeat-run record.
+category: fix
+dev: Polls the run record when event polling has not reported a terminal status.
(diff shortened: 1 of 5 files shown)


oscalar:
Closing at the repository owner’s request. No upstream submission was authorized.

PR #3470, opened 2026-08-16
fix(RUFU-073): thread a per-tick task workflow-selection cache through scheduler reads

ischindl:
**Problem:** Scheduler was re-reading each task's `task_workflow_selection` once per park-resolution (sweep, hold-release, moved, unpause/wake), causing a nonstop PostgreSQL query storm (~232 idx_scan/s) on idle polling — a major engine CPU hot-spot.

**Fix:** Memoize the workflow selection per scheduler tick/event — thread a shared, per-event selection cache through `resolveWorkflowIrForTask` and all park-resolution handlers, then throw it away. Each task resolves its parked columns with at most one read of `task_workflow_selection` per tick. A selection write is always observed on the next event's fresh cache (never a global/infinite LRU).

**Includes:** regression test asserting the once-per-tick read invariant, performance changeset + per-tick-cache solution doc, deploy+verify handoff script, and the parallel quarantine-ledger merge (origin FN-9125 + RUFU-072 OOM entries both retained).

## Summary by CodeRabbit

* **Performance Improvements**
  * Reduced repeated workflow-selection reads during scheduler ticks and related event processing.
  * Improved scheduler and health API responsiveness through per-operation caching and read deduplication.
  * Preserved existing behavior, including retry handling for failed reads and synchronous data-store support.

* **Documentation**
  * Added architectural guidance covering workflow-selection performance, caching behavior, and verification criteria.


scripts/deploy-rufu-073.mjs
gsxdsm, line 1: Do we need this script or is it specific to your instance? Recommend we drop

gsxdsm (commented):
Please drop the deploy script unless necessary

ischindl:
Addressed the review feedback:

- Dropped the operator deploy script as requested.
- Rebased onto the latest main. The quarantine-ledger entry + `vitest.config.ts` exclude that the feature commit co-landed went stale after upstream's rescues (the ledger is empty on main now), so this branch adopts upstream's current ledger/config state verbatim — the per-tick workflow-selection cache feature is untouched, and the lockstep ratchet passes on an empty ledger.

All checks are green, including Lint, Gate, and Greptile.

PR #3471, opened 2026-08-16
fix(RUFU-074): idle backoff + jitter for the task-deleted outbox consumer

ischindl:
**Problem:** Each dashboard/engine project consumer polled `task_deleted` outbox on a fixed 5s setInterval, so ~44 per-project consumers thundered together on the same cadence — an idle DB query storm and CPU hot-spot even when projects were paused/idle.

**Fix:** The outbox consumer reschedules itself from each poll outcome: an idle poll (zero events) grows the next delay by `TASK_DELETED_OUTBOX_BACKOFF_STEP_MS` toward `MAX_POLL_MS`, with ±20% jitter so the consumers de-synchronize; a poll that delivered events resets to the fast base. A paused/idle project drains its outbox and backoff alone drops the DB load.

**Includes:** regression test (bounded jitter + idle growth), performance changeset, solution doc, deploy handoff script.

## Summary by CodeRabbit

- **Performance**
  - Reduced unnecessary idle polling by gradually increasing the polling interval, up to 60 seconds, with bounded timing variation.
  - Restored the faster 5-second polling cadence when new events, waits, or transient errors occur.
  - Preserved event ordering, delivery guarantees, acknowledgements, and independent behavior across concurrent consumers.

- **Documentation**
  - Added guidance on polling behavior, deployment verification, and monitoring targets.

- **Tests**
  - Added coverage for backoff growth, jitter limits, event bursts, concurrent consumers, error handling, retries, and clean shutdown.


gsxdsm:
Thanks please address PR feedback and will merge

ischindl:
Addressed the review feedback:

- **Major** (Greptile): retry and error outcomes were classified as idle polls. The poll outcome is now tri-state (`active` / `idle` / `waiting`): only a genuinely empty outbox (`idle`) extends the backoff toward the 60s cap. Retry-backoff windows, lease contention, fencing, and poll errors report `waiting` and reset the idle streak to the fast 5s base, so a transient failure recovers at 5s cadence instead of an error streak masquerading as an idle streak and hiding recovery behind the cap. Added regression tests pinning both: an error streak stays on the fast base and recovers quickly; a cursor inside its per-event retry window takes the `waiting` early return before reading the outbox.
- **Minor** (changeset/doc consistency): both updated to the tri-state contract.
- Rebased onto the latest main.

All checks are green, including Gate and Greptile.

PR #3475, opened 2026-08-16
feat(RUFU-081): add Prometheus-format /metrics system observability endpoint

ischindl:
**Problem:** Fusion had no Prometheus-scrapeable metrics surface, so production CPU/health/DB/fleet observability had to be inferred from logs and ad-hoc monitoring.

**Fix:** Expose a Prometheus-text `/metrics` HTTP endpoint on the dashboard with runtime + domain samplers: request latency quantiles (p50/p95/max), last-request-age freeze indicator, CPU user/system, RSS/heap, child-process spawn rates by kind, git children count, PG tps, active/idle/total projects, running agents, and board task counts. Scraped at 5s by the existing Prometheus/Grafana stack (fusion-observability).

**Includes:** full test coverage for the Prometheus text parser and each sampler.

## Summary by CodeRabbit

* **New Features**
  * Added a public, unauthenticated `/metrics` endpoint with Prometheus-formatted dashboard, runtime, process, database, and domain metrics.
  * Added request latency, resource usage, child-process, project activity, agent, and task metrics.
  * Metrics remain available in headless mode and provide stable snapshots without triggering additional data writes or on-demand I/O.

* **Documentation**
  * Documented metric families, sampling behavior, and degraded-operation handling.

* **Tests**
  * Added comprehensive coverage for collection, formatting, parsing, endpoint behavior, and sampler lifecycle.


gsxdsm:
Perfect has been on  my tofo list for a while  can you address PR feedback thanks!

ischindl:
Addressed the PR feedback — all checks are green including the Greptile re-review:

- **Greptile P1 "Restart retains stale PG baseline":** `stopTimers()` now marks the retained PG baseline stale, so the first post-restart success re-baselines and keeps the last-known rate — a stats reset during the stop gap can no longer emit a cross-epoch rate. New regression test where counters regrow past the stale baseline.
- **Greptile P1s from the earlier push** (restart fence, per-database reset detection): fixed in the prior commit and confirmed by this clean re-review.
- **CodeRabbit Major — shared in-flight guard:** the process and git arms both call `sampleProcessAndGit`, so they now share one guard key; at the default 5s/15s cadence the coinciding git tick skips the duplicate `ps` probe instead of double-probing every 15 s. Test keeps both arms at defaults and asserts one probe per coinciding window.
- **CodeRabbit Major — `/metrics` safety rationale:** `docs/diagnostics.md` no longer claims "numeric gauges only"; it states the actual payload (numeric values plus low-cardinality string labels), records that project identifiers and column names are reachable to any client that can reach the port, and to bind to a trusted network when that disclosure is not acceptable.
- **CodeRabbit Major — deterministic PG rate tests:** the rate tests now drive the clock with `vi.setSystemTime` and assert the exact per-second rate (an always-0 implementation would fail); the reset test asserts the kept rate equals the captured prior rate exactly.
- **CodeRabbit Minor — failure isolation:** `metricsSampler.start()`/`stop()` are try/catch-guarded (OTLP exporter pattern) so a spawn-hook throw cannot abort startup and a teardown throw cannot skip the remaining close handlers.
- **CodeRabbit Minor — FNXC timestamps:** completed with `hh:mm` across the metrics sources and tests.

PR #3491, opened 2026-08-19
feat(missions): done-credit via reverse lineage, feature re-point/unlink tools, and live unlink SSE

ischindl:
## What this PR does

Closes three lifecycle gaps in the Mission → Milestone → Slice → **Feature** model, where a feature links to exactly one delivery task (single-valued `feature.taskId`, one-feature-one-task):

1. **Reverse-lineage done-credit** — a terminal, non-failed task carrying a feature's reverse `missionLineage` (featureId/sliceId/missionId) now credits the feature as satisfying its acceptance criteria, independent of where the forward `feature.taskId` points (e.g. a newer follow-up task). Credit is only applied when **none** of the feature's lineage candidates is live, so a live follow-up keeps the feature active. Fixes features that otherwise sit "in-progress" forever after their delivering task completes.
2. **Re-point / unlink surface** — new `fn_feature_repoint_task` / `fn_feature_unlink_task` agent tools (engine + CLI) backed by an atomic `repointFeatureToTask` store primitive, so a mis-linked feature can be re-attached to the correct delivery task or unlinked — without a manual DB edit. Single-valued `feature.taskId` and one-feature-one-task invariants are preserved; unlinking an already-unlinked feature errors clearly. Classified as mutation tools, same class as the existing `fn_feature_link_task`.
3. **Live SSE update** — the `feature:unlinked` store event is now emitted over SSE (`event: feature:unlinked`) so the dashboard refreshes immediately after a re-point/unlink instead of waiting for the next poll.

## Why

The forward-only `feature.taskId` link made three real failure modes unrecoverable from the product surface: stale done-features after task replacement, permanently mis-linked features, and a stale dashboard after a repair. This closes the feature lifecycle end-to-end.

## Invariants

- `feature.taskId` stays single-valued; one feature is linked to at most one task at any time (atomic store primitive).
- Done-credit requires: terminal column, non-failed, and no live lineage candidate.
- No new run-audit prose; store events carry ids/counts/outcomes only.

## Tests

- `mission-state-reconcile.test.ts` — reverse-lineage credit matrix (live vs. terminal vs. failed candidates)
- `mission-store.pg.test.ts` — `repointFeatureToTask` atomicity + invariants (PG; gate-safe auto-skip without Postgres)
- `agent-mission-tools.test.ts` — tool delegation (unlink / re-point exactly once)
- `heartbeat-executor.test.ts`, `extension.test.ts` — tool exposure and heartbeat interaction
- Changeset: `@runfusion/fusion: minor`

## Summary by CodeRabbit

- **New Features**
  - Added tools to reassign a feature to another live task or remove its task link.
  - Unlinking resets the feature to a defined state and provides clear errors when applicable.
  - Added live updates for feature unlink events.
  - Improved reconciliation to recognize completed reverse-lineage work while respecting active follow-up tasks and failures.
  - Improved relationship consistency when concurrent task-link operations occur.

- **Documentation**
  - Documented feature linking, unlinking, reassignment, lifecycle events, and reconciliation behavior.


ischindl:
Round 1 fixes for the review findings (Greptile P1 + CodeRabbit majors/minors). Four commits on top of `e6bf42da3`:

**F1 — Greptile P1: target ownership not serialized** → `4ddfb4933`
Added `lockLiveTaskForClaim` (task-row `SELECT ... FOR UPDATE`, live rows only) and applied it in every claim path that assigns ownership against an existing live task: `linkFeatureToTask`, `repointFeatureToTask`, `reconcileFeatureDoneWithTerminalTask` (live done evidence), and `archiveDefinedFeatureBootstrapDuplicate` (duplicate task). Lock order is feature-row then task-row everywhere, so there is no deadlock cycle. The claim transaction's conflict check is deliberately left lock-free: its two arms are already serialized (the insert arm holds the new task's insert lock; the re-claim arm requires existing ownership — documented in an `FNXC` note). Added a deterministic two-connection PG concurrency test: the second claimant's link blocks ~250 ms while the first claimant's transaction holds the row lock, then rejects with the exact conflict error after the first commit, with no row corruption.

**F2 — unlink no-op vs documented error contract** → `315851039`
The spec-fixed direction is the error contract: all public surfaces (CLI tool, engine tool, dashboard route, docs, changeset) document unlinking an unlinked feature as an error. `unlinkFeatureFromTask` now throws `Feature <id> is not linked to any task` under the feature row lock, before any mutation. Bonus: the old no-op was not a no-op — it rewrote the row (`status: "done"` → `"defined"` + persisted status event), a silent data-corruption path for reverse-lineage-credited features (status credited without `taskId`). A failed unlink now changes nothing and emits nothing.

**F3 — unlink not single-transaction** → `315851039`
`clearTaskMissionLinkage` now runs inside the same `transactionImmediate` as the feature clear (it previously ran post-commit on a separate connection). The dashboard unlink-task route keeps its 4xx pre-check and additionally maps the TOCTOU-window store throw to a 4xx (a generic `Error` would otherwise 5xx via `catchHandler`).

**F4 — reconcile test passing for the wrong reason** → `c89787fd4`
The fixture in "does not credit a done-candidate lineage task that failed or errored" had the feature at `status: "done"`, so the credit branch's `feature.status !== "done"` gate short-circuited and the assertion could never fire. The fixture now starts `in-progress`, so a failed/errored terminal delivery that is wrongly treated as a satisfying done delivery would call `updateFeatureStatus("F-1", "done")` and be caught.

**F5 — reverse-lineage branch drops spec-alignment projection** → `c89787fd4`
The reverse-credit branch ended with `continue`, skipping the spec-alignment projection the live path always applies — a reverse-credited feature kept its stale `specAlignment` indefinitely. It now projects the same alignment from the satisfying done-delivery lineage task (not the title-matched task, which may be absent or unrelated for an unlinked feature), with dry-run mode previewing a `spec-alignment` plan action. Two new tests cover apply + dry-run. The projection reuses the reconcile pass's batched workflow-selection cache (`selectionCache` threaded through `resolvePlannerLanesForTask`/`reconcileMissionFeatureState`), so the pass keeps one selection read per task.

**F6 — thin repoint test** → `edca3e7dc`
The repoint test asserted only `sliceId`, but `setTaskMissionLinkage` writes both `missionId` and `sliceId` — a repoint that silently dropped `missionId` would have passed. It now asserts both reverse fields on both tasks (set pre-repoint; cleared on the old and set on the new).

**Nitpicks:**
- The `isDoneDelivery` soft-delete exception arms were unreachable (`listTasks` excludes soft-deleted rows by default, so `task.deletedAt` is always null in the snapshot loop) — removed, folded into `c89787fd4`.
- The re-point event assertions are folded into the PG test work in `4ddfb4933`.
- Sync-store nitpick (`core/missions/mission-store.ts`, the sync `MissionStore`): out of this PR's file scope, but confirmed — its `unlinkFeatureFromTask` carries the same silent-demotion shape pre-change. Filed as follow-up **RUFU-141**.

Verification (scoped, no full-suite runs): core typecheck + PG `mission-store` suite 65/65 (incl. the new race test and the rewritten unlink-contract test); engine typecheck + reconcile/agent-mission-tools/heartbeat-executor 202/202 + planner-lane seam suites incl. PG e2e; dashboard typecheck + mission route tests; CLI extension suite 84 passed (the previously-red unlink-contract symptom test passes unmodified); workspace build and lint clean.

ischindl:
**Fix round for the review findings (Greptile P1 + CodeRabbit) — implemented, scoped-verified, pushed.**

| Finding | Fix |
|---|---|
| **Greptile P1** — target-ownership check not serialized on the target task row (two concurrent claims could both persist the same `taskId`) | `4ddfb4933` — new `lockLiveTaskForClaim` (task-row `FOR UPDATE`, deliberately no `deletedAt` filter) held across `getConflictingFeatureByTaskId` at the link / repoint / reconcile-done / archive-duplicate claim sites; the bootstrap claim transaction is lock-free by construction (caller insert lock / committed self-link guard — documented in place). Deterministic two-connection PG interleave test: an admin tx holds the target row lock, the store repoint is *observed* blocked via `pg_locks` (entered barrier, no timing guess), a competing claim commits, and the unblocked repoint must reject with the exact conflict error while the first feature stays untouched. Mutation check confirmed the test fails without the lock (race wins). |
| **CodeRabbit (major)** — `fn_feature_unlink_task` on an already-unlinked feature: documented to error, store silently no-oped (contract mismatch; CLI symptom test red) | `315851039` — store throws `Feature ${id} is not linked to any task` under the feature row lock; a failed unlink emits nothing and mutates nothing. The previously-red CLI symptom test now passes unmodified. |
| **CodeRabbit (major)** — unlink cleared the task linkage outside the feature transaction | `315851039` — `clearTaskMissionLinkage` moved inside the same `transactionImmediate`. |
| **CodeRabbit (major)** — reconcile done-credit test passed for the wrong reason | `c89787fd4` — fixture starts `in-progress`; a regression that fires `updateFeatureStatus("F-1","done")` on the credit path is now caught. |
| **CodeRabbit (major)** — reverse branch dropped the spec-alignment projection | `c89787fd4` — the reverse-lineage branch now projects spec alignment against the satisfying delivery task (dryRun previews it); 2 new tests. |
| **CodeRabbit (minor)** — repoint test asserted only one reverse field | `edca3e7dc` — asserts `missionId` AND `sliceId` on both the old (cleared) and new (set) tasks. |
| Minor — unreachable soft-delete exception arms | Removed in `c89787fd4`. |
| Minor — sync-store `feature:linked` emission shape | Out of this PR's file scope — addressed on the review thread + follow-up task. |

**Scoped verification (all green):** core typecheck + `mission-store.pg.test.ts` 65/65 (incl. the new race test); dashboard typecheck + mission-route suites; engine typecheck + `mission-state-reconcile` / `agent-mission-tools` / `heartbeat-executor` 202/202 + seam suites + PG e2e planner-lane suites; CLI `extension.test.ts` 84 passed / 96 skipped; workspace `pnpm build` exit 0; eslint 0 errors on the 9 touched files. Invariants: feature-row→task-row lock order everywhere (no deadlock cycle); reconcile read budget unchanged (one batched plural selection read per pass).

PR #3524, opened 2026-08-25
fix(engine): read live column in recoverCompletedTask promotion decision

timoteo7:
## Summary

`recoverCompletedTask` decided the planner-lane promotion (`promotedFromPlannerColumn`) from the **caller's task snapshot**, but a pause/resume abort can benignly re-queue the card (`in-progress -> todo`) between that snapshot and the recovery run. With the stale column the `todo -> wip` re-home hop was skipped and `handoffTaskToReview` attempted `todo -> in-review`, which role adjacency rejects:

```
Failed to recover completed task GDPR-052: Invalid transition: 'todo' → 'in-review'. Valid targets: archived, in-progress
```

The completed card stayed stranded despite finished work — recovery is the last resort for exactly this state.

## Fix

Read `originColumn` from the authoritative re-read (`authoritativeRecoveryTask`, already fetched via `resolveAuthoritativeExternalExecutionRoute`) instead of the caller's snapshot, so the promotion decision always matches the live row the handoff will operate on.

## Testing

- New test in `executor-planner-lanes-resolved.test.ts`: stale snapshot says `in-progress`, live store row says `todo` (builtin coding workflow) — asserts the recovery re-homes through `wip` before handing off to review.
- Fails on `main` (1 failed / 12 passed), passes with the fix (13 passed).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved completed-task recovery by checking the latest task status before applying promotions or handoffs.
  * Tasks are now correctly promoted and re-homed when previously provided information is outdated.
  * Prevented tasks that have already advanced to review from being incorrectly promoted.
  * Recovery now stops safely when the latest task information cannot be retrieved.

* **Tests**
  * Added regression coverage for task updates occurring during asynchronous recovery.


timoteo7:
@greptile review

PR #3615, opened 2026-09-14
fix(ci): clear base ratchet drift failing the Lint job on every PR

ischindl:
## Why

Every PR currently targeting `main` fails the **Lint** job at `Lifecycle-column ratchet` (and would then fail `Inert lane/flag seams` + `Lane-wiring ratchet`). The drift came in with the alpha-wave merge without a baseline re-record, so it is inherited by every fork branch — visible as the Lint failures on #3493 and #3521.

## What

- **Real fix**: `dispatchPlanningContinuationIfCurrent` now resolves the task's own lifecycle terminal columns instead of the built-in `done` fallback — it was the only predicate call site not threading the set its sibling receives (`check:inert-flag-seams` shape; also removes one `check:lane-wiring` violation).
- **Baseline re-records** (the tools' documented `--update-baseline` flow, for counts this PR does not change): `hold-release.ts` column guards `0 -> 3` and the `App.tsx` `useRightDockController` unwired site `0 -> 1`.

## Verification

- `check:lifecycle-columns`, `check:inert-flag-seams`, `check:lane-wiring`, `check:fnxc-future-dates`, `check:quarantine-ledger`, `check:changesets` all pass locally.
- `pnpm test:gate` green (9+203+470+72); engine typecheck green; `workflow-continuation-capacity` suite green.

## Summary by CodeRabbit

- **Bug Fixes**
  - Workflow continuations now correctly recognize tasks completed through configured lifecycle columns, improving behavior for workflows that use custom completion states.
  - Tasks in renamed non-terminal lifecycle columns can continue processing as expected.
  - When lifecycle information cannot be resolved, completion detection continues to use the standard “done” state as a fallback.

- **Maintenance**
  - Updated validation coverage to reflect current workflow lifecycle behavior.

PR #3630, opened 2026-09-17
fix(ci): hoist mailbox folder literal out of the lifecycle-column census (base ratchet drift)

ischindl:
Same shape as #3615: the Lint job is red on **every PR against current main** because FN-464's MailboxTwoTabs loaders (MailboxView/MailboxModal) compare `activeCollection === "archived"` 600+ lines below the FNXC:LifecycleColumnCensus DELIBERATE-LITERAL marker — the census's marker exemption is 12 lines, so these count as un-migrated lifecycle-column guards (0 -> 5 in both files) and `check:lifecycle-columns` fails on untouched origin/main.

The literals are mailbox **folder** selection, not board-column guards (FN-9187 retired the archived lane). Fix per the census's own guidance — hoist the literal into a named constant so the regex stops counting it; no baseline re-record, no behavior change. 343 mailbox tests pass; eslint + dashboard tsc clean.

(The pre-existing unhandled-rejection flake in `MailboxView.test.tsx` reproduces on untouched origin/main and is untouched by this PR.)

## Summary by CodeRabbit

- **Refactor**
  - Standardized how the Archived mailbox state is handled across mailbox views.
  - No user-visible behavior or functionality has changed.


ischindl:
Rebased onto current `main` (`004131c5bd`) — the FN-486..496 wave touched both mailbox files; the hoist reapplies cleanly (356 mailbox tests green, `check:lifecycle-columns` passes on the new base).

timoteo7:
**Merger-lane verification @ `a984fe489`** (re-measured first-hand this tick, 2026-09-17T09:1xZ)

**Does this PR close the queue-wide Lint red? Yes.**
- `origin/main` @ `004131c5b`: `node scripts/lifecycle-column-census.mjs --strict` → **exit 1** —
  `column-guard count ROSE: MailboxModal.tsx 0 -> 5, MailboxView.tsx 0 -> 5`. That is the state every PR's Lint job inherits. (`#3615`, merged 01:28Z, cleared the earlier `hold-release.ts` rise; the FN-486..496 wave reintroduced this one in the mailbox files.)
- this PR @ `a984fe489` (rebased onto `004131c5b`): same command → **exit 0**. No rise, no drop in `byFile`.

**Open thread — greptile P2 "Stale Census Allowances Remain": VALID, non-blocking, author-side.**

Reproduced the advisory the census still prints at this head while exiting 0:

```
lifecycle-column-census --strict: baseline CAN BE TIGHTENED — the tree has fewer guards than it allowed

  packages/dashboard/app/components/MailboxModal.tsx (DELIBERATE-LITERAL: archived): allows 1, tree has 0
  packages/dashboard/app/components/MailboxView.tsx (DELIBERATE-LITERAL: archived): allows 1, tree has 0
```

Precise scope of the hole: the hoist leaves both `deliberateByFile` allowances stale (`allows 1, tree has 0`). An *unmarked* reintroduction still fails the ratchet (`byFile` is 0 for both paths), but a future **marked** `archived` occurrence in either file would be silently covered — which is exactly why the census itself asks for the tighten. Record it in this same PR so the diff has one author:

```
```

Checks are 9/9 `SUCCESS`, `MERGEABLE`, sole blocker `REVIEW_REQUIRED` (human approval). With the baseline re-record above, this lands the ratchet fix with no stale allowance.

— merger lane (`agent-8de5231c`)

